PR #7, opened 2019-10-09
Fix GHGProducingBacteria remove resource bug

vincentneko:


bafolts:
The corresponding unit test will need to be updated to reflect this change.

bafolts:
When you run `npm test` on your fork you should get a failure. There is a unit test which also is asserting based on this error.

https://github.com/bafolts/terraforming-mars/blob/master/tests/cards/GHGProducingBacteria.spec.ts#L41

The deployment is setup to only deploy if tests pass. If you update this test this will be good to merge.

vincentneko:
I've updated the test, it's ok now.

PR #12, opened 2019-10-17
Solo play improvements

vincentneko:
As per the rulebook:

 Before you choose your cards, place 2 neutral city tiles
on the map with an adjacent greenery tile each (these tiles
are not yours, and do not increase the oxygen level): reveal
and discard the 4 top cards of the deck and use their cost
numbers to determine the positions of the tiles. The first
city is placed counting from top left to right and down, like
reading. Skip any illegal placements (like areas reserved
for ocean). For the second city you step backwards from
bottom right in the same fashion. Then you place the two
greeneries by counting the cost numbers and stepping
clockwise around each city, starting from top left, skipping
illegal placements.
Special case: If you choose to play Tharsis Republic this
game, you get M€ production for the 2 neutral cities even
though they are placed before you reveal your corporation.

And 

You have a neutral opponent that you can steal from, or
reduce any kind of resources and production from.

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -23,6 +23,7 @@ import { FundedAward } from "./FundedAward";
 import { Milestone } from "./Milestone";
 import { ResourceType } from "./ResourceType";
 import * as constants from "./constants";
+import { Color } from "./Color";
 
 import { ALL_CORPORATION_CARDS } from "./Dealer";
 
@@ -37,15 +38,16 @@ export class Game {
     private oxygenLevel: number = constants.MIN_OXYGEN_LEVEL;
     private passedPlayers: Set<Player> = new Set<Player>();
     private researchedPlayers: Set<Player> = new Set<Player>();
-    private spaces: Array<ISpace> = new OriginalBoard().spaces;
+	private originalBoard = new OriginalBoard();
+	private spaces: Array<ISpace> = this.originalBoard.spaces;
     private temperature: number = constants.MIN_TEMPERATURE;
 
     constructor(public id: string, private players: Array<Player>, private first: Player) {
         this.activePlayer = first;
-        // Single player game player starts with 14TR
+        // Single player game player starts with 14TR and 2 neutral cities and forests on board		
         if (players.length === 1) {
-            players[0].terraformRating = players[0].terraformRatingAtGenerationStart = 14;
-        }
+            this.setupSolo();
+		}
         const corporationCards = this.dealer.shuffleCards(ALL_CORPORATION_CARDS);
         // Give each player their corporation cards
         for (let player of players) {
@@ -658,5 +660,19 @@ export class Game {
         }
         return undefined;
     }
+	private setupSolo() {
+			this.players[0].terraformRating = this.players[0].terraformRatingAtGenerationStart = 14;
+			// Single player add neutral player and put 2 neutrals cities on board with adjacent forest
+			let neutral = new Player("neutral", Color.NEUTRAL, true);
+			let space1 = this.originalBoard.getRandomCitySpace();
+			this.addCityTile(neutral, space1.id, SpaceType.LAND);
+			const fspace1 = this.originalBoard.getForestSpace(this.getAdjacentSpaces(space1));
+			this.addTile(neutral, SpaceType.LAND, fspace1, { tileType: TileType.GREENERY });
+			let space2 = this.originalBoard.getRandomCitySpace(30);
+			this.addCityTile(neutral, space2.id, SpaceType.LAND);
+			const fspace2 = this.originalBoard.getForestSpace(this.getAdjacentSpaces(space2));
+			this.addTile(neutral, SpaceType.LAND, fspace2, { tileType: TileType.GREENERY });
+			return undefined;
+	}			
 }
 
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -38,13 +38,13 @@ export class Game {
     private oxygenLevel: number = constants.MIN_OXYGEN_LEVEL;
     private passedPlayers: Set<Player> = new Set<Player>();
     private researchedPlayers: Set<Player> = new Set<Player>();
-	private originalBoard = new OriginalBoard();
-	private spaces: Array<ISpace> = this.originalBoard.spaces;
+    private originalBoard = new OriginalBoard();
+    private spaces: Array<ISpace> = this.originalBoard.spaces;
     private temperature: number = constants.MIN_TEMPERATURE;
 
     constructor(public id: string, private players: Array<Player>, private first: Player) {
         this.activePlayer = first;
-        // Single player game player starts with 14TR and 2 neutral cities and forests on board		
+        // Single player game player starts with 14TR and 2 neutral cities and forests on board
         if (players.length === 1) {
             this.setupSolo();
         }
@@ -673,6 +673,6 @@ export class Game {
         const fspace2 = this.originalBoard.getForestSpace(this.getAdjacentSpaces(space2));
         this.addTile(neutral, SpaceType.LAND, fspace2, { tileType: TileType.GREENERY });
         return undefined;
-    }			
+    }
 }
 
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -47,7 +47,7 @@ export class Game {
         // Single player game player starts with 14TR and 2 neutral cities and forests on board		
         if (players.length === 1) {
             this.setupSolo();
-		}
+        }
         const corporationCards = this.dealer.shuffleCards(ALL_CORPORATION_CARDS);
         // Give each player their corporation cards
         for (let player of players) {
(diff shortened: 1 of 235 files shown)


bafolts:
This was a big miss on my part and single player suddenly makes a lot more sense.

src/cards/Hackers.ts
```diff
@@ -17,6 +17,12 @@ export class Hackers implements IProjectCard {
         return player.energyProduction >= 1;
     }
     public play(player: Player, game: Game) {
+		if (game.getPlayers().length == 1) {
```
bafolts: You will need to 4 spaces instead of tabs to match what is already in `master`. Some of the spacing across all of these files seems different. I would make sure it aligns properly.

src/Game.ts
```diff
@@ -658,5 +660,19 @@ export class Game {
         }
         return undefined;
     }
+	private setupSolo() {
+			this.players[0].terraformRating = this.players[0].terraformRatingAtGenerationStart = 14;
+			// Single player add neutral player and put 2 neutrals cities on board with adjacent forest
+			let neutral = new Player("neutral", Color.NEUTRAL, true);
+			let space1 = this.originalBoard.getRandomCitySpace();
```
bafolts: Why do you need a random space? You can draw 4 cards from the dealer and follow the steps outlined in the pull request description.

bafolts:
Overall this looks good. If the tests pass and the spacing aligns we can merge for now.

src/cards/Hackers.ts
vincentneko: Should be fixed now

src/Game.ts
vincentneko: It seems simpler to address the issue with a RNG.
The method described in the rule book is not final, if the space selected by the card is not valid, what should be done ? Bouncing the RNG was easier I guess. Same for adjacent forest.

vincentneko:
Space / indentations issues should be ok now. As for the tests, I have a lot of errors !
I tried to fix Hackers.spec.ts, please let me know if it is the good way to do it.

src/Game.ts
```diff
@@ -37,14 +38,15 @@ export class Game {
     private oxygenLevel: number = constants.MIN_OXYGEN_LEVEL;
     private passedPlayers: Set<Player> = new Set<Player>();
     private researchedPlayers: Set<Player> = new Set<Player>();
-    private spaces: Array<ISpace> = new OriginalBoard().spaces;
+	private originalBoard = new OriginalBoard();
```
bafolts: This is tabs and should be spaces.

src/OriginalBoard.ts
```diff
@@ -125,4 +125,35 @@ export class OriginalBoard {
             new Ocean(colCount++, rowCount, [SpaceBonus.TITANIUM, SpaceBonus.TITANIUM])
         );
     }
+	
+	public getRandomSpace(offset: number = 0): Space {
```
bafolts: These tabs need to be spaces.

src/cards/Cards.ts
```diff
@@ -445,6 +453,10 @@ export class BigAsteroid implements IProjectCard {
         return true;
     }
     public play(player: Player, game: Game) {
+        if (game.getPlayers().length == 1) {
+			player.titanium += 4;
```
bafolts: These tabs need to be spaces.

src/cards/CloudSeeding.ts
```diff
@@ -17,6 +17,11 @@ export class CloudSeeding implements IProjectCard {
         return game.getOceansOnBoard() >= 3 - player.getRequirementsBonus(game) && player.megaCreditProduction > -5;
     }
     public play(player: Player, game: Game) {
+        if (game.getPlayers().length == 1) {
+            player.megaCreditProduction--;
+            player.plantProduction += 2;		
+			return undefined;
```
bafolts: These tabs need to be spaces.

src/cards/EnergyTapping.ts
```diff
@@ -17,6 +17,11 @@ export class EnergyTapping implements IProjectCard {
         return true;
     }
     public play(player: Player, game: Game) {
+        if (game.getPlayers().length == 1) {
+            player.energyProduction++;
+            player.victoryPoints--;			
+			return undefined;
```
bafolts: These tabs need to be spaces.

src/cards/Fish.ts
```diff
@@ -24,6 +24,7 @@ export class Fish implements IActionCard, IProjectCard {
         player.victoryPoints += player.getResourcesOnCard(this);
     }
     public play(_player: Player, game: Game) {
+		if (game.getPlayers().length == 1) return undefined;
```
bafolts: These tabs need to be spaces.

src/cards/GiantIceAsteroid.ts
```diff
@@ -20,6 +20,22 @@ export class GiantIceAsteroid implements IProjectCard {
         return true;
     }
     public play(player: Player, game: Game) {
+        if (game.getPlayers().length == 1) {
+			return new AndOptions(
+				() => {
+					return game.increaseTemperature(player, 2);
```
bafolts: These tabs need to be spaces.

src/cards/GreatEscarpmentConsortium.ts
```diff
@@ -17,6 +17,10 @@ export class GreatEscarpmentConsortium implements IProjectCard {
         return player.steelProduction > 0;
     }
     public play(player: Player, game: Game) {
+        if (game.getPlayers().length == 1) {
+			player.steelProduction++;
```
bafolts: These tabs need to be spaces.

src/cards/Hackers.ts
```diff
@@ -17,6 +17,12 @@ export class Hackers implements IProjectCard {
         return player.energyProduction >= 1;
     }
     public play(player: Player, game: Game) {
+        if (game.getPlayers().length == 1) {
+            player.megaCreditProduction += 2;
+			player.energyProduction--;
```
bafolts: These tabs need to be spaces.

src/cards/HeatTrappers.ts
```diff
@@ -17,6 +17,11 @@ export class HeatTrappers implements IProjectCard {
         return true;
     }
     public play(player: Player, game: Game) {
+		if (game.getPlayers().length == 1) {
```
bafolts: These tabs need to be spaces.

src/cards/PowerSupplyConsortium.ts
```diff
@@ -17,6 +17,10 @@ export class PowerSupplyConsortium implements IProjectCard {
         return player.getTagCount(Tags.ENERGY) >= 2;
     }
     public play(player: Player, game: Game) {
+        if (game.getPlayers().length == 1) {
+			player.energyProduction++
```
bafolts: These tabs need to be spaces.

src/cards/RoboticWorkforce.ts
```diff
@@ -130,6 +130,10 @@ export class RoboticWorkforce implements IProjectCard {
                 const foundCard: IProjectCard = selectedCards[0];
                 // this is the only card which requires additional user input
                 if (foundCard.name === new BiomassCombustors().name) {
+                    if (game.getPlayers().length == 1)  {
+						player.energyProduction += 2;
```
bafolts: These tabs need to be spaces.

vincentneko:
I've made a grep / sed to replace all tabs occurences, I hope it is ok now.

bafolts:
Tests will need to pass before this can be merged, otherwise the deployment will fail to heroku.

vincentneko:
I plan to correct tests as well.
I 've fixed Hackers.spec.ts, please let me know if it is the good way to do it

vincentneko:
There is something I don't understand with tests.

For MiningExpedition.spec.ts, I've added a second player to avoid execution of the one player if statement of the main class: 

describe("MiningExpedition", function () {
    it("Should play", function () {
        const card = new MiningExpedition();
        const player = new Player("test", Color.BLUE, false);
        const player2 = new Player("test2", Color.RED, false);
        const game = new Game("foobar", [player, player2], player);
        const action = card.play(player, game);
        expect(action).not.to.eq(undefined);
        player.plants = 2;
        action.cb(player);
        expect(player.plants).to.eq(0);
        expect(player.steel).to.eq(2);
        expect(game.getOxygenLevel()).to.eq(1);
    });
});

I have an error message I can't explain : 

tests/cards/MiningExpedition.spec.ts:17:19 - error TS2345: Argument of type 'Player' is not assignable to parameter of type 'Player & ISpace'.
  Type 'Player' is not assignable to type 'ISpace'.

17         action.cb(player);

bafolts:
You have modified `MiningExpedition` so that when there is only one player `game.increaseOxygenLevel(player, 1)` is called since there is no player to remove plants from. This has created a code path which may indirectly end up returning a `SelectSpace`. This situation would arise when you increased the oxygen level and achieved the bonus to increase temperature. If you then increased temperature to a point where you had to select a space for an aquifer you need to select a space for the aquifer at this time. You are now expecting `action` to be an `instanceof SelectPlayer`.

vincentneko:
Tests are ok now.
I had to switch all tests to 2 players games to avoid random city placement that led to random test errors like "space already occupied"
Then all TR test had to be converted to a 2 players games (starting TR is 20 instead of 14)
Please review and let me know if we can merge.

PR #15, opened 2019-10-27
Fix Moss bug

vincentneko:
Cards descriptions can be checked here : https://ssimeonoff.github.io/cards-list


bafolts:
This projects needs the front end skills from ssimeonoff.

PR #16, opened 2019-10-28
Fix awards scoring in case of equality

vincentneko:


src/Game.ts
```diff
@@ -127,14 +127,26 @@ export class Game {
             throw "Unsupported award " + award;
         }
         players.sort((player1, player2) => getScore(player2) - getScore(player1));
-        if (players.length <= 2) {
+        if (getScore(players[0]) > getScore(players[1])) {
```
bafolts: This fix will not work if there is only one player. In that situation `players[1]` will be `undefined` which will cause a runtime error. I think the fix for this would be to change this check.

```
if (players.length === 1) { // one player gets the awayd
   players[0].victoryPoints += 5;
} else if (getScore(players[0]) === getScore(players[1])) { // there was a tie and more than one player
   players[0].victoryPoints += 5;
   players[1].victoryPoints += 5;
} else { // there wasn't a tie and more than one player
   players[0].victoryPoints += 5;
   players[1].victoryPoints += 2;
}
```

vincentneko: I just checked the rules and saw that awards and milestones are disabled in 1 player games.
So we can kill 2 birds with one stone here with         
if (this.players.length === 1) return;

Situations where there is a tie with more than 2 players need also to be adressed, hence the while loops.

Btw, I've finished awards rework, you will see all modifications / adds in the merge request.
Vitor will need some modifications after the merge.

src/Game.ts
```diff
@@ -91,62 +89,50 @@ export class Game {
...
-        } else {
-            players[0].victoryPoints += 5;
-            players[1].victoryPoints += 2;
+            players.shift();
+            if (players.length > 1) {
+                if (award.getScore(players[0], this) > award.getScore(players[1], this)) {
+                    players[0].victoryPoints += 3;
+                }  else {  // We have at least 2 rank 2 players
+                    let score = award.getScore(players[0], this);
+                    while (award.getScore(players[0], this) === score) {
```
bafolts: Eventually as `players.shift()` is called the length of `players` will reach `0`. When this happens `players[0]` will be `undefined`. This loop should also verify that `players.length > 0`.

src/Game.ts
```diff
@@ -91,62 +89,50 @@ export class Game {
...
+                        players[0].victoryPoints += 3;
+                        players.shift();
+                    }                
+                }
+            }    
+        } else { // We have at least 2 rank 1 players 
+            let score = award.getScore(players[0], this);
+            while (award.getScore(players[0], this) === score) {
```
bafolts: This while loop will also need to check that `players.length > 0`.

src/Player.ts
```diff
@@ -796,8 +796,8 @@ export class Player {
         });
     }
 
-    private fundAward(award: Award, game: Game): PlayerInput {
-        let upperCaseAward = String(award)[0].toUpperCase() + String(award).substring(1);
+    private fundAward(award: IAward, game: Game): PlayerInput {
+        //let upperCaseAward = String(award)[0].toUpperCase() + String(award).substring(1);
```
bafolts: This comment should be deleted.

src/Game.ts
vincentneko: done

src/Game.ts
vincentneko: Done

src/Game.ts
```diff
@@ -103,15 +103,15 @@ export class Game {
                     players[0].victoryPoints += 3;
                 }  else {  // We have at least 2 rank 2 players
                     let score = award.getScore(players[0], this);
-                    while (award.getScore(players[0], this) === score) {
+                    while (award.getScore(players[0], this) === score && players.length > 0) {
```
bafolts: This conditional gets parsed from left to right. The check for `players.length` needs to go before the call to `award.getScore`.

vincentneko: ok got it thanks

src/cards/prelude/Vitor.ts
```diff
@@ -15,29 +15,16 @@ export class Vitor implements CorporationCard {
...
-    public initialAction(player: Player, game: Game) {
-        return new OrOptions(
-            new SelectOption("Fund landlord award", () => {
-                game.fundAward(player, Award.LANDLORD);
+    public initialAction(player: Player, game: Game) {        
+        const freeAward = new OrOptions();
+        for (let award of ORIGINAL_AWARDS) {
+            freeAward.options.push(
+            new SelectOption("Fund "+ award.name +" award", () => {
+                game.fundAward(player, award);
```
bafolts: By the time this function is called `award` will be the last item of `ORIGINAL_AWARDS`. Everything else in this pull request looks fine. If you change this to the following it will work:

```
new SelectOption("Fund " + award.name + " award", game.fundAward.bind(game, player, award));
```

vincentneko: Thanks for the explanation. It's modified.

src/cards/prelude/Vitor.ts
```diff
@@ -15,29 +15,16 @@ export class Vitor implements CorporationCard {
...
-    public initialAction(player: Player, game: Game) {
-        return new OrOptions(
-            new SelectOption("Fund landlord award", () => {
-                game.fundAward(player, Award.LANDLORD);
+    public initialAction(player: Player, game: Game) {        
+        const freeAward = new OrOptions();
+        for (let award of ORIGINAL_AWARDS) {
+            freeAward.options.push(
+            new SelectOption("Fund "+ award.name +" award", () => {
+                game.fundAward.bind(player, award);
```
bafolts: This is also not right. The first argument to the bind function is the `thisArg`, for this the `thisArg` should be `game`. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind

vincentneko: My bad... I need some glasses I guess. Should be ok now

PR #17, opened 2019-11-03
Bug fixes and add TR to final score

vincentneko:
- Add TR to final score
- Fix final greenery bug
- Fix WaterSplittingPlant missing canAct
- Fix Tardigrades missing canAct

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -390,6 +390,11 @@ export class Game {
     private gotoEndGame(): void {
         this.phase = Phase.END;
 
+        // TR is converted in victory points
+        this.players.forEach((player) => {
+            player.victoryPoints += player.terraformRating;
+        });
+
         // Give players any victory points from cards
         this.players.forEach((player) => {
             player.playedCards.forEach((playedCard) => {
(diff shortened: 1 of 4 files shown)


src/Game.ts
```diff
@@ -390,6 +390,11 @@ export class Game {
     private gotoEndGame(): void {
         this.phase = Phase.END;
 
+        // TR is converted in victory points
+        this.players.forEach((player) => {
+            player.victoryPoints += player.terraformRating;
```
bafolts: Does the outside of the board represent a player's terraform rating or victory points? Does this make more sense to do the opposite and add victoryPoints to terraformRating?

src/cards/WaterSplittingPlant.ts
```diff
@@ -19,6 +19,9 @@ export class WaterSplittingPlant implements IProjectCard {
         return undefined;
     }
     public actionText: string = "Spend 3 energy to increase oxygen level by 1";
+    public canAct(player: Player): boolean {
+        return player.energy >= 3;
+    }
     public action(player: Player, game: Game) {
         if (player.energy < 3) {
```
bafolts: I would remove this `throw "Need 3 energy"`. I was throwing these errors before I added the `canAct` method.

vincentneko: done

src/Game.ts
vincentneko: The answer is both. Before end game the outside of the board represents a player's terraform rating but after end game, it represents victory points.
Scoring is based on victory points so I guess it make more sense to add TR to VP in the end. (TR is only one way to score VPs)

PR #20, opened 2019-11-05
Milestones rework

vincentneko:
Milestones rework (same as awards, to prepare hellas & elysium)

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -208,12 +208,12 @@ function createGame(req: http.IncomingMessage, res: http.ServerResponse): void {
 function getPlayer(player: Player, game: Game): string {
     const output = {
         cardsInHand: getCards(player, player.cardsInHand),
-        claimedMilestones: game.claimedMilestones.map((claimedMilestone) => { return { player: claimedMilestone.player.id, milestone: claimedMilestone.milestone } }),
+        claimedMilestones: game.claimedMilestones.map((claimedMilestone) => { return { player: claimedMilestone.player.id, milestone: claimedMilestone.milestone.name +" " } }),
         color: player.color,
         corporationCard: player.corporationCard ? player.corporationCard.name : undefined,
         energy: player.energy,
         energyProduction: player.energyProduction,
-        fundedAwards: game.fundedAwards.map((fundedAward) => { return { player: fundedAward.player.id, award: fundedAward.award } }),
+        fundedAwards: game.fundedAwards.map((fundedAward) => { return { player: fundedAward.player.id, award: fundedAward.award.name +" " } }),
         generation: game.getGeneration(),
         heat: player.heat,
         heatProduction: player.heatProduction,
--- a/server.ts
+++ b/server.ts
@@ -208,7 +208,7 @@ function createGame(req: http.IncomingMessage, res: http.ServerResponse): void {
 function getPlayer(player: Player, game: Game): string {
     const output = {
         cardsInHand: getCards(player, player.cardsInHand),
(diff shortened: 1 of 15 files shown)


server.ts
```diff
@@ -208,12 +208,12 @@ function createGame(req: http.IncomingMessage, res: http.ServerResponse): void {
 function getPlayer(player: Player, game: Game): string {
     const output = {
         cardsInHand: getCards(player, player.cardsInHand),
-        claimedMilestones: game.claimedMilestones.map((claimedMilestone) => { return { player: claimedMilestone.player.id, milestone: claimedMilestone.milestone } }),
+        claimedMilestones: game.claimedMilestones.map((claimedMilestone) => { return { player: claimedMilestone.player.id, milestone: claimedMilestone.milestone.name +" " } }),
```
bafolts: The suffix of the ' ' character doesn't make sense.

src/milestones/Planner.ts
```diff
@@ -0,0 +1,10 @@
+import { IMilestone } from "./IMilestone";
+import { Player } from "../Player";
+import { Game } from "../Game";
+
+export class Planner implements IMilestone {
+    public name: string = "Planner";
+    public canClaim(player: Player, game: Game): boolean {
+        return !game.milestoneClaimed(this) && player.cardsInHand.length >= 16;
```
bafolts: It is redundant to check `game.milestoneClaimed` in each `IMilestone` implementation. I would check this in the filtering loop instead. Otherwise we have to always remember to add this to any new milestones and if game rules ever changed and allowed to claim a milestone twice we would have to update each milestone.

vincentneko: Modified

server.ts
vincentneko: This space was added at the wrong place...

bafolts: There is also an unusual space character appended on this line.

src/components/ClaimedMilestone.ts
```diff
@@ -15,7 +15,7 @@ export const ClaimedMilestone = Vue.component("claimed-milestone", {
         if (player === undefined) {
             return createElement("span", { domProps: { innerHTML: "Unknown Player" } });
         }
-        return createElement("span", { domProps: { innerHTML: claimedMilestone.milestone + " " + player.name } });
+        return createElement("span", { domProps: { innerHTML: claimedMilestone.milestone + " " + player.name + " "} });
```
bafolts: Unusual space character appended on this line.

server.ts
vincentneko: removed

src/components/ClaimedMilestone.ts
vincentneko: This space is needed in order to separate multiple milestones.
Same for awards.
I think something is missing in component ClaimedMilestone because Milestone name is not displayed (there is undefined instead), I need to understand how props work in vue js

bafolts: You should be able to put a '&nsbp;` in the parent template if you want to separate these by a space.

vincentneko: Not sure to understand where I should put a '&nsbp;` in the parent template. 
(I assume parent template is playerHome)

Without the space I've added, awards are rendered like this :

Funded Awards
Scientist vincentBanker vincentMiner tim

With the space, it's like :
Scientist vincent Banker vincent Miner tim

bafolts: The template file would need an & n b s p ; element. I can do this in a separate commit.

PR #23, opened 2019-11-20
Update card template to use ssimeonoff html and css.

bafolts:
The ssimeonoff repository has an index of all cards. This leverages that projects html and css to improve upon the card template.


server.ts
```diff
@@ -49,11 +49,14 @@ const pngs: Map<string, Buffer> = new Map<string, Buffer>([
...
     if (req.url !== undefined) {
         if (req.method === "GET") {
+
             if (
                 req.url === "/" ||
                 req.url.startsWith("/game?id=") ||
                 req.url.startsWith("/player?id=")) {
                 serveApp(res);
+            } else if (req.url.startsWith("/assets/")) {       
+                serveResource(res, fs.readFileSync('.'+req.url));
```
bafolts: We will need to specify each image that can be loaded. As written this will load any file from the underlying file system which node has read access. For example if the provided url were `/assets/../../../anyFile.txt`.

vincentneko: Fixed, I've left only ssimeonoff pngs

PR #26, opened 2019-11-26
Prelude implementation part 1

vincentneko:
Fix requirements bugs
Add mechanism for "shoot again" cards
Modify solo with prelude

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -29,7 +29,6 @@ import {ORIGINAL_AWARDS} from './awards/Awards';
 import {IAward} from './awards/IAward';
 
 const INITIAL_ACTION: string = 'INITIAL';
-const PRELUDE_ACTION: string = 'PRELUDE';
 
 export class Player {
     public corporationCard: CorporationCard | undefined = undefined;
@@ -541,8 +540,7 @@ export class Player {
         "Select prelude card to play",
         this.preludeCardsInHand,
         (foundCards: Array<IProjectCard>) => {
-            this.playCard(game, foundCards[0]);
-            return undefined;
+            return this.playCard(game, foundCards[0]);
         },
         1,
         1
@@ -595,14 +593,14 @@ export class Player {
     }
 
     public playCard(game: Game, selectedCard: IProjectCard, howToPay?: HowToPay): PlayerInput | undefined { 
-        const whenDone = () => {
-          this.cardsInHand
-              .splice(
-                  this.cardsInHand
-                      .findIndex(
-                          (card) => card.name === selectedCard.name
-                      ), 1
-              );
+      const whenDone = () => {
+          var projectCardIndex = this.cardsInHand.findIndex((card) => card.name === selectedCard.name);
+          var preludeCardIndex = this.preludeCardsInHand.findIndex((card) => card.name === selectedCard.name);
+          if (projectCardIndex !== -1) {
+            this.cardsInHand.splice(projectCardIndex, 1);
+          } else if (preludeCardIndex !== -1) {
+            this.preludeCardsInHand.splice(preludeCardIndex, 1);
+          }
           this.addPlayedCard(game, selectedCard);
 
           if (howToPay !== undefined) {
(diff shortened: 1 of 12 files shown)


src/Player.ts
```diff
@@ -655,6 +654,11 @@ export class Player {
         }
         whenDone();
 
+        // Shoot again prelude cards 
+        if (selectedCard.name === "Eccentric Sponsor" || selectedCard.name === "Ecology Experts") {
```
bafolts: From what I understand of the rules the prelude cards should be played during the first generation.

```
After all corporations have been played and cards paid
for (step 7), there is an extra round (step 7b) where each
player plays their pair of picked Prelude cards in player
order, and discards their remaining 2 Prelude cards.
They work like green cards, and stay in play with their
tags visible.
After this, the first player begins the action phase
(step 8). Note that the whole setup counts as part of
generation 1.
```
 
 I don't see where we are doing this. It should be called out as a separate action if we are using the prelude extension.

bafolts: From the rules it seems there are prelude cards inserted into the main project card deck. A player should not have to pay for those as well since they have no cost. When those cards are drawn we need to factor in that they have no cost.

bafolts: After `playCorporationCard` we should direct the game logic to the prelude round where each player plays their prelude cards instead of how it is written today. If you remove this part from this pull request the rest looks fine and I could implement the logic for introducing the round where prelude cards are played.

vincentneko: This was prerequisite work before implementing prelude correctly.
My attempt to do it is in next commit.

vincentneko: About your second comment, prelude cards inserted into the main project card deck (House Printing for instance) are additional projects and should be treated like other project cards.

src/Player.ts
```diff
@@ -1159,11 +1225,26 @@ export class Player {
     }
 
     public takeAction(game: Game): void {
+
+      //Prelude cards have to be played first
+      if (this.preludeCardsInHand.length > 0) {
+        const input = this.playPreludeCard(game);
+        input.onend = () => {
+          this.actionsThisGeneration.add(PRELUDE_ACTION);
+          this.actionsTakenThisRound++;
```
bafolts: Why does actionsTakenThisRound need to be incremented twice? A player needs to play both their prelude cards. The `playPreludeCard` function seems to pick and play one prelude card. If it is simpler you could just play the two prelude cards for the player. There is no option not to play the two selected prelude cards so may as well play the cards (call card.play) the moment they are picked.

src/cards/prelude/HugeAsteroid.ts
```diff
@@ -6,13 +6,13 @@ import { IProjectCard } from "../IProjectCard";
...
     public tags: Array<Tags> = [];
-    public cost: number = 5;
     public name: string = "Huge Asteroid";
     public text: string = "Increase Temperature 3 steps. Pay 5 MC.";
     public requirements: undefined;
     public description: string = "Deep impact on Mars before too many move here";
     public play(player: Player, game: Game) {
-		return game.increaseTemperature(player, 3);
+      player.megaCredits -=5;	
+		  return game.increaseTemperature(player, 3);
```
bafolts: The spacing is strange here. I plan to eventually add eslint to catch these places automatically.

src/Player.ts
```diff
@@ -535,6 +536,19 @@ export class Player {
       return card.tags.indexOf(Tags.SPACE) !== -1;
     }
 
+    private playPreludeCard(game: Game): PlayerInput {
+      return new SelectCard(
+        "Select prelude card to play",
+        this.preludeCardsInHand,
+        (foundCards: Array<IProjectCard>) => {
+            this.playCard(game, foundCards[0]);
```
bafolts: @vincentneko I simplified this logic which was duplicated. I don't know what the "shoot again" logic was accomplishing. The major difference with prelude cards is they have no cost so I made the `HowToPay` argument optional to the new `playCard` function.

src/Player.ts
vincentneko: When actionsTakenThisRound was not incremented twice, a player could do 3 actions in a round (in a 2 player game)

bafolts:
Pass it in. I missed this.

PR #30, opened 2019-11-29
Prelude play another card cards

vincentneko:
Another attempt to implement those "play another card" cards.
Right now postAction only triggers a playProjectCard but could be modified in the future if needed.

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1192,6 +1192,8 @@ export class Player {
         this.setWaitingFor(input);
         this.postAction = false;
         return;
+      } else if (this.postAction && this.getPlayableCards(game).length === 0) {
+        this.postAction = false;
       }
       
       //Prelude cards have to be played first
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1183,7 +1183,7 @@ export class Player {
     public takeAction(game: Game): void {
 
       //Post Action (after some specific prelude cards have been played)
-      if (this.postAction) {
+      if (this.postAction && this.getPlayableCards(game).length > 0) {
         const input = this.playProjectCard(game);
         input.onend = () => {
           this.actionsTakenThisRound++;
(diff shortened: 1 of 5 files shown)


src/cards/prelude/EccentricSponsor.ts
```diff
@@ -9,6 +9,7 @@ export class EccentricSponsor extends PreludeCard implements IProjectCard {
     public name: string = "Eccentric Sponsor";
     public text: string = "Next card you play this turn will have a 25 MC reduction. Playing this card does not count as an action";
     public requirements: undefined;
+    public postPlay: boolean = true;
```
bafolts: This check would work as long as it determined if a player had at least one card they could play. If this were a function that took in an instance of `Player` and `Game` it should work fine. If a player has no cards they can play this function would return `false`.

vincentneko: I've added a check in the //Post Action section, as modifiers from Eccentric Sponsor or Ecology Experts might not be active if the getPlayableCards is done in the postPlay function

src/Player.ts
```diff
@@ -1177,6 +1182,18 @@ export class Player {
 
     public takeAction(game: Game): void {
 
+      //Post Action (after some specific prelude cards have been played)
+      if (this.postAction && this.getPlayableCards(game).length > 0) {
```
bafolts: If `postAction` is `true` and the result of `getPlayableCards(game)` is an empty array than the `postAction` property needs to be set to `false`. Also, if you played `Eccentric Sponsor` and then selected `Ecology Experts` to play I am not sure this would work. We would set `postAction` to `false` after playing the first card.

vincentneko: I've added a check if getPlayableCards is empty.
I'm not sure what you mean by "if you played Eccentric Sponsor and then selected Ecology Experts".
If I play  Eccentric Sponsor, I cannot select Ecology Experts as it is a prelude card, not a project card.

bafolts: Ah I understand now. There are prelude cards and prelude project cards. I didn't realize there were separate prelude project cards. Since you could never play both of these prelude cards in sequence what is here should work.

Issue #49, opened 2019-12-09
IMPORTED NITROGEN can block the game


vincentneko:
If there is no animal or microbe card to select the game will hang.

vincentneko:
Thanks for fixing this

PR #54, opened 2019-12-10
Better ImportedNitrogen #49

bafolts:
Implemented the canPlay method from being a tautology to checking if there are cards which can have the resources needed added to them.

### DIFF
--- a/src/cards/ImportedNitrogen.ts
+++ b/src/cards/ImportedNitrogen.ts
@@ -15,23 +15,38 @@ export class ImportedNitrogen implements IProjectCard {
     public canPlay(): boolean {
         return true;
     }
+    private giveResources(player: Player): undefined {
+        player.terraformRating++;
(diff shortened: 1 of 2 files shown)


alrusdi:
This fix seems to be incorrect. It's not REQUIRED to have ability to place Microbes and Animals to another cards it is an option. Rules page 9:
You may play a card that...
...
-  adds  resources  that  you  can’t  collect  (for  example  adding   microbe   resources   without   having   any   microbe   project to place them on).

alrusdi:
I believe, both "SelectCard" actions must be optional and we need to show these options only if someone is already played an applicable cards

vincentneko:
I second this

bafolts:
I updated the card logic so that a player is only presented with options that make sense. If they have no applicable cards they are only given the terraform rating and plants.

PR #61, opened 2019-12-12
Solo game final page

alrusdi:
I prepared special solo game final page.
My English is bad, so please change the texts in src/components/GameEnd.ts to be correct


server.ts
```diff
@@ -437,6 +447,23 @@ function serveApp(res: http.ServerResponse): void {
...
+  game.generation = 14;
+  game.setOxygenLevel(14);
+  game.setTemperature(7);
+  games.set(gameId, game);
+  playersToGame.set(player.id, game);
+
+  res.writeHead(302, {'Location': '/player?id='+player.id});
+  res.end();
+}
+
```
vincentneko: Not sure if we really need this in main code.
As global parameters are stored in constants file, I think it's easy enough for us to change them localy for testing purpose.

src/Game.ts
```diff
@@ -80,6 +80,20 @@ export class Game {
...
+      this.oxygenLevel = newOxygenLevel; 
+    }
+      
+    public setTemperature(newTemperature: number): void {
+      if (newTemperature < -30 || newTemperature > 8) {
+        throw "Incorrect temperature " + newTemperature.toString();
+      }
+      this.temperature = newTemperature; 
+    }
+
```
vincentneko: Not sure if we really need this in main code.
As global parameters are stored in constants file, I think it's easy enough for us to change them localy for testing purpose.

alrusdi:
Indeed. I removed this unnecessary debug code

Issue #67, opened 2019-12-14
Wrong lose conditions for solo game


alrusdi:
I've set constants to:
MAX_OXYGEN_LEVEL: number = 0;
MAX_OCEAN_TILES: number = 0;
MAX_TEMPERATURE: number = -26;

When I increase temperature 2 steps, game is ending but with the "Sorry, you lose."

I'm expected to win here since I've achieved terraforming before generation 14.

alrusdi:
Merged

PR #68, opened 2019-12-14
Fix solo win conditions; refs #67

alrusdi:
Correct count oceans and generation conditions for victory


alrusdi:
These methods are private. Furthermore they are server-side, and depend on game instance.
Should i do ajax call to figure out win conditions? My solution is stupid but it is a reasonable compromise IMHO

alrusdi:
Oh, I got your idea. I will concentrate checking code in new game public method and then get it with player info from server-side.

Issue #62, opened 2019-12-12
Multiplayer end game is bugged


vincentneko:
It was the first time I was able to reach end game in multiplayer.
I was playing with a friend, it was a real blast !!!

Some feedback about it:
- I played most part of the game on mobile
- We texted one another to say "your turn", so even before auto refresh, I guess browser notifications should be #1 improvement item on the list.

In the end, the game never finished and carry on with a list of things we can do even if we have passed.

Our MC keep increasing just like generations were continuing.

Here is a screenshot that might help...

![endgame](https://user-images.githubusercontent.com/56086992/70751041-e328ed00-1d2f-11ea-8542-e678fbba6839.png)

I've set up a game server on openshift. This one will not stop after a 20 min period of inactivity, so if you want to set up a "long play game", please let me know so we can play a game together !

bafolts:
Setting the max limits required for terraforming to really low values
should help expose this problem. I agree adding polling and notifications
is a must have.

alrusdi:
Please help me with this: https://github.com/bafolts/terraforming-mars/pull/61
It contains a code to fast start the game with given conditions (close to ending).
It has implementation of final page (for "end" game phase) with Victory Points table for each player. Also it disallows to visit /player page for finished game again, so it may resolve current issue. I need a review and grammar/spelling checks due to my poor English.

vincentneko:
I've narrowed the problem.
It is related with gotoFinalGreeneryPlacement.

If both player don't have enough plants to create a greenery, the game stops fine.

I've setup a 2 player games, each have 8 plants.

After all parameters are maxed (done some tweaking into constants so only 2 temp increase are needed), first player got the "Place any final greenery from plants"
and when choosing a space got "Was no player left to place final greenery" error message.

Selecting this option multiple times lead to add some greeneries on the board, visible from other player page.

I hope this helps

vincentneko:
I think I got it.

In gotoFinalGreeneryPlacement, there is a filter on player that can put a forest and then there is 
firstPlayer = this.getNextPlayer(players, firstPlayer);

I've added this line after the filter, I will cary on testing

      // If no players can place greeneries we are done
      if (players.length === 0) {
        this.gotoEndGame();
        return;
      }
      this.startFinalGreeneryPlacement(players[0]);

vincentneko:
This seems to work, I'm working on a merge request

PR #70, opened 2019-12-14
Issue #62

vincentneko:


vincentneko:
> I think it is little bit incorrect. Greenery placement should start in turn order of last generation. If board is almost filled with trees it may be important who will place the last possible greenery first.

You are right, I will add some sort to call the players in the right order.

Issue #88, opened 2019-12-16
Space Station doesn't reduce costs


dabewi:
The card _Space Station_ should reduce the amount of money you have to spend to buy a card with a space tag by 2. Sadly it doesn't reduce the costs. 

How to Reproduce: 

1. Buy Space Station
2. Play any other card with a space symbol

Cheer

alrusdi:
Merged

PR #91, opened 2019-12-16
Fix cards discount clientside validation; #88

alrusdi:


src/components/SelectHowToPayForCard.ts
```diff
@@ -36,12 +36,19 @@ export const SelectHowToPayForCard = Vue.component("select-how-to-pay-for-card",
     },
     methods: {
         getCardCost: function () {
+            for (const icard of this.player.cardsInHand) {
+                if (this.$data.card === icard.name) {
+                    return icard.calculatedCost
+                }
+            }
+            /*
             if (this.$data.card !== undefined) {
```
bafolts: I would remove this instead of leaving it in comments.

PR #98, opened 2019-12-17
Issue 81

alrusdi:
8 more cards fixed. Please merge, but do not remove a branch

### DIFF
--- a/src/TestingUtils.ts
+++ b/src/TestingUtils.ts
@@ -4,9 +4,13 @@ import * as constants from "./constants"
 import { SpaceType } from "./SpaceType";
 
 
-export const maxOutOceans = function(player: Player, game: Game): void {
+export const maxOutOceans = function(player: Player, game: Game, toValue: number = 0): void {
+    if (toValue < 1) {
+        toValue = constants.MAX_OCEAN_TILES;
+    }
+    
     for (const space of game.getSpaces(SpaceType.OCEAN)) {
-        if (game.getOceansOnBoard() >= constants.MAX_OCEAN_TILES) break;
+        if (game.getOceansOnBoard() >= toValue) break;
         game.addOceanTile(player, space.id)
     }
 };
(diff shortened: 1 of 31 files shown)


vincentneko:
@bafolts can you please review as both alrusdi and I have worked on this merge ? Thanks

src/TestingUtils.ts
```diff
@@ -0,0 +1,16 @@
+import { Player } from "./Player";
```
bafolts: This should go in `tests/TestingUtils.ts` as the class is only needed for files in `tests/`.

src/cards/Birds.ts
```diff
@@ -15,14 +15,30 @@ export class Birds implements IActionCard, IProjectCard {
...
+        return false;
+      }
       return game.getOxygenLevel() >= 13 - player.getRequirementsBonus(game);
     }
     public onGameEnd(player: Player) {
       player.victoryPoints += player.getResourcesOnCard(this);
     }
     public play(_player: Player, game: Game) {
       if (game.getPlayers().length == 1) return undefined;
+      if ( ! this.canPlay(_player, game)) {
```
bafolts: Calls to `canPlay` shouldn't be included in the call to `play`. It is assumed the card can be played once `play` is called.

src/cards/CloudSeeding.ts
```diff
@@ -11,7 +11,18 @@ export class CloudSeeding implements IProjectCard {
...
+    private playersWithHeatProduction(game: Game): boolean {
+      for (const player of game.getPlayers()) {
+        if (player.heatProduction >= 1) return true;
+      }
+      return false;
+    }
+    
     public canPlay(player: Player, game: Game): boolean {
+
+      if ( ! this.playersWithHeatProduction(game)) return false;
```
bafolts: https://github.com/bafolts/terraforming-mars/blob/master/src/cards/CloudSeeding.ts#L30

I would return the array of players with heat production and only allow those players to be selected. With that change an error no longer needs to be thrown and `canPlay` can check if the array is empty.

PR #99, opened 2019-12-17
Feature : Adding cards in hand number

pierrehilbert:
Not knowing the number of cards in the opponent hand is a problem to evaluate his possibilities.
With this feature, you will be able to see the number of cards without exposing which cards he has..


src/Player.ts
```diff
@@ -54,6 +54,7 @@ export class Player {
     public plants: number = 0;
     public plantProduction: number = 0;
     public cardsInHand: Array<IProjectCard> = [];
+    public cardsInHandNbr: number = 0;
```
bafolts: The `cardsInHandNbr` property isn't needed on `Player`. The rest looks great.

src/components/OtherPlayer.ts
```diff
@@ -33,6 +33,7 @@ export const OtherPlayer = Vue.component("other-player", {
             <h3 :style="'color:' + player.color" v-on:click="toggleDisplayed()">{{player.name}}</h3>
             <div v-if="displayed === true">
                 <div>Terraform Rating: {{player.terraformRating}}</div>
+                <div>Cards In Hand: {{player.cardsInHandNbr}}</div>
```
alrusdi: We can say just:
{{ player.cardsInHand.length }}

pierrehilbert: No, the attribute wasn't available for other players (take a look at the server.ts modification).
Having this new attribute allow to have the number of cards without being able to see which one in the console.

alrusdi: Oh, you right! My fault.

PR #104, opened 2019-12-18
Issue 81 More issues fixed

alrusdi:
Hooray! Just 9 issues remaining. We are close to finish this huge task.
Please review, merge, but do not remove branch


src/cards/Fish.ts
```diff
@@ -15,19 +15,40 @@ export class Fish implements IActionCard, IProjectCard {
   public name: string = 'Fish';
   public resourceType: ResourceType = ResourceType.ANIMAL;
   public cardType: CardType = CardType.ACTIVE;
+
+  private getPlayersWithPlantProduction(currentPlayer: Player, game: Game): Array<Player> {
+    var players = game.getPlayers().filter((p) => p.plantProduction > 0);
+    if (players.length === 2) {
```
bafolts: Why does this have to equal 2?

src/cards/HeatTrappers.ts
```diff
@@ -11,20 +11,46 @@ export class HeatTrappers implements IProjectCard {
     public tags: Array<Tags> = [Tags.ENERGY, Tags.STEEL];
     public name: string = "Heat Trappers";
     public cardType: CardType = CardType.AUTOMATED;
-    public canPlay(): boolean {
-        return true;
+
+    private getPlayersWithHeatProduction(currentPlayer: Player, game: Game): Array<Player> {
+        var players = game.getPlayers().filter((p) => p.heatProduction > 1);
+        if (players.length === 2) {
```
bafolts: This is becoming a pattern. I am not sure what is special about 2 players. I would think this should be `if (players.length > 1)`

alrusdi: You are right! Wrong decision and then copy paste.

alrusdi:
I resolved review issues and added fixes for remaining cards.
I everything is alright please merge and delete the branch. We still have some problems (with Protected Habitatas and Pets) but they need a sort of common solution. I will create new issue for it and ask @bafolts for help

Issue #58, opened 2019-12-11
Browser notifications


vincentneko:
It would be great if the game could push a browser notification when it's our turn to play in a multiplayer game

bafolts:
Can this be closed since https://github.com/bafolts/terraforming-mars/pull/111 was merged?

vincentneko:
It was working for Alrusdi but not for me, and it requires a https server, which I don't have.
I think it's working fine anyway so let's close it and if some problems arise, we can open a new issue

PR #111, opened 2019-12-19
Auto refresh and notifications; #58

alrusdi:


script.ts
```diff
@@ -32,6 +32,7 @@ const app = new Vue({
             xhr.onload = () => {
                 if (xhr.status === 200) {
                     app.$data.player = xhr.response;
+                    app.playerkey ++;
```
bafolts: What is this `playerkey` property doing? I don't see it used.

alrusdi: This forces to refresh window because otherwise Vue somehow does not reacts to changes.
I found it in your code https://github.com/bafolts/terraforming-mars/blob/master/src/components/WaitingFor.ts#L44

alrusdi:
Sorry, i merged my work with assets and css to this branch. Let's review all changes if you don't mind.

PR #115, opened 2019-12-20
Disallow oxygenLevel and temperature to get out of the bounds

alrusdi:


src/Game.ts
```diff
@@ -607,6 +607,10 @@ export class Game {
       if (this.oxygenLevel >= constants.MAX_OXYGEN_LEVEL) {
         return undefined;
       }
+      if (this.oxygenLevel + steps > constants.MAX_OXYGEN_LEVEL) {
+        if (steps == 1) return undefined;
```
bafolts: If `oxygenLevel + 1 > constants.MAX_OXYGEN_LEVEL` is `true` then `this.oxygenLevel === constants.MAX_OXYGEN_LEVEL`. This check for `steps === ` should be unnecessary. Steps will always be `2` here. I would change this to:

```
if (steps === 2 && this.oxygenLevel + steps > constants.MAX_OXYGEN_LEVEL) {
    return this.increaseOxygenLevel(player, 1);
}
```

src/Game.ts
```diff
@@ -624,6 +628,11 @@ export class Game {
       if (this.temperature >= constants.MAX_TEMPERATURE) {
         return undefined;
       }
+      if (this.temperature + 2 * steps > constants.MAX_TEMPERATURE) {
+        if (steps == 1) return undefined;
```
bafolts: Similar to my comment above I think this check for `1` is unnecessary. I would change this to:

```
if (steps > 1 && this.temperature + 2 * steps > constants.MAX_TEMPERATURE) {
    return this.increaseTemperature(player, steps - 1);
}
```

tests/Game.spec.ts
```diff
@@ -69,4 +70,38 @@ describe("Game", function () {
         expect(player3.victoryPointsBreakdown.awards).to.eq(3); // one shared 1st place
 
     });
+
+    it("Disallows to set temperature more than allowed maximum", function () {
```
bafolts: Thank you, these tests are helpful. Both `Player` and `Game` need a lot of tests. We should try to add tests like this during each change.

vincentneko: For awards, I think that when there is a tie for the first place, all players get 5 points, not 3. And also there is no second player to get points for this award.

Example : 4 player game, thermalist award:

Player 1 : 5 heat 
Player 2 : 11 heat
Player 3 : 9 heat
Player 4 : 11 heat

Results : both player 2 and 4 get 5 VP, players 1 and 3 get no VP

alrusdi: @vincentneko It's true, we should fix it. But is it comment somehow related to this PR? May be we should create another issue for it?

vincentneko: Yes you are right, the PR looks good, let's merge it and deal with the awards tie in a separate issue.

Issue #18, opened 2019-11-03
Drafting option


vincentneko:
Drafting option is currently missing and would be a great addition to the game.
Below is a extract from rule book:

Draft variant:
If more interaction is desired, the Draft variant
may be used. During the Research phase the players draft 4
cards to buy from instead of just drawing 4 to buy from:
Each player first gets 4 cards and chooses one to
draft, putting it aside and passing the rest to the next player.
Then you have 3 cards; set aside 1 of them and pass the rest
to the left. Then set aside 1 of the 2 you receive, pass the last
card, and finally receive your last card.
Then examine the 4 cards you’ve set aside (drafted)
and choose which to buy (3 M€ each) and which to discard.
Drafting is not used during the first generation
since the first Research phase is skipped (see setup). Cards
are passed clockwise during even-numbered generations
and counter-clockwise during odd-numbered generations.
This variant allows players to affect which cards
other players get access to, and increases your ability to
pursue a certain strategy, or predict which parameter will
rise first.
The Draft Variant adds some extra game time and
is not recommended for players new to Terraforming Mars.

alrusdi:
I need this option too. There few notes about implementation.
1) After each round draft direction must change
2) Each drafting turn you should be able to see which opponent will receive your cards

vincentneko:
First try to implement drafting (not as an option yet), please test it on branch draft-option and let me know if it's ok for you.

vincentneko:
Done and merged

PR #125, opened 2019-12-22
Draft variant - Issue #18

vincentneko:
Draft variant as an option

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -32,15 +32,18 @@ export class Game {
     public dealer: Dealer;
     public fundedAwards: Array<FundedAward> = [];
     public generation: number = 1;
+    private draftRound: number = 1;
     public phase: Phase = Phase.RESEARCH;
     private donePlayers: Set<Player> = new Set<Player>();
     private oxygenLevel: number = constants.MIN_OXYGEN_LEVEL;
     private passedPlayers: Set<Player> = new Set<Player>();
     private researchedPlayers: Set<Player> = new Set<Player>();
+    private draftedPlayers: Set<Player> = new Set<Player>();
     private originalBoard = new OriginalBoard();
     public spaces: Array<ISpace>;
     private temperature: number = constants.MIN_TEMPERATURE;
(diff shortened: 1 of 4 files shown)


src/Game.ts
```diff
@@ -441,13 +469,72 @@ export class Game {
...
       if (this.allPlayersHaveFinishedResearch()) {
         this.gotoActionPhase();
       }
     }
 
+    public playerIsFinishedWithDraftingPhase(player: Player, cards : Array<IProjectCard>): void {
+      this.draftedPlayers.add(player);
+      // Store cards for next player
+      this.unDraftedCards.set(player,cards);
+      //cards.forEach((card) => console.log("Phase "+ this.draftRound + " unDraftedCards for player " + player.name + " "+card.name));
```
bafolts: I would remove these debugging comments for the final commit.

src/Game.ts
```diff
@@ -441,13 +469,72 @@ export class Game {
...
+     
+      if (this.allPlayersHaveFinishedDraft() && this.draftRound < 3) {
```
bafolts: With this logic it seems like the game would get stuck when `this.allPlayersHaveFinishedDraft() === false`. When all players have not finished with the draft I would think the game moves on to the next player.

src/Player.ts
```diff
@@ -435,14 +436,44 @@ export class Player {
...
+            this.draftedCards.push(foundCards[0]);
+            cards = cards.filter((card) => card !== foundCards[0]);
+            game.playerIsFinishedWithDraftingPhase(this,cards);
+            return undefined;
+          }, 1, 1
+        )
+      );  
+  }  
+
+    public runResearchPhase(game: Game, draftVariant: Boolean): void {
```
bafolts: I would use `boolean` instead of `Boolean`.

src/Game.ts
bafolts: I see what happens here after further reading of this pull request. When not all players have finished draft I would `return` to exit the function early.

PR #137, opened 2019-12-24
Issue 121 preferences

alrusdi:
I've added floating sidebar with "fast travel" buttons and Preferences dialog to Hide/Show some elements on the page as @vincentneko suggested


src/components/Preferences.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    static saveValue(name: string, val: string): void {
+        const date = new Date();
+        const value = val;
+    
+        // Set it expire in 1 year
+        date.setTime(date.getTime() + (365 * 24 * 60 * 60 * 1000));
+    
+        // Set it
+        document.cookie = name+"="+value+"; expires="+date.toUTCString()+"; path=/";
```
bafolts: I would prefer to use `localStorage`.

PR #143, opened 2019-12-25
Fix ProtectedHabitas and Pets behavior on some cards

alrusdi:
We found some bugs with ProtectedHabitas and Pets
Fixes and additional test are in this PR

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -80,6 +80,7 @@ export class Player {
     public isAnyOtherPlayerHasPlants(game: Game): boolean {
       for (const player of game.getPlayers()) {
         if (player.id === this.id) continue;
+        if (player.hasProtectedHabitats()) continue;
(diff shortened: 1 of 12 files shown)


src/Player.ts
```diff
@@ -80,6 +80,7 @@ export class Player {
     public isAnyOtherPlayerHasPlants(game: Game): boolean {
```
bafolts: I will make this function `getOtherPlayersWithPlants`. Then this array can be used where below you have:

```js
      const playersToRemovePlantsFrom = game.getPlayers().filter(
        (p) => p.id != player.id && p.plants > 0 && ! p.hasProtectedHabitats()
      )
```

Any place where the count is needed the length of the array can be checked. Otherwise we will have to keep all of these functions in sync as we change this function in the future.

Issue #138, opened 2019-12-24
Improve VP breakdown


vincentneko:
VP breakdown is a great feature.
I would like to list all cards that have a variable VP contribution like security fleet, or animals cards along with the VP associated to them on the endgame page.

This will need updating all onGameEnd methods to return the value.

bafolts:
Renaming the `onGameEnd` method to `getVictoryPoints` would make sense. I am fairly certain most `onGameEnd` implementations are use for victory point calculations anyway.

vincentneko:
I did not saw your changes and I tried to upload my changes, the branch is a mess now, sorry.
I will try to revert my changes.

alrusdi:
Implemented and merged

PR #146, opened 2019-12-26
issue #138

vincentneko:
Now that we have getVictoryPoints, we can remove nonNegativeVPIcon variable and add this function to all cards using this variable. This will also add those cards in the breakdown VP details list.
Vitor is updated in this PR to reflect this change.

Please leave this branch up for the time I update all cards with nonNegativeVPIcon variable.


vincentneko:
> I dislike HTML part, bot other changes are very nice. Let's do better markup latter.

I agree on this !

Issue #151, opened 2019-12-26
Bug with Valley Trust Corporation


vincentneko:
At the beginning of every generation, the game ask you to choose a prelude card among 3 and play it. (this should happen only on generation 1)

I don't think it's related to this corp. The generation counter does not increment, at least in a solo player game.

alrusdi:
Fixed, merged

PR #152, opened 2019-12-26
issue #151

vincentneko:


src/Game.ts
```diff
@@ -383,6 +382,7 @@ export class Game {
       if (this.gameIsOver()) {
         this.gotoFinalGreeneryPlacement();
       } else {
+        this.generation++;
```
alrusdi: I think generation must be incremented before
if (this.gameIsOver()) {
this leads to that bug with never ending solo game

vincentneko:
> Drafting code needs more tests. @vincentneko can you please add one at least for this issue

Yes you are right, I will try to put something up.

src/Game.ts
vincentneko: Not sure about this.
I've made the change I want to push in this PR and my solo game went to generation 12, you can see it in the screenshot I've sent.

And also, before implementing drafting option, I think generation was incremented in gotoResearchPhase.

vincentneko:
> > Drafting code needs more tests. @vincentneko can you please add one at least for this issue
> 
> Yes you are right, I will try to put something up.

I've tested generation increment for solo and 2 players (draft and no draft options)
Not sure how I can test drafting by itself though. Is there a way to answer the SelectCard function the player has to choose when in drafting ?

src/Game.ts
alrusdi: I think it is good start. We will expand coverage someday.
You can set any answer to Input function by calling its cb function (CallBack). See example in Plantation.spec.ts

PR #166, opened 2019-12-29
Issue 165 bugfixes

vincentneko:


src/Player.ts
```diff
@@ -1245,24 +1245,22 @@ export class Player {
       //Post Action (after some specific prelude cards have been played)
       if (this.postAction && this.getPlayableCards(game).length > 0) {
         const input = this.playProjectCard(game);
-        input.onend = () => {
-          this.actionsTakenThisRound++;
-          this.takeAction(game);
-        };
+        input.onend = () => {};
```
bafolts: The `onend` property of `PlayerInput` is optional. You can not set the `onend` property altogether instead of setting to a noop.

src/cards/AdvancedEcosystems.ts
```diff
@@ -8,13 +8,13 @@ export class AdvancedEcosystems implements IProjectCard {
...
+    public canPlay(player: Player): boolean {
+      return (player.getTagCount(Tags.PLANT) > 0 &&
+      player.getTagCount(Tags.MICROBES) > 0 &&
+      player.getTagCount(Tags.ANIMAL) > 0);
     }
     public play(player: Player) {
-      if (player.getTagCount(Tags.PLANT) === 0 ||
-            player.getTagCount(Tags.MICROBES) === 0 ||
-            player.getTagCount(Tags.ANIMAL) === 0) {
+      if ( ! this.canPlay(player)) {
```
bafolts: I would remove this check and the corresponding throw. This pattern was in place before `canPlay` and I never removed all of them. The `canPlay` function is always called prior to playing and it is assumed if `play` is called the card can be played.

src/Player.ts
vincentneko: Thanks for the review. So, just to be sure, I can remove this line completly ?
input.onend = () => {};

Issue #164, opened 2019-12-29
Solo game rules


vincentneko:
According to the rules book, which is not compatible with our implementation (see "Should finish solo game before last generation if Mars is already terraformed" test)

5) You always play 14 generations (marked ’solo’).
In order to win, you need to complete terraforming (i.e
making the three global parameters reach their goal) before
the end of generation 14. After generation 14, you may
convert plants into greenery tiles, following normal rules
but without raising the oxygen, and finally you score VPs
to get as high a score as possible. If you have not completed
terraforming by the end of generation 14, you lose.

If prelude is used, it is 12 generations.

alrusdi:
I've misread this rule, sorry. Will fix it ASAP.

PR #167, opened 2019-12-30
Fix too erly solo game final; #164

alrusdi:


alrusdi:
> Looking into the code, I think isSoloModeWin function is no longer used.
> Can you please double check ?

It is used in server.ts near 297 line

vincentneko:
> > Looking into the code, I think isSoloModeWin function is no longer used.
> > Can you please double check ?
> 
> It is used in server.ts near 297 line

Ok got it thanks.

If I understand correctly, the solo game could stop before end of gen 12 or 14
Should it be 

`return this.preludeExtension ? this.generation === 12 : this.generation === 14;`

instead of

```
    public isSoloModeWin(): boolean {
      if ( ! this.marsIsTerraformed()) return false;
      return this.preludeExtension ? this.generation <= 12 : this.generation <= 14;
    }
```

alrusdi:
> Should it be
> 
> `return this.preludeExtension ? this.generation === 12 : this.generation === 14;`

I believe it's not. We use this function to detect if user win or lose on the final page. To win player must terraform mars and to do it in less than 14 (12) generations

vincentneko:
> > Should it be
> > `return this.preludeExtension ? this.generation === 12 : this.generation === 14;`
> 
> I believe it's not. We use this function only to detect if user win or lose on the final page. To win player must terraform mars AND to do it in less than 14 (12) generations

Ok my bad

PR #178, opened 2020-01-02
Fix solo game bug

vincentneko:


src/OriginalBoard.ts
```diff
@@ -157,7 +157,10 @@ export class OriginalBoard {
     public getRandomCitySpace(offset: number = 0): Space {
         while (true) {
             let space = this.getRandomSpace(offset);
-            if (space.tile === undefined && space instanceof Land && space.id !==  SpaceName.NOCTIS_CITY  ) {
+            if (space.tile === undefined 
+                && space instanceof Land 
+                && space.id !==  SpaceName.NOCTIS_CITY
+                && space.spaceType !== SpaceType.COLONY ) {
```
bafolts: The check for `space instanceof Land` should already be enough since `Colony` is not a `Land`. The issue has something to do with space id 37. The adjacent spaces to this space is two oceans and another space. In the situation that loops indefinitely the only land space is occupied.

vincentneko: You are right, I've overlooked this.

If first random city is in space 27 and its associated greenery is in space 36, then if second random city is in space 37, there is no free space for it's greenery

I will add a check for getRandomCitySpace

Issue #180, opened 2020-01-05
Aerobraked ammonia asteroid


KiqueGar:
Cannot be payed with titanium

vincentneko:
This one is missing space tag, PR incoming

PR #181, opened 2020-01-05
Issue #180 fix

vincentneko:


src/cards/AerobrakedAmmoniaAsteroid.ts
```diff
@@ -8,7 +8,7 @@ import { Resources } from "../Resources";
 
 export class AerobrakedAmmoniaAsteroid implements IProjectCard {
     public cost: number = 26;
-    public tags: Array<Tags> = [Tags.SCIENCE];
+    public tags: Array<Tags> = [Tags.SCIENCE, Tags.SPACE];
```
bafolts: I think this card should *only* have `Tags.SPACE` instead of `Tags.SCIENCE`.

PR #196, opened 2020-01-13
Venus next

vincentneko:
Thanks @alrusdi  for the Venus board and @pierrehilbert for the help on project cards.
Can you please review and maybe try a couple of game with the branch before merging ?


alrusdi:
I will check cards from "AerialMappers" to "LuxuryFoods"

alrusdi:
I've briefly checked all the cards of expansion and collected some bugs in this issue: #197

bafolts:
Will need the build to pass as a first step. Seems something is wrong with the tsconfig. I will take a look while on first pass of reviewing. There are a lot of files here so it may take some time. Assuming the underlying `Game` and `Player` haven't changed much there isn't much impact to the original game. I will do a first pass within the next two days.

src/Dealer.ts
```diff
@@ -565,14 +676,20 @@ export class Dealer {
...
         this.deck = this.shuffleCards(ALL_PROJECT_CARDS);
         if (this.usePreludeExtension) {
                 this.preludeDeck = this.shuffleCards(ALL_PRELUDE_CARDS);
                 this.deck.push(...ALL_PRELUDE_PROJECTS_CARDS);
                 this.deck = this.shuffleCards(this.deck);
         }
+        if (this.useVenusNextExtension) {
+            this.deck.push(...ALL_VENUS_PROJECTS_CARDS);
+            this.deck = this.shuffleCards(this.deck);
+    }
```
bafolts: The spacing here should be indented 4 more spaces.

src/Game.ts
```diff
@@ -319,19 +354,66 @@ export class Game {
...
+
+      if (this.first.getProduction(Resources.HEAT) > this.tempHeatProduction) {
+        this.first.setProduction(Resources.HEAT, this.tempHeatProduction - this.first.getProduction(Resources.HEAT));
+      }
+
+      // Check for Aphrodite corporation
+      if (this.tempVenusScaleLevel < this.venusScaleLevel 
+          && this.first.corporationCard !== undefined
+          && this.first.corporationCard.name === new Aphrodite().name) {
+            this.first.megaCredits += 2;
```
bafolts: Why is this needed here as well as where it is being incremented in `increaseVenusScaleLevel` function?

src/OriginalBoard.ts
```diff
@@ -44,6 +44,11 @@ export class OriginalBoard {
         this.spaces.push(new Colony(SpaceName.GANYMEDE_COLONY)); 
         this.spaces.push(new Colony(SpaceName.PHOBOS_SPACE_HAVEN));
 
+        this.spaces.push(new Colony(SpaceName.DAWN_CITY));
```
bafolts: This should only get added to the board if using the venus board.

src/Game.ts
```diff
@@ -44,18 +50,33 @@ export class Game {
     public gameLog: Array<String> = [];
     public gameAge: number = 0; // Each log event increases it
     private unDraftedCards: Map<Player, Array<IProjectCard>> = new Map ();
+    public awards = ORIGINAL_AWARDS;
+    public milestones = ORIGINAL_MILESTONES;
+
+    private tempMC: number = 0;
```
bafolts: Why do we have to store these temporarily? I don't remember the venus rules that well but don't see why we would need to reset these.

src/cards/ICard.ts
```diff
@@ -13,20 +13,21 @@ import { SelectPlayer } from "../inputs/SelectPlayer";
...
 export interface ICard {
     name: string;
     tags: Array<Tags>;
     play: (player: Player, game: Game) => PlayerInput | undefined;
-    action?: (player: Player, game: Game) => AndOptions | SelectAmount | SelectCard<IProjectCard> | SelectHowToPay | SelectPlayer | SelectSpace | undefined;
+    action?: (player: Player, game: Game) =>  OrOptions | SelectOption | AndOptions | SelectAmount | SelectCard<IProjectCard> | SelectHowToPay | SelectPlayer | SelectSpace | undefined;
     canAct?: (player: Player, game: Game) => boolean;
     getCardDiscount?: (player: Player, game: Game, card: IProjectCard) => number;
-    getRequirementBonus?: (player: Player, game: Game) => number;
+    getRequirementBonus?: (player: Player, game: Game, venusOnly?: boolean) => number;
```
bafolts: Instead of passing along `venusOnly` to this function signature I would create a `game.isVenusOnly()` or similar and leverage the `game` argument already being provided.

src/Game.ts
vincentneko: Because what happen during World Gouv phase is null and void. So if Venus was increased during WG phase by the player that owns Aphrodite, the bonus is rollbacked after WG phase, it needs to be reapplied.

src/Game.ts
vincentneko: What happen during World Gouv phase is null and void. For instance, if you place an ocean acting as World Gouv, and get 2 plants as placement bonus, you should not get them.
I think snapshoting and restoring resource is a good way to not interfere with existing code.

src/cards/ICard.ts
vincentneko: I'm not sure I understand what you mean. venusOnly parameter is used to trigger MorningStarInc corp capacity that gives the same bonus as Inventrix but only for cards with a Venus tag.

bafolts:
Your `.DS_STORE` file shouldn't be added. You can add that file to the `.gitignore`.

src/Game.ts
bafolts: Will this work for `Arctic Algae`? If a player acting as world government places an ocean the player of `Arctic Algae` would get two mega credits. Do they get these mega credits?

bafolts: If the first player plays` Arctic Algae` I think they should get their plants. Similar to how `Aphrodite` gets their bonus. Since when _anyone_ places an ocean tile, including the world government, they get 2 plants. This approach may work as a stop gap but we should make an issue to improve upon this. There are at least 2 cards that need to be called out so we may need to consider introducing a system for awarding bonus or ignoring bonus in certain situations or cards.

vincentneko: Yes you are right. I will put a quick fix for  Arctic Algae and we can do as you say for a more elegant solution.

PR #203, opened 2020-01-15
Update README.md

bafolts:
I started this project more than a year ago before the steam version was released. Upon news of the steam version being released I lost interest on this version and the project collected some dust for a while. While trying to keep my typescript skills up to date I got back into this project in the beginning of 2019. Once I got it to a functional place for the original board I was mostly losing interest and was planning to move on to a different board game in 2020. 

Somewhere along the way in 2019 more people showed up and started contributing! We need to include more updates to our `README.md` to call out that the prelude expansion and the venus expansion are now supported but I figured we should start with a contributors section since there are more people giving their time to this project for one reason or another. Your contributions have motivated me to keep with this project wherever it may go.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,4 +17,29 @@ npm install
 npm run start
 ```
 
-This will start the game server listening on the default port of 8080. If you then point a web browser to http://localhost:8080 you will be on the create game screen. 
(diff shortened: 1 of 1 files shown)


vincentneko:
Yes thats great. Can we add draft option too ?

PR #210, opened 2020-01-18
Select corps widget

vincentneko:
Allow to fine tune corporation list to play with, this is an option, disabled by default.
Add 3 new corps : Polyphemos, Utopia Invest and Lakefront Resorts


src/Game.ts
```diff
@@ -233,6 +242,12 @@ export class Game {
     private playCorporationCard(
         player: Player, corporationCard: CorporationCard
     ): void {
+
+      for (let corporation of [...ALL_CORPORATION_CARDS, ...ALL_PRELUDE_CORPORATIONS, ...ALL_VENUS_CORPORATIONS, ...ALL_COLONIES_CORPORATIONS]) {
```
bafolts: Why is this for loop needed? The argument `corporationCard` is already of type `CorporationCard`.

vincentneko: Yes but it comes from the "front" end, it only has the variables, not the functions.
If I omit this code, I get an error message like (don't remember exactly) "missing play function"

vincentneko:
I've added Stormcraft corporation.

vincentneko:
Please do not delete this branch, I have Arklight in progress.

PR #213, opened 2020-01-23
Add Arklight

vincentneko:
Last commit for Arklight corporation.

After this one I think a change is needed.
Initialy, the SelectCard, addResourceTo, etc... were made to interact with project cards.
Now that we have some corporations that can hold resources  (like Arklight, Stormcraft, Celestic), we need a better way to interact with them.
I will open an issue for it.


alrusdi:
Can we delete this branch now?

vincentneko:
Please wait, I will delete it after some checks. Thanks.

PR #220, opened 2020-01-25
Better AndOptions with only one save button!

bafolts:
I have been keeping this work on my computer and haven't had as much time to work on it. The overall pattern is in place of hiding the save button for children of `AndOptions`. The only work left is calling the save method of each child when the save button of the `AndOptions` is clicked. For `SelectSpace` we will have to store the selected space on the data model so that the parent `AndOptions` can call the save method for the situation where `SelectSpace` is a child of `AndOptions`. This makes the initial selection of corporation and project cards much simpler as it only involves one click.

### DIFF
--- a/src/components/SelectSpace.ts
+++ b/src/components/SelectSpace.ts
@@ -2,9 +2,8 @@
 import Vue, { VNode } from "vue";
 import { PlayerInputModel } from "../models/PlayerInputModel";
 
-
 export const SelectSpace = Vue.component("select-space", {
-    props: ["playerinput", "onsave", "showtitle"],
+    props: ["playerinput", "onsave", "showsave", "showtitle"],
     data: function () {
         return {};
     },
@@ -44,6 +43,7 @@ export const SelectSpace = Vue.component("select-space", {
                             (elTiles[j] as HTMLElement).onclick = null;
                         }
                         el_id = elTile.getAttribute("data_space_id");
+                        // TODO - Store this property for when showsave === false
                         this.onsave([[el_id]]);
                     }
                 }
--- a/src/components/SelectSpace.ts
+++ b/src/components/SelectSpace.ts
@@ -31,7 +31,7 @@ export const SelectSpace = Vue.component("select-space", {
         const clearAllAvailableSpaces = function() {
             const elTiles = document.getElementsByClassName("board_space");
             for (let i = 0; i < elTiles.length; i++) {
-                elTiles[0].classList.remove("board_space--available");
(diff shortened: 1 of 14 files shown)


src/components/SelectSpace.ts
```diff
@@ -43,8 +56,10 @@ export const SelectSpace = Vue.component("select-space", {
                         for (let j = 0; j < elTiles.length; j++) {
                             (elTiles[j] as HTMLElement).onclick = null;
                         }
-                        el_id = elTile.getAttribute("data_space_id");
-                        this.onsave([[el_id]]);
+                        this.$data.spaceId = elTile.getAttribute("data_space_id");
+                        if (this.showsave) {
```
bafolts: @alrusdi with this change when you pay for a standard project like `Greenery` you have to select the space and then click the save from the parent `OrOptions`. This leaves a time where the space is selected but not finalized. Can you add a selected space so that when the tile is clicked we add a `selected` class and style the space until saved.

alrusdi: Yes, i will do this change

bafolts: Start with most any company and then as a first action place a greenery.

bafolts: When you select the greenery project you will first have to click to select space and click on the space you would like to place the greenery. Then you will have to click the save from the OrOptions to pick the space.

alrusdi: I did it. But... But 5 clicks and 2 scrollings to just place the Greenery seems to be too much. Can we shortcut this scenario somehow?

src/components/SelectSpace.ts
```diff
@@ -43,8 +57,9 @@ export const SelectSpace = Vue.component("select-space", {
                         for (let j = 0; j < elTiles.length; j++) {
                             (elTiles[j] as HTMLElement).onclick = null;
                         }
-                        el_id = elTile.getAttribute("data_space_id");
-                        this.onsave([[el_id]]);
+                        this.$data.spaceId = elTile.getAttribute("data_space_id");
+                        elTile.classList.add("board_space--selected")
+                        this.saveData();
```
bafolts: @alrusdi I called to `saveData` which should save the space as it did before this change. If this `SelectSpace` is a child of an `OrOptions` it will trigger the `OrOptions`. If the `SelectSpace` is a child of an `AndOptions` the selected space will get stored with the `board_space--selected` until any remaining options are populated. If a user wants to change their selected space before they click save for an `AndOptions` they can click `Select Space` again and select another space.

PR #225, opened 2020-01-27
Add 4 new corporations

vincentneko:
Arcadian Communities
Recyclon
Splice
Pristar


src/Game.ts
```diff
@@ -858,6 +858,10 @@ export class Game {
       if (space.player !== undefined && space.player !== player) {
         throw new Error('This space is land claimed by ' + space.player.name);
       }
+      // Arcadian Communities
+      if (space.player !== undefined && space.player === player && player.isCorporation(CorporationName.ARCADIAN_COMMUNITIES)) {
```
bafolts: There is an `onTilePlaced` which should be able to handle this logic and avoid one off logic in `Game`.

src/OriginalBoard.ts
```diff
@@ -216,6 +216,15 @@ export class OriginalBoard {
       ) !== undefined;
     }
 
+    public getAvailableSpacesForMarker(player: Player): Array<ISpace> {
```
bafolts: I would filter this list to remove duplicates.

src/OriginalBoard.ts
```diff
@@ -216,6 +216,15 @@ export class OriginalBoard {
...
+      return this.getAvailableSpacesOnLand(player)
+      .filter(
+          (space) => this.getAdjacentSpaces(space).find(
+              (adj) => adj.player === player
```
bafolts: The marker for `Arcadian Communities` must be placed adjacent to one of your own tiles. This conditional must also check that `adj.tile` is not `undefined`.

src/Player.ts
```diff
@@ -849,7 +854,10 @@ export class Player {
             if (somePlayer.corporationCard !== undefined &&
                 somePlayer.corporationCard.onCardPlayed !== undefined
             ) {
-              somePlayer.corporationCard.onCardPlayed(somePlayer, game, selectedCard);
+              const actionFromPlayedCard: OrOptions | void = somePlayer.corporationCard.onCardPlayed(this, game, selectedCard);
```
bafolts: In my other PR for #220 where I hide the save button for children of `AndOptions` or even as a user of the game today, this long list of potential `OrOptions` that all need to be done for the parent `AndOptions` to save is most likely very confusing. This makes sense to add another `OrOptions` here but we may want to reconsider the usage of and `AndOptions` with a list of `OrOptions` here.

src/cards/promo/Recyclon.ts
```diff
@@ -0,0 +1,43 @@
...
+    public tags: Array<Tags> = [Tags.MICROBES, Tags.STEEL];
+    public startingMegaCredits: number = 38;
+    public resourceType: ResourceType = ResourceType.MICROBE;
+
+    public play(player: Player) {
+        player.setProduction(Resources.STEEL);
+        player.addResourceTo(this);
+        return undefined;
+    }
+    public onCardPlayed(player: Player, _game: Game, card: IProjectCard) {
```
bafolts: For corporation cards you need to check that the player playing the card owns the corporation.

https://github.com/bafolts/terraforming-mars/blob/master/src/cards/corporation/InterplanetaryCinematics.ts#L14

This is because we loop through each corporation card a player may have to call `onCardPlayed`.

https://github.com/bafolts/terraforming-mars/blob/master/src/Player.ts#L852

src/cards/turmoil/Pristar.ts
```diff
@@ -0,0 +1,32 @@
...
+import { Tags } from "../Tags";
+import { CorporationName } from '../../CorporationName';
+import { ResourceType } from '../../ResourceType';
+
+export class Pristar implements CorporationCard {
+    public name: string = CorporationName.PRISTAR;
+    public tags: Array<Tags> = [];
+    public startingMegaCredits: number = 53;
+    public resourceType: ResourceType = ResourceType.PRESERVATION;
+    public lastGenerationTR: number = 0;
```
bafolts: This modifier level for this property should be `private`.

src/Game.ts
vincentneko: In onTilePlaced, the owner of the space is already set to player, so there is no way to tell if there was a marker before and therefore give 3 MC to player.

src/OriginalBoard.ts
vincentneko: The marker can also be placed adjacent to another marker.

src/Player.ts
vincentneko: Not sure I understand what you mean here.
There is already an array of OrOptions defined before.

const actionsFromPlayedCard: OrOptions[] = [];

Can we adress this in another issue ?

PR #238, opened 2020-02-02
Hellas and Elysium maps

vincentneko:
First draft for Hellas and Elysium boards,
The interrupt mechanism has been used for the bonus ocean on Hellas. Therefore, it could colide with an interrupt from Philares.
I think it would be better to adress this in another issue.


.vscode/symbols.json
```diff
@@ -0,0 +1 @@
...\\PaymentWidgetMixin.ts":"2020-01-09T17:51:36.153Z","src\\components\\PlayerHome.ts":"2020-01-24T11:23:56.320Z","src\\components\\PlayerInputFactory.ts":"2020-02-01T14:24:39.422Z","src\\components\\PlayerResource.ts":"2019-12-23T15:05:30.787Z","src\\components\\PlayerResources.ts":"2019-12-23T15:05:30.787Z","src\\components\\Preferences.ts":"2019-12-29T08:20:15.195Z","src\\components\\SelectAmount.ts":"2020-02-01T14:24:39.422Z","src\\components\\SelectCard.ts":"2020-02-01T14:24:39.423Z","src\\components\\SelectHowToPay.ts":"2020-02-01T14:24:39.423Z","src\\components\\SelectHowToPayForCard.ts":"2020-02-01T14:24:39.424Z","src\\components\\SelectOption.ts":"2020-02-01T14:24:39.425Z","src\\components\\SelectPlayer.ts":"2020-02-01T14:24:39.425Z","src\\components\\SelectPlayerRow.ts":"2019-12-23T15:05:30.789Z","src\\components\\SelectSpace.ts":"2020-02-01T14:24:39.426Z","src\\components\\Tile.ts":"2019-12-23T15:05:30.789Z","src\\components\\Venus.ts":"2020-01-15T12:41:43.240Z","src\\components\\WaitingFor.ts":"2020-02-01T14:24:39.426Z","src\\inputs\\AndOptions.ts":"2020-01-24T11:10:59.831Z","src\\inputs\\HowToPay.ts":"2020-01-09T17:51:36.156Z","src\\inputs\\OrOptions.ts":"2020-01-24T11:1
```
bafolts: I would add this file to the `.gitignore`.

src/Board.ts
```diff
@@ -0,0 +1,189 @@
+import { ISpace } from "./ISpace";
+import { Player } from "./Player";
+import { SpaceType } from "./SpaceType";
+import { SpaceBonus } from "./SpaceBonus";
+import { SpaceName } from "./SpaceName";
+import { TileType } from "./TileType";
+
+export class Space implements ISpace {
```
bafolts: I would consider making this an abstract class if it needs to be a class and not an interface.

src/Game.ts
```diff
@@ -78,13 +81,26 @@ export class Game {
       private draftVariant: boolean = false,
       public venusNextExtension: boolean = false,
       customCorporationsList: boolean = false,
-      corporationList: Array<CorporationCard> = []
+      corporationList: Array<CorporationCard> = [],
+      public boardName: string = "original"
```
bafolts: This should use the enum `BoardName`.

src/Game.ts
```diff
@@ -858,6 +874,35 @@ export class Game {
       if (space.tile !== undefined) {
         throw new Error('Selected space is occupied');
       }
+      // Hellas special requirements ocean tile
+      if (space.id === SpaceName.HELLAS_OCEAN_TILE 
+          && this.board.getOceansOnBoard() < constants.MAX_OCEAN_TILES
+          && this.boardName === BoardName.HELLAS) {
+        if (!player.canAfford(6)) {
+          throw new Error('You must be able to pay 6 to place a tile here');
```
bafolts: This error being thrown is most likely going to wreak havoc on the game. The tile being placed `SelectSpace` function is going to catch this error while we are processing the selected space. Depending on the card which played the action this could leave the game in an awkward state. This was why I introduced `canPlay` and `canAct` to avoid errors being introduced. We may want to make a separate issue for this as well. Ideally the logic would be updated so that this tile can not be selected unless you have 6 mega credits.

src/awards/DesertSettler.ts
```diff
@@ -0,0 +1,13 @@
+import { IAward } from "./IAward";
+import { Player } from "../Player";
+import { Game } from "../Game";
+
+export class DesertSettler implements IAward {
+    public name: string = "DesertSettler";
+    public description: string = "Most tiles south of the equator (the four bottom rows)"
+    public getScore(player: Player, game: Game): number {
+        return game.board.spaces
+          .filter((space) => (space.player !== undefined && space.player === player)
```
bafolts: If these spaces need to have tiles this should check that the tile is not undefined.

src/awards/EstateDealer.ts
```diff
@@ -0,0 +1,16 @@
...
+import { Player } from "../Player";
+import { Game } from "../Game";
+import { SpaceType } from "../SpaceType";
+
+export class EstateDealer implements IAward {
+    public name: string = "EstateDealer";
+    public description: string = "Most tiles adjacent to ocean tiles"
+    public getScore(player: Player, game: Game): number {
+        return game.board.spaces
+          .filter((space) => (space.player !== undefined && space.player === player)
```
bafolts: If these need to be tiles we should check that tile is defined as well.

src/awards/Excentric.ts
```diff
@@ -0,0 +1,18 @@
+import { IAward } from "./IAward";
+import { Player } from "../Player";
+import { Game } from "../Game";
+
+export class Excentric implements IAward {
+    public name: string = "Excentric";
+    public description: string = "Most resources on cards"
+    public getScore(player: Player, _game: Game): number {
+        let score: number = 0;
+        if (player.corporationCard !== undefined && player.getResourcesOnCard(player.corporationCard) > 0) {
```
bafolts: To avoid calling `player.getResourcesOnCard` twice if `player.corporationCard` is defined you can add `player.getResourcesOnCard` to `score` since it should never be negative.

src/cards/LavaFlows.ts
```diff
@@ -16,12 +16,24 @@ export class LavaFlows implements IProjectCard {
     public name: string = "Lava Flows";
     public cardType: CardType = CardType.EVENT;
     public static getVolcanicSpaces(player: Player, game: Game): Array<ISpace> {
+        if (game.boardName === "original") {
```
bafolts: Use the `enum` value here instead of the string when comparing to `boardName`.

src/cards/NoctisCity.ts
```diff
@@ -22,7 +24,14 @@ export class NoctisCity implements IProjectCard {
         }
         player.setProduction(Resources.ENERGY,-1);
         player.setProduction(Resources.MEGACREDITS,3);
-        game.addCityTile(player, noctisSpace.id);
-        return undefined;
+        if (game.boardName === "original") {
```
bafolts: Always use the enum value when comparing to board name. If you want to avoid using enum you could define the board name type as `private boardName: "original" | "hellas" | "elysium"`

src/components/CreateGameForm.ts
```diff
@@ -135,12 +138,23 @@ export const CreateGameForm = Vue.component("create-game-form", {
...
                 <label>
                         <input type="checkbox" class="nes-checkbox" v-model="customCorporationsList"  v-on:click="toggleDisplayed()" />
                         <span>Use custom Corporation list ?</span>
                 </label>
+                
+                <br>
+                <label>Board:</label>
+                <div class="nes-select">
+                    <select :id="board" v-model="board">
+                        <option value="original">Original board</option>
```
bafolts: To make sure we stay in sync with the enum of `BoardName` here you should leverage the enum values here.

src/milestones/Diversifier.ts
```diff
@@ -0,0 +1,23 @@
...
+import { Tags } from '../cards/Tags';
+import { CardType } from "../cards/CardType";
+
+export class Diversifier implements IMilestone {
+    public name: string = "Diversifier";
+    public description: string = "Requires that you have 8 different tags in play"
+    public canClaim(player: Player, _game: Game): boolean {
+        let allTags: Array<Tags> = [];
+        if (player.corporationCard !== undefined && player.corporationCard.tags.length > 0) {
+          allTags.push(...player.corporationCard.tags);
```
bafolts: I don't think what is here will work as expected. I would use a set for simplicity. Add the tag name as a key of the set and then check that there are more than 8 keys in the set.

```
var set = {};
set[tags[0]] = 1;
return Object.keys(set) >= 8;
```

src/milestones/Legend.ts
```diff
@@ -0,0 +1,13 @@
+import { IMilestone } from "./IMilestone";
+import { Player } from "../Player";
+import { Game } from "../Game";
+import { CardType } from '../cards/CardType';
+
+export class Legend implements IMilestone {
+    public name: string = "Legend";
+    public description: string = "Requires that you have player 5 events"
```
bafolts: `s/player/played`

src/milestones/PolarExplorer.ts
```diff
@@ -0,0 +1,16 @@
+import { IMilestone } from "./IMilestone";
+import { Player } from "../Player";
+import { Game } from "../Game";
+
+export class PolarExplorer implements IMilestone {
+    public name: string = "Polar Explorer";
+    public description: string = "Requires that you have 3 tiles on the two bottom rows"
+    public canClaim(player: Player, game: Game): boolean {
+        return game.board.spaces
+          .filter((space) => (space.player !== undefined && space.player === player)
```
bafolts: This needs to check that there are tiles on the spaces.

src/milestones/Tactitian.ts
```diff
@@ -0,0 +1,20 @@
...
+import { HTML_DATA } from '../HTML_data';
+
+export class Tactitian implements IMilestone {
+    public name: string = "Tactitian";
+    public description: string = "Requires that you have 5 cards with requirements in play"
+    public canClaim(player: Player, _game: Game): boolean {   
+       let score: number = 0; 
+       player.playedCards.filter((card) => card.cardType !== CardType.PRELUDE).forEach((card) => {
+         let htmlData = HTML_DATA.get(card.name);
+         if (htmlData !== undefined && htmlData.includes('class="requirements')) {
```
bafolts: I would make an issue to implement this better. This should become a property on `ICard`. This will break the moment we change the classname and forget we have this here.

bafolts:
I will approve this pull request once the minor issues are addressed. At that point we can create the separate issues as they will make since at that point.

src/components/CreateGameForm.ts
vincentneko: Not sure how to pass the enum values to the vue js component, will carry on searching on this one

src/milestones/Diversifier.ts
vincentneko: Not exactly as you suggested, can you confirm that what I've done is ok ?

src/Game.ts
vincentneko: I think I've found a solution for this one.

src/milestones/Tactitian.ts
vincentneko: I was wondering if we could count the canPlay number of arguments but I got this error: 

Error processing input from player TypeError: 'caller', 'callee', and 'arguments' properties may not be accessed on strict mode functions or the arguments objects for calls to them
 
trying to do something like:

`if card.canPlay.length > 0 ...`

src/components/CreateGameForm.ts
vincentneko: Should be ok now

src/milestones/Tactitian.ts
bafolts: You _could_ count the number of arguments probably by using `card.canPlay.toSource()` and parsing the string representation of the function. I don't think we _should_ do that. Explicitly listing that the card has restrictions, or making the `canPlay` property optional and then checking that the property exists will be what we should do.

src/Game.ts
```diff
@@ -82,7 +82,7 @@ export class Game {
       public venusNextExtension: boolean = false,
       customCorporationsList: boolean = false,
       corporationList: Array<CorporationCard> = [],
-      public boardName: string = "original"
+      public boardName: string = BoardName.ORIGINAL
```
bafolts: This should be `public boardName: BoardName = BoardName.ORIGINAL`

.gitignore
```diff
@@ -4,3 +4,4 @@ node_modules/
 *.swp
 .nyc_output/
 styles.css
+.vscode/symbols.json
```
bafolts: You also need to remove this file from your commit.

src/HellasBoard.ts
```diff
@@ -106,4 +110,42 @@ export class HellasBoard extends Board{
             new Land(idx++, pos_x++, pos_y)
         );
     }    
+
+    public getAvailableSpacesForCity(player: Player): Array<ISpace> {
+        // Check for special tile
+        if (player.canAfford(6)) return super.getAvailableSpacesForCity(player);
+        // A city cannot be adjacent to another city
+        return this.getAvailableSpacesOnLand(player).filter(
```
bafolts: If the player can not afford 6 do you need to duplicate all the underlying logic? I would think you could filter on the super results as follows:

```
return super.getAvailableSpacesForCity(player).filter((space) => space.id !== SpaceName.HELLAS_OCEAN_TILE);
```

src/milestones/Tactitian.ts
vincentneko: I've made the  `canPlay` property optional.
Some cards need a` hasRequirements `anyway as a lot of cards does not have a requirement as defined by the game but need to be checked with `canPlay` nonetheless (cards that will reduce your energy production for instance)

PR #250, opened 2020-02-11
Leverage interrupt mechanism

vincentneko:


src/Game.ts
```diff
@@ -573,9 +573,12 @@ export class Game {
     public playerIsFinishedTakingActions(player: Player): void {
 
       // Interrupt hook
-      if (this.interrupt !== undefined) {
-        this.interrupt.player.setWaitingFor(this.interrupt.playerInput);
-        return;
+      if (this.interrupts.length > 0) {
+        let interrupt = this.interrupts.pop();
```
bafolts: Since the items are added with `push` using `shift` will ensure that the items in `interrupts` are processed in the order they were added.

src/Player.ts
```diff
@@ -1532,9 +1532,14 @@ export class Player {
     public takeAction(game: Game): void {
 
       //Interrupt action
-      if (game.interrupt !== undefined && game.interrupt.player === this) {
-        this.setWaitingFor(game.interrupt.playerInput);
-        return;
+      if (game.interrupts.filter(interrupt => interrupt.player === this).length > 0 ) {
```
bafolts: I would use `findIndex` and use the selector `interrupt => interrupt.player === this` to get the index of the _first_ interrupt. This will avoid having to filter the interrupts and the call to `pop` on that intermediate array. Using `pop` on that intermediate array also wont run the interrupts in the order they were added.

```
const interruptIndex = game.interrupts.findIndex(interrupt => interrupt.player === this);
if (interruptIndex !== -1) {
    this.setWaitingFor(game.interrupts.splice(interruptIndex, 1)[0].playerInput);
}
```

PR #260, opened 2020-02-14
Change GiantIceAsteroid

vincentneko:
Please do not merge.
This is a POC for interrupt rework. Please review and let me know if the implementation seems correct.
For testing purpose, GiantIceAsteroid is always available from card selection.
This needs to be tested in a multiplayer game.

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -608,6 +608,7 @@ export class Game {
         this.startActionsForPlayer(nextPlayer);
       } else {
         // Recursively find the next player
+        this.activePlayer = nextPlayer;
         this.playerIsFinishedTakingActions();
       }
     }
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -33,11 +33,7 @@ import { ElysiumBoard } from './ElysiumBoard';
 import { HellasBoard } from './HellasBoard';
 import { BoardName } from './BoardName';
 import { GiantIceAsteroid } from './cards/GiantIceAsteroid';
-
-export interface PlayerInterrupt {
-  player: Player,
-  playerInput: PlayerInput
-}
+import { PlayerInterrupt } from './interrupts/PlayerInterrupt';
 
 export class Game {
     public activePlayer: Player;
@@ -150,6 +146,10 @@ export class Game {
           this.playCorporationCard(player, new BeginnerCorporation());
         }
       }
+    }
+
+    public addInterrupt(interrupt: PlayerInterrupt): void {
+      this.interrupts.push(interrupt);
     }
 
     public getPreludeExtension(): boolean {
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -32,6 +32,7 @@ import {CardName} from './CardName';
 import { ElysiumBoard } from './ElysiumBoard';
 import { HellasBoard } from './HellasBoard';
 import { BoardName } from './BoardName';
+import { GiantIceAsteroid } from './cards/GiantIceAsteroid';
 
 export interface PlayerInterrupt {
   player: Player,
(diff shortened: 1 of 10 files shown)


alrusdi:
Giant asteroid seems to be playable. But little bit confusing. Can we change captions on buttons to "Select space for first ocean" and "Select space for second ocean"?

vincentneko:
> Giant asteroid seems to be playable. But little bit confusing. Can we change captions on buttons to "Select space for first ocean" and "Select space for second ocean"?

You are right, I will modify this.

src/Player.ts
```diff
@@ -272,6 +272,34 @@ export class Player {
...
+          return undefined;
+        }
+      );
+
+      let interrupt = {
+        player: this,
+        playerInput: selectOcean
+      };
+
+      selectOcean.onend = () => {
```
bafolts: This could be a recurring theme of ever interrupt. Creating a class for `Interrupt` would allow us to put the as the `onend` for every `Interrupt`.

src/cards/GiantIceAsteroid.ts
```diff
@@ -18,6 +16,10 @@ export class GiantIceAsteroid implements IProjectCard {
 
 
     public play(player: Player, game: Game) {
+
+        player.oceanSelector(game,'Select space for first ocean');
```
bafolts: I would separate adding the interrupt from creating the ocean selector.

```
game.addInterrupt(new SelectOcean("Select space for first ocean"));
game.addInterrupt(new SelectOcean("Select space for second ocean"));
```

would be ideal.

vincentneko: Not sure I understand what you mean here, can you make the change ?

vincentneko: Sleeping on this, I think I got your point. Please review and let me know if it's ok.

src/Game.ts
```diff
@@ -279,6 +281,7 @@ export class Game {
       for (let i = 0; i < 10; i++) {
         dealtCards.push(this.dealer.dealCard());
       }
+      dealtCards.push(new GiantIceAsteroid());
```
bafolts: This will need to be removed prior to merging.

src/Game.ts
```diff
@@ -570,7 +573,7 @@ export class Game {
       return players[(playerIndex + 1 >= players.length) ? 0 : playerIndex + 1];
     }
 
-    public playerIsFinishedTakingActions(player: Player): void {
+    public playerIsFinishedTakingActions(_player: Player): void {
```
bafolts: Instead of marking this argument as not used you should remove the argument or always pass in `game.activePlayer`.

src/Game.ts
```diff
@@ -980,6 +983,9 @@ export class Game {
       this.addTile(player, spaceType, this.getSpace(spaceId), {
         tileType: TileType.OCEAN
       });
+      if (this.pendingOceans > 0) {
+        this.pendingOceans--;
```
bafolts: This should be deprecated in `SelectOcean` since that is where it is incremented.

bafolts: Speaking of, this needs to be incremented in `SelectOcean`'

vincentneko: I had to create an intermediate method in game `addOceanInterrupt ` to call `SelectOcean` only if an ocean space is available on the board.

src/Game.ts
```diff
@@ -598,7 +608,7 @@ export class Game {
         this.startActionsForPlayer(nextPlayer);
       } else {
         // Recursively find the next player
-        this.playerIsFinishedTakingActions(nextPlayer);
+        this.playerIsFinishedTakingActions();
```
bafolts: If the next player has passed I think this will infinitely loop since we no longer pass in `nextPlayer`. This will have to be updated to not be recursive but to iterate until the next player who has not passed is found.

bafolts: The rest of this pull request looks good but the `playerIsFinishedTakingActions` function will have to be updated to not use recursion. Or as a quick fix and to ensure we continue to process any interrupts which may show you you could set `this.activePlayer = nextPlayer` prior to calling `this.playerIsFinishedTakingActions();`

vincentneko: You're right, great you spotted this !

Issue #254, opened 2020-02-12
Ecological Zone / Viral Enhancers


dabewi:
Hi,

the combination of Ecological Zone and Viral Enhancers might be buggy:

Viral Enhancers should generate 1 extra animal per tag on another card. Ecological Zone has two tags (animal and plant), so it should get 2 extra ressources from Viral Enhancers. Ecological Zone itself should also get two ressources from it's own two tags. Right know if it is played when Viral Enhancers are already played it only gets two ressources instead of 4 it should get (2 from the card itself and 2 from Viral Enhancers).

Screenshot: https://abload.de/img/2020-02-12_15h39_11qdjgr.png

Maybe the method to grant extra ressources is only triggered once and not per tag?

dabewi:
> Hi,
> 
> the combination of Ecological Zone and Viral Enhancers might be buggy:
> 
> Viral Enhancers should generate 1 extra animal per tag on another card. Ecological Zone has two tags (animal and plant), so it should get 2 extra ressources from Viral Enhancers. Ecological Zone itself should also get two ressources from it's own two tags. Right know if it is played when Viral Enhancers are already played it only gets two ressources instead of 4 (2 from the card itself and 2 from Viral Enhancers).
> 
> Screenshot: https://abload.de/img/2020-02-12_15h39_11qdjgr.png
> 
> Maybe the method to grant extra ressources is only triggered once and not per tag?

bafolts:
Both the `Ecological Zone` implementation and the `Viral Enhancers` implementation do seem wrong.

For `Ecological Zone` we should add one for each tag found which is the easier of the two fixes.

https://github.com/bafolts/terraforming-mars/blob/master/src/cards/EcologicalZone.ts#L41

For `Viral Enhancers` we also need to apply the effect per tag and not overall. In the situation where there is more than one tag that adds to the effect we will have to use an `AndOptions` so the user can select if they want a plant or to add a resource to the card.

https://github.com/bafolts/terraforming-mars/blob/master/src/cards/ViralEnhancers.ts#L17

This will lead to an `AndOptions` with multiple `OrOptions` which should work fine. With the recent change to not show the child save buttons the user should only have to click save once after selection all options.

vincentneko:
Fixed and merged

PR #265, opened 2020-02-15
Issue #254

vincentneko:
Fix a bug in Domed crater requirements
Fix Ecological zone
Quick and dirty fix for Viral Enhancers, if x tags are played, you either choose to add x resource to card or gain x plants.
The only card impacted by this is Ecological zone. All other cards either have only one tag or cannot hold resources.

### DIFF
--- a/src/cards/EcologicalZone.ts
+++ b/src/cards/EcologicalZone.ts
@@ -38,15 +38,13 @@ export class EcologicalZone implements IProjectCard {
     return this.hasGreeneryTile(player, game);
   }
   public onCardPlayed(player: Player, _game: Game, card: IProjectCard): void {
-    if (card.tags.indexOf(Tags.ANIMAL) !== -1 ||
-          card.tags.indexOf(Tags.PLANT) !== -1) {
-      player.addResourceTo(this);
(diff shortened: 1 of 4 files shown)


src/cards/EcologicalZone.ts
```diff
@@ -38,15 +38,13 @@ export class EcologicalZone implements IProjectCard {
...
-          card.tags.indexOf(Tags.PLANT) !== -1) {
-      player.addResourceTo(this);
-    }
+      player.addResourceTo(this, card.tags.filter((tag) => tag === Tags.ANIMAL || tag === Tags.PLANT).length);
   }
   public getVictoryPoints(player: Player): number {
     return Math.floor(player.getResourcesOnCard(this) / 2);
   }
   public play(player: Player, game: Game) {
+    player.addResourceTo(this, 2);
```
bafolts: `onCardPlayed` gets called whenever a card is played. This change will end up giving a player 4 resources instead of only 2.

vincentneko: Ok, thanks for the clarification, I was not sure about this.

Issue #271, opened 2020-02-17
Credicor doesn't work on standard projects


dabewi:
Hi,

we are currently playing a 2 player-game on the Elysium board. My opponent plays credicor and he noticed this: While the refund of credicor (4 MC for every card or standard project > 20 MC) works for cities it doesn't work for greeneries.

Can someone confirm that?

vincentneko:
I've just tested it and yes, I confirm I did not get a 4 MC cashback.
But it's also the case with city project, which is a good news imho...
Will investigate this.

vincentneko:
Got it !

`payForStandardProject` is only looping through played cards, not corporations.

Fix incomming

vincentneko:
Fixed and merged !

PR #274, opened 2020-02-18
Issue #271 Fix credicor standard projects bug

vincentneko:


bafolts:
This fix is right, but while we are here we should also add this hook here as well where it is missing.

https://github.com/bafolts/terraforming-mars/blob/master/src/Player.ts#L965

vincentneko:
> This fix is right, but while we are here we should also add this hook here as well where it is missing.
> 
> https://github.com/bafolts/terraforming-mars/blob/master/src/Player.ts#L965

Not sure I understand what you mean.
The change I've added is right at this position:

```
    private payForStandardProject(
        projectType: StandardProjectType,
        megaCredits: number,
        heat: number): void {
      this.megaCredits -= megaCredits;
      this.heat -= heat;

      if (this.corporationCard !== undefined && this.corporationCard.onStandardProject!== undefined) {
        this.corporationCard.onStandardProject(this, projectType);
      }

      for (const playedCard of this.playedCards) {
        if (playedCard.onStandardProject !== undefined) {
          playedCard.onStandardProject(this, projectType);
        }
      }
    }
```

Issue #277, opened 2020-02-19
Flodding allows you to remove MC from "owner" of Ocean tiles


K-Hartmann:
Since Ocean tiles don't have any ownership marker or owner, flooding should not work for already placed ocean tiles. It does though, as seen in the screenshot.
![2020-02-19_13h21_00](https://user-images.githubusercontent.com/41374984/74833911-b2aa9400-531a-11ea-92b5-1a19b4b74373.png)

alrusdi:
I will try to fix it

alrusdi:
Should be fixed now

PR #279, opened 2020-02-19
Fix Flooding card action; #277

alrusdi:


alrusdi:
> I think it would be good to fix the problem at the source as well.
> 

Indeed. I will fix it

PR #289, opened 2020-02-21
Colonies

vincentneko:


server.ts
```diff
@@ -513,7 +515,7 @@ function serveAsset(req: http.IncomingMessage, res: http.ServerResponse): void {
     res.write(fs.readFileSync('dist/main.js'));
   } else if (req.url === '/assets/Prototype.ttf') {
     res.write(fs.readFileSync('assets/Prototype.ttf'));
-  } else if (req.url.endsWith('.png')) {
+  } else if (req.url.endsWith('.png') || req.url.endsWith('.jpg') ) {
```
bafolts: A header is sent of `Content-Type: image/png` a few lines down. For a jpeg this is the wrong header to send.

src/Game.ts
```diff
@@ -995,6 +1041,9 @@ export class Game {
       this.addTile(player, spaceType, this.getSpace(spaceId), {
         tileType: TileType.OCEAN
       });
+      if (this.pendingOceans > 0) {
+        this.pendingOceans--;
```
bafolts: It doesn't make sense to decrease `pendingOceans` here since we didn't increase it around here. This will most likely lead to bugs where the `pendingOcean` count is wrong.

src/cards/colonies/Aridor.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    public initialAction(_player: Player, game: Game) {
+        let addColony = new OrOptions();
+        addColony.title = "Aridor first action - Select colony tile to add";
+        game.colonyDealer.discardedColonies.forEach(colony => {
+          const colonySelect =  new SelectOption(
+            colony.name + " - (" + colony.description + ")", 
+            () => {
+                game.colonies.push(colony);
+                game.colonies.sort((a,b) => (a.name > b.name) ? 1 : -1);
```
bafolts: This sort function is not going to sort in any way that seems helpful. I am assuming this wants to sort the colonies by name alphabetically. I would use:

```
a.sort((a, b) => a.name.localeCompare(b.name))
```

src/cards/colonies/Aridor.ts
```diff
@@ -0,0 +1,52 @@
...
+export class Aridor implements CorporationCard {
+    public name: string =  CorporationName.ARIDOR;
+    public tags: Array<Tags> = [];
+    public startingMegaCredits: number = 40;
+    public allTags = new Set();
+
+    public initialAction(_player: Player, game: Game) {
+        let addColony = new OrOptions();
+        addColony.title = "Aridor first action - Select colony tile to add";
+        game.colonyDealer.discardedColonies.forEach(colony => {
```
bafolts: Instead of `forEach` and then pushing each option I would use `addColony.options = game.colonyDealer.discardedColonies.map`

src/cards/colonies/SubZeroSaltFish.ts
```diff
@@ -0,0 +1,63 @@
...
+    public cardType: CardType = CardType.ACTIVE;
+    public resourceType = ResourceType.ANIMAL;
+
+    public canAct(): boolean {
+        return true;
+    }
+
+    private getPlayersWithPlantProduction(currentPlayer: Player, game: Game): Array<Player> {
+        var players = game.getPlayers().filter((p) => p.getProduction(Resources.PLANTS) > 0);
+        if (players.length > 1) {
```
bafolts: I have not read what this card does, but this function returns other players if there is more than 1 player with plant production. If there is only 1 player with plant production it could return the current player. I don't know if this is intentional or not.

src/cards/colonies/SubZeroSaltFish.ts
```diff
@@ -0,0 +1,63 @@
...
+        return game.getTemperature() >= -6 - (player.getRequirementsBonus(game) * 2);
+      }
+
+      public action(player: Player) {
+        player.addResourceTo(this);
+        return undefined;
+      }
+
+    public play(player: Player, game: Game) {
+        if (game.getPlayers().length == 1) return undefined;
```
bafolts: There is no linting on the project but we should always use `===` as it is more strict.

src/cards/prelude/PointLuna.ts
```diff
@@ -12,7 +12,9 @@ export class PointLuna implements CorporationCard {
     public startingMegaCredits: number = 41; //Should be 38 but the drawed card when played is payed 3 MC
     public onCardPlayed(player: Player, game: Game, card: IProjectCard) {
         if (player.corporationCard !== undefined && player.corporationCard.name === this.name && card.tags.indexOf(Tags.EARTH) !== -1) {
-			player.cardsInHand.push(game.dealer.dealCard());
+            for (let i = 0; i < card.tags.filter(tag => tag === Tags.EARTH).length; i++) {
```
bafolts: The for loop here will run through `card.tags.filter` during each iteration. You should store the variable for re-use:

```
for (let i = 0, length = card.tags.filter(tag => tag === Tags.EARTH).length; i < length; i++)
```

src/colonies/Colony.ts
```diff
@@ -0,0 +1,87 @@
...
+    onColonyPlaced: (player: Player, game: Game) => undefined | SelectSpace;
+    giveTradeBonus: (player: Player, game: Game) => void;
+    endGeneration: () => void;
+    increaseTrack(value?: number): void;
+    decreaseTrack(value?: number): void;
+}
+
+export abstract class Colony  {
+    public isActive: boolean = true;
+    public isVisited: undefined | Player = undefined;
```
bafolts: A better name for this property would be `visitor`.

src/colonies/ColonyDealer.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+    public drawColonies(players: number): Array<IColony> {
+        let count: number = players + 2;
+        if (players === 1) count = 4;
+        let tempDeck = this.shuffle(ALL_COLONIES_TILES);
+        for (let i = 0; i < count; i++) {
+            this.coloniesDeck.push(tempDeck.pop());
+        }    
+        this.discardedColonies.push(...tempDeck);
+        this.discardedColonies.sort((a,b) => (a.name > b.name) ? 1 : -1);
```
bafolts: This will not sort alphabetically which I am assuming are the intentions.

src/interrupts/SelectColony.ts
```diff
@@ -0,0 +1,41 @@
...
+import { IColony } from '../colonies/Colony';
+import { SelectOption } from '../inputs/SelectOption';
+
+export class SelectColony implements PlayerInterrupt {
+    public playerInput: PlayerInput;
+    constructor(
+        public player: Player,
+        public game: Game,
+        public openColonies: Array<IColony>,
+        public title?: string
```
bafolts: You can use `public tile: string = "Select where to build a colony";` and it will replicate what you have written below.

src/interrupts/SelectDiscard.ts
```diff
@@ -0,0 +1,31 @@
...
+            title = 'Select a card to discard';
+        }
+        this.playerInput = new SelectCard(title, player.cardsInHand, (foundCards: Array<IProjectCard>) => {
+            player.cardsInHand.splice(player.cardsInHand.indexOf(foundCards[0]), 1);
+            game.dealer.discard(foundCards[0]);
+            return undefined;
+          });
+        this.playerInput.onend = () => { 
+            if (game.activePlayer !== player) {
+                game.playerIsFinishedTakingActions();
```
bafolts: This can merge before the interrupts change. With the interrupts change this call to `takeAction` is unnecessary. Prior to the interrupts change this will most likely leave the game in an awkward state since there are high odds that `takeAction` will end up being called while another `takeAction` was called.

vincentneko:
> I only have minor changes suggested. This can merge prior to the interrupts change. Once merged I will make the modifications so this can work with the interrupts change. The interrupts change mostly just removes every `onend` property as it isn't necessary after the change.

Thanks for the review. I will try to address the requested changes asap so we can merge this in master.

src/cards/colonies/SubZeroSaltFish.ts
vincentneko: This one was shamelessly copied from fish

src/cards/colonies/SubZeroSaltFish.ts
vincentneko: Same, shamelessly copied from another card

src/colonies/ColonyDealer.ts
vincentneko: Commenting this line 
` this.coloniesDeck.sort((a,b) => (a.name > b.name) ? 1 : -1);`
is displaying colonies in a random order.

When it's not commented, the order colonies are displayed is always the same.

Also, I've noticed that colonies are "modified" between different games.
I've added `let gameColonies = ALL_COLONIES_TILES.slice();` to solve the issue. If you agree this one is necessary, when should do the same for `Dealer ` class as it could be the source of the bug reported.

src/Game.ts
vincentneko: I forgot to remove this one after refactoring the interrupt.

src/cards/colonies/Aridor.ts
vincentneko: Same as ColonyDealer.ts

src/cards/colonies/Aridor.ts
vincentneko: Same comment as the reply for ColonyDealer.ts

src/cards/colonies/Aridor.ts
vincentneko: map function is still unclear for me. I will try to understand it better tomorrow.

src/interrupts/SelectDiscard.ts
vincentneko: I've tested this one in a 2 player game, it was ok, but I guess the interrupts change will leave no chance for this to go wild.

src/cards/colonies/SubZeroSaltFish.ts
bafolts: This function seems wrong even for `Fish`. I will create an issue to address these.

server.ts
```diff
@@ -523,6 +525,16 @@ function serveAsset(req: http.IncomingMessage, res: http.ServerResponse): void {
...
     res.write(fs.readFileSync(reqFile))
+  } else if (req.url.endsWith('.jpg') ) {
+    const assetsRoot = path.resolve('./assets');
+    const reqFile = path.resolve(path.normalize(req.url).slice(1));
+
+    // Disallow to go outside of assets directory
+    if ( ! reqFile.startsWith(assetsRoot) || ! fs.existsSync(reqFile)) {
+      return notFound(req, res);
+    }
+    res.setHeader('Content-Type', 'image/jpg');
```
bafolts: This header should be `image/jpeg`

https://www.w3.org/Graphics/JPEG/

src/colonies/ColonyDealer.ts
bafolts: We will always need to call `slice()` to copy an array we do not want to modify. Most places in `Dealer` seem to be making a copy. The other bug reported may be somewhere outside of `Dealer`.

What order are you trying to display the colonies? The sort function `(a.name > b.name) ? 1 : -1)` _may_ sort the names alphabetically if there are no duplicate colony names. If there are duplicate colony names that sort function will fail.

vincentneko: About the order, I want colonies listed alphabetically by their name. So if present, Callisto will always displayed first, then Ceres, etc... There are no duplicate names.

I've solved the "modification" of colonies between games by nesting the ALL_COLONIES_TILES array inside `ColonyDealer` class. Not sure if we are facing the same problem with project cards.

Issue #272, opened 2020-02-17
Valley Trust provides four prelude cards instead of three


dabewi:
Hi,

Valley Trust seems to be buggy again. It provides 4 prelude cards and not only 3 as it should be:

https://abload.de/img/img_0839rtk6d.png

This error was already fixed (see https://github.com/bafolts/terraforming-mars/issues/215) but it seems that I'm not allowed to reopen that issue, therefore this new one.

Cheers!

bafolts:
I am able to recreate this on a 3 player game with player 1 as Valley Trust.

bafolts:
I am also able to recreate this on a 2 player game with player 1 as Valley Trust. It doesn't seem to reproduce on a 1 player game.

bafolts:
The problem here is that valley trust calls to our internal player `playCard` method which increments actions taken this turn and then calls again to take action. This is then putting valley trust back into play and allowing the user to select a card again. The fix here should be to implement a play card method that valley trust can use which doesn't increment the internal state of the player while playing the card.

bafolts:
Where there is smoke there is usually fire. The underlying issue here of a call to `takeAction` indirectly calling another `takeAction` will be a problem more than only here. To fix this issue fully it may finally be time to refactor `takeAction`. I see at least two other places that will end up with a similar issue:

https://github.com/bafolts/terraforming-mars/blob/master/src/Game.ts#L907
https://github.com/bafolts/terraforming-mars/blob/master/src/cards/promo/Philares.ts#L88

Depending on what initiates these actions we could end up calling into `takeAction` when we were already waiting for an action which will then lead us calling into `takeAction` twice. Not sure of solution yet but I have a grasp on the scope of the problem.

vincentneko:
I've tested Valley Trust in a 2 player game, this is now running as expected.
@bafolts  thanks for the `takeAction `rework, this is great !

PR #290, opened 2020-02-22
Use interrupts for sub-actions

bafolts:
To solve the problems we always seem to hit where the game gets stuck with no player taking an action and #272 where Valley Trust continually allows a user to perform the initial action I have refactored most actions to go through the interrupt system which was recently introduced.

This change removed the `onend` property of `PlayerInput` and replaces it with a callback parameter passed to `setWaitingFor`. This removes a lot of code duplication and should make the code easier to follow. Only one place now handles how many actions have been taken each turn and calling `takeAction`.

I still need to update the tests and do some final sanity checks and manual testing but putting this up as I work on those remaining steps.


src/Game.ts
```diff
@@ -826,14 +827,17 @@ export class Game {
...
-            this.board.getAvailableSpacesForOcean(player),
-            (space: ISpace) => {
-              this.addOceanTile(player, space.id);
-              return undefined;
-            }
-        );
+        this.interrupts.push({
+            player: player,
+            playerInput: new SelectSpace(
+                'Select space for ocean from temperature increase',
```
bafolts: I will use the select ocean interrupt for this.

bafolts:
A `hasPerformedInitialAction` shouldn't matter after this change. The moment the initial action is performed the actions taken this round is increased so we wont run it again. The flag would be redundant. Since only `takeAction` calls to `takeAction` we wont have two calls coming in like before.

bafolts:
I re-based this with the colony changes. As it will have wide reaching impacts to the game I will do another pass through tomorrow prior to merging.

bafolts:
OK. This change isn't that large in terms of lines of codes changed, it actually removes more lines than it adds, but is far reaching in it changes how every action is taken. This change should make the code easier to read and allows us to work with the interrupt system. Any issues this does introduce should be easy to fix and I plan to fix them quickly if and when they come in. With this change we can start moving a lot of the card actions to use interrupt which should be easier to test and allow us to add more game expansions if they show up.

bafolts:
:see_no_evil: :hear_no_evil: :speak_no_evil:

Issue #269, opened 2020-02-17
Suggestion: Keep the opponent's cards visable


dabewi:
Hi!

After an opponent made his move his ressources become invisible as his cards are getting collapsed. I'd like to suggest a functionality that keeps the opponents cards visible until the player manually collapses them. Because when I play I want to constantly check which of the drafted cards are beneficial for which opponent.

Maybe this could be made with a small checkbox "Keep the opponents cards visable", so that every player can decide whether he prefers the actual behaviour or not.

Cheers!

alrusdi:
Nice idea. it's little annoying when just opened opponent info collapsing when someone do something.

alrusdi:
Implemented, merged

PR #304, opened 2020-02-24
Global state for other players visibility; #269

alrusdi:


src/components/OtherPlayer.ts
```diff
@@ -8,20 +8,19 @@ export const OtherPlayer = Vue.component("other-player", {
...
-    data: function () {
-        return {
-            displayed: false
-        };
-    },
     methods: {
-        toggleDisplayed: function () {
-            this.displayed = !this.displayed;
+        toggleVisible: function () {
+            (this.$root as any).otherPlayersVisibility[this.player.id] = !(this.$root as any).otherPlayersVisibility[this.player.id];
```
bafolts: Can't the `otherPlayersVisibility` map be stored on the `data` property of this component? This would avoid the need to have to continually reference the root node.

alrusdi: I think no. This variable is reseting to default values when parent component (PlayerHome) gets new data from server. So we have an behavior before fix.

PR #319, opened 2020-02-26
Update robotic WF

vincentneko:
- Fix Heat Trapper behavior with robotic Workforce
- Add prelude cards, mining cards, Gyropolis to robotic Workforce
- Rework the 3 CONSORTIUM cards to work with interrupts
- Add prerequisite checks for energy consuming cards
- Fix regression with Mining guild (oceans with steel bonus did not increase steel production)
- Remove all references from distinct project cards


src/cards/RoboticWorkforce.ts
```diff
@@ -68,13 +80,23 @@ export class RoboticWorkforce implements IProjectCard {
     public canPlay(player: Player, game: Game): boolean {
         return this.getAvailableCards(player, game).length > 0;
     }
+    private miningSteelProduction: number = 0;
+    private miningTitaniumProduction: number = 0;
 
     private getAvailableCards(player: Player, game: Game): Array<IProjectCard> {
         const builderCards: Array<IProjectCard> = [
```
alrusdi: If we are using just names of the cards why should we init whole objects list here?

vincentneko:
```
    private miningSteelProduction: number = 0;
    private miningTitaniumProduction: number = 0;
```

Are used to determine which production to increase (steel or titanium) when you copy mining cards

src/cards/MiningArea.ts
```diff
@@ -29,9 +31,11 @@ export class MiningArea implements IProjectCard {
             game.addTile(player, foundSpace.spaceType, foundSpace, { tileType: TileType.SPECIAL });
             if (foundSpace.bonus.indexOf(SpaceBonus.STEEL) !== -1) {
                 player.setProduction(Resources.STEEL);
+                this.bonusResource = Resources.STEEL;
             }
```
bafolts: `else if` can be used here as it seems a space never has both a steel and a titanium bonus. If a space were to have both placement bonus this `bonusResource` wouldn't work anyway.

src/cards/corporation/MiningGuild.ts
```diff
@@ -5,13 +5,16 @@ import { CorporationCard } from "./CorporationCard";
...
+import { SpaceType } from "../../SpaceType";
 
 export class MiningGuild implements CorporationCard {
     public name: string = "Mining Guild";
     public tags: Array<Tags> = [Tags.STEEL, Tags.STEEL];
     public startingMegaCredits: number = 30;
     public onTilePlaced(player: Player, space: ISpace) {
-        if (space.player === player && (space.bonus.indexOf(SpaceBonus.STEEL) !== -1 || space.bonus.indexOf(SpaceBonus.TITANIUM) !== -1)) {
+        if (
+            (space.player === player || (space.spaceType === SpaceType.OCEAN && player.isCorporation(this.name) ))
```
bafolts: For a corporation card `onTilePlaced` you should always check that `player.isCorporation(this.name)`. I think you can change this check to:

```
if (player.isCorporation(this.name) && (space.bonus.indexOf(SpaceBonus.STEEL) !== -1 || space.bonus.indexOf(SpaceBonus.TITANIUM) !== -1))
```

PR #325, opened 2020-02-27
Plants remove rework

vincentneko:
Leverage interrupt mechanism for cards that remove plants.


src/Game.ts
```diff
@@ -206,6 +207,24 @@ export class Game {
       this.addInterrupt(new SelectResourceProductionDecrease(player, this, resource, count, title));
     }
 
+    public addResourceDecreaseInterrupt(player: Player, resource: Resources, count: number = 1, title?: string): void {
+      if (this.players.length === 1) {
+        return;
+      }
+      let candidates: Array<Player> = [];
+      if (resource === Resources.PLANTS) {
+        candidates = this.getPlayers().filter((p) => p.id != player.id && !p.hasProtectedHabitats && p.getResource(resource) > 0);
```
bafolts: At least in `master` the `hasProtectedHabitats` property has a value which is a function, this should call `!p.hasProtectedHabitats()`. I haven't finished reading through this pull request and I am not sure if this method has changed. Disregard if it's type has changed.

src/Player.ts
```diff
@@ -92,6 +92,17 @@ export class Player {
...
     }
 
+    public getResource(resource: Resources): number {
+      if (resource === Resources.MEGACREDITS) return this.megaCredits;
+      if (resource === Resources.STEEL) return this.steel;
+      if (resource === Resources.TITANIUM) return this.titanium;
+      if (resource === Resources.PLANTS) return this.plants;
+      if (resource === Resources.ENERGY) return this.energy;
+      if (resource === Resources.HEAT) return this.heat;
+      return 0;
```
bafolts: Instead of returning `0` we should throw an error here. This branch represents a `Resources` we are not considering. If some new resource is added at a later time it could fail fast to remind us to add the enum here.

src/cards/Asteroid.ts
```diff
@@ -3,30 +3,18 @@ import {Tags} from './Tags';
...
 import {Game} from '../Game';
-import {SelectPlayer} from '../inputs/SelectPlayer';
+import { Resources } from '../Resources';
+import { CardName } from '../CardName';
 
 export class Asteroid implements IProjectCard {
     public cost: number = 14;
     public tags: Array<Tags> = [Tags.SPACE];
-    public name: string = 'Asteroid';
+    public name: string = CardName.ASTEROID;
```
bafolts: Eventually once every `name` has been updated we can use `CardName` as the type of the `IProjectCard::name` property.

bafolts:
So far changed look good. Using the interrupts is leading to a lot less duplicated code which is a good thing.

src/Game.ts
vincentneko: Thanks, that was it !

PR #326, opened 2020-02-28
Add dialog-polyfill for browsers without dialog

bafolts:
This will add the dialog-polyfill for browsers that don't support dialog.


src/styles/common.less
```diff
@@ -1,4 +1,4 @@
-@import "./nes.min.less";
+@import (inline) "../../node_modules/nes.css/css/nes.min.css";
```
alrusdi: If we are using the file from node_modules may be it is good idea to remove nes.min.less from our styles directory?

src/styles/common.less
```diff
@@ -37,4 +36,4 @@ body {
 @import "./payments.less";
 @import "./venus.less";
 @import "./colonies.less";
-
+@import (inline) "../../node_modules/dialog-polyfill/dialog-polyfill.css";
```
alrusdi: It is better to import it right after nes.min.css. Does it work there?

src/components/PlayerHome.ts
```diff
@@ -15,6 +15,8 @@ import { Preferences } from "./Preferences"
 import { PlayerModel } from "../models/PlayerModel";
 import { Colony } from './Colony';
 
+const dialogPolyfill = require("dialog-polyfill");
```
alrusdi: Shouldn't it be initialized in top most component to be available project-wide? I mean that one from script.ts

bafolts: Webpack should only include the file once allowing us to import it this way. This avoids needing to pass the reference through `global` or finding it on the root node.

src/styles/common.less
bafolts: Yes, good call, I removed it.

PR #331, opened 2020-02-29
Style rework

alrusdi:
This is big one...


vincentneko:
> This is big one...

A really big one !!!

src/components/CreateGameForm.ts
```diff
@@ -78,11 +86,29 @@ export const CreateGameForm = Vue.component("create-game-form", {
...
         }, 
+        randomInteger: function(min: number, max: number) {
+            let rand = min - 0.5 + Math.random() * (max - min + 1);
+            return Math.round(rand);
+        },
         createGame: function () {
-            const players = this.$data.players.slice().map((player: any, index: number) => {
-                player.first = (this.$data.firstIndex === index);
+            if (this.randomFirstPlayer) {
+                this.firstIndex = this.randomInteger(0, this.playersCount-1);
```
bafolts: The `randomInteger` function seems unnecessarily complex. Using `Math.floor(Math.random() * this.playersCount)` should work just the same.

src/components/CreateGameForm.ts
```diff
@@ -78,11 +86,29 @@ export const CreateGameForm = Vue.component("create-game-form", {
...
+            }
+
+            // Set player name for solo mode
+            if (this.playersCount == 1 && this.players[0].name == "") {
```
bafolts: We don't have linting in place but I would always use `===` as it leads to a lot less headaches and hard to find bugs.

src/components/CreateGameForm.ts
```diff
@@ -112,120 +143,226 @@ export const CreateGameForm = Vue.component("create-game-form", {
...
-                <label>
-                    <input type="checkbox" class="nes-checkbox" v-model="corporations" :value="corporation"/>
-                    <span>{{corporation.name}}</span>
-                </label>    
+
+            <div class="create-game-players-cont" v-if="playersCount > 1">
+                <h2>Players</h2>
+                <div class="container">
+                    <div class="columns">
+                        <div class="form-group col6 create-game-player create-game--block" v-for="playerIndex in getPlayersRange()">
```
bafolts: I am 80% sure you can use `playerIndex in playersCount` and remove the `getPlayerssRange` function.

https://stackoverflow.com/questions/44617484/vue-js-loop-via-v-for-x-times-in-a-range

src/components/OrOptions.ts
```diff
@@ -56,13 +57,14 @@ export const OrOptions = Vue.component("or-options", {
                 this.onsave([copy]);
             }, false, false));
             subchildren.push(createElement("div", { style: { display: displayStyle, marginLeft: "30px" } }, [this.$data.childComponents[this.$data.childComponents.length - 1]]));
+            //subchildren.push(createElement("div", { style: { marginLeft: "30px" } }, [this.$data.childComponents[this.$data.childComponents.length - 1]]));
```
bafolts: I would remove the commented out lines in this file.

src/components/PlayerHome.ts
```diff
@@ -156,15 +158,17 @@ export const PlayerHome = Vue.component("player-home", {
                 </div>
             </div>
 
-            <div class="player_home_block player_home_block--actions nofloat"  v-if="!player.corporationCard">
-                <h2>Actions</h2>
+            <div class="player_home_block player_home_block--setup nofloat"  v-if="!player.corporationCard">
+                <h2>Prepare to the game:</h2>
```
bafolts: This phrase `Prepare to the game` doesn't make sense to me. Is there a better phrase we can use? What is listed below ends up being the initial card selection. Something like `Select initial cards` may make more sense.

src/components/CreateGameForm.ts
alrusdi: I agree with this in general. But this.playersCount is not a number (it is an Vue watchable object), so === operator always returns false in this case.

src/components/CreateGameForm.ts
alrusdi: Indeed! Thank you

src/components/CreateGameForm.ts
alrusdi: I reworked it in other way. Is that variant any better?

src/components/CreateGameForm.ts
bafolts: If it is a Vue watchable object and not a number then what is `== 1` verifying? If you check `this.$data.playersCount` that should be a number.

alrusdi: Uhh...This was not a Vue watchable object this was just string instead of number. And 1 == "1" because it is JS, you know. Thank you for your patience :-(

PR #344, opened 2020-03-03
Seeded games

vincentneko:
This will allow to set the random seed to reproduce a game.
Could be used for duplicate tournaments.
Drafted cards will remain the same too, as well as first player

### DIFF
--- a/src/Dealer.ts
+++ b/src/Dealer.ts
@@ -818,10 +818,16 @@ export class Dealer {
     private usePreludeExtension: boolean = false;
     private useVenusNextExtension: boolean = false;   
     private useColoniesNextExtension: boolean = false;
-    constructor(usePreludeExtension: boolean, useVenusNextExtension: boolean, useColoniesNextExtension : boolean) {
+    private seed: number = 0;
+    constructor(usePreludeExtension: boolean, useVenusNextExtension: boolean, useColoniesNextExtension : boolean, seed?: number) {
         this.usePreludeExtension = usePreludeExtension;
         this.useVenusNextExtension = useVenusNextExtension;
         this.useColoniesNextExtension = useColoniesNextExtension;
+        if (seed !== undefined) {
+            this.seed = seed;
+        } else {
+            this.seed = Math.random();
+        }
         this.deck = this.shuffleCards(ALL_PROJECT_CARDS);
         if (this.usePreludeExtension) {
             this.preludeDeck = this.shuffleCards(ALL_PRELUDE_CARDS);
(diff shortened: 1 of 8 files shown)


src/Dealer.ts
```diff
@@ -841,19 +847,25 @@ export class Dealer {
...
-            deck.push(copy.splice(Math.floor(Math.random() * copy.length), 1)[0]);
+            deck.push(copy.splice(Math.floor(this.seed * copy.length), 1)[0]);
         }
         return deck;
     }
     public discard(card: IProjectCard): void {
         this.discarded.push(card);
     }
-    public dealCard(): IProjectCard {
+    public dealCard(isResearchPhase: boolean = false): IProjectCard {
```
bafolts: It shouldn't matter whether we call `pop` or `shift`. As far as I can tell we are always passing `true` to this method as well. I am not sure why this is needed. Can you explain?

src/Dealer.ts
```diff
@@ -841,19 +847,25 @@ export class Dealer {
         const deck: Array<any> = [];
         const copy = cards.slice();
         while (copy.length) {
-            deck.push(copy.splice(Math.floor(Math.random() * copy.length), 1)[0]);
```
bafolts: This is continually pulling a random card from 0 to copy.length. I am not sure this works if a number is provided. You could end up pushing `undefined` into the deck.

vincentneko: I've tested it in a 3 player game, it seems to work.
Not sure if it will shuffle the cards in a significant way though

vincentneko:
> I am not sure what is trying to be accomplished with this pull request. Can you explain more?

The concept is to remove randomness. You can have more details here: 
https://en.wikipedia.org/wiki/Duplicate_bridge

For instance, let's say you play on Tharsis map and you got Tharsis Republic corporation along with Immigrant City and Protected Habitats, you are somehow a strong favorite to win the game.

Now if we take x groups of 4 players and these play the same game, we can compare the position of players 1, 2, 3 and 4 relatively. 
Let's say we have 4 groups of 4 players.
If player number 3 is finishing in 2nd position on table 4 but other players are finishing in 3rd or 4th position on table 1,2,3 this is a great performance.
I'm not sure yet how to grant points according to the game results.
After 3 or 4 games you can make a global ranking among players.

Now about pop vs shift for dealer. 
This is to simulate 2 drawing stacks in order to preserve card order for research / drafting phase.

For instance, if in game 1 someone is placing an ocean on the 2 bonus cards and not in game 2, they will not have the same cards for the research / drafting phase.

I hope this clarify.

src/Dealer.ts
vincentneko: This is to simulate 2 drawing stacks in order to preserve card order for research / drafting phase.

bafolts: Is this work completed? This function is always called with `true` which makes the argument unnecessary.

vincentneko: Yes it is.
The function is always called with false except in `player runDraftPhase` and `player runResearchPhase`

```
    public runResearchPhase(game: Game, draftVariant: boolean): void {
      let dealtCards: Array<IProjectCard> = [];
      if (!draftVariant) {
        dealtCards.push(
          game.dealer.dealCard(true),
          game.dealer.dealCard(true),
          game.dealer.dealCard(true),
          game.dealer.dealCard(true)
        );
      } else {
        dealtCards = this.draftedCards;
        this.draftedCards = [];
      }
```

PR #348, opened 2020-03-04
Fix regression bug

vincentneko:


src/Game.ts
```diff
@@ -968,9 +968,9 @@ export class Game {
         );
       }
 
-      if (tile.tileType !== TileType.OCEAN) {
+      //if (tile.tileType !== TileType.OCEAN) {
```
bafolts: I would remove these commented out lines.

Issue #355, opened 2020-03-06
Solo game broken if you win before last turn with Venus Next


lostgrail:
When all global parameters have been completed before the final generation, the game asks you to choose an action for WG Terraforming, but there are no options. As a result, the game breaks and cannot proceed. 

![0517EFC3-E24C-454E-8394-70637BB7D4D4](https://user-images.githubusercontent.com/7331135/76045116-3aeb8480-5f2a-11ea-9d05-cc1afe6d669d.png)

This game was Venus, Preludes, and colonies, so game would end on generation 12.

vincentneko:
As always, thanks for feedback.
Fixed and merged

Issue #351, opened 2020-03-05
Cannot use Celestic action on first turn


lostgrail:
Celestic does not show up as an action that can be activated on the first turn

![409AEEB0-6569-4D87-91E7-7E65E0140E99](https://user-images.githubusercontent.com/7331135/75941128-dc0c0980-5e5c-11ea-9926-7e7fefa01de0.jpeg)
![C8F9D7B1-542F-4006-B4A5-5FA8E386BC71](https://user-images.githubusercontent.com/7331135/75941133-e0382700-5e5c-11ea-8ede-2706d537dbc3.jpeg)
![445F6F44-227D-48FF-8E35-648D2FC04644](https://user-images.githubusercontent.com/7331135/75941137-e1695400-5e5c-11ea-9462-b6fec56f464f.jpeg)

vincentneko:
Fixed and merged

Issue #349, opened 2020-03-05
Only four colonies in a two player game


lostgrail:
Start a new two player game, and there are only four colonies. There should be five.

vincentneko:
Fixed and merged

Issue #288, opened 2020-02-21
Suggestion: Expandable logging


dabewi:
Hi,

in some situations, especially wih 3 or 4 players, when one players passes early it's quite difficult to trace back what the other players have done when the next generation starts, because the log only shows 10 lines.

I think it would be cool to have a button (e.g. with the label "Show more...") that increments the lines shown in the log by a fixed amount (e.g. for every click I get 5 more/older lines of the log).

Another idea would be to make the amount of lines of log dependent of the player count. So, in a 2 player game the log shows 10 lines, in da 3 player game 15, etc. 

But I think the button would be better.

What do you think?

dabewi:
Although I like the new style I have some ideas regarding the logging:

- Wouldn't it be possible to make a small window with a scroll bar and just log the entire game?
- If not, I think it would be nice to reduce the font size of the logging itself to make it more space-efficient
- I also vote for a longer log (at least more than 10 lines)

Greetings

vincentneko:
I agree with you. Now that we have the new css we should be able to look into this, will try to get something.

vincentneko:
Game log is now 50 entries long, and a scrollbar can be used.
@alrusdi thanks a lot for your help on this one

PR #363, opened 2020-03-07
Misc changes

vincentneko:
Fix Solo game broken if you win before last turn with Venus Next #355
Fix Cannot use Celestic action on first turn #351
Fix Only four colonies in a two player game #349
Suggestion: Expandable logging #288, thanks @alrusdi for this one
Enable Ants and Predators for solo game play

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -596,6 +596,15 @@ export class Player {
     }
 
     public worldGovernmentTerraforming(game: Game) {
+
+      // Test if this is needed, usefull for solo play
+      if (game.getTemperature() >= constants.MAX_TEMPERATURE 
+        && game.getOxygenLevel() >= constants.MAX_OXYGEN_LEVEL
+        && game.board.getOceansOnBoard() >= constants.MAX_OCEAN_TILES
+        && game.getVenusScaleLevel() >= constants.MAX_VENUS_SCALE) {
+          return;
+      }
+
       const action: OrOptions = new OrOptions();
       action.title = "Select action for World Government Terraforming";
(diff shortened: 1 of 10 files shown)


src/Player.ts
```diff
@@ -596,6 +596,15 @@ export class Player {
     }
 
     public worldGovernmentTerraforming(game: Game) {
+
+      // Test if this is needed, usefull for solo play
+      if (game.getTemperature() >= constants.MAX_TEMPERATURE 
+        && game.getOxygenLevel() >= constants.MAX_OXYGEN_LEVEL
+        && game.board.getOceansOnBoard() >= constants.MAX_OCEAN_TILES
+        && game.getVenusScaleLevel() >= constants.MAX_VENUS_SCALE) {
```
bafolts: If you are not playing the venus game this will always be false.

src/Player.ts
```diff
@@ -1446,7 +1455,7 @@ export class Player {
         game.getGeneration() === 1 &&
             this.corporationCard !== undefined &&
             this.corporationCard.initialAction !== undefined &&
-            !this.actionsThisGeneration.has(this.corporationCard.name) &&
+            !this.actionsThisGeneration.has("CORPORATION_INITIAL_ACTION") &&
```
bafolts: What was the issue with using `this.corporatinCard.name`?

vincentneko: It was consumming the corporation action.

src/Player.ts
vincentneko: World gov mechanism is tight with venus

PR #367, opened 2020-03-08
Create ColonyModel for serializing colonies for client side

bafolts:
I wasn't able to recreate the issue reported but we had a cyclical link between players and colonies we were attempting to serialize. I have created a `ColonyModel` which only includes the data we need on the client. This will be less data we send to the client.


src/DataAccessObject.ts
```diff
@@ -1,31 +0,0 @@
-
```
bafolts: Before I was going to give this project up a while back I was going to try to add code to save the game state to memory so that a game could be stored. This never went anywhere so this class can be removed. This class has a reference to `JSON.stringify` when I was looking for all places where we could have hit the cyclical reference error. Removing this class since it isn't used.

PR #373, opened 2020-03-11
Layout changes (more compact)

vincentneko:
- Display corporation card as first played card, no more dedicated area for it.
- No display for event cards played for other players (only count)
- Stack Prelude and Automated cards
- Stack event cards
- Cleanup in card component

Thanks to @pierrehilbert for the help !


src/components/OtherPlayer.ts
```diff
@@ -24,25 +38,29 @@ export const OtherPlayer = Vue.component("other-player", {
...
 
-                <div v-if="player.playedCards.length > 0" class="player_home_block">
+                <div v-if="player.playedCards.length > 0 || player.corporationCard !== undefined" class="player_home_block">
                     <h4>Played Cards</h4>
                     <div>
-                        <div v-for="card in player.playedCards" :key="card.name" class="cardbox">
+                        <div v-if="player.corporationCard !== undefined" class="cardbox">
+                            <card :card="player.corporationCard" :resources="player.corporationCardResources"></card>
+                        </div>
+                        <div v-for="card in getCardsByType(player.playedCards, ['blue'])" :key="card.name" class="cardbox">
```
bafolts: You should pass in `CardType` here instead of the string for the colors. If the `CardType` were to change these need to get updated as well.

PR #379, opened 2020-03-12
seeded games, take 2

vincentneko:
Second attempt to make games reproductibles, and keeping a high card shuffle.
This time with random-seed package !


vincentneko:
> What problem is this attempting to solve. It is replacing `Math.random()` with some other software based random number generator. I can't imagine we had a bug related to `Math.random()` not being random enough.

The issue was not with `Math.random()` but with how I try to implement seeded RNG the first time (which has been rolled back because cards where not shuffled enough). Now instead of `Math.random()` we use a seeded RNG which allow to reproduce the same game if you choose the same seed and game options.
This can be used for some kind of duplicate tournament (I gave details in a previous PR)

bafolts:
I really don't think this is doing what you think it is doing. I had concerns the last one may not as well. Is the goal that users from two different games will end up with the same starting deck with cards in the same order?

vincentneko:
> I really don't think this is doing what you think it is doing. I had concerns the last one may not as well. Is the goal that users from two different games will end up with the same starting deck with cards in the same order?

I agree that the last one was not good but I think this one is doing what is expected.
Yes the goal is, if you provide the same seed and take the same options, to have two different games  with the same starting deck with cards in the same order, and same cards offered for drafting.

The only difference I see is that we use a seeded RNG instead of normal RNG.

This could allow duplicate tournament play mode and to compare each player relative performance.

peteroupc:
Introducing seeded games (that is, associating a "seed" to a particular game considering) by involving a PRNG has the concern that the application [will be tied to that PRNG or its implementation](https://github.com/peteroupc/peteroupc.github.io/blob/master/random.md#manually-seeded-prngs).

In this case, this game configuration includes a particular deck permutation (arrangement of cards in the deck); in shuffling cases such as this, the [choice of PRNG](https://github.com/peteroupc/peteroupc.github.io/blob/master/random.md#shuffling) is particularly important.

Did you consider alternatives to seeded games, such as storing the full game configuration instead of a seed (possibly compressing that configuration and/or converting it to readable text and/or memorable words)?  These alternatives don't require using an RNG and may achieve the same goal of reproducible game configurations.

bafolts:
I would be in favor of @peteroupc suggestion. There are less than 1000 cards. If each card were assigned a number you could store the order of the deck in 1000 bytes or less even if you did something as simple as store some id for each card in an array. There is often a limit to the size of GET requests so we would need to keep this value low. There are not 1000 cards so we can most likely come up with some format to store the game configuration in under 1000 bytes.

bafolts:
There are under 1000 cards and will most likely always be. Using 10 bits to store the card id at each index we can store the order of the cards in 1250 bytes at most. This would be a worst case scenario if we ever had 1000 cards. We would need a way to encode this string so that we could utilize it on the client side in a query string. That would increase the storage size but I don't _think_ it would double it so we should still come under 2000 bytes. The game server doesn't care how long the query string size is so we should be fine. I can work on implementing this approach which would end up with two games being able to have the same cards in the same order in the starting deck for both corporation cards and project cards. The seed would have to be provided on the game creation screen and would be a string of characters probably between 1000-1500 characters long.

vincentneko:
> There are under 1000 cards and will most likely always be. Using 10 bits to store the card id at each index we can store the order of the cards in 1250 bytes at most. This would be a worst case scenario if we ever had 1000 cards. We would need a way to encode this string so that we could utilize it on the client side in a query string. That would increase the storage size but I don't _think_ it would double it so we should still come under 2000 bytes. The game server doesn't care how long the query string size is so we should be fine. I can work on implementing this approach which would end up with two games being able to have the same cards in the same order in the starting deck for both corporation cards and project cards. The seed would have to be provided on the game creation screen and would be a string of characters probably between 1000-1500 characters long.

This looks great !
Please take also into account that random first player must be the same for each game, same for colonies.

bafolts:
OK @vincentneko at some point a user will need to see a seed string to use for another game. Somewhere we will need to add a link to get the seed string for a game. The game function will have a function to generate the seed but I think the seed string will need the following:

```js
    public populateFromSeed(_seed: string): void {                                                                     
      // version - start as 0, first byte                                                                              
      // boardName - number in second byte                                                                             
      // first player index - number in third byte                                                                     
      // cards - number of cards in fourth byte                                                                        
      //       - each card is then 10 bits where card number is stored                                                 
      // corporationCards
      //       - number of corporation cards in next byte                                                              
      //       - each card is then 10 bits where card number is stored                                                 
      // preludeExtension
      //       - 1 bit set to 1 if using prelude extension                                                             
      // venusNextExtension
      //       - 1 bit set to 1 if using venus next extension                                                          
      // coloniesExtension                                                                                             
      //       - 1 bit set to 1 if using colonies extension                                                            
      // colonies 
      //       - number of colonies in next byte                                                                       
      //       - each colony is then 10 bits where colony number is stored                                             
    }   
```

If it is a 1 player game I think we should also store the neutral cities and forests. Let me know your thoughts if this would work and I can being to implement this. I will have to add a number id to each corporation card, project card, and colony to track them as a number and not by their name to save space.

vincentneko:
> OK @vincentneko at some point a user will need to see a seed string to use for another game. Somewhere we will need to add a link to get the seed string for a game. The game function will have a function to generate the seed but I think the seed string will need the following:
> 
> ```js
>     public populateFromSeed(_seed: string): void {                                                                     
>       // version - start as 0, first byte                                                                              
>       // boardName - number in second byte                                                                             
>       // first player index - number in third byte                                                                     
>       // cards - number of cards in fourth byte                                                                        
>       //       - each card is then 10 bits where card number is stored                                                 
>       // corporationCards
>       //       - number of corporation cards in next byte                                                              
>       //       - each card is then 10 bits where card number is stored                                                 
>       // preludeExtension
>       //       - 1 bit set to 1 if using prelude extension                                                             
>       // venusNextExtension
>       //       - 1 bit set to 1 if using venus next extension                                                          
>       // coloniesExtension                                                                                             
>       //       - 1 bit set to 1 if using colonies extension                                                            
>       // colonies 
>       //       - number of colonies in next byte                                                                       
>       //       - each colony is then 10 bits where colony number is stored                                             
>     }   
> ```
> 
> If it is a 1 player game I think we should also store the neutral cities and forests. Let me know your thoughts if this would work and I can being to implement this. I will have to add a number id to each corporation card, project card, and colony to track them as a number and not by their name to save space.

This looks solid !
I think we need to store how many players are linked to the game.
If not, cards distributed / draw will differ.

bafolts:
I still plan to do this at some point. If someone else implements the strategy should be to allow a user to export their game configuration while in a game so that another game can be created with the same inputs. On the game creation screen a string can then be provided to start a new game with the same game settings. It doesn't have to be complex at first and for the time being we could simply use `JSON.stringify` on some JSON representation of a `GameSettingsModel` class and have users copy and paste that string. Early optimization is bad so then if the feature is even used we could then optimize it to use smaller strings when exporting and importing.

bafolts:
This is taking up space. With the recent change added to serialize Game, Dealer, and Player we can probably come up with a simpler solution. Leveraging that database we would only need to use a key from that database to allow another game to use the same setup as a previously finished game.

Issue #284, opened 2020-02-20
Show all running games per server


dabewi:
Hi,

as a server admin I'd like to display all the running games with the corresponding access links. It occurs that access links get lost between mediums (e.g. between laptop and mobile phone) or that someone just forgets to immediately start the game and after a while doesn't have access to the link.

Therefore it would be helpful to list all the running games and the according links.

Cheers!

gomfunkel:
@bafolts I tried myself on this and submitted a pull request. Feel free to check it and either merge or get back to me with questions/comments.

gomfunkel:
Implemented and merged with pull request #393. Feel free to close this issue, @dabewi.

PR #393, opened 2020-03-14
Added page with overview of all games on the server (Issue 284)

gomfunkel:
Tried myself on a page displaying all games available on the server as requested in #284.

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -39,6 +39,7 @@ import { Phase } from './src/Phase';
 import { Resources } from "./src/Resources";
 import { CardType } from "./src/cards/CardType";
 
+const serverId = generateRandomServerId();
 const styles = fs.readFileSync('styles.css');
 const games: Map<string, Game> = new Map<string, Game>();
 const playersToGame: Map<string, Game> = new Map<string, Game>();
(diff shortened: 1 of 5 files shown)


gomfunkel:
Sounds like a good idea, although it feels a bit more like obscurity than actual security. Better than nothing I guess, though.

I will introduce the serverId for the following routes and update the pull request accordingly:
- /
- /games-overview
- /api/games

gomfunkel:
Introduced `serverId` to secure the game administration pages mentioned in my last post.

I'm not totally happy with this yet, as only a rather ugly `403` along with the error message _Not authorized_ is returned. It would be way nicer to delegate the authorization to the app and provide more beautiful error pages. 

What do you think? Is this necessary and if so should it be a new issue or solved here?

bafolts:
I don't think a custom 403 page is necessary. Anyone attempting to get to those URLs would be doing so manually since the application doesn't provide those links to players.

server.ts
```diff
@@ -56,7 +57,16 @@ function requestHandler(
   if (req.url !== undefined) {
     if (req.method === 'GET') {
       if (
-        req.url === '/' ||
+        req.url.replace(/\?.*$/, '') === '/' ||
```
bafolts: This all looks fine but I may have gotten ahead of myself. I would only enforce it for the `/games-overview` route. This will then still allow anyone to _create_ a game but only a server administrator can see the games running on a server.

gomfunkel:
Removed the `serverId` protection for the create game screen. Now, only the following routes are protected by it:
- `/games-overview`
- `/api/games`

Issue #255, opened 2020-02-12
Suggestion: Expanded end screen


dabewi:
Hi,

in think it would be an improvement for the game to have the possibility to expand the end screen to show e.g. the last moves (just show the log), the funded awards and the award rankings of the players. I'd also like to thee the board in the final state.

Right now if one opponent ends the game it's possible to miss the final moves he made because there is no way from the end screen back to the game.

In fact it would be sufficient to add a possibility to return to the game screen.

Cheers!

dabewi:
This is done, isn't it?

alrusdi:
No, i don't have a good solution for returning to the final game.

vincentneko:
Maybe returning to the final game is not required.
Adding the game log could be a nice addition, I'm also in favor of storing award rankings for each players.

dabewi:
I think returning to the game screen is not required. I initially thought that this would be the easiest solution, because any needed information is stated there. But it's not necessary.

The game log would be nice though. And a little bit more information about the award rankings would be neat, too.

gomfunkel:
Created pull request #395 to add the final game log to the end game screen. Award details are still not there.

gomfunkel:
Can someone assign me to this, please? I am going to implement the award part on the end screen as soon as I‘m done with the #286.

vincentneko:
> Can someone assign me to this, please? I am going to implement the award part on the end screen as soon as I‘m done with the #286.

Done

gomfunkel:
Implemented and merged to master with #439. Feel free to close this issue, @dabewi.

PR #395, opened 2020-03-15
Added game log to end game screen

gomfunkel:
Added the game log to the end game screen as requested in issue #255.

### DIFF
--- a/src/styles/game_end.less
+++ b/src/styles/game_end.less
@@ -24,3 +24,22 @@
         border-bottom: .05rem solid #7b7b7b !important;
     }
 }
+
+.game_end_block--log {
+    .log-panel {
+        max-height: 180px;
+        border-style:none;
+        font-size: 18px;
+        padding: 5px;
+        max-width: 767px;
+        background: #c8c8c8;
+
+        li {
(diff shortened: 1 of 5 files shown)


src/styles/game_end.less
```diff
@@ -24,3 +24,22 @@
         border-bottom: .05rem solid #7b7b7b !important;
     }
 }
+
+.game_end_block--log {
```
alrusdi: This style definition duplicates code from player_home.less
Please, reuse it, do not copy/paste

gomfunkel: I absolutely see your point but as all classnames are prefixed with the screen they are on I figured this is the preferred way you do it. If it's not: How do you handle reuse of styles in general, I can't image adding a `.player_home_block--log` class to the game end screen would be the preferred way, is it?

alrusdi: The best variant - is to move log styling to separate log.less file, and make its css-classes naming independent from page name.
Simplest variant is just modify player_home.less and add .game_end_block--log to .player_home_block--log class

gomfunkel: @alrusdi Fixed.

Issue #402, opened 2020-03-16
Immigrant City does not modify MC production correctly


gomfunkel:
When playing Immigrant City the MC production bonus is given twice so you end up with 0 change in MC production.

Current behevior:
- -2 MC production for playing the card (ImmigrantCity#play())
- +1 MC production after playing the card (ImmigrantCity#play())
- +1 MC production for placing the city tile (ImmigrantCity#onTilePlaced())

Should be:
- -2 MC production for playing the card (ImmigrantCity#play())
- +1 MC production for placing the city tile (ImmigrantCity#onTilePlaced())

I will fix this later tonight and submit a pull request.

gomfunkel:
Unable to reproduce the behavior, will investigate further.

gomfunkel:
Fixed and merged with pull request #403.

PR #403, opened 2020-03-16
Fixed erroneous MC production change (#402)

gomfunkel:
Fixed erroneous MC production change for Immigrant City. The card effect (+1 MC production) was apparently awarded in the `play()` method and then triggered again in the `onTilePlaced()` method resulting in a MC production change of _0_ instead of _-1_.

### DIFF
--- a/src/cards/ImmigrantCity.ts
+++ b/src/cards/ImmigrantCity.ts
@@ -29,8 +29,6 @@ export class ImmigrantCity implements IProjectCard {
             game.addCityTile(player, space.id);
             player.setProduction(Resources.ENERGY,-1);
             player.setProduction(Resources.MEGACREDITS, -2);
-            // Resolve onTilePlaced
-            player.setProduction(Resources.MEGACREDITS);
             return undefined;
         });
     }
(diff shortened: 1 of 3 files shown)


src/cards/ImmigrantCity.ts
```diff
@@ -29,8 +29,6 @@ export class ImmigrantCity implements IProjectCard {
             game.addCityTile(player, space.id);
             player.setProduction(Resources.ENERGY,-1);
             player.setProduction(Resources.MEGACREDITS, -2);
-            // Resolve onTilePlaced
-            player.setProduction(Resources.MEGACREDITS);
```
bafolts: This fix is accurate. One way or another we have a unit test which is looking for the production to be `0` in this case but I agree it should be `-1`. The build is not running on forks but this change will need to update the associated unit test. I will look into getting this to run from forks but can you run `npm test` for now and update the associated unit test?

https://github.com/bafolts/terraforming-mars/blob/master/tests/cards/ImmigrantCity.spec.ts#L28

gomfunkel:
Unit test was updated accordingly and `npm test` is now green after also cherry-picking the fixed test for Vitor by @vincentneko.

Issue #399, opened 2020-03-16
UI Problems after #397


gomfunkel:
A couple of UI issues I have noticed after #397 was merged: 

- Choosing black as the player color makes it very hard to read the headline/playername.
- The amount of resources you receive from a card is very hard to read (see 3 plants in the screenshot below.

<img width="293" alt="ecoline" src="https://user-images.githubusercontent.com/547035/76759321-bc91ac80-678b-11ea-9e7c-00c19407f93c.png">

- Text on the player overlay on the player home screen is very hard to read. Background probably needs to be darker.

<img width="863" alt="player_overlay" src="https://user-images.githubusercontent.com/547035/76760615-3fb40200-678e-11ea-8cc3-c5b6a565b161.png">

I will add to this as soon as I find more.

FreeLikeGNU:
Please see https://github.com/bafolts/terraforming-mars/pull/404

alrusdi:
this should be fixed now

PR #404, opened 2020-03-16
Issue 399 visuals fixed

FreeLikeGNU:
Resolved visual problems found in issue https://github.com/bafolts/terraforming-mars/issues/399
Screenshot_2020-03-16_11-51-16
Screenshot_2020-03-16_11-49-16


src/HTML_data.ts
```diff
@@ -5136,7 +5136,7 @@ export const HTML_DATA: Map<string, string> =
 ["EcoLine",`
     <div class="tag tag1 tag-plant"></div>
     <div class="corporationLabel">CORPORATION</div>
-    <div class="contentCorporation">
+    <div class="contentCorporation" style="color:#020202;">
```
bafolts: This doesn't seem to have been added to every `contentCorporation`. I would create a separate class for this style color and use it where it is being applied. Which corporations were picked to use this color if not all of them?

bafolts: This all looks fine to me once a CSS class is utilized for the `style="color:#020202;"`. I am not sure which corporations are getting this color but then the overall div class can be similar to `class="contentCorporation reasonForColor"`.

FreeLikeGNU: I've made the requested changes and fixed a few more things as well

alrusdi:
I've already fixed resources count visibility problem. Please remove changes from HTML_data - it will be a huge pain in the future. We should rework that module...

src/HTML_data.ts
```diff
@@ -5518,7 +5518,7 @@ export const HTML_DATA: Map<string, string> =
           on this card.)
           </div>
         </div>
-        <div  class="celestic" style="font-size:24px;box-shadow:6px 6px 6px grey;margin-left: 9px;
+        <div  class="celestic" font-size:24px;box-shadow:6px 6px 6px grey;margin-left: 9px;
```
bafolts: Does this work to not include the `style` attribute. This HTML was copied from the ssimoneoff cards repository and should be using Vue.js templates. Eventually each of these should move to their own template file. @alrusdi has done most of the CSS work up until this point so he is probably best to approve or reject these changes. I would like to see `HTML_data` eventually go away though and move towards Vue.js templates for each of these cards.

FreeLikeGNU: I've reverted to upstream HTML_data.ts however the "points" need to be moved in a few corporation cards such as Arklight and Pristar. Maybe resolved in another PR?
Screenshot_2020-03-17_11-47-26

alrusdi: Please do it by css, add class "card-arklight" and "card-pristar" in HTML_data.ts, then fix fontt size or positioning in css rules.

FreeLikeGNU: I've removed my changes to HTML_data.ts while a better way to fix the point boxes in corporations is under discussion.

Issue #400, opened 2020-03-16
Visualize used actions from blue cards


gomfunkel:
Right now it is not possible to see which blue card was already _used_ or _activated_ in the current generation without scrolling through the log. When having played an action from a blue card, this should be made visible on the card itself.

This would be helpful when looking through the cards of other players to see what options they still have in this generation for example.

gomfunkel:
@alrusdi What do you think about re-using the new drop shadow for selected cards for this use case as well, just using a different color? Something like this:

<img width="297" alt="highlight" src="https://user-images.githubusercontent.com/547035/76788945-2b86f980-67bc-11ea-8d93-3f94e5a66af5.png">

alrusdi:
This is good idea but it is meaning is not obvious. May be we should use some kind of bage with "activated" caption on it?

alrusdi:
In original ssimeonoff cards list https://ssimeonoff.github.io/cards-list he uses grey-out effect for selection. It can better represent inactive state to me

gomfunkel:
@alrusdi The grayed out variant seems interesting. I will have a look into this. Could you please assign me to this issue?

gomfunkel:
I have a branch with a working version of this feature which just needs a little cleanup. In it, blue cards are now grayed out when their action was used. This works for both your own cards and when looking at the cards of other players.

Having two blue cards played:
<img width="868" alt="Before" src="https://user-images.githubusercontent.com/547035/76878395-073b2380-6875-11ea-9725-511f21bb050c.png">

Having used the Tardigrades action:
<img width="860" alt="After" src="https://user-images.githubusercontent.com/547035/76878690-7b75c700-6875-11ea-8acc-3daca2599bdc.png">

alrusdi:
I like this variant. Please make PR, but actualize your version from our master branch to avoid merging conflicts

gomfunkel:
Implemented and merged to master with #409.

PR #409, opened 2020-03-17
Visualize used actions from blue cards (#400)

gomfunkel:
When using the action of a blue card the card is now grayed out so all players can easily see which action was already used and which is still available.

To realize this feature, the `Set<string> actionsThisGeneration` had to be turned into an `Array` as `Set`s are not reactive so the views could not react to changes when an action was used.

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -336,7 +336,7 @@ function getPlayer(player: Player, game: Game): string {
     boardName: game.boardName,
     colonies: getColonies(game.colonies),
     tags: player.getAllTags(),
-    actionsThisGeneration: player.getActionsThisGeneration()
+    actionsThisGeneration: Array.from(player.getActionsThisGeneration())
   } as PlayerModel;
   return JSON.stringify(output);
 }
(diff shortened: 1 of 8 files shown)


alrusdi:
Visual part is great! I've asked @bafolts to check serverside part

server.ts
```diff
@@ -335,7 +335,8 @@ function getPlayer(player: Player, game: Game): string {
     venusScaleLevel: game.getVenusScaleLevel(),
     boardName: game.boardName,
     colonies: getColonies(game.colonies),
-    tags: player.getAllTags()
+    tags: player.getAllTags(),
+    actionsThisGeneration: player.getActionsThisGeneration()
```
bafolts: As long as you map the `Set` to an `Array` here you shouldn't have to modify `src/Player.ts` or `src/cards/venusNext/Viron.ts` to use a `Set`. Only `src/components/**/*.ts` runs on the client and will have to use the `Array`. You can't send a `Set` through JSON so using an `Array` here makes sense. The `src/Player.ts` code runs on the server so it can stay a `Set`.

gomfunkel: Good point. Fixed this with the latest commit sticking to `Set` on the server and `Array` for client/server interaction.

Issue #376, opened 2020-03-12
Suggestion: Replace generic tiles with more specific ones


dabewi:
Hi,

if one of my opponents plays e.g. a commercial district or a capital I want to know where he played it, even after some generations. Sometimes the mars gets rather packed and with more city tiles it gets more tedious to track down which city tile represents a standard city and which one represents the capital (because I don't want to place an ocean tile near an opponents capital or a city near an opponents commercial district).

Therefore it would be nice to actually have specific symbols for the special tiles, or at least for some of them.

Cheers!

alrusdi:
This is done

rodrigomp84:
> This is done

Actually it seems the capital tile is still just a regular city tile. I agree with the OP that being able do tell them apart is quite important!

PR #346, opened 2020-03-04
Add some cards

vincentneko:
Please do not merge from now, I will add some commits here from time to time.
Feel free to help on this one. If you do so please let me know the cards you will take so we don't collide.

PR #414, opened 2020-03-18
More details for specific tiles; #376

alrusdi:


src/cards/CommercialDistrict.ts
```diff
@@ -37,7 +37,7 @@ export class CommercialDistrict implements IProjectCard {
             game.addTile(player, foundSpace.spaceType, foundSpace, {
               tileType: TileType.SPECIAL,
               card: this.name
-            });
+            }, false, "commercial_district");
```
vincentneko: Maybe CardName enum could be used here instead of a string ?

alrusdi: I can do it after #346 merge. Let's wait for it

alrusdi: @vincentneko i did it!

src/components/Tile.ts
```diff
@@ -11,7 +12,7 @@ export const Tile = Vue.component("bonus", {
         let tile_elements = [];
         let app = this;
 
-        let build_tile_css_class = (tile_type: TileType):string => {
+        let build_tile_css_class = (tile_type: TileType, tileDetails: string = ""):string => {
```
bafolts: You can use `CardName` here and every palce `tileDetails` is used. This `tileDetails` (and the others) would then become `tileDetails: CardName`. Since this is the card which placed the tile a better name may be `cardName`.

alrusdi: Isn't it better to add some new entries to TileType enum?
export enum TileType {
    GREENERY,
    OCEAN,
    CITY,
    
   MOHOLE_AREA,
   MINING_RIGHTS_MARKER,
   ...
}
Then i can remove tileDetails

bafolts: That is better, and there are symbols for each of those tile types that could be introduced.

Issue #407, opened 2020-03-17
Better tags, millestones and VPs tracker


alrusdi:
What about to do it in popup widget like our current Preferences?
Popup will contain a switchable tabs with different categories
Each tab will have a table with player names and parameters. Columns can be sortable by click. Default sorting can be configurable.
Millestones and Awards will have subtabs for each MIllestone and Award with description and current values
VPs tab will track oponnets and current VPs actual to the current move
![draft](https://user-images.githubusercontent.com/394311/76856822-eb844d00-6875-11ea-8aa9-ee6ab4994d14.jpg)

FreeLikeGNU:
I'd like to see the VP next to the TR points in the top bar.   This top bar would also have a button (the bottom border) that expands to show the other players VP, TR and resources just under the viewing players stuff for quick comparisons.

ssimeonoff:
Example of a simple row of the tags count fitting the current design:
![tags](https://user-images.githubusercontent.com/6917565/77117512-76aa4200-6a32-11ea-933e-d9336e319489.png)

vincentneko:
Can we close this one ?

alrusdi:
No please. Tags counter is just first (amathing) step to show more statistics in the game

vincentneko:
@alrusdi and @nwai90 
I've created `improve_ms_awards `branch. In this branch scores for milestones and awards are exposed for client.
Do you mind taking care of the client side as I'm pretty bad with this ?

![Screenshot 2020-07-21 13 32 54](https://user-images.githubusercontent.com/56086992/88051032-41387100-cb58-11ea-8997-621980341c57.png)

alrusdi:
I think stats data needs more generalization. Now we need an award counts. Tomorrow we will need Milestones counts. Next day we will need Turmoil Influence values...
May be it is better to have separate section in player response for stats? It can be a list with StatsModel structure (id="awards", name="Awards", scores=[...])
And then implement UI like panel with tab for each StatsModel item in that list

vincentneko:
In my opinion it makes sense to store Milestones / Awards additional information within Milestones / Awards.
This information can then be extracted on the client side to be displayed either with Milestones / Awards or somewhere else.

Issue #286, opened 2020-02-20
Suggestion: VP projection


dabewi:
Hi,

what do you think about a VP projection during game? I find it rather stressfull to maintain an overview over the points of each player during a match, especially with higher player numbers and more complex networks (ressources on cards with 1 VP per token or per 2 tokens or per 4 tokens), milestones and awards and the TR itself.

I think it would be a great improvement since it really shows the benefits of an online game (while you have to calculate the VP yourself if you play IRL).

What do you think about that? I guess it wouldn't be much work, because there already seems to be a method that calculates the VP. And the change is only that this method is called after each action of each player.

Cheers!

gomfunkel:
I have a working prototype for this feature request.

@bafolts (or someone else): Can you please assign me to this issue?

gomfunkel:
It's working (and looking imho) nicely now. Will open a pull request soonish.

<img width="567" alt="vp-projection" src="https://user-images.githubusercontent.com/547035/76944956-2df36b80-6902-11ea-91ec-e665d59fe1d3.png">

This is not yet in the preferences pane style as suggested by @alrusdi in #407. The foundations for it should be laid now, though.

alrusdi:
I like it - good addition for OtherPlayer info

dabewi:
Very fine, @gomfunkel !

gomfunkel:
Implemented and merged to master with #416. Feel free to close this issue, @dabewi.

PR #416, opened 2020-03-18
Added VP projection to player home and other players (#286)

gomfunkel:
Now displaying the current VP of each player on their home screen and of all players on their respective player info screens. The VP are updated after every player action and it should be possible to enhance this with a more detailed breakdown (see #407 for ideas).

See #286 for further details and screenshot.


gomfunkel:
Still need to look into the failing test regarding VP calculation. In the frontend everything looked fine.

gomfunkel:
In `Game.spec.ts:41` 64 victory points are added arbitrarily. This is not working anymore as the current VP are now continuously recalculated. I think it makes sense to change the test by removing the addition of VP, do you agree @bafolts or is there some important reason for the addition I fail to understand?

gomfunkel:
Changed test as described above.

ssimeonoff:
A quote from: https://boardgamegeek.com/thread/2389566/terraforming-mars-vs-star-wars-rebellion
"In Terraforming Mars all the players feel like they are winning until the end..." 
In my last 1vs1 game, I lost a lot, but until the end I had the feeling that I have a shot. 
In my last 4-player games, I thought I will win with a small margin and it turned out to be a 90-60-60-60 total ownage. 

Knowing the VP count to all players in real time, will that sweet "ignorance is bliss" experience. 
In the first example, if I knew that I am losing completely, I would lose interest in the game few whole generations before the end. 
In the second example ,knowing that you have won, will makes you stop trying.

If you are really implementing this feature, then the VP count should show only the player personal VP.

gomfunkel:
@ssimeonoff Interesting ideas on the topic, thanks for sharing!

Showing the personal VP is a no-brainer to me, it just makes life more easy as you do have the information anyways, both on the board and online.

Showing the VP of other players on the other hand is an interesting discussion. In the board game, this information is also available to everyone except for the VP from event cards. In online games the information is 100% available, as the log shows which event cards have been played. So if you really are interested you can already calculate this manually and I think many successful players do this anyways. Especially in online games as there is no real peer/time pressure and you would have all the time in the world to calculate the current VP.

How about making it optional (per game) whether or not you want to show the VP of other players using a switch on the game creation screen for example?

bafolts:
I would make it a game option. I agree with @ssimeonoff and even though it can be easily calculated I always get frustrated when players stop to calculate the VP.

Regarding the arbitrary 64 it was most likely a magic number. If the test passes with your modification I wouldn't worry about it.

src/Game.ts
```diff
@@ -687,55 +690,64 @@ export class Game {
     private gotoEndGame(): void {
       if (this.phase === Phase.END) return;
       this.phase = Phase.END;
+      this.calculateVictoryPoints();
+    }
+
+    public calculateVictoryPoints(): void {
```
bafolts: Instead of creating this `calculateVictoryPoints` function and needing to call it every place that victory points may be applied I would instead make `victoryPoints` a function on `Player` to `getVictoryPoints`. This function signature could then be:

```
public getVictoryPoints(game: Game): VictoryPointsBreakdown
```
This would avoid having to always call to `calculateVictoryPoints` and should move all the victory point calculation logic to one place. A reference to `Game` is always available in `server.ts` where we send this value to the client. The calculation logic in `Game.ts` would then move into this function. I would add the total `victoryPoints` as a property of `VictoryPointsBreakdown` to avoid having to send two values to the client related to the victory points.

gomfunkel: Removing `victoryPoint` from the `PlayerModel` and using only `VictoryPointsBreakdown` definitely makes sense to remove redundancy.

Everything else means that the VP are not re-calculated everytime there is a change but everytime they are requested. Not quite sure if that‘s the right way to do it but will look into it. 

Looks like a bigger refactoring at the first glance, though, so it will take a bit.

bafolts: I can do this part of the refactoring if you would like.

gomfunkel: Nah, thanks, I'm already at it and almost done, no worries, will push tonight or tomorrow.

gomfunkel:
@bafolts Refactored VP calculation as discussed, looking forward to your review. 

Tests checking for VP have been adjusted accordingly and EOS Chasma National Park was fixed along the way as well.

alrusdi:
@gomfunkel can you merge your branch with current master? There are conflicts...

gomfunkel:
@alrusdi Done!

PR #427, opened 2020-03-20
Issue 357

vincentneko:
This is a big one...
- Move resource count inside cards
- Pass resource count to select card components so that resource count is displayed in all circumstances


src/models/PlayerInputModel.ts
```diff
@@ -1,13 +1,14 @@
...
+import { ICard } from '../cards/ICard';
 
 export interface PlayerInputModel {
     amount: number | undefined;
     availableSpaces: Array<string> | undefined;
     canUseHeat: boolean | undefined;
     canUseSteel: boolean | undefined;
     canUseTitanium: boolean | undefined;
-    cards: Array<string> | undefined;
+    cards: Array<ICard> | undefined;
```
bafolts: This should be `cards: Array<CardModel> | undefined`. Anything on client side should use the models which attempt a encapsulate data we can serialize to JSON.

server.ts
```diff
@@ -414,15 +414,13 @@ function getWaitingFor(
       }
       break;
     case PlayerInputTypes.SELECT_HOW_TO_PAY_FOR_CARD:
-      result.cards = (waitingFor as SelectHowToPayForCard)
-          .cards.map((card) => card.name);
+      result.cards = (waitingFor as SelectHowToPayForCard).cards;
```
bafolts: I would create a function `getCards(cards: Array<ICard>): Array<CardModel>` and map these to a `CardModel` for the client.

vincentneko: > One minor change involving using `CardModel` for the client. Everything else looks great.



> One minor change involving using `CardModel` for the client. Everything else looks great.

I've just found a bug, cards cannot be paid with steel or titanium, I will investigate this.

vincentneko:
> CardType.AUTOMATED

Not sure how to handle this as the input object is a `ICard ` and does not have a `cardType` property...
Any suggestion is welcome

vincentneko:
> > CardType.AUTOMATED
> 
> Not sure how to handle this as the input object is a `ICard ` and does not have a `cardType`
> property...
> The card could also be a corporation.
> Any suggestion is welcome

Sleeping on this one, maybe `cardType `property should be moved to `ICard ` so corporations would also have their `cardType `
I'm merging as it is now, we could make the change with another issue.

PR #432, opened 2020-03-22
I18n

alrusdi:
This is draft implementation for i18n
I use custom v-i18n Vue directive.  
Any json files from src/locales used to get translations for UI strings
We will never translate anything on backend


src/directives/TrimWhitespase.ts
```diff
@@ -0,0 +1,8 @@
+
+export function trimEmptyTextNodes (el: any) {
```
bafolts: This should be `TrimWhitespace` and not `TrimWhitespase`.

index.html
```diff
@@ -2,6 +2,7 @@
 <html>
     <head>
         <title>Terraforming Mars</title>
+        <script src="/api/translations.js"></script>
```
bafolts: Loading self-executing javascript in this way can be difficult to test and lead to security issues. I would use `fetch` or `XMLHttpRequest` to request this JSON and store it where it is needed.

alrusdi: Main problem was a requirement to load all translations before Vue app start. After your great review i realized that we can build translations json befor server start. So i made make_locales.js. But it is still needed to load translations first. As result a lot of changes to app initialization (most code moved from script to components/App.ts)

src/directives/TrimWhitespase.ts
alrusdi: This is done

src/PreferencesManger.ts
```diff
@@ -0,0 +1,17 @@
+export class PreferencesManager {
```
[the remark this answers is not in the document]
alrusdi: Done

PR #435, opened 2020-03-22
Player and Tag boards redesigns

ssimeonoff:


ssimeonoff:
sorry, misclicked the fork update

Issue #430, opened 2020-03-21
Feature: Update document title upon poll response from the backend


ssimeonoff:
Suggestion: Update <title>Terraforming Mars</title> based on players' turns.

While api/waitingfor is polling:
- when you get {result: "WAIT", player: "Player 2"}, update the title to, e.g. "Player 2 is playing..."
- when you get {result: "GO", player: "Player 1"}, update the title to, e.g. "YOUR TURN"

Implementation is trivial with few lines of code.
https://developer.mozilla.org/en-US/docs/Web/API/Document/title

For better visualization when whose turn it is. Currently there is no notification if it is your turn if you dont keep the game as an active tab.

![titles](https://user-images.githubusercontent.com/6917565/77225588-b0816280-6b70-11ea-9d84-cbb8a1a6bd97.png)

dabewi:
I'd like that.

Additional idea: Make it so that the TM icon changes if it's your turn. Maybe it could get a blueish border or something like that. Background: I like to "pin" the tab with my TM games (because it takes several days to finish one game and with pinning them in the browser they can't get lost) and I only see the small icon. An altered icon would tell me right away that it's my turn (even if I don't see the "YOUR TURN" text because the tab is pinned).

alrusdi:
Title part is implemented and merged. I don't sure about icon. Is pinned tab requests server for changes each hafsecond of several days? It makes me sad :-)

gomfunkel:
@alrusdi The icon part would be an important addition. When pinning tabs, changing the title does not help at all, a changing icon would. I am not sure about the doubts you have about this: The requests are sent anyways (otherwise changing the title wouldn't make any sense either) and changing the icon is only about updating the favicon which is cached on the client.

gomfunkel:
@alrusdi Thank you for reopening this! I'm also having trouble with the title change which does not seem to be working (at least for me). I commented on the pull request.

nwai90:
@alrusdi Is this issue still relevant? I saw it was merged and the changes were reverted later on

alrusdi:
I think it is relevant (at least title part). It needs more careful implementation than mine :-)

PR #437, opened 2020-03-23
Update document title when turn state is changed; #430

alrusdi:


gomfunkel:
@alrusdi @vincentneko This is not working for me. I tried with two players and both of them are displaying "YOUR TURN" and are stuck there. Is it just me? Have you tested this throughly?

gomfunkel:
Apparently the title keeps saying "YOUR TURN!" until the other player actually takes an action.

PR #440, opened 2020-03-23
Colonies css

ssimeonoff:
Changes:
- Callisto UI bug fix: colony bonus was showing only 1 energy resource instead of 3
- Colonies box-shadow equal to the rest of cards/boards shadows (the huge shadows was used on black background with 3D transforms on /cards-list, not needed here)


alrusdi:
@ssimeonoff can you please merge current upstream master into your branch?

PR #441, opened 2020-03-23
Revert PR changes

alrusdi:


alrusdi:
I've reverted changes about title update PR. There more problems than profit

PR #443, opened 2020-03-23
[Feature] Adding saves to allow undo action

pierrehilbert:
What has been done: 
- Save the game state every time a turn is finished
- Add loadFromJson function to be able to rebuild some objects
- Add undo action when you already played one action (we should only be able to undo an action when it's still our turn)
- Only keep last save when the game is finished 

What should (and will) be improved: 
- [x] Create the table if not exists when the game start
- [ ] Auto-refresh after the undo action (we currently have to do it manually)
- [ ] Add a load game option on the main page to be able to continue a game after a server crash
- [ ] Adding PostgreSQL option instead of sqlite3


alrusdi:
@pierrehilbert please merge it with upsteam master branch

src/Game.ts
```diff
@@ -1166,5 +1187,207 @@ export class Game {
...
     }
+
+    // Function to save the current game state
+    private saveGameState(): void {
+      // Getting the file path
+      let path = require('path');
+      let dbPath = path.resolve(__dirname, '../../db/game.db');
+      let db = new sqlite3.Database(dbPath);
+      // Create the table that will store every saves
+      db.run('CREATE TABLE IF NOT EXISTS games(gameId varchar, saveId integer, game text)');
```
bafolts: The `gameId` column should be a primary key and `saveId` should be an index.

src/Game.ts
```diff
@@ -1166,5 +1187,207 @@ export class Game {
...
+      db.run(`INSERT INTO games(gameId, saveId, game) VALUES(?, ?, ?)`, [this.id, this.lastSaveId, JSON.stringify(this,this.replacer)], function(err: { message: any; }) {
+        if (err) {
+          return console.log(err.message);  
+        }
+      });
+      db.close();
+    }
+
+    // Custom replacer to transform Map and Set to Array
+    public replacer(key: any, value: any) {
```
bafolts: The type of `key` here I would think is `string` or `string | number` which will be more restrictive than using `any`. Similar to `value` I think we know this will be `Set | Map | string` or whatever we expect the inputs to be.

src/Game.ts
```diff
@@ -1166,5 +1187,207 @@ export class Game {
...
+      key = key +"";
+      if (typeof value === 'object' && value instanceof Set) {
```
bafolts: I don't think there is any need to also check that `typeof value === 'object'` if `value instanceof Set` that is all we care about, similar to checking for `instanceof Map` below.

src/Game.ts
```diff
@@ -1166,5 +1187,207 @@ export class Game {
...
+        if (err) {
+          return console.log(err.message);  
+        }
+      });
+      db.close();
+    }
+
+    // Custom replacer to transform Map and Set to Array
+    public replacer(key: any, value: any) {
+      key = key +"";
```
bafolts: If the goal here is to  make a `key` which may be a `number` a `string` using `key = String(key)` should be clearer.

src/Game.ts
```diff
@@ -1166,5 +1187,207 @@ export class Game {
...
+      else if(typeof value === 'object' && value instanceof Map) {
+        return Array.from(value.entries());
+      }
+      return value;
+    }
+
+    // Function to restore previous turn from the database
+    public restoreLastSave(): void {
+      // Getting the file path
+      let path = require('path');
```
bafolts: This should be imported once at the top of the file.

```
const path = import("path");
```

src/Game.ts
```diff
@@ -1166,5 +1187,207 @@ export class Game {
...
+      let dbPath = path.resolve(__dirname, '../../db/game.db');
```
bafolts: I would create a constant for this path that can be re-used. This way when this database file updates the location only need be updated once.

src/Game.ts
```diff
@@ -1166,5 +1187,207 @@ export class Game {
...
+      db.run(`DELETE FROM games WHERE gameId = ? AND saveId < ?`, [this.id, this.lastSaveId], function(err: { message: any; }) {
+        if (err) {
+          return console.log(err.message);  
+        }
+      });
+      db.close();
+    }
+
+    // Function used to rebuild each objects
+    public loadFromJSON(d: Game): Game {
```
bafolts: I would make an interface for `ISerialize` which defines the `loadFromJSON` and corresponding convert to JSON functions. This way we can standardize the function names. I am pretty sure you can use generics similar to:

```
interface ISerialize<T> {
    loadFromJSON(d: T): T;
    convertToJSON(): string;
}
```

src/Game.ts
```diff
@@ -1166,5 +1187,207 @@ export class Game {
...
+      // Assign each attributes
+      var o = Object.assign(this, d);
```
bafolts: We don't have a linting rule for this (it would be nice to have!) but we should use `const` if a variable is not going to change and try to use `let` over `var`.

src/Game.ts
```diff
@@ -1166,5 +1187,207 @@ export class Game {
...
+
+      // Rebuild every player objects
+      this.players = new Array<Player>();
```
bafolts: This is a great place to use `map` instead of `forEach` since you are mapping `d.players` to `this.players`.

```
this.players = d.players.map((element: Player)  => {});
```

This probably goes for most other array conversions as well.

src/Game.ts
```diff
@@ -1166,5 +1187,207 @@ export class Game {
...
+      this.milestones = [];
+      this.awards = [];
+      this.board = this.boardConstructor(d.boardName);
+      d.board.spaces.forEach((element: ISpace) => {
+        if(element.tile) {
+          var space = this.getSpace(element.id);
+          var tileType = element.tile.tileType;
+          var tileCard = element.tile.card;
+          if (element.player){
+            let playerIndex: number = this.players.map(function(x) {return x.id; }).indexOf(element.player.id);
```
bafolts: I would use `Array.prototype.findIndex` here and the other places you have used this.

```
let playerIndex: number = this.players.findIndex((player) => player.id === element.player.id);
```

src/components/OtherPlayer.ts
```diff
@@ -23,6 +23,24 @@ export const OtherPlayer = Vue.component("other-player", {
...
+            let count: number = 0;
+            for (let index = 0; index < this.player.playedCards.length; index++) {
+                if (this.player.playedCards[index].cardType === CardType.EVENT) {
+                    count++;
+                } 
+            }
+            return count;
+        },
+        getActiveCards: function() {
+            let cards: Array<CardModel> = [];
```
bafolts: This is a great place to use `Array.prototype.filter`:

```
return this.player.playedCards.filter((card) => card.cardType === CardType.ACTIVE));
```

src/components/OtherPlayer.ts
```diff
@@ -23,6 +23,24 @@ export const OtherPlayer = Vue.component("other-player", {
         },
         isVisible: function () {
             return (this.$root as any).getOtherPlayerVisibility(this.player.id);
+        },
+        getEventCount: function() {
+            let count: number = 0;
+            for (let index = 0; index < this.player.playedCards.length; index++) {
```
bafolts: This is another great place to use `Array.prototype.filter`:

```
return this.player.playedCards.filter((card) => card.cardType === CardType.EVENT).length;
```

src/Game.ts
pierrehilbert: Primary key can only be on unique column but you're right, I will add one PK on (gameId, saveId).

src/Game.ts
pierrehilbert: Good point!

src/Game.ts
pierrehilbert: It's just that I never use the key part and it's a bad solution to avoid an error...

src/Game.ts
pierrehilbert: You're right, I corrected it.

src/Game.ts
pierrehilbert: Yes and it will remove some duplicate code.

src/Game.ts
pierrehilbert: I changed it thx!

src/components/OtherPlayer.ts
pierrehilbert: Wasn't from my code but I include the modification in it ;-)

src/components/OtherPlayer.ts
pierrehilbert: Corrected it too.

src/Game.ts
pierrehilbert: It's cleaner thx!

pierrehilbert:
> @pierrehilbert please merge it with upsteam master branch

I merged it but I have a trouble because there is no translation.json file.

alrusdi:
@pierrehilbert run "npm run build" i made PR to fix it

src/Game.ts
pierrehilbert: @bafolts : do you have a solution or can I close this reviewed?

src/Game.ts
pierrehilbert: Done!

pierrehilbert:
I need help for the refresh part...
I don't know how I can trigger the ui refresh after the restore...

bafolts:
When one player wants to undo their last action another player is actively taking their turn. We don't poll the server today if you are the active player taking your turn but in order to support this we would need to poll the server to determine if it is no longer a player's turn due to an undo.

Assuming we have a stack of players actions if you are the player with the action on top of the stack you should see the undo button. If you are the player who gets interrupted by an undo I would show a dialog alert indicating which player undid their last action and then take you back to the waiting screen. We use modals one place already and I would re-use it here since whatever a player is doing on the screen they should stop doing if a user is going to undo. Hopefully players are chatting outside of the game (want to put off becoming a chat application as long as I can fend it off) so they know this is coming.

src/Dealer.ts
```diff
@@ -883,4 +883,36 @@ export class Dealer {
         }
         return result;
     }
+  
+    // Function to return a card object by its name
+    private getProjectCardByName(cardName: string): IProjectCard | undefined {
```
bafolts: This will have merge conflicts with the recent change #463 . The dealer now has `ICardFactory` instances and not instances of cards. This will need to copy (or re-use) what we do in `src/components/Card.ts`.

https://github.com/bafolts/terraforming-mars/blob/master/src/components/Card.ts#L43

I would add this as a method to `Dealer` and have `src/components/Card.ts` use it as well.

src/Game.ts
alrusdi: @pierrehilbert try to use underscore for unused arg https://stackoverflow.com/questions/38835001/skip-type-check-on-unused-parameters

src/Player.ts
```diff
@@ -1752,4 +1769,75 @@ export class Player {
       this.waitingFor = input;
       this.waitingForCb = cb;
     }
-}
\ No newline at end of file
+
+    // Function to return a corporation card object by its name
+    private getCorporationCardByName(cardName: string): CorporationCard | undefined {
```
bafolts: I would create static methods on `Dealer` for these places where we need an instance of a card. This logic is also duplicated in `src/components/Card.ts` and should be consolidated.

https://github.com/bafolts/terraforming-mars/blob/master/src/components/Card.ts#L20

```
Dealer.getCorporationCardByName
Dealer.getProjectCardByName
```
```

File: src/Player.ts [hunk 15]
```diff
@@ -1752,4 +1769,75 @@ export class Player {
...
+    }
+
+    // Function used to rebuild each objects
+    public loadFromJSON(d: Player): Player {
+      // Assign each attributes
+      let o = Object.assign(this, d);
+
+      // Rebuild generation played map
+      this.generationPlayed = new Map<string, number>();
+      d.generationPlayed.forEach((element: any) => {
```
bafolts: I am pretty sure you can provide your stored array to the `Map` constructor as follows.

```
this.generationPlayed = new Map<string, number>([d.generationPlayed]);
```

src/Player.ts
```diff
@@ -1752,4 +1769,75 @@ export class Player {
...
+        this.generationPlayed.set(element[0], element[1]);
+      });
+
+      // action this generation set
+      this.actionsThisGeneration = new Set<string>();
```
bafolts: I am pretty sure you can provide your stored array to the `Set` constructor as follows.

```
this.actionsThisGeneration = new Set<string>(d.actionsThisGeneration);
```

src/Player.ts
```diff
@@ -1752,4 +1769,75 @@ export class Player {
...
+      d.actionsThisGeneration.forEach((element: any) => {
+        this.actionsThisGeneration.add(element);
+      });
+
+      // Rebuild corporation card
+      this.corporationCard = this.getCorporationCardByName(d.corporationCard!.name)
```
bafolts: I would protect against `d.corporationCard` being `undefined`. Granted we are only calling this now once a corporation has been selected but really there is no reason to lock us into that state since `corporationCard` can be `undefined` prior to a user selecting their corporation.

```
if (d.corporationCard !== undefined) {
    this.corporationCard = this.getCorporationCardByName(d.corporationCard.name);
} else {
    this.corporationCard = undefined;
}
```

src/Player.ts
```diff
@@ -1752,4 +1769,75 @@ export class Player {
...
+    private getProjectCardByName(cardName: string): IProjectCard | undefined {
+      return ALL_PROJECT_CARDS.find((card) => card.name === cardName) 
+      || ALL_PRELUDE_CARDS.find((card) => card.name === cardName) 
+      || ALL_PRELUDE_PROJECTS_CARDS.find((card) => card.name === cardName)
+      || ALL_VENUS_PROJECTS_CARDS.find((card) => card.name === cardName)
+      || ALL_COLONIES_PROJECTS_CARDS.find((card) => card.name === cardName);
+    }
+
+    // Function used to rebuild each objects
+    public loadFromJSON(d: Player): Player {
```
bafolts: When loading from JSON this function is not provided a `Player`. All we know is that it is provided an `object`. We can make no assumptions about what was stored as we are parsing a string into this object with `JSON.parse`. All these `loadFromJSON` should be `d: object` or some defined `Model` representing the data format we stored.

src/Player.ts
```diff
@@ -1752,4 +1769,75 @@ export class Player {
...
+      this.actionsThisGeneration = new Set<string>();
+      d.actionsThisGeneration.forEach((element: any) => {
+        this.actionsThisGeneration.add(element);
+      });
+
+      // Rebuild corporation card
+      this.corporationCard = this.getCorporationCardByName(d.corporationCard!.name)
+
+      // Rebuild deal corporation array
+      this.dealtCorporationCards = d.dealtCorporationCards.map((element: CorporationCard)  => {
```
bafolts: Similar to my comment above regarding `object`. The items in `dealtCorporationCards` are not instances of `CorporationCard` but that of the format we used when storing them prior to calling JSON.stringify. Typescript takes our word when we provide a type. This should be `object` or some interface representing the data format we stored.

src/Player.ts
```diff
@@ -1752,4 +1769,75 @@ export class Player {
...
+        return this.getCorporationCardByName(element.name)!;
```
bafolts: If for whatever reason these functions return `undefined` we should throw an error of some kind instead of pushing on `undefined`. When I was doing my re-factor yesterday for changing cards to use factories it was very helpful in the other places where we throw instead of using `!`. We should avoid using `!` and `any` as much as possible.

src/colonies/ColonyDealer.ts
```diff
@@ -11,23 +11,25 @@ import { Miranda } from './Miranda';
 import { Pluto } from './Pluto';
 import { Enceladus } from './Enceladus';
 
+// ALL COLONIES TILES is now a const not and attribute of Colony Dealer
+export const ALL_COLONIES_TILES: Array<IColony> = [
```
bafolts: Moving this out from the private instance is going to cause the issue we ran into with cards where different games are going to re-use the same instance of a colony. If taking this approach we will need `IColonyFactory` similar to `ICardFactory` to get new instances of a colony.

src/Game.ts
pierrehilbert: Many thx !

src/Game.ts
```diff
@@ -1062,4 +1092,196 @@ export class Game {
       });
       return undefined;
     }
-}
\ No newline at end of file
+
+    // Function to save the current game state
+    private saveGameState(): void {
+      // Open the database connexion
+      let db = new sqlite3.Database(dbPath);
```
bafolts: If using a database I would keep an open connection to the database that can be re-used instead of always creating a connection to the database. For now you could make a singleton `Database` which provides access to this database. We could also hide the queries in this class removing the query logic from `Game` since game doesn't care how the data gets stored. I believe you can create this database in memory be providing your path as `:memory:` when calling `sqlite3.Database(":memory:")` to avoid writing to a file which is unnecessary for our usage case.

src/Game.ts
bafolts: If you never use `key` you can indicate the argument is unused by prefixing it's name with a `_`.

```
public replacer(_key: any, value: any)
```

src/Player.ts
```diff
@@ -1421,6 +1424,15 @@ export class Player {
       });
     }
 
+    // Propose a new action to undo last action
+    private undoTurnOption(game: Game): PlayerInput {
+      return new SelectOption("Undo Turn", () => {
+        //this.actionsTakenThisRound = 0;
+        game.restoreLastSave();
```
bafolts: If there is an error with `restoreLastSave` I would throw an error to allow the user to know we were unable to restore their last save. We should leverage transactions if possible with the database so if there is some error with our query none of it goes through.

src/Game.ts
pierrehilbert: Not using the memory one is a choice: it will allow us to load a save after a server crash for example.

bafolts:
I think we will eventually get there with the work being done in this pull request. Sorry if I am nit picky. I may be salty after yesterday where most things were broken and I had to do a large refactor. During that refactor I saw some pain points with our usage of `!` and `any` which led to bugs I only found once running the code instead of having typescript catch them. I need to add more linting rules since more contributors are showing up.

I would create a `Database` class for use with storing data and load from `:memory:` if possible to avoid needing to write to disk. We don't need the database file around when the game server is stopped so for the time being `:memory:` meets our needs.

src/Game.ts
bafolts: If not using memory we will eventually have a massive database on disk and may need a strategy at some point for removing old games.

pierrehilbert: I add a cleanSaves() function called at the end of the game to clean old saves but you're right, maybe we should add a purge system and add a creation_date.

src/Player.ts
pierrehilbert: Done

src/Dealer.ts
pierrehilbert: Done

src/Game.ts
```diff
@@ -1070,4 +1100,197 @@ export class Game {
...
+      return value;
+    }
+
+    // Function to restore previous turn from the database
+    public restoreLastSave(): void {
+      // Open the database connexion
+      let db = new sqlite3.Database(dbPath);
+
+      // Retrieve last save from database
+      db.get("SELECT game game FROM games WHERE game_id = ? AND save_id = ? ORDER BY save_id DESC", [this.id, this.lastSaveId],(err: { message: any; }, row: { game: any; }) => {
```
bafolts: Only the row with the maximum value for `save_id` is needed. I would include ` LIMIT 1` to indicate we are only concerned with the first item. This could allow the database to only give us the maximum and not need to sort all of the rows.

src/Game.ts
```diff
@@ -1070,4 +1100,197 @@ export class Game {
...
+      db.run("DELETE FROM games WHERE game_id = ? AND save_id < ?", [this.id, this.lastSaveId], function(err: { message: any; }) {
+        if (err) {
+          return console.log(err.message);  
+        }
+      });
+      db.close();
+    }
+
+    // Function used to rebuild each objects
+    public loadFromJSON(d: Game): Game {
```
bafolts: Hoping to clarify since I am still seeing `Game`, `Player`, and `Dealer` passed into these `loadFromJSON` methods. When you call `JSON.stringify` on a class you get the properties that can be turned into JSON in an `object`. What is being provided to these `loadFromJSON` is `object` that happens to have some of the properties of `Game`, `Player`, or `Dealer`. If you were to call `d.someFunction()` it would not exist but typescript will not know that it does not exist since we have told typescript that we are being provided `Game`. We should created `SerializedGame`, `SerializedPlayer` and `SerializedDealer` which indicates the fields that we can serialize. These can be interfaces. This will help us when refactoring and it will allow typescript to help us.

For example:

```
class A {
    public doSomething() {
    }
    public aField: number = 1;
}

var a = new A();
JSON.stringify(a);

// Since JSON has no representation of a function this string will not have a `doSomething` property. When we call JSON.parse we will have an object with an `aField` only.
```

This may work find right now but as time passes and we forget that this class does not get a `Game` but an `object` that looks like a game we will inevitably introduce some bug.

src/Player.ts
```diff
@@ -1421,6 +1423,19 @@ export class Player {
...
     }
 
+    // Propose a new action to undo last action
+    private undoTurnOption(game: Game): PlayerInput {
+      return new SelectOption("Undo Turn", () => {
+        try {
+          game.restoreLastSave();
+        }
+        catch(error){
+          console.log(error);
```
bafolts: This should be `console.error` or `console.warn` and we should provide a message to help find this.

bafolts:
OK. This all looks fine except the `loadFromJSON` functions. These functions need to take in the appropriate type signature. We will need to create an interface to represent what ends up getting serialized when we call `JSON.stringify`. Once we have that in place this looks good to merge to me. Each `loadFromJSON` should have an interface that only has the properties of `Game`, `Player`, `Dealer`, etc that end up in the `JSON.stringify` string. These can be interfaces.

src/Game.ts
pierrehilbert: I know that I don't have function after retrieving from the stringify and it's why I recreate object with this "loadFromJSON" functions.
If I understand well, you want me to add some Interface having only fields in it instead of "casting" as the real object before recreating it?

bafolts: Yes. This will allow typescript to do it's job of compile time type checking.

bafolts:
This now looks great to me. :pray: :shipit: :see_no_evil:

Issue #428, opened 2020-03-20
Feature request: Actions log to show the card preview for card-related actions(played/activated)


ssimeonoff:
![log](https://user-images.githubusercontent.com/6917565/77176117-e9173280-6ac3-11ea-83e1-b948a2ae1451.png)
In the log:
- Player name to be displayed in the player's color
- Card name to be displayed as it is in the card (matching colors and fonts)
- When the card name is clicked, the preview div is displayed (X dismisses it)

Notes:
I don't know the cards' names, even after 1000+ games, I know only a handful. 
When I played 1vs1 I kept the Cards List page open in another tab and was copy/past -ing there every card name my opponent played/activated.

alrusdi:
It is a good idea! We will implement it, but not right now

gomfunkel:
Started working on this (at least the coloring part, not yet the preview div) tonight and already made some progress. Would be great if someone could assign me to this.

Early Screenshot:
<img width="788" alt="colors" src="https://user-images.githubusercontent.com/547035/77365314-30b4ed00-6d56-11ea-9386-33592e148453.png">

I had to put the player names on a lighter background as well, otherwise it would be impossible to see the black player.

ssimeonoff:
@gomfunkel 
Is that an easy option to do as well: (generations breakdown)
![Untitled](https://user-images.githubusercontent.com/6917565/77366999-84750580-6d59-11ea-81c4-2f9c0d0cf233.png)

And also, is it possible to fix a bug in this pr as well: 
"End Turn" action does not log anything. 
E.g. "Player 1 turn ended"

gomfunkel:
Great ideas, @ssimeonoff! Should be possible as the log is just a list of strings as it is now. Will see what I can do.

gomfunkel:
I have implemented the following ideas from this issue and submitted pull request #446. 

* Using player colors
* Using card type colors and font
* Added log message when ending turn
* Added log message on new generation

Displaying the cards on click in the logs is still to be done.

Here’s what the log now looks like:
<img width="786" alt="Screenshot 2020-03-24 at 10 18 44" src="https://user-images.githubusercontent.com/547035/77419911-75c63700-6dc9-11ea-83b5-920f41ed3e4c.png">

pierrehilbert:
I have done some test and I have this result:
[...]

PR #446, opened 2020-03-24
Optimized log panel (#428)

gomfunkel:
Implemented the following enhancement ideas from #428:

* Using player colors
* Using card type colors and font
* Added log message when ending turn
* Added log message on new generation

See issue #428 for screenshots and details.

I am not totally happy with the amount of code this requires. If someone has a better idea on how to solve this, please let me know.


gomfunkel:
@alrusdi The formatting is actually done in the frontend (`LogPanel.ts`). The backend only adds the semantics (in form of html/xml tags), this is required to enable formatting. 

The log messages are rather arbitrary and don't follow a precise pattern (0-n players being mentioned in different places, 0-n cards being mentioned in different places, etc.). I'm not sure if it's better to solve this with a LogMessage interface or similar, at least I don't have a clear image in mind how the interface should look. Do you have a good idea?

It is likely possible to get rid of the `name` attribute and only rely on the `innerHTML` of the `log*` tags to make it more readable, though. Would that already help?

alrusdi:
> The backend only adds the semantics (in form of html/xml tags), this is required to enable formatting.
Why not to add this semantics in json format not html/xml tags?
So each log message can be represent with json like this:
```
{
    "type": "card_played",
    "text": "{} played {}"
    "args": [{"id": "123", "title": "player1"}, {"id": "01", "title": "Colonizer Training Camp"]
}
```
And Log component can render each message type differently (by using "if" statement inside template or even other components)

gomfunkel:
Looks like the responses could be a bit large regarding byte size but I do see the advantages in the interface approach as well. Using an interface it would definitely be more easy to add additional information/metadata (timestamps for example).

Any other opinions on this?

alrusdi:
> Any other opinions on this?
Optimization by response size should be considered latter. We can gzip it, or split requests as necessary. Premature optimization is the root of all evil ;-)

bafolts:
I think this approach is fine. If for whatever reason the size did prove to be an issue changing it to a JSON object later should not be difficult. This approach here is at least simple. If we were to use some JSON object we would only be making our own data structure which represented a vue.js template anyway. That being said, with this approach we will need to escape any values we put inside `""` or `><` or we would have invalid XML. That is the downside of this approach. If using JSON we wouldn't have to do this escaping, we would only need put into a JSON object, call `JSON.stringify` and then `JSON.parse` inside the log component on the client side. Escaping the string and the XML will require two functions which are not straight forward.

By the time I finished writing that paragraph I have talked myself into wanting to serialize to some `LogModel` JSON object that we pack on the server and then the component uses the `LogModel` to create the vue templates it needs. This will avoid us having to escape the data for the Vue.js template.

alrusdi:
> By the time I finished writing that paragraph I have talked myself into wanting to serialize to some `LogModel` JSON object that we pack on the server and then the component uses the `LogModel` to create the vue templates it needs. This will avoid us having to escape the data for the Vue.js template.

It is exactly what i want. Of course i think about i18n too. It is much easier to translate base message + params than one big blackbox string.

gomfunkel:
OK, got it. Will introduce a model for the log (messages) and update this pull request as soon as I'm done.

gomfunkel:
Refactored the game logging according to our discussion, @bafolts and @alrusdi.

alrusdi:
@gomfunkel can you please merge upstream master branch into your branch?

gomfunkel:
@alrusdi Conflicts resolved.

alrusdi:
I like this version. I can try translate all the messages into Russian after merge and check if this logging work for us perfectly

PR #451, opened 2020-03-24
Fix resource decrement

bradwerth:
These changes allow the SelectHowToPay inputs to refund an appropriate amount of MegaCredits when reducing the amount of a resource allocated to payment. The canonical example:

* The player is paying for a 10 MC card with space tag, and the player has 4 titanium.
* The player allocates 3 titanium and sees 1 MC will be spent.
* The player increases titanium to 4 and sees 0 MC will be spent.
* The player decreases titanium to 3.

Without these changes, the MC will be set to 3, resulting in a situation where the player might accidentally overpay. With these changes, the MC will be set to 1, which is the minimum MC needed to fund the purchase with the 3 titanium.

Additionally, these changes prevent a player from paying more MC than a card costs. They can no longer allocate MC up to the amount they have; only enough to pay for the card. They can still overpay with resources.

### DIFF
--- a/src/components/PaymentWidgetMixin.ts
+++ b/src/components/PaymentWidgetMixin.ts
@@ -2,8 +2,8 @@
 
 export const PaymentWidgetMixin = {
     "methods": {
-        getCardCost: function (): number {
-            return (this as any).player.megaCredits;
+        getMegaCreditsMax: function (): number {
+            return Math.min((this as any).player.megaCredits, (this as any).getCardCost());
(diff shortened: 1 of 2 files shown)


bradwerth:
I see that this doesn't decrement MC properly for payments ON cards, such as Water Import From Europa. I'll work on a fix.

bradwerth:
I'll withdraw this until the fixes are ready.

Issue #461, opened 2020-03-25
Some cards start with resources on them randomly


ssimeonoff:
Forced Precipation has 1 floater even before played.
![90721949_905081343295362_6905203674092404736_n](https://user-images.githubusercontent.com/6917565/77518778-a320ec80-6e7e-11ea-8776-1d1f248a4bc3.png)

Rotator Impacts force you to REMOVE one assteroid on Gen 1. 
![90776823_232631741127119_7577169548277710848_n](https://user-images.githubusercontent.com/6917565/77518785-a4eab000-6e7e-11ea-901d-16850de84636.png)

The bug was observed with other cards holding floaters and microbes. 

E.g. Played Aeriel Mappers round one. Started with 4 floaters on it.

So far, I have experienced the bug only in Solo.

alrusdi:
I believe there are two different bugs.
Obviously wrong logic in RotatorImpacts catAct method, and some kind of multiple games on the same server clash in Forced Precipation

alrusdi:
I've fixed RotatorImpacts
Resources on cards is magic for me. I can't reproduce it on the localhost.
@bafolts @vincentneko can it be result of initialisation of cards on module level in Dealer.ts
I mean, can different instances of the Game use the same instances on Card?

alrusdi:
Domn... It is reproductible:
1) Add this code in Game constructor to put the same card into hand of player
`      const venusCards = ALL_VENUS_PROJECTS_CARDS.slice()
      first.cardsInHand.push(venusCards[12]);`
2) Start two solo games with Venus extension enabled
3) In first game play and activate Forced Precipitation card to have resources on it
4) In second game just refresh the page and you will see unwanted resources on Forced Precipitation card

ssimeonoff:
This is a critical bug: 
Game 1: The player with Aridor played tags and got production as expected.
Game 2: The player with Aridor played Jovian and Space tags, but got nothing, because in another ongoing game (Game1), Aridor already has those tags played. The Science tag gave 1 production as expected, because it is not played in Game1.
![games](https://user-images.githubusercontent.com/6917565/77533370-0a4a9b00-6e97-11ea-82e8-78825f073f4e.png)

bafolts:
There was a recent change which moved the logic for tracking resources on a card from the game to the instances of the cards themselves. The dealer is providing different games the same instance of the card now as @alrusdi suggested. The change will need to be reverted or the dealer will have to provide new instances of the cards to players.

[...]

PR #463, opened 2020-03-25
Use card factories to generate new card instances for Dealer for #461

bafolts:
We are sharing instances of cards between games. A recent change to store the resource count on the cards exposed this issue. This change introduces an `ICardFactory` interface which has the constructor for a given `CardName`.

```js
interface ICardFactory<T> {
    cardName: CardName;
    factory: new () => T;
}
```

The bulk of this work is porting everything to use these factories.


bafolts:
@alrusdi or @vincentneko I will work on this in 8-9 hours or so and finish the remainder of the conversion. The interface is introduced and I only need convert all of the cards. If someone else beats me to finishing the conversion that is fine.

bafolts:
@alrusdi or @vincentneko this is ready to merge. A new instance of a card is always provided to the dealer for a game to avoid issues where we are storing state on a card.

PR #471, opened 2020-03-26
Compact award and milestones

alrusdi:
I've joined Milestones list with Claimed Milestones list. Now they take less place and we have less code duplication.
The same done for Awards
awards_and_milestones


rzulian:
@alrusdi I would suggest to add the claimed milestones and funded awards also in the not expanded lists, so you can have the status at a glance without needing to expand the lists, which is something that you don't do often?

Something like:

Awards: Terraformer Xxxx, Mayor yyy
Milestones: Banker, Miner

alrusdi:
> @alrusdi I would suggest to add the claimed milestones and funded awards also in the not expanded lists

Nice idea! I will add it to this PR

alrusdi:
I've implemented visuals suggested by @ssimeonoff 
Снимок экрана от 2020-03-27 15-02-59
Снимок экрана от 2020-03-27 15-02-52

rzulian:
@alrusdi Awesome!

server.ts
```diff
@@ -321,27 +323,52 @@ function createGame(req: http.IncomingMessage, res: http.ServerResponse): void {
   });
 }
 
+function getMilestones(game: Game): Array<ClaimedMilestoneModel> {
+  const allMilestones = game.milestones;
+  const claimedMilestones = game.claimedMilestones;
+  let milestoneModels: Array<ClaimedMilestoneModel> = [];
+
+  for (let idx in allMilestones) {
+    let claimed = claimedMilestones.filter((m) => m.milestone.name === allMilestones[idx].name)
```
bafolts: Since you only want the first item you should use `.find` and not `.filter`. This is also slightly more efficient as the iteration stops after the first hit.

server.ts
```diff
@@ -321,27 +323,52 @@ function createGame(req: http.IncomingMessage, res: http.ServerResponse): void {
...
+  return milestoneModels;
+}
+
+function getAwards(game: Game): Array<FundedAwardModel>  {
+  const allAwards = game.awards;
+  const fundedAwards = game.fundedAwards;
+  let awardModels: Array<FundedAwardModel> = [];
+
+  for (let idx in allAwards) {
+    let funded = fundedAwards.filter((a) => a.award.name === allAwards[idx].name)
```
bafolts: Same as my comment above, using `.find` closer represents what this is trying to do.

src/components/Award.ts
```diff
@@ -13,17 +13,29 @@ export const Award = Vue.component("award", {
         },
         isVisible: function () {
             return (this.$root as any).getVisibilityState("awards_list");
+        },
+        getNameCss: function(awardName: string): string {
+            return "ma-name ma-name--" +  awardName.replace(' ', '-').toLowerCase();
```
bafolts: I don't know if this is a problem but in javascript this usage of replace will only replace the first instance. If you want to replace all space characters you will need to use a regular expression for what you want to replace.

Issue #418, opened 2020-03-18
Suggestion: Display events in opponents cards


dabewi:
Hi,

I just recently noticed that the events that opponents had played are not displayed when I take a look at the cards of an specific opponent.

I want to suggest to return to displaying them, because otherwise I don't have access to all information that I would have access to in a RL board game, e.g. the number of played events (for one of the milestones) and the VP on them.

Alternatively, if you want to continue hiding the events, please introduce a label which states the amount of played event cards.

Cheers!

ssimeonoff:
In the real game, you don't look into other player's events - you can only ask how many they have.

And about the count,  it is already implemented as a side effect. 
The new Tag count board is visible for for other players and there you have an Event tag, which effectively tells you how many Events are played by that player.

alrusdi:
@ssimeonoff great explanation! I think we don't need the feature to display opponents event cards.

PR #472, opened 2020-03-26
Improved player layout

rzulian:
I was moving milestones and awards on top, and I've noticed that @alrusdi worked on this.
So I've only reordered few things and corrected few bugs:
1. last actions after the turn order.
2. moved other player's cards after player's cards
3. enabled event cards for other players


src/components/OtherPlayer.ts
```diff
@@ -67,7 +67,7 @@ export const OtherPlayer = Vue.component("other-player", {
                         </div>
 
                         <stacked-cards :cards="getCardsByType(player.playedCards, [getAutomatedCardType(), getPreludeCardType()])" ></stacked-cards>
-
+                        <stacked-cards :cards="getCardsByType(player.playedCards, [getEventCardType()])" ></stacked-cards>
```
alrusdi: This is incorrect. In real games we don't see other player's event cards.

src/components/PlayerHome.ts
```diff
@@ -89,11 +89,16 @@ export const PlayerHome = Vue.component("player-home", {
                     <global-parameters :oceans_count="player.oceans" :oxygen_level="player.oxygenLevel" :temperature="player.temperature" v-trim-whitespace></global-parameters>
 
                     <div v-if="player.players.length > 1" class="player_home_block--milestones-and-awards">
-                        <milestone :milestones_list="player.milestones" expanded=false />
-                        <award :awards_list="player.awards" expanded=false />
+                        <milestone :milestones_list="player.milestones" :expanded=false />
```
alrusdi: Remove expanded prop it is not used anymore

src/components/PlayerHome.ts
```diff
@@ -89,11 +89,16 @@ export const PlayerHome = Vue.component("player-home", {
...
+                        <award :awards_list="player.awards" :expanded=false />
```
alrusdi: Remove expanded prop it is not used anymore

src/components/PlayerHome.ts
```diff
@@ -159,6 +154,14 @@ export const PlayerHome = Vue.component("player-home", {
                     <stacked-cards :cards="getCardsByType(player.playedCards, [getAutomatedCardType(), getPreludeCardType()])" ></stacked-cards>
                     <stacked-cards :cards="getCardsByType(player.playedCards, [getEventCardType()])" ></stacked-cards>
                 </div>
+
```
alrusdi: Other player components should be right under Turn order because they are becoming visible by clicking on player name in Turn order widget

alrusdi: Can you please give us a screenshot of how it will be look like after merging this changes?

src/components/OtherPlayer.ts
rzulian: Well, it's actually correct and it has largely debated elsewhere. Fryxelius, the designer, stated here https://boardgamegeek.com/thread/1956557/article/28537673#28537673 that the event cards are turned face down for practically reasons (not counting tags), but information is not hidden. In any case, the work the @gomfunkel is doing on the log is largely overcoming that need, which is have a clue on what the last event played by an opponent is all about.
On the other side, if you think about it, also the VPs shouldn't be calculated for your same reason.
What do you prefer I do?

gomfunkel: This was also part of our discussion regarding real-time VP calculation and one of the reasons we decided to make displaying the VP of other players optional. It was also talked about in #418.

I do understand people trying to keep it a bit more unpredictable/interesting by hiding information but after all the information is there already, just not accessible easily in all cases. Personally I'd vote for showing the event cards but that's just my two cents.

alrusdi: Reasonable. Let's show them. If some users will not want them we will make them optionsl

src/components/PlayerHome.ts
rzulian: This is the current PR layout:
Terraforming Mars current

This is the full layout I would suggest for a large screen:
Terraforming Mars proposed full screen

Other player components are always visible on the bottom. Maybe a shortcut on the right  preferences_cont to jump to them?

I know that layout is always debatable. This is the feedback I've got from my play group.

alrusdi: I don't like position of other player far below its activation link. If @bafolts and @vincentneko is ok with this, i will accept it too.

rzulian: My idea is to have them always visible, but below your board.
The main reason is that you usually deal with the main board with last actions, and your board.
Then you need to check what other are doing or their statuses.
But again, nothing will be perfect ;)

ssimeonoff: My private layout is similar to what you have done. 
1. Move the Venus board away and increase the global bars size (so when the browser zoom is 75% to be able to see them properly

2. Other players areas are capped at 1200x1200 (seeing 4 columns and 3 rows) are inlined and lighter (I tricked myself few times by looking at another player's board)

3. And many other minor css changes, e.g.


Now, when I think of it, maybe I should PR it...

rzulian: @ssimeonoff Very nice! How does it look the Player board? Tags, Resources, Actions, Last actions, Cards in hand and played cards?

alrusdi: @ssimeonoff does it work for small screens too?

ssimeonoff: The Venus board: No, should change the display to "flex" instead of "absolute" - currently it goes off screen on smaller screens. 
The other players boards - should not affect the current layout. 
And the css is trivial.

ssimeonoff: Trying to add some flare to it:
ui

PR #473, opened 2020-03-26
Small cards: Fixing stacked cards

pierrehilbert:


src/styles/preferences.less
```diff
@@ -75,7 +75,9 @@
  /* Reduce all cards size */
  .preferences_small_cards .filterDiv {
     transform: scale(0.8);
-    margin: -22px -15px;
+    margin-top: -22px;
```
alrusdi: Just:
margin: -22px 15px 0 15px;
please :-)

pierrehilbert: Done

PR #475, opened 2020-03-27
Allow to choose board randomly when creating a game

gomfunkel:
This adds the option to play on a random board (from the three available) when creating a game. For times when you just can't decide...


pierrehilbert:
I like this idea too!

PR #517, opened 2020-03-31
UI overhaul 

ssimeonoff:
UI changes to the boards.

css

Squashed it, but then had to fix two other small things...


alrusdi:
Do we even need this color bar with current player name? Let's remove it!

ssimeonoff:
@alrusdi, I didn't add it, someone else did - I just add max-width to it, because it was spanning through the whole page, as q easy fix.

Issue #496, opened 2020-03-29
BUG: Players with the same player name will confuse the log window color assignment.


FreeLikeGNU:
Replicate:
Give more than one player the same name.

![Screenshot_2020-03-29_11-12-56](https://user-images.githubusercontent.com/674095/77856788-5b67d100-71ae-11ea-81bc-503507fa8448.png)

![Screenshot_2020-03-29_11-19-39](https://user-images.githubusercontent.com/674095/77856933-6e2ed580-71af-11ea-8b23-b250ca660294.png)

FreeLikeGNU:
Seems as though the log only records and refers to player by name.  Thus the last entry added that matches is parsed and used to select the color of the log text.

gomfunkel:
The reason for this is that all log messages simply save the _name_ of the player. Generating the log iterates over all players trying to find one with a matching name. We then use the color of the first one who matches.

See [LogPanel.ts#L31](https://github.com/bafolts/terraforming-mars/blob/94a86f47fb2108aaddc1b6b205c362f5bf7414b0/src/components/LogPanel.ts#L31)

To fix this, we need to save a more unique identifier when recording the log, for example the `playerId`. Then replacing it with the according name and color when rendering the log. We could do that but need to make sure that we are not (accidentally) exposing the `playerId` to other players.

gomfunkel:
@FreeLikeGNU Fixed and merged with #523.

FreeLikeGNU:
It's working well now thank you!

PR #523, opened 2020-03-31
Use playerId for log messages (#496)

gomfunkel:
Fixed issue #496 to always use the correct player color. Also includes the fix from pull request #521 regarding the missing colony name.

Please check if this exposes the `playerId` in a way we don't want to.


alrusdi:
I don't sure if we really need this. It is better to validate CreateGame form to disallow identical names and colors for players.

alrusdi:
@gomfunkel please resolve conflicts with current upstream branch and we will merge your PR

FreeLikeGNU:
> ... I don't want to tell people they can't play if they have the same name.

Especially if their names happen to be "Bruce"!

gomfunkel:
@alrusdi Done!

PR #525, opened 2020-04-01
Layout improvement

rzulian:
I've did something wrong and Improved player layout #472 didn't come through

### DIFF
--- a/src/components/OtherPlayer.ts
+++ b/src/components/OtherPlayer.ts
@@ -66,6 +66,7 @@ export const OtherPlayer = Vue.component("other-player", {
                         </div>
 
                         <stacked-cards :cards="getCardsByType(player.playedCards, [getAutomatedCardType(), getPreludeCardType()])" ></stacked-cards>
+                        <stacked-cards :cards="getCardsByType(player.playedCards, [getEventCardType()])" ></stacked-cards>                    
 
 
                     </div>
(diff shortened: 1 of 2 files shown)


rzulian:
@alrusdi could you please merge this?

alrusdi:
> @alrusdi could you please merge this?

I'm sorry i still don't agree with suggested otherPlayer position. I think @bafolts or @vincentneko should do their review

rzulian:
My preference is to have control of the board and my cards, and if needed to scroll down to see other players. In the current layout I have to scroll everything from boar to my cards, because in between there are the others. But again, is my usability preference.

FreeLikeGNU:
I'd love for blocks to be moveable to a players preference.  Implemented with something like https://mauricius.github.io/vue-draggable-resizable/?path=/story/basic--basic-component-with-form-inside

rzulian:
I'm still thinking that the best solution for the playability is to have on hand the most used sections (i.e the board, the log, your board and cards on hand). @alrusdi : if you don't like it, can we at least think to make it an option?

alrusdi:
> I'm still thinking that the best solution for the playability is to have on hand the most used sections (i.e the board, the log, your board and cards on hand). @alrusdi : if you don't like it, can we at least think to make it an option?

1. I like the part with events cards. If you provide it in separate PR - we will accept it immediately
2. Do you have good idea of how to implement optional block positions without copypasting the blocks?

rzulian:
@alrusdi I've reverted the layout changes. Now there is only the event card for the other players

PR #527, opened 2020-04-01
fix Inventor's Guild can select a card without money

rzulian:


src/cards/InventorsGuild.ts
```diff
@@ -23,8 +23,9 @@ export class InventorsGuild implements IActionCard, IProjectCard {
     }
     public action(player: Player, game: Game) {
         const dealtCard = game.dealer.dealCard();
+        const canSelectCard = (player.megaCredits + (player.canUseHeatAsMegaCredits ?  player.heat : 0))  >=3 ;
         return new SelectCard(
-          "Select card to keep or none to discard",
+          canSelectCard ? "Select card to keep or none to discard" : "You cannot pay this card" ,
```
bafolts, line 28: It _does_ seem like passing in `0` for `maxCardsToSelect` will end up working -- as far as I can tell looking at the code. That is a nice coincidence versus being planned for. The only thing I would change with this is to leverage the `canAfford` function of `Player` which will remove this duplicated logic.

```
const canSelectCard = player.canAfford(player.cardCost);
```

rzulian: My bad, I din't write a proper test for this. I was assuming that `maxCardsToSelect` was working :(
In fact, we have to let the user see the card, even if he cannot afford it.

bafolts: Looking at the source I think they will. We disable the checkbox if if `card.length >= this.maxCardsToSelect` so they won't be able to select it which is what we want.

https://github.com/bafolts/terraforming-mars/blob/master/src/components/SelectCard.ts#L29

Leveraging `canAfford` is more important and being sure to use `player.cardCost` as some corporations modify how much a player needs to spend to purchase a card.

rzulian: Got it! BTW, which is you way of working here? Do I have to have a test in place before to PR?

rzulian: @bafolts You meant 
```
const dealtCard = game.dealer.dealCard();
const canSelectCard = player.canAfford(dealtCard.cost);
```
?

src/cards/InventorsGuild.ts
```diff
@@ -23,8 +23,9 @@ export class InventorsGuild implements IActionCard, IProjectCard {
     }
     public action(player: Player, game: Game) {
         const dealtCard = game.dealer.dealCard();
+        const canSelectCard = player.canAfford(dealtCard.cost);
```
bafolts, line 26: This should be `player.cardCost` and not `dealtCard.cost`.

rzulian: Right, my bad

PR #530, opened 2020-04-01
Toggle background image

FreeLikeGNU:
Toggle the background image from preferences

### DIFF
--- a/src/components/PlayerHome.ts
+++ b/src/components/PlayerHome.ts
@@ -56,7 +56,6 @@ export const PlayerHome = Vue.component("player-home", {
     },
     template: `
         <div id="player-home">
-          <div class="background">
            <h2 :class="'game-title player_color_'+ player.color" v-i18n>Terraforming Mars</h2>
            <h1 :class="'player_bg_color_'+ player.color">{{player.name}}</h1>
             <section>
@@ -211,7 +210,6 @@ export const PlayerHome = Vue.component("player-home", {
                     </div>
                 </div>
             </div>
-           </div>
         </div>
     `
 });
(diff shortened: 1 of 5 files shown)


src/components/PlayerHome.ts
```diff
@@ -56,6 +56,7 @@ export const PlayerHome = Vue.component("player-home", {
     },
     template: `
         <div id="player-home">
+          <div class="background">
```
alrusdi, line 59: Why do we need additional tag her?

src/styles/player_home.less
```diff
@@ -6,6 +6,9 @@
 #player-home {
     padding-left: 20px;
     border-left: 1px solid #adacac;
+}
+
+.background {
```
alrusdi, line 11: This class seems to be not needed

src/styles/preferences.less
```diff
@@ -91,3 +91,9 @@
     margin-left: 38px;
  }
 
+/* Remove background */
+.preferences_remove_background .background {
```
alrusdi, line 95: Why not just:
#player-home.preferences_remove_background
?

FreeLikeGNU: that did not seem to work but:
`.preferences_remove_background #player-home`

src/styles/player_home.less
FreeLikeGNU: requested changes committed, thank you!

src/components/PlayerHome.ts
FreeLikeGNU: requested changes committed, thank you!

PR #532, opened 2020-04-02
Fix not considering player's titanium production

rzulian:
Wasn't considering player's titanium production

### DIFF
--- a/src/cards/AsteroidMiningConsortium.ts
+++ b/src/cards/AsteroidMiningConsortium.ts
@@ -13,8 +13,9 @@ export class AsteroidMiningConsortium implements IProjectCard {
     public name: CardName = CardName.ASTEROID_MINING_CONSORTIUM;
     public canPlay(player: Player, game: Game): boolean {
       if (game.getPlayers().length > 1 
-        && game.getPlayers().filter(p => p.id !== player.id && p.getProduction(Resources.TITANIUM) > 0).length === 0 ) {
-            return false; //No other player to reduce resource from
+        && ( player.getProduction(Resources.TITANIUM)==0 ||
+        game.getPlayers().filter(p => p.id !== player.id && p.getProduction(Resources.TITANIUM) > 0).length === 0 )) {
+            return false; //PLayer hasn't production or No other player to reduce resource from
       }
       return player.getProduction(Resources.TITANIUM) >= 1;
     }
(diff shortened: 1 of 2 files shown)


src/cards/AsteroidMiningConsortium.ts
```diff
@@ -13,8 +13,9 @@ export class AsteroidMiningConsortium implements IProjectCard {
     public name: CardName = CardName.ASTEROID_MINING_CONSORTIUM;
     public canPlay(player: Player, game: Game): boolean {
       if (game.getPlayers().length > 1 
-        && game.getPlayers().filter(p => p.id !== player.id && p.getProduction(Resources.TITANIUM) > 0).length === 0 ) {
-            return false; //No other player to reduce resource from
+        && ( player.getProduction(Resources.TITANIUM)==0 ||
```
alrusdi, line 16: Use === operator please

src/cards/AsteroidMiningConsortium.ts
```diff
@@ -12,11 +12,9 @@ export class AsteroidMiningConsortium implements IProjectCard {
     public cardType: CardType = CardType.AUTOMATED;
     public name: CardName = CardName.ASTEROID_MINING_CONSORTIUM;
     public canPlay(player: Player, game: Game): boolean {
-      if (game.getPlayers().length > 1 
-        && game.getPlayers().filter(p => p.id !== player.id && p.getProduction(Resources.TITANIUM) > 0).length === 0 ) {
-            return false; //No other player to reduce resource from
-      }
-      return player.getProduction(Resources.TITANIUM) >= 1;
+      const othersHaveTitaniumProduction = game.getPlayers().length === 0 ||
```
bafolts, line 15: It a game has *no* players there would definitely be no player to remove titanium production from. I don't see the value in checking for a game with no players. It is a safe assumption to assume the game has players.

bafolts: There is no need to check how many players there are. This `canPlay` could return:

```
return game.getPlayers().find(p => p.id !== player.id && p.getProduction(resources.TITANIUM) > 0) !== undefined || player.getProduction(Resources.TITANIUM) > 0;
```

rzulian: this would be enough `return game.getPlayers().filter(p => p.getProduction(Resources.TITANIUM) > 0).length  >= 1 `, but you need to use `player` in canPlay

Issue #511, opened 2020-03-31
Undo action doesn't work


alrusdi:
How to reproduce:
1) Start 2 player game (both with Beginner corporation)
2) Play **Power plant** stadart project as first player action
3) Do Undo Turn 

As result game crashing with error message like this:
```
/src/database/SQLite.js:28
var gameToRestore = JSON.parse(row.game);
                                               ^
TypeError: Cannot read property 'game' of undefined
```

alrusdi:
@pierrehilbert can you check it?

pierrehilbert:
Yes, we had the same problem yesterday night and I just found the root cause.
I'm patching it right now.

FreeLikeGNU:
Played Research Outpost as first player of generation, placed tile and then used Undo action.  The tile was removed, card and MC/steel returned but I lost my turn, second player of generation then played.

pierrehilbert:
Will take a look!

FreeLikeGNU:
two player, prelude, hellas map, no draft. I think this was second or third gen.

pierrehilbert:
Okay, I found it! The problem was that I saved the game after the research phase and it's not working the same way.
I found another solution, I will make a PR in few minutes.

alrusdi:
Fxd. Thank you!

PR #543, opened 2020-04-03
Saves: Adding a load game page. Resolve Issue #511

pierrehilbert:


server.ts
```diff
@@ -186,6 +190,41 @@ function apiGetGames(req: http.IncomingMessage, res: http.ServerResponse): void
 
 }
 
+function LoadGame(req: http.IncomingMessage, res: http.ServerResponse): void {
```
alrusdi, line 193: Can you please rename it to loadGame for consistency?

pierrehilbert: Sorry I was pretty that it was "CreateGame" before and it's why I did the same, I'm correcting right now!

PR #561, opened 2020-04-05
Show turn order on initial cards selection page

alrusdi:


pierrehilbert:
We already have the information with the link order but yes, could be great to have it on the card selection page.

Issue #562, opened 2020-04-05
Robotic Workforce does not work on Cheung Shing MARS


alrusdi:
Can't copy production box from corporation cards.

alrusdi:
Fxd

PR #564, opened 2020-04-05
Allow to select corporation cards for Robotic Workforce; #562

alrusdi:


src/cards/RoboticWorkforce.ts
```diff
@@ -87,7 +88,16 @@ export class RoboticWorkforce implements IProjectCard {
             CardName.SPACE_PORT
         ];
 
-        const availableCards = player.playedCards.filter((card) => {
+        const corporationCardNames = [
```
bafolts, line 91: Using a `Set` would better represent what this data is being used for. You can put this array in the constructor and then use `Set.prototype.has` instead of `Array.prototype.includes`.

src/cards/RoboticWorkforce.ts
```diff
@@ -170,11 +186,12 @@ export class RoboticWorkforce implements IProjectCard {
                 }
 
                 // Mining resource definition
-                if (foundCard.name === CardName.MINING_AREA || foundCard.name === CardName.MINING_RIGHTS ) {
-                    if (foundCard.bonusResource !== undefined && foundCard.bonusResource === Resources.STEEL) {
+                if (foundCard.name === CardName.MINING_AREA || foundCard.name === CardName.MINING_RIGHTS) {
+                    const bonusResource = (foundCard as any).bonusResource;
```
bafolts, line 190: Having a linting rule to avoid `any` would be really helpful. Instead of casting to `any` you should be able to cast to `IProjectCard` here which defines the `bonusResource` property.

bafolts: Typescript wont be able to determine the type of `bonusResources` since `any` was used. If you use `IProjectCard` with type casting that will allow typescript to know that `bonusResouce` is of type `Resources | undefined`.

alrusdi: @bafolts can you help me with it? It is pretty severe bug. Obvious fix for it is to add Corporation cards to available cards. But typing system doesn't allow to do it... Should be better implementation - i just don't see it

src/cards/RoboticWorkforce.ts
alrusdi: Fixed

PR #573, opened 2020-04-06
Added promo card Rego Plastics (#539)

gomfunkel:
Includes pull requests #569, #570, #571 and #572, please merge those first.

PR #572, opened 2020-04-06
Added promo card Mercurian Alloys (#539)

gomfunkel:
Includes pull requests #569, #570 and #571, please merge those first.

PR #571, opened 2020-04-06
Added promo card Dusk Laser Mining (#539)

gomfunkel:
Includes pull requests #569 and #570, please merge those first.

PR #570, opened 2020-04-06
Added promo card Snow Algae (#539)

gomfunkel:
Includes pull request #569, please merge #569 first.

PR #574, opened 2020-04-06
Added promo card Interplanetary Trade (#539)

gomfunkel:
Includes pull requests #569, #570, #571, #572 and #573, please merge those first.

### DIFF
--- a/src/cards/promo/InterplanetaryTrade.ts
+++ b/src/cards/promo/InterplanetaryTrade.ts
@@ -31,7 +31,10 @@ export class InterplanetaryTrade implements IProjectCard {
                 uniqueTags.add(tags);
             }
         }
-        player.setProduction(Resources.MEGACREDITS, (uniqueTags.size + wildcardCount));
+        // Only count wildcards up to the max amount of tag types existing (minus events and wildcards)
+        const availableTags = uniqueTags.size + wildcardCount;
+        const existingTags = Object.keys(Tags).length - 2;
+        player.setProduction(Resources.MEGACREDITS, Math.min(availableTags, existingTags));
         return undefined;
     }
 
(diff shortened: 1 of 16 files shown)


src/cards/promo/InterplanetaryTrade.ts
```diff
@@ -0,0 +1,42 @@
+import { IProjectCard } from '../IProjectCard';
```
alrusdi, line 1: Double quotes

src/cards/promo/InterplanetaryTrade.ts
```diff
@@ -0,0 +1,42 @@
...
+            player.corporationCard.tags.forEach((tag) => allTags.push(tag));
+        }
+        player.playedCards.filter((card) => card.cardType !== CardType.EVENT).forEach((card) => {
+            card.tags.forEach((tag) => { 
+                allTags.push(tag);
+            });
+        });
+        for (const tags of allTags) {
+            if (tags === Tags.WILDCARD) {
+                wildcardCount++;
```
alrusdi, line 29: I think you should calc all existing tags and increment this only if there exists a unused tags.
I mean, wild tag is not the tag itself - it is a copy of another tag.
Quote form internets:
"The wild tag becomes another tag of your choice on your turn only. So if you already have all other tags, it is not a unique tag"

gomfunkel:
Will be updated with change requests from #570, #571, #572 and #573 before merging.

src/cards/promo/InterplanetaryTrade.ts
gomfunkel: Fixed.

src/cards/promo/InterplanetaryTrade.ts
gomfunkel: Fixed and added additional test for it.

gomfunkel:
@alrusdi Updated the pull request with (hopefully) all change requests from #570, #571, #572, #573 and this one. Further cards will be branched off directly from master for easier handling. Sorry again.

Issue #539, opened 2020-04-03
Add official promo cards


alrusdi:
We are missing these:

- [x] Penguins
- [x] Small Asteroid
- [ ] Self-Replicating robots
- [x] Snow Algae
- [x] Dusk Laser Mining
- [x] Project Inspection
- [x] Energy Market
- [x] Hi-Tech Lab
- [x] Interplanetary Trade
- [x] Law Suit
- [x] Mercurian Alloys
- [x] Orbital Cleanup
- [x] Rego Plastics
- [x] Saturn Surfing
- [x] Stanford Torus

Political Alliance (Turmoil)

vincentneko:
Penguins, Small Asteroid, Self-Replicating robots and Snow Algae missing as well.

https://ssimeonoff.github.io/cards-list 
with promo filter will display all these cards

alrusdi:
> Penguins, Small Asteroid, Self-Replicating robots and Snow Algae missing as well.

Indeed. I found their name in CardName enum, but there no real cards for them :-)

gomfunkel:
@alrusdi @vincentneko If no one started this yet feel free to assign it to me. Should find some time to add the cards soonish.

alrusdi:
@gomfunkel  It would be great. Are you ready to write unit tests for it? :-)

gomfunkel:
@alrusdi Sure thing. No cards without tests. I might even write the tests before implementing the cards... 😉

alrusdi:
You are welcome! Do it one by one to make reviews easy

gomfunkel:
@alrusdi Do you prefer one pull request per card or one commit per card?

alrusdi:
One pull request per card will be awesome!

alrusdi:
@gomfunkel great work! But can you see now why i requested to PR cards one by one? Unfortunately, the card with a bug is on the top of the stack :-(

alrusdi:
@gomfunkel thank you very much for this great work!
I think "Self replicating robots" is too complicated for our version of the game. We can live without it.

gomfunkel:
@alrusdi You're welcome! 

We should create a new issue for _Political Alliance_ which can be implemented once Turmoil is finished or along with the Turmoil cards. 

Regarding _Self-Replicating Robots_: Yes, that's a complicated one. I do have an idea on how to implement the card and will have a try at it. Can't promise anything, though.

PR #579, opened 2020-04-07
Added promo card Orbital Cleanup (#539)

gomfunkel:
This time just one card...


src/cards/promo/OrbitalCleanup.ts
```diff
@@ -0,0 +1,37 @@
+import { IProjectCard } from '../IProjectCard';
```
alrusdi: Double quotes please

src/cards/promo/OrbitalCleanup.ts
```diff
@@ -0,0 +1,37 @@
...
+        return player.getProduction(Resources.MEGACREDITS) >= -3;
+    }
+
+    public play(player: Player) {
+        player.setProduction(Resources.MEGACREDITS, -2);
+        return undefined;
+    }
+
+    public canAct(): boolean {
+        return true;
```
alrusdi, line 25: It's probably better to disallow to activate it if player has no science tags

tests/cards/promo/OrbitalCleanup.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+        const player = new Player("test", Color.BLUE, false);
+        expect(card.canPlay(player)).to.eq(true);
+        expect(card.play(player)).to.eq(undefined);
+        expect(player.getProduction(Resources.MEGACREDITS)).to.eq(-2);
+    });
+    it("Should act", function () {
+        const card = new OrbitalCleanup();
+        const player = new Player("test", Color.BLUE, false);
+        player.playedCards.push(new Research());
+        player.playedCards.push(new AdvancedAlloys());
```
alrusdi, line 27: Can you add here one card with "wildcard" tag to be sure if it is working?

src/cards/promo/OrbitalCleanup.ts
gomfunkel: I think it's better this way. If you are able to use the action without any science tags available you have the chance to use it for a strategic advantage: You can use the action (without doing anything) and then end your turn to wait without needing to pass. I think the rules would allow this as well.

tests/cards/promo/OrbitalCleanup.spec.ts
gomfunkel: Done.

PR #581, opened 2020-04-08
Issue 576 more player details

alrusdi:
Clickable player names on the final screen, and ability to see player's page


src/components/GameEnd.ts
```diff
@@ -89,7 +89,7 @@ export const GameEnd = Vue.component("game-end", {
                         </thead>
                         <tbody>
                             <tr v-for="p in getSortedPlayers()">
-                                <td><span :style="getPlayerColorStyle(p)">{{ p.name }}</span></td>
+                                <td><a :href="'/player?id='+p.id+'&noredirect'" :style="getPlayerColorStyle(p)">{{ p.name }}</a></td>
```
bafolts, line 92: What is `noredirect`? Is this trying to accomplish what `target=blank` would do?

alrusdi: No, by default player's home will be "redirected" to the-end page. This is a parameter to prevent this behavior for the links from final page

PR #598, opened 2020-04-09
Other player details cleanup

FreeLikeGNU:
Clicking anywhere in other player window will close it.

Screenshot_2020-04-09_09-37-13

### DIFF
--- a/src/components/OtherPlayer.ts
+++ b/src/components/OtherPlayer.ts
@@ -26,8 +26,8 @@ export const OtherPlayer = Vue.component("other-player", {
     },
     template: `
         <div> 
-            <div v-show="isVisible()" class="other_player_cont menu" v-on:click="hideMe()">
-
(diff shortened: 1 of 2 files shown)


src/components/OtherPlayer.ts
```diff
@@ -26,13 +26,10 @@ export const OtherPlayer = Vue.component("other-player", {
     },
     template: `
         <div> 
-            <div v-show="isVisible()" class="other_player_cont menu">
-                <button class="btn btn-sm btn-error other_player_close" v-on:click="hideMe()"><i class="icon icon-cross"></i></button>
```
alrusdi, line 30: Can you restore the button as a hint for desktop users? Hide by click anywhere is not intuitive action.

FreeLikeGNU: Close with larger button restored.
Screenshot_2020-04-10_07-52-10

PR #605, opened 2020-04-10
Added promo card Law Suit (#539)

gomfunkel:


src/cards/promo/LawSuit.ts
```diff
@@ -0,0 +1,42 @@
...
+        let possibleLawSuits: Array<Player> = [];
+        for (let playerName of possibleLawSuitsNames) {
+            possibleLawSuits.push(game.getPlayer(playerName));
+        }
+        return new SelectPlayer(possibleLawSuits, "Select player to sue (steal 3 MC from)", (suedPlayer: Player) => {
+            player.setResource(Resources.MEGACREDITS, Math.min(3, suedPlayer.getResource(Resources.MEGACREDITS)));
+            suedPlayer.setResource(Resources.MEGACREDITS, -3, game, player);
+            const cardIndex = player.playedCards.findIndex((element) => element.name === this.name);
+            player.playedCards.splice(cardIndex, 1);
+            suedPlayer.playedCards.push(new LawSuit());
```
alrusdi, line 32: It is better to use "this" here, not "new LawSuit()". Or you have some reasons for it?
It "this" is not available you can use an instance of the card returned by splice()

gomfunkel: You are absolutely right. It already was `this` but then I experimented a bit and forgot to change it back. Thanks for spotting this!

PR #616, opened 2020-04-12
Fix set

vincentneko:
Replace the 2 sets created for law suit by a single array.
I've tested small asteroid in a 2 player game, did not experience any issue.

Also remove Venus option from World Gov if Venus next is not activated

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -80,8 +80,7 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
     public tradesThisTurn: number = 0;
     public colonyTradeOffset: number = 0;
     public colonyTradeDiscount: number = 0;
-    public resourcesRemovedThisGenerationByPlayer: Set<string> = new Set<string>();
-    public productionDecreasedThisGenerationByPlayer: Set<string> = new Set<string>();
+    public removingPlayers: Array<Player> = [];
 
     constructor(
         public name: string,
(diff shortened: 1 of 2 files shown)


src/Player.ts
```diff
@@ -80,8 +80,7 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
     public tradesThisTurn: number = 0;
     public colonyTradeOffset: number = 0;
     public colonyTradeDiscount: number = 0;
-    public resourcesRemovedThisGenerationByPlayer: Set<string> = new Set<string>();
-    public productionDecreasedThisGenerationByPlayer: Set<string> = new Set<string>();
+    public removingPlayers: Array<Player> = [];
```
alrusdi, line 83: This can break Player serialization. Did you check it?

vincentneko: No, not sure on how to do this.
Right now I guess removingPlayers content will be lost in case of restore.

alrusdi: I am checking it...

alrusdi: I didn't find any problem. And it is probably ok to lose removingPlayers info after restore because it is valid only 1 generation. @pierrehilbert can probably clarify it.

vincentneko: I've discussed this with @pierrehilbert . Having an array of players inside player can be messy and uses a lot of information. I will switch to an array on playerIds, it will be easier to restore.

pierrehilbert: The problem was that there is an Array (or a Set) of Player and when you are using the loadFromJSON function, you are not rebuilding the Player object.
You will have some troubles when trying to access to a Player method because of that.
The new version with the array of string will be perfect, we should do it more often.

Issue #624, opened 2020-04-16
In solo game: using WGT to get bonus ocean, shows cards to buy early


thenewjuniorexecutive:
Solo game
Eight oceans out
Used WGT to raise temp from -2 to 0 at end of gen 11
Was given four cards to buy
Was then prompted to place the bonus ocean

The bonus ocean should be placed before anything happens in gen 12

vincentneko:
@bafolts I think one of your commit fixes this, can you confirm ?

bafolts:
Yes, this was closed by #657

PR #632, opened 2020-04-17
Fix WGT for solo mode; #624

alrusdi:


alrusdi:
This is duplicated in another PR

PR #636, opened 2020-04-19
Issue 635 atmoscoop fix

alrusdi:


src/cards/venusNext/Atmoscoop.ts
```diff
@@ -14,7 +14,7 @@ export class Atmoscoop implements IProjectCard {
     public cost: number = 22;
     public tags: Array<Tags> = [Tags.JOVIAN, Tags.SPACE];
     public name: CardName = CardName.ATMOSCOOP;
-    public cardType: CardType = CardType.EVENT;
+    public cardType: CardType = CardType.AUTOMATED;
```
bafolts, line 17: This change is right and if was in it's own pull request could merge. Need to look into the change in `src/Game.ts` as that isn't as clear.

src/Game.ts
```diff
@@ -568,6 +568,17 @@ export class Game implements ILoadable<SerializedGame, Game> {
     }
 
     public doneWorldGovernmentTerraforming() {
+      // Interrupt hook
+      if (this.interrupts.length > 0) {
```
bafolts, line 572: This is mostly duplicated code from:

https://github.com/bafolts/terraforming-mars/blob/master/src/Game.ts#L702

You should create a function that takes a callback for what to do once an action is complete.

```js
private flushAllPlayerInputs(cb: () => void): void {
    if (this.interrupts.length > 0) {
        const interrupt = this.interrupts.shift();
        if (interrupt !== undefined) {
            interrupt.player.setWaitingFor(interrupt.playerInput, cb);
        } else {
            throw new Error("Unexpected empty interrupt in array");
        }
    }
}
```

The `PlayerInterrupt` interface dictates that both `player` and `playerInput` are defined so we shouldn't have to protect against `undefined`.

Then what you have here is:

```js
this.flushAllPlayerInputs(() => this.doneWorldGovernmentTerraforming());
```

alrusdi:
This seems to be complete mess. I will redo it

Issue #633, opened 2020-04-18
Bug: Ecological Zone card enters the game with the wrong amount of resources


Carcassonner:
In today's game, the card entered the game with 4 resources instead of 2.
It may be counting previously played plant/animal tags (after entering Ecological zone, I had 4 of them in total)

alrusdi:
It is not easy reproducible. @Carcassonner can you provide more details? If game has been played on my server then game id will be good enough
Also it may not count wild tags. I don't sure if it is correct behavior.

Carcassonner:
https://tm.alrusdi.ru/game?id=2971682b5706
This effect should not apply to wild tags, and should not be applied to previously played tags.

alrusdi:
Steps to reproduce:
1) Play Viral Enhancers
2) Play Ecological Zone
You will get either  2 plants or 2 animals
Expected to get either 2 plants or **ONE** animal

thenewjuniorexecutive:
Viral Enhancers can also place two animals on Eco Zone

Carcassonner:
By Viral Enhancerss, you get a plant for a plant tag and an animal/plant for an animal tag. Only animals accumulate on the "Ecological zone" card

thenewjuniorexecutive:
Viral Enhancer's text "Effect: When you play a plant, microbe, or an animal tag, including this, gain 1 plant or add 1 resource to THAT CARD"

So you get a resource on the played card or a plant for each trigger of Viral Enhancers. There is no restriction that the resource gained must match the tag played. It's just that this is the only plant tag on a card that holds resources (and there are no animal tags on cards that hold microbes or vice versa.) Once it's added to the "stack", the trigger of VE should not care what type tag it was triggered by, only whether the card holds resources.

This is the best source I can find for this, there's no like from Jacob on this summary, but there was no disagreement from anyone: https://boardgamegeek.com/thread/2091170/article/30459429#30459429

Carcassonner:
Ok, I agree. In this case, the "get 1 animal and 1 plant" option is missing.

bafolts:
What I plan to implement for this is allowing the user to select what bonus they want to apply for each tag. If they have 2 tags they will select what to apply twice. The solution we have in right now where we assume the user wants all the bonus to go to the same place was mostly for efficiency and it was before we had the interrupt system in place which makes this easier to implement.

vincentneko:
Is this issue still relevant ?

bafolts:
[...]

PR #641, opened 2020-04-22
Apply each tag bonus separately for ViralEnhancers

bafolts:
This change applies any bonus for viral enhancers separately.


bafolts:
This is half the work required for #633 , a similar fix will be put in place for EcologicalZone after this is merged.

rzulian:
@alrusdi I'm not sure this is the correct rule. I had submitted a PR for the same bug. I wasn't aware of yours ;)

rzulian:
@alrusdi well, I think the confusion here is because the italian version of this card says "when you play a card with  plant...", which is completely different.

PR #643, opened 2020-04-22
fix one resource only, regardless of the tags

rzulian:
the test was correct but it wasn't testing the edge case


bafolts:
I do not think this fix is right:

https://boardgamegeek.com/thread/1650981/viral-enhancers

> Dirk is right:
When you play a card with plant/microbe/animal tags, you choose:
- put a resource cube on the card you just played (if possible)
- or take a plant resource on your player board

Each tag counts, so if you play Advanced Ecosystems, Viral enhancer would let you take 3 plant resources (one for each tag). But since Adv. Ecosystems is not a card that collects resources you cannot but anything on it.

The thought behind Viral Enhancer was simple enough, but it was not easy to formulate the effect in an easy way!

rzulian:
yes, you are right. The wording in the Italian card is different and the mean is completely different :(

PR #644, opened 2020-04-22
Fix marsuniversity

rzulian:
Mars University doesn't trigger itself, because it doesn't state "including this one"


rzulian:
also this one has been triggered by the italian translation

bafolts:
The italian translation may be wrong, In english the card reads "including this".

rzulian:
yes, you are right. The wording is different and the mean is completely different :(

PR #646, opened 2020-04-30
Translate more parts of the card

bafolts:
There are more players showing up who speak better languages than english. I noticed we don't translate the cards. I have put in this first change to translate direct matching parts of the card. I will put in another check next to efficiently find strings that are close and replace the translations we have for those so that we do not need a direct match.

### DIFF
--- a/src/components/Card.ts
+++ b/src/components/Card.ts
@@ -101,7 +101,7 @@ export const Card = Vue.component("card", {
         }
     },
     template: `
-    <div class="filterDiv" v-html=this.getData()></div>
+    <div class="filterDiv" v-i18n v-html=this.getData()></div>
     `
 });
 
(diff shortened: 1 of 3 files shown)


alrusdi:
I love the idea to be able to translate a cards, but we need more work on it.
Снимок экрана от 2020-04-30 15-21-48

I think we need to add a css class like  "lang-**de**" or "lang-**ru**" to body tag
And then css class like "card-**rego-plastics**" to each card to be able to fine-tune each card appearance for specific language. I mean, to make smaller font in cards titles for Russian language for example.

We need to get rid of Prototype font. It is amazing but it does not support French, German and Russian letters. Ubuntu font is good enough replacement.

bafolts:
My major concern is the text itself and translating it. I leave how it is displayed up to those who can read that language.

PR #647, opened 2020-04-30
Add self replicating Robots

vincentneko:
Last card from Promo set

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -435,7 +435,8 @@ function getPlayer(player: Player, game: Game): string {
     showOtherPlayersVP: game.showOtherPlayersVP,
     actionsThisGeneration: Array.from(player.getActionsThisGeneration()),
     fleetSize: player.fleetSize,
-    tradesThisTurn: player.tradesThisTurn
+    tradesThisTurn: player.tradesThisTurn,
+    selfReplicatingRobotsCardCost: player.getSelfReplicatingRobotsCardCost(game)
   } as PlayerModel;
   return JSON.stringify(output);
 }
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -39,6 +39,7 @@ import {SerializedPlayer} from "./SerializedPlayer";
 import {LogMessageType} from "./LogMessageType";
 import {LogMessageData} from "./LogMessageData";
 import {LogMessageDataType} from "./LogMessageDataType";
+import { SelfReplicatingRobots } from './cards/promo/SelfReplicatingRobots';
 
 export class Player implements ILoadable<SerializedPlayer, Player>{
     public corporationCard: CorporationCard | undefined = undefined;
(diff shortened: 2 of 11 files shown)


vincentneko:
> Uuugh... May be we leave it alone? :-)
> It can be more than one card with robots
> Effect: **Cards** here may be played as if from hand with its cost reduced by the number of resources on it

You're right, I was thinking you could only put a card at a time...

> Uuugh... May be we leave it alone? :-)
> It can be more than one card with robots
> Effect: **Cards** here may be played as if from hand with its cost reduced by the number of resources on it

vincentneko:
I would be happy to have this card in the game, even with a small modification so only one card can be attached to it.
And maybe improve it after to have the exact game board implementation.

bafolts:
At some point it would be helpful to display the card(s) which have been used for the card.

server.ts
```diff
@@ -435,7 +435,8 @@ function getPlayer(player: Player, game: Game): string {
     showOtherPlayersVP: game.showOtherPlayersVP,
     actionsThisGeneration: Array.from(player.getActionsThisGeneration()),
     fleetSize: player.fleetSize,
-    tradesThisTurn: player.tradesThisTurn
+    tradesThisTurn: player.tradesThisTurn,
+    selfReplicatingRobotsCardCost: player.getSelfReplicatingRobotsCardCost(game)
```
bafolts, line 439: You should add `selfReplicatingRobotsCardCost` to `PlayerModel`.

src/Player.ts
```diff
@@ -905,6 +906,18 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
       }
     }
 
+    public getSelfReplicatingRobotsCardCost(game: Game) : number {
+      if (this.playedCards.filter(card => card.name === CardName.SELF_REPLICATING_ROBOTS).length === 1) {
```
bafolts, line 910: It makes more sense to use `find` here rather than `filter`. Filtering twice is unnecessary. The `find` function also stops when it gets a result and doesn't have to traverse the full array.

src/Player.ts
```diff
@@ -1063,6 +1076,18 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
         } else if (preludeCardIndex !== -1) {
           this.preludeCardsInHand.splice(preludeCardIndex, 1);
         }
+
+        // Remove card from Self Replicating Robots
+        if (this.playedCards.filter(card => card.name === CardName.SELF_REPLICATING_ROBOTS).length === 1) {
```
bafolts, line 1081: Same comment regarding using `find`.

src/Player.ts
```diff
@@ -1606,7 +1631,18 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
     }
 
     private getPlayableCards(game: Game): Array<IProjectCard> {
-      return this.cardsInHand.filter((card) => {
+      let candidateCards: Array<IProjectCard> = [...this.cardsInHand];
+      // Self Replicating robots check
+      if (this.playedCards.filter(card => card.name === CardName.SELF_REPLICATING_ROBOTS).length === 1) {
```
bafolts, line 1636: Every place you are doing this check it is simpler to do as follows:

```
const card = this.playedCards.find(card => card.name === CardName.SELF_REPLICATING_ROBOTS);
if (card !== undefined) {

}
```

src/components/SelectHowToPayForCard.ts
```diff
@@ -50,10 +50,8 @@ export const SelectHowToPayForCard = Vue.component("select-how-to-pay-for-card",
                     return icard.calculatedCost
                 }
             }
-            // We should always find a card
-            // It may make more sense to throw
-            // an error here that the card wasn't found
-            return 41; // max card cost
+            // If not found, it is self replication robot stored card
```
bafolts, line 53: fwiw this comment isn't necessarily true, we are _hoping_ that it is self replicating robot.

vincentneko:
> Can you add an option to not include promo cards then?

Deal !

Issue #655, opened 2020-05-01
Bug: Prelude card let MC become negative number


Solo899815:
Before playing prelude card, if a player has no MC in the board, playing prelude card that need to pay MC will cause the MC become negative number. And the only thing you can do in round 1 is pass, even you have enough steel/titanium to play a building / space tag card.

for example: [photo link](https://drive.google.com/file/d/1vgSHJk1809uhHsWAZ-SHvf4_P4BCvPZS/view)

Solo899815:
Forget to say: Many thanks to build up this awesome project !!

vincentneko:
Thanks for the feedback. The link requires an authorization, can you put it somewhere else ?

vincentneko:
I was searching for an answer but did not find one. Imho, if someone has a prelude card with some M€ decrease, if the player does not have enough M€, the card must be discarded.

Solo899815:
> Thanks for the feedback. The link requires an authorization, can you put it somewhere else ?

How about now?

Solo899815:
> I was searching for an answer but did not find one. Imho, if someone has a prelude card with some M€ decrease, if the player does not have enough M€, the card must be discarded.

I retry the solo mode with using "PhoboLog" as corporation. I can keep 10 cards in hand and give a -7MC at first round as a result.  Is it possible to add MC validation before start the game? (I know it need a lot of work)

vincentneko:
Thanks for the feedback. We've just fixed prelude that could lead to negative M€, what you describe is another problem we need to solve !

vincentneko:
I'm closing this one and opening a new one.

Issue #648, opened 2020-04-30
Two Bugs. Hackers and Miranda 


Marhabal:
1) Miranda. Played animal card with prelude. Aridor then added Miranda. Contrary to rules Miranda did not activate for almost the entirety of the game despite there being an animal card in play. Its an odd situation, I know, but I guess there was no check at the end of the turn to activate the colony.

2) Two player game. Played hackers on an opponent with -1 MC. It would not allow me to lower his MC and in fact forced me to lower my own MC :(, in effect creating a net 0 MC prod gain for the card player, in addition to the -1 VP and -1 energy Prod. I guess the card checks for 0.

Marhabal:
Just to clarify because I realize i worded this funny. 1) The animal card with the prelude had resources 2) "Miranda did not activate for almost the entirety of the game" - Miranda only activated when a second animal card with resources was played.

Marhabal:
And thanks for this awesome project !!!

vincentneko:
PR incoming for second issue

vincentneko:
Second issue is fixed.

vincentneko:
Fixed and merged

PR #656, opened 2020-05-02
Fix Aridor colony activation bug

vincentneko:
Issue #648 part 1
And also fix for issue #655

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1943,7 +1943,7 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
       // Prelude cards have to be played first
       if (this.preludeCardsInHand.length > 0) {
         // Remove unplayable prelude cards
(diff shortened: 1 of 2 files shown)


src/Player.ts
```diff
@@ -1942,6 +1942,18 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
  
       // Prelude cards have to be played first
       if (this.preludeCardsInHand.length > 0) {
+        // Remove unplayable prelude cards
+        this.preludeCardsInHand.forEach(card => {
```
bafolts, line 1946: This should use a filter instead of splicing the array we are iterating.

```js
this.preludeCardsInHand = this.preludeCardsInHand.filter(card => card.canPlay!==undefined && card.canPlay(this, game));
```

vincentneko:
> I would use a filter versus modifying an array being iterated by forEach.

Great idea, thanks !

PR #657, opened 2020-05-02
During world government phase take any interrupts prior to next phase

bafolts:
Flushes any pending interrupts which may have happened during a world government phase prior to finishing the world government phase.


vincentneko:
Let's merge this one as well now...

PR #663, opened 2020-05-03
Restore games from server start

vincentneko:


src/database/SQLite.ts
```diff
@@ -16,16 +16,43 @@ export class SQLite implements IDatabase {
...
+            // Should be duplicate column error
+            return;
+          }
+        });  
+    }
+
+    getAllPendingGames(): Array<string> {
+        let allGames: Array<string> = [];
+        let sql = "SELECT distinct game_id game_id FROM games WHERE status = 'running'";
+        this.db.all(sql, [], (err, rows) => {
```
bafolts, line 31: I am assuming the `setTimeout` of 3 seconds was needed because the `allGames` array is being returned before it is populated. The `getAllPendingGames` function will need to take a callback function. The `this.db.all` function is not synchronous. The callback will need to be called inside this function once we have pushed all the rows.

src/database/SQLite.ts
```diff
@@ -16,16 +16,43 @@ export class SQLite implements IDatabase {
...
+            rows.forEach((row) => {
+              allGames.push(row.game_id);
+            });
+        });
+        return allGames;
     }
 
     cleanSaves(game_id: string, save_id: number): void {
-        // DELETE all saves expect last one
+        // DELETE all saves exept last one
```
bafolts, line 44: s/exept/except

server.ts
```diff
@@ -213,6 +213,26 @@ function loadGame(req: http.IncomingMessage, res: http.ServerResponse): void {
   });
 }
 
+function loadAllGames(): void {
+  let allGames = Database.getInstance().getAllPendingGames();
```
bafolts, line 217: This will need to become:

```
Database.getInstance().getAllPendingGames(function (err, allGames) {

})
```

src/database/SQLite.ts
```diff
@@ -16,37 +16,61 @@ export class SQLite implements IDatabase {
...
 
-            return true;
+            return cb(err);
         });
+    }    
+
+
+    cleanSaves(game_id: string, save_id: number): void {
+        // DELETE all saves except last one
+        this.db.run("DELETE FROM games WHERE game_id = ? AND save_id < ?", [game_id, save_id], function(err: { message: any; }) {
```
bafolts, line 58: If these queries need to run one after the other you should chain them in the callbacks.

```
this.db.run("DELETE FROM games WHERE ...", function (err) {
   if (err) {
      return;
   }
   this.db.run("UPDATE games SET status = ...."
}):
```

vincentneko: Yes it's better to run the delete before the update, will modify it.

vincentneko: It's not working, I have an error message from my IDE :
`Property 'db' does not exist on type 'RunResult'.ts(2339)`

It should no be a problem anyway because the database should queue the 2 concurrent statements (the delete should put a lock preventing the update to execute before the delete is finished)

PR #698, opened 2020-05-12
Improved for optional played cards and others player board

rzulian:
I've tested this version which seems to be a good compromise.
The others players board are always shown, but you can hide/show played cards, which can speed up the middle-end game phase.
I've moved the log on to, right after the awards and milestones and above the turn order and above the player's board.

Screenshot 2020-05-12 at 15 15 36

Screenshot 2020-05-12 at 15 15 46

### DIFF
--- a/src/components/OtherPlayer.ts
+++ b/src/components/OtherPlayer.ts
@@ -22,12 +22,18 @@ export const OtherPlayer = Vue.component("other-player", {
         },
         isVisible: function () {
             return (this.$root as any).getVisibilityState("other_player_"+this.player.id);
+        },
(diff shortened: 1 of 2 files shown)


rzulian:
@alrusdi Any chance you can review this?

vincentneko:
I've just checked out this change.
I must admit I'm not very fond of it, not being able to click on players name and having to scroll all the way down feels awkward.
As this behavior is not optable out, I would rather not merge it into master.

alrusdi:
@rzulian Sorry, i am not into this feature as you remember.

PR #723, opened 2020-05-27
Finish translation.

glass4pk:
Finish translation by compare to ru. but still find so many sport needs to be translate.
How  to add more translations into it?
Can I add more .json file to this folder, copy the strings and translate?

like:

image

image


src/Game.ts
```diff
@@ -212,8 +212,9 @@ export class Game implements ILoadable<SerializedGame, Game> {
       for (const player of players) {
         if (!player.beginner) {
           // Failsafe for exceding corporation pool - Minimum is 12
-          if (this.startingCorporations * this.players.length > 12) {
-            this.startingCorporations = 2;
+          // Minimum increase to 25 by ChainLee
+          if (this.startingCorporations * this.players.length > 25) {
```
bafolts, line 216: The translations in this file are fine. This seems unrelated to the translations change. Can we do this in a separate pull request if at all. What is this doing?

vincentneko: This looks like a local tweak to be able to see all corporations cards at once.

bafolts: I will remove this tweak and merge. The rest looks fine.

bafolts: Actually this is on the master branch of a fork. I will wait for the fork owner to update.

alrusdi: @glass4pk wants to merge into non master branch. I believe it is pretty save to approve the PR and then edit it in our repository to correct that code

Issue #726, opened 2020-05-28
Suggestion: Minor changes for player actions list


nwai90:
1. If there is only one claimable milestone, its position in the list of player actions is not fixed due to only the milestone's name being displayed without the action's title.

**Suggestion:** Always display the same title for consistency even if there is only one milestone - it will be selected by default when the "Claim Milestone" action is clicked

<img width="593" alt="Screenshot 2020-05-28 at 1 44 13 PM" src="https://user-images.githubusercontent.com/2408094/83108918-1222f780-a0f3-11ea-9d25-05be9dadf68e.png">

<img width="488" alt="Screenshot 2020-05-28 at 2 15 11 PM" src="https://user-images.githubusercontent.com/2408094/83108930-18b16f00-a0f3-11ea-890c-9a2010a6ee6d.png">

2. "Pass" is sometimes made the default selected option under certain circumstances. Related to https://github.com/bafolts/terraforming-mars/issues/510

**Suggestion:** Always fix the terminal positions of "End Turn", "Pass" and "Undo Turn" actions, in that order.

**Rationale:** "End Turn" is usually preferable to "Pass" (e.g. for tactical stalling), and "Undo Turn" should be selected only as a last resort in the event of a misclick.

<img width="489" alt="Screenshot 2020-05-28 at 2 39 42 PM" src="https://user-images.githubusercontent.com/2408094/83109362-db99ac80-a0f3-11ea-9e41-3047cb2c219d.png">

nwai90:
@bafolts @vincentneko Thoughts? I'm happy to submit a PR for this

vincentneko:
This is looking great !

nwai90:
Ok, will submit a PR for it soon. The above suggested ordering is based on my own experience while playing, wondering if others have similar/other thoughts about it

PR #730, opened 2020-05-28
Sort actions

nwai90:
**Scope:**
- Fix milestones' positions in player actions list and ensure consistent title
- Fix terminal sequence of "End Turn", "Pass" and "Undo Turn" in player actions list

**Ref:** https://github.com/bafolts/terraforming-mars/issues/726

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -2020,19 +2020,9 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
         );
       }
 
-      action.options.push(
-          this.passOption(game)
-      );
-
       if (this.cardsInHand.length > 0) {
         action.options.push(
             this.sellPatents(game)
-        );
-      }
-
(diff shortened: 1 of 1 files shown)


src/Player.ts
```diff
@@ -2145,12 +2142,19 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
...
       action.options.sort((a, b) => {
-        if (a.title > b.title) {
-          return 1;
-        } else if (a.title < b.title) {
-          return -1;
-        }
-        return 0;
+        if (a.title === b.title) return 0;
+
+        if (a.title === "Undo Turn") return 1;
```
bafolts, line 2147: Instead of putting this into the sort I would add the undo, pass, and end turn options after calling to sort the other options . After sorting options by title the undo, pass, and end turn can be added by calling `action.options.unshift`.

nwai90: 👍 Will update

PR #731, opened 2020-05-28
Fix crashes if certain corps are selected without required expansions

nwai90:
**Scope:**
- Handle edge cases causing game to crash by not triggering corp's first action if the required expansion is not selected
- Affected corps: Aridor, Celestic and MSI 
- Don't increment turn action counter if corp could not take its first action - player should still be able to do two actions on turn 1 in this case

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1991,10 +1991,9 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
             player: this,
             playerInput: input
           });
-
-          this.actionsTakenThisRound++; // only increment if corp can take its corp first action
         }
         this.actionsThisGeneration.add("CORPORATION_INITIAL_ACTION");
+        this.actionsTakenThisRound++;
         this.takeAction(game);
         return;
       }
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1991,9 +1991,10 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
             player: this,
             playerInput: input
           });
+
+          this.actionsTakenThisRound++; // only increment if corp can take its corp first action
         }
         this.actionsThisGeneration.add("CORPORATION_INITIAL_ACTION");
-        this.actionsTakenThisRound++;
         this.takeAction(game);
         return;
       }
--- a/src/cards/venusNext/Celestic.ts
+++ b/src/cards/venusNext/Celestic.ts
@@ -16,7 +16,7 @@ export class Celestic implements IActionCard, CorporationCard, IResourceCard {
     public resourceCount: number = 0;
 
     public initialAction(player: Player, game: Game) {
-        if (game.hasCardsWithResource(ResourceType.FLOATER)) {
+        if (game.hasCardsWithResource(ResourceType.FLOATER, 2)) {
             for (let foundCard of game.drawCardsByResource(ResourceType.FLOATER, 2)) {
                 player.cardsInHand.push(foundCard);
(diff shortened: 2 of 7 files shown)


src/Player.ts
```diff
@@ -1991,9 +1991,10 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
             player: this,
             playerInput: input
           });
+
+          this.actionsTakenThisRound++; // only increment if corp can take its corp first action
         }
         this.actionsThisGeneration.add("CORPORATION_INITIAL_ACTION");
-        this.actionsTakenThisRound++;
```
bafolts, line 1996: This will not work for every corporation. For example, the `Splice` corporation returns `undefined` but takes an action without an interrupt.

https://github.com/bafolts/terraforming-mars/blob/master/src/cards/promo/Splice.ts#L16

If for some reason a player selects a corporation where the initial action ends up being a noop I would still count that as an action taken this round.

src/cards/venusNext/Celestic.ts
```diff
@@ -16,9 +16,14 @@ export class Celestic implements IActionCard, CorporationCard, IResourceCard {
     public resourceCount: number = 0;
 
     public initialAction(player: Player, game: Game) {
-        for (let foundCard of game.drawCardsByResource(ResourceType.FLOATER, 2)) {
-            player.cardsInHand.push(foundCard);
+        if (game.venusNextExtension || game.coloniesExtension) {
```
bafolts, line 19: This class is the only caller for `drawCardsByResource`. The check for `venusNextExtension` or `coloniesExtension` being `true` is being used assuming that if either of those is `true` that there will be cards with floater resources. We should iterate through the deck to see if there are any floater cards instead.

src/cards/venusNext/MorningStarInc.ts
```diff
@@ -11,9 +11,14 @@ export class MorningStarInc implements CorporationCard {
     public startingMegaCredits: number = 50;
 
     public initialAction(player: Player, game: Game) {
-        for (let foundCard of game.drawCardsByTag(Tags.VENUS, 3)) {
-            player.cardsInHand.push(foundCard);
+        if (game.venusNextExtension || game.coloniesExtension) {
+            for (let foundCard of game.drawCardsByTag(Tags.VENUS, 3)) {
```
bafolts, line 15: Similar to my comment above I would iterate the deck to see if there are any venus cards. You could make a function on game to `hasVenusCards`.

src/Player.ts
nwai90: 👌 I see, will revert this part

src/cards/venusNext/MorningStarInc.ts
vincentneko: Can you make this more generic, like `hasTagCards(tag: Tags)` ?
This could be used for EXPERIMENTAL FOREST or ACQUIRED SPACE AGENCY, maybe someday we will implement custom project card selection.

Same with `hasResourceCards(resourceType: ResourceType)` ?

vincentneko: After a second thought and always with custom project card selection in mind, it might be better to have the exact count of a given tag or resource card, and then test if we have enough to allow the draw.

nwai90: Good idea, will update so we can always ensure that there are sufficient eligible cards to be drawn

PR #732, opened 2020-05-28
Fix english typos

bradwerth:
This makes a number of changes to both visible text and to symbols to fix English language typos.


bradwerth:
Added a fixup for the issue blocking the build.

PR #735, opened 2020-05-29
Improvement: Log revealed cards and discarded cards

nwai90:
**Context:**
- Add missing logs for several corporations' first actions
- For actions that explicitly reveal cards, the cards drawn into hand should be made known to other players
- Also removes Inventrix's unnecessary first action click in https://github.com/bafolts/terraforming-mars/issues/749 and logs Prelude revealed draws

Screenshot 2020-06-01 at 3 26 34 PM

Screenshot 2020-06-01 at 3 54 17 PM

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -54,6 +54,7 @@ import { CardName } from "./CardName";
 import { Turmoil } from "./turmoil/Turmoil";
 import { PartyName } from "./turmoil/parties/PartyName";
 import { IParty } from "./turmoil/parties/IParty";
+import { ColonyName } from "./colonies/ColonyName"
 
 export interface GameOptions {
   draftVariant: boolean;
@@ -1345,6 +1346,58 @@ export class Game implements ILoadable<SerializedGame, Game> {
 
     public hasCardsWithResource(resource: ResourceType, requiredQuantity: number = 1) {
       return this.dealer.deck.filter((card) => card.resourceType === resource).length >= requiredQuantity;
+    }
+    
+    public logCorpFirstAction(player: Player, colonyName?: ColonyName) {
+      if (player.corporationCard !== undefined) {
+        let message = "";
+
+        switch (player.corporationCard.name) {
+          case CardName.INVENTRIX:
+            message = "drew 3 cards"
+            break;
+
+          case CardName.THARSIS_REPUBLIC:
+            message = "placed a City tile"
+            break;
+
+          case CardName.CELESTIC:
+            const floaterCards = player.cardsInHand.filter((card) => card.resourceType === ResourceType.FLOATER).slice(-2)
+            message = "drew 2 floater cards: " + floaterCards.map((card) => card.name).join(", ")
+            break;
+
+          case CardName.MORNING_STAR_INC:
+            const venusCards = player.cardsInHand.filter((card) => card.tags.includes(Tags.VENUS)).slice(-3)
+            message = "drew 3 Venus cards: " + venusCards.map((card) => card.name).join(", ")
+            break;
+
+          case CardName.ARIDOR:
+            message = "added a new Colony tile: " + colonyName
+            break;
+
+          case CardName.PHILARES:
+            message = "placed a Greenery tile"
+            break;
+
+          case CardName.ARCADIAN_COMMUNITIES:
+            message = "placed a Community (player marker)"
+            break;
+
+          case CardName.SPLICE:
+            const drawnCard = player.cardsInHand.filter((card) => card.tags.includes(Tags.MICROBES)).slice(-1)[0]
+            message = "drew a Microbe card: " + drawnCard.name
+            break;
+        
+          default:
+            break;
+        }
+
+        this.log(
+          LogMessageType.DEFAULT,
+          "${0} " + message,
+          new LogMessageData(LogMessageDataType.PLAYER, player.id)
+        );
+      }
     }
 
     private setupSolo() {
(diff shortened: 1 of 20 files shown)


nwai90:
After pushing a second commit to make the card names clickable, the names are always blank even though I can see them in debug logs, and the empty spaces are clickable. Am I doing something incorrectly? 🤔

Screenshot 2020-05-29 at 8 31 45 PM

nwai90:
@vincentneko @bafolts Could you look through this? I fixed the issue in third commit; it was caused by the cards not being played or in hand yet at that point in time. Sample screenshot of change:

Screenshot 2020-05-31 at 8 22 11 PM

src/Game.ts
```diff
@@ -1346,6 +1355,76 @@ export class Game implements ILoadable<SerializedGame, Game> {
     public hasCardsWithResource(resource: ResourceType, requiredQuantity: number = 1) {
       return this.dealer.deck.filter((card) => card.resourceType === resource).length >= requiredQuantity;
     }
+    
+    public logCorpFirstAction(player: Player, colonyName?: ColonyName) {
+      if (player.corporationCard !== undefined) {
+        let message = "", drawnCards = undefined;
+        const corpName = player.corporationCard.name;
+
+        if (corpName === CardName.CELESTIC) {
```
bafolts, line 1364: This logic should be in the cards themselves and not the game class. This sets an unsustainable precedent where we have logic or separate cards in the game class.

nwai90: I updated

vincentneko:
@bafolts can you please review / approve this PR ? Thanks

src/Game.ts
```diff
@@ -1306,6 +1306,11 @@ export class Game implements ILoadable<SerializedGame, Game> {
           result.push(projectCard);
         } else {
           this.dealer.discard(projectCard);
+          this.log(
+            LogMessageType.DEFAULT,
+           "${0} was discarded",
+           new LogMessageData(LogMessageDataType.CARD, projectCard.name)
```
vincentneko, line 1312: I'm not sure there is some value to display discarded cards here.

src/Game.ts
```diff
@@ -1321,10 +1326,23 @@ export class Game implements ILoadable<SerializedGame, Game> {
...
```
vincentneko, line 1332: I'm not sure there is some value to display discarded cards here.

src/cards/colonies/Poseidon.ts
```diff
@@ -23,6 +26,12 @@ export class Poseidon implements CorporationCard {
               colony.name + " - (" + colony.description + ")", 
               () => {
                   colony.onColonyPlaced(player, game);
+                  game.log(
+                    LogMessageType.DEFAULT,
```
vincentneko, line 30: I think we can rework how the game log colony placement.
Only one call to `game.log` in `Colony.addColony(colony: IColony, player: Player, game: Game): void` should be necessary.
I can take care of this in a separate PR.

src/Game.ts
nwai90: I believe there is some value here, as discarded cards from "reveal and draw" effects are meant to be public information. IMO, knowing that specific cards have been discarded (and thus are no longer in the draw pile) could help players to adjust strategy.

src/cards/colonies/Poseidon.ts
nwai90: @vincentneko Thanks for the feedback, I'll push this change here

src/Game.ts
vincentneko: In the "real" game, cards need to be seen so the card effect can be resolved. The card revelation is a side effect I guess, not something sought after for strategic purpose. What I foresee is log pollution and extra time taken by players who want to take advantage of this information.

nwai90: This is a good point, let me see if I can do something about it. Some possibilities I could think of:

1) Control this using a flag (game option), disabled by default
2) Collapse each set of public discards into a single log, passing all discarded cards as a string (though this will not be clickable to bring up the card)

nwai90: @vincentneko What do you think about this approach instead? I consolidated it into a single line, which players can click on to view the list of discarded cards. This way we can still provide the information without cluttering the log

Screenshot 2020-06-10 at 11 12 54 AM

vincentneko: This is great !

jijiqw:
image
What if  I  already have  some  floater cards in my hand , the  log  should  show  exactly  the  cards revealed  in  initial action , not  from my hand

jijiqw:
image

This change  may be  better , I think . 
By the way ,  I was about to add these " revealed cards log"  and "first  action  log "  myself  when   saw  you pull  request , this greatly saves my time .
   Thanks for your work !

nwai90:
> image
> What if I already have some floater cards in my hand , the log should show exactly the cards revealed in initial action , not from my hand

Yes, it should show the cards revealed in the initial action. The drawn cards are appended to hand, so here we log the names of the last two floater cards added, e.g.:

Screenshot 2020-06-12 at 12 13 06 AM

jijiqw:
> > image
> > What if I already have some floater cards in my hand , the log should show exactly the cards revealed in initial action , not from my hand
> 
> Yes, it should show the cards revealed in the initial action. The drawn cards are appended to hand, so here we log the names of the last two floater cards added, e.g.:
> 
> Screenshot 2020-06-12 at 12 13 06 AM

I get it , Thanks

Issue #423, opened 2020-03-19
Issue with market manipulation


insectman:
Currently market manipulation allows to move colony track back and forth on the same colony tile although it's explicitly stated in the card text that that card action should be applied to two different colonies.

alrusdi:
@vincentneko can you look at this?

vincentneko:
This one will be a pain to modify. Right now the 2 actions are not correlated.
If we want to implement the card rules exactly, we need to pass the first option results to the second.

I don't see this as a potential exploit to the game so this issue will get a low priority to me.

vincentneko:
Fixed by @nwai90 
Thanks

PR #748, opened 2020-05-31
Suggestion: Fix minor issue for Market Manipulation

nwai90:
**Ref:** https://github.com/bafolts/terraforming-mars/issues/423

Screenshot 2020-05-31 at 4 28 33 PM

This fix involves minimal code change, so we don't have to rewrite and pass additional arguments around


nwai90:
@vincentneko We could consider this as an interim solution to prevent duplicates with minimal code change

nwai90:
> I think the card now needs a canPlay function so we are sure there is a colony to increase and another to decrease. Without it and with all colonies maxed out or at minimum, it might struck the game.

Good catch, will make this change as well

nwai90:
@vincentneko I updated, could you have a look?

Issue #749, opened 2020-05-31
General wish list, Interface ideas and perhaps some bugs


PhantomFFR:
Foreword: I have yet to play anything different than “Solo Challenge” with Prelude, promo cards and Random Board, so all my comments are in respect to it. I play using Firefox on Windows 10, without  touch input, but with a wide screen. I also have almost no understanding of programming. But after playing Solo Challenge about 50 times in the past week, here are some repeated thought of mine. I hope it’s alright to have them all in one post rather than split up into individual posts.

-----

Interface:

Related to #360: Though I’m not concerned about the sorting of the actions, but rather the default selection of Actions. As of now “standard projects” is defaulted to, even if you have playable project cards, that would rarely be the correct choice. 
Personally I feel the default selection order should be:
0 “Buy Milestone” (Though this is more guesswork due to only playing the physical board game with other people)
1 “Converting heat” (As you likely have to think less about than when “converting plants”, though perhaps this is a good safeguard (for card requirements) to have “converting plants” first after all.)
2 “Converting plants”
3 “Perform an action from a played card” or “Play a patent”
4 “Play a patent” or “Perform an action from a played card”
5 “Standard Project”
6 “Fund Award” (Perhaps this could be ahead of “Standard Project”, but again, I have yet to play this implementation other than Solo)
7 “Pass”
8 “Sell Patents”

Related to #726 and #510, or perhaps even contrarian to #510: 
Once you are out of funds/actions, “sell patents” should not be the default action, you’d barely want to do that at all, whereas passing will be what you mostly do. It being default is made a bit more annoying by the fact, that when you then select “Pass” as your action, the “Save” button moves up the screen and away from your cursor. 
Perhaps (after reading through #510), my issue is actually more about the “Save button” moving away from the cursor, so remedying this would be swell.

-----

Wish list:

It would be nice if the layout was taking landscape screen formats more into account (e.g. having the game board to the right of the “Actions” or the “cards in hand” to the right of the game board (and “Actions” below the game board).

Perhaps making the allowed spaces on the board where to may place a tile even more obvious would be a good idea, I was confused for two generations why I can’t place my greenery tile, totally forgetting about my Mohole area.

Placement bonus for metal is very similar to the background colour of the map and can be hard to actually notice.

It seems that “Select 3 cards” at the beginning of the game if playing Inventrix does not need to be a click. 

[...]

PR #756, opened 2020-06-02
Improvement: Payment UI automatically uses other available resources

nwai90:
**Ref:** https://github.com/bafolts/terraforming-mars/issues/749

### Context
If a player cannot pay for a card using only MC, other valid and available resources will be used for payment in the following order:

1. Microbes from Psychrophiles for Plant cards (first priority since they can be removed by Ants)
2. Floaters from Dirigibles
3. Steel ("greedy" implementation, use as much as possible without overpaying by default)
4. Titanium ("greedy" implementation, use as much as possible without overpaying by default)
5. Heat (last priority since it can be used for temperature raise)

Screenshot 2020-06-02 at 11 06 48 AM

Screenshot 2020-06-02 at 12 29 27 PM

Screenshot 2020-06-02 at 9 49 54 PM

Screenshot 2020-06-02 at 10 29 03 PM

### DIFF
--- a/src/components/SelectHowToPay.ts
+++ b/src/components/SelectHowToPay.ts
@@ -34,30 +34,19 @@ export const SelectHowToPay = Vue.component("select-how-to-pay", {
         app.$data.cost = app.playerinput.amount;
         app.$data.megaCredits = (app as any).getMegaCreditsMax();
 
-        app.setDefaultHeatValue();
         app.setDefaultSteelValue();
         app.setDefaultTitaniumValue();
+        app.setDefaultHeatValue();
       });
     },
     methods: {
         hasWarning: function () {
             return this.$data.warning !== undefined;
         },
-        setDefaultHeatValue: function() {
-          // automatically use available heat for Helion if not enough MC
-          if (!this.canAffordWithMcOnly() && this.canUseHeat()) {
-              this.$data.heat =  this.$data.cost - this.player.megaCredits;
-          } else {
-              this.$data.heat = 0;
-          }
-
-          let discountedCost = this.$data.cost - this.$data.heat;
-          this.$data.megaCredits = Math.max(discountedCost, 0);
-        },
         setDefaultSteelValue: function() {
           // automatically use available steel to pay if not enough MC
           if (!this.canAffordWithMcOnly() && this.canUseSteel()) {
-              let requiredSteelQty = Math.ceil(Math.max(this.$data.cost - this.player.megaCredits - this.$data.heat, 0) / this.player.steelValue);
+              let requiredSteelQty = Math.ceil(Math.max(this.$data.cost - this.player.megaCredits, 0) / this.player.steelValue);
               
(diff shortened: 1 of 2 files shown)


bafolts:
I will check this out tomorrow. From a quick pass through it looks good. As this can be a critical component of the game I will give it more of a read through prior to approving. Thanks for this!

vincentneko:
This feature is great !

One might argue that default usage of resources could be nonoptimal if you have cards like Security Fleet or Space Elevator, etc...

It is the case in your last image of examples.
A better allocation I think would be to use one more titanium and one less steel so you can use the card ability. (this is not a request to enable this automatically !) 

Maybe some players like to allocate resource manually also (not my case)

So if you could add this behavior as a game option like "Smaller cards" or "Magnify cards on hover" it would be better of two.

I may ask too much, if it's the case I'm ok to merge this in master as it is.

nwai90:
> This feature is great !
> 
> One might argue that default usage of resources could be nonoptimal if you have cards like Security Fleet or Space Elevator, etc...
> 
> It is the case in your last image of examples.
> A better allocation I think would be to use one more titanium and one less steel so you can use the card ability. (this is not a request to enable this automatically !)
> 
> Maybe some players like to allocate resource manually also (not my case)
> 
> So if you could add this behavior as a game option like "Smaller cards" or "Magnify cards on hover" it would be better of two.
> 
> I may ask too much, if it's the case I'm ok to merge this in master as it is.

@vincentneko Yes, resources can always be adjusted manually after the default allocation. As it stands, the current change will save quite a number of clicks for the user.

I could make the auto-allocation a game option that can be disabled. Not too sure if this will be a useful option to have though.

I do like the idea of auto-reserving 1 steel for Space Elevator and 1 titanium for Security Fleet if possible, but it may be better to tackle these special scenarios in a separate PR as this is a fairly large change in terms of scope.

vincentneko:
From my perspective, I do not like the idea of auto-reserving 1 steel for Space Elevator and 1 titanium for Security Fleet. This should be left to players decision according to me. As you mentioned, if you want to go this way it should be done in a separate PR.

Once again, I really like the change (I suggested it some time ago). 

About the usefulness of the option, maybe we can ask other people feedback. @alrusdi any thought about this ?

alrusdi:
I think this is a great feature and it does not need to be optional.
To take into count special cases (like Security Fleet or Space Elevator) seems to be too complicated. It is ok to do it by hand.

vincentneko:
> I think this is a great feature and it does not need to be optional.
> To take into count special cases (like Security Fleet or Space Elevator) seems to be too complicated. It is ok to do it by hand.

Fair enough, let's wait for @bafolts review / approval and let's merge this in master.

vincentneko:
> > I think this is a great feature and it does not need to be optional.
> > To take into count special cases (like Security Fleet or Space Elevator) seems to be too complicated. It is ok to do it by hand.
> 
> Fair enough, let's wait for @bafolts review / approval and merge this in master.

PR #758, opened 2020-06-03
Cache db directory for heroku

bafolts:
We added a local database on disk for storing games that are being played for the undo operation. When the app starts we attempt to repopulate games from that database. The idea was on our heroku server we would load any games which were ongoing. Reading through the heroku documentation today I believe we have to define our `db` directory as a `cacheDirectories` in our top level `package.json` to keep the database file around.

Eventually this database will grow too large to be stored on disk and we will need to consider pruning it in some way. For the time being this should allow games to be played while we update `master` without losing the games.

https://devcenter.heroku.com/articles/nodejs-support#custom-caching

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,5 +48,6 @@
   "bugs": {
     "url": "https://github.com/bafolts/terraforming-mars/issues"
   },
-  "homepage": "https://github.com/bafolts/terraforming-mars#readme"
+  "homepage": "https://github.com/bafolts/terraforming-mars#readme",
+  "cacheDirectories": ["db"]
 }


Solo899815:
Is there any coding method to store the record as "[game id].db" files , instead of just a file "game.db" ?

bafolts:
This would be a coding method to store the database file under any name we desired. Originally the database file was only being unused for the undo operation and I figured we could use it for restarting the game server. Trying to maintain the file on disk is futile and will require more work than going with a separate database server. What I plan to finish soon is propping up a free heroku postgres database server that will be used when the app is running in _production_. When running the game locally the SQLite file can be used. This will still allow the game to run locally and it will allow me to better manage the game when it runs on heroku without needing access to the disk on the device. Heroku allows me to grant other users access to the postgres instance as well so others could assist with managing the database without needing disk access to the machine where the game server is running. I can't imagine we would ever get enough users, but this would also help with scaling.

Issue #684, opened 2020-05-06
Venus bar appears over Turmoil board in Chrome on Macbook


thenewjuniorexecutive:
With both Venus Next & Turmoil selected in Chrome  81.0.4044.138 with Sierra 10.12.6 (16G2136) on a 1920 x 1080 display, the Venus bar obscures the Turmoil board.
![Screen Shot 2020-05-06 at 11 41 52 AM](https://user-images.githubusercontent.com/62528913/81201425-417e9280-8f93-11ea-8a3c-8cd87d1e6ec4.png)

Marhabal:
I noticed if you "restore down" the window size slightly the turmoil board will slide below. The venus board only covers it when you have window maximized.

vincentneko:
@alrusdi I think this one is fixed with new global design ?

alrusdi:
yes it is

PR #763, opened 2020-06-04
Global params revamp; #684

alrusdi:
YA attempt to make compact global parameters
More translations for colonies
Highlight hint for SelectSpace action
Снимок экрана от 2020-06-04 22-15-32
Снимок экрана от 2020-06-04 22-16-35


src/components/BoardMixin.ts
```diff
@@ -18,7 +18,7 @@ export const BoardMixin = {
             return boardSpaces;
         },
         getMainSpaces: function (): Array<SpaceModel> {
-            return (this as any).getAllSpaces().filter((s: SpaceModel) => {return parseInt(s.id) < 70})
+            return (this as any).getAllSpaces().filter((s: SpaceModel) => {return s.spaceType != "colony"})
```
bafolts, line 21: We should use `SpaceType.COLONY` instead of this string.

funkybaboon:
@alrusdi The free ocean at 0°C is missing on the new map.

alrusdi:
@funkybaboon yes, i've fixed it already. Waiting for PR approovement

ssimeonoff:
@alrusdi , молодец!

Best UI for the global params. (first time I saw it [here](https://scontent-arn2-1.xx.fbcdn.net/v/t1.15752-9/66753914_315487199169615_5355324404177829888_n.jpg?_nc_cat=104&_nc_sid=b96e70&_nc_ohc=bQjAccJyLbEAX-asq1n&_nc_ht=scontent-arn2-1.xx&oh=7a260546ca75048632ea8451c8e550c6&oe=5F1ABC4C) )

nwai90:
@alrusdi Suggestion: Why not use a different colour instead of applying opacity? Attached an example below, I feel it's quite clear that oxygen is at 0%, Venus is at 2%, and temperature is at -26 degrees.

Screenshot 2020-06-25 at 8 19 03 PM

alrusdi:
@nwai90 but problem is how to distinquish a **-26** and **-24** values, not how to read exact value. Prev variant was more readable.

nwai90:
@alrusdi I see. Was thinking in relation to https://github.com/bafolts/terraforming-mars/issues/777, that full opacity might be helpful in this direction.

ssimeonoff:
Maybe add colour and text-shadow to the active ones, something like: 
Untitled

alrusdi:
@nwai90 look at it! It can be great addition to the frame around current value

nwai90:
@ssimeonoff Thanks for illustration! I'll update and post

ssimeonoff:
wouldn't it be better to keep the inactive ones with the opacity 0.4 as they were?

nwai90:
> wouldn't it be better to keep the inactive ones with the opacity 0.4 as they were?

It's subjective; 0.4 makes it quite hard to see IMO. If not full opacity, perhaps a higher value closer to 1 would improve readability (e.g. 0.8)

PR #774, opened 2020-06-06
Add created_time to games table

vincentneko:
Add a created_time that default to now() to the table. It will be useful for housekeeping.
If the postgres database is already up and running it will not do anything, the table could be modified from the command line. I can take care of this if needed.

### DIFF
--- a/src/database/PostgreSQL.ts
+++ b/src/database/PostgreSQL.ts
@@ -15,7 +15,7 @@ export class PostgreSQL implements IDatabase {
             }
         });
         this.client.connect();
-        this.client.query("CREATE TABLE IF NOT EXISTS games(game_id varchar, save_id integer, game text, status text default 'running', PRIMARY KEY (game_id, save_id))", (err) => {
+        this.client.query("CREATE TABLE IF NOT EXISTS games(game_id varchar, save_id integer, game text, status text default 'running', created_time timestamp default now(), PRIMARY KEY (game_id, save_id))", (err) => {
             if (err) {
                 throw err;
             }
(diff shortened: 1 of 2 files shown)


bafolts:
This is smart. I will put this live at some point soon. Hopefully this stabilizes some of the longer running games while we push out new releases. The free database I believe has 10k rows and a storage limit of some kind. Assuming not that many people are playing I can prune older games on some background interval while the game server is running. If a lot of people are playing I will have to upgrade to the larger postgresql that heroku provides.

bafolts:
OK. I restarted the server on heroku to use postgresql. We should now get some idea of how many games are being played.

bafolts:
There are 459 unique games on the postgres server and 4,000 rows used. The free postgres instance used only supports 10,000 rows. We will reach this quota soon. I manually added this column to be able to prune games that have an older creation time. Hobby basic supports 10,000,000 rows. I will update to this limit. That should cover us for a lot of games and allow us to prune games that are a few days old.

bafolts:
I upgraded us from heroku-dev to heroku-basic so we now have a 10,000,000 row limit which I am assuming gives us enough runway to delete games which are a week old and advertise that we will delete games that take longer than a week to play from `README.md`. I will wait a week to see how many rows we end up with.

vincentneko:
Around 500 games is an impressive number. It's great because it means the game engine is somehow stable, we have very few bug reports.
Can we make a `games_history `table to move records from `games `?
I would like to be able to do some statistics sometimes. If we used `jsonb` datatype, it would allow queries to parse json data directly.
If you give me access to the database I can take care of this.

PR #785, opened 2020-06-09
Added timestamps to log entries

gomfunkel:
Changed the bullet in front of log level entries to a clock face which shows the timestamp for the log message on hover.

timestamps

### DIFF
--- a/src/components/LogPanel.ts
+++ b/src/components/LogPanel.ts
@@ -74,9 +74,10 @@ export const LogPanel = Vue.component("log-panel", {
             return '';
         },
         parseMessage: function(message: LogMessage) {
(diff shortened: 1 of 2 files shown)


nwai90:
Looks nice, Daniel! On the subject of log display, I'm wondering if there is a reason we limit it to show only the last 50 lines?

vincentneko:
> Looks nice, Daniel! On the subject of log display, I'm wondering if there is a reason we limit it to show only the last 50 lines?

It's an arbitrary number I've set. Except for resource consumption (memory) there is no special reason to this value.

nwai90:
> > Looks nice, Daniel! On the subject of log display, I'm wondering if there is a reason we limit it to show only the last 50 lines?
> 
> It's an arbitrary number I've set. Except for resource consumption (memory) there is no special reason to this value.

@vincentneko I see, thanks! It might be interesting to be able to export the entire log at game end, allowing a played game to be reproduced accurately on a physical board using the log's data.

PR #790, opened 2020-06-09
[WIP] Game option to randomize milestones and awards

nwai90:


src/Game.ts
```diff
@@ -1452,11 +1480,10 @@ export class Game implements ILoadable<SerializedGame, Game> {
         return player.loadFromJSON(element);
       });
 
-
       // Rebuild milestones, awards and board elements
-      this.milestones = [];
+      this.milestones = []; // TODO: Store and fetch M&A data on rebuild
       this.awards = [];
-      this.board = this.boardConstructor(d.boardName);
+      this.board = this.boardConstructor(d.boardName, false);
```
nwai90, line 1486: @vincentneko Here's where we'd need to fetch stored milestones and awards for the game if `randomMA = true`, else if can be constructed normally if this option is not enabled

vincentneko: I will create a specific branch for it.
randomize_milestones_awards
I took a quick glimpse of the changes, I did not see any specifics for Venus Next award and milestone (hoverlord / venuphile)

nwai90: Ah yes, I didn't add those two so they will be included by default in every game that has Venus expansions selected. Do you think we should include them in the randomization too?

vincentneko: Yes, I think when Venus Next is enabled, these 2 should be added to the respective pools.

PR #791, opened 2020-06-09
Show players as active while drafting

gomfunkel:
This shows all players still drafting as being active in the turn order section. This way it is easy to spot who finished drafting and who still needs to do so.

The activity state is not refreshed automatically for everyone but needs a page refresh, similar to the logging of who bought how many cards. If anyone has a good idea on how to refresh all components for all players, please let me know (or simply go for it).


gomfunkel:
The test doesn't fail for me locally and I can't see a reason why it should. Can you please have a look at that, @vincentneko?

vincentneko:
> The test doesn't fail for me locally and I can't see a reason why it should. Can you please have a look at that, @vincentneko?

This is strange, it happens to me as well. Re-launching the checks fixed the issue.
@bafolts Any idea about this ?

alrusdi:
@gomfunkel can you please merge the master branch? Then i will try to do a review

Issue #792, opened 2020-06-10
Bug: Celestic should draw any card that has floater icon(s) anywhere


nwai90:
Currently, cards like Air Scrapping Expedition, Aerosport Tournament and Stratospheric Birds are being discarded by Celestic

**Ref:**
https://boardgamegeek.com/thread/1960449/question-about-celestic-and-wg-phase
https://boardgamegeek.com/thread/1870830/celestic-corporation-initial-card-draw-and-stratos

nwai90:
I will submit a PR to fix this issue

PR #793, opened 2020-06-10
Fix: Celestic should be able to draw any card with floater icon(s)

nwai90:
Ref: https://github.com/bafolts/terraforming-mars/issues/792

### DIFF
--- a/src/cards/venusNext/Celestic.ts
+++ b/src/cards/venusNext/Celestic.ts
@@ -16,25 +16,16 @@ export class Celestic implements IActionCard, CorporationCard, IResourceCard {
     public resourceCount: number = 0;
 
     private static readonly floaterCards: Set<CardName> = new Set<CardName>([
-        CardName.AERIAL_MAPPERS,
         CardName.AEROSPORT_TOURNAMENT,
         CardName.AIR_SCRAPPING_EXPEDITION,
         CardName.AIR_RAID,
         CardName.AIRLINERS,
         CardName.ATMOSCOOP,
-        CardName.DEUTERIUM_EXPORT,
-        CardName.DIRIGIBLES,
-        CardName.EXTRACTOR_BALLOONS,
         CardName.FLOATER_LEASING,
         CardName.FLOATER_PROTOTYPES,
         CardName.FLOATER_TECHNOLOGY,
-        CardName.FLOATING_HABS,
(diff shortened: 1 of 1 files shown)


nwai90:
> I'm afraid this one will not work. `resourceType` main purpose is to indicate that the card is able to hold tokens for this type of resource. Adding this property to other cards may have some side effects, like appearing in list where they could receive a resource for instance.
> 
> I think a simple approach is to create an enum of floater related cards (other cards than resourceType Floater) and add them to the Celestic test.

I see, will make the change to go with this approach then

vincentneko:
I think the list is missing 8 cards.
I've used https://ssimeonoff.github.io/cards-list with floater filter to get the list.

Maybe the enum should only list cards that do not have resourceType as floater and the test could be something like :

if (Celestic.floaterCards.has(card.name) || card.resourceType === ResourceType.FLOATER) {

nwai90:
> I think the list is missing 8 cards.
> I've used https://ssimeonoff.github.io/cards-list with floater filter to get the list.
> 
> Maybe the enum should only list cards that do not have resourceType as floater and the test could be something like :
> 
> if (Celestic.floaterCards.has(card.name) || card.resourceType === ResourceType.FLOATER) {

Will update. Where's the floater filter though? Could not find it on the page

vincentneko:
> > I think the list is missing 8 cards.
> > I've used https://ssimeonoff.github.io/cards-list with floater filter to get the list.
> > Maybe the enum should only list cards that do not have resourceType as floater and the test could be something like :
> > if (Celestic.floaterCards.has(card.name) || card.resourceType === ResourceType.FLOATER) {
> 
> Will update. Where's the floater filter though? Could not find it on the page

Just type floater in the "Search by text..." field =)

nwai90:
> > > I think the list is missing 8 cards.
> > > I've used https://ssimeonoff.github.io/cards-list with floater filter to get the list.
> > > Maybe the enum should only list cards that do not have resourceType as floater and the test could be something like :
> > > if (Celestic.floaterCards.has(card.name) || card.resourceType === ResourceType.FLOATER) {
> > 
> > 
> > Will update. Where's the floater filter though? Could not find it on the page
> 
> Just type floater in the "Search by text..." field =)

All good, done!

PR #800, opened 2020-06-12
Also randomize Venus milestone and award if option selected

nwai90:
@vincentneko I updated to randomize Venus milestone and award as well, if the option is selected

Screenshot 2020-06-12 at 3 07 32 PM


vincentneko:
@nwai90 I've also pushed a change to include Venus MA before yours.
Can you please check there are no things remaining from my commit ?

nwai90:
@vincentneko Checked, LGTM

PR #802, opened 2020-06-12
Improvement: Swap action order to save clicks

nwai90:
**Context:** If a player is able to spend resources on cards added previously, he will usually want to do that in most cases, barring certain exceptions like Hoverlord award and Excentric milestone.

**Scope:**
- Make "Remove resource to do X" option the default action if available
- Skip confirmation click if played card has only 1 choice of action available

Screenshot 2020-06-12 at 11 17 49 PM

### DIFF
--- a/src/cards/colonies/RedSpotObservatory.ts
+++ b/src/cards/colonies/RedSpotObservatory.ts
@@ -43,10 +43,9 @@ export class RedSpotObservatory implements IProjectCard, IResourceCard {
             return undefined;
         });
 
-        opts.push(addResource);
-
         if (this.resourceCount > 0 ) {
             opts.push(spendResource);
+            opts.push(addResource);
         } else {
             return addResource;
         }
--- a/src/cards/colonies/RedSpotObservatory.ts
+++ b/src/cards/colonies/RedSpotObservatory.ts
@@ -36,12 +36,8 @@ export class RedSpotObservatory implements IProjectCard, IResourceCard {
         const addResource = new SelectOption("Add 1 floater on this card", () => this.addResource());
         const spendResource = new SelectOption("Remove 1 floater on this card to draw a card", () => this.spendResource(player, game));
 
-        if (this.resourceCount > 0 ) {
-            opts.push(spendResource);
-            opts.push(addResource);
-        } else {
-            this.addResource();
-        }
+        opts.push(spendResource);
+        opts.push(addResource);
 
         return new OrOptions(...opts);
     }
--- a/src/cards/colonies/RedSpotObservatory.ts
+++ b/src/cards/colonies/RedSpotObservatory.ts
@@ -32,25 +32,29 @@ export class RedSpotObservatory implements IProjectCard, IResourceCard {
         }
 
         var opts: Array<SelectOption> = [];
-        const addResource = new SelectOption("Add 1 floater on this card", () => {
-            this.resourceCount++;
-            return undefined;
-        });
 
-        const spendResource = new SelectOption("Remove 1 floater on this card to draw a card", () => {
-            this.resourceCount--;
-            player.cardsInHand.push(game.dealer.dealCard());
-            return undefined;
-        });
+        const addResource = new SelectOption("Add 1 floater on this card", () => this.addResource());
+        const spendResource = new SelectOption("Remove 1 floater on this card to draw a card", () => this.spendResource(player, game));
 
         if (this.resourceCount > 0 ) {
             opts.push(spendResource);
             opts.push(addResource);
         } else {
-            return addResource;
+            this.addResource();
         }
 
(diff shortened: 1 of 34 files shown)


FrankJones314:
Yeah, I don't know whether this would be too much work, but it would be nice to have an OPTION (key word is "option") to have it automatically deduct two microbes to raise the temperature or oxygen or whatever.  

This is tricky though, because, when a bonus is up for grabs, it's not uncommon to add a 3rd, and then do the "remove 2" next gen, with the eventual result being able to still do the same number of terraforms, but delaying the first one by one gen without delaying the second one.  

But then, if I forget that I have this  setting toggled on, which is likely, I'll forget to switch it off temporarily.   

Similar thoughts - the steam game always bugged me (Hah!) by always making me select which card I'm removing a microbe from for my ants card.  If I only have tardigrades, it's always going to be tardigrades.   I always  thought the game would benefit from a "same as last time" checkbox that one can just leave checked until he wishes or needs to unchecked it.  So, it would automatically eat one of my tardigrades when I activate ants, but then if my opponent plays decomposers, I can uncheck the default tardigrades before I do the action again.   

So, maybe being to select a default target or action is good.  Same with Symbiotic Fungus, I'm always adding it to "ants" if all i have is ants and decomposers.  (Trickier if I have one of the cards like regolith eaters of GHG-producing bacteria).   

Any card that adds an animal - is there any reason not to just default add this to any card that is 1 VP per animal?  Who cares whether it is on birds or predators or fish or penguins or livestock.

nwai90:
> Yeah, I don't know whether this would be too much work, but it would be nice to have an OPTION (key word is "option") to have it automatically deduct two microbes to raise the temperature or oxygen or whatever.
> 
> This is tricky though, because, when a bonus is up for grabs, it's not uncommon to add a 3rd, and then do the "remove 2" next gen, with the eventual result being able to still do the same number of terraforms, but delaying the first one by one gen without delaying the second one.
> 
> But then, if I forget that I have this setting toggled on, which is likely, I'll forget to switch it off temporarily.
> 
> Similar thoughts - the steam game always bugged me (Hah!) by always making me select which card I'm removing a microbe from for my ants card. If I only have tardigrades, it's always going to be tardigrades. I always thought the game would benefit from a "same as last time" checkbox that one can just leave checked until he wishes or needs to unchecked it. So, it would automatically eat one of my tardigrades when I activate ants, but then if my opponent plays decomposers, I can uncheck the default tardigrades before I do the action again.
> 
> So, maybe being to select a default target or action is good. Same with Symbiotic Fungus, I'm always adding it to "ants" if all i have is ants and decomposers. (Trickier if I have one of the cards like regolith eaters of GHG-producing bacteria).
> 
> Any card that adds an animal - is there any reason not to just default add this to any card that is 1 VP per animal? Who cares whether it is on birds or predators or fish or penguins or livestock.

Yes, Hoverlord award and Excentric milestone are exceptions which will require the player to select "Add a resource to this card". IMO in most other cases, it's probably better to have the default as "Remove resources" whenever this option is available.

Ants - this is addressed in https://github.com/bafolts/terraforming-mars/pull/803 to make it similar to Predators (auto select target if there is only one valid choice)

Symbiotic Fungus - I can look into default selection when the player selects "Add 2 microbes to another card" if there is only one valid target for adding microbes.

Animals - I agree that we should automatically add animals if there is only one valid target card. If there are multiple possible targets, it might be a bit more subjective. One example: If I have Pets and my opponent has Viron + Predators, I may prefer to add animals to the protected Pets card instead, even if I have other 1 VP/animal cards available.

src/cards/colonies/RedSpotObservatory.ts
```diff
@@ -32,28 +32,31 @@ export class RedSpotObservatory implements IProjectCard, IResourceCard {
...
-            this.resourceCount--;
-            player.cardsInHand.push(game.dealer.dealCard());
-            return undefined;
-        });
 
-        opts.push(addResource);
+        const addResource = new SelectOption("Add 1 floater on this card", () => this.addResource());
+        const spendResource = new SelectOption("Remove 1 floater on this card to draw a card", () => this.spendResource(player, game));
 
         if (this.resourceCount > 0 ) {
```
vincentneko, line 39: I think this if will always be true as the value is tested before.

src/cards/colonies/RedSpotObservatory.ts
```diff
@@ -32,28 +32,31 @@ export class RedSpotObservatory implements IProjectCard, IResourceCard {
...
             opts.push(spendResource);
+            opts.push(addResource);
         } else {
-            return addResource;
+            this.addResource();
         }
```
vincentneko, line 44: I think this if is unnecessary as the value is already tested above and it will be always true.

Issue #787, opened 2020-06-09
Feature Request: Random Milestones and Awards


nwai90:
Ref: https://ssimeonoff.github.io/ma-randomizer

We could implement this as a game option without taking synergy into account first, and add synergy considerations at a later stage.

<img width="874" alt="Screenshot 2020-06-09 at 9 43 30 PM" src="https://user-images.githubusercontent.com/2408094/84155282-c0874f00-aa9a-11ea-83f5-5f6871f6a77b.png">

<img width="804" alt="Screenshot 2020-06-09 at 9 43 57 PM" src="https://user-images.githubusercontent.com/2408094/84155288-c2e9a900-aa9a-11ea-915d-78e10d32b845.png">

nwai90:
@vincentneko Could you help with the database side for this? I can push a draft PR first; we would have to store milestones and awards data for games that have randomized M&A, in order for them to be properly rebuilt.

vincentneko:
> @vincentneko Could you help with the database side for this? I can push a draft PR first; we would have to store milestones and awards data for games that have randomized M&A, in order for them to be properly rebuilt.

Yes I can help with the database side.

nwai90:
Thanks, I raised a draft PR https://github.com/bafolts/terraforming-mars/pull/790

vincentneko:
@nwai90  I've pushed a commit on the branch that should handle game reload and clone with random  MA. Please review and test.

nwai90:
> @nwai90 I've pushed a commit on the branch that should handle game reload and clone with random MA. Please review and test.

@vincentneko Tested, works perfect 👍I added a minor improvement here: https://github.com/bafolts/terraforming-mars/pull/806

vincentneko:
Great, I think you can submit a PR. Now we need someone to review !

nwai90:
Closing as changes are merged

PR #809, opened 2020-06-13
Add game option to randomize milestones and awards

nwai90:
**Scope:**
- Add option to randomize milestones and awards (default: OFF)
- Without Venus Next: Randomize all 5 milestones and awards
- With Venus Next: Randomize all 6 milestones and awards
- Milestones and awards set for each game will persist across reloads and predefined cloned games. Thanks @vincentneko for help!

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -377,7 +377,8 @@ function createGame(req: http.IncomingMessage, res: http.ServerResponse): void {
         startingCorporations: gameReq.startingCorporations,
         soloTR: gameReq.soloTR,
         clonedGamedId: gameReq.clonedGamedId,
-        initialDraftVariant: gameReq.initialDraft
+        initialDraftVariant: gameReq.initialDraft,
+        randomMA: gameReq.randomMA
       } as GameOptions;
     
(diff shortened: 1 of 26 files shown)


ap3h3ad:
Hi,

maybe this should contain some logic to ensure that there aren't too many milestones and awards that benefit each other (and make it easier to get several of them with the same tactics), like Benefactor and Terraformer or Gardener and Cultivator or Builder and Contractor.

Maybe the logic of ssimeonoffs MA Generator (https://ssimeonoff.github.io/ma-randomizer) could be used here.

nwai90:
> Hi,
> 
> maybe this should contain some logic to ensure that there aren't too many milestones and awards that benefit each other (and make it easier to get several of them with the same tactics), like Benefactor and Terraformer or Gardener and Cultivator or Builder and Contractor.
> 
> Maybe the logic of ssimeonoffs MA Generator (https://ssimeonoff.github.io/ma-randomizer) could be used here.
> 
> Or does it already contain such restrictions?

Agree. Synergy will be done separately in a different PR as mentioned here: https://github.com/bafolts/terraforming-mars/issues/787

So that incremental changes can be isolated

Issue #801, opened 2020-06-12
Sixth player


AlexBurkov:
I know original game has 5 players max, but there are 6 players from us so we need possibility to play six of it. Can you add one extra player?

nwai90:
@bafolts @vincentneko Is this something you are okay to support? I can submit a PR to allow up to 6 players per game

vincentneko:
> @bafolts @vincentneko Is this something you are okay to support? I can submit a PR to allow up to 6 players per game

Yes I don't see why we shouldn't

vincentneko:
Merged

PR #810, opened 2020-06-13
Add support for sixth player

nwai90:
**Ref:** https://github.com/bafolts/terraforming-mars/issues/801

Seems to work fine when tested locally. Issues like having too few corporations in the base game to deal out 3 to each of the 6 players have already been previously handled.

6p_1

6p_2

Screenshot 2020-06-14 at 4 32 13 AM

Screenshot 2020-06-14 at 4 28 31 AM

Screenshot 2020-06-14 at 4 28 09 AM


AlexBurkov:
When I added sixth player here I had to add images for cube and ship for colony. Can you check it in your PR? In other cases it was ok because we played it a lot of times.

nwai90:
> When I added sixth player here I had to add images for cube and ship for colony. Can you check it in your PR? In other cases it was ok because we played with six players a lot of times. And I used this color #800080. I think it's more light color:)
> image

Added the new assets!

PR #826, opened 2020-06-16
 fixbug:Tactician can be claimed with fewer than 5 cards with require…

jijiqw:
fixbug:Tactician can be claimed with fewer than 5 cards with requirements in play

### DIFF
--- a/src/cards/colonies/FloaterLeasing.ts
+++ b/src/cards/colonies/FloaterLeasing.ts
@@ -1,17 +1,17 @@
 import { IProjectCard } from "../IProjectCard";
 import { Tags } from "../Tags";
-import { CardType } from '../CardType';
+import { CardType } from "../CardType";
 import { Player } from "../../Player";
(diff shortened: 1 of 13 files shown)


src/cards/colonies/FloaterLeasing.ts
```diff
@@ -1,17 +1,17 @@
...
 export class FloaterLeasing implements IProjectCard {
     public cost: number = 3;
     public tags: Array<Tags> = [];
     public name: CardName = CardName.FLOATER_LEASING;
     public cardType: CardType = CardType.AUTOMATED;
-
+    public hasRequirements = false;
     public canPlay(player: Player): boolean {
         return player.getResourceCount(ResourceType.FLOATER) >= 3;
     }
```
vincentneko, line 17: I think all these line are to be removed.
The card can always be played.

jijiqw: updated now  ¦-D

Issue #845, opened 2020-06-19
Feature Request: Add promo corp Pharmacy Union


nwai90:
<img width="734" alt="Screenshot 2020-06-20 at 5 51 23 AM" src="https://user-images.githubusercontent.com/2408094/85182542-99f7be00-b2bb-11ea-94d3-79077b674441.png">

**Effect:** When ANY microbe tag is played, including these, add a disease here and lose 4 MC or as much as possible. When you play a science tag, remove a disease from this card and raise your TR 1 step OR if there are no diseases here, you may raise your TR 3 steps and TURN THIS CARD FACE DOWN (lose this corporation's tags and effects).

nwai90:
Closed as done

PR #846, opened 2020-06-20
Add Pharmacy Union promo corp

nwai90:
**Ref:** https://github.com/bafolts/terraforming-mars/issues/845

Screenshot 2020-06-20 at 6 33 06 PM


src/HTML_data.ts
```diff
@@ -6793,6 +6793,29 @@ export const HTML_DATA: Map<string, string> =
...
       </div>
 `],
+[CardName.PHARMACY_UNION,`
+      <div class="tag tag1 tag-microbe"></div>
+      <div class="tag tag2 tag-microbe"></div>
+      <div class="corporationLabel">CORPORATION</div>
+      ##RESOURCES##
+      <div class="venus-icon corporation-icon"></div>
+      <div class="contentCorporation">
+        <div class="corporationEffectBox" style="height:150px;margin-top:130px;">
```
alrusdi, line 6803: Can you please do not use inline styles?
We have a lot of legacy here, but for new cards it is better to avoid it

nwai90: Sure, I updated to use classes

src/cards/promo/PharmacyUnion.ts
```diff
@@ -0,0 +1,82 @@
...
+                );
+                return undefined;
+            } else {
+                const availableOptions: OrOptions = new OrOptions();
+
+                availableOptions.options.push(
+                    new SelectOption('Turn this card face down and gain 3 TR', () => {
+                        this.tags = [];
+                        player.increaseTerraformRatingSteps(3, game);
+                        this.onCardPlayed = () => {};
```
vincentneko, line 66: I think we need a boolean to store if the corporation has been disabled.
Using the undo will reset the gain 3 TR option and reinstate the tags, we need to reset the tags and the onCardPlayed based on the boolean value.

nwai90: I updated; the tags of disabled corporation will also be excluded from `player.getTagCount()`

src/HTML_data.ts
```diff
@@ -6800,19 +6800,19 @@ export const HTML_DATA: Map<string, string> =
       ##RESOURCES##
       <div class="venus-icon corporation-icon"></div>
       <div class="contentCorporation">
-        <div class="corporationEffectBox" style="height:150px;margin-top:130px;">
-          <div class="corporationEffectBoxLabel" style="margin-bottom:5px">EFFECT</div>
+        <div class="corporationEffectBox pharmacy_union-corp-effect-box">
```
alrusdi, line 6803: You don't need to apply specific css classes for each block.
Each card has a parent class like card-pharmacy-union.
So it is perfectly correct to just remove inline styles from html and then apply them in css filest like this:

`.card-pharmacy-union .corporationEffectBox {height:150px;margin-top:130px;}`

Do you want me to make an example of it on this card? Or it is crear and you can fix it?

nwai90: @alrusdi Thanks! I'll give it a shot

nwai90: Could you make the example? Can't seem to get it working right

alrusdi: Ok, lets's merge it. Then i will make new markup and you will take it as s reference for upcoming cards

Issue #849, opened 2020-06-20
Bug: Immigrant City not playable at -4 MC production (or -5 for Tharsis)


nwai90:
Official ruling at https://boardgamegeek.com/thread/1655663/article/31621425#31621425.

The card should be playable at
(A) -4 MC production for all players, or
(B) -5 MC production for Tharsis

nwai90:
@vincentneko @bafolts I drafted a fix for (A) at https://github.com/nwai90/terraforming-mars/commit/aa697c9310ae103d525024d4d6c1f0ed8ebfcc97.

Do you have suggestions on how we could best handle (B)? The commented out test currently fails due to Tharsis' effect of raising MC production after city tile placement.

nwai90:
Closing as should be resolved in https://github.com/bafolts/terraforming-mars/pull/861

jijiqw:
![image](https://user-images.githubusercontent.com/16096558/86467620-47f97280-bd68-11ea-9d82-8fe83b25c335.png)
Can't figure it out when the  property mcProductionAfterDecrease  would be reduced to  -7 or  lower ,  I would greatly appreciate it if you can  helping me out of my confusion

nwai90:
Tharsis playing it at -5 MC prod reduces it to -7. Increased back to -5 by its own effect and by Tharsis's corp effect

jijiqw:
I have tried serverl time in local ,  if Tharsis playing  at -5 MC ,  the  "onTilePlaced" func would be called  in  "game.addCityTile" , 
so the  mcProduction would be  increased to -3  and then  reduced to  -5 by  player.setProduction after "game.addCityTile",

nwai90:
👌 I'll check again

PR #860, opened 2020-06-23
Fix: Playability of Immigrant City

nwai90:
**Ref:** https://github.com/bafolts/terraforming-mars/issues/849

**Scope:**
A) Anyone can play -4 MC production, and end up with -5 MC production after playing.
B) Tharsis can play at -5 MC production, and remain at -5 MC production after playing.


nwai90:
Closed as found a bug

PR #861, opened 2020-06-23
Fix edge cases for Immigrant City

nwai90:
Ref: https://github.com/bafolts/terraforming-mars/issues/849


nwai90:
@alrusdi That was fast! Was about to leave a comment to ask for feedback haha. I think we could use `player.shouldTriggerCardEffect` for any future edge cases that we might encounter as well

PR #868, opened 2020-06-24
Add logs to expansion cards

nwai90:
**Scope:**
- Add logs to remaining cards
- Extend `logGainPlants` to `logGainStandardResource`

### DIFF
--- a/src/cards/ElectroCatapult.ts
+++ b/src/cards/ElectroCatapult.ts
@@ -60,13 +60,11 @@ export class ElectroCatapult implements IActionCard, IProjectCard {
     }
     
     private log(game: Game, player: Player, resource: Resources) {
-      const displayStr = resource === Resources.PLANTS ? resource.slice(0, -1) : resource;
-
       game.log(
         LogMessageType.DEFAULT,
         "${0} spent 1 ${1} to gain 7 MC",
         new LogMessageData(LogMessageDataType.PLAYER, player.id),
-        new LogMessageData(LogMessageDataType.STRING, displayStr)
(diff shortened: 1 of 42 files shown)


src/cards/ElectroCatapult.ts
```diff
@@ -60,12 +59,14 @@ export class ElectroCatapult implements IActionCard, IProjectCard {
       return 1;
     }
     
-    private log(game: Game, player: Player, resource: string) {
+    private log(game: Game, player: Player, resource: Resources) {
+      const displayStr = resource === Resources.PLANTS ? resource.slice(0, -1) : resource;
+
```
vincentneko, line 64: I don't understand why this line is required, can you please explain ?

nwai90: The current code causes `"steel"` to be displayed as `"stee"` without the `"l"` as the last character was removed in all cases. This line is to display `Resources.PLANTS` as `"plant"`, removing the `"s"`. Minor quibble haha

vincentneko: Thanks for your answer. I overlock a string was also an array !
I'm pretty sure this will not for for some other languages than english, may I suggest we use resource icon here instead of text ?

If you prefer we handle this in another issue, I can finish review here, just let me know.

nwai90: @vincentneko I will remove this line. Displaying as "1 plants" instead of "1 plant" is not a show stopper, we can handle it properly in another issue to address localisations as well

vincentneko: Great, I will create an issue to suggest using icons for log.

Issue #777, opened 2020-06-06
Suggestion: Clearly indicate end of game to players


nwai90:
Some ideas:
- Show a temporary message (e.g. "Mars is terraformed!") to players
- Style all global parameter values to a different colour when Mars is terraformed

FrankJones314:
For solo mode this would also be helpful (similar to the steam version; one of the few things they did correctly).  I had two games that I "lost" (I counted them as wins) because I had literally 175 MC and spent them all on cities and greeneries not realizing I needed one more temperature bump.  

Having the portion of the corresponding semicircle around the board for each of the three parameters completely change color when it is completed might be an idea also.

nwai90:
Handled in https://github.com/bafolts/terraforming-mars/pull/893. Possibly improve alert UI / use animation in future

PR #874, opened 2020-06-25
Improve visibility of global parameter current levels

nwai90:
**Ref:** https://github.com/bafolts/terraforming-mars/issues/777

**Suggestion:** Use a different color plus text-shadow instead of opacity, so that it's clearer what the current level for each parameter is.

Screenshot 2020-06-26 at 4 26 40 PM

### DIFF
--- a/src/styles/globs.less
+++ b/src/styles/globs.less
@@ -13,6 +13,7 @@
 
 	.val-is-active {
 		color: gold;
+		text-shadow: 2px 3px 2px #474747, 2px 2px 2px black;
 	}
 
 	.global-numbers-oceans {
--- a/src/styles/globs.less
+++ b/src/styles/globs.less
@@ -8,17 +8,17 @@
 	
 
 	.global-numbers-value {
-		filter: opacity(.4);
+		filter: opacity(1.0);
 	}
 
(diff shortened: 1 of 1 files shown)


alrusdi:
I don't like these changes. They make values more readable, but it is much harder to see where the current level of parameter is.
It is easier to understand in black and white.
new_colors

Better solution can be to make param block little longer to make more space for digits and remove background. For active (already reached) values there can be bright background with very digits on it.
more_readable
Unfortunately i don't have much time to make working prototype of it. But i will try ASAP

nwai90:
I think it's best to keep the original backgrounds which are more visual, instead of using black and white. Would be great to highlight the current parameter level in a way that it clearly stands out from all other values

alrusdi:
Yes! Some kind of frame around current value also can be a solution
85719857-6d5ffe00-b722-11ea-8810-2e10bfead7f2

alrusdi:
@nwai90 should i accept these changes or you will add a frame around current value too?

nwai90:
@alrusdi Sure let's add the frame later if this is not sufficient? I see @ssimeonoff's suggestion of shadow has helped a lot

PR #883, opened 2020-06-29
Add Diversity Support promo card (?)

nwai90:
Screenshot 2020-06-29 at 9 08 32 AM
Screenshot 2020-06-29 at 11 11 14 AM

**Note:** This has at least two possibilities:
- Total of 9 resources collected on cards
- 9 different cards with resources on them

It is not entirely clear at the moment which one is the actual card effect

### DIFF
--- a/src/CardName.ts
+++ b/src/CardName.ts
@@ -385,6 +385,7 @@ export enum CardName {
     MAGNETIC_SHIELD = "Magnetic Shield",
     MELTWORKS = "Meltworks",
     MOHOLE_LAKE = "Mohole Lake",
+    DIVERSITY_SUPPORT = "Diversity Support",
     INDUSTRIAL_CENTER = "Industrial Center",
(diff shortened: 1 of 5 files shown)


vincentneko:
I'm pretty sure that the 6 basic resources (MC, steel, plants...) are included here, if not reaching 9 seems almost impossible (ResourceType enum has 9 elements)
So my guess is having at least 9 different resources (on cards or basic resources)

nwai90:
> I'm pretty sure that the 6 basic resources (MC, steel, plants...) are included here, if not reaching 9 seems almost impossible (ResourceType enum has 9 elements)
> So my guess is having at least 9 different resources (on cards or basic resources)

Yes, that's a possibility as well. There are many ways this could go, will try to get more clarification first on whether it is number of different types of resources, number of cards with resources, or total number of resources

PR #887, opened 2020-06-29
Optimize json and clone game API

vincentneko:
WARNING : this change will make all pending games incompatible with reload. The database must be purged before server restart (drop table games)

Only @bafolts should merge this in master.

This one is big...

Add a players column to the games table. This will avoid clone game API to parse a lot of JSON to extract players count.
Also add an index that will speed up clone game API query.

Use player ids instead on player objects in turmoil, colonies and main game class.

Average 3 player games, with all expansions : JSON save file is 14 701 lines long
After update, JSON save file is 5444 lines long

It will save some space on the database and make JSON processing for reload faster.

Game.first is still a Player because changing this will require to update all create game calls...

### DIFF
--- a/src/database/PostgreSQL.ts
+++ b/src/database/PostgreSQL.ts
@@ -15,17 +15,23 @@ export class PostgreSQL implements IDatabase {
             }
         });
         this.client.connect();
-        this.client.query("CREATE TABLE IF NOT EXISTS games(game_id varchar, save_id integer, game text, status text default 'running', created_time timestamp default now(), PRIMARY KEY (game_id, save_id))", (err) => {
+        this.client.query("CREATE TABLE IF NOT EXISTS games(game_id varchar, players integer, save_id integer, game text, status text default 'running', created_time timestamp default now(), PRIMARY KEY (game_id, save_id))", (err) => {
             if (err) {
                 throw err;
             }
             console.log("Connected to PostgreSQL");
         });
+        this.client.query("CREATE INDEX IF NOT EXISTS games_i1 on games(save_id)", (err) => {
+            if (err) {
+                throw err;
+            }
+            console.log("Connected to PostgreSQL");
+        });        
     }
 
     getClonableGames( cb:(err: any, allGames:Array<IGameData>)=> void) {
         const allGames:Array<IGameData> = [];
-        const sql = "SELECT distinct game_id game_id, game FROM games WHERE status = 'running' and save_id = 0 order by game_id asc";
+        const sql = "SELECT distinct game_id game_id, players players FROM games WHERE status = 'running' and save_id = 0 order by game_id asc";
 
         this.client.query(sql, (err, res) => {
             if (err) {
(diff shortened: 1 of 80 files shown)


pierrehilbert:
We discussed a lot about this subject, many thanks for that!!

bafolts:
This is a great change. I don't think we necessarily have to lose every running game in order to introduce this change. If I don't grow too frustrated I will:

1. Take down the server
2. Run a script on the postgres database which adds the column for player count
3. Run a script to populate the player count column and  modified the stored JSON to match the new format
4. Start the server by merging this to master

If at some point during this process I grow frustrated or mess something up I will just wipe the games table and start fresh.

src/database/PostgreSQL.ts
```diff
@@ -15,17 +15,23 @@ export class PostgreSQL implements IDatabase {
...
         });
         this.client.connect();
-        this.client.query("CREATE TABLE IF NOT EXISTS games(game_id varchar, save_id integer, game text, status text default 'running', created_time timestamp default now(), PRIMARY KEY (game_id, save_id))", (err) => {
+        this.client.query("CREATE TABLE IF NOT EXISTS games(game_id varchar, players integer, save_id integer, game text, status text default 'running', created_time timestamp default now(), PRIMARY KEY (game_id, save_id))", (err) => {
             if (err) {
                 throw err;
             }
             console.log("Connected to PostgreSQL");
         });
+        this.client.query("CREATE INDEX IF NOT EXISTS games_i1 on games(save_id)", (err) => {
```
bafolts, line 24: I already created this index at some point:

```
    Column    |            Type             | Collation | Nullable |     Default     
--------------+-----------------------------+-----------+----------+-----------------
 game_id      | character varying(32)       |           | not null | 
 save_id      | integer                     |           | not null | 
 game         | text                        |           |          | 
 status       | text                        |           |          | 'running'::text
 created_time | timestamp without time zone |           |          | now()
 players      | integer                     |           |          | 
Indexes:
    "games_pkey" PRIMARY KEY, btree (game_id, save_id)
    "save_id" btree (save_id)
```

bafolts:
Updating the data was taking far too long. There were over 40k rows. My computer isn't fast enough to work through those in a timely manner. I wiped the games table. Any games that were playing were lost.

bafolts:
I displayed the heroku maintenance mode during the swap so anyone playing an active game should have at least gotten the notice that I was doing maintenance and is hopefully less upset about the loss. This change will help with the frugal cheapest heroku hosting and posgres data plan being used.

jijiqw:
emmmmm ,  we really don't need to change so much code ,  I  suggest another implementation 
1. add a prop stringifyPlayers  in class  Game
image
2. clear  stringifyPlayers before  anywhere call  saveGameState
image
3.add some codes  in  replacer func
image

the  stringifyPlayers prop records the players  that  have been stringified ,and  ensure every player stringified  once, the  other player ref  would be replaced with "{id: player.id}" . 
we can rebuild 'players' by code like :
`const player = this.players.find((player) => player.id === element.player!.id);`
image

jijiqw:
It  's simple, compatible and also decrease  JSON save file lines
image

PR #890, opened 2020-06-29
Various fixes

alrusdi:
Option to start server without 404 errors output to stdout. Start it like:
**npm run start -- hide-not-found-warnings**
Fix typo in russian localization
Show ruling party name on Turmoil board
Снимок экрана от 2020-06-29 20-45-35


vincentneko:
Maybe it would be better not to hide some errors by default and therefore use a -- hide-not-found-warnings ?

alrusdi:
> Maybe it would be better not to hide some errors by default and therefore use a -- hide-not-found-warnings ?

Done!

vincentneko:
> > Maybe it would be better not to hide some errors by default and therefore use a -- hide-not-found-warnings ?
> 
> Done!

Thanks =)

server.ts
```diff
@@ -805,7 +805,9 @@ function getGame(game: Game): string {
 }
 
 function notFound(req: http.IncomingMessage, res: http.ServerResponse): void {
-  console.warn('Not found', req.method, req.url);
+  if (process.argv.includes("hide-not-found-warnings")) {
+    console.warn('Not found', req.method, req.url);
```
vincentneko, line 809: I think the if should be reversed

PR #893, opened 2020-06-30
Add one-time alert when Mars is terraformed

nwai90:
**Ref:** https://github.com/bafolts/terraforming-mars/issues/777

- Show an alert only once when Mars is fully terraformed
- Don't show the alert on game_end screen
- Basic alert for now, in future this can be a nicer styled onscreen animation with sound

Screenshot 2020-06-30 at 6 48 56 PM

### DIFF
--- a/src/components/Board.ts
+++ b/src/components/Board.ts
@@ -11,8 +11,12 @@ class GlobalParamLevel {
     }
 }
 
+class AlertDialog {
+    static shouldAlert = true;
+}
+
 export const Board = Vue.component("board", {
-    props: ["spaces", "venusNextExtension", "venusScaleLevel","boardName", "oceans_count", "oxygen_level", "temperature"],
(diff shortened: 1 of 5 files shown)


alrusdi:
@nwai90 Can you please make an option (enabled by default) for it in Preferences dialog?

nwai90:
> @nwai90 Can you please make an option (enabled by default) for it in Preferences dialog?

@alrusdi Sure, I can add it. Do we want to introduce a preference option for this though? As it is just one alert throughout the entire game.

Screenshot 2020-06-30 at 9 03 01 PM

alrusdi:
@nwai90 i don't know. It will be little annoying to me personally. If it is not hard to do - please do it :-)

nwai90:
> @nwai90 i don't know. It will be little annoying to me personally. If it is not hard to do - please do it :-)

@alrusdi Done, I added the option to PreferencesManager

PR #898, opened 2020-07-01
An option to set number of rounds of initial draft

alrusdi:
initial_draft_rounds


alrusdi:
@vincentneko can you please review this PR too? I need it for our local tournament. Or you don't like it? Can i modify it somhow if so?

alrusdi:
> @vincentneko can you please review this PR too? I need it for our local tournament. Or you don't like it? Can i modify it somhow if so?

Will it be ok to you if i will change it dynamically when user changes players count? As result it will be more flexible

alrusdi:
@vincentneko please look at it. Now a default behavior should be the same as before

PR #906, opened 2020-07-03
Wild tag count for turomail ruling bonus fixed and a new property increaseTerraformRatingThisGeneration for UN and Pristar skill

zcy921716:


src/Game.ts
```diff
@@ -862,9 +861,7 @@ export class Game implements ILoadable<SerializedGame, Game> {
 
       this.players.forEach((player) => {
         player.terraformRatingAtGenerationStart = player.getTerraformRating();
-        if(player.corporationCard?.name === CardName.PRISTAR){
-          (player.corporationCard as Pristar).lastGenerationTR = player.getTerraformRating();
-        }
+        player.increaseTerraformRatingThisGeneration = false;
```
nwai90, line 864: `hasIncreasedTerraformRatingThisGeneration` or simply `hasIncreasedTRThisGeneration` might be a clearer name.

Using this property could allow us to phase out `terraformRatingAtGenerationStart` at some point too, since it no longer seems to be needed after this change.

alrusdi:
@nwai90 test failure. Can you please fix it. Is it real problem?

nwai90:
> @nwai90 test failure. Can you please fix it. Is it real problem?

@zcy921716 The Pristar test is failing:

```
1) Pristar
       Should play:

      AssertionError: expected 10 to equal 16
      + expected - actual

      -10
      +16
      
      at Context.<anonymous> (dist/tests/cards/turmoil/Pristar.spec.js:24:50)
```

I think you can remove lines 18-21 in the test; we should not need to call Pristar's `onProductionPhase` twice

src/Game.ts
zcy921716: Thank you. Name has been changed to `hasIncreasedTerraformRatingThisGeneration`. 

 Yes I think `terraformingRatingAtGenerationStart` is no longer be needed now. But I am not sure if it might cause other issues so I just leave it as before.

zcy921716:
> @nwai90 test failure. Can you please fix it. Is it real problem?

I updated the test file now it can pass the test!

zcy921716:
> > @nwai90 test failure. Can you please fix it. Is it real problem?
> 
> @zcy921716 The Pristar test is failing:
> 
> ```
> 1) Pristar
>        Should play:
> 
>       AssertionError: expected 10 to equal 16
>       + expected - actual
> 
>       -10
>       +16
>       
>       at Context.<anonymous> (dist/tests/cards/turmoil/Pristar.spec.js:24:50)
> ```
> 
> I think you can remove lines 18-21 in the test; we should not need to call Pristar's `onProductionPhase` twice

Yes I have already updated the test file. Thank you for figuring it out.

src/Game.ts
nwai90: Yup, it will cause some issues if removed directly. We can plan to deprecate both `terraformingRatingAtGenerationStart` and `lastGenerationTR` later on

zcy921716:
delegates related card bug fixed (more than 7 delegates)

src/cards/turmoil/MartianMediaCenter.ts
```diff
@@ -27,13 +27,13 @@ export class MartianMediaCenter implements IProjectCard {
         return undefined;
     }
 
-    public canAct(player: Player): boolean {
-        return player.canAfford(3);
+    public canAct(player: Player, game: Game): boolean {
+        return (player.canAfford(3) && game.turmoil!.getDelegates(player.id) > 0);
```
nwai90, line 31: 👍 This logic of checking if a player has delegates is used in several places. We can add a method in `Turmoil.ts` for it:

```
public hasAvailableDelegates(playerId: PlayerId | "NEUTRAL"): boolean {
    return this.delegate_reserve.filter(p => p === playerId).length > 0;
}
```

zcy921716: Yep we can. But I feel `getDelegates` might be enough at this stage. Like for CulturalMetropolis we still need to know the number of available delegates.

PR #908, opened 2020-07-03
Add Big Box promo cards

nwai90:
(Only need to examine the last 3 commits here as everything else has been checked previously)


src/cards/promo/PharmacyUnion.ts
```diff
@@ -0,0 +1,84 @@
...
+    public onCardPlayed(player: Player, game: Game, card: IProjectCard): void {
+        if (this.isDisabled) return undefined;
+        
+        if (card.tags.includes(Tags.MICROBES)) {
+            const microbeTagCount = card.tags.filter((cardTag) => cardTag === Tags.MICROBES).length;
+            player.addResourceTo(this, microbeTagCount);
+            player.megaCredits = Math.max(player.megaCredits - microbeTagCount * 4, 0)
+        }
+            
+        if (player.isCorporation(CorporationName.PHARMACY_UNION) && card.tags.includes(Tags.SCIENCE)) {
```
vincentneko, line 51: This will not work with Research card (2 sciences tags)
Mars University has the proper logic to handle this and can be used a source of inspiration

nwai90: 👍 Great catch, will update

nwai90: @vincentneko Updated, can you see if this is ok?

alrusdi:
I have a question about that replacement cards. Are they playable now or we have a naming conflict?

nwai90:
> I have a question about that replacement cards. Are they playable now or we have a naming conflict?

They should be playable, since they replace the original cards. Will test it for a bit to make sure that the 3 replacement cards can be played without issues

nwai90:
I added missing HTML data for the 3 promo cards and fixed the incorrect name for Asteroid Deflection System

Screenshot 2020-07-04 at 12 03 40 PM
Screenshot 2020-07-04 at 11 52 42 AM
Screenshot 2020-07-04 at 12 06 09 PM
Screenshot 2020-07-04 at 12 25 35 PM

nwai90:
Also updated tile placement logic for Great Dam promo:

Screenshot 2020-07-04 at 12 46 41 PM

The tile has to be placed adjacent to an ocean.

src/styles/cards.less
```diff
@@ -1607,6 +1607,12 @@ input[type=radio]:checked + .filterDiv {
   }
 }
 
+.card-beam-from-a-thorium-asteroid {
+    .title {
+      font-size: 13px;
```
nwai90, line 1612: Possibly automate this styling in future if needed, e.g. use `font-size: 14px` if title > 20 characters, and `font-size: 13px` if title > 24 characters

Issue #918, opened 2020-07-05
Landlord image used for Estate Dealer and Desert Settler awards


samyelias:
The image used for both Estate Dealer and Desert Settler are the one for Landlord.

Screenshot: https://i.imgur.com/WtrmOWq.png

alrusdi:
Fixed & Merged

PR #919, opened 2020-07-05
Fix wrong images for awards

nwai90:
**Ref:** https://github.com/bafolts/terraforming-mars/issues/918

Screenshot 2020-07-05 at 2 09 26 PM


nwai90:
@alrusdi Is this ok or should we just change the class names in `src/styles/player_home.less`?

PR #920, opened 2020-07-05
DRY and cleanup all tests except Turmoil

nwai90:


alrusdi:
@nwai90 there are failing tests. Are they related to this changes?

nwai90:
> @nwai90 there are failing tests. Are they related to this changes?

I had a look. It seems that some tests are flaky when running on CI, although they pass locally. Examples:

- Turmoil.spec.ts
- LavaFlows.spec.ts
- LavaTubeSettlement.spec.ts

nwai90:
@alrusdi I fixed it by adding a function to reset the board across tests, but I think the intermittent test failures are worth a relook (especially for Turmoil). Not sure if it is related to config or other causes. Do you know if the tests are running in parallel on CI?

alrusdi:
I don't think if they are run in parallel. But some problems is there. And i am experiencing them from time to time on localhost too.

Issue #931, opened 2020-07-06
Bug : edge case for MANUTECH


jijiqw:
1.the same as [https://github.com/bafolts/terraforming-mars/pull/853](https://github.com/bafolts/terraforming-mars/pull/853), plays with corp MANUTECH  should be able to play Moss with 0 plants and Nitrophilic Moss with 0 plant.

2. the player with corp MANUTECH should be able to place the city tile on the south pole .  Get 3 MC for 3MC prod according to MANUTECH‘s effect  , then he have 6MC to pay for  
![image](https://user-images.githubusercontent.com/16096558/86616587-a46ac500-bfe8-11ea-873c-4e2bb22aad47.png)
![image](https://user-images.githubusercontent.com/16096558/86616585-a3399800-bfe8-11ea-8b34-caee3e9270ad.png)

It ‘s a litte complex , there are many cards which place a city tile on mars need be changed.
Could anyone help fix this one?  If none , I would try to fix myself later.

nwai90:
You could define `bonusMc` property for each affected card and add it to `player.megaCredits` if that player's corporation is Manutech. This change will probably touch a lot of cards

jijiqw:
> You could define `bonusMc` property for each affected card and add it to `player.megaCredits` if that player's corporation is Manutech. This change will probably touch a lot of cards

It  seems to be a good idea !

BTW,  another case:
The player has 2MC  , then play a card that increase 2MC pro ， place a tile on south pole , place a ocean which adjacent to another ocean to get 2MC bonus .  at last  he can pay  6 . 
A lot of work to do if we want to strictly  implement the rule

nwai90:
The `bonusMc` is currently used for some Prelude cards to resolve issues related to play order in https://github.com/bafolts/terraforming-mars/issues/761, you can use that implementation as a reference. I think the normal edge case can be handled with this approach.

For the other case:
```
The player has 2MC , then play a card that increase 2MC pro ， place a tile on south pole , place a ocean which adjacent to another ocean to get 2MC bonus . at last he can pay 6 .
A lot of work to do if we want to strictly implement the rule
```

I think this is quite an extreme edge case, albeit a valid scenario. Maybe deal with it later

alrusdi:
Why do we read the rules like this? Player must met all requirements before playing the card isn't it?

nwai90:
[...]

PR #933, opened 2020-07-06
Fix edge case: Manutech can play Moss cards with 1 less plant required

nwai90:
**Ref:** https://github.com/bafolts/terraforming-mars/issues/931

Partial fix (point 1 only)


jijiqw:
Manutech can play NitrophilicMoss  with 0  plant

PR #958, opened 2020-07-09
Fix SelectParty when sending delegate from reserve

nwai90:
The default value for `fromLobby` in `SelectParty` is true. It seems that it should be false here when paying 5 MC to send a delegate from the reserve, not from the lobby.


vincentneko:
Yes you are right.
No bug is occurring right now because lobby will always be empty when you pay 5 MC to send a delegate from reserve.

vincentneko:
I will not merge it now as it is not a blocking bug, we can wait for another PR to merge all together.

PR #960, opened 2020-07-09
Merge in-game alerts into single option

nwai90:


alrusdi:
It is interesting. But what if i want overpayment warning, but don't want "mars is terraformed" notification?
More options - more flexibility isn't it?

nwai90:
> It is interesting. But what if i want overpayment warning, but don't want "mars is terraformed" notification?
> More options - more flexibility isn't it?

True, this could also lead to having too many options though. Especially when there are more alerts added (e.g. Pass confirmation, other types of confirmations).

I'm fine with it both ways but might be too micro and perhaps overkill to be configuring each alert individually. Steam app also merges showing or hiding of alerts into a single config option which looks cleaner.

alrusdi:
I'm not against it if everyone else like it. Let's wait a review from other contributors.

src/components/SelectHowToPay.ts
```diff
@@ -2,6 +2,7 @@
 import Vue from "vue";
 import { HowToPay } from "../inputs/HowToPay";
 import { PaymentWidgetMixin } from "./PaymentWidgetMixin";
+import { PreferencesManager } from "./PreferencesManger";
```
bafolts, line 5: This file should be named `PreferencesManager` and not `PreferencesManger`.

nwai90: 👍 I renamed the file

PR #971, opened 2020-07-11
add  promo cards into RoboticWorkforce 's copy list

jijiqw:
There might be more cards need be added into,I have no time to check it all


nwai90:
@jijiqw I think you missed out the promo card `Field-Capped City`, could you add that in as well?

jijiqw:
Updated, thanks!

PR #988, opened 2020-07-14
Rename BribedCommittee and SuccessfulOrganisms

nwai90:
Note: This is a breaking change


vincentneko:
@bafolts As mentioned before, this will break games, I'm not sure if the database needs to be flushed before because some games will not reload...

jijiqw:
image
what about add some code like that ?
Cause there are thousands of games on my server, It's  a little unacceptable if we need to flush the database.

vincentneko:
@nwai90 Can you please resolve the conflicts ?

bafolts:
If there is an issue deserializing a game the game is ignored.

https://github.com/bafolts/terraforming-mars/blob/master/src/database/PostgreSQL.ts#L110

If these referenced names ended up not being found they would have been ignored while the server was starting up.

Issue #936, opened 2020-07-06
Option to have 5 Random Milestones / Achievements


duckmammal:
Right now, when playing with the Venus Expansion and Randomizer, there are 6 milestones and 6 awards, and Venus is not required to be among them. My group likes to play with 5 of each, even when Venus is in play.

nwai90:
Closed as resolved in https://github.com/bafolts/terraforming-mars/pull/993

PR #993, opened 2020-07-15
Add option to remove 6th milestone and award for Venus games

nwai90:
**Ref:** https://github.com/bafolts/terraforming-mars/issues/936

- This option is only shown if Venus Next expansion is selected (default: ON)
- If deselected, the game will have only 5 milestones and awards, which can optionally be randomized

**Note:** Needs help to add this field on the database side also

Screenshot 2020-07-15 at 4 38 58 PM


duckmammal:
Great job! You're all doing such good work to close tickets made by strangers like me! It's really appreciated!

Issue #1000, opened 2020-07-16
Bug:buy cards with  corporation  helion


jijiqw:
![image](https://user-images.githubusercontent.com/16096558/87618843-62125800-c74d-11ea-8709-aa3310fd4816.png)
Can't  adjust mc  by  '+-' button  , however  'pay by heat'  buttons  work fine .

nwai90:
Should be resolved in https://github.com/bafolts/terraforming-mars/pull/1001

PR #1001, opened 2020-07-16
Fix: Allow Helion to adjust MC amount when buying cards during Research

nwai90:
Ref: https://github.com/bafolts/terraforming-mars/issues/1000

Screenshot 2020-07-16 at 5 36 34 PM

(Checking by title to avoid introducing too many changes)

### DIFF
--- a/src/components/PaymentWidgetMixin.ts
+++ b/src/components/PaymentWidgetMixin.ts
@@ -28,6 +28,7 @@ export const PaymentWidgetMixin = {
         },
         reduceValue: function (target: string, to: number): void {
             let currentValue: number = (this as any)[target];
+            const isResearchPhase = (this as any).$data.isResearchPhase;
 
             if (currentValue === 0) return;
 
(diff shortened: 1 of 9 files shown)


src/components/PaymentWidgetMixin.ts
```diff
@@ -36,12 +37,19 @@ export const PaymentWidgetMixin = {
...
             let currentValue: number = (this as any)[target];
             let maxValue: number = (this as any).player[target];
-            if (target === "megaCredits") maxValue = this.getMegaCreditsMax();
+            const isResearchPhase = (this as any).$data.isResearchPhase;
+
+            if (isResearchPhase && target === "megaCredits") {
+                maxValue = (this as any).player.cardCost * 4;
+            } else if (target === "megaCredits") {
+                maxValue = this.getMegaCreditsMax();
+            }
```
jijiqw, line 51: globalevent may decrease players's mc up to 20 ,  cardCost * 4 may greater than player.megaCredits

nwai90: GlobalEvents don't trigger during the Research phase. This change is specific to `player.runResearchPhase()`. `player.cardCost` is used to account for Terralabs and Polyphemos.

If player attempts to overspend MC an error will be thrown: `"You don't have that many mega credits"`

nwai90:
@vincentneko @alrusdi @bafolts I found another bug on `master`, also fixed in this PR. If you choose to buy 0 cards as Helion during Research phase but decide to "spend" n heat, that heat gets incorrectly converted into MC.

alrusdi:
@nwai90 is it ready to be merged? I don't see any problems

nwai90:
It should be ready; I removed the `wip` label.

Issues like overpayment and underpayment are already handled in existing code. I think we can improve it as we go along.

src/components/PaymentWidgetMixin.ts
jijiqw: It's Ok ,thanks

PR #1005, opened 2020-07-17
Robots strike back

vincentneko:
This one is long overdue !
Self Replicating Robots can now store multiple cards.
Stored cards under SRR are displayed in a separate section, with counters.

### DIFF
--- a/src/cards/promo/SelfReplicatingRobots.ts
+++ b/src/cards/promo/SelfReplicatingRobots.ts
@@ -3,12 +3,17 @@ import { CardName } from '../../CardName';
 import { CardType } from '../CardType';
 import { Tags } from '../Tags';
 import { Player } from '../../Player';
-import { ResourceType } from '../../ResourceType';
 import { SelectCard } from '../../inputs/SelectCard';
 import { Game } from '../../Game';
 import { LogMessageType } from '../../LogMessageType';
 import { LogMessageData } from '../../LogMessageData';
 import { LogMessageDataType } from '../../LogMessageDataType';
+import { OrOptions } from "../../inputs/OrOptions";
+
+export interface RobotCard {
+    card: IProjectCard;
+    resourceCount: number;
+}
 
 export class SelfReplicatingRobots implements IProjectCard {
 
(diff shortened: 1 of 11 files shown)


src/components/SelectHowToPayForCard.ts
```diff
@@ -52,13 +52,14 @@ export const SelectHowToPayForCard = Vue.component("select-how-to-pay-for-card",
...
     methods: {
         getCardCost: function () {
-            for (const icard of this.player.cardsInHand) {
+            let cards = this.player.cardsInHand.concat(this.player.selfReplicatingRobotsCards);
+            for (const icard of cards) {
                 if (this.$data.card.name === icard.name) {
                     return icard.calculatedCost
                 }
             }
             // If not found, it should be self replication robot stored card
```
nwai90, line 61: I think we can remove these comments as they are no longer relevant with this change

src/components/OtherPlayer.ts
```diff
@@ -64,8 +64,16 @@ export const OtherPlayer = Vue.component("other-player", {
                         <stacked-cards :cards="getCardsByType(player.playedCards, [getAutomatedCardType(), getPreludeCardType()])" :player="player"></stacked-cards>
                         <stacked-cards :cards="getCardsByType(player.playedCards, [getEventCardType()])" :player="player"></stacked-cards>
                     </div>
-                </div> 
+                </div>
 
+                <div v-if="player.selfReplicatingRobotsCards.length > 0" class="player_home_block">
+                    <span> Self-Replicating robots cards </span>
```
nwai90, line 70: Minor suggestion: Capitalize `robots` here and in `PlayerHome.ts` as well

src/cards/promo/SelfReplicatingRobots.ts
```diff
@@ -68,9 +74,35 @@ export class SelfReplicatingRobots implements IProjectCard {
...
+                'Select card to double robots resource', robotCards,
+                (foundCards: Array<IProjectCard>) => {
+                  for (let targetCard of this.targetCards) {
+                    if (targetCard.card.name === foundCards[0].name) {
+                      targetCard.resourceCount = targetCard.resourceCount * 2;
+                    }
+                  }
+                  game.log(
+                    LogMessageType.DEFAULT,
+                    "${0} doubled resources on ${1} with ${2}",
```
nwai90, line 95: It may be clearer as `${0} doubled resources on ${1} from ${2} to ${3}`, so that there is no need to scroll down to view the card.

We probably don't need to include `this.name` here as this log message will appear together with `"PLAYER used Self-Replicating Robots action"`.

nwai90:
Screenshot 2020-07-17 at 7 34 08 PM

Would be a bonus to also display robots resource count here (avoids having to scroll down)

vincentneko:
> Screenshot 2020-07-17 at 7 34 08 PM
> 
> Would be a bonus to also display robots resource count here (avoids having to scroll down)

This is not directly linked with SRR, I think it should be handled in another PR.

vincentneko:
@nwai90 Thanks for the feedback, I've updated, would you please check if everything is ok for you ?

nwai90:
@vincentneko Everything LGTM 👍

Issue #1023, opened 2020-07-19
Venus Soils prompt incorrect


duckmammal:
The prompt asks the user to select a card to which it will add 1 resource. The card is supposed to (and does) add 2 resources, not 1, but the prompt after playing Venus Soils does not reflect this.

nwai90:
Resolved in https://github.com/bafolts/terraforming-mars/pull/1025

PR #1025, opened 2020-07-19
Fix prompt for Venus Soils

nwai90:
Ref: https://github.com/bafolts/terraforming-mars/issues/1023

### DIFF
--- a/src/HTML_data.ts
+++ b/src/HTML_data.ts
@@ -4048,7 +4048,7 @@ export const HTML_DATA: Map<string, string> =
               </div>
               <div class="resource microbe" style="margin-left:20px;"></div><div class="resource microbe"></div>*
               <div class="description">
-                (Raise Venus 1 step. Increase your Plant production 1 step. Add 2 Microbes to ANOTHER card)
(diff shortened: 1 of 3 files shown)


src/HTML_data.ts
```diff
@@ -4048,7 +4048,7 @@ export const HTML_DATA: Map<string, string> =
               </div>
               <div class="resource microbe" style="margin-left:20px;"></div><div class="resource microbe"></div>*
               <div class="description">
-                (Raise Venus 1 step. Increase your Plant production 1 step. Add 2 Microbes to ANOTHER card)
+                (Raise Venus 1 step. Increase your Plant production 1 step. Add 2 Microbes to ANOTHER card.)
```
alrusdi, line 4051: Can you please change it in russian translation too?
If you will make new PR's from main repository i'll be able to add commits with that kind of fixes myself. Do you have any reasons to use your own repository instead of bafolts' one?

nwai90: @alrusdi I added the missing `.`. Sure, I can push new branches to origin if that's better ;)

PR #1032, opened 2020-07-21
Tag overview and corp names

meussiebeaucoup:
- Initial version for tag overview grid
- Initial version for displaying corp name under player name

I did a quick implementation of these two features, we feel that it saves us a lot time of going over every player's boards, especially when dealing with milestones / awards. 

I didn't come round to making the tag overview optional via the configuration screen at the start of the game, but I can imagine some people might want to have the option to turn it off.

Feel free to use whatever you can, even if just the concept.


vincentneko:
Thanks for your contribution.
I think theses 2 features should be activated or deactivated with the preference panel.

Screenshot 2020-07-22 08 10 41

And as a personal preference, I would set the background color to the player's color for the tag overview.

meussiebeaucoup:
Agreed, I'll try and add it over the weekend.

vincentneko:
There is something strange with this MR, as it aims to merge 151 commit into master.
@meussiebeaucoup can you please clone master again and only apply the changes related with Tag overview and corp names ?

meussiebeaucoup:
Hi Vincent,

Looks like something went wrong with an previous commit, my apologies. I've committed a new version, and I think I've resolved it, looking at the number of files changed / commit number. If there is still something let me know so I can try and resolve it.

I've made some changes to the initial commit, adding the tag overview to the option section and cleaning up the table a little bit, through spacing and background color:

image

image

nwai90:
@meussiebeaucoup I updated the wiki to include a tooltip link for "Tag Overview" (https://github.com/bafolts/terraforming-mars/wiki/Variants#tag-overview). Changes LGTM overall 👍 

We could probably improve the styling for corporation names, e.g.

1) Use a constant box width and maybe ellipsis for text-overflow? Green box is a lot longer than Red box
2) Remove the index (`1.`, `2.`, etc.) as it seems redundant to display here

Screenshot 2020-08-09 at 11 26 24 PM

meussiebeaucoup:
I've added the ellipses overflow for the player's name in the tag grid, I can also see a need for it when displaying the corp. name in the player box. I agree that the player numbers don't add a whole lot and can be removed.

I can add these changes, if there is a consensus about the width of the players-box names ;)

vincentneko:
Hi @meussiebeaucoup 
Thanks for the changes. The MR now seems good from a commit perspective.
The layout is great.
I think the "Tag Overview" does not belong to the game options panel, and it should be moved to the preference panel.
A preference to display corp names could also be a great addition.

Ronnoc:
Maybe add a new row for current vp is a good idea?

vincentneko:
> Maybe add a new row for current vp is a good idea?

meussiebeaucoup:
I can move it to the preferences, no problem.

The reason I added it to game options was that I thought it had an impact on the flow of the game, since it makes it a bit easier to keep track of other players' tags, maybe there are some cases in which players find that it takes away from the nature of the game (same as showing other players their VP score or award/milestone counters)

vincentneko:
> I can move it to the preferences, no problem.
> 
> The reason I added it to game options was that I thought it had an impact on the flow of the game, since it makes it a bit easier to keep track of other players' tags, maybe there are some cases in which players find that it takes away from the nature of the game (same as showing other players their VP score or award/milestone counters)

I see your point. I consider this public information (unlike VP scores), as it is already available for each player, so I think it's more a preference than an option. If someone has a different opinion please let us know.

meussiebeaucoup:
What do you guys think about this? 

image

I've added the 'cards in hand' column and a VP column, the latter is hidden when the 'Real time VP' option is disabled.

vincentneko:
> What do you guys think about this?
> 
> image
> 
> I've added the 'cards in hand' column and a VP column, the latter is hidden when the 'Real time VP' option is disabled.

This is great !

meussiebeaucoup:
Another week, another update.

- I've pushed the new table layout
- Some styling has been done to the player boxes as suggested in the comments above, feel free to discuss:

image

vincentneko:
Thanks for the update.
Having the corporation name in the player box in turn order could be another preference. I'm not sure myself if I like it or not.
About the tag overview table, I think zeroes are overloading the display and maybe displaying nothing instead of 0 would help.

meussiebeaucoup:
The corp names is a preference option:

image

image

I could reserve the options, so the text would say 'show player corporation names', instead of 'hide', but I felt this kept it more in line with the other options.

I tried it with an empty text box instead of displaying a '0' , but it doesn't look that great, especially in the early game when there are few cards played:

image

vincentneko:
Sorry, I've overlooked the other preference. Having corp displayed as default if ok.
I agree with you, dash instead of zero or nothing is much clear.

nwai90 (approved):

vincentneko (approved):

PR #1034, opened 2020-07-22
fix bug : red policy of some cards

jijiqw:


jijiqw:
Could you please rerun the tests ? @nwai90

src/cards/venusNext/ForcedPrecipitation.ts
```diff
@@ -41,8 +41,8 @@ export class ForcedPrecipitation implements IActionCard,IProjectCard, IResourceC
 
         const addResource = new SelectOption("Pay 2 to add 1 floater to this card", () => this.addResource(player));
         const spendResource = new SelectOption("Remove 2 floaters to raise Venus 1 step", () => this.spendResource(player, game));
-
-        if (this.resourceCount > 1 && game.getVenusScaleLevel() < MAX_VENUS_SCALE) {
+        const canAffordRed = !PartyHooks.shouldApplyPolicy(game, PartyName.REDS) || player.canAfford(REDS_RULING_POLICY_COST);
+        if (this.resourceCount > 1 && game.getVenusScaleLevel() < MAX_VENUS_SCALE && canAffordRed) {
```
nwai90, line 45: @jijiqw Should this also apply to `src/cards/venusNext/ExtractorBalloons.ts:38`? Only push the option to spend floaters to raise Venus if player can afford to pay while Reds are in power

jijiqw: Updated, we just add a resource and return undefined   if there is only one choice .

PR #1042, opened 2020-07-22
Update LocalHeatTrapping.ts

SliceOfBread:
You cannot use megaCredits as heat. To play this card you MUST have 5 heat.
The tricky part is if the player can use heat as megaCredits then they not only need 5 heat but ALSO heat + megaCredits >= 5 + cardCost
I fixed the return value and removed the spurious code that deducted MC if there was not enough heat.

### DIFF
--- a/src/cards/LocalHeatTrapping.ts
+++ b/src/cards/LocalHeatTrapping.ts
@@ -91,6 +91,8 @@ export class LocalHeatTrapping implements IProjectCard {
             );
           }
         
+        player.heat -= 5
+        
         if (availableActions.options.length === 1) return availableActions.options[0].cb();
         return availableActions;
     }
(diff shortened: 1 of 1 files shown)


src/cards/LocalHeatTrapping.ts
```diff
@@ -26,7 +26,7 @@ export class LocalHeatTrapping implements IProjectCard {
 
         // Helion must be able to pay for both the card and its effect
         if (player.canUseHeatAsMegaCredits) {
-          return player.heat + player.megaCredits >= requiredHeatAmt + player.getCardCost(game, this)
+          return (player.heat >= requiredHeatAmt) && (player.heat + player.megaCredits >= requiredHeatAmt + player.getCardCost(game, this))
```
nwai90, line 29: 👍 This is the correct approach. We still need to add `player.heat -= 5` on line 94 below though, so that 5 heat will be spent

SliceOfBread: Yes! My mistake and fixed.

PR #1044, opened 2020-07-22
Improve ms awards

vincentneko:
This will display current score for each player for milestones and awards.
The layout is far from perfect...
This will help a lot for some milestones / awards (looking at you Tycoon, Celebrity and Excentric)
It can be deactivated through game preferences.

Thanks @pierrehilbert for the sort !

### DIFF
--- a/src/milestones/Builder.ts
+++ b/src/milestones/Builder.ts
@@ -6,7 +6,10 @@ import { Tags } from "../cards/Tags";
 export class Builder implements IMilestone {
     public name: string = "Builder";
     public description: string = "Having at least 8 building tags in play"
+    public getScore(player: Player, _game: Game): number {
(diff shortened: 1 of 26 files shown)


src/milestones/Builder.ts
```diff
@@ -6,7 +6,10 @@ import { Tags } from "../cards/Tags";
 export class Builder implements IMilestone {
     public name: string = "Builder";
     public description: string = "Having at least 8 building tags in play"
+    public getScore(player: Player, _game: Game): number {
+        return player.getTagCount(Tags.STEEL);
+    }
     public canClaim(player: Player, _game: Game): boolean {
```
nwai90, line 12: Suggestion: Remove unneeded `_game` arguments from milestones that don't need them in `canClaim`

PR #1063, opened 2020-07-25
Fix venus global events

vincentneko:


vincentneko:
@nwai90 Can you please review again ? I've added missing global event (hopefully a simple one !)

src/turmoil/globalEvents/MicrogravityHealthProblems.ts
```diff
@@ -0,0 +1,22 @@
+import { IGlobalEvent } from './IGlobalEvent';
+import { GlobalEventName } from './GlobalEventName';
+import { PartyName } from '../parties/PartyName';
+import { Game } from '../../Game';
+import { Resources } from '../../Resources';
+import { Turmoil } from '../Turmoil';
+
+export class MicrogravityHealthProblems implements IGlobalEvent {
+    public name = GlobalEventName.MICROGRAVITY_HEALTH_PROBLEMS;
+    public description = "Lose 5 MC for each colony (max 5, then reduced by influence).";
```
nwai90, line 10: @vincentneko Should this be `Lose 3 MC for each colony`? I saw it is -3 on line 19 below

vincentneko: Good catch 👍

Issue #969, opened 2020-07-11
using Iron/Titanium to reduce a card cost can spend excess Iron/Titanium


Arsenki-33:
Minor UI issue.
When you take the **play a card** action, you often have the option to spend Iron or Titanium instead, or in addition to MC.
However the UI seems to let you spend **one extra material** even though the cost is already reduced to zero.

example: You want to pay for a cost 12 card the UI will let you spend 7 iron instead of 6 which is the useful limit.

PS: I'm completely new to GH, this is my first post. Congrats for the good work on thihs great boardgame.

nwai90:
Fixed and merged in https://github.com/bafolts/terraforming-mars/pull/1070

PR #1070, opened 2020-07-27
Fix overpayment of resources

SliceOfBread:
I believe this fixes issue #969 by ignoring attempts to use more resources once MC need is 0.

### DIFF
--- a/src/components/PaymentWidgetMixin.ts
+++ b/src/components/PaymentWidgetMixin.ts
@@ -54,8 +54,6 @@ export const PaymentWidgetMixin = {
             if (target === "floaters") maxValue = (this as any).playerinput.floaters;
             if (currentValue === maxValue) return;
 
-            if (target !== "megaCredits" && (this as any)["megaCredits"] === 0) return;
-            
             const realTo = (currentValue + to <= maxValue) ? to : maxValue - currentValue;
             (this as any)[target] += realTo;
 
--- a/src/components/PaymentWidgetMixin.ts
+++ b/src/components/PaymentWidgetMixin.ts
@@ -54,6 +54,8 @@ export const PaymentWidgetMixin = {
             if (target === "floaters") maxValue = (this as any).playerinput.floaters;
             if (currentValue === maxValue) return;
 
+            if (target !== "megaCredits" && (this as any)["megaCredits"] === 0) return;
+            
             const realTo = (currentValue + to <= maxValue) ? to : maxValue - currentValue;
             (this as any)[target] += realTo;
 
--- a/src/components/SelectHowToPay.ts
+++ b/src/components/SelectHowToPay.ts
@@ -157,6 +157,26 @@ export const SelectHowToPay = Vue.component("select-how-to-pay", {
               htp.megaCredits = 0;
             }
 
+            if (requiredAmt > 0 && totalSpentAmt > requiredAmt) {
+                let diff = totalSpentAmt - requiredAmt;
+                if (htp.titanium && diff >= this.player.titaniumValue) {
+                    this.$data.warning = "You cannot overspend titanium";
+                    return;
+                }
+                if (htp.steel && diff >= this.player.steelValue) {
+                    this.$data.warning = "You cannot overspend steel";
+                    return;
+                }
+                if (htp.heat && diff >= 1) {
+                    this.$data.warning = "You cannot overspend heat";
+                    return;
+                }
+                if (htp.megaCredits && diff >= 1) {
+                    this.$data.warning = "You cannot overspend megaCredits";
+                    return;
+                }
+            }
+            
             const showAlert = PreferencesManager.loadValue("show_alerts") === "1";
 
             if (requiredAmt > 0 && totalSpentAmt > requiredAmt && showAlert) {
--- a/src/components/SelectHowToPayForCard.ts
+++ b/src/components/SelectHowToPayForCard.ts
@@ -268,6 +268,34 @@ export const SelectHowToPayForCard = Vue.component("select-how-to-pay-for-card",
                 return;
             }
 
+            if (totalSpentAmt > this.getCardCost()) {
+                let diff = totalSpentAmt - this.getCardCost();
+                if (htp.titanium && diff >= this.player.titaniumValue) {
+                    this.$data.warning = "You cannot overspend titanium";
+                    return;
+                }
+                if (htp.steel && diff >= this.player.steelValue) {
+                    this.$data.warning = "You cannot overspend steel";
+                    return;
+                }
+                if (htp.floaters && diff >= 3) {
+                    this.$data.warning = "You cannot overspend floaters";
+                    return;
+                }
+                if (htp.microbes && diff >= 2) {
+                    this.$data.warning = "You cannot overspend microbes";
+                    return;
+                }
+                if (htp.heat && diff >= 1) {
+                    this.$data.warning = "You cannot overspend heat";
+                    return;
+                }
+                if (htp.megaCredits && diff >= 1) {
+                    this.$data.warning = "You cannot overspend megaCredits";
+                    return;
+                }
+            }
+
             const showAlert = PreferencesManager.loadValue("show_alerts") === "1";
             
             if (totalSpentAmt > this.getCardCost() && showAlert) {


vincentneko:
This is looking good to me, just tested, it fixes the issue. @alrusdi and @nwai90 can you confirm this change will not induce side effects ?

nwai90:
👌 I'll check this today

nwai90:
@SliceOfBread There is a bug: Helion cannot increase heat amount when MC amount is at 0 during research phase. The below change should fix it. I also found another issue re Helion, but will raise that separately.

Screenshot 2020-07-28 at 8 44 26 AM

`if (target !== "megaCredits" && (this as any)["megaCredits"] === 0 && !isResearchPhase) return;`

SliceOfBread:
So research phase does not do default payment like card pay. 
I also think the original problem is worse. If a card can be paid by two alternatives to MC then it can be overpaid. For example, a card costs 9. Spend 8 heat and then 1 titanium; heat is overspent by 2. I will verify tomorrow. If true,  fix will be more complicated.

nwai90:
> So research phase does not do default payment like card pay.
> I also think the original problem is worse. If a card can be paid by two alternatives to MC then it can be overpaid. For example, a card costs 9. Spend 8 heat and then 1 titanium; heat is overspent by 2. I will verify tomorrow. If true, fix will be more complicated.

Yes, I noticed that it is more complicated for Research phase due to Helion's ability to use heat for payment. We should probably fix that in a separate PR

SliceOfBread:
One way to fix this would be to remove my previous fix and add at the end of the addValue routine a check to see if anything is being overspent and, if so, remove the overspend. (This would also fix the bug @nwai90 found.) But deciding what to remove can get tricky if there are multiple choices (e.g. add 1 titanium and now steel is overspent by 1 or heat is overspent by 2). Anything else I'm missing?

nwai90:
@SliceOfBread The other issue https://github.com/bafolts/terraforming-mars/issues/1075 should have been resolved in https://github.com/bafolts/terraforming-mars/pull/1077.

For the above bug referenced in this PR that prevents Helion from using heat during Research phase, it could earlier be addressed by updating the code to:

`if (target !== "megaCredits" && (this as any)["megaCredits"] === 0 && !isResearchPhase) return;`

Can you see if this change is still relevant now that https://github.com/bafolts/terraforming-mars/pull/1077 has been merged?

SliceOfBread:
In my limited testing, there is no need for your suggestion as Helion can now pay with heat during research.

If we reframe the bug as "Players can overpay resources when playing cards" then the bug still exists but is much more limited then before. You have to look at an extreme case such as one player has Ants (steal microbe from another player) and the other player uses this bug to get rid of his last microbe. 

I'm going to ask about this on the BGG forum as there is no clear rule about it that I can find.

SliceOfBread:
I have confirmation from the game designer that you are not allowed to pay excess resources.

nwai90:
> I have confirmation from the game designer that you are not allowed to pay excess resources.

Thanks @SliceOfBread . I saw the thread at https://boardgamegeek.com/thread/2474183/how-much-can-you-overpay-things confirming that the most valuable resources are counted first.

I believe the current fix is largely sufficient to prevent excess overspending of other resources. There is one other bug with this approach though. Attached an example below:

Screenshot 2020-07-29 at 8 45 38 AM

Here, without any available discounts, I should be strictly limited to spending 2 titanium to pay for this card. At this point, if I increase MC amount from 0 to 1, I can now increase the titanium amount to 3 (which should be disallowed as per the ruling).

SliceOfBread:
I will look into that tomorrow. I've already put together some code that I hope to finish tomorrow to fix this.

SliceOfBread:
After much thought, I think the easiest thing to do would be to allow users to select any combination of resources they want limited only by the amount the have. The automatic update of MC should remain also. This keeps things quick.

To prevent the over payment, I suggest we add checks in the SelectHowToPay/SelectHowToPayForCards files that would ensure overpayment does not occur in the same way that we now warn of lack of resources or underpayment. Pop up a warning "You are trying to pay too much steel!" and let the user fix it. What do you think?

SliceOfBread:
I added the above update and tested it locally. It works for card payment and for paying to activate a card. It does not do anything for overpayment in research for Helion (I can't find that code). 
And my apologies but the change in PaymentWidgetMixin.ts should be removed but I can't figure out how to do that.

nwai90:
Payment for research phase has some separate code that prevents overspending by deducting only what is necessary to pay for the purchased cards.

@vincentneko @alrusdi Could you have a look at the latest changes in this PR? The new validations introduced here should eliminate overspend exploits 👍 

I saw two minor issues that aren't game-breaking - the implication is that some manual adjustments will be required during payment.

Screenshot 2020-07-30 at 10 27 11 AM

vincentneko:
> Payment for research phase has some separate code that prevents overspending by deducting only what is necessary to pay for the purchased cards.
> 
> @vincentneko @alrusdi Could you have a look at the latest changes in this PR? The new validations introduced here should eliminate overspend exploits 👍
> 
> I saw two minor issues that aren't game-breaking - the implication is that some manual adjustments will be required during payment.
> 
> Screenshot 2020-07-30 at 10 27 11 AM



> Payment for research phase has some separate code that prevents overspending by deducting only what is necessary to pay for the purchased cards.
> 
> @vincentneko @alrusdi Could you have a look at the latest changes in this PR? The new validations introduced here should eliminate overspend exploits 👍
> 
> I saw two minor issues that aren't game-breaking - the implication is that some manual adjustments will be required during payment.
> 
> Screenshot 2020-07-30 at 10 27 11 AM

This is looking good to me.
As a personal preference, I don't mind having to adjust payment for special cases, this is part of the game imho.

SliceOfBread:
The previous interface only updated MC after a change to any other resource (and this is still the case). It speeds things up as usually you just adjust steel/titanium and then you are done. It would be fairly easy to add an update if the choice is between two resources as the change of one directly affects the other but very limited usefulness. For 3 or more, you are just guessing.

SliceOfBread:
> And my apologies but the change in PaymentWidgetMixin.ts should be removed but I can't figure out how to do that.

Uhm, I figured this out by just deleting the code I had added.

vincentneko:
What is the state of this merge request ? Is there still some work to do on it or is it ready for review / merge ?

nwai90:
@vincentneko I just tested this against the latest version of `master` branch.

We should revert/remove the latest commit `2135a470ab65a5a7609f104f8d23d90cdd7bd376` for it to properly prevent overpayment. The change in that commit is required to prevent any resource from being increased beyond its allowed limit, otherwise we can freely increase any resource to any value:

Screenshot 2020-08-03 at 3 33 27 PM

With that latest commit ("No change") removed, there are some minor caveats as highlighted in earlier comments, but the checks introduced sufficiently prevent overspending of resources 👍

vincentneko:
> @vincentneko I just tested this against the latest version of `master` branch.
> 
> We should revert/remove the latest commit `2135a470ab65a5a7609f104f8d23d90cdd7bd376` for it to properly prevent overpayment. The change in that commit is required to prevent any resource from being increased beyond its allowed limit, otherwise we can freely increase any resource to any value:
> 
> Screenshot 2020-08-03 at 3 33 27 PM
> 
> With that latest commit ("No change") removed, there are some minor caveats as highlighted in earlier comments, but the checks introduced sufficiently prevent overspending of resources 👍

@nwai90 Thanks for the information. Should we add this line:
`if (target !== "megaCredits" && (this as any)["megaCredits"] === 0) return`;
or this one 
`if (target !== "megaCredits" && (this as any)["megaCredits"] === 0 && !isResearchPhase) return;`
?

nwai90:
@vincentneko It should be `if (target !== "megaCredits" && (this as any)["megaCredits"] === 0 && !isResearchPhase) return;` as Helion needs to able to spend heat during research phase without spending any MC.

vincentneko:
> @vincentneko It should be `if (target !== "megaCredits" && (this as any)["megaCredits"] === 0 && !isResearchPhase) return;` as Helion needs to able to spend heat during research phase without spending any MC.

@nwai90 Ok thanks for the information. Please hold on this MR. I'm currently working on a refactor of SelectHowToPay, and research phase should no longer be different from other situations...

SliceOfBread:
Without that line, the code works. Overpayment is not possible. 

With the line, you can lower MC to 0 and then you can't raise anything else. That is why I removed it.

vincentneko:
@nwai90 I think we can now merge this MR (Helion buying cards during research phase is now in line with other pay panels)
Can you please confirm ?

nwai90:
> @nwai90 I think we can now merge this MR (Helion buying cards during research phase is now in line with other pay panels)
> Can you please confirm ?

@vincentneko It seems good, let's merge it 👍

PR #1087, opened 2020-07-30
Reds' ruling bonus  for solo

jijiqw:
RULING BONUS: The player with lowest TR gains 1 TR. Ties are friendly. In solo, you receive 1 TR if you **have TR 20 or below**.


jijiqw:
Global Event: SnowCover  's    name property  is incorrectly set to SPONSORED_PROJECTS

Issue #1088, opened 2020-07-30
Typo of astrodrill action explanation


charlielam:
Refer to #1026 , the action should be "OR", not "AND"

![image](https://user-images.githubusercontent.com/6376909/88911267-34222d00-d290-11ea-9458-5ba60b9e3a64.png)

nwai90:
Resolved in https://github.com/bafolts/terraforming-mars/pull/1089

PR #1089, opened 2020-07-30
Update Astrodrill description

nwai90:
**Ref:** https://github.com/bafolts/terraforming-mars/issues/1088

Screenshot 2020-07-30 at 6 21 09 PM


nwai90:
@jijiqw You might want to update description in `src/locales/cn/bigboxpromo.json` also

jijiqw:
ok

PR #1092, opened 2020-07-31
Fixed restricted area tile misalignment

chosta:
Fixed an issue where the Restricted area card tile was misaligned

restricted-bug


nwai90:
@chosta Could you relook at the change? I just tested it on desktop using Chrome browser and see the following result:

after

Before (without any changes):

before

chosta:
@nwai90 Interesting.. I tested it on every possible browser / resolution and it seemed ok.
What OS are you on? I will check on another PC and will figure it out. Thanks for noticing.

nwai90:
> @nwai90 Interesting.. I tested it on every possible browser / resolution and it seemed ok.
> What OS are you on? I will check on another PC and will figure it out. Thanks for noticing.

I'm using Chrome on MacOS Mojave version 10.14.6. Could be an OS issue.

chosta:
> > @nwai90 Interesting.. I tested it on every possible browser / resolution and it seemed ok.
> > What OS are you on? I will check on another PC and will figure it out. Thanks for noticing.
> 
> I'm using Chrome on MacOS Mojave version 10.14.6. Could be an OS issue.

It is an OS issue :)
I'll do the right thing and add a proper .png instead of the symbol. Will update the change as soon as I can. Thanks again.

nwai90:
> > > @nwai90 Interesting.. I tested it on every possible browser / resolution and it seemed ok.
> > > What OS are you on? I will check on another PC and will figure it out. Thanks for noticing.
> > 
> > 
> > I'm using Chrome on MacOS Mojave version 10.14.6. Could be an OS issue.
> 
> It is an OS issue :)
> I'll do the right thing and add a proper .png instead of the symbol. Will update the change as soon as I can. Thanks again.

Thanks @chosta for the fix!

chosta:
@nwai90 can you please check if it looks ok on your side. It turned out to be quite a nasty thing - just my luck :)
The ascii cancel symbol could not be centered on my OS (ubuntu) using css unless I add a hack / fix and since I can't test on MacOs (soon I'll be able to) the best solution I found was to add a custom .png asset and hope it would look ok on all platforms / browsers / reslolutions.
Let me know if there is still a problem.

nwai90:
> @nwai90 can you please check if it looks ok on your side. It turned out to be quite a nasty thing - just my luck :)
> The ascii cancel symbol could not be centered on my OS (ubuntu) using css unless I add a hack / fix and since I can't test on MacOs (soon I'll be able to) the best solution I found was to add a custom .png asset and hope it would look ok on all platforms / browsers / reslolutions.
> Let me know if there is still a problem.

@chosta It's looking great now, thanks! 👍 

Screenshot 2020-07-31 at 10 26 04 PM

PR #1096, opened 2020-08-01
Enable clone finished game and auto prune

vincentneko:
- Finished games will now appear in the clonable game list
- Auto pruning of unfinished games. When a game is finished, a query will be launched to purge other unfinished games that are older than 10 days.
@bafolts this should reduce the memory footprint and maybe avoid to manually purge the table. 10 days is maybe too high, we can change this if needed.

### DIFF
--- a/src/database/PostgreSQL.ts
+++ b/src/database/PostgreSQL.ts
@@ -35,7 +35,7 @@ export class PostgreSQL implements IDatabase {
 
     getClonableGames( cb:(err: any, allGames:Array<IGameData>)=> void) {
         const allGames:Array<IGameData> = [];
-        const sql = "SELECT distinct game_id game_id, players players FROM games WHERE status = 'running' and save_id = 0 order by game_id asc";
+        const sql = "SELECT distinct game_id game_id, players players FROM games WHERE save_id = 0 order by game_id asc";
 
         this.client.query(sql, (err, res) => {
             if (err) {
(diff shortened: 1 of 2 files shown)


src/database/PostgreSQL.ts
```diff
@@ -130,8 +130,14 @@ export class PostgreSQL implements IDatabase {
                     console.error("PostgreSQL:cleanSaves2", err2);
                     throw err2;
                 }
-            });
+            });           
         });
+        // Purge unfinished games older than 10 days
+        this.client.query("DELETE FROM games WHERE created_time > now() - interval '10 days' and status = 'running'", function(err: { message: any; }) {
```
bafolts, line 136: Should this be `created_time < now() - interval '10 days'`? I assume this would delete games _newer_ than 10 days. I will have to double check this.

vincentneko: Yes you are right ! Will update asap

vincentneko:
> Don't merge this until the postgres query is fixed. As written this will delete games newer than 10 days old.

Updated.
I've also added an index on created_time column (@bafolts I think you already created this one)

src/database/PostgreSQL.ts
```diff
@@ -30,7 +30,13 @@ export class PostgreSQL implements IDatabase {
                 throw err;
             }
             console.log("Connected to PostgreSQL");
-        });        
+        });
+        this.client.query("CREATE INDEX IF NOT EXISTS games_i2 on games(created_time )", (err) => {
+            if (err) {
+                throw err;
+            }
+            console.log("Connected to PostgreSQL");
```
alrusdi, line 38: Is it message really needed?

vincentneko: > Is it message really needed?

Nope, just removed it

src/database/PostgreSQL.ts
```diff
@@ -130,8 +133,14 @@ export class PostgreSQL implements IDatabase {
                     console.error("PostgreSQL:cleanSaves2", err2);
                     throw err2;
                 }
-            });
+            });           
         });
+        // Purge unfinished games older than 10 days
+        this.client.query("DELETE FROM games WHERE created_time < now() - interval '10 days' and status = 'running'", function(err: { message: any; }) {
```
bafolts, line 139: I will sanity check this query. I wouldn't assume these queries are running in series. The call to `this.client.query` is asynchronous and these queries may end up running after the function returns. I don't think this is a problem for our use case here but if we continue to add queries it could lead to difficult to track down bugs.

PR #1101, opened 2020-08-03
Select how to pay cleanup

vincentneko:
This one is quite big and could have some impact on the game...
It also removes more than 300 lines of code !

All calls to `SelectHowToPay `have been removed and will now use the `addSelectHowToPayInterrupt `mechanism which is consistent with Helion pay with heat.
This is also true for the research phase.

Some extra testing would be appreciated along with the review !

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -1019,19 +1019,21 @@ export class Game implements ILoadable<SerializedGame, Game> {
 
     public playerIsFinishedWithResearchPhase(player: Player): void {
       this.researchedPlayers.add(player.id);
-      if (this.allPlayersHaveFinishedResearch() && this.interrupts.length === 0) {
-        this.gotoActionPhase();
-      } else if (this.allPlayersHaveFinishedResearch() && this.interrupts.length > 0 ) {
+      if (this.allPlayersHaveFinishedResearch()) {
+        if (this.interrupts.length === 0) {
+          this.gotoActionPhase();
+        } else {
         // Resolve research interrupt (Helion player)
-        let interrupt = this.interrupts.shift();
-        if (interrupt !== undefined && interrupt.playerInput !== undefined) {
-          if (interrupt.beforeAction !== undefined) {
-            interrupt.beforeAction();
-          }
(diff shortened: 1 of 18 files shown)


src/Game.ts
```diff
@@ -1019,8 +1019,20 @@ export class Game implements ILoadable<SerializedGame, Game> {
 
     public playerIsFinishedWithResearchPhase(player: Player): void {
       this.researchedPlayers.add(player.id);
-      if (this.allPlayersHaveFinishedResearch()) {
+      if (this.allPlayersHaveFinishedResearch() && this.interrupts.length === 0) {
```
bafolts, line 1022: I would write this:

```
if (this.allPlayersHaveFinishedResearch()) {
    if (this.interrupts.length === 0) {
        this.gotoActionPhase();
    } else {
        // other logic
    }
}
```

src/Player.ts
```diff
@@ -996,28 +996,11 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
...
         const purchasedCardsCost = this.cardCost * selectedCards.length;
-
-        if (htp.heat > 0 && this.canUseHeatAsMegaCredits) {
-          this.heat -= Math.min(htp.heat, purchasedCardsCost); // prevent overspending
-          let megaCreditsToRemove = purchasedCardsCost - htp.heat;
-          if (megaCreditsToRemove > 0) this.megaCredits -= megaCreditsToRemove;
-        } else {
-          this.megaCredits -= purchasedCardsCost;
-        }  
+        game.addSelectHowToPayInterrupt(this, purchasedCardsCost, false, false, "Select how to pay " + purchasedCardsCost + " for purchasing " + selectedCards.length + " card(s)");
```
nwai90, line 1003: @vincentneko Could we skip this step if `selectedCards.length` is 0? Came across this scenario while testing with Helion and it seems unnecessary

Screenshot 2020-08-07 at 11 08 06 PM

src/Game.ts
vincentneko: updated

src/Player.ts
vincentneko: updated

Issue #1107, opened 2020-08-03
Negative megacredits on Hellas


etiennepetit:
I had 18 megacredits, playing on Hellas. Then, I played Lava Flows (cost: 18M) on the southmost tile with ocean and -6M as placement bonus. I was actually allowed to place the tile there and ended up with -6M, after placing the ocean.

vincentneko:
EcologicalZone may be subject to the same bug.

nwai90:
Closed as resolved in https://github.com/bafolts/terraforming-mars/pull/1108

PR #1108, opened 2020-08-03
Fix: Hellas bonus ocean space should not be selectable if player can't afford

nwai90:
**Ref:** https://github.com/bafolts/terraforming-mars/issues/1107

Edge cases (e.g. additional MC obtained **after** placing the bonus ocean) are not accounted for in this PR.

### DIFF
--- a/src/Board.ts
+++ b/src/Board.ts
@@ -6,7 +6,7 @@ import { SpaceBonus } from "./SpaceBonus";
 import { TileType } from "./TileType";
 import { Game } from "./Game";
 import { BoardName } from "./BoardName";
-import { MAX_OCEAN_TILES } from "./constants";
+import { MAX_OCEAN_TILES, HELLAS_BONUS_OCEAN_COST } from "./constants";
 
 export abstract class Space implements ISpace {
     constructor(public id: string, public spaceType: SpaceType, public bonus: Array<SpaceBonus>, public x: number, public y: number ) {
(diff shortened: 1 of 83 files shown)


src/Board.ts
```diff
@@ -138,8 +141,17 @@ export abstract class Board {
         );
     }    
 
-    public getSpaces(spaceType: SpaceType): Array<ISpace> {
-        return this.spaces.filter((space) => space.spaceType === spaceType);
+    public getSpaces(spaceType: SpaceType, player: Player, game: Game): Array<ISpace> {
+        let spaces = this.spaces.filter((space) => space.spaceType === spaceType);
+
+        if (spaceType ===  SpaceType.LAND
+            && this.getOceansOnBoard() < MAX_OCEAN_TILES
```
vincentneko, line 148: Not sure why there is this condition about oceans.

vincentneko:
I think most logic is already there for special ocean bonus for Hellas map.

Hellas map features methods like 

`getAvailableSpacesForCity `or `getAvailableSpacesOnLand` that will filter the space if the player cannot afford 6.

nwai90:
@vincentneko I simplified the code. It's correct that the 6 MC must be paid to place a tile there even if there are no oceans left: https://boardgamegeek.com/thread/1842647/article/27873537#27873537

Hence, we don't need to check for max oceans as well.

PR #1111, opened 2020-08-03
Add simple log showing tile placement's coordinates

nwai90:
Example:

Screenshot 2020-08-04 at 12 28 19 AM

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -1445,7 +1445,7 @@ export class Game implements ILoadable<SerializedGame, Game> {
       }
       
       this.tilePlaced(space);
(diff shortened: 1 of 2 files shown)


vincentneko:
It would be great to know the tile type (city, ocean, etc...)

nwai90:
@vincentneko I updated so that it will show greenery, city, ocean or special tile. Would that be sufficient?

Screenshot 2020-08-04 at 5 24 00 PM

chosta:
How does the coords system work exactly?
I've done these two placements on Hellas and they don't seem correct to me:
bug-coors?
with log:
bug-coords2

PR #1123, opened 2020-08-06
Global params UI

ssimeonoff:
globals

Rework a bit of the global params:
- Venus/Oxygen/Temp active values styling to look as a slider input
- Venus/Oxygen/Temp active styling is ONLY for the current values (as in the real game there is a cube only in the current values)
- Oceans countdown value (as in the real game starting from 9)
- Oceans count when reach 0 to show a checkmark instead of 0
oceans

# TODO
I want 
`completed` to be displayed when there no oceans left, but it is not rendered as html when added to the html template. 
Can someone help me with it? (currently the word "Done" is displayed when oceans=0)


chosta:
Great design!

ssimeonoff:
Forgot to mention that I snuck in also: res counter minor UI change to make it blend in with the card more (it will overlap with text on some cards, but we all know the cards already)
res

PR #1124, opened 2020-08-07
Add game results

vincentneko:
Create game_results table in postgreSQL database only.
A row will be inserted for each finished game.

Tested locally on my laptop with a local postgreSQL database with a solo game.

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -1175,7 +1175,11 @@ export class Game implements ILoadable<SerializedGame, Game> {
       let scores:  Array<Array<String>> = [];
       this.players.forEach(player => {
         let result : Array<String> = [];
-        result.push(player.corporationCard!.name);
+        let corponame: String = "";
+        if (player.corporationCard !== undefined) {
+          corponame = player.corporationCard.name;
+        }
+        result.push(corponame);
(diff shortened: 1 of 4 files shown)


src/Game.ts
```diff
@@ -1170,6 +1177,16 @@ export class Game implements ILoadable<SerializedGame, Game> {
 
     private gotoEndGame(): void {
       Database.getInstance().cleanSaves(this.id, this.lastSaveId);
+      let scores:  Array<Score> = [];
+      this.players.forEach(player => {
+        let corponame: String = "";
```
nwai90, line 1182: Minor suggestion: It's clearer to rewrite as

```
const corporation = player.corporationCard as CorporationCard;
scores.push({corporation: corporation.name, playerScore: player.victoryPointsBreakdown.total });
```

Is there ever a situation where `player.corporationCard` will be `undefined`?

vincentneko: yes, `player.corporationCard` is undefined before player chooses it.
I've applied your suggestion, I agree it is much clearer.

vincentneko: Shite, it makes the tests fail.

PR #1135, opened 2020-08-11
Award winner name is always visible

alrusdi:


nwai90:
@alrusdi One question about this, what if the award winner is tied? Showing a single name could be misleading

chosta:
Also, make sure you set a limited length of the username field (with overflow: ellipsis or smth. like that) as player names like "Viswanathan Anand" might break the design :)
here is an untested suggestion:
```
.long-name {
  display: inline-block;
  max-width: 50px;
  white-space: nowrap;
  overflow: hidden;
  text-overflow: ellipsis;
}
```

alrusdi:
> The current logic shows funder name rather than winner name.
> 
> Screenshot 2020-08-11 at 5 43 51 PM

Oh! You are right. I will not do this task then (logic and output is complicated - it is much easier to just click and see what is happening inside).

Issue #1130, opened 2020-08-11
"Smaller cards" selection problem 


chosta:
@ssimeonoff 
It seems that "smaller cards" options breaks the new (and awesome) card highlighting

![smaller-cards](https://user-images.githubusercontent.com/836179/89868440-7a925880-dbbb-11ea-9ab5-b8158ac08010.png)

Also I noticed that in the initial phase the board is showing NaN for oceans.

![nan](https://user-images.githubusercontent.com/836179/89868578-b3323200-dbbb-11ea-85be-e9fea50648bf.png)

chosta:
Also, I think the yellow that we used to mark the already terraformed global parameters looked good. It gives the feeling that you've made progress. The new marker is awesome but I'd suggest we add the coloring back :)

ssimeonoff:
ugh, yes, please remove the 
`transform` property for selected cards.

chosta:
We can close this one

PR #1137, opened 2020-08-11
Minor fixes after ui update

chosta:
#1130
+Fixed oceans NaN in initial phase
+Smaller cards scale properly on selection now

### DIFF
--- a/src/components/Board.ts
+++ b/src/components/Board.ts
@@ -1,6 +1,6 @@
 
 import Vue from "vue";
-import * as constants from '../constants';
+import * as constants from "../constants";
 import { BoardSpace } from "./BoardSpace";
 import { SpaceModel } from "../models/SpaceModel";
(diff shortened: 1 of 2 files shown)


src/components/Board.ts
```diff
@@ -107,7 +107,8 @@ export const Board = Vue.component("board", {
             }
         },
         oceansValue: function() {
-            const leftover = constants.MAX_OCEAN_TILES - this.oceans_count;
+            const oceans_count = this.oceans_count || 0;
+            const leftover = constants.MAX_OCEAN_TILES - oceans_count;
             if (leftover == 0) {
```
alrusdi, line 112: this comparison operator is also should be ===

chosta: Yes, I think prettier will solve these issues, I've been checking how it behaves in this project and will soon add a suggestion on some .esprettier and .eslint settings

PR #1154, opened 2020-08-13
Fix search for life

chosta:
+ fixed home screen stars background
+ card can't be acted once contact is made
+ easter egg upon contact with the aliens


src/components/Card.ts
```diff
@@ -109,10 +110,14 @@ export const Card = Vue.component("card", {
...
             }
             return cssClass;
+        },
+        lifeFound: function (card: CardModel): boolean {
+            return card.name === CardName.SEARCH_FOR_LIFE && card.resources !== undefined && card.resources > 0
         }
     },
     template: `
     <div :class="getCardCssClass(card)">
+        <img v-if="lifeFound(card)" class="little-green-men" src="assets/martian.png" />
```
nwai90, line 120: 👽

src/cards/SearchForLife.ts
```diff
@@ -31,7 +31,7 @@ export class SearchForLife implements IActionCard, IProjectCard, IResourceCard {
         return undefined;
     }
     public canAct(player: Player): boolean {
-        return player.canAfford(1);
+        return player.canAfford(1) && this.resourceCount === 0;
```
nwai90, line 34: This is incorrect, you can always choose to spend 1 MC to reveal the top card even if there are already resources on SearchForLife. It's just that the card gives 3 VP no matter how many resources are on it, as long as it has at least 1 resource.

tests/cards/SearchForLife.spec.ts
```diff
@@ -18,11 +18,17 @@ describe("SearchForLife", function () {
         expect(card.canAct(player)).to.eq(false);
     });
 
+    it("Can't act if found", function () {
```
nwai90, line 21: Related to above comment, we should remove or modify this test. The card action can always be used as long as the player can afford it.

chosta:
On 1ce4523:
+ I've added an `actionOrderWeight` property to the cards with the idea that some of them appear in the action queue before others
+ In the case of Search For Life we start with a high weigth - meaning the card will appear with high priority in the actions queue
+ Once a microbe tag is found the card will get a 0 action weight so that it goes back to the queue of action cards

We can use this approach for any card and should you approve of this concept I'll refactor the sorting algorithm in `Player.ts`

vincentneko:
I have a mixed feeling with this one (actionOrderWeight)
First I'm not sure I would like the actions order to change dynamically during a game, cards are displayed in a row and a card being displayed first or last has a little benefit IMHO.
Second, we have to decide which action is better than another, and I think this should be left to the player to decide.
And one last thing, which is very edgy : I might want to try to add another resource on search for life in order to boost the Excentric award.

nwai90:
> I have a mixed feeling with this one (actionOrderWeight)
> First I'm not sure I would like the actions order to change dynamically during a game, cards are displayed in a row and a card being displayed first or last has a little benefit IMHO.
> Second, we have to decide which action is better than another, and I think this should be left to the player to decide.
> And one last thing, which is very edgy : I might want to try to add another resource on search for life in order to boost the Excentric award.

@vincentneko I see your point. Agree that we don't necessarily have to add actionOrderWeight here.

This could complicate things down the road as more cases are introduced where action weights are readjusted. It's probably a reasonable tradeoff that all actions are shown in the same order everytime, and the player decides which action(s) to take.

chosta:
Thanks for the feedback. I do agree with what you both say but let me try to address your concerns one last time before scrapping the weighted action functionality.
+ The whole thing stems from a player feedback (discord) that had a game with early tap success and got a bit frustrated that he saw the action on top of the queue until the end of the game. I think it made sense and we decided to do something about it:

+ The `actionOrderWeight` property was never meant to be used for one card only. I could have made an conditional check for SEARCH_FOR_LIFE and put it at the end of the queue if res > 0 but I saw an opportunity there. It could be added to any other card but it is optional. For me playing AI central or any free card draw action is almost 100% the first preferrable action on my turn. We can easily do this now by just adding a high value of actionWeight to the card and _et voila_ :)

+ I want to stress on the `optional` part. Unless we want to specifically make a card move up or down the queue, we don't have to add the property to every single card.

+ Having the option to set card order priority would allow interesting solutions like allowing the player to set his / her card order (maybe drag&drop UI for action cards to set personal card order)

+ I don't feel that adding the property would increase complexity that much as it probably would be used on just a couple of cards and even if something goes wrong it will not break the app but might just change the action order.

+ Finally, all we are doing here is helping the player take turns faster and reduce frustration from unwanted behavior.

I hope these points changed your mind but I trust your judgement so I'm ok with ditching the property just say the word :)

vincentneko:
If we could put another option in preference panel like "allow reorder of blue card actions", I would be ok with this.
The problem I see is that `actionOrderWeight ` is server side and will happen for all players, whether they like this feature or not.

chosta:
Ok. I'll scrap the actionWeight and just leave the iconography fix here.
For the future we can have the option for any player to choose whether they want to rearrange actions but this would probably need a different architecture. Let's leave it for now and think about it later.

vincentneko:
> Ok. I'll scrap the actionWeight and just leave the iconography fix here.
> For the future we can have the option for any player to choose whether they want to rearrange actions but this would probably need a different architecture. Let's leave it for now and think about it later.

Please also leave the Easter egg, it's great !

PR #1160, opened 2020-08-14
Unify colony trade fleets images

alrusdi:


chosta:
I believe there is a bug where the color of the colony ship is displayed properly before taking off but not when put on the colony tile. They all look the same (3 different players)
@alrusdi 
ships

alrusdi:
@chosta i will check it

Issue #1161, opened 2020-08-14
Luna Governor Interactions


duckmammal:
Bringing back up the issues from #1148 because the ticket was closed before being considered.

Earth Office (and probably Terractor too) should have 2x 3 MC = 6 MC total discount for Luna Governor, making it be free, not 1 MC.

As @nwai90 points out in a reply https://github.com/bafolts/terraforming-mars/issues/1148#issuecomment-673341445

> @vincentneko I saw that the designer gave a thumbs up to agree that Teractor and Earth Office should give discount per Earth tag here: https://boardgamegeek.com/thread/2173302/article/31568994#31568994
> 
> It's also discussed in a few other threads like this one: https://boardgamegeek.com/thread/2070625/two-tags-and-discounts
> 
> This was a non-issue before Luna Governor was released; it seems that the wordings of Teractor and Earth Office are actually misleading as they are intended to count per Earth tag.

Having seen these rulings before, and agreeing with @nwai90 , I'd like to reopen the case, hopefully to not have it summarily dismissed this time.

nwai90:
Resolved in https://github.com/bafolts/terraforming-mars/pull/1164

PR #1164, opened 2020-08-15
Fix discount for Earth Office and Teractor

nwai90:
**Ref:** https://github.com/bafolts/terraforming-mars/issues/1161

Luna Governor should be discounted by 6 MC as confirmed by the game designer.

We could also consider to reword the effect text on Earth Office and Teractor to state the correct effect clearly, since this is not possible on the physical print version.


vincentneko:
We could use the same wording as Cheung Shing MARS or VALLEY TRUST
>> (Effect: When you play an Earth tag, you pay 3 MC less for it.)

PR #1175, opened 2020-08-16
Fix:when a player with corp MiningGuild places an ocean act as the WG

jijiqw:
Fix:
1、when a player with corp MiningGuild places an ocean act as the WG (when he is the first palyer), his STEEL prod increase one by mistake
2、when create a new game , the players can't get  promo corps for choose  even if the "Promos" option has been checked

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -253,6 +253,11 @@ export class Game implements ILoadable<SerializedGame, Game> {
         corporationCards.push(...ALL_TURMOIL_CORPORATIONS.map((cf) => new cf.factory()));
       }  
 
+      // Add Promo stuff
+      if (this.promoCardsOption) {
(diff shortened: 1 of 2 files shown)


src/Game.ts
```diff
@@ -253,6 +253,11 @@ export class Game implements ILoadable<SerializedGame, Game> {
         corporationCards.push(...ALL_TURMOIL_CORPORATIONS.map((cf) => new cf.factory()));
       }  
 
+      // Add Promo stuff
+      if (this.promoCardsOption) {
+        corporationCards.push(...ALL_PROMO_CORPORATIONS.map((cf) => new cf.factory()));
+      }  
+
```
vincentneko, line 260: Not sure why we have this change here, it does not seems to be related with the issue.

jijiqw: It's another one, I have updated the comment now

PR #1183, opened 2020-08-18
Implement player TR handicap option

nwai90:
Ref: https://github.com/bafolts/terraforming-mars/issues/1139

Screenshot 2020-08-18 at 4 45 47 PM


src/Player.ts
```diff
@@ -107,7 +107,8 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
     constructor(
         public name: string,
         public color: Color,
-        public beginner: boolean) {
+        public beginner: boolean,
+        public handicap: number) {
```
alrusdi, line 111: Isn't it possible to provide here an default value to avoid that huge amount of tests changes?

nwai90: Ah yes, overlooked this. Will add, it's much easier that way

Issue #1074, opened 2020-07-28
Feature request: Random Milestones/Awards toggle


mloballer:
Hello, my game group is wondering if it would be possible when choosing Random Milestones/Awards to be able to toggle some milestones off. We find that in our game Tycoon is never ever going to get claimed in a 3 player game. Also, I am sure many may disagree with me, but I hate planner, although I can see how in some rare strategies it's viable. But either way would be nice to remove some of them from the random pool, thanks!

kevinb9n:
I would suggest closing this and, if it is ever important enough to anyone, consider reopening #1697 instead.

PR #1184, opened 2020-08-18
Colonies filter for game creation form

alrusdi:


src/components/ColoniesFilter.ts
```diff
@@ -0,0 +1,59 @@
...
+    new Io(),
+    new Luna(),
+    new Miranda(),
+    new Pluto(),
+    new Titan(),
+    new Triton()
+
+];
+
+export const ColoniesFilter = Vue.component("corporations-filter", {
```
nwai90, line 31: Should this be `"colonies-filter"`? Similarly for line 34 below (`customColoniesList`)

chosta:
This looks awesome. Can we use the same approach here:
https://github.com/bafolts/terraforming-mars/issues/1074

I think it might be good since we are doing the full-customization anyway.

src/components/ColoniesFilter.ts
alrusdi: Fixed. Thank you!

Issue #1125, opened 2020-08-08
Feature request: Toggle "Click on player name to see details" on and off


Elonka:
First, thanks for all the work on this, my group is loving it!

Right now, we can click on a player's name to see their cards/score, and then have to "X" to close out that screen. Would it be possible to make this a toggle, so click once on their name to open another player's info, and then click again on their name to close it? Thanks!

chosta:
Yes, this one has been bugging me for a while so I guees its time has come :)

PR #1188, opened 2020-08-18
Toggle player on name click

chosta:
#1125

In order to test it properly, you need 3+ players.
+ You will be able to "pin" the player of your choice by clicking on the [p] after their name (this will become an icon should we approve this change)
+ You will still be able to have all players displayed one next to the other if you click their names
+ Pinning will close all active player windows

Let me know what you think


src/components/PlayerHome.ts
```diff
@@ -117,8 +123,11 @@ export const PlayerHome = Vue.component("player-home", {
                         <span class="help_tip" v-i18n>(click on player name to see details)</span>
                     </h2>
                     <div class="player_item" v-for="(p, idx) in player.players" v-trim-whitespace>
-                        <div class="player_name_cont" :class="getPlayerCssForTurnOrder(p, true)">
-                            <span class="player_number">{{ idx+1 }}.</span><a v-on:click.prevent="showPlayerDetails(p)" class="player_name" :class="getPlayerCssForTurnOrder(p, false)" href="#">{{ p.name }}</a>
+                        <div style="display:inline">
```
alrusdi, line 143: Please do not use inline styles. Add new css class here

src/components/PlayerHome.ts
```diff
@@ -52,11 +52,28 @@ export const PlayerHome = Vue.component("player-home", {
 
             (this.$root as any).setVisibilityState("other_player_" + player.id, true);
         },
-        hasPinIcon: (player: PlayerModel): boolean => {
-            console.log(player);
-            console.log(player.id)
-            console.log(window.location.href, "HREF");
-            return window.location.href.split("=")[1] !== player.id;
+        pinPlayer: function(player: PlayerModel) {
+            //if (player.id === this.player.id) return;
```
alrusdi, line 56: Please remove all commented code. We have git history for it

src/styles/player_home.less
```diff
@@ -186,6 +186,10 @@
             text-shadow:1px 1px 1px #222, -1px -1px 1px #666;
         }
     }
+    .player_pin {
+        cursor: pointer;
+        // background: url("assets/pin.png");
```
alrusdi, line 191: Why background is commented? Shouldn't we provide width and height according to image size?

alrusdi (changes requested):

src/components/PlayerHome.ts
chosta: Maybe I wasn't clear enough on this. The pull request needs approval on the functionality. Or basically "is this a good idea". The code will be "clean" in the final version.

src/styles/player_home.less
chosta: I wanted to make sure that we want to do this before doing the icon as it takes some time that I don't want to waste if we scrap the idea :)

src/components/PlayerHome.ts
chosta: will do, thanks

chosta:
@vincentneko @alrusdi
Sorry for pushing a uncomplete code sample but I wanted to make sure we are ok with this functionality. 
I'll try to send complete pull requests next time.
Just let me know if you are ok with this UX improvement and I'll fix the rest.
Thanks.

chosta:
@nwai90 @alrusdi 
I need some help with Vue. Here is what we have so far:
+ having a pin for each player (not the one on focus)
+ the pin would hide all other players' data, except for the one that is pinned
- once pinned the pin icon needs to change to the one that shows that the player is pinned (it needs another click at the moment to work)
I could not manage to make it work. The data.pinnedPlayerId needs to change on each click (which it does only half of the time). I am missing something here. I've done mostly React where the state is changed in one direction, so I assume vue has the same thing with data (state) and it could be changed in a method (which I try to do in methods.pinPlayer(). Please, helpzzz :)

alrusdi:
> @nwai90 @alrusdi
> I need some help with Vue. Here is what we have so far:
> 
>     * having a pin for each player (not the one on focus)
> 
>     * the pin would hide all other players' data, except for the one that is pinned
> 
> 
>     * once pinned the pin icon needs to change to the one that shows that the player is pinned (it needs another click at the moment to work)
>       I could not manage to make it work. The data.pinnedPlayerId needs to change on each click (which it does only half of the time). I am missing something here. I've done mostly React where the state is changed in one direction, so I assume vue has the same thing with data (state) and it could be changed in a method (which I try to do in methods.pinPlayer(). Please, helpzzz :)

I think the game refresh is rewriting your state. You can use a global state like it is done in the Milestone component

alrusdi:
> Would it be ok If I merge master onto my branch and resolve the conflicts?

Yes! It is must be done

chosta:
@nwai90 please, review if you have the time. I can't assign reviewers

alrusdi:
@chosta everything seems to be alright now. Can you please show us a demo somewhere?

PR #1201, opened 2020-08-23
Hover show description over res box

chosta:
+ fix for some cards that could not have their description read because of res box

### DIFF
--- a/src/HTML_data.ts
+++ b/src/HTML_data.ts
@@ -536,7 +536,7 @@ export const HTML_DATA: Map<string, string> =
         <div class="tag tag1 tag-microbe "></div>
         <div class="tag tag2 tag-science "></div>
         <div class="card-number">034</div>
-        <div class="content ">
+        <div class="content hover-hide-res">
             <div class="requirements ">4% O2</div>
             <div class="red-arrow "></div> <div class="microbe resource "></div><br>
             OR <div class="microbe resource "></div><div class="microbe resource "></div> <div class="red-arrow "></div> <div class="tile temperature-tile "></div>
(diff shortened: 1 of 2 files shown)


alrusdi:
Why do we need it?
We already have something like this when "Magnify card descriptions on hover" option is enabled
sc3
sc1
sc2

chosta:
We were discussing this in discord. A couple of users complained its impossible to read on some cards and we decided that a hover - toggle view would be a good fix. 
I see the benefit of the option for "magnify description" but that is 3 clicks too many to check just a single card. You need to click on options -> select magnify -> read the text -> deselect magnify.
I'm ok with cancelling this change but I also don't see a problem letting it stay as it is set for only ~15 cards and the effect is immediate and intuitive.
@vincentneko @ssimeonoff let us know what you think

nwai90:
If it's only ~15 cards and we know the list of cards, another option might be to rework the css for those cards so that their descriptions are not blocked by the resource counter. That would be visually friendlier for new players as well / those playing with a new expansion for the first few times.

chosta:
@nwai90 It would be very hard for Pharmacy Union for example:
https://ssimeonoff.github.io/cards-list#corp40

nwai90:
> @nwai90 It would be very hard for Pharmacy Union for example:
> https://ssimeonoff.github.io/cards-list#corp40

True, this is an anomaly 🙃

chosta:
So what do we do with this one :) ?

alrusdi (approved):
let's merge it. My workaround described above is not working for now.

PR #1206, opened 2020-08-24
issue_1195

vincentneko:
- Add Floaters and Microbes to CanAfford method (fix Reds policy bug)
- Fix a bug where MC could go negative

Edge case : If I'm able to play a card because of microbes or floaters (with dirigibles) with Reds ruling but I did not spend them and therefore I lack MC after to pay Red, TR will not be increased and MC will not be spent.


nwai90 (approved):
Looks great @vincentneko

PR #1210, opened 2020-08-26
Initial draft v2

vincentneko:
- Rework initial draft, following Kyshantry guidelines (see below)
- Remove Initial Draft rounds option and parameter.

start with a hand of 5, pick one, pass the remainder left
you receive 4, pick one, pass the remainder left
lather rinse and repeat
til that pack is gone
then start with another hand of 5, pick one, pass right
lather rinse repeat

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -967,7 +967,12 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
             game.dealer.dealCard(true),
             game.dealer.dealCard(true));
         } else {
-          cards = this.dealtProjectCards;
+          cards.push(
+            game.dealer.dealCard(true),
+            game.dealer.dealCard(true),
+            game.dealer.dealCard(true),
+            game.dealer.dealCard(true),            
+            game.dealer.dealCard(true));
         }
       } else { cards = passedCards}      
 
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -957,6 +957,12 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
       });
     }
 
+    public dealCards(game: Game, quantity: number, cards: Array<IProjectCard>) {
+      for (let i = 0; i < quantity; i++) {
+        cards.push(game.dealer.dealCard());
+      }      
+    }
+
     public runDraftPhase(initialDraft: boolean, game: Game, playerName: String, passedCards?: Array<IProjectCard>): void {
       let cards: Array<IProjectCard> = [];
       if (passedCards === undefined) {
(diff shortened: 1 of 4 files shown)


alrusdi:
My group and me needs full draft by one card very much :-( Can we have it as option somehow?

alrusdi (approved):

vincentneko:
> My group and me needs full draft by one card very much :-( Can we have it as option somehow?

My bad, I thought this implementation on initial draft would be better than the current one.
Having options for initial draft seems very complicated to handle at the moment, I think we may have to choose one for now.

I will setup a pool in discord to get feedback.

alrusdi:
> > My group and me needs full draft by one card very much :-( Can we have it as option somehow?
> 
> My bad, I thought this implementation on initial draft would be better than the current one.
> Having options for initial draft seems very complicated to handle at the moment, I think we may have to choose one for now.
> 
> I will setup a pool in discord to get feedback.

No-no! I've missread description. Your variant seems to be very good and it is better than current.

nwai90 (approved):

src/Player.ts
```diff
@@ -967,7 +967,12 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
             game.dealer.dealCard(true),
             game.dealer.dealCard(true));
         } else {
-          cards = this.dealtProjectCards;
+          cards.push(
+            game.dealer.dealCard(true),
```
nwai90, line 971: These blocks appear at 3 different places in the code for this file; it may be cleaner to extract them into a method e.g. `public dealCardsToPlayer(game: Game, quantity: number)`

vincentneko: @nwai90  I've updated, please let me know if it's ok.

src/Player.ts
```diff
@@ -957,6 +957,12 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
       });
     }
 
+    public dealCards(game: Game, quantity: number, cards: Array<IProjectCard>) {
+      for (let i = 0; i < quantity; i++) {
+        cards.push(game.dealer.dealCard());
```
nwai90, line 962: @vincentneko LGTM. Are there any implications if `isResearchPhase` is false (by default) here? In the original (removed) code it was `game.dealer.dealCard(true)`

vincentneko: @nwai90 Good catch ! Will update.

rzulian:
@vincentneko did you created the pool on the drafting variants? What's the results. Your new draft 5+5 is nice, but the draft 10 is more "complete", as you can see 34 cards out of 40, instead of 28 in your version.

Issue #1211, opened 2020-08-26
Bug: Trade Envoy and Trading Colony needs option for Europa


vsrisuknimit:
In certain situation, energy production is preferable to plant production. For example, someone might need energy production to claim generalist or to trade in future turns. 

This issue was previously closed without resolution here: https://github.com/bafolts/terraforming-mars/issues/630

vincentneko:
Thanks @nwai90 for fixing. Merged.

PR #1212, opened 2020-08-26
Make colony track increase from card effects optional for Europa

nwai90:
**Ref:** https://github.com/bafolts/terraforming-mars/issues/1211

- Add option to increase colony track for Europa or do nothing
- Add log to show actual number of increased steps (1 or 2)

Screenshot 2020-08-26 at 11 50 24 PM
Screenshot 2020-08-26 at 11 50 44 PM


vincentneko (approved):
This looks great !

nwai90:
@vincentneko I pushed another change to handle the edge case where the number of steps that can be increased is less than the player's `colonyTradeOffset`. This should ensure the game log shows the correct number of steps increased.

PR #1213, opened 2020-08-26
1. Fix :  Mons Insurance

jijiqw:
1. Fix :  monsInsuranceOwner  needn't  pay 3mc when someone removes his own resources (red-bordered resource icons cards ,such as  fish  )

2. Improvement: merge duplicate targets of  flooding

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -255,7 +255,7 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
       }      
 
       // Mons Insurance hook
-      if (game !== undefined && game.monsInsuranceOwner !== undefined && amount < 0 && fromPlayer !== undefined) {
+      if (game !== undefined && game.monsInsuranceOwner !== undefined && amount < 0 && fromPlayer !== undefined && fromPlayer !== this) {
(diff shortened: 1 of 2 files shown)


src/Player.ts
```diff
@@ -255,7 +255,7 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
       }      
 
       // Mons Insurance hook
-      if (game !== undefined && game.monsInsuranceOwner !== undefined && amount < 0 && fromPlayer !== undefined) {
+      if (game !== undefined && game.monsInsuranceOwner !== undefined && amount < 0 && fromPlayer !== undefined && fromPlayer !== this) {
```
nwai90, line 258: 👍  It would be great to add a test for this to guard against regressions

nwai90 (approved):

src/Player.ts
jijiqw: Sorry but I have no idea how to add a test  for this.

nwai90: 👌 I'll add one later after it's merged

PR #1227, opened 2020-08-29
Side Menu css changes

ssimeonoff:
Untitled

## Changes
- the title "Terraforming Mars" is part from the side menu
- Player name and colour are displayed
- (additional: star background is fixed with fading to black)

## TODO ( Need help)
Don't know how to expose "player" so the colour and player name are hardcoded at the moment.
Also if Solo, #303030 as background and "Solo Challenge" as text to be used.


vincentneko:
@ssimeonoff Great design ! I've added name and color as props to Preferences

vincentneko:
Options are truncated and not left aligned anymore :

image

chosta:
I know this PR will probably go through but I must comment for my own sake. Horizontal text is a no-go for me. UI needs to be helping users understand the scenery in an effortless manner. Tilting your head or making your brain arrange the letters in an unusual way (unless we are using hieroglyphs its from left to right or right to left) is an extra effort + brain computing power. It might seem like a minor thing but these things stack up and make clutter. I think we are going in the right direction for the UI so that is why I suggest to discuss:
+ get rid of TERRAFORMING MARS
+ player name = player color

I'm just trying to make us think a bit more and avoid the trap of visual appeal > UX :)

ssimeonoff:
Eye candy > functionality :)
A no-text option can be:
91647542-2bbe3d00-ea5c-11ea-8273-b76a585fab72

chosta:
Clean, lean + I bet the sidebar shortcuts would be used more on the top left instead of bot left :)

ssimeonoff:
ok, current "final" design: 
Untitled

vincentneko (changes requested):
The preference panel is not aligned to the left, can you please revert to previous state ?

ssimeonoff:
Did I break something? It looks fine to me. 
I didn't change the panel css (only added back its width, because I removed it accident in a previous commit)

vincentneko:
> Did I break something? It looks fine to me.
> I didn't change the panel css (only added back its width, because I removed it accident in a previous commit)

Strange. Did you see my screenshot ? The lines in preference panel are centered instead of left aligned...

ssimeonoff:
screenshot? where?

vincentneko:
> screenshot? where?

image

ssimeonoff:
oh (I was thinking for the menu itself).

vincentneko (approved):
great, thanks for fixing

Issue #1228, opened 2020-08-30
Bug: Volcanic site labels missing


vsrisuknimit:
The label for Volcanic sites are missing.

![image](https://user-images.githubusercontent.com/14239220/91648492-98bedc00-ea36-11ea-870d-bdd1ae64085b.png)

vincentneko:
Fixed by @nwai90

PR #1229, opened 2020-08-30
Update map name references from original to tharsis

nwai90:
Ref: https://github.com/bafolts/terraforming-mars/issues/1228


vincentneko (approved):
Thanks for fixing !

Issue #1225, opened 2020-08-28
Minor feature request: hire raiders and sabotage showing how much resource to remove


vsrisuknimit:
Hire raiders and sabotage should, once a target is selected, show how many resources of that type can be removed. This is to prevent people from accidentally sabotaging titanium of someone who had none, for example.

nwai90:
Fixed and merged

PR #1230, opened 2020-08-30
Improve Hired Raiders and Sabotage

nwai90:
Ref: https://github.com/bafolts/terraforming-mars/issues/1225

- Reduce from two clicks to one by directly showing all available options with exact quantities

Screenshot 2020-08-30 at 11 23 39 AM

Screenshot 2020-08-30 at 11 22 53 AM

### DIFF
--- a/src/cards/HiredRaiders.ts
+++ b/src/cards/HiredRaiders.ts
@@ -28,7 +28,7 @@ export class HiredRaiders implements IProjectCard {
             );
         }
 
-        const availablePlayerTargets = game.getPlayers();
+        const availablePlayerTargets = game.getPlayers().filter((p) => p.name !== player.name);
         let availableActions = new OrOptions();
 
         availablePlayerTargets.forEach((target) => {
--- a/src/cards/HiredRaiders.ts
+++ b/src/cards/HiredRaiders.ts
@@ -1,11 +1,9 @@
-
 import { Game } from "../Game";
 import { Player } from "../Player";
 import { IProjectCard } from "./IProjectCard";
 import { CardType } from "./CardType";
 import { Tags } from "./Tags";
 import { OrOptions } from "../inputs/OrOptions";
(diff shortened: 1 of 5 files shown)


src/cards/HiredRaiders.ts
```diff
@@ -31,18 +28,35 @@ export class HiredRaiders implements IProjectCard {
...
-                selectedPlayer.setResource(Resources.STEEL, -2, game, player);
-                return undefined;
-            }),
-            new SelectPlayer(game.getPlayers(), "Select player to steal up to 3 mega credits", "Steal MC", (selectedPlayer: Player) => {
-                player.megaCredits += Math.min(3, selectedPlayer.megaCredits);
-                selectedPlayer.setResource(Resources.MEGACREDITS, -3, game, player);
-                return undefined;
-            })
-        );
+        const availablePlayerTargets = game.getPlayers();
```
nwai90, line 31: Can we filter out the current player here or is there a situation where it makes sense to steal/remove resources from yourself? This shouldn't trigger Mons Insurance effect, but not sure if there are other scenarios

vincentneko: I don't see any scenario where it could be beneficial to target oneself.
The card can be played even opponent has no steel nor MC so I would filter it. (and it would be consistent with Virus)

vincentneko: A very very edge case could be that nobody as steel nor MC but you want to play the card for Legend milestone...
But then you would have no MC left to claim it (except if you are Helion...)

vincentneko: The card is playable even with no target so I would filter current player.

vincentneko (approved):

src/cards/HiredRaiders.ts
nwai90: 👍 I added the filter. If there are no targets it will simply return `undefined`

PR #1243, opened 2020-09-01
Add Colony count to Tag Overview

nwai90:
Screenshot 2020-09-01 at 5 34 26 PM


vincentneko (approved):

nwai90:
@vincentneko @alrusdi  The CSS doesn't seem to work when Colonies is not selected. Any idea what the problem is?

src/styles/player_home.less
```diff
@@ -60,14 +60,17 @@
...
 
-
 .tags-grid {
     display: grid;
     grid-template-columns: max-content repeat(19, 50px);
 
     grid-row-gap: 10px;
     margin-top: 10px;
 
+    .hide-colony-count {
```
alrusdi, line 70: Shouldn't it be like:
&..hide-colony-count
?

PR #1250, opened 2020-09-02
Fix board with undo

vincentneko:
This should fix the changing board with undo.
With this I was not able to reproduce the issue.


nwai90:
👍 @vincentneko The board no longer changes when Undo action is done

nwai90 (approved):

PR #1258, opened 2020-09-06
Rework game options, take 2

vincentneko:
Previous commit was rollbacked because of a bug. This bug has been fixed.

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -118,27 +118,10 @@ export class Game implements ILoadable<SerializedGame, Game> {
     public colonyDealer: ColonyDealer | undefined = undefined;
     public pendingOceans: number = 0;
     public lastSaveId: number = 0;
-    private draftVariant: boolean;
     public soloMode: boolean = false;
-    public corporateEra: boolean = true;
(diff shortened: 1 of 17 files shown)


src/Game.ts
```diff
@@ -118,27 +118,10 @@ export class Game implements ILoadable<SerializedGame, Game> {
     public colonyDealer: ColonyDealer | undefined = undefined;
     public pendingOceans: number = 0;
     public lastSaveId: number = 0;
-    private draftVariant: boolean;
     public soloMode: boolean = false;
-    public corporateEra: boolean = true;
```
bafolts, line 123: The `gameOptions` property is not reflected on `SerializedGame`. These changes should be kept in sync with `SerializedGame`. Reading this code I am fairly sure that the database is going to have the `gameOptions` property populated and all these removed properties kept in sync. For a sanity check I would try restoring a game from the database upon a server re-start and undo operation.

bafolts (approved):

src/Game.ts
vincentneko: I've updated SerializedGame

PR #1279, opened 2020-09-10
Docker Update

jrbarronumd:
A couple tweaks to the Docker installation method.

1. Add /usr/app/db directory to Dockerfile mkdir command so it can inherit proper permissions to be accessed by 'tfm' user created by Dockerfile
2. Delete docker-compose.yml.  It references a Dockerfile that does not pull from the latest version of the game.  I don't think it is necessary to include this file at all.  I think including detailed instructions for various docker install methods is sufficient.
3. Replaced link to a newer set of Docker install instructions.

@chosta recommended I share this effort, and asked that I tag @pierrehilbert as well.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM node:lts-alpine
 
 EXPOSE 8080
 
-RUN mkdir -p /usr/src/app \
+RUN mkdir -p /usr/src/app/db \
    && addgroup -S tfm \
    && adduser -S -D -h /usr/src/app tfm tfm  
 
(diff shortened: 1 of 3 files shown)


bafolts:
This looks reasonable to me. I never use the docker stuff but the addition of `/db/` makes sense for storing the local database file.

vincentneko:
@lotooo can you look into this one please ?

lotooo:
The PR looks great.

I am just wondering if instead of deleting the `docker-compose.yml` for good it wouldn't be better to update `image: lotooo/terraforming-mars` (which is, as you mentioned, outdated -- version from the day I wrote the docker-compose.yml) with `build: .`

That way people can choose to use the simple solution of docker-compose.

(Another option would be to automatically build a docker image on each commit/tags from this repo instead of my fork)

jrbarronumd:
I'm open to updating the `docker-compose.yml` and simplifying the basic d-c portion of the instructions to utilize the file.  That seems easier than building a new image with each update?

pierrehilbert (changes requested):
I like the suggestion to keep the docker-compose.
If you add it, the rest is perfect for me.

jrbarronumd:
I'll revise the .yml and the instructions to follow suit.

alrusdi:
Closing it as duplicate

PR #1282, opened 2020-09-13
Ares initial commit, adds 11 of the 24 cards.

kberg:


src/HTML_data.ts
```diff
@@ -703,6 +725,25 @@ export const HTML_DATA: Map<string, string> =
...
+    <div class="content">
+      <div class="points points-big">1</div>
+      <div class="requirements requirements-max">max 4% O2</div>
+      <div class="production-box">
+        <div class="money production">1</div>
+      </div>
+      <div class="tile special-tile" style="font-size:30px;margin-left:20px;">&#x2642;</div>*
+      <div class="description">
+        (Oxygen must be 4% or less. Place this tile NEXT TO NO OTHER TILE. 
+          The tile grants an grants an <b>adjacency bonus</b> of  of 1M€. Increase your MC production 1 step.)
```
chosta, line 743: I think you have on extra "grants" here

src/components/CreateGameForm.ts
```diff
@@ -367,6 +371,11 @@ export const CreateGameForm = Vue.component("create-game-form", {
                                 <i class="form-icon"></i> <span v-i18n>Promos</span>&nbsp;<a href="https://github.com/bafolts/terraforming-mars/wiki/Variants#promo-cards" class="tooltip" target="_blank">&#9432;</a>
                             </label>
 
+                            <label class="form-switch">
+                                <input type="checkbox" v-model="aresOption">
+                                <i class="form-icon"></i> <span v-i18n>Ares</span>&nbsp;<a href="https://boardgamegeek.com/thread/2218211/tm-ares-fan-expansion-more-interaction-map-mars" class="tooltip" target="_blank">&#9432;</a>
```
chosta, line 376: This is just a note on closing tags. I personally prefer empty tags to be closed immediately:
`<i class="form-icon" />`
It saves space but throughout the project it is done with a closing tag so I guess consistancy is more important. It comes to a personal preference in the end but as developers we like stuff to be as concise as possible

src/cards/ICard.ts
```diff
@@ -22,6 +22,7 @@ export interface IActionCard {
     canAct: (player: Player, game: Game) => boolean; 
 }
 
+// A card that stores resources on it.
```
chosta, line 25: We avoid comments as much as possible. The variable, methods and class names should speak for themselves. In case there is something really funky (which should be avoided) we add a proper explanation.

chosta (commented):
This looks good to me. Lets see how the additional logic for the card would work. The weather events should also be fun to play around with. Please, keep the implementation steps of similar size, it is easier to review. Please, also check my comments on the env and vscode setup in discord.

kberg:
> This looks good to me. Lets see how the additional logic for the card would work. The weather events should also be fun to play around with. Please, keep the implementation steps of similar size, it is easier to review. Please, also check my comments on the env and vscode setup in discord.

Excellent, thank you so much!

I'm about to head to sleep, so I'll leave these questions and they'll either be answered or not in the morning, but these questions will let me hack on it in the morning.

What's your team practice on making minor changes to this branch?  EG, if I just add a commit that says "address chosta's comments" /shrug?  Will this all be, comprsesed into a single commit before it gets into the main branch? I'm adept enough at git to get through this, but might need a little guidance, particularly as you have preferences.

src/cards/ICard.ts
kberg: OK. While I personally strongly disagree with this approach, I'm also willing to go along - no problem. Also, I'm a bit surprised at how readable this code is.

kberg:
Actually I don't have as many questions as I thought. Heh.

nwai90:
> > This looks good to me. Lets see how the additional logic for the card would work. The weather events should also be fun to play around with. Please, keep the implementation steps of similar size, it is easier to review. Please, also check my comments on the env and vscode setup in discord.
> 
> Excellent, thank you so much!
> 
> I'm about to head to sleep, so I'll leave these questions and they'll either be answered or not in the morning, but these questions will let me hack on it in the morning.
> 
> What's your team practice on making minor changes to this branch? EG, if I just add a commit that says "address chosta's comments" /shrug? Will this all be, comprsesed into a single commit before it gets into the main branch? I'm adept enough at git to get through this, but might need a little guidance, particularly as you have preferences.

@kberg This project is using merge rather than squash for PRs; it's fine to push commits solely to address PR comments. It will be easier to review when each commit has a defined scope and descriptive title 👍

src/CardName.ts
```diff
@@ -1,445 +1,457 @@
...
-  RECRUITMENT = "Recruitment",
-  RED_TOURISM_WAVE = "Red Tourism Wave",
-  SPONSORED_MOHOLE = "Sponsored Mohole",
-  SUPPORTED_RESEARCH = "Supported Research",
-  WILDLIFE_DOME = "Wildlife Dome",
-  VOTE_OF_NO_CONFIDENCE = "Vote Of No Confidence",
-  LAW_SUIT = "Law Suit",
-  STANFORD_TORUS = "Stanford Torus",
-  SATURN_SURFING = "Saturn Surfing",
+    ACQUIRED_COMPANY = "Acquired Company",
```
nwai90, line 2: It's preferable to do massive formatting changes in their own commit / a separate PR so that it is easier to review. This should be ok for now though

kberg: That was a pure mistake on my part, it's not fixed in a subsequent commit in this PR?

src/components/CreateGameForm.ts
kberg: ack.

src/cards/ICard.ts
kberg: Removed.

src/HTML_data.ts
kberg: Fixed, thanks.

src/CardName.ts
kberg: Seems it is, so resolving, LMK if I'm mistaken.

vincentneko (approved):

PR #1288, opened 2020-09-13
Implementing most of the bonus behavior, tests for cards

kberg:
A couple of functional pieces of this need to be addressed which in subsequent changes
1: tests for the changes to Game.ts
2. Implementing cost deductions - the player needs to know they have an additional cost to place tiles next to Nuclear Zone, for instance. canAfford is too early, and addTile is too late. Probably somewhere in SelectSpace, which is going to be a New Thing I Have To Learn.
3. Turns out my AdjacencyBonus struct isn't expressive enough: BioEngineeringFacility grants an adjacency bonus of two different resource types.

Some other notes about the code:
1. Everything up to "Merge branch 'ares' into ares-bonus" is the prior PR. I don't know how to stage this PR lined up behind the last one.
2. I didn't want to add adjacency-specific bonuses to SpaceBonus, and I didn't want to repeat them. So I did what you see -- create an AresSpaceBonus. It's ... it's OK. I don't love it. I tried some weird things with enum unions but that was impossible to get past the compiler. I'm still relatively new to TS.
3. Game.ts line 1497 is where most of the interesting work is, please let me know what you think about the structure. Is Game.ts getting too large? Time for a refactor? :D
4. Adjacency is stored on the space -- when a tile is placed on a space, the space gets the adjacency bonus. This is better than my original approach which dealt with a custom callback on cards. Felt unnecessary.
5. How did I implement adjacency bonuses while maintaining some degree of code sharing between an Ares card and its base game counterpart? I did it by adding an adjacencyBonus attribute to all the base game cards, which is extended by the Ares counterpart. Just compare Capital and CapitalAres.
6. MiningArea and MiningRights have special behavior that doesn't fit into #5.


src/Game.ts
```diff
@@ -1473,6 +1474,37 @@ export class Game implements ILoadable<SerializedGame, Game> {
...
       }
 
+      let extraCosts: number = 0;
+      this.board.getAdjacentSpaces(space).forEach((adjacentSpace) => {
+        if (adjacentSpace.adjacency) {
+          extraCosts += adjacentSpace.adjacency.cost || 0;
+        }
+      });
+      if (extraCosts > 0) {
+        throw new Error("Implement making opponent pay " + extraCosts + " MC");
```
chosta: I guess this is where things get tricky. Maybe add #TODO so we don't have to wonder if this should be an actual error or you reminding yourself to implement a functionality.

src/Game.ts
```diff
@@ -1473,6 +1474,37 @@ export class Game implements ILoadable<SerializedGame, Game> {
...
+
+      // From this point forward the tile can be placed.
+
+      this.board.getAdjacentSpaces(space).forEach((adjacentSpace) => {
+        if (adjacentSpace.tile) {
+          if (adjacentSpace.tile.tileType === TileType.OCEAN) {
+            player.megaCredits += player.oceanBonus;
+          }
+
+          // Grant adjacency bonuses for cards from Ares.
```
chosta: Since it's only for Ares, you can add a conditional if (aresExtension) here. I presume we can have additional adjacency features in the future but for better readability and to make sure we don't ever go to this code block unless we are using ares. Also, whenever we want to make a fix for ares it would be really easy to find all the ares related stuff by a global search for aresExtension === true or smth. similar. That is just food for though here.

chosta:
I've skimmed most of the code and it looks reasonable but this has to be checked in more detail, probably @nwai90 can help especially with the Game.ts part. I'll try to address your concerns here:

> Everything up to "Merge branch 'ares' into ares-bonus" is the prior PR. I don't know how to stage this PR lined up behind the last one.

We have to find a better way to handle this. I feel like having the ares branch constantly merged can't be a big problem since it's not in production. So I suggest don't make a new pull request and work on a single PR branch until someone merges it to ares and then make a new one. There are commits hanging here from your previous PR and it can get confusing for people not following the progress. @alrusdi maybe you can merge ares-bonus to ares?

> I didn't want to add adjacency-specific bonuses to SpaceBonus, and I didn't want to repeat them. So I did what you see -- create an AresSpaceBonus. It's ... it's OK. I don't love it. I tried some weird things with enum unions but that was impossible to get past the compiler. I'm still relatively new to TS.

There is a way to merge enums in typescript and it's easy as const A = {...B, ...C}. It's better to have things separated in a logical way so you can keep the AresSpaceBonus and SpaceBonus separated. Adding anything to existing enums will most probably break current functionality so better be safe.

> Game.ts line 1497 is where most of the interesting work is, please let me know what you think about the structure. Is Game.ts getting too large? Time for a refactor? :D

Its big, yes. It needs refactor, yes. Is it the time to do it, maybe. What we can do is try to slowly abstract functionality and move out of the main file. No need to be done at once. So you can try and think of how to handle spacing behavior in a separate file/module and we can comment.

> Adjacency is stored on the space -- when a tile is placed on a space, the space gets the adjacency bonus. This is better than my original approach which dealt with a custom callback on cards. Felt unnecessary.

It should be on the space yes

> How did I implement adjacency bonuses while maintaining some degree of code sharing between an Ares card and its base game counterpart? I did it by adding an adjacencyBonus attribute to all the base game cards, which is extended by the Ares counterpart. Just compare Capital and CapitalAres.

This is good I think. @nwai90 can confirm

Looks like a good progress, hope I was helpful.

src/Game.ts
```diff
@@ -1469,35 +1473,91 @@ export class Game implements ILoadable<SerializedGame, Game> {
...
+              adjacentSpace.player.megaCredits += 1;
+        
+              var typeToLog: string = "";
+              if (bonus.spaceBonus) {
+                typeToLog = bonus.spaceBonus.toString();
+                for (var idx = 0; idx < bonus.units ; idx++) {
+                  this.grantSpaceBonus(player, bonus.spaceBonus);
+                }
+              } else if (bonus.aresSpaceBonus) {
+                var type: AresSpaceBonus = bonus.aresSpaceBonus;
```
nwai90, line 1517: I'd suggest to move this part to a separate handler that specifically handles Ares-related logic if possible; it will keep `Game.ts` clean.

e.g.
```
else if (bonus.aresSpaceBonus) {
    aresHandler.awardSpaceBonus(args);
}
```

src/Game.ts
```diff
@@ -1469,35 +1473,91 @@ export class Game implements ILoadable<SerializedGame, Game> {
             `Select a valid location ${space.spaceType} is not ${spaceType}`
         );
       }
+     
+      let extraCost: number = 0;
```
nwai90, line 1477: This seems to be Ares-specific logic as well (similar to below comment)

src/ISpace.ts
```diff
@@ -3,13 +3,18 @@ import { SpaceBonus } from "./SpaceBonus";
...
 import { Player } from "./Player";
+import { AdjacencyBonus } from "./cards/ares/AdjacencyBonus";
 
 export interface ISpace {
     id: string;
     spaceType: SpaceType;
     tile?: ITile;
     player?: Player;
     bonus: Array<SpaceBonus>;
+    adjacency?: {
```
nwai90, line 14: 👍

src/cards/ares/AdjacencyBonus.ts
```diff
@@ -0,0 +1,25 @@
+import { SpaceBonus } from "../../SpaceBonus";
+
+export class AdjacencyBonus {
```
nwai90, line 3: One possibility for cases like `BioEngineeringFacility` might be to use an array of `SpaceBonus`es instead of `units`

tests/cards/ares/CapitalAres.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+    }
+    player.setProduction(Resources.ENERGY,2);
+    expect(card.canPlay(player, game)).to.eq(true);
+
+    const action = card.play(player, game);
+    expect(action instanceof SelectSpace).to.eq(true);
+    expect(player.getProduction(Resources.ENERGY)).to.eq(0);
+    expect(player.getProduction(Resources.MEGACREDITS)).to.eq(5);
+
+    debugger;
```
nwai90, line 34: Let's remove this line before merge

vincentneko:
This is great !

A few comments:
- About Implementing cost deductions, this looks similar to me with bonus ocean and extra fee of 6 MC with Hellas map.
I would add some code in `Board.ts` in `getAvailableSpacesForxxx` methods to filter spaces if player cannot afford the extra fee.

- About the bonus animal for placement, you can use `Game.addResourceInterrupt` (look at Miranda for instance), it will save a few lines.

- Mohole placement bonus should be 2 heat.

src/Game.ts
kberg: I'm adding reminders here: https://gist.github.com/kberg/4bc227e403c23751559cdd5b084fa89f for the moment. But you're right, a TODO is reasonable.

src/Game.ts
kberg: Will do.

kberg:
@vincentneko, thanks for the notes, and fixed Mohole Area.

src/Game.ts
kberg: Done, WDYT?

src/Game.ts
kberg: Removed.

tests/cards/ares/CapitalAres.spec.ts
kberg: Done, whoops.

src/cards/ares/AdjacencyBonus.ts
kberg: Yeah, you're right of course. I'll fix that separately if you're OK with it.

kberg:
OK! I think this is good to go (unless you have other concerns.)

nwai90:
@kberg Great! It's much cleaner now with the Ares-specific logic in its own file. There are a couple of suggestions from @vincentneko (the first two bullet points) that can be addressed later on, but otherwise LGTM 👍

kberg:
@nwai90 yes, I'll most definitely address those first two bullet points separately, and shortly after this goes in. (And some more of the comments @chosta made.) Thanks.

kberg:
Thanks!

Issue #1295, opened 2020-09-14
Bug: influence tracker


vsrisuknimit:
In the picture shown below, red player has two influence. This is incorrect.

Normally, you can achieve a maximum of 3 influence (1 for Chairman, 1 for Party Leader of the Dominant Party, 1 for any amount of non-Leader delegates in the Dominant Party).

The player confirmed that they did not have Event Analysts.

I believe there is a bug in the third case of influence. It should only count non-Leader delegate.

![image](https://user-images.githubusercontent.com/14239220/93080423-acd02380-f65b-11ea-844c-9b796effe033.png)

nwai90:
Fixed and merged

PR #1296, opened 2020-09-14
Fix influence count

nwai90:
Ref: https://github.com/bafolts/terraforming-mars/issues/1295


pierrehilbert (commented):
Everything look great, thanks.
Just a suggestion: should we remove the delegate from the list and only display it as the party leader?

nwai90:
@pierrehilbert We could do that; it would change the current visual experience though. The current view is more intuitive IMO as it clearly shows how many delegates are needed to claim/overtake party leadership.

alrusdi (approved):

PR #1301, opened 2020-09-15
Ares bonus

kberg:
Simplify the AdjacenyBonus API, fix the AresSpaceBonus enum so it doesn't conflict with the SpaceBonus enum.

This is getting a little complicated - I'm sure the reviews will be unpleasant because some of this is a merge from base code. Perhaps we should discuss an alternative. (Anyone up for some sort of chat?)


src/ares/AresSpaceBonus.ts
```diff
@@ -1,4 +1,5 @@
 export enum AresSpaceBonus {
-  MC,
-  ANIMAL,
+  // Numbering ensures that these bonuses don't conflict with SpaceBonus's indexes.
```
chosta, line 2: Not sure why would they be in conflict. Is this connected to the problems you mentioned about the enum unions?

src/ares/AresHandler.ts
```diff
@@ -21,8 +21,9 @@ export class AresHandler {
     this.game = game;
   }
 
-  isAresSpaceBonus = (a: SpaceBonus | AresSpaceBonus) : a is AresSpaceBonus => {
+  public static isAresSpaceBonus(a: SpaceBonus | AresSpaceBonus) : a is AresSpaceBonus {
     assert(Object.values(AresSpaceBonus).length == 2);
```
chosta, line 25: I understand the purpose of this but surely there is a better way to check if an enum is of type AresSpaceBonus, or maybe I'm wrong..
Problem here is that if AresSpaceBonus changes in size, you have to === {new_number}
Just looks a bit off to me. I guess someone with more TS exp can help here @nwai90 ?

src/ares/AresSpaceBonus.ts
kberg: That's right.

src/ares/AresHandler.ts
kberg: I really tried a lot of stuff, but three hours is long enough.

I tried your const suggestion, but then the enums seemed to no longer be the original enums assignable back to their originals. I tried the X | Y, and the hasOwnProperty test but the complier complained. Anyway, this works, as evidenced by passing tests, and now it's rife for someone who hates this the most to optimize this if they like. :D

chosta: Ok, if it works (I see it does by the sanity test) - don't touch :)

kberg: Just to be clear, I don't love it either. And the tests mean yes, please feel free to touch!

kberg:
Help me on this one, please -- what's preventing this PR from being accepted?

vincentneko:
> Help me on this one, please -- what's preventing this PR from being accepted?

It's strange to me to see that this MR will push 93 commits. I would expect to see only your changes.
As it's a separate branch, let's merge it !

kberg:
Ugh, yet another rookie mistake. I'm sorry, @vincentneko. Up until last night it was fine, but then I spent a long time last night cleaning up my branches, and that included catching up to master/HEAD. >:-( Sorry, thanks.

vincentneko:
> Ugh, yet another rookie mistake. I'm sorry, @vincentneko. Up until last night it was fine, but then I spent a long time last night cleaning up my branches, and that included catching up to master/HEAD. >:-( Sorry, thanks.

no problem. I've done my share of mess with git too =)

PR #1302, opened 2020-09-15
Separating the interrupt portion of SelectResourceCard

kberg:
You'll notice from this the new SRC that it's not actually subclassing SelectCard, but using a newInstance. I'd actually prefer to subclass it but am having just a little difficulty.

Actually now that I think about it, it might be doable, but in the meantime is this an acceptable change in spirit?


kberg:
sorry, I'm gonna figure out this bit of the problem, why I keep breaking the build and still make PRs.

chosta:
@kberg just FYI, CI breaks on random flaky tests sometimes. Its not always you :)

kberg:
Thanks. No, I have finally included eslint in vscode, and that ought to help. Mea culpa.

kberg:
I'm not quite clear what's blocking this change -- it's just about people getting to it, yes? Is "Merging is blocked" relevant?

kberg:
For context, this branch https://github.com/kberg/terraforming-mars/tree/ares-cards is a big buildup of ares cards that does rely on this change. So if it's going to get dropped I'll need a new solution, hence the request for a response.

vincentneko:
> I'm not quite clear what's blocking this change -- it's just about people getting to it, yes? Is "Merging is blocked" relevant?

Hi, I just needed time to review it.
The point is I don't understand what this change is trying to achieve.
Moving `SelectResourceCard `from interrupts to inputs feels strange. 

inputs folder hold all PlayerInput siblings, `SelectResourceCard `is not a `PlayerInput `

Can we setup a chat session so maybe I can help on this part : 
"For context, this branch https://github.com/kberg/terraforming-mars/tree/ares-cards is a big buildup of ares cards that does rely on this change. So if it's going to get dropped I'll need a new solution, hence the request for a response."
Or if you have an example.

vincentneko:
In the meantime, I will improve `SelectResourceCard` so that when there is only 1 possibility, it will resolve it by itself instead of prompting only one card to select. This would simplify cards like SYMBIOTIC FUNGUS or IMPORTED NITROGEN.

kberg:
Thank you Vincent. I'm grateful for your help. I will try to reach you on discord -- maybe what I'm doing is nonsensical, and yet it seems to make sense to me. /shrug I'm happy to learn.

kberg:
They're used in these two files

https://github.com/kberg/terraforming-mars/blob/ares-cards/src/cards/ares/BioFertilizerFacility.ts
https://github.com/kberg/terraforming-mars/blob/ares-cards/src/cards/ares/BioengineeringEnclosure.ts

I think part of the problem continues to be that I don't really see the difference between an input and an interrupt - if an interrupt would work in those bits of card code, great - I had trouble testing that kind of thing.

vincentneko:
> They're used in these two files
> 
> https://github.com/kberg/terraforming-mars/blob/ares-cards/src/cards/ares/BioFertilizerFacility.ts
> https://github.com/kberg/terraforming-mars/blob/ares-cards/src/cards/ares/BioengineeringEnclosure.ts
> 
> I think part of the problem continues to be that I don't really see the difference between an input and an interrupt - if an interrupt would work in those bits of card code, great - I had trouble testing that kind of thing.

I'm glad to help.
I will have a look at this.

You can also check this method in `game` class : `addResourceInterrupt`
I will submit a PR with a improvement of it as well as another method to trigger the same interrupt:
`public addSelectResourceCardInterrupt(player: Player, count: number = 1, resourceType: ResourceType, resourceCards: Array<ICard>, title?: string): void`

PR #1304, opened 2020-09-15
Selected card css

ssimeonoff:
The current blurry box-shadow selection is barely visible on mobile. 
A label is much better. 
Instead of using .js to add/remove the label, a pseudo ::after background is used for it.

Suggested design:
Untitled

### DIFF
--- a/src/styles/cards.less
+++ b/src/styles/cards.less
@@ -68,7 +68,7 @@
 
 input[type=checkbox]:checked + .filterDiv,
 input[type=radio]:checked + .filterDiv {
-  box-shadow: 0 0 0 2px #303030, 0 0 0px 5px #c97e23;
+  box-shadow: 0 0 0 2px #303030, 0 0 0px 6px #c97e23;
 }
 
 input[type=checkbox]:checked + .filterDiv::after,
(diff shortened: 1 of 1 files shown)


alrusdi:
Amazing! @ssimeonoff can you hide the label on mouse over to be able to read the selected card text?

ssimeonoff:
@alrusdi , I get the idea, but the short answer is NO due to the UX - after the initial click/selection the expected behaviour is the label to appear, but since the mouse is not moved away yet the :hover is still active, which hides the label - it makes no sense.

I have another another idea now to try: 
- add another state combining:hover:after
- label width is small when the card is hovered (to the right side of the card)
- transition animation for the hover state

alrusdi (approved):
Let's do it latter when you will have a time. Current version is great.

FreeLikeGNU:
The thick ribbon used to display a single check mark makes it inconvenient if not sometimes impossible to see card text.  This is especially a problem when there is only one card to select an it is selected automatically, such as having only one playable project card in hand. Does a selection indicator have to block the card?  I'm sorry for this feedback being so late. RL has been painful lately.

PR #1306, opened 2020-09-16
Awards and milestones synergy

alrusdi:
Applied @ssimeonoff  's synergy matrix to choose random awards and milestones


chosta:
This looks great. I know it's hard to make a test for a function returning random output but I feel that maybe we need some sanity checks here. I'm sure you've manually tested but my gut tells me that there might be some special case where things combine in such a way that a milestone or award is not treated fairly. Just food for thought.

src/MASynergy.ts
```diff
@@ -0,0 +1,83 @@
...
+    }
+    return arr
+}
+
+function getNumbersRange(start: number, end: number): Array<number> {
+    return Array.from(Array(end + 1 - start).keys()).map(n => n + start)
+}
+
+export function getRandomMilestonesAndAwards(withVenusian: boolean = true, requiredQty: number = 5) {
+    var synergy = 0;
```
chosta, line 62: Just for the sake of code consistancy maybe we try to stick to
`let` and `const` instead of `var`

src/MASynergy.ts
```diff
@@ -0,0 +1,83 @@
...
+    var limit = 30;
+    let output: Array<number> = [];
+    while(synergy == 0 || synergy > limit) {
+        synergy = 0;
```
chosta, line 66: Isn't synergy already = 0?

chosta:
Here is the [link](https://docs.google.com/spreadsheets/d/1PIUzO9XLF-MHSrTDP9QU2VW3ce7hJNciAR-G7N-RBUA/edit#gid=330638583) for MA randomization. Reviewers might find it useful.

src/MASynergy.ts
```diff
@@ -0,0 +1,83 @@
...
+    return arr
+}
+
+function getNumbersRange(start: number, end: number): Array<number> {
+    return Array.from(Array(end + 1 - start).keys()).map(n => n + start)
+}
+
+export function getRandomMilestonesAndAwards(withVenusian: boolean = true, requiredQty: number = 5) {
+    var synergy = 0;
+    var limit = 30;
```
chosta, line 63: Maybe we should also think ahead and let the user increase the limit. Some people, might want to be able to have a more synergistic games.

alrusdi: I don't like this idea. In the future we can allow user to select exact milestones and awards.

chosta: Yes, it might be confusing as an option for people that do not understand it. Was just thinking out loud. It's good as it is.

src/MASynergy.ts
```diff
@@ -0,0 +1,83 @@
+let NAMES = ["TERRAFORMER", "MAYOR", "GARDENER", "BUILDER", "PLANNER",
```
nwai90, line 1: It seems that we are maintaining two sets of data which are really intended to reference the same things. This could be simplified as follows, so that we don't have to retype all names manually here:

```
import { ELYSIUM_AWARDS, HELLAS_AWARDS, ORIGINAL_AWARDS, VENUS_AWARDS } from "./awards/Awards";
import { ELYSIUM_MILESTONES, HELLAS_MILESTONES, ORIGINAL_MILESTONES, VENUS_MILESTONES } from "./milestones/Milestones";

const ITEMS = (ELYSIUM_MILESTONES.concat(HELLAS_MILESTONES, ORIGINAL_MILESTONES, VENUS_MILESTONES) as any).concat(ELYSIUM_AWARDS, HELLAS_AWARDS, ORIGINAL_AWARDS, VENUS_AWARDS);
```

It's a little hackish, but less prone to manual errors.

src/MASynergy.ts
```diff
@@ -0,0 +1,83 @@
...
+"GENERALIST", "SPECIALIST", "ECOLOGIST", "TYCOON", "LEGEND",
+"DIVERSIFIER", "TACTICIAN", "POLAR EXPLORER", "ENERGIZER", "RIM SETTLER",
+"HOVERLORD",
+"LANDLORD", "SCIENTIST", "BANKER", "THERMALIST", "MINER",
+"CELEBRITY", "INDUSTRIALIST", "DESERT SETTLER", "ESTATE DEALER", "BENEFACTOR",
+"CULTIVATOR", "MAGNATE", "SPACE BARON", "EXCENTRIC", "CONTRACTOR",
+"VENUPHILE"];
+
+let SYNERGIES = [
```
nwai90, line 10: Minor: I would use `const` for values like these that should never be reassigned

src/Game.ts
```diff
@@ -392,28 +393,23 @@ export class Game implements ILoadable<SerializedGame, Game> {
...
-      if (hasVenus) availableMilestones = availableMilestones.concat(VENUS_MILESTONES);
+    public setRandomMilestonesAndAwards(hasVenus: boolean, requiredQty: number) {
+      let MA_Info = getRandomMilestonesAndAwards(hasVenus, requiredQty)
 
-      availableMilestones = availableMilestones.filter((milestone) => !this.milestones.includes(milestone));
+      let availableMilestones = ELYSIUM_MILESTONES.concat(HELLAS_MILESTONES, ORIGINAL_MILESTONES, VENUS_MILESTONES);
 
-      const shuffledMilestones = availableMilestones.sort(() => 0.5 - Math.random());
-      this.milestones.push(...shuffledMilestones.slice(0, requiredQty));
+      this.milestones.push(...availableMilestones.filter(m => MA_Info.milestones.includes(m.name.toUpperCase())));
```
nwai90, line 401: If the change in `src/MASynergy.ts` is made, this can be simplified to:

```
const availableMilestones = ELYSIUM_MILESTONES.concat(HELLAS_MILESTONES, ORIGINAL_MILESTONES, VENUS_MILESTONES);
this.milestones.push(...availableMilestones.filter(m => MA_Info.milestones.includes(m)));
```

Same for `availableAwards` below

src/MASynergy.ts
alrusdi: There are names order is important. But i get an idea. Will use the names from modules

kberg:
> This looks great. I know it's hard to make a test for a function returning random output but I feel that maybe we need some sanity checks here. I'm sure you've manually tested but my gut tells me that there might be some special case where things combine in such a way that a milestone or award is not treated fairly. Just food for thought.

I've dealt with random number generation in a recent app by providing a fake seed and a fake generator which worked well enough for my limited case and also made for reportable bugs: https://github.com/kberg/trains-randomizer/blob/master/src/app/random.ts

You can do something similar, and even something more customized just to return specific entities in tests just by abstracting the RNG.

alrusdi:
Ups... Simplest test implementation pointed me to the fact that prev logic is completely incorrect.
Sum of synergies lower than 30 can be taken from awful combinations like Terraformer+Benefactor (9) + Builder+Contractor (9) + other zeros = 18
So new logic disallows synergy of any pair to be greater than 1

nwai90:
@alrusdi Thanks for implementing, looks great!

I saw chosta's earlier comment; we could introduce variable synergy threshold in future if requested. Current implementation looks fine for now 👍 

Note: There is some chance that the unit test `"Generates random milestones and awards"` may fail occasionally if either the randomised milestones or randomised awards happen to be the same for both generated games. It is extremely unlikely to happen though.

nwai90 (approved):

PR #1308, opened 2020-09-16
Add prominent generation marker

chosta:
Several players complained that the current generation is very hard to find (true). It would be improved with the new game log but this might take some time. Until then i've come up with this solution:
1. Added gen marker after the Game Log (renamed from Last actions)
2. Added cards played number in label
3. Added consistent UI background with the elements mentioned above. It is the same as the log background implying that this is dynamic data. This could change in the future but I feel it does the job for now:
gen-marker


src/components/PlayerHome.ts
```diff
@@ -146,26 +146,20 @@ export const PlayerHome = Vue.component("player-home", {
...
-                    </h2>
+                    <h2 :class="'player_color_'+ player.color"> <span v-i18n>Cards In Hand</span><span class="label-additional">{{ player.cardsInHandNbr }}</span></h2>
                     <div v-for="card in player.cardsInHand" :key="card.name" class="cardbox">
                         <card :card="card"></card>
                     </div>
                 </div>
 
                 <div class="player_home_block player_home_block--cards">
-                    <h2 :class="'player_color_'+ player.color" v-i18n>Played Cards</h2>
+                    <h2 :class="'player_color_'+ player.color"><span>Played Cards</span><span class="label-additional">{{ getPlayerCardsPlayed(player, true) }}</span></h2>
```
alrusdi, line 162: isn't v-i18n needed for span tag here?

alrusdi (approved):

src/components/PlayerHome.ts
chosta: yes

PR #1219, opened 2020-08-27
New feature: a simple implement of voice promt

jijiqw:
Add an option "voice promt" in PreferencesManager , it play an "ding dong" sound  when you change to active player. just like the notification  "It's your turn!"

PR #1312, opened 2020-09-17
Add sound for active player notification

nwai90:
First-cut attempt at cleaning up and fixing an issue in the previous PR by @jijiqw in https://github.com/bafolts/terraforming-mars/pull/1219.

This is a POC for adding a sound when it is the active player's turn. If this approach is viable we can gradually introduce more sounds that can be customised and exported as individual functions.

### DIFF
--- a/src/SoundManager.ts
+++ b/src/SoundManager.ts
@@ -1,0 +1,66 @@
+const frequencies = [
+    196.00,
+    220.00,
+    246.94,
+    261.63,
+    293.66,
+    329.63,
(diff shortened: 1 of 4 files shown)


alrusdi:
Does it work on mobile devices?

nwai90:
> Does it work on mobile devices?

Yes to some extent, only on supported browsers that have implemented it: https://caniuse.com/?search=audio-api

alrusdi (approved):

PR #1318, opened 2020-09-18
Improve addResourceInterrupt

vincentneko:
addResourceInterrupt will automatically add resource to card if there is only one card. (and put a message in the log)

Add` addSelectResourceCardInterrupt(player: Player, count: number = 1, resourceType: ResourceType, resourceCards: Array<ICard>, title?: string): void {`

for future rework and maybe to support Ares development

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -537,6 +537,15 @@ export class Game implements ILoadable<SerializedGame, Game> {
       }
     }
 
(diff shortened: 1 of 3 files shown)


nwai90 (approved):

kberg:
Thanks for this. I'm rebuilding my repo and starting again.

vincentneko:
> addSelectResourceCardInterrupt

Yes we can add a 
if (resourceCards.length === 0) return;

PR #1319, opened 2020-09-18
Allow undo action for every action of remainning player if all other players passed

alrusdi:


nwai90:
@alrusdi There is some problem with it, the active player seems to not be able to pass

Screenshot 2020-09-18 at 5 35 17 PM

alrusdi:
> @alrusdi There is some problem with it, the active player seems to not be able to pass
> 

Yep, logic was incorrect. Can you please take another look at it?

vincentneko:
> > @alrusdi There is some problem with it, the active player seems to not be able to pass
> 
> Yep, logic was incorrect. Can you please take another look at it?

This part of the app is very tricky. I would advise extreme caution if we make some changes.

alrusdi:
Let's reject it then

vincentneko:
> Let's reject it then

Not necessarily but maybe we should deploy on a test server

PR #1325, opened 2020-09-19
Simplify tsconfig-test with wildcards.

kberg:
TESTED=npm test passed 968 tests, which matches the prior commit.


alrusdi (approved):
Great!

PR #1335, opened 2020-09-20
Restructure Game.addTile to be executed in more logical steps.

kberg:
This will make it simpler to implement Ecological Survey and Geological Survey cards from Ares and as a side benefit increases comprehension.

I'd like to put in the main branch first as I'd like to keep that order if possible.

TESTED=all tests still pass.

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -60,7 +60,6 @@ import { LogHelper } from "./components/LogHelper";
 import { ColonyName } from "./colonies/ColonyName";
 import { getRandomMilestonesAndAwards } from "./MASynergy";
 
-
 export interface Score {
   corporation: String;
   playerScore: number;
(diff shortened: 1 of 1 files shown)


chosta:
A rare case where comments can be allowed to stay. I'd remove the numbering though as we might have shuffling in the future and we all know that code flows top to bottom :)
Also, I recommend using prettier for vs code as I see some blank lines that would automatically be removed.

kberg:
The numbers are intentional -- that method goes through five clear stages,
and it's important that there's an indication that those comments are
actually correlated. Something else such as "next phase", or, as you're not
opposed to code documenting, actually split the function into five pieces,
and then addTile looks like this:

addTile {
  validatePlacement();
  collectFees();
  setupForBonuses();
  placeTile();
  collectBonuses();
}

chosta:
Makes sense

kberg:
I'm wondering if someone could accept this change.

nwai90 (approved):

kberg:
gasp thank you so much.

>

PR #1340, opened 2020-09-20
Fix circular JSON bug

vincentneko:


bafolts (approved):
I am going to to wrap all the http request routes in try...catch and return a 500 if we get an error. This will prevent every game from breaking in the future when one issue like this goes out.

PR #1344, opened 2020-09-21
Added spectator link for each player

chosta:
+ added a "SPECTATE" link next to each player's name in start game screen
+ works for after the draft by making the player in view only poll "waitingFor" but not able to take actions
+ it shows drafted cards for player in view
- in inital phase cards are not seen we can fix that later with the improvement of initial phase screen

! please, do not merge to master. I'd like to have it on the `beta` branch if possible and have a couple of games just to be sure all is OK

Also -> formatting. In `waitingFor.ts`, ignore the formatting changes and inspect these lines only:
        - if (this.waitingfor === undefined) {
        + if (this.waitingfor === false) {

### DIFF
--- a/src/components/PlayerHome.ts
+++ b/src/components/PlayerHome.ts
@@ -21,7 +21,7 @@ export const hidePlayerData = (root: any, player: PlayerModel) => {
 };
 
 export const PlayerHome = Vue.component("player-home", {
-    props: ["player"],
+    props: ["player", "spectating"],
     components: {
         "board": Board,
         "card": Card,
@@ -67,6 +67,23 @@ export const PlayerHome = Vue.component("player-home", {
                 fleetsRange.push(i);
             }
             return fleetsRange;
+        },
+        getWaitingFor: function (): boolean {
+            if (this.spectating) return false;
+
+            return this.player.waitingFor || false;
+        },
+        getWaitingForClasses: function (): string {
+            let classes = [
+                "player_home_block",
+                "player_home_block--actions",
+                "nofloat",
+            ];
+            if (this.spectating) {
+                classes.push("hidden");
+            }
+
+            return classes.join(" ");
         },
     },
     mounted: function () {
@@ -123,10 +140,10 @@ export const PlayerHome = Vue.component("player-home", {
                 
                 <players-overview class="player_home_block player_home_block--players nofloat:" :player="player" v-trim-whitespace />
                  
-                <div class="player_home_block player_home_block--actions nofloat">
+                <div :class="getWaitingForClasses()">
                     <a name="actions" class="player_home_anchor"></a>
                     <h2 :class="'player_color_'+ player.color" v-i18n>Actions</h2>
-                    <waiting-for v-if="player.phase !== 'end'" :players="player.players" :player="player" :waitingfor="player.waitingFor"></waiting-for>
+                    <waiting-for v-if="player.phase !== 'end'" :players="player.players" :player="player" :waitingfor="getWaitingFor()"></waiting-for>
                 </div>
 
                 <div class="player_home_block player_home_block--hand" v-if="player.draftedCards.length > 0">
@@ -196,8 +213,8 @@ export const PlayerHome = Vue.component("player-home", {
                 </div>
 
                 <h2 :class="'player_color_'+ player.color" v-i18n>Select initial cards:</h2>
-
-                <waiting-for v-if="player.phase !== 'end'" :players="player.players" :player="player" :waitingfor="player.waitingFor"></waiting-for>
+                
+                <waiting-for v-if="player.phase !== 'end'" :players="player.players" :player="player" :waitingfor="getWaitingFor()"></waiting-for>
 
                 <h2 :class="'player_color_'+ player.color" v-i18n>Game details</h2>
 
--- a/index.html
+++ b/index.html
@@ -24,6 +24,7 @@
                     v-if="screen === 'player-home'"
                     :player="player"
                     :key="playerkey"
+                    :spectating="spectating"
                 ></player-home>
                 <game-end
                     v-if="screen === 'the-end'"
--- a/server.ts
+++ b/server.ts
@@ -49,7 +49,10 @@ const styles = fs.readFileSync("styles.css");
 const games: Map<string, Game> = new Map<string, Game>();
 const playersToGame: Map<string, Game> = new Map<string, Game>();
 
-function processRequest(req: http.IncomingMessage, res: http.ServerResponse): void {
+function processRequest(
+    req: http.IncomingMessage,
+    res: http.ServerResponse
+): void {
     if (req.url !== undefined) {
         if (req.method === "GET") {
             if (req.url.replace(/\?.*$/, "").startsWith("/games-overview")) {
@@ -65,6 +68,7 @@ function processRequest(req: http.IncomingMessage, res: http.ServerResponse): vo
                 req.url.startsWith("/solo") ||
                 req.url.startsWith("/game?id=") ||
                 req.url.startsWith("/player?id=") ||
+                req.url.startsWith("/spectate?id=") ||
                 req.url.startsWith("/the-end?id=") ||
                 req.url.startsWith("/load")
             ) {
@@ -655,7 +659,7 @@ function getWaitingFor(
         max: undefined,
         microbes: undefined,
         floaters: undefined,
-        coloniesModel: undefined
+        coloniesModel: undefined,
     };
     switch (waitingFor.inputType) {
         case PlayerInputTypes.AND_OPTIONS:
@@ -691,7 +695,7 @@ function getWaitingFor(
             break;
         case PlayerInputTypes.SELECT_COLONY:
             result.coloniesModel = (waitingFor as SelectColony).coloniesModel;
-            break;            
+            break;
         case PlayerInputTypes.SELECT_HOW_TO_PAY:
             result.amount = (waitingFor as SelectHowToPay).amount;
             result.canUseSteel = (waitingFor as SelectHowToPay).canUseSteel;
--- a/src/components/App.ts
+++ b/src/components/App.ts
@@ -11,6 +11,7 @@ export const mainAppSettings = {
     "data": {
         screen: "empty",
         playerkey: 0,
+        spectating: false,
         componentsVisibility: {},
         game: {
             players: [],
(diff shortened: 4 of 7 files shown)


chosta:
@vincentneko @nwai90

vincentmoreaudba:
I see a potential security issue. Giving your player id to another person and with changing the url, he can easily access your private page.

chosta:
> I see a potential security issue. Giving your player id to another person and with changing the url, he can easily access your private page.

Well spotted. This is kind of an experimental feature at the moment and we want to mainly use it for admin work. They way observing was done so far was to just watch over any player id, which needed constant refreshing.
We can easily encrypt the player_id in the spectator link and read it in the backend in the next iteration.

vincentneko:
Using the same player id for both pages is a major security issue. It's very easy to change the url and access player main page...

alrusdi:
> Using the same player id for both pages is a major security issue. It's very easy to change the url and access player main page...

Even the player_id of other players in json response from server is a security issue.

src/components/PlayerHome.ts
```diff
@@ -68,6 +68,23 @@ export const PlayerHome = Vue.component("player-home", {
             }
             return fleetsRange;
         },
+        getWaitingFor: function (): boolean {
+            if (this.spectating) return false;
```
alrusdi, line 72: "Waiting for" data must be exluded from server response

chosta:
If we have to be technical - any user with basic dev tools knowledge can remove the "hidden" class from the waitingFor element and have access to the player's actions. The question here is not if this is secure enough - it's not (And as @alrusdi noted there is a bigger issue in the response) - its whether we want to follow this direction or have something else done. At the moment with this functionality any admin that creates a game can observe any player. I find it useful for the upcoming tournament. We can improve security later. 
On another note, I enjoy watching games (sometimes even more than playing them) so this could be a nice thing to have.
I already have ideas how to protect the id or the player, so let me know.
Also if someone merged this into `beta` we can all try a quick run and discuss.

src/components/PlayerHome.ts
chosta: In general or in this case?

alrusdi: In this case. Spectator is not allowed to see waiting for even in server json response. It is not enough to just hide in UI

alrusdi:
> I already have ideas how to protect the id or the player, so let me know.

We need to protect all private endpoints by some sort of auth. All the limitations of what observer can see or what he can do must be limited on server side, not client side

src/components/PlayerHome.ts
chosta: Ok, let me see what I can do

chosta:
> > I already have ideas how to protect the id or the player, so let me know.
> 
> We need to protect all private endpoints by some sort of auth. All the limitations of what observer can see or what he can do must be limited on server side, not client side

That is true but this requires additional work on security of the whole app. Maybe with the remote login that we want to implement, this would be possible.
So I am asking this question:
Are we ok with this as a non-secure feature (at this moment the app is not 100% secure anyway) that can be later improved to have proper endpoint auth or you want to scrap this?
If we close this PR the way to observe a game would again include constant refreshing and seeing other players' actions.
The value of this PR is in the utility of being able to observe a game without wanting to sabotage it. We already put the trust in the players' hands by sending the admin link to all when we create games in discord. Let me know.

alrusdi:
I am positive about spectator mode.
But i would like to see it as a separate page/template. This page should use separate api calls to get info from server side. On server side it can use the same functions as for actual players but data must be explicitly filtered.

chosta:
> I am positive about spectator mode.
> But i would like to see it as a separate page/template. This page should use separate api calls to get info from server side. On server side it can use the same functions as for actual players but data must be explicitly filtered.

This was my original idea but I decided I could do it faster with this approach. Let's hear some other opinions and can decide how to approach this.

bafolts:
The only concern I had with this was similar to @vincentneko with displaying the player id. We most likely do expose this id in some of our JSON responses. Eventually we may want to consider having a public id that exposes information you could see while playing with a player in the same room and a private id that exposes information only that player would see if you were in the same room with them such as cards in their hand. The spectator link should get the public player id if we went this route. Any JSON should also have the public id.

chosta:
Thanks for the comments. I think we just close this one and do it the correct way with a separate module that reuses PlayerHome blocks but does not allow for foul play with player ids and hidden blocks. I'd like to ask for one last thing before closing though. If we remove the "SPECTATE" link from the game page, there will be no opportunity for people to watch a game unless they know that replacing `/spectate` with `/player` would spectate the game. I just really like to be able to do that even with the security concerns. Sorry if I repeat myself but there are people that are observing games even now by just opening the player's link and refreshing like crazy. With this a game observation would be more enjoyable and it would be used mostly by admins.

alrusdi:
If we will work with safe public_player_id then there no reason to remove the "SPECTATE" link from the game page

kberg:
The only concern I had with this was similar to @vincentneko with displaying the player id. We most likely do expose this id in some of our JSON responses. Eventually we may want to consider having a public id that exposes information you could see while playing with a player in the same room and a private id that exposes information only that player would see if you were in the same room with them such as cards in their hand. The spectator link should get the public player id if we went this route. Any JSON should also have the public id.



> The only concern I had with this was similar to @vincentneko with displaying the player id. We most likely do expose this id in some of our JSON responses. Eventually we may want to consider having a public id that exposes information you could see while playing with a player in the same room and a private id that exposes information only that player would see if you were in the same room with them such as cards in their hand. The spectator link should get the public player id if we went this route. Any JSON should also have the public id.

Funny! I was just thinking about coming here to say the same thing. Either secure-hash the current player's ID into a public ID, or just randomly generate a second one and, like @bafolts says, store it along.

kberg:
In fact I think an implementation is simpler: keep all the existing code, the current player's ID is no longer a secret. Add a new ID that represents the web URL, and store that as the map to the player's ID.

chosta:
@kberg I think just hashing the player id to make a public one would be the best solution. It could be used to obfuscate any player id outside of the backend and db. I'll work on that this week and take a review / feedback gladly.

chosta:
So, I got to the point where there are some options for this, maybe you can help choose the correct path:
1) I know @bafolts prefers not to use external libs so what node has in store is the 'crypto' module. This should work in theory but we have to store the content of the encrypted function in the db (iv + content, this is the recommended way):
```import * as crypto from "crypto";

const algorithm = "aes-256-ctr";
const secretKey = "vOVH6sdmpNWjRRIqCc7rdxs01lwHzfr3";
const iv = crypto.randomBytes(16);

export const encrypt = (text: string) => {
    const cipher = crypto.createCipheriv(algorithm, secretKey, iv);
    const encrypted = Buffer.concat([cipher.update(text), cipher.final()]);

    return {
        iv: iv.toString("hex"),
        content: encrypted.toString("hex"),
    };
};

export const decrypt = (hash: any) => {
    const decipher = crypto.createDecipheriv(
        algorithm,
        secretKey,
        Buffer.from(hash.iv, "hex")
    );

    const decrpyted = Buffer.concat([
        decipher.update(Buffer.from(hash.content, "hex")),
        decipher.final(),
    ]);

    return decrpyted.toString();
};```

2) Second option is to use [public-id](https://www.npmjs.com/package/public-id) pretty straightforward if you ask me but an external dependency.

3) Generate a second id a.k.a. spectator id and use it for specator mode only.

Let me know what you guys think about it.

kberg:
Now that I've done some minor DB stuff I still think a second random
spectator ID would be simpler than creating a new secret key, etc. If I get
around to it, I'll take a stab at that myself. Or, pile on to your proposal.

>

alrusdi:
I vote for additional public player_id too

vincentneko:
We need to agree on what we want to do here. I don't like the idea of another people being able to see the cards I have in hand.
If we go in this direction, only a single spectator link per game is needed, and we can advertise games in progress from another page in the app with the links.

Generating another playerId is the way to go imo.

chosta:
I'm going to wait for the merge from #1434 and see how to work on top of it.
@vincentneko I'm also OK for you to take over as I feel you have more in-depth knowledge on the server side. Just a couple of notes:
Vincent mentioned that he doesn't want his cards in hand to be seen. That is true for the case where a spectator link is accessable to the public. But consider the following two situations:
+ Someone wants to teach the game to a new player. Being able to see the cards in hand would be perfect for explaining and discussing options with the player.
+ We have a tournament a spectator is streaming a game. He/she can be commenting on the game and the choices that the player might make. Think about Heartstone tournaments, how commentators are discussing possible next moves

So we can have two modes: public and commentator.

Let me know what you think.

vincentneko:
> I'm going to wait for the merge from #1434 and see how to work on top of it.
> @vincentneko I'm also OK for you to take over as I feel you have more in-depth knowledge on the server side. Just a couple of notes:
> Vincent mentioned that he doesn't want his cards in hand to be seen. That is true for the case where a spectator link is accessable to the public. But consider the following two situations:
> 
> * Someone wants to teach the game to a new player. Being able to see the cards in hand would be perfect for explaining and discussing options with the player.
> * We have a tournament a spectator is streaming a game. He/she can be commenting on the game and the choices that the player might make. Think about Heartstone tournaments, how commentators are discussing possible next moves
> 
> So we can have two modes: public and commentator.
> 
> Let me know what you think.

Yes that makes sense too. Public mode will be easier // more useful to implement first imo.
And for commentator you could then share your private link, I don't see any drawbacks doing this.

kberg:
So, what's the point of a per-player spectator link if the cards are
hidden? If that's the case (which I agree with) you can make a single
spectator link for the game.

chosta:
> So, what's the point of a per-player spectator link if the cards are hidden? If that's the case (which I agree with) you can make a single spectator link for the game.
The value is that you can spectate every action that is happening on the board. It's like you are observing the game Log without having to frantically refreshing the page since when it's "your" turn it does not refresh until an action is taken by the player in view. You have to try and spectate a game to see what I mean.

kberg:
Gotcha.

bafolts:
Waiting on review for https://github.com/bafolts/terraforming-mars/pull/1428 which adds a separate route for pulling game logs. With this route a spectator wouldn't have to refresh the entire page to view logs and could potentially continually query the game logs route with perhaps a separate UI for this game logs route.

chosta:
I'm closing this as it is. I've revisited your comments and I strongly feel that a separate spectator page with a different UI should be the correct way to do it. Once I'm done refactoring the Card component, I'll come back to it (unless someone else does it before that). Thanks for the great feedback!

PR #1345, opened 2020-09-21
Reveal milestones, awards and board by default

chosta:
+ reveal milestones, awards and board by default 

The `Award.ts` and `Milestone.ts` just got two comments removed and some formatting.


nwai90 (approved):
LGTM @chosta . If the awards and milestones are closed by the user, they will be revealed upon page refresh but it's not an issue imo

PR #1327, opened 2020-09-19
Big push of Ares cards

kberg:
All the cards are defined. All the tiles except hazards are defined. There are empty test suites for all the cards.

Most of the remaining work is listed in TODOs which I will now be working from.
(Though details about oceans and hazards hasn't been marked as TODOs, and will be a significant amount of work in AresHandler.)

PR #1352, opened 2020-09-22
Hazards

kberg:
I am sending this out in order to specifically get feedback on the behavior I have defined: a tile which can be replaced with another tile on top. It makes for some peculiarities in Board.ts, which, if you don't want, I'll need to know, because the same behavior is gong to apply to oceans, which I am about to start.


kberg:
Sorry! Dropped switched branch.

kberg:
Note that this depends on the cards2 integration, so when that goes in this will be smaller.

src/Board.ts
```diff
@@ -219,7 +228,9 @@ export abstract class Board {
     }
 
     protected canPlaceTile(space: ISpace): boolean {
-        return space !== undefined && space.tile === undefined && space instanceof Land;
+        return space !== undefined &&
+            (space.tile === undefined || (space.tile.hazard || true)) &&
```
nwai90, line 232: @kberg `(space.tile.hazard || true)` looks strange here. This expression always returns `true`. Do you mean to use `===`?

src/TileType.ts
```diff
@@ -18,5 +18,16 @@ export enum TileType {
 
     DEIMOS_DOWN,
     GREAT_DAM,
-    MAGNETIC_FIELD_GENERATORS
+    MAGNETIC_FIELD_GENERATORS,
+
+    BIOFERTILIZIER_FACILITY,
```
nwai90, line 23: Minor: Typo? `BIOFERTILIZER_FACILITY`

src/components/BoardSpace.ts
```diff
@@ -17,11 +17,35 @@ let tileTypeToCssClass = new Map<TileType, string>([
     [TileType.RESTRICTED_AREA, "restricted_area"],
     [TileType.DEIMOS_DOWN, "deimos_down"],
     [TileType.GREAT_DAM, "great_dam"],
-    [TileType.MAGNETIC_FIELD_GENERATORS, "magnetic_field_generators"]
+    [TileType.MAGNETIC_FIELD_GENERATORS, "magnetic_field_generators"],
+    [TileType.BIOFERTILIZIER_FACILITY, "biofertilizer_facility"],
```
nwai90, line 21: It may be better to move these to `tileTypeToCssClassAresOverride` as they are specific to `Ares` expansion, even though they do not replace existing base game tiles

src/Game.ts
```diff
@@ -1305,6 +1307,9 @@ export class Game implements ILoadable<SerializedGame, Game> {
       if (this.oxygenLevel === 8 || (steps === 2 && this.oxygenLevel === 9)) {
         return this.increaseTemperature(player, 1, isWorldGov);
       }
+      if (this.gameOptions.aresExtension) {
```
nwai90, line 1310: 👍 This is great!

nwai90:
@kberg This PR includes some recent changes like `SelectColony`. Could you rebase against `origin/master` so that it's easier to review?

src/Board.ts
kberg: Maybe? It's been years since I did a heavy Javascript project and truishness eludes me. That expression is clearer if you look at the negative case:

You may NOT place a tile if the space has a tile. Unless it's a hazard tile, in which you may place over hazard tiles -- you'll just need to pay for them later.

if (space.tile.hazard === undefined):
  canPlace = true
if (space.tile.hazard ===  true)
  canPlace = true.
if (space.tile.hazard === false)
  canPlace = false.

src/TileType.ts
kberg: /shamecube fixing.

src/components/BoardSpace.ts
kberg: Yeah. Yeah, I thought about that.  I chose this approach as it matches with DEIMOS_DOWN, GREAT_DAM, MAGNETIC_FIELD_GENERATORS, which are also expansion tiles.

If you go that route, then overrides will get complicated -- especially if you start to add multiple expansions. Anyway, I think this is the right thing, sincerely. You say "may" so you're either being kind or unsure. Either way, let me know!

src/Game.ts
kberg: Thanks!

kberg:
@nwai90 wrote 
> @kberg This PR includes some recent changes like SelectColony. Could you rebase against origin/master so that it's easier to review?

Actually I think a better approach would be: accept #1328 which is a merge from bafolts/origin/master to bafolts/ares/master, just through my repo since I can't do that myself. (In fact, isn't that just a no-brainer?)

Then accept #1327, #1335. They're smaller and more focused. There's a good chance I'll have to deal with a conflict merge or two, but nothing I can't handle.

kberg:
Thanks for the comments, though, this is encouraging!

src/Board.ts
nwai90: @kberg I see. Would this be clearer?

```
if (space === undefined) return false;
if (!(space instanceof Land)) return false;

const spaceIsUnoccupied : boolean = space.tile === undefined;
const spaceHasHazard : boolean = space.tile.hazard === true;

return spaceIsUnoccupied || spaceHasHazard;
```

nwai90: About the first condition:

```
if (space.tile.hazard === undefined):
canPlace = true
```

Is there any scenario where you can place a tile on a space that already has a tile which is not a hazard? If not, the above change should handle all cases

kberg: Yes, but it's complicated: the additional case is when you upgrade oceans. However in that case it's controlled by a card, and so while strictly speaking board.canPlaceTile(space) will be false, those cards won't use this method to check.

vincentneko:
There is a lot of conflicts with this one, is it still relevant after the other pushes to ares branch ?

kberg:
Not relevant. I'll close it - sorry for wasting your time

PR #1353, opened 2020-09-22
Fix trade bug

vincentneko:


chosta:
@vincentneko looks good to me. I'm just wandering how did it occur. Was it due to the recent trade colonies update?

pierrehilbert:
The problem is that selected the colony and paying for the trade are managed separately. Before, a colony was selected by default so the problem didn't occurs.

alrusdi (approved):

pierrehilbert (approved):
You fixed the issue by moving the paiement part into the select colony.
It seems to be perfect for me.

PR #1361, opened 2020-09-22
Added variables.less and mixins.less and setup inital values

chosta:
I started doing some work on the new Log visuals and decided to cleanup something that has been bothering me first:

+ added `variables.less` - a file holding values such as player colors, default action color, warning color, etc.
+ added `mixins.less` - similar to variables but for more complex combinations (like color and opacity)
+ started replacing some of the css values with their variable representation

Note: Should we accept this change we can try to use and replace color, shrift, etc. with variables whenever possible. I'd also suggest we do spacing in increment of 4px / 8px / 16px.

Reasoning: Control over the app outlook with a single source of truth (similar to theming).

Food for though: We have been discussing the change of colors in discord to a different palette. We could easily do that but I guess we have to stick to the original colors and look for rgb values close to them. The biggest concern is black that needs special attention (it is grey apart from the player marker). So my question is - do we ever want to escape the set of colors named "red", "yellow", "green" etc, and replace them with player "one", "two",  "three" and allow for any rgb values for backgrounds, borders and text shadows?

@alrusdi @ssimeonoff @nwai90

### DIFF
--- a/src/styles/common.less
+++ b/src/styles/common.less
@@ -2,11 +2,18 @@
 @import (inline) "../../node_modules/spectre.css/dist/spectre-exp.min.css";
 @import (inline) "../../node_modules//spectre.css/dist/spectre-icons.min.css";
 @import (inline) "../../node_modules/dialog-polyfill/dialog-polyfill.css";
+@import "./variables.less";
+@import "./mixins.less";
 
 html,
 body {
     height: 100%;
     margin: 0px;
+}
+.btn {
+    color: @text_action !important;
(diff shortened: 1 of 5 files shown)


chosta:
@alrusdi @ssimeonoff I need this reviewed to continue with my Log / Action view rework. This PR is basically organisational work that IMO should go to master so we all can start reusing variables and mixins in a smart way. Thanks.

alrusdi (approved):
It looks great.

PR #1369, opened 2020-09-23
mobile width

ssimeonoff:
Just making the mobile width to be enough to fit 4 cards.
A good side-effect fix is that 6th milestone/award will also be on the same row.
Another is that 2 colonies fits now on the same row (it was one....no comment)
Untitled


chosta:
Sounds promising. Let me play around with it tomorrow and will review.

alrusdi (approved):
Great!

PR #1377, opened 2020-09-25
[WIP] Add Concede button for 2p games

nwai90:
WIP. Need to hook things up to end the 2p game once a player has confirmed they wish to concede.

image

### DIFF
--- a/src/cards/community/CrashingGanymedeIntoMars.ts
+++ b/src/cards/community/CrashingGanymedeIntoMars.ts
@@ -1,0 +1,49 @@
+import { Tags } from "../Tags";
+import { CardName } from '../../CardName';
+import { CardType } from "../CardType";
+import { IProjectCard } from "../IProjectCard";
+import { Player } from "../../Player";
+import { Game } from "../../Game";
+import { LogMessageData } from "../../LogMessageData";
+import { LogMessageDataType } from "../../LogMessageDataType";
+import { LogMessageType } from "../../LogMessageType";
+import * as constants from "../../constants";
+import { SpaceType } from "../../SpaceType";
+
+export class CrashingGanymedeIntoMars implements IProjectCard {
+    public cost: number = 0;
+    public tags: Array<Tags> = [Tags.SPACE];
+    public name: CardName = CardName.CRASHING_GANYMEDE_INTO_MARS;
+    public cardType: CardType = CardType.EVENT;
+
+    public play(player: Player, game: Game) {
+        player.hasConceded = true;
+        this.maxOutAllGlobalParams(player, game);
+
+        game.log(
+            LogMessageType.DEFAULT,
+            "${0} crashed Ganymede into Mars!",
+            new LogMessageData(LogMessageDataType.PLAYER, player.id)
+        );
+
+        return undefined;
+    }
+
+    public getVictoryPoints() {
+        return -99;
+    }
+
+    private maxOutAllGlobalParams(player: Player, game: Game) {
+        (game as any).oxygenLevel = constants.MAX_OXYGEN_LEVEL;
+        (game as any).temperature = constants.MAX_TEMPERATURE;
+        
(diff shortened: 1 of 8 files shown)


alrusdi:
Why don't you make it a normal action? We can add general option like "confirmable" to common action interface

chosta:
I wouldn't put it there to be seen all the time. It will be bugging your brain on every action. It should be hard to reach too as "conceding" is never fun for the players that you play against, so it should be discrouraged. Maybe add it to options and keep the red color.As @alrusdi proposed, confirmable action is a good idea to have anyway. A game should be conceded only if all but one person "confirm" that they want to.

nwai90:
Agree that it should not be in the list of main actions (that was my first thought as well). IMO the first version should only be done for 2p games as the desired behaviour is less clear-cut in games with 3 or more players.

alrusdi:
@nwai90 Add one more option for action! Allow it to be "**hidden**". Then anywhere on client side implement the button which activates this hidden action

kevinb9n:
FYI, I've observed toxic arguments in the forums of every single gaming site I've been on that has a concede button. Different players have different concession etiquettes that are often incompatible.

You will quickly find one poster saying "I ban players for doing X" and another saying "I ban players for NOT doing X". It's weird how strongly people feel that their way of looking at concession is the only right way.

Having a game creation option for "allow conceding" doesn't fully address it either. Really, the only way to avoid the butthurt I've seen is for players to *communicate* with each other and indicate either "I'm willing to concede now if you want" or "I'd prefer to concede now if it's okay", and then come to agreement.

(A game might do well to require an action from both sides before terminating the game -- first one player clicks "offer to resign" then the other clicks "accept resignation", but I would be unsurprised if you'd rather not bother with all that at this point.)

kevinb9n:
(Also, conceding CAN apply to 3p+ games with the rule that whoever concedes first accepts last place, and so forth, but these players are obligated to keep playing until N-1 players have conceded. I feel like this wouldn't really be used much.)

src/cards/community/CrashingGanymedeIntoMars.ts
```diff
@@ -0,0 +1,49 @@
...
+import { SpaceType } from "../../SpaceType";
+
+export class CrashingGanymedeIntoMars implements IProjectCard {
+    public cost: number = 0;
+    public tags: Array<Tags> = [Tags.SPACE];
+    public name: CardName = CardName.CRASHING_GANYMEDE_INTO_MARS;
+    public cardType: CardType = CardType.EVENT;
+
+    public play(player: Player, game: Game) {
+        player.hasConceded = true;
```
bafolts, line 20: I agree with @kevinb9n that this feature may do more harm than good.

That being said if this is going to be implemented I would avoid needing a one off `hasConceded` property stored on `Player`. If you were to call `Game::gotoEndGame` that should do what the card is intending and automatically end the game. This will be the first situation I am aware of where we would instantly go to the end of the game. If any player has a defined `getWaitingFor()` it should be cleared as no further actions need to be taken.

https://github.com/bafolts/terraforming-mars/blob/master/src/Game.ts#L1228

You will need to skip final greenery placement and other final generation events. This card makes no mention of maxing out global parameters but seems to apply -70 victory points. As this is a community card may be better off only including this card with the community expansion enabled. If there were community expansion specific options you could include this card by default as an option. If every player had this card by default it would solve the problem of what to do on three player games. With this project card included by default would also avoid needing a concede button added. If at any time a player wants to play this project card they can play it.

https://boardgamegeek.com/image/4836682/dabluesheep

nwai90:
Great points! I'm not a fan of the current approach as well as it felt a bit forced. Would suggest not implementing it for now due to the gaming etiquette issue + it's probably too much work to implement confirmation steps from both players.

Looking at Discord activity over the past few weeks it seems that conceding is currently done by mutual agreement from both players to abandon the game - in which case it gets auto deleted after 10 days. I will close this for now until there is a stronger use case for it.

PR #1381, opened 2020-09-25
[WIP] When playing Ants, log the name of the player whose card was taken.

kberg:
I just need to either write a test or manually test it.


alrusdi (approved):

kberg:
So, turns out this is broken. Which is why I listed it as WIP.

kberg:
Rougly line 73 should be "var otherPlayer =..." and line 78 should reference otherPlayer.

alrusdi:
@kberg i'm sorry. Will revert it

Issue #1152, opened 2020-08-13
Red Policy Bug: cannot play card if no mc left to pay Reds policy despite placement bonus


vsrisuknimit:
**Bug**: The app does not allow you to play a card if you are left with too little mc to pay the Reds policy -3mc/TR, even though you could potentially get some mc back from tile placement next to oceans.

**Example scenarios**: 

(1)start with 16 mc, play a card Black Polar Dust costing 15 mc, 1 mc left, place ocean next to another ocean, gain a TR, gain 2 mc back, now has 3 mc left, pay 3mc to the Reds policy.

(2)start with 10 mc, play Nuclear Zone, 0 mc left, place the nuclear zone tile next to three oceans, gain 6 mc back, raise temp 2 steps, pay 6 mc to the Reds policy

This situation could also apply to gaining mc back from rebate: Media Group, Optimal Aerobraking. I have not tested. Those should be easier to fix though.

**Suggestion**: It would be very hard to check all possible ways to gain mc for this situation. While Reds are running, make a special case for cards that give TR AND allow tile placement. Give the user a warning of much they have to pay at the end of this action. At the end of their action, ask them to pay (with mc or heat if Helion). Also give the user of the option, "cannot pay". If the user pick this option, just force undo the action.

Source: https://boardgamegeek.com/thread/2352956/question-about-turmoilreds

nwai90:
One other possible approach is to precompute the MC gain from each tile placement spot, and use this info to decide (1) whether a card is playable / whether an action can be performed, and (2) filter out the spots that don't meet the criteria.

So a card could be playable but only have one or two valid placement spot that give sufficient bonus MC from adjacent oceans to cover any required cost/tax. In general I see three scenarios:

- Player's remaining MC + placement MC bonus < 9 for Hellas bonus ocean tile while Reds are ruling and there are oceans remaining
- Player's remaining MC + placement MC bonus < 6 for Hellas bonus ocean tile while Reds are not ruling or Turmoil expansion is not selected
- Player's remaining MC + placement MC bonus < 3 * TR steps raised for other cards/actions that place any type of tile while Reds are ruling

vsrisuknimit:
@nwai90 Things to consider with your solution:
(1) For Giant Ice Asteroid or Ice Asteroid: will you precompute all possible combos for the two ocean placement?
(2) On Hellas map, Helion could gain heat (near south pole or center ocean) from tile placement and also use that as mc to pay back the Reds.

PR #1382, opened 2020-09-25
Add Hygiea, Iapetus, Mercury, Titania, Venus and Leavitt colony tiles

nwai90:
Screenshot 2020-09-29 at 11 26 33 AM

Modified from: https://boardgamegeek.com/thread/2086154/high-orbit-rogue-colonies-8-new-fan-colony-tilesve/page/6

**Notes:**
1) `player.cardDiscount` is introduced: Set to 1 by Iapetus trade and reset to 0 at generation end
2) Building a colony on Mercury allows player to gain the trade income of any active colony in play, without triggering `beforeTrade` or `afterTrade` effects.
- Colony marker does not move, trade fleet is not spent and colony bonuses are not awarded
- `usesTradeFleet` property is introduced for this purpose; default value is `true` to ensure all existing colony behaviour is unchanged
3) `player.colonyVictoryPoints` is introduced to track Titania VP as its own score item
4) `Venus` colony is activated only when a Venus card that collects resources enters play. Reds ruling bonus is applied to Venus raise on colony placement.
5) `player.scienceTagCount` is introduced for Leavitt


src/Game.ts
```diff
@@ -914,6 +917,7 @@ export class Game implements ILoadable<SerializedGame, Game> {
       this.passedPlayers.clear();
       this.someoneHasRemovedOtherPlayersPlants = false;
       this.players.forEach((player) => {
+        player.cardDiscount = 0;
```
pierrehilbert, line 920: Add comment to be sure that we will remember why it's here.

pierrehilbert (approved):
I think comments could be helpful when we introduce new things in other class (like Player.ts) but the colony seems nicely build.
Vincent is the colony expert, he should give you the final validation

vincentneko (approved):
This is looking good to me

vincentneko:
@nwai90 can you please resolve the conflict so we can merge this one in master ?

nwai90:
@vincentneko Done

src/Player.ts
```diff
@@ -2040,12 +2046,17 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
...
           && colony.colonies.indexOf(this.id) === -1
-          && colony.isActive);      
-          if (openColonies.length > 0) {
-            standardProjects.options.push(
-                this.buildColony(game, openColonies)
-            );
-          }
+          && colony.isActive);
+
+        if (redsAreRuling && !this.canAfford(constants.BUILD_COLONY_COST + constants.REDS_RULING_POLICY_COST)) {
```
nwai90, line 1930: Just realised that this logic should be applicable to Europa colony as well. Happened in a real game which allowed the Reds tax to be avoided. Building a colony on Europa to place an ocean should be disallowed if Reds are ruling and the player cannot pay the mandatory 3 MC tax for the TR increase.

Europa is less straightforward due to the additional 2-12 bonus MC that could potentially be gained from placing adjacent to oceans, which we don't currently factor in. A quick and partial fix would be to add a check for Europa here.

I will submit a separate draft PR that attempts to implement part of the proper logic.

Ref:
https://boardgamegeek.com/thread/2352956/question-about-turmoilreds
https://github.com/bafolts/terraforming-mars/issues/1152

alrusdi: Damn Reds :-(

PR #1385, opened 2020-09-25
Further Ants repair, and minor test change.

kberg:
OK, this I tested, :D. Manually, but it's right.


kberg:
Huh, not how I thought it would work. Anyway, merge conflict should be resolved.

alrusdi (approved):

PR #1404, opened 2020-09-27
Safely parse bad log messages so as to not destroy the log.

### LINKED ISSUE #1398: Make GameLog resilient to thrown errors
I wrote some bad logging code. Which is -- look -- that happens. But when the bad logging entry makes it down to the client, and LogPanel doesn't know what to do with it, no log appears.

So while I should fix all my logs, LogPanel needs to be safer.

FYI it was parseData, this line:

            if (data.type !== undefined && data.value !== undefined) {

when data was undefined.

There are things that are safe to do. Happy to take a crack at this.

kberg:
TESTED=all manual

Fixes #1398.

I created a bad entry like so

game.log(
    LogMessageType.DEFAULT,
    "${0} played ${1} ${2}",
    new LogMessageData(LogMessageDataType.PLAYER, this.id),
    new LogMessageData(LogMessageDataType.CARD, card.name));

Some of the testing had to be done by changing data live in the
browser's debugger to verify extreme cases will preserve the log.

For example

"[{"type":1,"value":"adfe8b3a97d8"},{"type":2,"value":"Rover Construction"},null,{}]"

renders as

"BUG: Unparseable message: ${0} played ${1} ${2}, (1) adfe8b3a97d8, (2) Rover Construction, undefined, (undefined) undefined"

Which is ugly, but better than a broken log. Also, if the log entry is
that broken, you have bigger problems. But now it's fairly resilient.

### DIFF
--- a/src/components/LogPanel.ts
+++ b/src/components/LogPanel.ts
@@ -75,13 +75,36 @@ export const LogPanel = Vue.component("log-panel", {
             }
             return "";
         },
+        // Called in the event that a bad log message comes down. Does its best to return something.
+        safeMessage: function(message: LogMessage) {
+            try {
+                if (message === undefined) {
+                    return "undefined";
+                }
+                if (message.data === undefined) {
+                    return `BUG: Unparseable message: ${message.message}`;
+                }
+                var data = message.data.map(datum => {
+                    return (datum === undefined)
+                        ? "undefined"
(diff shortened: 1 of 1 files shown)


nwai90 (approved):

kberg:
AWESOME thank you.

PR #1406, opened 2020-09-27
new icons UI

ssimeonoff:
This is the big Cards UI overhaul. 

Action Points:
- replace all Unicode char and homemade images with official images
- fix minor UI bugs of misaligned elements

Changing the icons will affect others elements too. 
For example, the Player board will look something like: 
Untitled.

TODO: 
Many of the cards are still "ugly", so if you see an ugly fix - fix it please.


alrusdi:
@nwai90 @vincentneko i like it! But what do you think Jakob will say about it? Is it ok to use an official icons?

ssimeonoff:
@alrusdi,
Those are not the official ones. :)
They are fan made as well.
See - https://boardgamegeek.com/thread/2433226/article/34902171#34902171 
Made by Giulio Baldi, as it seems. 
Jacob has endorsed the project himself so we should be good (as long we don't use any copyright flavour text and images)

alrusdi (approved):
It is amazing! Let's merge it and fix remaining cards iteeratively

Issue #1408, opened 2020-09-27
Discovery of microbial life in the Martian soil is a major event, surely worthy of being mentioned in the game log :-)


kevinb9n:
I think it should output a message in the game log saying something like "Life has been found!"

nwai90:
@kevinb9n It's a minor change yes. I'll add it in as `{PLAYER} found life!`

nwai90:
Implemented

kevinb9n:
Thanks!

PR #1409, opened 2020-09-28
Add log when SearchForLife action is successful

nwai90:
Ref: https://github.com/bafolts/terraforming-mars/issues/1408


kberg (approved):
Since taking the Search for Life action is documented, this is probably all you need. Short and sweet.

alrusdi (approved):

PR #1412, opened 2020-09-28
Refactored all h2 in player home to be controlled by DynamicTitle

chosta:
Continuing the work on the Log element but a much needed refactor on a commonly repeated element done first:
+ all h2 titles are now DynamicTitle with props and can be controlled in one place
+ will try to keep this rework in small increments so this one is doing just the refactor part


src/components/PlayerHome.ts
```diff
@@ -90,7 +92,7 @@ export const PlayerHome = Vue.component("player-home", {
             
             <div v-if="player.phase === 'end'">
                 <div class="player_home_block">
-                    <h2 v-i18n>This game is over!</h2>
+                    <dynamic-title :title="'This game is over!'" :color="player.color"/>
```
alrusdi, line 95: Title syntax looks little odd.
May be it's better to use it like this:
`<dynamic-title :color="player.color">This game is over!</dynamic-title>`
?

chosta: It's more concise and we have better control this way.I prefer elements that can be contolled via the props. Since title should always be a "string" we can make a check in the props block. Ofc, who would add a Boolen there instead... It's a matter of preference I guess and mine comes from too much react and too much props validation.
I can change it to what you propose if you insist :)

alrusdi: May be pass the strings as literal then?
`<dynamic-title title="This game is over!" :color="player.color"/>`

chosta: Wait, you can do that :D
Ok, I need to do some vue tutorials

src/components/PlayerHome.ts
```diff
@@ -92,7 +92,7 @@ export const PlayerHome = Vue.component("player-home", {
             
             <div v-if="player.phase === 'end'">
                 <div class="player_home_block">
-                    <dynamic-title :title="'This game is over!'" :color="player.color"/>
+                    <dynamic-title :title="This game is over!" :color="player.color"/>
```
alrusdi, line 95: Does it work?
Should be `title="This game is over!"` not `:title="This game is over!"` 
: symbol adds dynamic expression evaluation for this variable

alrusdi:
@chosta please fix merging conflict. It is conficting with my changes aimed to show limit of generations for solo games :-( I don't sure how to fix it

chosta:
will do

src/components/PlayerHome.ts
```diff
@@ -127,7 +131,11 @@ export const PlayerHome = Vue.component("player-home", {
                 <players-overview class="player_home_block player_home_block--players nofloat:" :player="player" v-trim-whitespace />
                  
                 <div class="player_home_block player_home_block--log player_home_block--hide_log nofloat" v-if="player.gameLog.length > 0">
-                    <dynamic-title title="Game log" :color="player.color" :withAdditional="true" :additional="'generation' + player.generation" />
+                    <dynamic-title v-if="player.players.length > 1" title="Game log" :color="player.color" :withAdditional="true" :additional="'generation' + player.generation" />
+                    <h2 v-if="player.players.length === 1" :class="'player_color_'+ player.color">
```
alrusdi, line 135: You may use v-else here to avoid additional check

chosta:
@alrusdi this one is good to go IMO. Just a note on what I did there. There was no need for solo title to go into dynamic-title as too much logic needed to be passed around. This dynamic-title is going to soon be extinct anyway - due to this generation marker:
log

In case we have a solo game, the numbers will be shown from 1 to 12/14 and blink will start on 12/14. For now I kept both 'h2' and 'dynamic-title' versions and added a condition on the nr of players.

src/components/PlayerHome.ts
chosta: ok, will do now

alrusdi (approved):

PR #1414, opened 2020-09-29
Disable inputs while waiting for server response

alrusdi:


src/components/WaitingFor.ts
```diff
@@ -89,11 +89,20 @@ export const WaitingFor = Vue.component("waiting-for", {
         }
         const input = new PlayerInputFactory().getPlayerInput(createElement, this.players, this.player, this.waitingfor, (out: Array<Array<string>>) => {
             const xhr = new XMLHttpRequest();
+            const root = (this.$root as any);
+            if (root.isServerSideRequestInProgress) return;
+            let els  = window.document.querySelectorAll(".wf-root .btn-submit");
```
chosta, line 94: My issue with these selectors is that all devs need to be aware that these are the buttons that would be disabled on game load. If we somehow change the class name or use a different lib it would need a change here too. I'd rather do it for each and every button separately by getting the global "loading" state variable and disable the button / input.
This is the quick way but not sure it's optimal.
I can go through every button or element that needs to be disabled on loading and do it for you.

alrusdi: @chosta very constructive point! Go ahead, give us a PR to this branch (server-resp-wait)

chosta: Ok, let me see what I can do

chosta: Can you merge `master`. I'll give it a try now.

alrusdi: > Can you merge `master`. I'll give it a try now.

Done

alrusdi:
Closing it. @chosta will make new PR from scratch

Issue #1418, opened 2020-09-29
[Bug] Some icons missing after icon update


Maelstrom1:
Not all of the new icons seem to have imported properly....
Can't confirm that this happens to more than this, but I found this in the current heroku version
![chrome_kvWyDvFDN4](https://user-images.githubusercontent.com/20371049/94621230-f042a880-027d-11eb-9af1-bbed5123b233.jpg)
Seems to be fine on the cards list area.

vsrisuknimit:
Colony, trade fleet, and trade still use the old icons.
![image](https://user-images.githubusercontent.com/14239220/94676138-dee2b600-02e8-11eb-852d-0beafa55b691.png)

vsrisuknimit:
Venusian plants, thermophiles, freya biodomes have venus icons significantly raised.

![image](https://user-images.githubusercontent.com/14239220/94676166-e86c1e00-02e8-11eb-9e02-4940b999a6f8.png)

![image](https://user-images.githubusercontent.com/14239220/94676233-020d6580-02e9-11eb-95e0-ee58371da4da.png)

![image](https://user-images.githubusercontent.com/14239220/94676257-0afe3700-02e9-11eb-8b0f-8c5cc028d179.png)

vsrisuknimit:
These cards use black venus V instead of white on blue.

- Airscrapping expedition
- Hydrogen to Venus
- Stratopolis

![image](https://user-images.githubusercontent.com/14239220/94676308-22d5bb00-02e9-11eb-8ae1-3177d8d43498.png)

![image](https://user-images.githubusercontent.com/14239220/94676322-2701d880-02e9-11eb-82f9-6526d5c68363.png)

vsrisuknimit:
Atmoscoop icons are all messed up.

![image](https://user-images.githubusercontent.com/14239220/94676625-a2fc2080-02e9-11eb-949e-dd3c1bbcbbb5.png)

vsrisuknimit:
Rotator Impacts misses the titanium icon for the cost of action

![image](https://user-images.githubusercontent.com/14239220/94676675-b27b6980-02e9-11eb-808f-66f30f39dc89.png)

vsrisuknimit:
Asteroid resource cards have both A and asteroid icon, overlapping.

- Asteroid Hollowing
- Asteroid Deflection System
- Comet Aiming
- Directed Impactors
 
This card only has A and not asteroid icon.

- Rotator Impacts

[...]

PR #1427, opened 2020-09-30
More markup fixes by #1418 issue report

alrusdi:


src/styles/cards.less
```diff
@@ -923,7 +924,7 @@ filter: invert(1) drop-shadow(0px 1px 1px rgba(0,0,0,0.5)) drop-shadow(0px 1px 0
 .card-icon {
   position: absolute;
   margin-left: 15px;
-  margin-top: -14px;
+  margin-top: -8px;
```
ssimeonoff, line 927: make it margin-top: -10px;
At -8px the icon overlaps the resource icon and it is not vertically centered (top half of over the resource and the bottom half is over it)

alrusdi: Done

ssimeonoff (approved):

Issue #1110, opened 2020-08-03
Feature Request: Game log download


nwai90:
Request is to be able to download a text file containing all game logs at game end

SliceOfBread:
If the log contains all the choices made then the log can be used to recreate the game which can be very useful for recreating bugs. It also allows for easy Undo (delete the last choice made then re-run the log). In the versions of the Web App boards that I've done, I kept two logs: One human readable and one more machine readable.

They look like this:
### Human readable
Anna gains 40 MC;
Anna increases Steel production by 1;
Anna gains 5 Steel;
Bob gains 36 MC;
Bob gains 3 Plants;
Bob increases Plants production by 1;
Bob increases Plants production by 1;
Anna places Ocean tile (r:5 c:4 loc:3);
--Anna increases TR production by 1;
--Anna gains 2 Plants;
Bob spends 14 MC;
Bob increases Temp to -28;
--Bob increases TR production by 1;


### Same thing in machine version:
numplayers:2:Anna:Bob:
asset:1:40:MC:
production:1:1:Steel:
asset:1:5:Steel:
asset:2:36:MC:
asset:2:3:Plants:
production:2:1:Plants:
production:2:1:Plants:
tile:1:3:Ocean:
asset:2:-14:MC:
increase:2:1:temp:

bafolts:
If someone knows the format they want this log to be I can write the node code to provide the file download link via a game id.

nwai90:
The ideal scenario would be that a game can be recreated (on a physical board or otherwise) by relying purely on the game logs.

One gap at the moment is that the exact locations of placed tiles are not captured in the logs yet, e.g. `Red placed a tile on (x, y) and gained 2 plants`. I will send a PR to add a basic log to capture coordinates of placed tiles.

trilz:
Yeah, this would be great. If we could instead dump it into human readable/machine version and parse it out, it could unlock some potentially cool analysis.

kevinb9n:
I almost filed this myself today! I strongly agree that a *complete* and machine-readable game log would be an awesome thing to have.

How do we feel about JSON format? This would be easily parseable by anyone for any purpose, whatever language they wanted to program in.

I would be happy to start drafting a schema if that's okay?

Of course, it needs to record:

* all game configuration options, including the specific values chosen for any randomized options
* every single decision made by players
[...]

PR #1428, opened 2020-09-30
refactor: show all log entries

bafolts:
We have had an arbitrary limit placed on how much we log for game log entries. With #1110 there are users that want to see the entire log. I am curious to see how our memory will run on the heroku instance if we remove this limit. As long as this doesn't put us above the heroku memory limit it could be fine. I can keep an eye on the memory usage on heroku and remove if it is a problem.

### DIFF
--- a/src/routes/Route.ts
+++ b/src/routes/Route.ts
@@ -1,0 +1,31 @@
+
+import * as http from "http";
+
+export class Route {
+    public badRequest(req: http.IncomingMessage, res: http.ServerResponse): void {
+        console.warn("bad request", req.url);
+        res.writeHead(400);
+        res.write("Bad request");
+        res.end();
+    }
+    public notFound(req: http.IncomingMessage, res: http.ServerResponse): void {
+        if (!process.argv.includes("hide-not-found-warnings")) {
+            console.warn("Not found", req.method, req.url);
+        }
+        res.writeHead(404);
+        res.write("Not found");
+        res.end();
+    }
+    public internalServerError(req: http.IncomingMessage, res: http.ServerResponse, err: unknown): void {
+        console.warn("internal server error", req.url, err);
+        res.writeHead(500);
+        res.write("Internal server error " + err);
+        res.end();
+    }
+    public notAuthorized(req: http.IncomingMessage, res: http.ServerResponse): void {
+        console.warn("Not authorized", req.method, req.url);
+        res.writeHead(403);
+        res.write("Not authorized");
+        res.end();
+    }
+}
--- a/server.ts
+++ b/server.ts
@@ -64,7 +64,7 @@ function processRequest(req: http.IncomingMessage, res: http.ServerResponse): vo
                     route.notAuthorized(req, res);
                     return;
                 } else {
-                    serveApp(res);
+                    serveApp(req, res);
                 }
             } else if (
                 req.url === "/" ||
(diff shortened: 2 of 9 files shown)


chosta:
I'd suggest to create a function that gets the log by a given generation if possible.
Me and @ssimeonoff are working on an improved Log overview and he had the following idea:
Have an idicator of which generation is and load the log up to any other on clicking the corresponding number. I think this would solve the issue of memory overload as people would rarely want to see the whole log.
log

kberg:
I'm also interested in storage, and latency impact.

bafolts:
OK. I will re-visit this. Having a separate api route to get the logs should avoid the latency impact. With some form of `/api/game/logs` that provides the logs we can avoid having to send every log to the client.

kberg:
:thumbsup

bafolts:
@chosta I added a separate route for pulling logs of `/api/game/logs`. This change migrates the `LogPanel` to use the route. This should help with pulling logs by generation. This should help with latency and the entire log will never be sent to the client unless the client were to query for the entire log.

I also addressed two warnings which vue was logging to the developer console of missing key.

src/routes/Route.ts
```diff
@@ -0,0 +1,31 @@
+
+import * as http from "http";
+
+export class Route {
```
alrusdi, line 4: Isn't class name little surprising?
As far as i know "routing" in web-apps is traditionally a way to map the Path to the View function.
All functions in this suggested class are Views actually.

Anyway it is great improvement to start the server.ts refactoring. This file takes alot of responsibility. It smells.

bafolts: I have always referred to parts of an api as route. So the current `/api/game` would be the `GameRoute`. This class is meant to act as a utility of sorts to provide common responses that routes will need. There could be a better name here.

kberg: Seems like a fine name, at least for now. A one-sentence comment at the top?

kberg (approved):
Having been in a shockingly high-latency network for a month (sattelite to a main house with a wifi repeater to the guest house, with a private VPN on my end because I don't think the owner understands network security) I do worry about this latency from a network perspective. Games took up to two minutes to load. (I told you it was bad.)

How often should that be your priority? Not a ton, I'm sure, but hundreds or thousand of log messages can be a bear to manage in all parts of the app.

I don't object to this change, but it's something to keep an eye on is all.

src/routes/Route.ts
alrusdi: > I have always referred to parts of an api as route. 

I'm worrying about naming if someday we will add **express** as server side framework. Their term Routing means something else than here https://expressjs.com/en/api.html#router But i'm not insisting of course - just trying to discuss it.

bafolts:
The `main.js` is 2.6MB since latency has never been a consideration. There is a quick win which should help you. We build with webpack setup in development mode. Using production mode cuts the build down to 1.36MB. Every player is downloading that file at least once prior to playback. I can put up a change to build in production mode if an environment variable is specified. The development build we use now comes with vue.js debug logging client side.

If you are running locally using a value of `production` here should give you a quick win.

https://github.com/bafolts/terraforming-mars/blob/master/webpack.config.js#L4

src/routes/Route.ts
bafolts: Using express makes sense and would help with people contributing. Should probably start to migrate towards express. As a history lesson I started this project out of boredom while commuting. I had a goal of taking on no dependencies. I don't think it would be too painful to wedge in express and use it for this new route while the old stuff used native `http`. Then eventually everything can port over to express. People are most likely more familiar with the documented express versus the jalopy code in server.ts

bafolts:
Merged this with master. Will manually test more tomorrow prior to merging. Will do work to introduce `express` for the server in a separate PR.

PR #1434, opened 2020-10-01
Hide player id's

vincentneko:
Warning : This one will break any pending game

Use color instead of playerId in otherPlayers data.
PlayerId will not be visible anymore in Vue data and therefore someone will not be able to access other players pages.

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -720,7 +720,7 @@ function getWaitingFor(
                     if (player === "NEUTRAL") {
                         return "NEUTRAL";
                     } else {
-                        return player.id;
+                        return player.color;
                     }
                 }
             );
--- a/server.ts
+++ b/server.ts
@@ -609,7 +609,7 @@ function getPlayer(player: Player, game: Game): string {
         deckSize: game.dealer.getDeckSize(),
         randomMA: game.gameOptions.randomMA,
         actionsTakenThisRound: player.actionsTakenThisRound,
-        passedPlayers: Array.from(game.getPassedPlayers()), // JSON stringify does not honor sets
+        passedPlayers: game.getPassedPlayers(),
         preludeExtension: game.gameOptions.preludeExtension,
     } as PlayerModel;
     return JSON.stringify(output);
@@ -703,7 +703,7 @@ function getWaitingFor(
             break;
         case PlayerInputTypes.SELECT_PLAYER:
             result.players = (waitingFor as SelectPlayer).players.map(
-                (player) => player.id
+                (player) => player.color
             );
             break;
         case PlayerInputTypes.SELECT_SPACE:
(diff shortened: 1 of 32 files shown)


chosta:
This is very nice! The only general remark is that maybe a separate PlayerIndex set from 0 to 5 would be better. Also, we are kind of stuck to "red", "black", etc. as identifiers for the different players. This might be an issue if we want black to be another color (in terms our brains mapping colors to identifiers, otherwise it shouldn't) . But I know that this would take a lot of refactoring so just mentioning it here as food for thought.

src/Game.ts
```diff
@@ -1421,8 +1421,12 @@ export class Game implements ILoadable<SerializedGame, Game> {
       return this.generation;
     }
 
-    public getPassedPlayers():Set<PlayerId> {
-      return this.passedPlayers;
+    public getPassedPlayers():Array<Color> {
+      let passedPlayersColors: Array<Color> = [];
```
bafolts, line 1425: This map as well use `this.passedPlayers.map((player) => this.getPlayerById(player).color)` as this is mapping the data.

kberg:
I agree, this is a nice idea. And can solve enough of the spectator problem to make it live.

src/Player.ts
```diff
@@ -775,7 +776,7 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
           throw new Error("Invalid players array provided");
         }
         const foundPlayer = pi.players.find(
-            (player) => player.id === input[0][0]
+            (player) => player.color === input[0][0]
```
bafolts, line 779: The only games I think which would break are games where users have the browser open and they can take an input still using id. To not break these users if this check were to use `(player) => player.color === input[0][0] || player.id === [0][0]` I think there would be no introduced errors.

server.ts
```diff
@@ -703,7 +703,7 @@ function getWaitingFor(
             break;
         case PlayerInputTypes.SELECT_PLAYER:
             result.players = (waitingFor as SelectPlayer).players.map(
-                (player) => player.id
+                (player) => player.color
```
bafolts, line 706: The `SELECT_DELEGATE` option will also need to use `player.color` instead of `player.id`.

bafolts (changes requested):
Also need to include `SELECT_DELEGATE` as that is using the player id instead of color. If we check for both the `id` and `color` when matching on the input as this first rolls out we should avoid any intermediate issues during the swap.

src/Game.ts
vincentneko: I get a "Property 'map' does not exist on type 'Set<string>'" when I try to switch to map

server.ts
vincentneko: Good catch !

PR #1438, opened 2020-10-01
Server resp wait buttons

chosta:
@alrusdi please check if this approach looks good to you.
+ created Button component (capital B so we can differentiate in case we ever need to use a normal button markup)
+ added props including disabled on demand and on server waiting (by default all buttons are disabled on wait but there might be some that don't need that hence -> disableOnServerBusy)
+ type and size are self explanatory, I can go on and replace all buttons and have them controlled by one component

- I messed up the PR, had to merge master first but didn't realize I was doing it on my copy of your branch so let me know how we can fix that. Maybe you just merge master again and see what happens


alrusdi:
@chosta it is ok. I would call component "SubmitButton" but Button is ok too

alrusdi:
Closing it. @chosta will make new PR from scratch

PR #1441, opened 2020-10-01
Complete the transition to the new log API.

kberg:
Fixing each card was going to be a bit of a hassle, so I did them separately. Also the subsequent changelist removes the "newLog" method name so everything looks normal again. **I recommend reading each commit separately.**

### DIFF
--- a/src/cards/Ants.ts
+++ b/src/cards/Ants.ts
@@ -7,9 +7,6 @@ import {Game} from "../Game";
 import {SelectCard} from "../inputs/SelectCard";
 import {ResourceType} from "../ResourceType";
 import { CardName } from "../CardName";
-import { LogMessageType } from "../LogMessageType";
-import { LogMessageData } from "../LogMessageData";
-import { LogMessageDataType } from "../LogMessageDataType";
 
 export class Ants implements IActionCard, IProjectCard, IResourceCard {
     public cost: number = 9;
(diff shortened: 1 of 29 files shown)


src/cards/Ants.ts
```diff
@@ -71,19 +68,9 @@ export class Ants implements IActionCard, IProjectCard, IResourceCard {
...
-          new LogMessageData(LogMessageDataType.CARD, card.name)
-        );  
+        game.log("${0} removed a microbe from ${1}'s ${2}", b => b.player(player).player(otherPlayer).card(card));
       } else {
-        game.log(
-          LogMessageType.DEFAULT,
-          "${0} removed a microbe from ${1}",
-          new LogMessageData(LogMessageDataType.PLAYER, player.id),
-          new LogMessageData(LogMessageDataType.STRING, "Neutral Player"));
+        game.log("${0} removed a microbe from ${1}'s ${2}", b => b.player(player).string("Neutral Player"));
```
bafolts, line 73: The first argument should still be `${0} removed a microbe from ${1}`

kberg: You're right. Hey, don't do more review until I do a proper scan. My bad.

bafolts: The rest all look fine. This will be a nice change.

kberg: Well, thanks for doing the rest of them.

And thanks, having written logs, this will at least be minimal. I'm used to older versions of Java -- Typescript is still relatively new to me, but I _really_ like how simple and clear Typescript can make an API like this.

bafolts (changes requested):
This is great change! Spotted one typo.

kberg:
OK PTAL.

bafolts (approved):

Issue #1388, opened 2020-09-25
Playwrites test flaky?


kberg:
From the main branch:

Playwrights
       Can replay other player's event:

      AssertionError: expected 1 to equal 2
      + expected - actual

      -1
      +2

kberg:
@nwai90 this seems to still be happening. https://github.com/bafolts/terraforming-mars/pull/1442/checks?check_run_id=1196738089

bafolts:
I think I spot the issue here. If you happen to draw `TechnologyDemonstration` upon playing it will be removed from your hand throwing off the count by one.

https://github.com/bafolts/terraforming-mars/blob/master/src/Player.ts#L1231
https://github.com/bafolts/terraforming-mars/blob/master/src/cards/community/Playwrights.ts#L43

Will put up fix.

kberg:
It's back: https://github.com/bafolts/terraforming-mars/pull/1479/checks?check_run_id=1218531222

 1) Playwrights
       Can replay other player's event:

      AssertionError: expected 1 to equal 2
      + expected - actual

      -1
      +2
      
      at Context.<anonymous> (dist/tests/cards/community/Playwrights.spec.js:1:9452)

bafolts:
We meet again.

PR #1442, opened 2020-10-01
Reduce duplicated discard action

kberg:
Review and commit this one after #1441 (as that's where the first two commits come from.)

TESTED=manually tested Celestic.


bafolts:
This is another great change. Will need to merge after #1441

kberg:
The test failure is a flaky test. Reported in bug #1388

bafolts (approved):

PR #1317, opened 2020-09-18
Refactor the way cards are laid out -- separated by expansion.

kberg:
TESTING=npm test, played full solo game.
BROKEN=I didn't win the solo game.

Turns out the tricky part was coming up with a structure that I could share with components without feeling like I've inserted too much content into the front-end.

PR #1444, opened 2020-10-02
Move cards out of Dealer and into CardManifests.

kberg:
This clarifies the code structure and localizes some common behavior across the entire codebase.

This supercedes #1317 which was too much of a mess,  @vincentneko. While I manually verified 1317, I haven't verified this one.


kberg:
>  While I manually verified 1317, I haven't verified this one.
... yet. :D

bafolts:
I will give this another pass through tomorrow. Looks great on surface.

On a completely unrelated topic I was surprised to see `var` used. I need to add both `no-var` and `prefer-const`.

https://eslint.org/docs/rules/no-var
https://eslint.org/docs/rules/prefer-const

kberg:
Agree.

>

kberg:
So this PR is about 20 days old if you include the original PR which I had to abandon (nobody's fault but my own.) Is it too much to ask for a closer look at it?

bafolts (approved):
I somehow lost track of this PR, sorry. Looks great! I didn't verify each card moved to the right module but the separation is nice.

PR #1447, opened 2020-10-02
Request for feedback: Multiset

kberg:
In a couple of cases I found I needed really wanted a Multiset type object, so I finally wrote one.

My request: Please review https://github.com/bafolts/terraforming-mars/commit/c9cb11f23540b75cd94e36c62385b407c337a508 which specifically builds and uses the Multiset. Right now it's utils/ which means in order to use it I had to change tsconfig.json.

Let me know your preferences about where the file should live and whether that change to tsconfig.json is acceptable.


src/utils/Multiset.ts
```diff
@@ -0,0 +1,31 @@
+export class Multiset<T> {
+    private map: Map<T, Number> = new Map();
```
bafolts, line 2: With typescript you can use `Map<T, number>` which should make more sense for the use case of this class. This will avoid needing to call `valueOf` throughout this class.

src/utils/Multiset.ts
```diff
@@ -0,0 +1,31 @@
...
+    }
+    add(key: T, count: number = 1) : number {
+        var val = (this.get(key) || 0) + count;
+        this.map.set(key, val);
+        return val;
+    }
+
+    get(key: T): number | undefined {
+        if (this.map.has(key)) {
+            return this.map.get(key)!.valueOf();
```
bafolts, line 15: With using `number` instead of `Number` this function can be:

```
get(key: T): number | undefined {
    return this.map.get(key);
}
```

bafolts (commented):
The Multiset class looks like a good addition. Using `Number` instead of `number` will make it simpler. In typescript the generic argument doesn't need to be an `Object` and any type can be used.

src/utils/Multiset.ts
kberg: I tried, but VSCode seemed to dislike my use of <T, number>. I'll try again.

kberg: Well, I'll be. That's much better.

bafolts (approved):
Multiset looks great.

PR #1453, opened 2020-10-02
chore: add no-var and prefer-const eslint rules

bafolts:
I noticed a pull request yesterday that was using `var`. The usage of `var` can lead to bugs. This requires `let` or `const`. Only `src/*.ts` is being linted. I will include other PRs to put every file under the linter. This should help us catch bugs.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -17,6 +17,8 @@ module.exports = {
   ],
   'rules': {
     'eqeqeq': ['error', 'always'],
-    'quotes': ['error', 'double']
+    'quotes': ['error', 'double'],
+    'no-var': 'error',
+    'prefer-const': 'error'
   },
 };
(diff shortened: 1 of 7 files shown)


.eslintrc.js
```diff
@@ -17,6 +17,8 @@ module.exports = {
   ],
   'rules': {
     'eqeqeq': ['error', 'always'],
-    'quotes': ['error', 'double']
+    'quotes': ['error', 'double'],
```
kberg, line 20: Would you object to a pass through the code to clean some of the remaining single-quotes or bad eqs/

kberg (approved):

.eslintrc.js
bafolts: No. Changing to `eslint src/**/*.ts` which we should have shows 905 problems with 895 potentially fixable with the `--fix` option. Will have to manually fix the others.

PR #1454, opened 2020-10-02
Hide player Id (new PR without conflicts)

vincentneko:


nwai90 (approved):

src/Player.ts
```diff
@@ -775,7 +776,7 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
           throw new Error("Invalid players array provided");
         }
         const foundPlayer = pi.players.find(
-            (player) => player.id === input[0][0]
+            (player) => player.color === input[0][0] || player.id === [0][0].toString()
```
bafolts, line 779: This must be a typo, this should be `input[0][0]` and not `[0][0].toString()`

bafolts: This check also needs to be added for `SelectDelegate`

https://github.com/bafolts/terraforming-mars/blob/master/src/Player.ts#L797

src/components/overview/PlayersOverview.ts
```diff
@@ -37,7 +37,7 @@ export const PlayersOverview = Vue.component("players-overview", {
         },
         getPlayerOnFocus: function (): Player {
             return this.player.players.filter(
-                (p: Player) => p.id === this.player.id
+                (p: Player) => p.id === this.player.color
```
bafolts, line 40: This will need to check that `p.color === this.player.color`

src/components/overview/PlayersOverview.ts
```diff
@@ -14,7 +14,7 @@ export const getCurrentPlayerIndex = (
 ): number => {
     let currentPlayerIndex: number = 0;
     players.forEach((p: Player, index: number) => {
-        if (p.id === player.id) {
+        if (p.id === player.color) {
```
bafolts, line 17: This will need to check that `p.color === player.color`

src/components/overview/PlayersOverview.ts
```diff
@@ -14,7 +14,7 @@ export const getCurrentPlayerIndex = (
 ): number => {
     let currentPlayerIndex: number = 0;
     players.forEach((p: Player, index: number) => {
```
bafolts, line 16: fwiw a client side component will only ever have `PlayerModel` as a type and not `Player`.

src/components/LogPanel.ts
```diff
@@ -48,7 +48,7 @@ export const LogPanel = Vue.component("log-panel", {
             if (data.type !== undefined && data.value !== undefined) {
                 if (data.type === LogMessageDataType.PLAYER) {
                     for (let player of this.players) {
-                        if (data.value === player.id) {
+                        if (data.value === player.color) {
```
bafolts, line 51: This will need the same intermediate workaround for the transition of `data.value === player.color || data.value === player.id`. Games running during the transition could have both `id` or `color` stored for this value.

src/Player.ts
```diff
@@ -199,12 +199,13 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
...
         this.megaCredits += retribution;
         game.getPlayerById(game.monsInsuranceOwner).setResource(Resources.MEGACREDITS,-3);
+        let color = game.getPlayerById(game.monsInsuranceOwner).color;
         if (retribution > 0) {
           game.log("${0} received ${1} MC from ${2} owner (${3})", b =>
               b.player(this)
               .number(retribution)
               .cardName(CardName.MONS_INSURANCE)
-              .playerId(game.monsInsuranceOwner!));
+              .playerId(color));
```
bafolts, line 208: I would use `.player(getPlayerById(game.monsInsuranceOwner))` here.

server.ts
```diff
@@ -754,7 +754,7 @@ function getPlayers(players: Array<Player>, game: Game): Array<PlayerModel> {
             energyProduction: player.getProduction(Resources.ENERGY),
             heat: player.heat,
             heatProduction: player.getProduction(Resources.HEAT),
-            id: player.id,
+            id: player.color,
```
bafolts, line 757: If possible we should remove `id` from `PlayerModel`. Anything in `src/components/` at that point should only make reference to `PlayerModel` and not `Player`.

bafolts (changes requested):
Two minor suggestions and one typo spotted.

As a note and change that could be done separately. Anything in `src/components/` should only make reference to `PlayerModel`. I see there are some references to `Player` which could lead to bugs during runtime when the type is actually `PlayerModel`.

src/Player.ts
vincentneko: If I remove the` .toString()` I get a "This condition will always return 'false' since the types 'string' and 'number' have no overlap." error

src/components/overview/PlayersOverview.ts
vincentneko: Thanks for the clarification. Maybe we can work on this in another PR ?

bafolts: Since I have awoken from my slumber I may tackle this in separate PR as well

bafolts (approved):
This is a great change since it no longer exposes player ids. I can't think of what this could break but this PR could very well have something we aren't considering. Once merged we may end up needing to react to a change. New games will most likely be fine but games that are in progress and have both color and id stored could have issues.

PR #1455, opened 2020-10-02
Card elements to vue components

chosta:
So I was talking to @vsrisuknimit about having an option to toggle between flat design for the icons (old) and the new one (shiny). We just need to switch directories for the assets but this is really hard to do in runtime unless we do some heavy document-wide js replace of attributes. What we want is background-image("assets/tags") to be easily switched to background-image("assets/new-tags") or something similar.
This would be possible with vue by doing something like:
`computed: {
     imageUrl() {
         return require('./path/to/image');
     }
   }`
...but you all know that HTML_DATA holds hardcoded html for the cards. We all know that this needs to be improved at some point. I know that @bafolts was talking about but I just decided to give it a stab. So the following commit is the beginning of what I think needs to be done:
+ add card-title component (already done, this way of doing small vue comps is the way imo)
+ add card-cost component
+ add card-tag component
+ add card-icon component (which expansion)
+ leave the div with the content as it is and replace all the html above it with the vue components and the info from the CardModel

Please, let me know if you are ok with that. I can make all the components' logic relatively fast and we can then decide what to do with the card content.

I also think we can have a schema for each card instead of storing things in the model (like tags and cost). By having a schema we can even try to explain the content of cards behavior.

Let me know if you like the approach and if I should continue.
@vincentneko @alrusdi @nwai90


nwai90:
@chosta This approach looks reasonable imo 👍 Agree that the remainder of the card should be left as is, since that tends to be a highly customised component as well (both content and layout wise)

src/components/Card.ts
```diff
@@ -119,11 +123,17 @@ export const Card = Vue.component("card", {
             return card.name === CardName.SEARCH_FOR_LIFE && card.resources !== undefined && card.resources > 0
         }
     },
+    mounted: function() {
+        console.log(this.card)
```
bafolts, line 127: Remove before flight.

server.ts
```diff
@@ -531,13 +531,13 @@ function getAwards(game: Game): Array<FundedAwardModel> {
...
 function getCorporationCard(player: Player): CardModel | undefined {
-    if (player.corporationCard === undefined) return undefined;
-
+    if (player.corporationCard === undefined) return undefined; 
     return {
         name: player.corporationCard.name,
         resources: player.getResourcesOnCard(player.corporationCard),
         calculatedCost: 0,
         cardType: CardType.CORPORATION,
+        tags: player.corporationCard.tags
```
bafolts, line 540: Since `tags` is a static field we shouldn't have to send it here. Similar to `cardType`. We have these values available on the client stored with each instance of `ICard`. The only thing we should need on `CardModel` are values that can change such as `resources` and `calculatedCost`. The `name` is used for finding the card client side. It is redundant to send this to the client as the value is static.

src/components/Card.ts
```diff
@@ -119,11 +123,17 @@ export const Card = Vue.component("card", {
...
+    },
     template: `
     <div :class="getCardCssClass(card)">
         <img v-if="lifeFound(card)" class="little-green-men" src="assets/martian.png" />
         <div class="card_resources_counter" v-if="card.resources !== undefined">RES:<span class="card_resources_counter--number"> {{ card.resources }}</span></div>
-        <div class="card-content-wrapper" v-i18n v-html=this.getCardContent()></div>
+        <div class="card-content-wrapper" v-i18n>
+            <card-title :title="card.name" :type="card.cardType"/>
```
bafolts, line 134: This is a good change. Doing this in parts may be simpler. I would remove every title from `HTML_DATA`. The big win with this change is the reduction of the built library as HTML is removed from `HTML_DATA`. As well as not having to update every card in `HTML_DATA` to make a change to the title bar.

src/components/card/CardTag.ts
```diff
@@ -0,0 +1,22 @@
+import Vue from "vue";
+
+export const CardTag = Vue.component("card-tag", {
+    props: {
+        title: {
+            type: String,
```
bafolts, line 6: Unrelated, but it is rare to ever need `String` or `Number` as opposed to `string` or `number`. I plan to expand the eslint rules soon to every file. This should help us for people more familiar with terraforming-mars or languages other than typescript.

bafolts: Oh wait, just realized this may be a vue deal. Disregard. Eslint rule still valid. I have that on the brain.

bafolts:
This approach is what I was envisioning. We will want to store the static card data with the built `main.js` that the client loads initially and only send dynamic data from the server through `CardModel`. The final `main.js` size is 2.6MB and HTML_DATA is 358k. Most excited to see how this decreases the size of `main.js` as most the data in HTML_DATA is redundant HTML.

src/components/Card.ts
chosta: Absolutely :)
I'd even recommend adding a no-console in the eslint rules. We can always use a console.log with a suppresion comment when needed.

server.ts
chosta: I was trying to figure out how to get the tags and thought the CardModel was the correct approach as it already had some values. After looking at the code with more thought, your explanations make perfect sense. I'll just use the getCard() method and use the ICard properties (no need for the schema I was thinking about, we have all the info there)

src/components/Card.ts
chosta: It's just one of the things that needs to be done. I can't focus on anything else knowing there is this monster, so lets kill it.

bafolts: Slay the beast! I would do this title component first and remove all the title rows from the HTML_DATA and then move on to the next part. Will help with reviewing. If anyone is updating HTML_DATA will also help with merge conflicts.

chosta:
Having capital letters for Card is opinionated but better imo. We've started refactoring a lot of things, this might help us differentiate what is new vs. old.

chosta:
Concerns from db7a15d:
+ not sure about `getCardExtensionByName`
+ had to make changes in ICard, CorporationCard, IProjectCard interfaces (always makes me uncomfortable)
+ added containers around some of the elements (like tags) and css needs to be refactored too (better do it now anyway)

Next:
Move out dead code
Here. We. Go.

src/cards/ICard.ts
```diff
@@ -41,4 +41,6 @@ export interface ICard {
     onTilePlaced?: (player: Player, space: ISpace, game: Game) => void;
     resourceType?: ResourceType;
     resourceCount?: number;
+    cost?: number;
+    cardType?: CardType;
```
bafolts, line 45: Every card may not have a `cost` but I am pretty sure every card could have a `cardType`. Could make this `cardType: CardType` potentially.

chosta:
Still to-do:
+ corporation icons for expansion
+ bug with resources shown on cards
+ a couple of corps with broken design (point luna)

Taking a break now

src/components/Card.ts
```diff
@@ -1,129 +1,258 @@
...
-        getCardCssClass: function (card: CardModel): string {
-            var cssClass = "filterDiv card-" + card.name.toLowerCase().replace(/ /g, "-");
+        getTags: function (): Array<String> | undefined {
+            return this.getCard()?.tags;
+        },
+        getCost: function (): number | null {
+            const cost = this.getCard()?.cost;
+            const type = this.getCardType();
+            return cost === undefined || type === CardType.PRELUDE
+                ? null
```
bafolts, line 224: I would use `undefined` here and not introduce `null`.

src/components/card/CardExpansion.ts
```diff
@@ -0,0 +1,36 @@
...
+            } else if (this.expansion === CorporationGroup.VENUS_NEXT) {
+                classes.push("venus-icon");
+            } else if (this.expansion === CorporationGroup.TURMOIL) {
+                classes.push("turmoil-icon");
+            } else if (this.expansion === CorporationGroup.COMMUNITY) {
+                classes.push("community-icon");
+            } else if (this.expansion === CorporationGroup.PRELUDE) {
+                classes.push("prelude-icon");
+            } else if (this.expansion === "corporation") {
+                classes.push("corporate-icon");
```
bafolts, line 27: Why is "corporation" left and not `CorporationGroup.CORPORATION`?

src/components/card/CardResourceCounter.ts
```diff
@@ -0,0 +1,14 @@
...
+export const CardResourceCounter = Vue.component("CardResourceCounter", {
+    props: {
+        amount: {
+            type: Number,
+            required: true,
+        },
+    },
+    template: `
+        <div class="card-resources-counter">
+            <span class="card-resources-counter-label>res:</span><span class="card-resources-counter-number"> {{ amount }}</span>
```
bafolts, line 11: This class name string is not closed. Should be `<span class="card-resources-counter-label"`

src/components/card/CardTag.ts
```diff
@@ -0,0 +1,28 @@
+import Vue from "vue";
+import { Tags } from "../../cards/Tags";
+
+export const CardTag = Vue.component("CardTag", {
+    props: {
+        index: {
+            type: Number,
+            required: true,
+            validator: (i) => i < 4,
```
bafolts, line 9: Why is this 4? If this property used?

src/components/card/CardTitle.ts
```diff
@@ -0,0 +1,45 @@
...
+            } else if (this.type === CardType.PRELUDE) {
+                classes.push("background-color-prelude");
+            }
+            return classes.join(" ");
+        },
+    },
+    template: `
+        <div class="card-title">
+            <div v-if="isPrelude()" class="prelude-label">prelude</div>
+            <div v-if="isCorporation()" class="corporation-label">corporation</div>
```
bafolts, line 41: This can be `v-else-if`

bafolts:
I gave this a run locally. This will be a nice change the size of main.js goes from 1.4MB to 1.3MB! This is looking great.

src/cards/ICard.ts
chosta: Making it a required field actually found two corps that I didn't add the property to. TS doing the hard work here :)

src/components/Card.ts
chosta: I see that troughought the app undefined is the preferred null value. I'm totally ok with it and will use it too but was wandering if there is any specific reason for it?

src/components/card/CardExpansion.ts
chosta: The CorporationGroup didn't have it as an option but I should have just added it...

src/components/card/CardTag.ts
chosta: I believe you can't have more than 4 tags per card.

src/components/card/CardTitle.ts
chosta: It was intended this way but I can refactor it further down the line.

chosta:
This should be good to go. I hope we can merge it soon because conflicts are a possibility with any change on the files I've touched.

chosta:
@nwai90 I had to work on some of the new cards, please pull and double check if all good. Thanks

nwai90:
> @nwai90 I had to work on some of the new cards, please pull and double check if all good. Thanks

@chosta Pulled and tested locally. Everything works great 👍

src/cards/community/AerospaceMission.ts
```diff
@@ -12,36 +12,37 @@ export class AerospaceMission extends PreludeCard implements IProjectCard {
...
-                  c1.onColonyPlaced(player, game);
-                  c2.onColonyPlaced(player, game);
-                  return undefined;
-                }
-              );
+        openColonies.forEach(function (c1) {
+            openColonies.slice(1).forEach(function (c2) {
+                if (c1.name !== c2.name) {
+                    const description =
+                        "Build colonies on ${c1.name} (${c1.description}) and ${c2.name} (${c2.description})";
```
bafolts, line 23: This needs to merge with the fix for this string that should be a template string.

bafolts (approved):
Once merging with master everything here looks like it will work. This is a step in the direction of smaller `main.js`. A bulk of the size is from the translations file (411k!). We always pull down every language and not only the desired language.

chosta:
Hm, didn't know translations were that heavy on the size..

bafolts:
I created #1562 . A slight tweak to how we are storing the data should drop the size.

src/cards/community/AerospaceMission.ts
```diff
@@ -12,36 +12,45 @@ export class AerospaceMission extends PreludeCard implements IProjectCard {
...
-                () => {
-                  c1.onColonyPlaced(player, game);
-                  c2.onColonyPlaced(player, game);
-                  return undefined;
-                }
-              );
+        openColonies.forEach(function (c1) {
+            openColonies.slice(1).forEach(function (c2) {
+                if (c1.name !== c2.name) {
+                    const description =
```
kberg, line 22: What happened here?

src/cards/corporation/PhoboLog.ts
```diff
@@ -3,12 +3,15 @@ import { Tags } from "../Tags";
...
 import { CorporationCard } from "./CorporationCard";
-import { CardName } from '../../CardName';
+import { CardName } from "../../CardName";
+import { CardType } from "../CardType";
 
 export class PhoboLog implements CorporationCard {
     public name: CardName = CardName.PHOBOLOG;
     public tags: Array<Tags> = [Tags.SPACE];
     public startingMegaCredits: number = 23;
+    public cardType: CardType = CardType.CORPORATION;
```
kberg, line 13: The fact that this implements CorporationCard means you're repeating yourself. I get if CorporationCard is an interface, but this could be automatically set by a base class, and then these changes would instead just extend that base class, which defines common behavior for corporation cards.

src/components/card/CardTag.ts
kberg: Add a comment, please, so the next person doesn't ask the same question. I'd have the very same question.

src/HTML_data.ts
```diff
@@ -902,12 +642,7 @@ export const HTML_DATA: Map<string, string> =
...
       </div>
 `],
-[CardName.MINE,`
-      <div class="title background-color-automated">Mine</div>
-      <div class="price">4</div>
-      <div class="tag tag1 tag-building"></div>
-      <div class="corporate-icon project-icon"></div>
-      <div class="card-number">056</div>
+[CardName.MINE,` 
       <div class="content">
```
kberg, line 646: Is every card still wrapped in <div class=content>? If so, there's no need for this to repeat, they can be removed.

src/cards/community/AerospaceMission.ts
chosta: Esprettier happened. It came to me unformatted after a merge and this is what happens to + separated string when using prettier. At least its consistent :)
I am using it on all my newly created files and on some of the already existing ones. For me it's a must in such a project but since I'm the only one using it at the moment, I'm doing the herd immunity approach, where I format some files with every PR :)
I am switching it off for big files like Player.ts as it will reformat all (make it readable and consistent) but will be hard to review.

src/cards/corporation/PhoboLog.ts
chosta: I agree with you. It looked superfluous to me too but I'll do it in next PR where I will be dealing with the content of the `HTML_data.ts` file.

src/HTML_data.ts
chosta: Handling of the `<div class="content"/>` will be handled in the next PR. This file will be deprecated.

chosta:
@kberg I love what you did with the Dealer and Deck manifests. Great improvement!

src/components/card/Card.ts
```diff
@@ -13,51 +13,63 @@ import { CardExtraContent } from "./CardExtraContent";
...
+    ALL_PRELUDE_DECKS,
+    ALL_PROJECT_DECKS,
+} from "../../cards/AllCards";
 import { Deck, Decks } from "../../Deck";
-import { ICardFactory } from "../../cards/ICardFactory";
 import { GameModule } from "../../GameModule";
-import { MiningExpedition } from "../../cards/MiningExpedition";
 
 function getCorporationCardByName(cardName: string): ICard | undefined {
-    if (cardName === (new BeginnerCorporation()).name) {
```
chosta: @kberg this is the work of esprettier, I strongly suggest you try it :)

kberg (approved):

PR #1459, opened 2020-10-02
remove unused graphics, and styles

vsrisuknimit:
- Remove old unused graphics from assets
- Remove old styles that is no longer in use such as pinned, unpinned, rt-count.


alrusdi (approved):
Thank you! It is very important to do regular cleanup ><

PR #1460, opened 2020-10-03
Update MASynergy for readability

kberg:
Take a look. The goal is to make the MASynergy algorithm clear. If this moved forward, I'd actually sort the binding calls by decreasing synergy, and eventually suggest removing the 2d array -- that's the point of the unit test -- to show that the two versions are equivalent.


src/MASynergy.ts
```diff
@@ -1,7 +1,38 @@
...
 import { ELYSIUM_MILESTONES, HELLAS_MILESTONES, ORIGINAL_MILESTONES, VENUS_MILESTONES } from "./milestones/Milestones";
+import { Planner } from "./milestones/Planner";
+import { PolarExplorer } from "./milestones/PolarExplorer";
+import { RimSettler } from "./milestones/RimSettler";
+import { Specialist } from "./milestones/Specialist";
+import { Tactician } from "./milestones/Tactician";
+import { Terraformer } from "./milestones/Terraformer";
+import { Tycoon } from "./milestones/Tycoon";
 
 const MA_ITEMS = [
```
bafolts, line 37: Does the `MA` in `MASynergy` stand for `MilestoneAwards`? If a goal of this change is readability it is probably worthwhile to change the name of this file to `MilestoneAwardsSynergy`.

tests/MASynergy.spec.ts
```diff
@@ -0,0 +1,8 @@
+import { expect } from "chai";
+import { buildSynergies, SYNERGIES } from "../src/MASynergy";
+
+describe("MASynergy", function () {
+    it("verify that hand-built synergies is the same as the predefined array", function () {
+        expect(buildSynergies()).is.eql(SYNERGIES);
```
bafolts, line 6: This change looks great to me. I stumbled upon this file while adding eslint updates and I had no idea what the `SYNERGIES` array represented. With this change I have a much better idea what the array represents and would have some shot of addressing a bug. I am still guessing that "MA" stands for milestones and awards but today I read it as ma' synergy or mother synergy which I don't think it stands for.

alrusdi:
I've expected prom this PR two amazing things:
1) Get rid of SYNERGIES matrix
2) Add more unit tests
@kberg is it achievable in this PR?

src/MASynergy.ts
kberg: It does. Will git and github know to show the proper diff?

tests/MASynergy.spec.ts
kberg: Which is correct.

kberg:
> I've expected prom this PR two amazing things:
> 
> 1. Get rid of SYNERGIES matrix
>    @kberg is it achievable in this PR?

Well, this is a request for feedback PR, so let me answer it this way: In short, yes, of course it can. But someone else wanted it this way, so before I delete the matrix, I'd want to know whether anyone objects to its removal.

> 2. Add more unit tests
>    @kberg is it achievable in this PR?
What kind of test would you want in this case? I am having a hard time envisioning a good test here.

alrusdi:
>  But someone else wanted it this way, so before I delete the matrix, I'd want to know whether anyone objects to its removal.

Who wanted? It is imported from @ssimeonoff repo. It is always there if we will need it someday

src/MASynergy.ts
bafolts: Github may struggle to show the diff but if you used `git mv` to move the file then `git` will know the file moved.

bafolts (approved):
Will miss old ma' synergy. The goals of this PR have been met by this reader.

kberg:
> > But someone else wanted it this way, so before I delete the matrix, I'd want to know whether anyone objects to its removal.
> 
> Who wanted? It is imported from @ssimeonoff repo. It is always there if we will need it someday

Actually nobody actively suggested keeping it -- someone was pointing out the advantage of its relationship to a tracking spreadsheet.

Issue #1456, opened 2020-10-02
I think all global parameter indicators should count upwards


kevinb9n:
Notice that the temperature, O2, and venus tracks all show you what the current value of that parameter IS. But the Oceans display tells you how many oceans there are NOT. I think it's an unnecessary inconsistency. Whenever I look at that number I have to subtract it from 9. A new player might look quickly, see a 5, and mistakenly think they should be able to play Algae.

Instead of counting down from 9 to 0, could it count up from "0/9" up to "9/9"? (The "/9" is a reminder for new players, and for players who aren't sure what Dry Deserts does with the ocean it removes.)

The only argument I can guess at for the current behavior is that it's similar to what you see with a physical game. But I think it's clear that mimicking the physical reality has not generally been a design principle here (case in point: party leaders seeming to exist in two places at once).

PR #1462, opened 2020-10-03
count up oceans

vsrisuknimit:
- Count up the ocean value in ? / 9. 
- I believe it is better this way. Easier to understand cards with ocean requirement as well. You don't need to do math in your head to see there are six oceans to play Artic algae.

image

From #1456


nwai90:
The interesting thing is that it used to be this way (counting up). I agree it is good to have it as such (1) for consistency with other global parameters, and (2) less work for the player mentally as they don't need to subtract on their own

kberg:
Oh, yes, please.

bafolts (approved):

PR #1465, opened 2020-10-03
fix: flaky playwright unit test

bafolts:
This addresses #1388 

The failed run:

https://github.com/bafolts/terraforming-mars/pull/1442/checks?check_run_id=1196738089

Was failing on:

```
  1) Playwrights
       Can replay own event:

      AssertionError: expected 3 to equal 4
      + expected - actual

      -3
      +4
```

This would fail if when playing `TechnologyDemonstration` the first time a `TechnologyDemonstration` card were drawn. When repeating the playing of the card the `TechnologyDemonstration` card would be removed from hand so the final count would be 3 instead of 4. I modified the test to use an action that increases terraform rating instead as it should avoid these side effects.


kberg (approved):
Thank you so much!

Issue #1367, opened 2020-09-23
Better default selections for how to pay for things?


kevinb9n:
When it asks how I want to pay for something, it seems that the default selection is to prefer spending MC, and then only spend metals or spendable card resources (e.g. Psychrophiles) when I don't have enough MC.

That seems backwards to me as a default? Except in the case of Helion/Heat, these resources *usually* aren't good for any other purpose. I think you would usually want to spend these resources first (short of overpaying) - or, say, to only have to click once to hold back a single resource for something like Space Elevator.

As a result, whenever I get this prompt, the first thing I do is blindly mash the up arrow on the metals until they top out - but being careful not to go to far (and for example end up using a titanium worth 4 as a single MC). The state I end up in is what seems like should have been the default.

vsrisuknimit:
@kevinb9n I disagree with always spending steels and titaniums. As you said Space Elevator, Jovian Lantern, Security Fleets and other cards require metals to activate. It would be more annoying to have to deduct titanium every time you play space card until you activate Jovian Lantern. You might accidentally forget to do that. Even in the case of psychrophiles, sometimes you want Ants to eat them instead of keep them to win Eccentric.

Perhaps an alternative is to have a MAX button next to steel or titanium (and microbe/floater/heat payment). This will use (1) as much metals as you have, but (2) not overpay.

nwai90:
**Current logic:**
1. No auto selection of other resources if user can afford with MC alone
2. Auto select to spend other resources (e.g. steel/titanium) if user cannot afford with MC alone
3. Will never overpay by default

I agree with @vsrisuknimit that a MAX button would be a good idea as long as we maintain condition (3) - the default selection should never result in overpayment

kevinb9n:
Whatever you want to do here is fine but let me just plead my case a wee bit :-)

The fact that we're here enumerating interesting cases where you don't want to use up resources first, I think actually *supports* my point: that these are indeed the *special* cases.

(Another special case might be the interval between when you acquire Advanced Alloys until you play it.)

Certainly either default behavior can be dangerous when that default is accepted carelessly. I'm here because I *think* that leaving these resources unspent is *more* often the dangerous mistake. There are always other cards/standard-projects/milestones/etc. that you want to buy that generation that won't accept the resources, and it's sad to realize you could have had them if you'd remembered to use your metals etc. earlier.

This could be a toggled preference of some sort. But, given that there is always a confirmation step, a configurable default might be overkill.

[...]

Issue #1464, opened 2020-10-03
Payment widget should not permit a state in which any resource is being discarded for no value


kevinb9n:
Proposal: the payment widget should at all times obey the invariant that each amount selected is a local minimum: you cannot decrement any value by 1 and still be able to purchase the card. Put another way, you can never change a single value by itself; something else must always change in response. That makes sense because what you're choosing is not really the amounts, but how to divide up a cost into various buckets. Making one bigger should always make another smaller.

It's not stated clearly in the game rules or FAQ, but I think everyone understands that when a card says "spend 3 energy" you cannot just voluntarily spend 4 instead (and thus get your Factorum to kick in that way). I don't think the payment widget ever deals with energy, but the principle should be the same - you don't just throw stuff away for no reason.

I think you shouldn't be able to get the widget into this state, whether by typing, up-arrowing, or hitting max.

nwai90:
Fixed and merged

kevinb9n:
As #1466 states, it only "partially" addresses this issue. See:

![image](https://user-images.githubusercontent.com/934551/95024209-7c661e80-0636-11eb-85b1-2e4c0b949fa1.png)

kevinb9n:
Verified still not fixed.

kevinb9n:
If it's not bad behavior, I'd like to broaden this request slightly:

It would be really fantastic if the payment widget simply never permitted ANY invalid states to exist: either the "you are unnecessarily throwing away resources" state already described, or the "you haven't paid enough" state.

It would be a big relief to just be able to mess with the value and submit. Right now I often feel uneasy like I have to check carefully what it's doing.

One thing that might help: consider making the megacredits box special, and not permit any direct changes to it at all. Users only change the other resources and it always calculates the MC in reaction to that. If not that, consider at least not allowing any TYPING, so that arrowing can simply skip over invalid values.

kevinb9n:
It was shown on discord maybe a week ago or so that at least the backend is preventing such payments from occurring, with an error message.

Today it was claimed this error is perhaps not working for Stormcraft. Screenshot was shown of selecting 7 Heat 1 Floater but it's not clear whether the game accepted that payment.

PR #1466, opened 2020-10-03
fixing max button to not overspend

vsrisuknimit:
From this issue #1367 , the max button should not overpay. Currently it spends all available resources, not taking the card price into account.

This PR fixes that issue.

image

Before clicking max:

image

After clicking max:

image

### DIFF
--- a/src/components/PaymentWidgetMixin.ts
+++ b/src/components/PaymentWidgetMixin.ts
@@ -81,7 +81,7 @@ export const PaymentWidgetMixin = {
         },
         setMaxValue: function (target: string): void {
             let currentValue: number = (this as any)[target];
-            let cardCost: number = (this as any).$data.cost;
+            const cardCost: number = (this as any).$data.cost;
             let amountHave: number = (this as any).player[target];
 
             let amountNeed: number = -1;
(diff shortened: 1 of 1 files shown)


vsrisuknimit:
#1464 should also be addressed partly if not completely by this as well.

src/components/PaymentWidgetMixin.ts
```diff
@@ -81,12 +81,20 @@ export const PaymentWidgetMixin = {
         },
         setMaxValue: function (target: string): void {
             let currentValue: number = (this as any)[target];
-            let maxValue: number = (this as any).player[target];
+            const cardCost: number = (this as any).$data.cost;
+            let amountHave: number = (this as any).player[target];
 
-            if (target === "microbes") maxValue = (this as any).playerinput.microbes;
-            if (target === "floaters") maxValue = (this as any).playerinput.floaters;
+            let amountNeed: number = -1;
```
nwai90, line 87: Minor: Can directly do `let amountNeed = cardCost;` here so that only one `if` check is needed below on line 88

nwai90 (approved):

PR #1469, opened 2020-10-04
Replace solo city tile placement with a card-draw mechanic.  Also rem…

### LINKED ISSUE #1467: FR: use deck-based random initial city selection for solo.
The official solo rules [ref](https://www.fryxgames.se/TerraformingMars/TMRULESFINAL.pdf) for placing random initial tiles is

> 1) Before you choose your cards, place 2 neutral city tiles on the map with an adjacent greenery tile each (these tiles are not yours, and do not increase the oxygen level): reveal and discard the 4 top cards of the deck and use their cost numbers to determine the positions of the tiles. The first city is placed counting from top left to right and down, like reading. Skip any illegal placements (like areas reserved for ocean). For the second city you step backwards from bottom right in the same fashion. Then you place the two greeneries by counting the cost numbers and stepping clockwise around each city, starting from top left, skipping illegal placements.

The current design doesn't do this, it relies on a random number generator in `Board.getRandomCitySpace().`

Ares needs similar behavior, but instead I wrote code that matches the intent of the game rules:
```
public getAvailableSpaceByOffset(distance: number, direction: -1 | 1) {
    // By doing an additional filter on space.tile, it skips over hazards.
    var spaces = this.getAvailableSpacesOnLand(undefined).filter(space => !space.tile);
    var idx = (direction === 1) ? distance : (spaces.length - (distance + 1));
    return spaces[idx];
}
```
If people would prefer it I can replace Board.getRandomCitySpace with this algorithm (and the other half, which draws a card and computes that distance accordingly.)

```
function randomlyPlaceHazard(game: Game, tileType: TileType, direction: 1 | -1) {
    var card = game.dealer.dealCard();
    game.log(
        LogMessageType.DEFAULT,
        "Dealt and discarded ${0} (cost ${1}) to place a hazard",
        new LogMessageData(LogMessageDataType.CARD, card.name),
        new LogMessageData(LogMessageDataType.STRING, card.cost.toString())
    );

    var distance = card.cost - 1;
    distance = Math.max(distance, 0); // Some cards cost zero.
    var space = game.board.getAvailableSpaceByOffset(distance, direction);
    if (space === undefined) {
        throw new Error("Couldn't find space when card cost is " + card.cost);
    }
    AresHandler.putHazardAt(space, tileType);
    return space;
}
```

If there's interest I will assemble a PR to bafolts/master.

kberg:
…ove some now-unused and unnecessary methods.

Fixes #1467

### DIFF
--- a/src/Board.ts
+++ b/src/Board.ts
@@ -142,9 +142,9 @@ export abstract class Board {
         return this.spaces.filter((space) => space.spaceType === spaceType);
     }
 
-    protected getRandomSpace(offset: number): ISpace {
-        return this.spaces[Math.floor(Math.random() * 30) + offset];
-    }
+    // protected getRandomSpace(offset: number): ISpace {
+    //     return this.spaces[Math.floor(Math.random() * 30) + offset];
+    // }
 
     public getEmptySpaces(): Array<ISpace> {
         return this.spaces.filter((space) => space.tile === undefined);
@@ -205,20 +205,17 @@ export abstract class Board {
         return out;
     }
 
-    public getRandomCitySpace(offset: number): Space {
-        let safety = 0;
-        // avoid bugs which would lock node process
-        while (safety < 1000) {
-            const space = this.getRandomSpace(offset);
-            if (this.canPlaceTile(space) && this.getAdjacentSpaces(space).filter(sp => sp.tile?.tileType === TileType.CITY).length === 0 && this.getAdjacentSpaces(space).find(sp => this.canPlaceTile(sp)) !== undefined) {
-                return space;
-            }
-            safety++;
-        }
-        throw new Error("space not found for getRandomCitySpace");
+    // If direction is 1, count from the top left. If -1, count from the other end of the map.
+    public getNthAvailableLandSpace(
+        distance: number, 
+        direction: -1 | 1,
+        predicate: (value: ISpace) =>  boolean = _x => true) {
+        const spaces = this.spaces.filter((space) => space.spaceType === SpaceType.LAND && space.tile === undefined).filter(predicate);
+        const idx = (direction === 1) ? distance : (spaces.length - (distance + 1));
+        return spaces[idx];
     }
-
-    protected canPlaceTile(space: ISpace): boolean {
+    
+    public canPlaceTile(space: ISpace): boolean {
         return space !== undefined && space.tile === undefined && space instanceof Land;
     }
 
--- a/src/Board.ts
+++ b/src/Board.ts
@@ -142,10 +142,6 @@ export abstract class Board {
         return this.spaces.filter((space) => space.spaceType === spaceType);
     }
 
-    // protected getRandomSpace(offset: number): ISpace {
-    //     return this.spaces[Math.floor(Math.random() * 30) + offset];
-    // }
-
     public getEmptySpaces(): Array<ISpace> {
         return this.spaces.filter((space) => space.tile === undefined);
     }
@@ -209,8 +205,11 @@ export abstract class Board {
     public getNthAvailableLandSpace(
         distance: number, 
         direction: -1 | 1,
+        player: Player | undefined = undefined,
         predicate: (value: ISpace) =>  boolean = _x => true) {
-        const spaces = this.spaces.filter((space) => space.spaceType === SpaceType.LAND && space.tile === undefined).filter(predicate);
+        const spaces = this.spaces.filter((space) => {
+            return this.canPlaceTile(space) && (space.player === undefined || space.player === player);
+        }).filter(predicate);
         const idx = (direction === 1) ? distance : (spaces.length - (distance + 1));
         return spaces[idx];
     }
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -1706,24 +1706,40 @@ export class Game implements ILoadable<SerializedGame, Game> {
       // Single player add neutral player
       // put 2 neutrals cities on board with adjacent forest
       const neutral = new Player("neutral", Color.NEUTRAL, true, 0);
-      const space1 = this.board.getRandomCitySpace(0);
-      this.addCityTile(neutral, space1.id, SpaceType.LAND);
-      const fspace1 = this.board.getForestSpace(
-          this.board.getAdjacentSpaces(space1)
-      );
-      this.addTile(neutral, SpaceType.LAND, fspace1, {
-        tileType: TileType.GREENERY
-      });
-      const space2 = this.board.getRandomCitySpace(30);
-      this.addCityTile(neutral, space2.id, SpaceType.LAND);
-      const fspace2 = this.board.getForestSpace(
-          this.board.getAdjacentSpaces(space2)
-      );
-      this.addTile(neutral, SpaceType.LAND, fspace2, {
-        tileType: TileType.GREENERY
-      });
+
+      function placeCityAndForest(game: Game, direction: -1 | 1) {
+        const space1 = game.getSpaceByOffset(direction);
+        game.addCityTile(neutral, space1.id, SpaceType.LAND);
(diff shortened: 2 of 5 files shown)


kberg:
TESTED=unit tests and a manual run-through of the solo game. It must be broken because it's the first solo game I've won.

src/Board.ts
```diff
@@ -205,20 +205,17 @@ export abstract class Board {
...
-            }
-            safety++;
-        }
-        throw new Error("space not found for getRandomCitySpace");
+    // If direction is 1, count from the top left. If -1, count from the other end of the map.
+    public getNthAvailableLandSpace(
+        distance: number, 
+        direction: -1 | 1,
+        predicate: (value: ISpace) =>  boolean = _x => true) {
+        const spaces = this.spaces.filter((space) => space.spaceType === SpaceType.LAND && space.tile === undefined).filter(predicate);
```
bafolts, line 213: Technically a space is only available if no player has claim to it. When a player claims a space with the `LandClaim` card the `tile` property will be `undefined` but `player` will be defined. This should also check that `space.player === undefined`.

https://github.com/bafolts/terraforming-mars/blob/master/src/cards/LandClaim.ts#L25

src/Board.ts
```diff
@@ -142,9 +142,9 @@ export abstract class Board {
         return this.spaces.filter((space) => space.spaceType === spaceType);
     }
 
-    protected getRandomSpace(offset: number): ISpace {
-        return this.spaces[Math.floor(Math.random() * 30) + offset];
-    }
+    // protected getRandomSpace(offset: number): ISpace {
```
bafolts, line 145: May as well remove this it can live on in git history and our memories.

src/Board.ts
```diff
@@ -205,20 +205,17 @@ export abstract class Board {
...
+        direction: -1 | 1,
+        predicate: (value: ISpace) =>  boolean = _x => true) {
+        const spaces = this.spaces.filter((space) => space.spaceType === SpaceType.LAND && space.tile === undefined).filter(predicate);
+        const idx = (direction === 1) ? distance : (spaces.length - (distance + 1));
+        return spaces[idx];
     }
-
-    protected canPlaceTile(space: ISpace): boolean {
+    
+    public canPlaceTile(space: ISpace): boolean {
```
bafolts, line 218: The `LandClaim` card is frustrating in that a player can claim a tile and at that point only that player can place a tile. While we are here a better name for this function may be `isLandTileEmpty`.

src/Game.ts
```diff
@@ -1706,25 +1706,41 @@ export class Game implements ILoadable<SerializedGame, Game> {
...
+  
+      const distance = Math.max(card.cost-1, 0); // Some cards cost zero.
+      const space = this.board.getNthAvailableLandSpace(distance, direction,
+        space => {
+          const adjacentSpaces = this.board.getAdjacentSpaces(space);
+          return adjacentSpaces.filter(sp => sp.tile?.tileType === TileType.CITY).length === 0 && // no cities nearby
+             adjacentSpaces.find(sp => this.board.canPlaceTile(sp)) !== undefined;  // can place forest nearby
+        });
+      if (space === undefined) {
+          throw new Error("Couldn't find space when card cost is " + card.cost);
```
bafolts, line 1739: This will be helpful in the case there is some issue here. Unit tests should end up running through this logic multiple times with randomly drawn cards. :+1:

src/Game.ts
```diff
@@ -1706,25 +1706,41 @@ export class Game implements ILoadable<SerializedGame, Game> {
...
+      placeCityAndForest(this, -1);
+
       return undefined;
     }
 
+    public getSpaceByOffset(direction: -1 | 1, type = "tile") {
+      const card = this.dealer.dealCard();
+      this.log("Dealt and discarded ${0} (cost ${1}) to place a ${2}", b => b.card(card).number(card.cost).string(type));
+  
+      const distance = Math.max(card.cost-1, 0); // Some cards cost zero.
```
bafolts, line 1731: Why does this need to subtract 1 from the cost? Cards that cost 0 and cards that cost 1 would both end up using the same tile.

bafolts (commented):
This looks right. Curious why we need to subtract 1 from the cost instead of using the cost directly.

src/Board.ts
kberg: You're right. This code is only used in two places at the moment, and if I make this change it will (mildly) break one of them. Or I just need to reinterpret the rules.

1. In the solo-setup scenario land claim won't have occurred, so no risk there if this is added. Also a player using Land Claim in a solo game better have a deep deep discount to want to play it! ha ha!
2. When placing hazards in Ares, some happens during set-up (same as 1. above) but some happens mid-game. Should a land claim actually prevent erosion from occurring? I believe the spirit of Land Claim overrides the literal rule on the card in this case. But I can be convinced otherwise.

As we're chatting I realize that by making the change you suggest would solve an unrelated problem having to do with how I implemented the Ares card Desperate Measures. _I have a truly remarkable description of this case which this margin is too small to contain._

src/Board.ts
kberg: Whoops yes of course.

src/Board.ts
bafolts: If someone were to claim land that ended up with an environmental hazard it makes sense to me that the claim to the land by a corporation would be irrelevant. Any time a corporation makes claim to a land they risk anything that may happen to that land on the planet. The claim to the land will only need to be considered when another corporation wants to use the tile. Not checking for the land claim here makes sense to me as it is a tough luck scenario if a corporation had made a claim to the land.

kberg: I'll have to go back and fix this when we look to integrate Ares properly. Thanks.

src/Board.ts
kberg: Well, again, this method is only used for solo placement. Land Claim never really took part in this. I suggest a comment and doubling back to fix this later in the name of setting up a clean contract. Of course I created this problem by making it a public method, so if you still suggest it, just say that you feel it's worth doing now.

src/Game.ts
kberg: ack thanks. I wrote it when my code failed and I was like what is going on here??? That's when I was reminded that cards can have raw cost zero.

src/Board.ts
bafolts: I don't feel it is worth doing now. Only curious why we subtract one from the project cost when determining distance.

src/Game.ts
kberg: Yes, they would. People aren't used to counting from zero. (I'm not going to poll people using this repository, they don't count! :D) If you drew a card of value three, you would naturally look at the third space from the top left corner of the board, not the fourth. As a result, I think zero is the outlier.

It's not as if this meaningfully increases the risk of tiles placed in the wrong space (1.4% [5 out of 363] project cards cost 0), and will visually make sense when actually running the app. Tell you what: take a look at a solo game's log and correlate it to the map and you'll see what I mean.

kberg: https://boardgamegeek.com/thread/1680256/article/30559949#30559949

I'm heading out for the day. Ta!

bafolts (approved):
Awesome :shipit:

kberg:
> This looks right. Curious why we need to subtract 1 from the cost instead of using the cost directly.
Oh, I see. Subtracting one means distance from the upper right -- and there we count from zero. Perhaps it's still unnecessary and I'm blind to it. But I'm going to need to integrate this with the Ares branch and I'll take another look.

Or perhaps you just mean, why Math.max(cost - 1, 0)?

Well, again, in this case 0 represents the upper left (or bottom right) corner. cost - 1 then equates a card of cost 1 with the first space in the upper right corner. For cards that cost zero, Math.max(-1, 0) will also be 0. Does that explain it?

PR #1470, opened 2020-10-04
Fix build failure due to new linter rule

alrusdi:
I've broken master branch by blindly accepting some approved PRs
Below are fixes for the problems i found so far


kberg (changes requested):
Did you mean to include make_source.py?

alrusdi:
> Did you mean to include make_source.py?

no-no! It is mistake of cource

bafolts:
Trying to think of the most painless way to introduce these linting rules. I feel there is enough value with using `let` and `const` versus `var` that the pain is worth it. I thought I had this rule turned on initially. This should have covered both `src/Player.ts` and `src/Game.ts` which get a lot of modifications as those files are very large. I will do the change per directory hoping to avoid these continual conflicts. Sorry for any pain in the meantime.

alrusdi:
> Trying to think of the most painless way to introduce these linting rules. 
We need them anyway. Clean code is better code!

src/locales/initial_source.json
```diff
@@ -0,0 +1,10554 @@
+[
+    {
+        "model": "translations.source",
+        "pk": 1,
+        "fields": {
```
bafolts, line 5: This file also seems unrelated to fixing linting rule failure.

kberg:
> > Trying to think of the most painless way to introduce these linting rules.
> 
> We need them anyway. Clean code is better code!

Strongly agree.

src/locales/initial_source.json
alrusdi: Yes, it is fixed too

bafolts (approved):

kberg (approved):

PR #1471, opened 2020-10-04
Use BoardBuilder to replace duplication across board generators.

kberg:
Code is easier to read and it paves the way for more interesting boards.

I most certainly need to add more tests, and it's possible that I overdid it with how Hellas continues to keep its protected ocean space. But my goodness it's easier to understand.

There's further simplifications that can come from this, but first I want to see if this will be generally accepted.


src/BoardBuilder.ts
```diff
@@ -0,0 +1,103 @@
+import { Ocean, Land, BoardColony } from "./Board";
+import { ISpace } from "./ISpace";
+import { Random } from "./Random";
+import { SpaceBonus } from "./SpaceBonus";
+import { SpaceName } from "./SpaceName";
+
+export class BoardBuilder {
+    private rng: Random;
+    // This builder assumes the map has nine rows, of growing tiles [5,6,7,8,9,8,7,6,5].
```
bafolts, line 9: A rare moment where this is a palindrome. Would be wrong not to use palindrome in the comment in some way.

src/BoardBuilder.ts
```diff
@@ -0,0 +1,103 @@
...
+            }  
+        }
+
+        this.spaces.push(new BoardColony(SpaceName.STANFORD_TORUS));
+
+        return this.spaces;
+    }
+
+    public shuffleArray(array: Array<Object>): void {
+        const sorted = this.unshufflableSpaces.sort((a, b) =>  a < b ? a : b);
```
bafolts, line 59: :warning: the call to sort will sort `this.unshuffableSpaces` in place.

src/BoardBuilder.ts
```diff
@@ -0,0 +1,103 @@
...
+                idx++;
+            }  
+        }
+
+        this.spaces.push(new BoardColony(SpaceName.STANFORD_TORUS));
+
+        return this.spaces;
+    }
+
+    public shuffleArray(array: Array<Object>): void {
```
bafolts, line 58: If the type of the items in the `Array` isn't important it may make sense to use `unknown`. At that point no assumptions can be made as to the type of the items in the array.

bafolts:
This change is accepted by me. Looks great!

kberg:
> This change is accepted by me. Looks great!

Great. I'll do more work on it. Thanks.

src/BoardBuilder.ts
kberg: There are at least three shuffle algorithms in the codebase. I won't quit until there's only one.

kberg:
Manual layout verification. New maps are on the LHS.

**Tharsis**
tharsis-side-by-side

**Hellas**

hellas-side-by-side

**Elysium**

elysium-side-by-side

Huzzah! Something wrong! So I had to correct it.

elysium-corrected

src/BoardBuilder.ts
```diff
@@ -0,0 +1,109 @@
...
+import { Random } from "./Random";
+import { SpaceBonus } from "./SpaceBonus";
+import { SpaceName } from "./SpaceName";
+
+export class BoardBuilder {
+    private rng: Random;
+
+    // This builder assumes the map has nine rows, of tile counts [5,6,7,8,9,8,7,6,5].
+    //
+    // "Son I am able, " she said "though you scare me."
```
bafolts, line 12: :+1:

bafolts (approved):
This is great. :shipit:. Has inspired me to put more love into performance on the server!

PR #1476, opened 2020-10-04
refactor: include git log information for version

bafolts:
This has a few changes related to how the server operates. 

Included with the `StartScreen` component is now the commit hash to display some form of version for users. It will appear as:

`e48ba89a Sun, 4 Oct 2020 15:40:43 -0600`

This should be a simple solution and not get in the way of contributors while allowing users to know which version of the application they are using. If we decide to update what this displays in the future the mechanism will be in place for displaying the value.

I also updated the webpack configuration to link to the source map. I noticed that all of our javascript was attempting to reach the source map through `webpack://` when the developer console was open. With this change the developer console should now link back to out typescript.

I also updated webpack to use production mode which runs minify on our javascript. This leads to a small main.js file. The main.js is now 1.3MB down from 2.6MB. With the map file now in place this should not impact developers as the map file still shows the original line numbers. For users without their devtools open they will now only have to load 1.3MB.

I also changed some of the `readFileSync` to `readFile` so they work asynchronously. As written all other requests would be blocked until the `readFileSync` call were to complete. Node is good at input and output and can get off the main thread while reading files.

### DIFF
--- a/src/components/StartScreen.ts
+++ b/src/components/StartScreen.ts
@@ -22,7 +22,7 @@ export const StartScreen = Vue.component("start-screen", {
     <a class="start-screen-link start-screen-link--chat" href="https://discord.gg/fWXE53K" target="_blank" v-i18n>Join us on Discord</a>
     <div class="start-screen-header start-screen-link--languages">
       <language-switcher />
-      <div>{{version}}</div>
+      <div class="start-version">version: {{version}}</div>
     </div>
   </div>
 </div>`
--- a/src/components/StartScreen.ts
+++ b/src/components/StartScreen.ts
@@ -2,29 +2,28 @@ import Vue from "vue";
 import {LanguageSwitcher} from "./LanguageSwitcher";
 
 export const StartScreen = Vue.component("start-screen", {
+    props: ["version"],
     components: {
         LanguageSwitcher
     },
     template: `
-        <div class="start-screen">           
-            <div v-i18n class="start-screen-links">
-                <div class="start-screen-header start-screen-link--title">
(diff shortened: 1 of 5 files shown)


src/components/StartScreen.ts
```diff
@@ -2,29 +2,28 @@ import Vue from "vue";
...
+    <a class="start-screen-link start-screen-link--new-game" href="/new-game" v-i18n>New game</a>
+    <a class="start-screen-link start-screen-link--solo" href="/solo" v-i18n>Solo challenge</a>
+    <a class="start-screen-link start-screen-link--cards-list" href="https://ssimeonoff.github.io/cards-list" v-i18n>Cards list</a>
+    <a class="start-screen-link start-screen-link--board-game" href="https://boardgamegeek.com/boardgame/167791/terraforming-mars" v-i18n>Board game</a>
+    <a class="start-screen-link start-screen-link--about" href="https://github.com/bafolts/terraforming-mars" v-i18n>About us</a>
+    <a class="start-screen-link start-screen-link--changelog" href="https://github.com/bafolts/terraforming-mars/wiki/Changelog" v-i18n>Whats new?</a>
+    <a class="start-screen-link start-screen-link--chat" href="https://discord.gg/fWXE53K" target="_blank" v-i18n>Join us on Discord</a>
+    <div class="start-screen-header start-screen-link--languages">
+      <language-switcher />
+      <div>{{version}}</div>
```
kberg, line 25: Version: {{version}}
?

kberg (approved):
Thanks for the source-map fix, by the way.

bafolts:
I will need to re-visit this. Heroku isn't pulling this value.

PR #1478, opened 2020-10-04
Adding global parameters indicator

vsrisuknimit:
Indicator for global parameters on the side bar.

Without Venus.

image

With Venus.

image

@ssimeonoff please help.


vsrisuknimit:
- Zoom out works fine. Zooms in too closely make the number move weirdly, but at that zoom level, the side bar misbehaves anyway (you cannot see the colony or card buttons).
- The number is currently clickable as <slot></slot> is under cards button. I don't know how to fix it, sorry.
https://github.com/bafolts/terraforming-mars/blob/eb5afa1ff456db55baebcdc059a3526301309761/src/components/Preferences.ts#L115-L118

vsrisuknimit:
I would like to hide the Venus icon if Venus is not in play. Not sure how to do that since I have that information in Playerhome.ts but not in Preferences.ts where the image is called.

alrusdi (changes requested):
Why do we need an additional image?
All images are already in assets/global-parameters and in tiles/

chosta:
@alrusdi I will help @vsrisuknimit with this one in a private session.

alrusdi:
> Also I'm not digging how wide the sidebar is becomming
I think @ssimeonoff already made it big enough to fit global params

ssimeonoff:
Now I see why you needed those margin-top for the param values :)

I will fix it. Currently the global values are rendered into the cards , instead of being passed as props to Preferences.

ssimeonoff:
Untitled
Changes: 
- SideBar reduced back to 60px width
- Venus param not shown when Venus is not in play
- The chosen order for the params follow the main map layout: top - Venus, bottom: oceans)

alrusdi (approved):
Great!

PR #1479, opened 2020-10-04
Add filter to debug UI card list.

kberg:
filter


src/components/DebugUI.ts
```diff
@@ -16,17 +21,20 @@ export const DebugUI = Vue.component("debug-ui", {
...
+        },
+        filtered: function(cardName: string):boolean {
+            return this.$data.filterText.length == 0 || cardName.toUpperCase().indexOf(this.$data.filterText.toUpperCase()) > -1;
         }
     },
     template: `
         <div class="debug-ui-container">
             <section class="debug-ui-cards-list">
-                <h2>Cards list</h2>
+                <h2>Cards list <input placeholder="filter" v-model="filterText"></h2>
```
alrusdi, line 35: Add classes "form-input form-inline" to the input

src/components/DebugUI.ts
```diff
@@ -7,6 +7,11 @@ export const DebugUI = Vue.component("debug-ui", {
     components: {
         Card,
     },
+    data: function() {
+        return {
+          filterText: ""
```
alrusdi, line 12: Can you take initial value of filterText from URL like "/debug-ui/?search-card=venusian" ahen component is "mounted"?

alrusdi (changes requested):

src/components/DebugUI.ts
kberg: I didn't realize this was holding up the review, sorry. Can that be done in a follow-up? I'm not sure I want to start diving into URL parameters yet.

src/components/DebugUI.ts
alrusdi: @kberg and what about this change?

kberg: > Add classes "form-input form-inline" to the input
Done, good, thanks!

> @kberg and what about this change?

My head doesn't have that anymore. Still an issue?

alrusdi (approved):

PR #1487, opened 2020-10-05
[WIP] Fix: Some plays that can benefit from ocean adjacency bonus are blocked by Reds tax

nwai90:
**Ref:** https://github.com/bafolts/terraforming-mars/issues/1152

**Example scenario for 1 card:** Artificial Lake should be playable here. There is a valid spot that gives 4 MC upon placement. Current logic deems the card to be unplayable.

Screenshot 2020-10-05 at 6 05 08 PM

**Result:** Player is left with 1 MC (4 MC - 3 MC paid for Reds tax)

Screenshot 2020-10-05 at 6 19 21 PM

This is a draft of a possible partial fix for the one-tile scenario. However it gets a lot more complicated as some cards can place 2-3 tiles on the board (e.g Ice Asteroid, Deimos Down / Giant Ice Asteroid triggering the bonus ocean).

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1112,6 +1112,20 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
       return Math.max(cost, 0);
     }
 
+    public getMaxOceanPlacementBonus(game: Game) : number {
+      const spaces = game.board.getAvailableSpacesOnLand(this);
+      let maxRebate = 0;
+
+      spaces.forEach((space) => {
+        const adjacentSpaces = game.board.getAdjacentSpaces(space);
+        const oceanAdjacencyBonus = adjacentSpaces.map((s) => s.spaceType === SpaceType.OCEAN && s.tile ? this.oceanBonus : 0).reduce((acc, val) => acc + val);
+        
+        if (oceanAdjacencyBonus > maxRebate) maxRebate = oceanAdjacencyBonus;
+      })
+
+      return maxRebate;
+    }
+
     private addPlayedCard(game: Game, card: IProjectCard): void {
       this.playedCards.push(card);
       game.log("${0} played ${1}", b => b.player(this).card(card));
@@ -1830,6 +1844,7 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
         }
 
         maxPay += this.megaCredits;
+
         return maxPay >= this.getCardCost(game, card) &&
                    (card.canPlay === undefined || card.canPlay(this, game));
       });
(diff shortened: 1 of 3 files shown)


kberg:
This problem also gets complicated with Ares, and I've been thinking about
that. Including that I reordered addTile so that payments occur before
placements.

Getting tile placement costs and rewards just right is tricky, and I don't
think there's a simple solution -- a more complicated one perhaps.

Or these are two separate issues, and I misunderstood yours.

>

nwai90:
> This problem also gets complicated with Ares, and I've been thinking about that. Including that I reordered addTile so that payments occur before placements. Getting tile placement costs and rewards just right is tricky, and I don't think there's a simple solution -- a more complicated one perhaps. Or these are two separate issues, and I misunderstood yours.
> […](#)

@kberg There are two issues here - (1) Europa colony placement bonus can avoid Reds tax, (2) Some legal plays are blocked due to ocean adjacency bonuses not being factored in. 

I believe your scenario is related to (2). The complexity comes with cards like Deimos Down Promo / GIA, which can place 2-3 tiles on the board in succession. It should be possible to resolve this algorithmically; I will have a think about it. My other thought was to store a value for ocean adjacency bonuses for each board space and update that with every tile placement, but that's a bigger change. The edge case of Lakefront Resorts could be checked for and handled in the same logic block.

kberg:
Documenting my research: The Reds ruling policy is
> You lose 3M€ for each step your TR is raised. If you have less than 3M€, you may **not** raise your TR.
So, it shouldn't stop a player from playing a tile, just should stop them from gaining the TR bonus. And by having the 2 oceans, well it's a matter of whether the TR goes up before the adjacency bonuses kick in. (player-preferred order, I'm sure.) _Is this assessment wrong?_

I notice a mass proliferation of code throughout all the cards to accommodate the reds tax. Surely this whole thing could start by removing the reds cost from all the canPlay from cards, and instead implement it as part of the code that increases terraforming ratings. Again, if it's too late, and they can't afford it, they lose the TR bump. But that's what undo is for.

kberg:
WDYT?

nwai90:
@kberg Undoing the TR bump alone would not be sufficient though, as certain plays (e.g. placing a greenery or ocean tile) still alter the board state or give additional VP / effects (e.g. Lakefront with ocean placements, Philares with adjacency bonuses, etc). I believe this check has to be done prior to making the play as it should not be allowed in the first place, since it is strictly not a valid play.

kberg:
As a reference, this is it: https://boardgamegeek.com/thread/2196388/article/31885882#31885882

kberg:
Is this still relevant @nwai90?

nwai90:
> Is this still relevant @nwai90?

Still relevant; not had time to look beyond the one-tile placement scenario though. In the most extreme cases a single action can place up to 3 tiles onto the board.

nwai90:
Closing as superceded by https://github.com/bafolts/terraforming-mars/pull/1864

PR #1490, opened 2020-10-05
Moved buttons to vue

chosta:
done: 
+ moved all  `button` elements to a `Button` vue component
+ by default all buttons are disabled + loading when waiting for server response
+ multiple props to control the layout, functionality and behavior of the buttons

comments:
+ we might use a better loading indicator @ssimeonoff 
+ the .btn css is overriding btn-error and btn-success properties, I decided to not change functionality and do another pull request to resolve this issue
+ I think titanium max does not work properly (in comparison to steel) @nwai90 

I've tested all the changes manually but any other manual test is welcome.
@alrusdi @kberg @bafolts


chosta:
99% of the work is in 1760c532a19c123c1929911d41efe8b903f25492

src/components/AndOptions.ts
```diff
@@ -24,43 +33,56 @@ export const AndOptions = Vue.component("and-options", {
...
-                    on: { click: () => { this.saveData(); } } 
-                }, 
-                playerInput.buttonLabel
-            );
+            const saveBtn = createElement(Button, {
+                props: {
+                    title: playerInput.buttonLabel,
+                    type: "submit",
+                    size: "normal",
+                    onClick: this.saveData,
```
bafolts, line 79: I would prefer to still use `() => { this.saveData(); }`. The fat arrow function maps the `this` argument automatically. Leaving this function exposed as `this.saveData` I am not sure what the `this` argument will end up being. Using the fat arrow function keeps the context within this class.

src/components/CardsFilter.ts
```diff
@@ -58,15 +63,15 @@ export const CardsFilter = Vue.component("cards-filter", {
...
+        } 
     },
     template: `
     <div class="cards-filter">
         <h2 v-i18n>Cards to exclude from the game</h2>
         <div class="cards-filter-results-cont" v-if="selectedCardNames.length">
             <div class="cards-filter-result" v-for="cardName in selectedCardNames">
                 <label>{{ cardName }}</label>
-                <button class="btn btn-error btn-sm" v-on:click.prevent="removeCard(cardName)"><i class="icon icon-cross"></i></button> 
+                <Button size="small" type="error" :onClick="removeCard.bind(null, cardName)" v-html="getCrossHtml()" />
```
bafolts, line 74: I am not sure how this works. Providing `null` as the `thisArg` to the `removeCard` function should end up causing a runtime exception.We can't lose scope.

src/components/LogPanel.ts
```diff
@@ -144,7 +147,7 @@ export const LogPanel = Vue.component("log-panel", {
             </div>
         </div>
         <div class="card-panel" v-if="cards.length > 0">
-            <button class="btn btn-sm btn-error other_player_close" v-on:click="hideMe()"><i class="icon icon-cross"></i></button>
+            <Button size="big" type="close" :onClick="hideMe" align="right" />
```
bafolts, line 150: In other places this `getCrossHtml` was provided to `Button`. I assume that also need to be passed here.

src/components/OrOptions.ts
```diff
@@ -1,71 +1,130 @@
...
+                            style: { "margin": "5px 30px 10px" },
+                            "class": "wf-action",
+                        },
+                        [
+                            createElement(Button, {
+                                props: {
+                                    title: $t(option.buttonLabel),
+                                    type: "submit",
+                                    size: "normal",
+                                    onClick: this.saveData,
```
bafolts, line 120: I would also use fat-arrow function as it encapsulates the `this` argument.

src/components/OtherPlayer.ts
```diff
@@ -19,11 +19,14 @@ export const OtherPlayer = Vue.component("other-player", {
...
         },
+        getCrossHtml: function() {
+            return "<i class='icon icon-cross' /i>";
+        }
     },
     template: `
         <div> 
             <div v-show="isVisible()" class="other_player_cont menu">
-                <button class="btn btn-lg btn-error other_player_close" v-on:click="hideMe()"><i class="icon icon-cross"></i></button> 
+                <Button size="big" type="close" :onClick="hideMe" align="right" />
```
bafolts, line 29: I assume this also needs the cross HTML provided.

src/components/SelectColony.ts
```diff
@@ -25,7 +25,7 @@ export const SelectColony = Vue.component("select-colony", {
             <colony :colony="colony"></colony>
```
bafolts, line 25: Does this need the `Button` component listed as a dependency?

src/components/WaitingFor.ts
```diff
@@ -89,16 +89,19 @@ export const WaitingFor = Vue.component("waiting-for", {
         }
         const input = new PlayerInputFactory().getPlayerInput(createElement, this.players, this.player, this.waitingfor, (out: Array<Array<string>>) => {
             const xhr = new XMLHttpRequest();
+            const root = (this.$root as any);
+            if (root.isServerSideRequestInProgress) return;
```
bafolts, line 93: I would add a `console.warn` here when this request is ignored. If there were ever a bug regarding how `isServerSideRequestInProgress` is being set we would end up failing silently which could be very difficult to track down. With a warning logged we could spot this warning.

src/components/WaitingFor.ts
```diff
@@ -89,16 +89,19 @@ export const WaitingFor = Vue.component("waiting-for", {
...
+           
+            root.isServerSideRequestInProgress = true;
             xhr.open("POST", "/player/input?id=" + (this.$parent as any).player.id);
             xhr.responseType = "json";
             xhr.onload = () => {
```
bafolts, line 98: `xhr.onerror` needs to also set `isServerSideRequestInProgress` to `false`.

src/components/common/Button.ts
```diff
@@ -0,0 +1,119 @@
...
+                ].includes(item),
+        },
+        onClick: {
+            type: Function,
+            default: () => {},
+        },
+    },
+    methods: {
+        getClasses: function (): string {
+            const classes: Array<String> = ["btn"];
```
bafolts, line 50: This should be `Array<string>` as this pushes `string` to the array and not `String`.

src/components/common/Button.ts
```diff
@@ -0,0 +1,119 @@
...
+
+            //size
+            if (this.size === "tiny") {
+                classes.push("btn-tiny");
+            }
+            if (this.size === "small") {
```
bafolts, line 56: This can also be `else if (this.size === "small")`

src/components/common/Button.ts
```diff
@@ -0,0 +1,119 @@
...
+                }
+            }
+
+            return classes.join(" ");
+        },
+        getDisabled: function (): boolean {
+            if (this.disableOnServerBusy) {
+                return (
+                    this.disabled ||
+                    (this.$root as any).isServerSideRequestInProgress
```
bafolts, line 94: As i am only vaguely familiar with vue.js I am curious. How does this template know it needs to be redrawn when `isServerSideRequestInProgress` changes. Anytime the data changes on the app does every child component redraw?

bafolts (changes requested):
Spotted a few places of concern. 

* We should provide fat arrow function whenever possible as it retains the `this` argument. Some of the onClick handlers were binding to `null` for the `this` argument which leaves me perplexed as to how that works. We may be getting lucky that there only happens to be one instance of that class.

* There seems to be some leftover cross HTML that may have existed before a type of "close" was a thing.

* When isServerSideRequestInProgress changes on app does vue then redraw the entire application? This seems heavy handed for a button that may need to show "loading". Is there a better way to isolate this property perhaps to a `Button` singleton? If vue has a way to find all buttons it may be simpler to iterate over each button and set this property. We should try to keep this property specific to Button.

src/components/LogPanel.ts
chosta: I ditched the getHtml and added types that provide the html conditionally. I should learn to group commit changes better so reviews are easier.

src/components/SelectColony.ts
chosta: Well spotted. I though I tested this manually...I realised it's called on rare occasions like Aridor's first action.

src/components/common/Button.ts
chosta: it should,yes

src/components/AndOptions.ts
chosta: That is good point.

src/styles/payments.less
```diff
@@ -47,6 +47,6 @@
     color: #666;
 }
 
-.max_button {
+/* .max_button {
```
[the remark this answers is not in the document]
chosta: done

src/components/CardsFilter.ts
chosta: There are two ways to pass functions with parameters as props in vue. The first is what you see and null is replacing the event which is the default parameter if we do it like that. This does work. The recommended Vue way is to use $emit but I couldn't make it work. I'll give it one more try and ask for help if I don't succeed.
https://stackoverflow.com/a/39479405/83336

src/components/LogPanel.ts
chosta: Nevermind, I actually left it in one place. Well spotted

src/components/WaitingFor.ts
chosta: Makes sense. Not sure I did it correct as I don't see any warning. Please, check 
0107b3d

src/components/common/Button.ts
chosta: I'll comment below

src/components/CardsFilter.ts
alrusdi: Have you tried something like :onClick="_=>removeCard(cardName)" ?

chosta:
> * We should provide fat arrow function whenever possible as it retains the `this` argument. Some of the onClick handlers were binding to `null` for the `this` argument which leaves me perplexed as to how that works. We may be getting lucky that there only happens to be one instance of that class.

Need to try using $emit for parent to child function with params as a prop, as it is the recommended vue way. This one works too, though. I've done it in react many times and I figured that it shold be safe but I recognize your concerns.



> * There seems to be some leftover cross HTML that may have existed before a type of "close" was a thing.

Done



> * When isServerSideRequestInProgress changes on app does vue then redraw the entire application? This seems heavy handed for a button that may need to show "loading". Is there a better way to isolate this property perhaps to a `Button` singleton? If vue has a way to find all buttons it may be simpler to iterate over each button and set this property. We should try to keep this property specific to Button.

I am using my intuition here again coming from my experience from React. I presume that global state in vue works in a similar way as context (also redux) in React. The approach `it may be simpler to iterate over each button` was initially introduced by @alrusdi who iterated (hardcore) trough all .btn elements and added the "loading" class. This whole PR was done to avoid the approach :) .Now that we have a Button as Vue we might have a solution that iterates the elements just not sure how atm. I'll research a bit but any help here is needed from more experienced Vue users.

src/components/CardsFilter.ts
chosta: what kind of sorcery is this :D

bafolts (approved):
Code wise nothing stands out anymore. Once rolled out if there is some bug I assume it will show up as some click event not working as expected.

PR #1491, opened 2020-10-05
Separating fan-made content

vsrisuknimit:
Separating official expansions from fan-made content.
I would also like to add a button for turning all official expansions at once, but I am not sure yet how to do.

image


vsrisuknimit:
@chosta could you give some hint how to make a toggle to turn on all official expansion?

kberg:
Enabling all official expansions at once can be done in a separate PR. Shouldn't be hard. Maybe I'll do it tonight.

kberg (approved):

vsrisuknimit:
Sounds good. I won't add anything more then.

Issue #1489, opened 2020-10-05
Inventrix gets +/- 2 steps on venus


Maelstrom1:
Incredibly self explanatory. Inventrix was allowed to play cards that should not have been playable because its ability does not include Venus; that's exclusive to Morning Star.

![](https://media.discordapp.net/attachments/692836080323526698/762474620317859860/chrome_RZ2t8UheAr.png)

![](https://media.discordapp.net/attachments/692836080323526698/762474622872190976/chrome_uhXUQ4865X.png)

![](https://media.discordapp.net/attachments/692836080323526698/762474625736114197/chrome_QSvRNEmZa8.png)

alrusdi:
Not a bug actually
https://boardgamegeek.com/thread/1850152/question-authors-adaptation-technology-special-des
**"..Adaptation Technology, Special Design, Inventrix also affect Venus requirements."**

kevinb9n:
Can we call this a display bug then? Either we can

* change the card to say "global params" instead of calling out 3 of the 4, or
* dynamically make the card add "/ Venus" only when Venus is in use?

nwai90:
@kevinb9n Yes, what's not possible on the physical print version can be done on the web app 👍  Agree that stating "global parameters" is a more generic and clearer term.

alrusdi:
Let's reopen it then

kevinb9n:
Thanks! Oh, I said "the card" but I believe we are really talking about three cards.

nwai90:
Implemented

PR #1494, opened 2020-10-05
Fix global requirement adjustment box

vsrisuknimit:
- Changed global requirement from "temp/O2/ocean" to "global requirements" so that it implies Venus ad well.
- Fixed the link of community-corp wiki page.

image

image


vsrisuknimit:
This PR addresses #1489.

kberg:
Speaking as someone who doesn't use Venus, I'd much prefer the original. If you wanted to add "Venus" to the global requirements box, that wouldn't bother me as much. Plus, if cards eventually become data driven like Dani is planning, Venus can be optionally added and removed.

src/HTML_data.ts
```diff
@@ -5231,7 +5231,7 @@ export const HTML_DATA: Map<string, string> =
               </div>
               
               <div class="description ">
-                  (Effect: Your temperature, oxygen and ocean requirements are +2 or -2 steps, your choice in each case.)
+                  (Effect: Your temperature, oxygen, ocean, and venus requirements are +2 or -2 steps, your choice in each case.)
```
alrusdi, line 5234: Please change this text in all translations too

alrusdi (changes requested):

src/HTML_data.ts
kberg: "venus requirements" should be "Venus requirements."

vsrisuknimit: Fixed

alrusdi:
@vsrisuknimit What about Special Design card? Is it included?

alrusdi (approved):

PR #1496, opened 2020-10-05
Add a simple select-all to CreateGameForm.

kberg:


vsrisuknimit:
@kberg it works great. I think it would be better if the toggle is smaller and on the same line as the word "official", toward the right.

kberg:
Funny I thought so too.

The thing about it I don't like is if you select one or two items, then All, and the unselect All, it should restore the values that existed before.

And if you manually select all the items, follow-along by selecting all, et cetera.

vsrisuknimit:
It is fine. Not a big deal. No need to restore. Definitely no need to add None or Invert to this as is the case for corp-filter.

kberg:
I can fix it with this but I don't like using `style` inline.

diff --git a/src/components/CreateGameForm.ts b/src/components/CreateGameForm.ts
index 80945907..dadbfaef 100644
--- a/src/components/CreateGameForm.ts
+++ b/src/components/CreateGameForm.ts
@@ -350,13 +350,14 @@ export const CreateGameForm = Vue.component("create-game-form", {
 
                         <div class="create-game-options-block col3 col-sm-6">
                             <h4 v-i18n>Expansions</h4>
-                            <div class="expansion-label">Official</div>
+                            <div class="expansion-label">Official
 
-                            <label class="form-switch">
+                            <label class="form-switch" style="display: inline; margin-left: 20px;">
                                 <input type="checkbox" name="allOfficialExpansions" v-model="allOfficialExpansions" v-on:change="selectAll()">
                                 <i class="form-icon"></i> <span v-i18n>All</span>
                             </label>
 
+                            </div>
                             <label class="form-switch">
                                 <input type="checkbox" name="corporateEra" v-model="corporateEra">
                                 <i class="form-icon"></i> <span v-i18n>Corporate Era</span>

alrusdi (approved):

PR #1500, opened 2020-10-06
checkmark for side bar global parameters

vsrisuknimit:
What I did:
- Check mark for the global parameters on the side bar

What I would like to do:
-  ~~make the checkmark green instead. I don't know how to do css inside v-html.~
- ~~change GEN count to LAST instead of a number after the three main global parameters are reached.~~
- ~~not use hard code number of max ocean, max oxygen, etc. Is the only way to avoid is to pass more props?~~

image


vsrisuknimit:
I was debugging. I set the max temp and max ocean to be -26 and 2 to test.

nwai90:
@vsrisuknimit You can use the constants `MAX_OCEAN_TILES`, `MAX_OXYGEN_LEVEL` and `MAX_TEMPERATURE`. Same for Venus

vsrisuknimit:
@nwai90 would it know inside vue in preferences.ts?

vsrisuknimit:
Thanks! Using global constants now.

vsrisuknimit:
- Added green check instead.
- Decide not to LAST of GEN, users can already see 3 check marks for the main parameters. That might be something to consider for solo games.

Good to go now.

src/components/Preferences.ts
```diff
@@ -92,28 +92,28 @@ export const Preferences = Vue.component("preferences", {
         },
         getOceanCount: function(): string{
             if (this.oceans === MAX_OCEAN_TILES){
-                return `<img width="18" src="/assets/checkmark.png" alt="completed">`;
+                return `<img width="18" src="/assets/checkmark.png" style="filter: hue-rotate(280deg);" alt="completed">`;
```
alrusdi, line 95: Please move width and filter properties to css (less) file

alrusdi (changes requested):

src/components/Preferences.ts
```diff
@@ -92,28 +92,28 @@ export const Preferences = Vue.component("preferences", {
...
             } else {
                 return `${this.oceans}`;
             } 
         },
         getTemperatureCount: function(): string{
             if (this.temperature === MAX_TEMPERATURE){
-                return `<img width="18" src="/assets/checkmark.png" alt="completed">`;
+                return `<img width="18" src="/assets/checkmark.png" style="filter: hue-rotate(280deg);" alt="completed" style="">`;
```
alrusdi, line 102: Try to make **alt** property translatable.
1) Import $t function from directives/i18n
2) Use it like **:alt="$t('Completed!')"**

alrusdi (changes requested):

src/components/Preferences.ts
vsrisuknimit: Moved it to css

src/components/Preferences.ts
vsrisuknimit: Added translatable alt text.

vsrisuknimit:
Added new icon and moved the png file to assets/misc folder.

alrusdi (approved):

Issue #1497, opened 2020-10-05
Virus should be playable when no one has animals or plants


kevinb9n:
In fact, I saw this bug in Solo, where it's doubly true that Virus should always be playable as long as you have the 1 money.

nwai90:
Fixed by @vsrisuknimit and merged

PR #1503, opened 2020-10-06
virus should always be playable

vsrisuknimit:
Virus should always be playable even when it does have a target. (You may want to play it to gain mc back from Media Group.) Current code prevents that.

### DIFF
--- a/src/cards/Virus.ts
+++ b/src/cards/Virus.ts
@@ -1,4 +1,3 @@
-
 import { IProjectCard } from "./IProjectCard";
 import { Tags } from "./Tags";
 import { CardType } from "./CardType";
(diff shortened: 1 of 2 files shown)


src/cards/Virus.ts
```diff
@@ -20,11 +19,7 @@ export class Virus implements IProjectCard {
     public cardType: CardType = CardType.EVENT;
     public hasRequirements = false;
     public canPlay(player: Player, game: Game): boolean {
-        return this.getPossibleTargetCards(player, game).length +
-            game.getPlayers().filter((p) => 
-                ((p.id !== player.id && !p.plantsAreProtected()) 
-                || p.id === player.id)
-                && p.plants > 0).length > 0;
+        return true;
```
nwai90, line 22: We can also add a simple test for it in `Virus.spec.ts`

```it("Should play", function () {
    expect(card.canPlay()).to.eq(true);
}```

vsrisuknimit: added

vsrisuknimit:
This resolves #1497.

nwai90 (approved):

PR #1507, opened 2020-10-06
Fix toggling of turmoil policies

alrusdi:


nwai90 (approved):

chosta:
Yep, that is on me. I've somehow skipped it when merging to the new without playerIndex state.

PR #1508, opened 2020-10-06
adjust ocean location, fix path for ocean checkmark

vsrisuknimit:
- Fix path to ocean checkmark.
- Adjust the ocean location, number and check mark to be more center.


vsrisuknimit:
@alrusdi

src/components/Board.ts
```diff
@@ -111,8 +111,8 @@ export const Board = Vue.component("board", {
         oceansValue: function() {
             const oceans_count = this.oceans_count || 0;
             const leftover = constants.MAX_OCEAN_TILES - oceans_count;
-            if (leftover === 0) {
-                return `<img width="26" src="/assets/circle-checkmark.png" :alt="$t('Completed!')">`
+            if (leftover === 8) {
```
alrusdi, line 114: It's probably wrong!

alrusdi (changes requested):

src/components/Board.ts
vsrisuknimit: sorry. I added that to debug and forgot to remove it. Very sloppy. My bad.

alrusdi:
it will rewrite new image of mars background. I am rejecting it.

PR #1509, opened 2020-10-06
Fix Mars-pattern baldness

vsrisuknimit:
- Remove the blank space above when Venus is not in used.
- Move Stanford Torus to the top

What I would like to do:
- Conditional placement for Stanford Torus. Move it to the right when Venus next is not in used. I do not know how to do that yet.

Screen Shot 2020-10-06 at 9 14 45 AM

Screen Shot 2020-10-06 at 9 15 01 AM


alrusdi:
I don't like it >< May be it is better to keep Venus bar but grey-out it and add little transparency?

vsrisuknimit:
Does it weird without Venus track? @alrusdi

alrusdi:
> Does it weird without Venus track? @alrusdi

To me it is little strange. Can we try gray-outed version? I will provide a demo

chosta:
> To me it is little strange. Can we try gray-outed version? I will provide a demo

I second greyed out version

vsrisuknimit:
With Venus

image

Without Venus

image

@chosta @alrusdi

alrusdi (approved):

PR #1519, opened 2020-10-06
Fix new graphics on corporations

vsrisuknimit:
- Update the corporations with new graphics.

Some examples:

Old
image

Updated
image
image
image

### DIFF
--- a/src/HTML_data.ts
+++ b/src/HTML_data.ts
@@ -5124,26 +5124,16 @@ export const HTML_DATA: Map<string, string> =
         <div class="corporationEffectBoxLabel">EFFECT</div>
         X<div class="resource heat"></div> :
         <div class="resource money">X</div>
-        <div class="description" style="text-align:center;margin-top:0px;">
+        <div class="description helion-effect">
           (Effect: Your may use heat as MC. You may not use MC as heat.)
         </div>
       </div>
-      <div style="font-size:34px;
-                      width:140px;
-                      text-align: center;
-                      color: #020202;
-                      margin-top:5px;
-                      border:2px solid black;
-                      margin-left:8px;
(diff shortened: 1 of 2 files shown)


src/HTML_data.ts
```diff
@@ -5139,7 +5139,7 @@ export const HTML_DATA: Map<string, string> =
                       box-shadow: 6px 6px 6px grey;
                       border-radius:2px;">
                   helion</div><br>
-      <div class="production-box production-box-size3" style="margin-left:20px;margin-top:-50px;">
+      <div class="production-box production-box-size3" style="margin-left:20px;margin-top:-10px;">
```
alrusdi, line 5142: Please move inline styles to cards.less

alrusdi (changes requested):

src/HTML_data.ts
vsrisuknimit: @alrusdi There are inline styles everywhere in corporation cards. Do you want me to move just this one box specifically or everywhere?

Full Helion code for example:
https://github.com/bafolts/terraforming-mars/blob/6ec98ef60abb6d5988e1d20cf80b00cf4a8a2901/src/HTML_data.ts#L5119-L5150

alrusdi: Not all, but if you take some card to fix please do full cleanup of that card. 
In this case move all inline styles of Helion corp to cards.less
If each of us will do it - someday we will finish that cleanup

vsrisuknimit: @alrusdi I moved all inline style for Helion to card.less. I fixed many other corps too in terms of graphics but I will deal with their styles later if that is okay.

alrusdi (approved):

PR #1521, opened 2020-10-06
Fixes #1518

### LINKED ISSUE #1518: Player name link does not work on game results page
We got rid of player_id there. I don't sure what to do with it. Just remove that link?

pocc:
Removes the link to the player that no longer exists as game has ended. It might make sense to add this back in if we add a replay option from the perspective of a player, like in Board Game Arena.

### DIFF
--- a/src/components/GameEnd.ts
+++ b/src/components/GameEnd.ts
@@ -94,7 +94,7 @@ export const GameEnd = Vue.component("game-end", {
                         </thead>
                         <tbody>
                             <tr v-for="p in getSortedPlayers()">
-                                <td><a :href="'/player?id='+p.id+'&noredirect'" :style="getPlayerColorStyle(p)">{{ p.name }}</a></td>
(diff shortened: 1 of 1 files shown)


bafolts (approved):
This should resolve the bug. Once the game has ended we could provide the player id. We have mapped the color to the id but once the game has ended it should be fine to expose the player ids. Another option for this fix would be to make `id` on `PlayerModel` optional. If the game phase has ended then set `id` to the player id.

https://github.com/bafolts/terraforming-mars/blob/master/server.ts#L775

During the roll-out of using the color instead of the id the id was set to color. As by now every client most likely has updated to a version with color could change the `PlayerModel` id to be `undefined | string`.

pocc:
The annoying part for me is that there is a link, but it doesn't do anything. It would be possible to expose the player ID, but what purpose does that serve?

bafolts:
True, with the new game logs route it may make sense to query for the full game log and display it.

PR #1523, opened 2020-10-07
GameLoader class for loading games

bafolts:
Whenever the server is started we call to load every game from the database. Until every game has loaded players will end up getting 404 responses and may end up losing their game state. As we push changes to master more often games are getting interrupted more often. This change loads a game or player only when requested instead of always loading every game and player.  With this in place the restart shouldn't impact games as much. Instead of returning 404, players wont get a response until the game has been loaded from database.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "extends": "@istanbuljs/nyc-config-typescript"
   },
   "scripts": {
-    "build": "node make_locales.js && eslint src/*.ts && lessc src/styles/common.less styles.css && tsc --build tsconfig.json && webpack --config webpack.config.js",
+    "build": "node make_locales.js && eslint src/*.ts src/database/**/*.ts && lessc src/styles/common.less styles.css && tsc --build tsconfig.json && webpack --config webpack.config.js",
     "start": "node make_locales.js && lessc src/styles/common.less styles.css && tsc --build tsconfig.json && webpack --config webpack.config.js && node dist/server.js",
     "pretest": "tsc --build tsconfig-test.json",
     "test": "mocha --recursive dist/tests",
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -366,7 +366,11 @@ export class Game implements ILoadable<SerializedGame, Game> {
 
     // Function to retrieve a player by it's id
     public getPlayerById(id: string): Player {
-      return this.players.filter(p => p.id === id)[0];
+      const player = this.players.find(p => p.id === id);
+      if (player === undefined) {
+        throw new Error(`player ${id} does not exist on game ${this.id}`);
+      }
+      return player;
     }
 
     // Function to return an array of players from an array of player ids
--- a/src/database/GameReloader.ts
+++ b/src/database/GameReloader.ts
@@ -1,0 +1,108 @@
+
+import { Color } from "../Color";
+import { Database } from "./Database";
+import { Game } from "../Game";
+import { Player } from "../Player";
+
+export class GameReloader {
+    private loaded = false;
+    private loading = false;
+    private readonly pendingGame = new Map<string, Array<(game: Game | undefined) => void>>();
+    private readonly pendingPlayer = new Map<string, Array<(game: Game | undefined) => void>>();
+    constructor (private games: Map<string, Game>, private playersToGame: Map<string, Game>) {}
+
+    public start(): void {
+        if (this.loaded === true) {
+            console.warn("already loaded, ignoring");
+        } else if (this.loading === true) {
+            console.warn("already loading, ignoring");
+            return;
+        }
+        this.loading = true;
+        this.loadAllGames();
+    }
+
+    public getByGameId(gameId: string, cb: (game: Game | undefined) => void): void {
+        if (this.loaded === true) {
+            cb(this.games.get(gameId));
+            return;
+        }
+        const pendingGame = this.pendingGame.get(gameId);
+        if (pendingGame !== undefined) {
+            pendingGame.push(cb);
+        } else {
+            this.pendingGame.set(gameId, [cb]);
+        }
+    }
+
+    public getByPlayerId(playerId: string, cb: (game: Game | undefined) => void): void {
+        if (this.loaded === true) {
+            cb(this.playersToGame.get(playerId));
+            return;
+        }
+        const pendingPlayer = this.pendingPlayer.get(playerId);
+        if (pendingPlayer !== undefined) {
+            pendingPlayer.push(cb);
+        } else {
+            this.pendingPlayer.set(playerId, [cb]);
+        }
+    }
+
+    private onGameLoaded(gameId: string, playerId: string): void {
+        const pendingGames = this.pendingGame.get(gameId);
+        if (pendingGames !== undefined) {
+            for (const pendingGame of pendingGames) {
+                pendingGame(this.games.get(gameId));
+            }
+            this.pendingGame.delete(gameId);
+        }
+        const pendingPlayers = this.pendingPlayer.get(playerId);
+        if (pendingPlayers !== undefined) {
+            for (const pendingPlayer of pendingPlayers) {
+                pendingPlayer(this.playersToGame.get(playerId));
+            }
+            this.pendingPlayer.delete(playerId);
+        }
+    }
+
+    private onAllGamesLoaded(): void {
+        this.loading = false;
+        this.loaded = true;
+    }
+
+    private loadAllGames(): void {
+        Database.getInstance().getGames((err, allGames) => {
+            if (err) {
+                console.error("error loading all games", err);
+                this.onAllGamesLoaded();
+                return;
+            }
+            let loaded = 0;
+            allGames.forEach((game_id) => {
+                const player = new Player("test", Color.BLUE, false, 0);
+                const player2 = new Player("test2", Color.RED, false, 0);
+                const gameToRebuild = new Game(game_id, [player, player2], player);
+                Database.getInstance().restoreGameLastSave(
+                    game_id,
+                    gameToRebuild,
+                    (err) => {
+                        loaded++;
+                        if (err) {
+                            console.error(`unable to load game ${game_id}`, err);
+                            return;
+                        }
+                        console.log(`load game ${game_id}`);
+                        this.games.set(gameToRebuild.id, gameToRebuild);
+                        for (const player of gameToRebuild.getPlayers()) {
+                            this.playersToGame.set(player.id, gameToRebuild);
+                            this.onGameLoaded(gameToRebuild.id, player.id);
+                        }
+                        if (loaded === allGames.length) {
+                            this.onAllGamesLoaded();
+                        }
+                    }
+                );
+            });
+        });
+    }
+}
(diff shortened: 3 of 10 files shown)


src/database/GameReloader.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    private loadAllGames(): void {
+        Database.getInstance().getGames((err, allGames) => {
+            if (err) {
+                console.error("error loading all games", err);
+                this.onAllGamesLoaded();
+                return;
+            }
+            let loaded = 0;
```
kberg, line 80: loaded locally here is a number hiding a class instance of the same name. eslint have an opinion about this?

src/database/GameReloader.ts
```diff
@@ -0,0 +1,108 @@
...
+        }
+    }
+
+    private onAllGamesLoaded(): void {
+        this.loading = false;
+        this.loaded = true;
+    }
+
+    private loadAllGames(): void {
+        Database.getInstance().getGames((err, allGames) => {
```
kberg, line 74: Does getGames go in a particular order? I wouldn't be surprised if loading newer games first would be more effective.

src/database/GameReloader.ts
```diff
@@ -0,0 +1,108 @@
+
+import { Color } from "../Color";
+import { Database } from "./Database";
+import { Game } from "../Game";
+import { Player } from "../Player";
+
+export class GameReloader {
```
kberg, line 7: As a WIP, seems pretty good. What are your plans as far as tests in the long run?

src/database/GameReloader.ts
bafolts: There is the `no-shadow` rule which may catch this for `let`. I can re-name if the linting doesn't catch.

https://eslint.org/docs/rules/no-shadow

Since `src/database` is rarely touched I will also apply the eslint to `src/database/**/*.ts` while here. :+1:

src/database/GameReloader.ts
bafolts: Well I'll be. We don't sort at all right now. Sorting by newer games first should be a simple solution to ensure games being played aren't left waiting long if at all and buy time. Will most likely start by doing this sort at first for this change. Right now we dump the JSON into the database in the least efficient way which makes trying to find the player id for a game slow. Need to store player ids with each game so that when we only have player id we can find the game to load.

src/database/GameReloader.ts
bafolts: I can write unit tests for this :+1: As things move out of `server` they are easier to test in isolation like this.

src/database/GameReloader.ts
kberg: Sorry I should add, my main issue is that they were also different types. Anyway, thanks.

src/database/GameReloader.ts
kberg: But until now, that wasn't an issue, so I'm not surprised. Anyway, thanks.

src/database/GameReloader.ts
kberg: Thanks. BTW, not exactly the same kind of test, but I wrote a unit test for Ares to make sure my wacky board behavior change didn't cause issues in between save and load: https://github.com/bafolts/terraforming-mars/blob/ares/tests/ares/AresStorageTest.spec.ts

I haven't seen much in terms of serialization tests, so at least now there's a pattern, in case you want to copy it.

kberg: LMK when you want me to re-review this.

src/Game.ts
```diff
@@ -366,7 +366,11 @@ export class Game implements ILoadable<SerializedGame, Game> {
 
     // Function to retrieve a player by it's id
     public getPlayerById(id: string): Player {
-      return this.players.filter(p => p.id === id)[0];
+      const player = this.players.find(p => p.id === id);
+      if (player === undefined) {
+        throw new Error(`player ${id} does not exist on game ${this.id}`);
```
bafolts, line 371: This sucks to have to add but filtering and returning the 0th index item can return `undefined`. This return type is wrong. Various places calling this all assume they get a result. I threw error for time being but this return type needs to change to `Player | undefined`.

src/database/GameReloader.ts
bafolts: I am not in love with the name of this class `GameReloader` but I added tests. Will remove the `wip` tag as I think this is ready. Will keep do not merge as I want to run more manual tests for my sanity before merging.

src/database/GameReloader.ts
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    private onAllGamesLoaded(): void {
+        this.loadingGames = false;
+        this.loadedGames = true;
+        // any pendingPlayer or pendingGame callbacks
+        // are waiting for a train that is never coming
+        // send them packing. call their callbacks with
+        // undefined and remove from pending
+        for (const pendingGame of Array.from(this.pendingGame.values())) {
```
bafolts, line 87: while reading through this I need to follow my own comment and remove these from the pending maps as well.

package.json
```diff
@@ -34,7 +34,7 @@
     "extends": "@istanbuljs/nyc-config-typescript"
   },
   "scripts": {
-    "build": "node make_locales.js && eslint src/*.ts && lessc src/styles/common.less styles.css && tsc --build tsconfig.json && webpack --config webpack.config.js",
+    "build": "node make_locales.js && eslint src/*.ts src/database/**/*.ts && lessc src/styles/common.less styles.css && tsc --build tsconfig.json && webpack --config webpack.config.js",
```
kberg, line 37: Unrelated, but is it right for build to have an eslint step but not start?

src/database/GameReloader.ts
```diff
@@ -0,0 +1,136 @@
...
+import { Game } from "../Game";
+import { Player } from "../Player";
+
+export class GameReloader {
+    private loadedGames = false;
+    private loadingGames = false;
+    private readonly games = new Map<string, Game>();
+    private readonly pendingGame = new Map<string, Array<(game: Game | undefined) => void>>();
+    private readonly pendingPlayer = new Map<string, Array<(game: Game | undefined) => void>>();
+    private readonly playersToGame = new Map<string, Game>();
```
kberg, line 13: Could just be your personal style, but you don't mean playerToGame?

src/database/GameReloader.ts
```diff
@@ -0,0 +1,136 @@
...
+        }
+        const pendingGame = this.pendingGame.get(gameId);
+        if (pendingGame !== undefined) {
+            pendingGame.push(cb);
+        } else {
+            this.pendingGame.set(gameId, [cb]);
+        }
+    }
+
+    public getByPlayerId(playerId: string, cb: (game: Game | undefined) => void): void {
```
kberg, line 50: What do you think about playersToGame being playerToGameId, and this function is getGameIdByPlayerId? I think this is just my style peeking through, and nothing else.

src/database/GameReloader.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    public addGame(game: Game): void {
+        this.games.set(game.id, game);
+        for (const player of game.getPlayers()) {
+            this.playersToGame.set(player.id, game);
+        }
+    }
+
+    public getGameIds(): Array<string> {
+        return Array.from(this.games.keys());
```
kberg, line 34: How much do you care that getGameIds() before the initial query returns will return an empty list?

src/database/GameReloader.ts
kberg: Just GameLoader. It'll be fine!

src/database/GameReloader.ts
```diff
@@ -0,0 +1,136 @@
...
+        }
+        const pendingPlayer = this.pendingPlayer.get(playerId);
+        if (pendingPlayer !== undefined) {
+            pendingPlayer.push(cb);
+        } else {
+            this.pendingPlayer.set(playerId, [cb]);
+        }
+    }
+
+    private onGameLoaded(gameId: string, playerId: string): void {
```
kberg, line 63: What's the typescript concurrency story here? Is that something to worry about here? /ignorant

tests/database/GameReloader.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+        const expectedGameId = "foo";
+        expectedGameIds = [expectedGameId];
+        const reloader = new GameReloader();
+        let actual: Game | undefined;
+        reloader.getByGameId(expectedGameId, (game) => {
+            actual = game;
+        });
+        expect(actual).to.eq(undefined);
+        reloader.start();
+        expect(actual).not.to.eq(undefined);
```
kberg, line 45: expect(actual).to.be.undefined.
expect(actual).to.not.be.undefined

I found these and they seem really nice. Not to mention

https://medium.com/building-ibotta/testing-arrays-and-objects-with-chai-js-4b372310fe6d

for things like the expectedGameIds two lines down.  .to.have.members ignores array order, which is kind of nice!

kberg:
I don't see anything here I dislike (besides the comments I left.) Thank you for the tests - they're good.

src/database/GameReloader.ts
bafolts: The sequel can use GameReloaded

package.json
bafolts: The `start` is currently for heroku or anyone who only wants to run the game. The eslint step is only needed for people developing the application. This allows `eslint` to be in `devDependencies` instead of `dependencies`. This is less that heroku or anyone has to download each time when running to install only `dependencies`. There are a few other `dependencies` which should be moved to `devDependencies` for this separation.

src/database/GameReloader.ts
bafolts: Our styles are in alignment. That is a better name.

src/database/GameReloader.ts
bafolts: The only route which ends up calling for this data is `/api/games` which is an administration route. I ported this from `server.ts` since it needed access to `this.games`. I preferred this this return the games Map.

src/database/GameReloader.ts
bafolts: Javascript is single threaded. The only worry I have is blocking this single thread. The calls to the postgres client are asynchronous as they happen through I/O off the event loop and when queries complete will put our callback on the event loop and work through sequentially. Javascript has opposite problem from multi threaded environments which have to worry about synchronization. Always trying to get asynchronous and not block everything else.

tests/database/GameReloader.spec.ts
bafolts: Ah thank you. I work in various javascript environments on daily basis, I tried `toBeDefined` which exists in `jasmine` and then just went with what I saw work quickly. I will switch to this.

bafolts:
OK. Removed do not merge label. This should be ready to go. Ran it through a few game actions. Loaded a game I saw come in on the logs before it had loaded. Loaded a game that would never exist and finally got 404 once everything loaded. While loading able to create a new game.

kberg (approved):

kberg:
This is nice.

PR #1525, opened 2020-10-07
Fresh batch of changes for Ares.

kberg:
Sorry, I neglected to separate upstream master. I think that ship has sailed for this pass.


alrusdi:
@kberg why don't you write directly to bafolts:ares? You don't need the PR to write to non-master branches.

kberg:
> @kberg why don't you write directly to bafolts:ares? You don't need the PR to write to non-master branches.

Ah didn't realize I could. OK, in it goes (after I retitle it.)

Issue #1457, opened 2020-10-02
I think tag requirements should show the icons


kevinb9n:
When viewing a card that requires a Jovian tag, there should be *some* way to find out what a Jovian tag looks like.

(I'm hoping to use this app to teach friends who don't already know the game, so I'm looking for interface elements that may trip them up.)

If this isn't possible on the card itself, then there should be *something* you can use to find out the icon-name correspondence for tags, such as clicking a tag in the tag tracker.

vsrisuknimit:
@kevinb9n would alt text on the icons in the tags overview table help?

alrusdi:
We have small Jovian icon. Why not to use it in requirements box instead of text?

kevinb9n:
As long as there is *some* way to find this correspondence for both desktop and mobile users then I think the bug is probably resolved.

It still seems inconsistent that only in the case of requirements the cards diverge from their physical counterparts and choose text over icons. That text is then just duplicated below. If icons were used in requirement boxes, then you would know both icon form and word form from just looking at the card itself.

vsrisuknimit:
@kevinb9n Does this help with your issue? http://terraforming-mars.herokuapp.com/help-iconology

I will work on integrating tags into the requirement box next.

kevinb9n:
That guide is excellent - how can we make that available for players to pop up as needed?
I have minor suggestions about it if you want them somewhere. :-)

vsrisuknimit:
Hi @kevinb9n here is fine.
- I hope to edit the help page further. Add a few more things.
- I would like to add a shortcut to the help page on the side bar. A question mark button.

kevinb9n:
Ah, well, I already did put them onto the PR before I saw this. And it's a lot, but LMK if you want me to make em myself. (I should really start being useful around here.)

A ? in the sidebar is great - so the user can pull it up as they need it.

kevinb9n:
I still wonder why we use text for requirements.

One weird effect of the current decision is that Celestic is claiming it won't draw the card Aerosport Tournament, but it actually does.
https://ssimeonoff.github.io/cards-list#CORP14#214

kberg:
I basically agree with this request. Commenting to make it unstale.

kberg:
I really want to implement this, it's bubbling up near the top of my list.

kberg:
This is fixed!

kevinb9n:
FWIW, I really like the results.

PR #1530, opened 2020-10-07
help page

vsrisuknimit:
What I did:
- Add a page that show game iconography

image

What I would like to do next:
- Add a link to the side bar. A question mark button. Put it at the very bottom but above the gear. 
- Click on that will link to this page
- Make the global parameter center aligned.
- Fix this page in general cause I don't know what I am doing.

This will partially address #1457 .

### DIFF
--- a/src/components/HelpIconology.ts
+++ b/src/components/HelpIconology.ts
@@ -11,28 +11,28 @@ export const HelpIconology = Vue.component("help-iconology", {
     template: `
         <div class="help-iconology-container">
             <div class="row" style="display: flex;">
-                <div class="column" style="flex: 25%; padding: 10px;">
+                <div class="column">
                     <section class="help-iconology-tags">
                     <h2>Card Tags</h2>
-                        <div class="resource-tag tag-building"></div> : Building tag<br>
-                        <div class="resource-tag tag-space"></div> : Space tag<br>
-                        <div class="resource-tag tag-science"></div> : Science tag<br>
-                        <div class="resource-tag tag-power"></div> : Power tag<br>
-                        <div class="resource-tag tag-earth"></div> : Earth tag<br>
-                        <div class="resource-tag tag-jovian"></div> : Jovian tag<br>
-                        <div class="resource-tag tag-venus"></div> : Venus tag<br>
-                        <div class="resource-tag tag-plant"></div> : Plant tag<br>
-                        <div class="resource-tag tag-microbe"></div> : Microbe tag<br>
-                        <div class="resource-tag tag-animal"></div> : Animal tag<br>
-                        <div class="resource-tag tag-city"></div> : City tag<br>
-                        <div class="resource-tag tag-wild"></div> : Wild tag<br>
-                        <div class="resource-tag tag-event"></div> : Event tag<br>
+                        <div class="resource-tag tag-building"></div> : Building<br>
+                        <div class="resource-tag tag-space"></div> : Space<br>
+                        <div class="resource-tag tag-science"></div> : Science<br>
+                        <div class="resource-tag tag-power"></div> : Power<br>
+                        <div class="resource-tag tag-earth"></div> : Earth<br>
+                        <div class="resource-tag tag-jovian"></div> : Jovian<br>
+                        <div class="resource-tag tag-venus"></div> : Venus<br>
+                        <div class="resource-tag tag-plant"></div> : Plant<br>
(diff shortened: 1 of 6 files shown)


src/components/HelpIconology.ts
```diff
@@ -0,0 +1,75 @@
...
+export const HelpIconology = Vue.component("help-iconology", {
+    components: {
+        Card
+    },
+    methods: {
+        
+    },
+    template: `
+        <div class="help-iconology-container">
+            <div class="row" style="display: flex;">
```
kberg, line 13: How comfortable are you with defining CSS and classes? If you're not, I can step in. What I'd suggset is adding help-iconology.less in src/styles with this content

.help-iconology-container .row {
  display: flex;
}

.help-iconology-container .column {
  flex: 25%;
padding: 10px;
}

src/components/HelpIconology.ts
```diff
@@ -0,0 +1,75 @@
...
+                <div class="column" style="flex: 25%; padding: 10px;">
+                    <section class="help-iconology-tags">
+                    <h2>Card Tags</h2>
+                        <div class="resource-tag tag-building" /> : Building tag<br>
```
kberg, line 17: No need to add the word "tag" after each as it's obvious they're tags, being in the tags section.

kberg (changes requested):
LMK if you have questions.

kberg:
Oh also!

This is very helpful -- thank you! I could see adding this off the player's page with an <a target="_blank"> tag.

src/components/HelpIconology.ts
vsrisuknimit: Moved

src/components/HelpIconology.ts
```diff
@@ -0,0 +1,75 @@
...
+                        <div class="resource-tag tag-animal" /> : Animal tag<br>
+                        <div class="resource-tag tag-city" /> : City tag<br>
+                        <div class="resource-tag tag-wild" /> : Wild tag<br>
+                        <div class="resource-tag tag-event" /> : Event tag<br>
+                    </section>
+                </div>
+                <div class="column" style="flex: 25%; padding: 10px;">                        
+                    <section class="help-iconology-resources">
+                    <h2>Resources</h2>
+                        <div class="resource money" /> : Mega credits<br>
```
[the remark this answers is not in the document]
vsrisuknimit: fixed

src/components/HelpIconology.ts
vsrisuknimit: deleted

kberg (approved):

kberg:
This is terrific. Thank you!

kevinb9n:
Yes, this is really fantastic!

I have small suggestions. Sorry for picking, I really do appreciate this. Please ignore any of the below that you don't personally find helpful.

First some slight corrections:

* It's just "terraform rating".
* "Megacredits" is one word. The rules sometimes camel-case it "MegaCredits" but not always. Perhaps put "(M€)" after it.
* TR/Colony/TradeFleet need some heading of their own ("Other"?), as they're not resources.
* Visual: dunno if you noticed the Greenery tile is randomly wider than the others.

Possible additions - just ignore the ones you find unnecessary?

* Project card (seen on map and some cards)
* Delegate
* Victory Point should be in here using whatever style is currently in use.
* Perhaps a hollow production box can be shown, labeled as "Production"?

Suggested changes:

* I think you can omit the Fighter icon for the same reason you're omitting Camp and Preservation resources - they're single-use.
* Separate sections labeled "Standard Resources" and "Card-based Resources" would be helpful (e.g. for PhilAres).
* I would like it very much if we could set "Wild tag" and "Event tag" apart from the rest somehow, even if just by a little extra space, as it is sometimes important to understand that they *are not tags*. I know I am a little more bothered by this than most, but I actually always put "tag" in quote marks when talking about these, or use the phrase "Event symbol"/"Wild symbol".
* Among the 11 actual tags, it would also be useful to set Venus apart a bit, even if not actually labeling it "(expansion only)". (I'd still keep Earth and Jovian just before it.) The effect of this and the previous suggestion is to make it easier to understand why the most M€ production you can ever get from Interplanetary Trade in a no-expansions game is 10. That's _very_ nonobvious as things stand.
* SUPER NITPICK ALERT, under global parameters, it's a little better to say "Oceans" with an s.

kevinb9n:
Important addendum: if you prefer, please feel free to just tell me which ones you disagree with and wait for me to try hacking up the file myself.

PR #1536, opened 2020-10-08
Modifies the create game form styles

Lynesth:
A few images will probably be the best to show what it does. The changes are also much clearer in French because sentences are longer but I wouldn't want to hurt your eyes with it everywhere :)

Please test it and let me know if I missed something, like the reason those widths were set.

## Current version at different sizes:
ApplicationFrameHost_2020-10-08_13-32-50
ApplicationFrameHost_2020-10-08_13-32-23
ApplicationFrameHost_2020-10-08_13-32-35
-----
## This PR at same sizes:
ApplicationFrameHost_2020-10-08_13-30-58
ApplicationFrameHost_2020-10-08_13-31-38
ApplicationFrameHost_2020-10-08_13-32-02

### DIFF
--- a/src/components/CreateGameForm.ts
+++ b/src/components/CreateGameForm.ts
@@ -341,7 +341,7 @@ export const CreateGameForm = Vue.component("create-game-form", {
                     </div>
 
                     <div class="columns">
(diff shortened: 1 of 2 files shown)


kberg:
:+1:

alrusdi (approved):

PR #1541, opened 2020-10-08
Add translation to log messages parameters 

Lynesth:
This will translate every parameters of the log messages except if their type is set as either `PLAYER` or `RAW_STRING` (new type currently only used for the game id).

Tested it both in English and in French, seems to work without issues.

### DIFF
--- a/src/components/LogPanel.ts
+++ b/src/components/LogPanel.ts
@@ -28,6 +28,7 @@ export const LogPanel = Vue.component("log-panel", {
             }
         },
         parseCardType: function(cardType: CardType, cardName: string) {
+            cardName = $t(cardName);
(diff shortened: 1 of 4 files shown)


src/components/LogPanel.ts
```diff
@@ -28,6 +28,7 @@ export const LogPanel = Vue.component("log-panel", {
             }
         },
         parseCardType: function(cardType: CardType, cardName: string) {
+            cardName = $t(cardName);
             if (cardType === CardType.EVENT) {
                 return "<log-card class=\"background-color-events\">"+cardName+"</log-card>";
```
alrusdi, line 33: Shouldn't we translate cardName in this block too?

alrusdi (approved):

Issue #1371, opened 2020-09-23
Giant Ice Asteroid vs Arctic Algae


kevinb9n:
Lost the game id, but a couple nights ago I played Giant Ice Asteroid on my wife, who owned 3 plants. It seems to have tried to deduct 6 plants first (actually removing all 3 that she had), and *then* process Arctic Algae to give her the 4 plants from my two ocean placements. So she ended up with 4 plants, when I thought she should have ended up with only 1.

I believe the official rule is that the active player gets to choose the order, but the game can easily guess which order I'd want, or it could handle each step of the action card one by one.

I suppose the current behavior is consistent with an approach of executing an entire action atomically, then checking for any effects. And that could be intentional. A bummer in this case though!

kberg:
Checking the rulebook, page 9, A, 2. 

"Then the lower panel of the card shows you any immediate effects. These effects (and other effects triggered by the card you just played) are performed in any order you choose."

kberg:
I would imagine getting that to work in a way you expect would require a meaningful change to the way the code is structured (from declarative to storing a list of effects, and then changing the UI to ask for resolution order. Probably not trivial, but a fundamental change to game structure.

kevinb9n:
Just to be clear, I'm also okay with it picking an order itself, if it could pick one more obviously favorable to the active player.

vsrisuknimit:
@kevinb9n from #703 It was fixed. The current order is temp, ocean, ocean and plant destruction. 

See https://github.com/bafolts/terraforming-mars/blob/95a79236dee9857b863d0bd357642fb39d66cc74/src/cards/GiantIceAsteroid.ts#L31

Are you sure about what you saw?

kevinb9n:
Well, I feel about as sure of it as anyone feels about anything. :-) We stared at it for a while. I fairly distinctly recall the log even saying explicitly that 3 plants were removed.

kevinb9n:
Yup, just saw it again. I think this screencap proves it?

![image](https://user-images.githubusercontent.com/934551/94335795-c6923480-ff92-11ea-86a2-1fcb1111faf9.png)

kevinb9n:
Just to be very clear, Deimos Down *did* take 6 plants, leaving me with zero, then I had 2/4/6 from Arctic Algae as she placed her oceans, and then... nothing, I just got to keep them!

nwai90:
Confirmed - this issue is reproducible.

**Scenario:**
[...]

PR #1544, opened 2020-10-08
Fix resource decrease interrupts

Lynesth:
This is in relation with #1371 (and other reports on discord) where interaction between Arctic Algae and multiple astronomical objects such as Comet or Giant Ice Asteroid is not working as intended.

The issue is that interrupts are computed when the card is played. So we end up removing less plants than we should have (probably other interactions could be impacted by that, I'm not sure tbh).

This PR - please bear with me - fixes this by using a method that was used only for the Pluto Colony interrupt as far as I can tell.
This effectively moves candidates computation to the moment the interrupt is actually processed.

Note that I never used Typescript before stumbling upon this repo so there might be things I did really wrong there, feel free to edit the PR directly. That might also not be the best way to fix this, I don't really know about that though.

Tested it a bit on my end, seems to work and didn't break anything in my test games.

-------

Edit: It breaks tests. I might take a look at it tomorrow.

### DIFF
--- a/src/interrupts/SelectResourceDecrease.ts
+++ b/src/interrupts/SelectResourceDecrease.ts
@@ -9,7 +9,7 @@ import { SelectOption } from "../inputs/SelectOption"
 import { SelectPlayer } from "../inputs/SelectPlayer";
 
 export class SelectResourceDecrease implements PlayerInterrupt {
-    public playerInput: PlayerInput;
+    public playerInput: PlayerInput | undefined;
     constructor(
         public player: Player,
         public candidates: Array<Player>,
@@ -37,7 +37,7 @@ export class SelectResourceDecrease implements PlayerInterrupt {
 		}
 
 		if (this.candidates.length === 0) {
-			return;
+			this.playerInput = undefined;
 		} else if (this.candidates.length === 1) {
 			const qtyToRemove = Math.min(this.candidates[0].plants, this.count);
 			if (this.resource === Resources.PLANTS) {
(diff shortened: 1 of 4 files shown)


Lynesth:
Actually broke something, so I fixed it.

I'm confident (as much as I can be at 3:46 am) that this should now be working as intended and that tests fail because of how they are checking things (like checking that no interrupts are created if there's no targer while now there's still one that's created even if it doesn't really fire afterwards).

src/interrupts/SelectResourceDecrease.ts
```diff
@@ -26,4 +28,58 @@ export class SelectResourceDecrease implements PlayerInterrupt {
...
+
+    public beforeAction(): void {
+		if (this.resource === Resources.PLANTS) {
+			this.candidates = this.game.getPlayers().filter((p) => p.id !== this.player.id && !p.plantsAreProtected() && p.getResource(this.resource) > 0);
+		} else {
+			this.candidates = this.game.getPlayers().filter((p) => p.id !== this.player.id && p.getResource(this.resource) > 0);
+		}
+
+		if (this.candidates.length === 0) {
+			this.playerInput = undefined;
```
bafolts, line 40: I can help out here as this change makes sense to me. I want a goal of this project to be ease of contributing so hopefully it wasn't too painful to add this typescript!

I am weary to introduce setting `playerInput` to `undefined` once it is in the interrupts array. It seems we already have a lot of code duplication around `beforeAction` which we should put into some function. Since this `beforeAction` modifies state I don't think it should return a value. I am thinking we introduce `canAct` here. If after `beforeAction` the result of `canAct` were false we would ignore the interrupt.

I will make a branch on `bafolts` to ease with contributions. After that can you change the target branch to that branch to get the code there?

Lynesth: @bafolts I actually rewrote everything today.
Would you be available to talk about it real quick on Discord before I write too much things here ?

bafolts:
I moved this to `bafolts`

https://github.com/bafolts/terraforming-mars/tree/fix-resource-decrease-interrupts

PR #1547, opened 2020-10-09
Reorder fields in Game and Player into logical clusters

kberg:
No functional change at all, purely so it's clear what's in Game and Player and why.

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -56,57 +56,86 @@ import { ColonyModel } from "./models/ColonyModel";
 export type PlayerId = string;
 
 export class Player implements ILoadable<SerializedPlayer, Player>{
-    public corporationCard: CorporationCard | undefined = undefined;
     public id: PlayerId;
-    public canUseHeatAsMegaCredits: boolean = false;
-    public shouldTriggerCardEffect: boolean = true;
-    public plantsNeededForGreenery: number = 8;
+    private waitingFor?: PlayerInput;
+    private waitingForCb?: () => void;
+  
+    // Corporate identity
+    public corporationCard: CorporationCard | undefined = undefined;
+    // Used only during set-up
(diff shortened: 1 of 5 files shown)


src/Player.ts
```diff
@@ -56,57 +56,86 @@ import { ColonyModel } from "./models/ColonyModel";
...
+    // Cards
+    public dealtCorporationCards: Array<CorporationCard> = [];
+    public dealtProjectCards: Array<IProjectCard> = [];
+    public dealtPreludeCards: Array<IProjectCard> = [];
     public cardsInHand: Array<IProjectCard> = [];
     public preludeCardsInHand: Array<IProjectCard> = [];    
     public playedCards: Array<IProjectCard> = [];
     public draftedCards: Array<IProjectCard> = [];
     public removedFromPlayCards: Array<IProjectCard> = [];
+    // TODO(kberg): Recast to Map<CardName, number>, make sure it survives JSONification.
```
bafolts, line 107: Only typescript knows about `CardName`. The generated javascript for the enum ends up as (from dist/src/CardName.js)

```js
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
var CardName;
(function (CardName) {
    CardName["ACQUIRED_COMPANY"] = "Acquired Company";
    CardName["ADAPTATION_TECHNOLOGY"] = "Adaptation Technology";
```
Should be fine to use `CardName` here if this compiles.

bafolts (changes requested):
Should be fine to implement the TODO. Using `CardName` in the `Map` will be great change. May get compile error on the way back from database.

https://github.com/bafolts/terraforming-mars/blob/master/src/Player.ts#L2223

src/Player.ts
kberg: For sure, but I'll do it in separate pull request.

bafolts (approved):

Issue #1474, opened 2020-10-04
Permit playing an action that legally has no effect


kevinb9n:
thenewjuniorexecutive Today at 10:33 AM
"Taking Directed Impactors action with a resource on it and temp maxed forces you to pay for another resource. Removing a resource is a valid stall in this situation since raising maxed params is not required to complete an action.'

I believe this is correct, and I think I saw this with Aquifer Pumping as well.

kevinb9n:
Oh: the same applies to Symbiotic Fungus when you have no microbe-accepting card. Etc.

(The third exception to the no-partial-actions rule is removing resources from an opponent, I just don't know of any actions that have that as their right-hand-side.)

vsrisuknimit:
@kevinb9n Ants and Predators can remove their own and add back to themselves.

kevinb9n:
That is correct... but is it related to this issue? That describes an action that has two effects, which both actually happen.

vsrisuknimit:
@kevinb9n sorry I misunderstood your third exception. Please ignore me.

bafolts:
This change will most likely need to fix various cards as the issue is spotted. As the `canPlay` and `canAct` methods were implemented this changed. There was a mindset that if the action doesn't make sense we should not show the card as a user most likely doesn't want to play it. I agree though that we should allow any action that is legally possible even if it isn't clear why someone would want to take that action.

nwai90:
Yes, stall actions do have some value in late game. Will fix along the way as we discover them

nwai90:
Fixed and merged

PR #1548, opened 2020-10-09
Permit stall actions for some cards

nwai90:
Ref: https://github.com/bafolts/terraforming-mars/issues/1474

Minimal change. Exit condition for max is already handled in `game.addOceanInterrupt` and `game.increaseTemperature`.

### DIFF
--- a/src/cards/AquiferPumping.ts
+++ b/src/cards/AquiferPumping.ts
@@ -20,9 +20,9 @@ export class AquiferPumping implements IActionCard, IProjectCard {
     }
     public canAct(player: Player, game: Game): boolean {
       const oceansMaxed = game.board.getOceansOnBoard() === MAX_OCEAN_TILES;
-      if (oceansMaxed) return false;
-
       let oceanCost = 8;
+
+      if (oceansMaxed) return player.canAfford(oceanCost, game, true, false);
 
       if (PartyHooks.shouldApplyPolicy(game, PartyName.REDS)) {
         return player.canAfford(oceanCost + REDS_RULING_POLICY_COST, game, true, false);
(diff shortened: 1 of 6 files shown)


src/cards/AquiferPumping.ts
```diff
@@ -20,10 +20,10 @@ export class AquiferPumping implements IActionCard, IProjectCard {
     }
     public canAct(player: Player, game: Game): boolean {
       const oceansMaxed = game.board.getOceansOnBoard() === MAX_OCEAN_TILES;
-      if (oceansMaxed) return false;
-
       let oceanCost = 8;
 
+      if (oceansMaxed) return player.canAfford(oceanCost, game, true, false);
```
kberg, line 25: So let me make sure I understand: a player is going to spend 8MC and yet not be able to place the ocean? I appreciate the ability to take no-op actions, is this really such a valid place? If it is I trust you.

But more important, would you add a comment above this, and each canPlay method that describes why the behavior is non-streamlined? `// #1474: Permit playing an action that legally has no effect` for instance.

nwai90: @kberg Yes, it has been pointed out that this could have value as a stall action. This is permitted as part of the official 3 exceptions of "stuff that can be ignored _if you cannot_ do them" - (1) gaining resources on cards, (2) raising a global parameter, (3) removing another player's plants

kberg: *fascinating*

kberg (approved):
Consider the comment I suggested. Or not -- it sounds like something experienced players ought to know.

nwai90:
> Consider the comment I suggested. Or not -- it sounds like something experienced players ought to know.

Do we really need those extra comments though? Not sure about this one; the one-liner is quite self-explanatory imo

kberg:
> > Consider the comment I suggested. Or not -- it sounds like something experienced players ought to know.
> 
> Do we really need those extra comments though? Not sure about this one; the one-liner is quite self-explanatory imo

We don't. I hoped it would be clear by my comment that we don't. I'm good - thak you.

PR #1549, opened 2020-10-09
Card rewards are not drawn until the end of an action.

kberg:
Related to #1487, first looking at how to make it possible for players to play out a full action only to realize they can't afford it. One issue is that players can exploit such an option by getting their card-drawing bonuses. Well, if you were playing a physical game of Terraforming Mars, would your friends let you draw cards before you could prove you could afford them? So why not here too?

No tests yet, just an RFC.


src/Player.ts
```diff
@@ -2181,6 +2183,10 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
       }
 
       this.setWaitingFor(action, () => {
+        while (this.cardsEarned > 0) {
```
bafolts, line 2186: I am not sure upon earning a bonus for placing a tile that this would always get executed. Upon placing the tile the bonus is that project cards can be drawn. Another option would be to push to the interrupts array for drawing the project card. This would ensure any interrupts already in the queue would happen first. Not entirely sure what the problem being solved is here. Can you provide example? If you place a tile and earned the bonus at that moment you could draw the cards.

kberg: > Not entirely sure what the problem being solved is here. Can you provide example? If you place a tile and earned the bonus at that moment you could draw the cards.

Yeah -- I had a chat on discord yesterday with @nwai90  about #1487, and ways to fix it.

Me:
> The problem comes down to being one of defining what the app should do in best effort. So I see two options: 1. Be perfectly predictive. 2. Be reasonably predictive, but permissive, and also forgiving.

nwa90:
> Yeah. Many players here value correctness highly (reasonably so)

Me:
> I can perfectly understand why they are frustrated if a valid move is disallowed
>
> But it's not particularly maintainable to have code on every card asking if Reds is in power. So that's why I think #2 is better.

nwa90:
> It's possible. This was discussed before

Me:
> Better to allow moves that eventually prove invalid, than to prevent valid moves.

nwa90:
> But the consensus is that this allows exploits. For example, I place a tile that draws cards. That's the main one I know of

Me:
> No but here's a thought: don't draw cards in the middle of the action. Just denote that the player gains cards. At the end of the turn, the player gains their cards.
> Well it's currently both the best and worst suggestion. Let's find a better one.
> I'll toss a prototype of that in a branch so it can at least be reviewed and even tested if people care.


> Imagine you were playing a real game of TM
> On a table
> Around other human beings. :cry:
> And that thing happened: draw two cards, pay 3M€
> And you're like "I'm drawing two cards suckers!!"
> And they're like "where's your 3M€?"
> And you're like "suck it I'm broke!"
> No way you would get to see those cards before you paid.

nwa90:
> Let's try it and see what others think. They might end up finding some edge cases or things we didn't consider. Sometimes it is a bit of a landmine

bafolts: Instead of drawing the cards here I think it will have to be at the top of `takeAction` function once the interrupts queue is empty.

https://github.com/bafolts/terraforming-mars/blob/master/src/Player.ts#L2007

Being around human beings. I read about that in history books.

kberg: > Instead of drawing the cards here I think it will have to be at the top of `takeAction` function once the interrupts queue is empty.

You sure about that? I did it, but -- won't that risk causing the action to delay until the start of the player's next turn?

bafolts: I was worried about the opposite. Risking a user never drawing their cards. I will have to read through source some more.

bafolts: Where you have it after 

https://github.com/bafolts/terraforming-mars/pull/1549/commits/657c50057de0031b13c8d0c0289e8e8843f8f214

Looks right to me. Previously resetting `cardsEarned` to `0` at the top of `takeAction` seems like we would clear `cardsEarned` which may have not been `0`;

kberg: @bafolts OK that's reassuring. Clearly I'm working outside my comfort element -- can you advise me on a good test here?

bafolts: The `player.takeAction` calls to `game.playerIsFinishedTakingActions` when the player can not take another action. Otherwise `player.takeAction` performs an action asynchronously and then calls to `player.takAction`. Where you have the cards earned makes sense as the interrupts have all been cleared and then the function is starting to calculate if there is an action the player can take.

I can advice on _a_ test. As both `Game` and `Player` quickly grew as contributors showed up I can make no guarantee it is _good_. If a test were to set `game.cardsEarned` to `X` and `player.takeAction` were called I would expect the player to then have `X + cardsStartedWith` cards in their hand. If `player.actionsTakenThisRound` is `2` when called the `game.playerIsFinishedTakingActions()` should synchronously be called. In a two player game if this was done on player 1. The first player should have their cards and the action player should now be player 2.

The interrupts that are ran at the top of `playerIsFinishedTakingActions` I believe are running the interrupts for any players that may be able to take actions based on the actions another player has taken. It is probably worth adding some comment here as it took me a while to realize what it was doing.

https://github.com/bafolts/terraforming-mars/blob/master/src/Game.ts#L1104

If one of these interrupts allowed a player to draw a card they would have to be given there card after the interrupts are cleared here as well.

Lynesth:
You might want to draw your cards before doing something else though.
Example:
 - First action, you play **Large Convoy**. Let's say you have **Birds** or **Livestock** so you'll obviously get the 4 animals (who wouldn't, really).
 - You draw 2 cards and one of them is **Nitrophilic Moss**. You are interested in playing this, but you don't have plants!
 - You decide to place your Ocean on a 2 plants board spaac to be able to play that card as your second action.

As you said, I think it's important that the game is as correct as it can be. But moving card draw to the end to be able to rollback actions doesn't seem to be the solution (#1692 was just opened yesterday on that matter).
I agree with @nwai90 that we should be able to tell the user beforehand if he's going to be able to do a specific action when Reds are ruling. We already discussed this a bit and, even though it's going to be a lot of work, I believe it will probably be preferable than allowing users to start actions that they cannot complete.

But that's all just my opinion eh :)

kberg:
I haven't looked at this for some time. Yes, I think what you say is reasonable, @Lynesth if an outlier. In the long-run, giving players a chance to order their action steps, **and also** define action steps as ones from which a player cannot back out. Drawing cards is probably the one and only case where undo would be blocked.

Feels like `fast, cheap, good: pick two.`

kberg:
I think the next thing to do is just abandon this. Any concerns?

kberg:
Yeah, I still had hopes for it. But yeah.

PR #1551, opened 2020-10-09
Improve game creation page

vsrisuknimit:
- Remove radio box and toggle. Create clickable label instead.
- Add icons for expansions.
- Add color for boards, move board column next to expansion.

image


kberg:
That looks great.

vsrisuknimit:
still working on turning WGT by default upon clicking venus expansion. Not working yet. Need help @kberg .

src/components/CreateGameForm.ts
```diff
@@ -162,6 +162,17 @@ export const CreateGameForm = Vue.component("create-game-form", {
             this.turmoil = this.$data.allOfficialExpansions;
             this.promoCardsOption = this.$data.allOfficialExpansions;
         },
+        getBoardColor: function(boardName: String): String {
```
chosta, line 168: 1) Better to be explicit here `getBoardColorClass` - we get a string but we don't know what it is
2) We use lowercase for `string` `boolean` etc. types. Capital letters for classes like `Player`

src/components/CreateGameForm.ts
```diff
@@ -319,6 +330,7 @@ export const CreateGameForm = Vue.component("create-game-form", {
 
                 <div class="container create-game-options">
 
+                    <!--CREATE PLAYERNAME AND COLOR HERE-->
```
chosta, line 336: Comments only for non-trivial code if possible. Also html comments are using bytes.

src/components/CreateGameForm.ts
```diff
@@ -340,151 +352,183 @@ export const CreateGameForm = Vue.component("create-game-form", {
...
+                                </label>
+                            </template>
 
-                            <label class="form-switch" v-if="playersCount > 1">
-                                <input type="checkbox" v-model="randomFirstPlayer">
-                                <i class="form-icon"></i> <span v-i18n>Random first player</span>
+                            <template v-if="playersCount === 1">
+                            <input type="checkbox" v-model="soloTR" id="soloTR-checkbox">
+                            <label for="soloTR-checkbox">
+                                <span v-i18n>63 TR solo mode</span>&nbsp;<a href="https://github.com/bafolts/terraforming-mars/wiki/Variants#tr-solo-mode" class="tooltip" target="_blank">&#9432;</a>
```
chosta, line 470: avoid using `&nbsp;` it's almost always better to have a markup on an element that offsets a couple of pixels

src/styles/common.less
```diff
@@ -103,13 +103,14 @@ h2 {
     filter: drop-shadow(4px 4px 4px black);
     text-shadow: -1px -1px 6px #aaa4;
 }
-
+/*
```
chosta, line 106: We can get rid of comments in css.

src/styles/create_game_form.less
```diff
@@ -100,6 +100,129 @@
...
+    min-width: 125px;
+    border: 3px solid black;
+}
+
+.create-game-page-board-column label:hover {
+    border: 3px dashed gray;
+}
+
+.create-game-page-board-column .create-game-tharsis {
+    background-color: rgba(238, 121, 43, 0.5);
```
chosta, line 216: Please, check `variables.less` and `mixins.less` and try to add colors for the different expansions as constants. We should aim for reusable code in every possible way :)

chosta (changes requested):
Great work! Visually appealing and very intuitive.

src/components/CreateGameForm.ts
vsrisuknimit: It was non-trivial to me at first why it was on top not at the bottom.

src/components/CreateGameForm.ts
vsrisuknimit: It was already there for ? question mark help icon. Should I replace it or add a class on top of it?

chosta:
No need, just a recommendation for the future

src/styles/create_game_form.less
vsrisuknimit: added

vsrisuknimit:
Solo selected:

image

More than one player selected:

image

vsrisuknimit:
One more thing to add then I am done. Working on the color selection.

vsrisuknimit:
@alrusdi Could you have a look at this?

chosta:
Looks good to me. Let's see what @alrusdi has to say about it.

alrusdi:
I like it!

vsrisuknimit:
@alrusdi Thanks!

It is good to go if someone approves of the code change.

alrusdi (approved):

PR #1556, opened 2020-10-10
Purge after Max Days variable

pocc:
2 commits that do the following:

* Makefile for VPSes to restart a server more easily
* Create .env variable for how many days to purge


src/database/SQLite.ts
```diff
@@ -102,13 +102,14 @@ export class SQLite implements IDatabase {
...
         });
-        // Purge unfinished games older than 10 days
-        this.db.run("DELETE FROM games WHERE created_time < strftime('%s',date('now', '-10 day')) and status = 'running'", function(err: { message: any; }) {
-            if (err) {
-            return console.warn(err.message);  
-            }
-        });        
-
+        // Purge unfinished games older than MAX_GAME_DAYS days
+        if (process.env.MAX_GAME_DAYS) {
```
nwai90, line 106: @pocc Is the intention here to not delete unfinished games at all if this variable is not specified, or to keep the default behaviour of deleting unfinished games older than 10 days?

Makefile
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
```
bafolts, line 1: The max days environment variable makes sense. The introduction of this Makefile seems unrelated. I am not sure how many people would use this file. This seems like a personal file for starting and stopping the server.

src/database/SQLite.ts
pocc: That's exactly right. The intention here is to default to not deleting unfinished games at all if the variable is not specified. People who are setting up a server may not be aware that these deletions are happening. I was in this boat until I looked at the code.

So it better to purge periodically as an explicit option.

bafolts: This seems right to me. The SQLite shouldn't take up that much space unless you were playing tens of thousands of games locally. This was added for the Postgres server where we do see tens of thousands of games played as a simple garbage collection mechanism. I can set this environment variable on the heroku instance for use with Postgres. I plan to introduce a change soon where a player or game is queried from the database if it isn't found locally in memory each time. This should avoid the issue we have today where most games in the database are automatically reloaded.

src/database/SQLite.ts
```diff
@@ -102,13 +102,14 @@ export class SQLite implements IDatabase {
...
+            this.db.run("DELETE FROM games WHERE created_time < strftime('%s',date('now', '-" + process.env.MAX_GAME_DAYS + " day')) and status = 'running'", function(err: { message: any; }) {
```
bafolts, line 107: Concatenating this string poses the risk of a malicious attack. If we can use `?` here and pass in this value that would be ideal. Otherwise I would use `+ Number(process.env.MAX_GAME_ENDS) + ` to ensure that we only pass in a number and not a malicious query string.

pocc: This has been fixed. The syntax should now be correct.

Makefile
pocc: I've removed the Makefile. Talking to people on the discord server, I seem to be the among the one or two who uses a VPS (most people use heroku). Nodemon makes more sense here because it won't restart the server but will automatically update it if it detects changes in files.

bafolts (approved):

Issue #1557, opened 2020-10-10
AEROSPACE MISSION got incomprehensibility info


Ronnoc:
Build colonies on ${c1.name} (${c1.description}) and ${c2.name} (${c2.description})
https://terraforming-mars.herokuapp.com/player?id=994741a3f361

nwai90:
Fixed and merged

PR #1558, opened 2020-10-10
Fix incomprehensible option text on Aerospace Mission

nwai90:
Ref: https://github.com/bafolts/terraforming-mars/issues/1557


pierrehilbert (approved):
Perfect for me.

kberg:
Huh - what was broken?

bafolts:
The ${} syntax only works within a template string with the back tick and not double quote characters. This should have used ` and not " originally.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

PR #1559, opened 2020-10-10
defensive coding measures on new cards

bafolts:
The heroku application crashed this morning after the introduction of the new community cards. Unfortunately the cheap hosting plan used only allows me to go back 1500 log entries which isn't including the runtime exception which caused the crash. I have added defensive coding measures to protect the new cards from taking down the server. One of the cards when played without turmoil would have crashed the server then `turmoil` was not defined.

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1326,10 +1326,8 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
             }
         }
 
-        const preludesWithPlayCardEffects = [CardName.ECOLOGY_EXPERTS, CardName.ECCENTRIC_SPONSOR, CardName.VALUABLE_GASES];
-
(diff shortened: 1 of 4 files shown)


kberg (approved):
I am curious what happened here! But OK. I'll read initial the PR comment more closely.

kberg:
I'll leave merging to you.

bafolts:
The `!` operator assumes the property will be defined. I don't have evidence for what caused the crash but if `game.turmoil` were undefined a runtime exception could be thrown. As javascript is single threaded and the heroku instance only runs one instance with no daemon to restart the process this could take down the process. For the fix with arrays typescript only knows of the type of the items in the array and can't require protecting against undefined. Need to enforce the linting rules on every file which can enforce both of these checks.

PR #1560, opened 2020-10-10
Modify the way interrupts work

Lynesth:
## Modified interrupts
They are now processed when they need to be displayed instead of when the card if played. It allows them to be aware of the results from previous interrupts in the chain and fixes stuff like #1371.

## ~~Energy Tapping and Power Supply Consortium~~
~~When fixing tests for the new interrupts, I noticed those cards were only playable if a player had some energy production while it should be possible to play them without (and decrease the same production you are increasing, yours). This is now fixed.~~
_Edit: This has been moved to another PR._


src/Game.ts
```diff
@@ -632,56 +579,39 @@ export class Game implements ILoadable<SerializedGame, Game> {
...
-            }),
-            new SelectOption("Skip removing plants", "Confirm", () => {
-              return undefined;
-            })
-          )});
-        } else {
-          candidates[0].setResource(resource, -qtyToRemove, this, player);
-          return;
+    public runNextInterrupt(cb: () => void, findByPlayer?: Player): boolean {
+      if (this.interrupts.length < 0) {
```
bafolts, line 590: This most likely was supposed to be `this.interrupts.length === 0`. The length of the array will never be negative. When there are no interrupts there is nothing to run.

src/interrupts/SelectResourceCard.ts
```diff
@@ -1,34 +1,59 @@
 import { Game } from "../Game";
-import { PlayerInput } from "../PlayerInput";
+import { PlayerInput } from '../PlayerInput';
 import { Player } from "../Player";
 import { PlayerInterrupt } from "./PlayerInterrupt";
 import { SelectCard } from "../inputs/SelectCard";
 import { ResourceType } from "../ResourceType";
 import { ICard } from "../cards/ICard";
+//import { Tags } from "../cards/Tags";
```
bafolts, line 8: This line should be removed as it isn't being used.

src/interrupts/SelectResourceDecrease.ts
```diff
@@ -1,29 +1,77 @@
-
 import { Game } from "../Game";
-import { PlayerInput } from "../PlayerInput";
+import { PlayerInput } from '../PlayerInput';
```
bafolts, line 2: An ask like this can be frustrating and so far everything in this pull request is looking great. The `eslint` rules haven't been pushed to every directory yet but they soon will be. That rule is going to dictate the usage of double quotes versus single quotes. Until the linting rule is in place using single quote is fair game but using double quotes would avoid needing to update this in the future. I think some IDEs are enforcing the `.eslintrc.js` which enforces double quotes.

src/interrupts/SelectResourceProductionDecrease.ts
```diff
@@ -1,29 +1,45 @@
...
-              return undefined;
-            }
-        );
-    };
+    ){}
+
+    public generatePlayerInput() {
+        let candidates: Array<Player> = [];
+        if (this.resource === Resources.MEGACREDITS) {
+            candidates = this.game.getPlayers().filter((p) => p.getProduction(this.resource) >= this.count - 5);
```
bafolts, line 21: Is this `-5` attempting to determine which players have production that isn't already at the minimum? The `Player::addProduction` method will cap production at `-5` if an attempt is made to push it lower than the minimum. If for some reason a player wanted to decrease production of a player which were already at `-5` we should allow it as there is a push to allow players to make any legal move.

#1474

bafolts (commented):
This all looks great and it should calculate the values when the interrupt is going to run versus storing it initially. I only spotted minor issues where I left comments. I am going to do another sanity read through of this tomorrow as it is a larger change.

Also it may be worthwhile to modify the API to be:

```
export interface PlayerInterrupt {
   player: Player,
   playerInput: () => PlayerInput
}
```

This would avoid needing to have `generatePlayerInput` be optional. It would also enforce that _every_ interrupt is always loading the input when it is needed. There are interrupts we are creating in place that also need to be created when needed.

https://github.com/bafolts/terraforming-mars/blob/master/src/cards/promo/PharmacyUnion.ts#L87

What is here is better than what was here before though. Will give another sanity read through tomorrow.

src/interrupts/SelectResourceProductionDecrease.ts
Lynesth: That line (and the one below) checks that you can at least reduce the production by `this.count`, otherwise the player isn't a valid target since production decrease is mandatory (as opposed to resource removal).

Example:
2 players game, you have +3 M€ production and your opponent has -4.
If you wanted to play HACKERS, the only valid target for the M€ production decrease would be yourself (`-4 is not >= to (2 - 5)`).

src/Game.ts
Lynesth: Indeed an mistake here. This is fixed.

kberg:
This looks really good. I'm going to have to get my head around some of it a bit more when it makes it in, but I like seeing deleted code!

src/interrupts/SelectResourceProductionDecrease.ts
Lynesth: Actually I wonder if `Player::addProduction` shouldn't instead throw an exception if you try to reduce below the minimum since that is not allowed at all in the game and production decrease is always mandatory (except for the SABOTAGE Global Event).

Lynesth:
> Also it may be worthwhile to modify the API to be

Would you rather have it now or as part of another PR once this one is merged?

bafolts:
Another PR is fine. The pre-set `playerInput` in the use cases where the interrupt is provided with `{}` wont be made any worse with this. This will improve the other usages.

src/interrupts/SelectOcean.ts
```diff
@@ -1,30 +1,33 @@
...
-        public title?: string,
+        public title: string = 'Select space for ocean tile',
         public isWorldGov: boolean = false
-    ){
-        if (title === undefined) {
-            title = 'Select space for ocean tile';
+    ){}
+
+    public generatePlayerInput() {
+        if (this.game.board.getOceansOnBoard() + 1 >= constants.MAX_OCEAN_TILES) {
```
bafolts, line 20: When there are 8 oceans I believe this check will avoid allowing the placement for the final ocean. I think this should be.

```
if (this.game.board.getOceansOnBoard() >= constants.MAX_OCEAN_TILES)
```

src/Game.ts
```diff
@@ -549,20 +548,10 @@ export class Game implements ILoadable<SerializedGame, Game> {
...
-            return;
-      }
       this.addInterrupt(new SelectHowToPayInterrupt(player, amount, title, canUseSteel, canUseTitanium));
     }
 
     public addOceanInterrupt(player: Player, title?: string, isWorldGov: boolean = false): void {
-      if (this.board.getOceansOnBoard() + this.pendingOceans  >= constants.MAX_OCEAN_TILES) {
-        return;
-      }
-      this.pendingOceans++;
```
bafolts, line 565: The `pendingOceans` property is no longer used. We should remove it from `Game` and `SerializedGame` while making this change.

src/Game.ts
```diff
@@ -632,56 +579,39 @@ export class Game implements ILoadable<SerializedGame, Game> {
...
-        } else {
-          candidates[0].setResource(resource, -qtyToRemove, this, player);
-          return;
+    public runNextInterrupt(cb: () => void, findByPlayer?: Player): boolean {
+      if (this.interrupts.length === 0) {
+        return false;
+      }
+      let interrupt;
+      if (findByPlayer !== undefined) {
+        const interruptIndex: number = this.interrupts.findIndex(interrupt => interrupt.player === findByPlayer);
```
bafolts, line 595: This was copied but we should always compare on `player.id`. While here lets update with `interrupt.player.id === findByPlayer.id`.

src/interrupts/SelectResourceProductionDecrease.ts
bafolts: This is fine to keep. I saw you copied it during this pass through.

src/interrupts/RemoveOcean.ts
```diff
@@ -6,24 +6,23 @@ import { ISpace } from '../ISpace';
...
-        if (title === undefined) {
-            title = 'Select an Ocean tile to remove from board';
-        }
+    ){}
+
+    public generatePlayerInput() {
         this.playerInput = new SelectSpace(
-            title,
-            game.board.getOceansTiles(),
+            this.title,
```
bafolts, line 19: Technically there could be no tiles to remove. If `this.game.board.getOceansOnBoard() === 0` this should set `playerInput` to `undefined`.

bafolts (changes requested):
OK! I gave this another read through and the only thing that stands out to me is the `SelectOcean` interrupt.

I think the check should be:

```typescript
if (this.game.board.getOceansOnBoard() >= constants.MAX_OCEAN_TILES) {
    this.playerInput = undefined;
    return;
}
```

Consider the scenario where `constants.MAX_OCEAN_TILES` is `1` and `this.game.board.getOceansOnBoard()` is `0`. With the current logic a player would never be able to select the space for the final ocean.

This change moves code out of `src/Game.ts` AND removes duplication. After this slight tweak for oceans this should be ready to ship.

My other comments would all be nice to have versus needed.

src/interrupts/SelectOcean.ts
Lynesth: Yep that's a complete oversight, thanks.

src/Game.ts
Lynesth: Fixed it. I'll try to take some time to make a PR that only does that across the board since I spotted some other places this is the case.

bafolts (approved):
:shipit: 

This is a nice change for players in that resources aren't calculated until the interrupt is used.

src/Game.ts
```diff
@@ -542,55 +530,13 @@ export class Game implements ILoadable<SerializedGame, Game> {
       }
     }
 
-    public addSelectResourceCardInterrupt(player: Player, count: number = 1, resourceType: ResourceType, resourceCards: Array<ICard>, title?: string): void {
```
kberg, line 545: This function is used in Ares. Please restore it?

kberg: (It was actually written for Ares.)

Lynesth: Since it wasn't in use anywhere, I simply removed it.
Should I put it back in or do you add it to the Ares branch ?

src/interrupts/SelectHowToPayInterrupt.ts
```diff
@@ -1,26 +1,36 @@
...
     constructor(
         public player: Player,
         public amount: number,
         public title: string = "Select how to pay for " + amount + " MCs",
         public canUseSteel: boolean,
         public canUseTitanium: boolean,
-    ){
+    ){}
+
+    public generatePlayerInput() {
```
kberg, line 17: So, here's my concern about this structure: it breaks the expectation of what a PlayerInterrupt is. Is it a PlayerInterrupt? But instead you're not going to interrupt sometimes? It's breaking the isa-nature of this instance.

I agree that you're simplifying something (I wish your PR description said what it was more than "New interrupts" -- for a while there I thought you were adding new interrupts) but now that I'm looking at this critically, I see this flaw.

OK, so maybe there's another way to do this that preserves the Is-a of the `PlayerInterrupt`. Is the goal of moving logic in here to defer execution? If so I might suggest creating a superclass to `PlayerInterrupt` let's call it `DeferredAction` and it's nothing more than a `(game: Game) => void`.

Now, we change game.interrupts from Array<PlayerInterrupt> to Array<DeferredAction>.

And then this looks like
```
class ChargePlayer implements DeferredAction {
    constructor(
        public player: Player,
        public amount: number,
        public title: string,
        public canUseSteel: boolean
        public canUseTitanium: boolean) {};

  cb(game: Game) {
     // Someone should make a method called Player.mustPayCash(amount): boolean
     if ((!this.player.canUseHeatAsMegaCredits || this.player.heat === 0) &&
            (!this.canUseSteel || this.player.steel === 0) &&
            (!this.canUseTitanium || this.player.titanium === 0)) {
                this.player.megaCredits -= this.amount;
        }
    } else {
      game.addInterrupt(new SelectHowToPayInterrupt(player, amount, title, ....)
    }
}
```

I know you have put a lot into this already. Let me know your thoughts.

Lynesth: I guess I understand what you mean. I honestly didn't pay much attention to the naming of objects.
I'll try to explain my chain of thoughts while excluding the only exception that the interrupt for trading with Pluto is.

### Current implementation
The only thing a `PlayerInterrupt` does is to craft a `PlayerInput`.
When created, they get added to an array called `game.interrupts`.
The function that iterates over `game.interrupts` simply calls `player.setWaitingFor`, passing the crafted input as argument.

So, for example,  when you play **Giant Ice Asteroid**, its `play` function does the following:
```ts
game.increaseTemperature(player, 2);
game.addOceanInterrupt(player, "Select space for first ocean");
game.addOceanInterrupt(player, "Select space for second ocean");
game.addResourceDecreaseInterrupt(player, Resources.PLANTS, 6);
```

This will, for each `add*Interrupt`, generate a `PlayerInput` based on the state of the game at that time, and push it in the `game.interrupts`.
It means that, as soon as you play the card, it selects a list of targets for the plants removal (and the max amount you can remove from each target) instead of after the 2 Oceans have been placed (which could change the amount of plants amongst players - placement bonus + **Arctic Algae**).

-----
This is the reason we cannot make those calculations when adding them to the stack and why they must be made at the time action is supposed to be taken.

-----

### Back to the PR
When I made that PR, I just thought of `game.interrupts` as a stack of actions that may prompt a player to do something.
Now, I understand the semantic issue with keeping the name "interrupt" for an object that might not interrupt anything.

I would happily rename it for something like `deferredAction`. It would also probably help with some other timing issues like **Helion** getting 3MC from **Standard Project** before chosing how to pay for it.

Here's what I think it could look like:
#### SelectOcean (currently a PlayerInterrupt)
```ts
export class SelectOceanDA implements DeferredAction {
    constructor(
        public player: Player,
        public game: Game,
        public title: string = 'Select space for ocean tile'
    ){}

    public execute(): void {
        if (this.game.board.getOceansOnBoard() >= constants.MAX_OCEAN_TILES) {
            return;
        }
        this.player.setWaitingFor(new SelectSpace(
            this.title,
            this.game.board.getAvailableSpacesForOcean(this.player),
            (space: ISpace) => {
                this.game.addOceanTile(this.player, space.id, SpaceType.OCEAN);
                return undefined;
            }
        ));
    }
}
```
#### Game::runNextInterrupt
```ts
    public runNextDeferredAction(cb: () => void, findByPlayer?: Player): boolean {
        // This function returns true if an action was executed, false otherwise
        if (this.deferredActions.length === 0) {
            return false;
        }
        let da;
        if (findByPlayer !== undefined) {
            const daIndex: number = this.deferredActions.findIndex(da => da.player.id === findByPlayer.id);
            if (daIndex < 0) {
                return false;
            }
            da = this.deferredActions.splice(daIndex, 1)[0];
        } else {
            da = this.deferredActions.shift();
        }
        if (da === undefined) {
            return false;
        }
        da.execute();
        cb();
        return true;
    }
```
#### Player::addGreenery (example for Helion + Standard Technology)
```ts
    private addGreenery(game: Game): PlayerInput {
        return new SelectOption(
            "Greenery (" + constants.GREENERY_COST + " MC)",
            "Confirm",
            () => {
                game.defer(new SelectHowToPayDA(this, constants.GREENERY_COST, false, false, "Select how to pay for Greenery project"));
                game.defer(new SelectGreenery(this, game));
                game.defer({ game, execute: () => { this.onStandardProject(StandardProjectType.GREENERY) } });
                game.log("${0} used ${1} standard project", b => b.player(this).standardProject("Greenery"));
                return undefined;
            }
        );
    }
```

-----
This was a pretty long comment, and there's probably mistakes in the code I wrote on the fly.
Also I'm very far from being a Typescript guru nor a JS one and my view on it might also not fit best practice or this project's ways of doing things.

Anyway, let me know what you think.

bafolts:
I like this class of `DeferredAction`. I would change `game.interrupts` to `game.deferredActions` with `Array<DeferredAction>`. I don't think that `DeferredAction` needs to get into the business of modifying state. Similar to cards I think the `execute` should return `PlayerInput | undefined`. If it returns `PlayerInput` the player needs to provide more information and we can populate `waitingFor` if we want. If there is no action needed by the player we can move on to the next deferred action. This should make these deferred actions easier to unit test as we don't need to create some mock `setWaitingFor` to see what value ends up being populated. I also think this could allow us to serialize these and not lose them on server restarts.

We [lose the interrupts](https://github.com/bafolts/terraforming-mars/blob/master/src/Game.ts#L1761) on server restart. If there is any chance of eventually loading these from JSON we will need to avoid storing a reference to `Player` in the constructor. If we use `Player.id` or `Game.id` this will allow us to repopulate these values later when this interrupt is finally executed. If these constructors only worked with player id and game id we should be able to reconstruct these when the server restarts. I wouldn't consider this deserialization with this change but I would only use player id and game id in the constructors for deferred actions so in the future we can deserialize in another change.

I have never been in love with `runNextDeferredAction` as it both modifies state and returns a value. I think there are at least two functions hiding here. One of them being if there is a deferred action and the other to run the deferred action. With my suggest I think this then becomes functions like.

```
    public getNextDeferredAction(): DeferredAction | undefined {
        return this.deferredActions[0];
    public getNextDeferredActionForPlayer(playerId: string): DeferredAction | undefined {
        return this.deferredActions.find(player => player.id === playerId);
    }
    private removeDeferredAction(action: DeferredAction): void {
        this.deferredActions.splice(this.deferredActions.indexOf(action, 1);
    }
    public runDeferredAction(action: DeferredAction, cb: () => void): void {
        if (input !== undefined) {
            this.setWaitingFor(input, () => {
                this.removeDeferredAction(action);
                cb();
            });
        } else {
            this.removeDeferredAction(action);
            cb();
        }
    }
```

Then when calling this from `Game` where we only need to flush the queue we would call:

```
if (this.getNextDeferredAction() !== undefined) {
    this.runDeferredAction(this.getNextDeferredAction(), () => this.whatToDoNext());
}
```

Within player when they are player specific we call:

```
if (this.getNextDeferredActionForPlayer(this.id) !== undefined) {
    this.runDeferredAction(this.getNextDeferredActionForPlayer(this.id), () => this.whatToDoNext());
}
```

Lynesth:
Ok so that's basically renaming everything interrupts to deferredActions and exploding the runNextInterrupt into 4 different smaller functions.

That sounds pretty good to me.

Lynesth:
The PR also still has a few functions like `addSelectHowToPayInterrupt`, `addOceanInterrupt`, etc.

Do you want to transform them as well (`game.addSelectHowToPayDeferredAction()`) or remove them since they don't have logic and go with something like `game.defer(new SelectHowToPayDeferredAction())` ?

bafolts:
Removing them for a `game.defer` makes sense now.

src/interrupts/SelectHowToPayInterrupt.ts
kberg: > This was a pretty long comment, and there's probably mistakes in the code I wrote on the fly.
> Also I'm very far from being a Typescript guru nor a JS one and my view on it might also not fit best practice or this project's ways of doing things.

Yeah, but it's good. I appreciate your long write-up here, but now I fully understand the problem you're solving.

I have no problem therefore about the state of the PR as is, though I'll now do yet another look-through with a new lens, and an understanding of a different architecture at the other end -- which doesn't have to be in this PR as far as I'm concerned, but I defer completely to @bafolts here.

kberg:
```
public getNextDeferredAction(): DeferredAction | undefined {
public getNextDeferredActionForPlayer(playerId: string): DeferredAction | undefined {
private removeDeferredAction(action: DeferredAction): void {
public runDeferredAction(action: DeferredAction, cb: () => void): void {
```

This sounds like something that can become a class on its own.

src/SerializedGame.ts
```diff
@@ -34,7 +34,6 @@ export interface SerializedGame {
     
     oxygenLevel: number;
     temperature: number;
-    pendingOceans: number;
```
kberg, line 37: Intentional to remove a field from SerializedGame?

src/interrupts/RemoveOcean.ts
```diff
@@ -6,23 +6,25 @@ import { ISpace } from "../ISpace";
...
-    public playerInput: PlayerInput;
+    public playerInput?: PlayerInput;
     constructor(
         public player: Player,
         public game: Game,
-        public title?: string,
-    ){
-        if (title === undefined) {
-            title = "Select an Ocean tile to remove from board";
+        public title: string = 'Select an Ocean tile to remove from board'
```
kberg, line 13: Double-quotes, please!

src/interrupts/SelectCity.ts
```diff
@@ -6,19 +6,21 @@ import { ISpace } from '../ISpace';
...
-    public playerInput: PlayerInput;
+    public playerInput?: PlayerInput;
     constructor(
         public player: Player,
         public game: Game,
         public title: string = 'Select space for city tile'
-    ){
+    ){}
+
+    public generatePlayerInput() {
```
kberg, line 16: Perhaps I missed this and I'll go back to your description from earlier, but I would have assumed that instead of a method which has to be called to generate internal state, for example, consider:

_PlayerInterrupt.ts_
```
export class PlayerInterrupt {
  protected player: Player;
  private playerInput?: PlayerInput;
  private generated: boolean = false;

  public getPlayerInput() {
    if (!generated) {
      this.playerInput = this.generateInput();
      this.generated = true;
    }
    return this.playerInput();
  }

  // Subclasses implement this.
  public generateInput(): PlayerInput | undefined { }
}
```

And then in Game.ts, replace

```
interrupt.generatePlayerInput?.();
  if (interrupt.playerInput) {
```

with 
```
const playerInput = interrupt.getPlayerInput();
if (playerInput) {
}
```

and SelectColonyInterrupt, as an example, no longer has additional demands on it.

```
export class SelectColonyInterrupt *extends* PlayerInterrupt {
    constructor(
        public player: Player,
        public game: Game,
        public openColonies: Array<IColony>,
        public title: string = "Select where to build a colony"
    ) {
      super(player);
    }

    *protected* generateInput() {
        let coloniesModel: Array<ColonyModel> = this.game.getColoniesModel(this.openColonies);
-        this.playerInput = new SelectColony(this.title, "Build", coloniesModel, (colonyName: ColonyName) => {
+       return new SelectColony(this.title, "Build", coloniesModel, (colonyName: ColonyName) => {
            this.openColonies.forEach(colony => {
                if (colony.name === colonyName) {
                    colony.onColonyPlaced(this.player, this.game);
                }
                return undefined;
            });
            return undefined;
        });            
    }
}    
```

kberg: By the way, this includes an assumption that because you separated generatePlayerInput was because you wanted that to be re-entrant, which is why I created the `generated` guard. If I'm wrong, feel free to ignore that part.

I just get nervous about putting the demand on a sublcass implementer to have to change internal state on every instance, eg DRY.

kberg (commented):
More tomorrow.

src/SerializedGame.ts
bafolts: This property is no longer used after this change. It was being used prior when there were interrupts holding a spot to place an ocean.

src/interrupts/SelectHowToPayInterrupt.ts
bafolts: I approved this as I think it is at least better from what we offer today as it calculates values when the interrupt is executed versus when the executed is queued. I don't think it is far off from a `DeferredAction` approach. Could do that in a separate pull request. I am a big fan of the `DeferredAction` or some class which returns the `PlayerInput` through a function like `ICard` since we should be able to serialize those and restore them from the database. Once that change is introduced we can most likely re-populate the `interrupts` queue from the database.

src/SerializedGame.ts
kberg: Gotcha.

kberg:
I attempted to resolve the conflicts, but only resolved two of the four -- not very helpful in retrospect -- mea culpa. Sorry about that.

kberg (approved):

Lynesth:
I agree with you both that this isn't an optimal solution yet and I'm happy to work on another PR improving this with your feedback and ideas :)

kberg:
Good work. Done!

bafolts:
This is nice. Once we have some `DeferredAction` or similar class I can work on serializing these to and from the database instead of losing them. If you go far enough back in git history you will see that every action used to involve complicated nested `OrOption`, `AndOption`, and `Select*` and there was no idea of an interrupt. The interrupts got us away from that. This change now puts this in an even better place!

PR #1561, opened 2020-10-10
eslint fixes.

kberg:


src/components/DebugUI.ts
```diff
@@ -49,7 +49,7 @@ export const DebugUI = Vue.component("debug-ui", {
             return allItems;
         },
         filtered: function(cardName: string):boolean {
-            return this.$data.filterText.length == 0 || cardName.toUpperCase().indexOf(this.$data.filterText.toUpperCase()) > -1;
+            return this.$data.filterText.length === 0 || cardName.toUpperCase().indexOf(this.$data.filterText.toUpperCase()) > -1;
```
bafolts, line 52: To enable eslint across the board I am debating enforcing it on every file with the `--fix` option enabled. With `--fix` enabled it should help with the merge conflicts once introduced where the code can get merged without passing the same lint file. Some of the errors can't be fixed with `--fix` and will have to fix those in a separate pull request. I am hesitant to enable `--fix` as it modifies code automatically without the user knowing.

The other option is to put up pull request with all the fixes and deal with the pain for whoever ends up merging the first few times. As I type this out I am thinking I lean towards the option of all at once but I can prep people on when it is coming.

bafolts (approved):

PR #1566, opened 2020-10-11
Remove isWorldGov parameter, use a real definition for Solar Phase in the Phase enum.

### LINKED ISSUE #1564: Mining Guild shouldn't activate credit during WGT.
Bug report:
> Bug: mining guild increase steel prod 1 step   when other players  place a tile and  gain steel/titanium placement bonus
> if anyone want to fix this , don't forget : MiningGuild places an ocean act as the WG shouldn't increase his steel prod too

kberg:
It's possible this could break running games.

I wrote this after trying to fix #1564 by passing `isWorldGov` to yet several more places. Defining a concrete solar phase seemed like a better approach.

Let me know what additional tests you might like, or whether I should stop going down this path.


kberg:
As a comparison, see https://github.com/kberg/terraforming-mars/compare/main...kberg:bug1564?expand=1, which pushes `isWorldGov` around and https://github.com/kberg/terraforming-mars/compare/worldgov...kberg:1564-b?expand=1 which checks the current phase.

Ignoring the eslint changes, of course.

package.json
```diff
@@ -34,6 +34,7 @@
     "extends": "@istanbuljs/nyc-config-typescript"
   },
   "scripts": {
+    "clean": "rm -r ./dist/ ./styles.css",
```
bafolts, line 37: :+1:

kberg: Oops -- wrong branch. I wondered where that went. :D I'll create a new one for that.

bafolts (approved):
This is a great change. I haven't played with all of these expansions. I now don't have to know what world government is while looking at most interrupts. Someone calling to add the interrupts in the future also doesn't have to consider whether it is for world government.

bafolts:
The `game.PHASE` is stored in database. When the game server is restarted we will have JSON representation of `waitingFor` but we couldn't have serialized `waitingForCb` as it is a function. Most games on the public heroku are probably breaking during each code push to `master`. 

I assume we end up throwing:

https://github.com/bafolts/terraforming-mars/blob/master/src/Player.ts#L2219

I wouldn't worry about this breaking games in progress on the heroku server as I have no idea how many even use the world government expansion. As I am looking at this now will also need some strategy for dealing with these `waitingFor`. Even the undo operation is most likely leaving games in states we aren't expecting.

kberg:
@bafolts if this is still OK, lemme know -- or merge the PR? This was RFC/DNS before.

bafolts:
Will need some strategy for serializing the "waitingFor" for each player. Storing this phase in the database should help with server restarts as we now will know which phase the game was in. Hopefully most people are running server locally for games or servers which don't update as often.

PR #1567, opened 2020-10-11
Fix: Mining Guild placement bonus for WGT

nwai90:
This fixes https://github.com/bafolts/terraforming-mars/issues/1564 caused by incorrect timing of setting `space.player` to `undefined` and adds a unit test for WGT placements not triggering Mining Guild's effect.

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -1547,6 +1547,7 @@ export class Game implements ILoadable<SerializedGame, Game> {
         space.player = undefined;
       }
 
+      // Mining Guild tile placement effects should resolve before removing space.player for Oceans
       this.players.forEach((p) => {
         if (p.corporationCard !== undefined &&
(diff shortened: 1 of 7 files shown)


kberg:
Thanks, whoops.

src/Game.ts
```diff
@@ -1594,6 +1588,10 @@ export class Game implements ILoadable<SerializedGame, Game> {
           }
         });
       });
+
+      if (tile.tileType === TileType.OCEAN) {
```
kberg, line 1592: The next person like me is going to look at this and wonder why it's here versus where space.player is set -- that it has to be set between then and now is not obvious.

In fact, I still don't understand why.

All of this is to say that a comment explaining this specific timing issue would go a long way. Won't block submission for it tho.

kberg (approved):

PR #1568, opened 2020-10-11
Fix Stratospheric Birds + Dirigibles edge case interaction

Lynesth:
It doesn't really change the core of what's happening since it was working properly.
But this way it lets you spend only the correct amount of floaters instead of letting you spend all of them and fixing it by paying an additional 3M€.

It adds a bit of code here and there (couldn't think of a less intrusive way of doing it) so it might not be that good. Let me know.

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1103,7 +1103,7 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
             cards.push(
               {
                 resources: targetCard.resourceCount,
-                resourceType: undefined,
+                resourceType: undefined, // Card on SRR cannot gather its own resources (if any)
                 name: targetCard.card.name,
                 calculatedCost: this.getCardCost(game, targetCard.card),
                 cardType: card.cardType,
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1854,7 +1854,7 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
         if (dirigibles !== undefined 
            && dirigibles.resourceCount
            && card.tags.indexOf(Tags.VENUS) !== -1) {
-            maxPay += (dirigibles.resourceCount - +(card.name === CardName.STRATOSPHERIC_BIRDS)) * 3;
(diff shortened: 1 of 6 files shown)


nwai90:
@Lynesth Will this prevent a player from spending all available floaters on Dirigibles? We should only prevent this behaviour if Dirigibles is the only card with floaters that the player has.

Lynesth:
@nwai90 Yep it's only supposed to work that way.
The idea is that it checks that you don't spend all your available floaters when playing Stratospheric Birds.

Lynesth:
@nwai90 Oh I see what I did wrong. Will fix.

Lynesth:
Ok so it's finally fixed (or at least it should).
The previous commit was done too quickly and I didn't check that everything was properly working, sorry about that.

To get access to all the info needed in the payment widget component, I added a `resourceType` property to the `CardModel` interface.

Thanks @nwai90 for your time on Discord with this 👍 .

src/Player.ts
```diff
@@ -1103,6 +1103,7 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
             cards.push(
               {
                 resources: targetCard.resourceCount,
+                resourceType: undefined,
```
Lynesth, line 1106: Just a precision here (since it was asked).

This card is on the SRR and to add its real resourceType would require more code changes as RobotCard (which targetCard is) doesn't have that property.

Also, since it's there, the resources it is actually "gathering" are the SRR's ones, not its own (which it cannot gather at that time) so it would be even more misleading. Hence the choice to set it to undefined.

If you'd rather have the real one here, or that I create a new SRR resourceType for example, let me know.

nwai90: @Lynesth This should be fine as is 👍 Creating a new SRR resourceType might be more misleading as well.

It may be useful to leave a small comment here as this is the only property that is intentionally being set to a different value than what would normally be expected.

nwai90 (approved):

src/Player.ts
Lynesth: Added a comment.

PR #1569, opened 2020-10-11
Fix some cards in case players have the same name

Lynesth:


nwai90:
Thanks for this one! We could also use `player.color` here as it is guaranteed to be unique upon game creation.

I was also thinking about preventing duplicate player names from occurring in the first place but that should be a separate issue.

src/cards/HiredRaiders.ts
```diff
@@ -28,7 +28,7 @@ export class HiredRaiders implements IProjectCard {
             );
         }
 
-        const availablePlayerTargets = game.getPlayers().filter((p) => p.name !== player.name);
+        const availablePlayerTargets = game.getPlayers().filter((p) => p !== player);
```
bafolts, line 31: This needs to use `p.id` to match. The `===` operator will only compare that two players are the same instance of a class which isn't necessarily always true. We serialize and de-serialize JSON objects into players and have no guarantee these will be the same instance.

bafolts (changes requested):
This is a good catch but should compare on the player `id`.

```
> { id: 1} === { id : 1 }
false
```

The equality operator only tells us if the objects are the same instance in this situation which may be true but for our needs comparing by `id` will be safer as we serialize to JSON and back often.

bafolts (approved):

PR #1570, opened 2020-10-11
improve player color selection box

vsrisuknimit:
- Improve player color selection box
- The color of the whole player box changes in response to what color is selected making it easier to check who is going to play what color.

image

Please don't merge until after #1551 is merged. I would like to hide the Beginner Corp and TR boost until an option is selected.


vsrisuknimit:
What it looks like currently. No gray bar, more spacing.

image

src/styles/create_game_form.less
```diff
@@ -103,3 +127,72 @@
...
+  
+.create-game-colorbox-yellow {
+    background-color: @player_yellow;
+}
+  
+.create-game-colorbox-green {
+    background-color: @player_green;
+}
+  
+.create-game-colorbox-blue {
```
chosta, line 187: these classes are repeating with the ones in player_home.less

I'd suggest you use .player_bg_color_red etc. instead
Actually you can move // normal backgrounds,  // backgroungs with opacity and // normal backgrounds
to `common.less` so it's easier for people to spot and reuse them

src/styles/create_game_form.less
```diff
@@ -152,7 +151,7 @@
     text-align: center;
     vertical-align: middle;
     padding: 4px 6px;
-    margin: 2px;
+    margin: 0px 6px;
```
chosta, line 154: It's better to use increments of 4px 8px 16px for consistency. By having spacing that is a factor of 4px the app will have a much better overall feel. One of these days, I'll add variables with these amounts.

vsrisuknimit:
Please ignore this PR. Git hates me again. I will try to make a new clean one.

PR #1572, opened 2020-10-11
Ocean spot blue highlight

ssimeonoff:
Before and After:
map

As minor side fixes:
- fixed Poseidon
- removed box-shadow from the corporation effect box (some content were overflowing and it looked ugly, e.g. corporations with VP on them)
corp


vsrisuknimit:
Well defined border of the effect box looks better in my opinion. It is better to be able to distinguish clearly what is the effect and what is the starting resources.

Everything else the right one looks better. (50:50 on red border. Again I like clear border.)

vsrisuknimit:
Should we consider adding similar highlight for volcanic sites as well? Perhaps just red border not the filling like oceans.

chosta:
Please, merge #1455 before this :)

kberg:
> Should we consider adding similar highlight for volcanic sites as well? Perhaps just red border not the filling like oceans.

Red and Yellow borders are a visual cue for Ares -- is it possible not to double-use those?

(I know this is oddly in contrast with my comment about color-blindness in another PR, it's a tricky thing, for sure.)

ssimeonoff:
Ares - one more reason empty spots NOT to have borders, but only background is some colour.

nwai90 (approved):

Issue #1286, opened 2020-09-13
Edge case for Pharmacy Union


nwai90:
Ref: https://boardgamegeek.com/thread/2451064/article/35805086#35805086

When a card with both Microbe and Science tags is played by Pharmacy Union's owner, the player should be allowed to choose the order of tag resolution

vsrisuknimit:
Note that you still lose 4 mc no matter what.

Issue #1179, opened 2020-08-17
GMO Contract + Pharmacy Union ordering issue


duckmammal:
When a PU plays a card with a microbe on it, They often times try to make themselves poor first so they don't have to pay the 4 MC fee. 

With a GMO Contract in play, they should be able to get 2 MC after losing 4 MC. 

It's always better to lose all the money first before gaining any money back if PU is playing a science tag.

PR #1573, opened 2020-10-11
Added onCardPlayedPriority to cards

### LINKED ISSUE #848: Bug: Mars University interrupt triggers before Olympus Conference
**Context:** If player has both cards and Olympus Conference's effect would result in drawing a card from playing a Science tag, its effect should be resolved first before Mars University's cycle effect

Lynesth:
The game will order the cards by ascending priority (smaller is "faster") before calling their `onCardPlayed` method.

This was originally made to fix #848 but I decided to check [all currently available cards with an effect](https://ssimeonoff.github.io/cards-list#CORP23#C24#C41#CORP24#031#CORP22#CORP19#128#074#X14#CORP37#CORP38#CORP40#073#T06#CORP01#CORP05#CORP12#131#185#259#109).
 I only found 4 cards that interact with each other and gave 3 of them a high priority.

Now, if you play:
 - **Mars University** while having **Olympus Conference** down:
   - **Olympus** draw then **Mars University** discard.
 - A Science card while having **Mars University** and **Olympus Conference** down:
   - **Olympus** draw then **Mars University** discard.
 - A 20+ cost Science/Earth card while having **Mars University**, **Olympus Conference**, **Spin-off Department** and **Point Luna** down:
   - **Olympus**/**Point Luna**/**Spin-off** draws then **Mars University** discard.

And so on.

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1302,7 +1302,10 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
 
         this.addPlayedCard(game, selectedCard);
 
-        for (const playedCard of this.playedCards) {
+        const playedCardsOrdered = this.playedCards.filter((card) => card.onCardPlayed !== undefined)
+            .sort((a, b) => ((a.onCardPlayedPriority || 999) < (b.onCardPlayedPriority || 999) ? -1 : 1));
+
+        for (const playedCard of playedCardsOrdered) {
             if (playedCard.onCardPlayed !== undefined) {
                 const actionFromPlayedCard: OrOptions | void = playedCard.onCardPlayed(this, game, selectedCard);
                 if (actionFromPlayedCard !== undefined) {
--- a/src/cards/ICard.ts
+++ b/src/cards/ICard.ts
@@ -37,6 +37,7 @@ export interface ICard {
     getRequirementBonus?: (player: Player, game: Game, venusOnly?: boolean) => number;
     getVictoryPoints?: (player: Player, game: Game) => number;
     onCardPlayed?: (player: Player, game: Game, card: IProjectCard) => OrOptions | void;
+    onCardPlayedPriority?: number;
     onStandardProject?: (player: Player, projectType: StandardProjectType) => void;
     onTilePlaced?: (player: Player, space: ISpace, game: Game) => void;
     resourceType?: ResourceType;
--- a/src/cards/OlympusConference.ts
+++ b/src/cards/OlympusConference.ts
@@ -17,6 +17,7 @@ export class OlympusConference implements IProjectCard, IResourceCard {
     public resourceType: ResourceType = ResourceType.SCIENCE;
     public resourceCount: number = 0;
     public name: CardName = CardName.OLYMPUS_CONFERENCE;
+    public onCardPlayedPriority: number = 10;
 
     private runInterrupts(player: Player, game: Game, scienceTags: number): void {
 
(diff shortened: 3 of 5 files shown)


nwai90:
@Lynesth Would this also apply for https://github.com/bafolts/terraforming-mars/issues/1179? Pharmacy Union triggers before GMO contract

src/Player.ts
```diff
@@ -1302,7 +1302,10 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
 
         this.addPlayedCard(game, selectedCard);
 
-        for (const playedCard of this.playedCards) {
+        const playedCardsOrdered = this.playedCards.filter((card) => card.onCardPlayed !== undefined)
```
bafolts, line 1305: Using some defined order for these actions should be more effective than relying on the order the cards were played. Any property we put on `ICard` will end up going to the database since we only run `JSON.stringify` on the game instance and don't consider what values are static versus dynamic. At some point we could eventually only send a `SerializedCard` to the database which would map `ICard` to `SerializedCard` and could avoid this property.

It should be simpler across the board if this just wrote the logic out using `CardName`. We should also consider corporation cards in this logic. If we first collected all the `ICard` from `this.playedCard` and the corporation cards from all players corporation cards we could pass this array through a function which orders them and then iterate this array to add the interrupts.

```typescript
private orderCardPlayedActions(cards: Array<ICard>): Array<ICard> {
    return actions.sort((card) => {
        if (card.name === CardName.OLYMPUS_CONFERENCE || card.name === CardName.SPINOFF_DEPARTMENT || card.name === POINT_LUNA) {
            return 1;
        }
        return 0;
    });
}
```

bafolts (changes requested):

kberg:
Before this goes in, let me ask -- is it better to ask the player to lay out the order as they see fit?

bafolts:
We could consider animating the card in some way. At that point a player would have to click on the card to trigger the action. At that point we could put some marker on the card. That would indicate to the user they have taken the action and allow them to dictate the order.

Lynesth:
I thought there're only a couple of edge cases where player's driven order matters (like playing a Science/Microbe with Pharmacy Union #1286).

In the examples of this PR, there's no reason you wouldn't want to draw cards before picking the one you'd want to discard.

If we start prompting the player with a OrOptions that takes each `PlayerInterrupt` (aka `deferredAction` soon™) as options and let the player pick them one by one, games are bound to take much longer. I feel like they shouldn't be given a choice unless it actually matters.

kberg:
> I feel like they shouldn't be given a choice unless it actually matters.
As do I -- but knowing this maybe is tricky.

kberg:
What is this waiting on?

Lynesth:
@kberg 
@bafolts asked for some changes but I completely forgot about it. I liked the fine tuning allowed by having a priority property but that's probably overkill.

After taking more time to verify things, it seems that only **Olympus Conference** (**OC**) needs to be activated before **Mars University** since the others (**Spin-off Department**, **Point Luna**) don't defer actions and directly push cards in your hand.

I also thought about the bigger picture of having some `onCardPlayed` trigger before the action of the card you are playing, like **OC** drawing a card before **Invention Contest** or **Sponsored Academies** effects are resolved.

I think that adding a second parameter to `game.defer` to unshift the action instead of pushing it could be the way to do things. It would then be used in specific cards' `onCardPlayed`, like **OC** or **Pharmacy Union** (could easily fix what @nwai90 mentioned above when that's done).

Lynesth:
If #1737 is merged I'll close this one.

PR #1575, opened 2020-10-11
Fix Pharmacy Union paying 3MC for its free card

Lynesth:
The free Science card is not its first action but is part of its `play()` method.
Therefore, that card was taken into account when calculating cost of initial hand of cards, effectively making the player start with 43MC instead of 46.

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -797,12 +797,12 @@ export class Game implements ILoadable<SerializedGame, Game> {
       }
 
       player.corporationCard = corporationCard;
-      corporationCard.play(player, this);
       player.megaCredits = corporationCard.startingMegaCredits;
       if (corporationCard.name !== new BeginnerCorporation().name) {
         const cardsToPayFor: number = player.cardsInHand.length;
         player.megaCredits -= cardsToPayFor * player.cardCost;
       }
+      corporationCard.play(player, this);
 
       // trigger other corp's effect, e.g. SaturnSystems,PharmacyUnion,Splice
       for (const somePlayer of this.getPlayers()) {


nwai90 (approved):

src/Game.ts
```diff
@@ -797,12 +797,12 @@ export class Game implements ILoadable<SerializedGame, Game> {
       }
 
       player.corporationCard = corporationCard;
-      corporationCard.play(player, this);
       player.megaCredits = corporationCard.startingMegaCredits;
       if (corporationCard.name !== new BeginnerCorporation().name) {
         const cardsToPayFor: number = player.cardsInHand.length;
         player.megaCredits -= cardsToPayFor * player.cardCost;
```
bafolts, line 803: I don't think that this change will be this straight forward. The `TerralabsResearch` corporation has a `cardCost` of `1` that it needs set prior to charging for cards. Similarly `Polyphemos` has a `cardCost` of `5`. Any corporation which manipulates the card cost has to be applied prior to charging for cards.

bafolts (changes requested):
I am not sure what the solution is yet but this will break corporations which modify `cardCost`.

src/Game.ts
bafolts: If you use the local variable `cardCost` instead of `player.cardCost` this change should work.

Lynesth: Fixed.

Lynesth:
Fixed the issue that arised with **Terralabs Research** and **Polyphemos**.
Also added tests to all 3 (including **Pharmacy Union**) to check for starting hand card cost.

bafolts (approved):

PR #1579, opened 2020-10-11
atmoscoop wording fix

vsrisuknimit:
Many users complains that they often accidentally click the wrong action on Atmoscoop due to the second option being hidden below card option for adding floaters and the confirmation button.

I reworded the actions of Atmoscoop to bring the global parameters to the front of the sentence. See below.

What I would like to do also is to not preselect any of the option when: 
1) More than one floater cards AND
2) Temp and Venus are not maxed out.

This way, the user will first encounter the first option before seeing the cards they need to select for adding floaters. The second option won't be as hidden below the first one as before.

image


vsrisuknimit:
@nwai90 Can you help with not preselecting any option under the specific if statement above?

vsrisuknimit:
It does not seem like there is way not to pre-select, then this is a good as it will get without adding a number of clicks more.

What would be best is AndOption(OrOption(temp, venus), floater-card-adding) so the floater is only shown once. I tried but I can't do OpOption under AndOption.

It is good to merge if someone approves the change.

nwai90:
@vsrisuknimit Localisation file also needs to update to replace with the new text strings (e.g. `src/locales/fr/ui.json`)

vsrisuknimit:
@nwai90 added French translation.

nwai90 (approved):

PR #1588, opened 2020-10-12
Improve player color selection

vsrisuknimit:
- Color boxes instead of radio buttons for picking colors.
- The entire player box changes color to match, making it easier to tell who is playing what color.
- Hiding TR boost and Beginner Corp option under Beginner Options

image

### DIFF
--- a/src/components/CreateGameForm.ts
+++ b/src/components/CreateGameForm.ts
@@ -9,6 +9,7 @@ import { ColoniesFilter } from "./ColoniesFilter";
 import { ColonyName } from "../colonies/ColonyName";
 import { CardsFilter } from "./CardsFilter";
 import { Button } from "../components/common/Button";
+import { playerColorClass } from "../utils/utils";
 
 interface CreateGameModel {
     firstIndex: number;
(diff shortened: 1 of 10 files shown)


vsrisuknimit:
Still has to refactor the color classes as @chosta suggested. More commit coming.

vsrisuknimit:
@chosta . I finished moving classes to common.less. Please check again. I deleted the last PR after my git rebase added a thousand commits.

chosta:
Looks good to me. I like the name change to 'translucent', too.

vsrisuknimit:
Example: 

image

image

vsrisuknimit:
image

What it looks like top and bottom together. Tagging @alrusdi for review. Good to go if you think it is good.

kberg:
I'm a little worried about this being particularly unfriendly to people with color blindness issues. Having the names of the colors would go a long way.

vsrisuknimit:
@kberg I can do something like this but it is unclear how that would help. Users will still have to deal with the problem in the main games of cubes being the same color, even if they can tell the colors apart due to labels on the selection page.

We need to a setting for users to pick their own color palette.

image

vsrisuknimit:
Labels with colorblind simulation.

image

kberg:
Oh I love it! You're the best.

src/components/CreateGameForm.ts
```diff
@@ -361,7 +361,7 @@ export const CreateGameForm = Vue.component("create-game-form", {
 
                     <div class="create-game-page-container">
                         <div class="create-game-page-column" v-if="! isSoloModePage">
-                            <h4 v-i18n># of Players</h4>
+                            <h4 v-i18n>№ of Players</h4>
```
alrusdi, line 364: Please replace original string "Players count" with proposed "№ of Players" in all translation files

alrusdi (approved):

src/components/CreateGameForm.ts
vsrisuknimit: All set. I did not change the translation though.

PR #1590, opened 2020-10-12
Dynamically set cardTitle widths and trim Promo text from titles

nwai90:
Screenshot 2020-10-13 at 1 56 40 AM
Screenshot 2020-10-13 at 1 56 48 AM
Screenshot 2020-10-13 at 2 07 10 AM

- Set cardTitle width dynamically based on title length
- Drop "Promo" text from promo card titles
- Fix CSS for ProjectWorkshop

### DIFF
--- a/src/components/card/CardTitle.ts
+++ b/src/components/card/CardTitle.ts
@@ -34,20 +34,26 @@ export const CardTitle = Vue.component("CardTitle", {
                 classes.push("background-color-prelude");
             }
 
-            if (title.length > 26) {
+            const trimmedTitle = this.getCardTitleWithoutPromoText(title);
+
+            if (trimmedTitle.length > 26) {
                 classes.push("title-smaller");
-            } else if (title.length > 23) {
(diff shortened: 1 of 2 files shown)


chosta (approved):

kberg (approved):

src/components/card/CardTitle.ts
```diff
@@ -32,14 +33,27 @@ export const CardTitle = Vue.component("CardTitle", {
...
+            if (trimmedTitle.length > 26) {
+                classes.push("title-smaller");
+            } else if (trimmedTitle.length > 23) {
+                classes.push("title-small");
+            }
+
             return classes.join(" ");
         },
+        getCardTitleWithoutPromoText(title: string): string {
+            if (title.toLowerCase().includes("promo")) return title.substring(0, title.length - 6);
```
nwai90, line 48: Minor note: If there are ever future cards that include "promo" in their name this should update to check for strings that end with "promo" instead. It is fine for now though

PR #1600, opened 2020-10-13
Add :promo to promo cards, hide it from LogPanel and CardTitle. Used in Ares, too.

kberg:


src/components/LogPanel.ts
```diff
@@ -29,16 +29,22 @@ export const LogPanel = Vue.component("log-panel", {
...
+                color="background-color-events";
             } else if (cardType === CardType.ACTIVE) {
-                return "<log-card class=\"background-color-active\">"+cardName+"</log-card>";
+                color="background-color-active";
             } else if (cardType === CardType.AUTOMATED) {
-                return "<log-card class=\"background-color-automated\">"+cardName+"</log-card>";
+                color="background-color-automated";
             } else if (cardType === CardType.PRELUDE) {
-                return "<log-card class=\"background-color-prelude\">"+cardName+"</log-card>";
+                color="background-color-prelude";
```
alrusdi, line 41: Shouldn't this function return the **color** variable somewhere?

src/components/card/CardTitle.ts
```diff
@@ -44,16 +44,16 @@ export const CardTitle = Vue.component("CardTitle", {
 
             return classes.join(" ");
         },
-        getCardTitleWithoutPromoText(title: string): string {
-            if (title.toLowerCase().includes("promo")) return title.substring(0, title.length - 6);
-            return title;
+        getCardTitleWithoutSuffix(title: string): string {
+            // split will always return at least one element.
```
alrusdi, line 48: Comment is not necessary

src/CardName.ts
```diff
@@ -410,9 +410,9 @@ export enum CardName {
     PENGUINS = "Penguins",
     SMALL_ASTEROID = "Small Asteroid",
     SNOW_ALGAE = "Snow Algae",
-    DEIMOS_DOWN_PROMO = "Deimos Down Promo",
-    MAGNETIC_FIELD_GENERATORS_PROMO = "Magnetic Field Generators Promo",
-    GREAT_DAM_PROMO = "Great Dam Promo",
+    DEIMOS_DOWN_PROMO = "Deimos Down:promo",
```
bafolts, line 413: Only seeing this `:promo` in the `CardName` it would be near impossible to know that in `LogPanel` this was driving logic. This is the case today with the ` Promo` suffix. When used in `ares` is this again used to override cards? Currently the `projectCardsToRemove` are always replaced with some other card. If `projectCardsToRemove` were `projectCardsToReplace` and provided the mapping from `DEIMOS_DOWN` to `DEIMOS_DOWN_PROMO` we could then look up the card name in that map. If we find the card name in the values of that map we can display the key.

Then in `LogPanel` the `parseCardType` could use `function (cardType: CardType, cardName: CardName)` which it should be as the value is not only a `string`.

kberg: > When used in ares is this again used to override cards?
Yes, that was the agreement early in the FR.

I think the solution is more complicated than what your proposal would fix as all of it is predicated on the CardName enum's value being a display name, and also as a unique key in maps (eg HTML_data.ts)

I suggest leaving this as is following along with a correction later.

src/components/LogPanel.ts
kberg: It's a bad name. Changing color to className.

src/components/card/CardTitle.ts
kberg: Mmm.... yeah OK.

alrusdi (approved):
Great.
You don't need to make additional changes to rerun the failed tests btw!
Just press **red cross** of failed build, then press **Details**, in the top right corner of the build details hit **Re-run jobs**

PR #1604, opened 2020-10-13
making all buttons the same height

vsrisuknimit:
- Minor UI adjustment on the create game page to satisfy Simon's OCD :P
- More padding when Beginner Option is turned on.

### DIFF
--- a/src/components/CreateGameForm.ts
+++ b/src/components/CreateGameForm.ts
@@ -428,7 +428,7 @@ export const CreateGameForm = Vue.component("create-game-form", {
                             <template v-for="boardName in boards">
                                 <input type="radio" :value="boardName" name="board" v-model="board" :id="boardName+'-checkbox'">
                                 <label :for="boardName+'-checkbox'" class="expansion-button">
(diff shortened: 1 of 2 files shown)


src/components/CreateGameForm.ts
```diff
@@ -428,7 +428,7 @@ export const CreateGameForm = Vue.component("create-game-form", {
                             <template v-for="boardName in boards">
                                 <input type="radio" :value="boardName" name="board" v-model="board" :id="boardName+'-checkbox'">
                                 <label :for="boardName+'-checkbox'" class="expansion-button">
-                                    <span :class="getBoardColorClass(boardName)">&#x2B22;</span><span class="capitalized" v-i18n>{{ boardName }}</span>
+                                    <span :class="'create-game-board-hexagon '+getBoardColorClass(boardName)">&#x2B22;</span><span class="capitalized" v-i18n>{{ boardName }}</span>
```
alrusdi, line 431: Why not to add this class in getBoardColorClass function?

vsrisuknimit: You are right. I moved.

alrusdi (approved):

PR #1608, opened 2020-10-14
refactor: use gzip for main.js

bafolts:
Well our javascript seems to like being compressed.

```
Built at: 10/13/2020 6:12:38 PM
         Asset      Size  Chunks                          Chunk Names
       main.js  1.33 MiB       0  [emitted]        [big]  main
    main.js.gz   275 KiB          [emitted]        [big]  
   main.js.map  2.81 MiB       0  [emitted] [dev]         main
main.js.map.gz   524 KiB          [emitted]        [big]  
```

Now users will only need to download 275kb to load the app.


kberg:
Seems good to me.

kberg (approved):
before:
image

after: 
image

PR #1613, opened 2020-10-14
Fix Energy Tapping and Power Supply Consortium play requirements

Lynesth:
You can increase your own energy production and then decrease it.
There's really little reasons you would do that (maybe playing Energy Tapping into Magnetic Field or delaying your action late game) but it would be valid. So there we go.

### DIFF
--- a/src/cards/EnergyTapping.ts
+++ b/src/cards/EnergyTapping.ts
@@ -1,29 +1,25 @@
-
-import {IProjectCard} from './IProjectCard';
-import {Tags} from './Tags';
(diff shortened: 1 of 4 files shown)


nwai90 (approved):

nwai90:
> You can increase your own energy production and then decrease it.
> There's really little reasons you would do that (maybe playing Energy Tapping into Magnetic Field or delaying your action late game) but it would be valid. So there we go.

👍 Magnate award and several other possible reasons too

bafolts (approved):

PR #1614, opened 2020-10-14
change class of end game player name

vsrisuknimit:
- Change the player username class to be the same as in log.

New: image

Previous:
image

### DIFF
--- a/src/components/GameEnd.ts
+++ b/src/components/GameEnd.ts
@@ -18,8 +18,8 @@ export const GameEnd = Vue.component("game-end", {
         isSoloGame: function (): boolean {
             return this.player.players.length === 1;
         },
-        getPlayerColorStyle: function (player: PlayerModel): string {
-            return "color: " + player.color;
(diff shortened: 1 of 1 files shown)


nwai90:
This looks great! Makes the player names a lot more visible 👍

kberg:
Looks great.

src/components/GameEnd.ts
```diff
@@ -18,8 +18,8 @@ export const GameEnd = Vue.component("game-end", {
         isSoloGame: function (): boolean {
             return this.player.players.length === 1;
         },
-        getPlayerColorStyle: function (player: PlayerModel): string {
-            return "color: " + player.color;
+        getEndGamePlayerColorClass: function (player: PlayerModel): string {
```
chosta, line 21: @vsrisuknimit this could be :) 
`playerColorClass("bg", player.color)`

p.s. maybe add it as a todo for a next PR and do a separate commit for it

PR #1615, opened 2020-10-15
create settings.json for passing server settings to client

bafolts:
During the recent change to display the server version on the client I took a rudimentary approach. There was a request to pass another server variable to the client to configure how often the client checks for updates. This change creates a `settings.json` similar to the `translations.json`. This file is then passed to the `App` component making it available on the client. In the future if we need to pass more data from server to client we can add it here.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,8 +36,8 @@
   },
   "scripts": {
     "clean": "rm -r ./dist/ ./styles.css",
-    "build": "node make_locales.js && eslint src/*.ts src/database/**/*.ts && lessc src/styles/common.less styles.css && tsc --build tsconfig.json && webpack --config webpack.config.js",
-    "start": "node make_locales.js && lessc src/styles/common.less styles.css && tsc --build tsconfig.json && webpack --config webpack.config.js && node dist/server.js",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -36,8 +36,8 @@
   },
   "scripts": {
     "clean": "rm -r ./dist/ ./styles.css",
-    "build": "node make_locales.js && eslint src/*.ts src/database/**/*.ts && lessc src/styles/common.less styles.css && tsc --build tsconfig.json && webpack --config webpack.config.js",
-    "start": "node make_locales.js && lessc src/styles/common.less styles.css && tsc --build tsconfig.json && webpack --config webpack.config.js && node dist/server.js",
+    "build": "node make_static_json.js && eslint src/*.ts src/database/**/*.ts && lessc src/styles/common.less styles.css && tsc --build tsconfig.json && webpack --config webpack.config.js",
```
kberg, line 39: This has much in common with start, why not just call npm build first? /shrug

kberg (approved):
Works correctly locally.

image

Verified on Heroku:
image

package.json
bafolts: Yes. It is looking like that makes the most sense.  I will put PR up for that.

PR #1616, opened 2020-10-15
npm run clean now doesn't return an error code if rm does nothing.

kberg:
Result is that `npm run clean && npm run build` will build even if the work area is already clean.

```
$ npm run clean && npm run build

> terraforming-mars@1.0.0 clean /home/***/***/terraforming-mars
> rm -r ./dist/ ./styles.css || true

rm: cannot remove './dist/': No such file or directory
rm: cannot remove './styles.css': No such file or directory

> terraforming-mars@1.0.0 build /home/***/***/terraforming-mars
> node make_locales.js && eslint src/*.ts src/database/**/*.ts && lessc src/styles/common.less styles.css && tsc --build tsconfig.json && webpack --config webpack.config.js
```


package.json
```diff
@@ -35,7 +35,7 @@
     "extends": "@istanbuljs/nyc-config-typescript"
   },
   "scripts": {
-    "clean": "rm -r ./dist/ ./styles.css",
+    "clean": "rm -r ./dist/ ./styles.css || true",
```
Lynesth, line 38: Why not simply use -f ?
`"clean": "rm -rf ./dist/ ./styles.css", `

bafolts (approved):
Windows keeps gaining more unix support and I am not sure where it stands these days but there is a popular `rimraf` which would work across every operating system.

https://www.npmjs.com/package/rimraf

package.json
kberg: if there are permission issues deleting dist and styles.css, that's a problem. I'd rather minimize destructiveness.

PR #1617, opened 2020-10-15
Remove CorporationGroup and CorporationNames.

kberg:
I don't like how this mixes in CorporationFilter yet, and it's lost the module name in the process, but that's recoverable. @chosta, please take a look.


kberg:
Mea culpa, I forgot a few things. Should be good now.

chosta (approved):
Looks good. As long as the debug-ui opens up and you see the expansion icons it should be fine. I like the approach, reduces boilerplate a lot.

src/components/CorporationsFilter.ts
```diff
@@ -1,67 +1,44 @@
...
-import { CORP_ERA_CARD_MANIFEST, BASE_CARD_MANIFEST } from "../cards/StandardCardManifests";
-import { PROMO_CARD_MANIFEST } from "../cards/promo/PromoCardManifest";
-import { TURMOIL_CARD_MANIFEST } from "../cards/turmoil/TurmoilCardManifest";
-import { CardManifest } from "../cards/CardManifest";
-import { COMMUNITY_CARD_MANIFEST } from "../cards/community/CommunityCardManifest";
+import { ALL_CARD_MANIFESTS, ALL_CORPORATION_DECKS, MANIFEST_BY_MODULE } from "../cards/AllCards";
+import { GameModule } from "../GameModule";
 
+function corpCardNames(module: GameModule): Array<CardName> {
+    return MANIFEST_BY_MODULE.get(module)!.corporationCards.cards.map(cf => cf.cardName);
```
bafolts, line 8: Using the `!` operator is dangerous. If `module` is not found in the map this will throw a runtime exception. If `MANIFEST_BY_MODULE.get(module) === undefined` this can return `[]`.

src/components/CorporationsFilter.ts
```diff
@@ -1,67 +1,44 @@
...
+            cardsByModuleMap: cardsByModuleMap,
             customCorporationsList: false,
             selectedCorporations: [
-                ...this.corporateEra ? corpCardNames.original : [],
-                ...this.prelude ? corpCardNames.prelude : [],
-                ...this.venusNext ? corpCardNames.venus : [],
-                ...this.colonies ? corpCardNames.colonies : [],
-                ...this.turmoil ? corpCardNames.turmoil : [],
-                ...this.promoCardsOption ? corpCardNames.promo : [],
+                ...this.corporateEra ? cardsByModuleMap.get(GameModule.CorpEra)! : [],
```
bafolts, line 25: The `!` operator is dangerous. I would replace these with:

```
...this.corporateEra ? (cardsByModuleMap.get(GameModule.CorpEra) || []) : []
```

src/components/CorporationsFilter.ts
```diff
@@ -1,67 +1,44 @@
...
     methods: {
         getItemsByGroup: function (group: string): Array<CardName> {
             if (group === "All") return allItems.slice();
 
-            const corps = this.corporationGroups.find((g) => g.title === group);
-            if (corps === undefined) return [];
+            const corps = this.cardsByModuleMap.get(group as GameModule);
                         
-            return (corps.items as any).slice()
+            return corps!.slice();
```
bafolts, line 41: If `corps` is `undefined` this can return `[]`.

src/components/card/CardExpansion.ts
```diff
@@ -1,6 +1,15 @@
 import Vue from "vue";
-import { CorporationGroup } from "../../CorporationName";
+import { GameModule } from "../../GameModule";
 
+const MODULE_TO_CSS: Map<string, string> = new Map([
```
bafolts, line 4: This generic type should be able to use `Map<GameModule, string>`

bafolts (changes requested):
I would not use `!` operator. Especially on the server this can end up killing the node process as it introduces `undefined` where it wasn't expected. The enum consolidation is good.

src/components/card/CardExpansion.ts
kberg: There were some issues when I tried that - can't remember what. Mind if I add a TODO?

src/components/CorporationsFilter.ts
kberg: Yeah, I saw you having issues with ! before, but here me out. MANIFEST_BY_MODULE is a map keyed by GameModule, so under what conditions could this actually occur? That's why I think it's safe here. My personal experience is that returning [] under what is a guaranteed impossible occurrence means that when it happens it hides the error.

If you disagree and you have a reasonable rebuttal, let me know. I'm also not going to fight for it, but it's important you know my reasons -- a few more of these! are going to come up with Ares (particularly inside AresHandler -- but more on that another time.)

src/components/CorporationsFilter.ts
kberg: Same -- if this happens there's a real error. LMK your thoughts.

You also don't have to warn me about the ! ongoing you can at least save those characters in your feedback, and Ill be happy to discuss / address them appropriately.

src/components/CorporationsFilter.ts
kberg: In this particular case -- is it possible `group as GameModule` could create an item that [i]isn't[/i] a GameModule? Given that CorporationsFilter is a purely internalized class, that just shouldn't happen. I know I'm repeating myself, but again if that happens, it's an error -- why not raise it as one?

src/components/CorporationsFilter.ts
kberg: I see that you marked it for all of them -- feel free to not have to deeply respond to each one as if I asked them independently -- feel free to respond to all of them right here, or something.

kberg:
OK. Is there a way to report this as a warning? So we can check logs for invalid behavior instead of silently accepting it?

kberg:
> I would not use `!` operator. Especially on the server this can end up killing the node process as it introduces `undefined` where it wasn't expected. The enum consolidation is good.

Ah wait a second -- this isn't the server. It's the client. Same concerns?

src/components/CorporationsFilter.ts
bafolts: This is a front end component so it wouldn't take down the game server. It is most likely not possible as written today but in the future during some re-factor it broke in some way this could end up throwing that `slice` is not a function of `undefined`. I don't think the user would see this thrown error. If this were to return `[]` the user most likely also wouldn't see any error to know that something went wrong either. Since typescript tries to warn us that the item can be `undefined` I always find it strange when `!` is used as typescript is most likely going to be right at some point in the future and there will be a runtime exception here. If this instead returned an array I guess we wouldn't spot it either unless we logged a warning. As this is client side I am not as concerned as this at least wont crash the game server which would take down all other games running on the process. Could consider using `console.assert` as this is asserting that `corps` will be defined. At that point if it were ever an issue there would be an error message logged to the console.

bafolts (approved):
I am less concerned with the `!` usage on client as that will potentially error on that component and not kill the node process if it goes uncaught. I don't think we will show these errors to the user so if `!` were used or `[]` returned either way the user is most likely going to experience a bug.

kberg:
I fixed as many as I could -- hope I caught them all.

kberg:
Incidentally, had a little bit more of a chat with Brian off-list, I'll stop using ! everywhere.

PR #1622, opened 2020-10-15
allowing custom corp even without expansion

vsrisuknimit:
**Problem**: Currently, you cannot play any expansion corp without the expansion it comes from. For example, you cannot play Manutech without Venus Next expansion. Previously, this was possible.

**Source**: #1444 changed the code such that custom corps only filter on top of the corps available by selected expansion(s) instead of all corps as it was previously.

**Fix:** Getting CorporationCard from the custom corp list and use this instead of the corps available by selected expansion(s). 

I tested it, it works, however, two problems.
1) This is probably a bad code, there must be a way of doing this. Do I need to use Decks.findByName()? 
2) I would like to throw an error but I don't think I can do that in a constructor. I get "Unexpected Server Response" instead of the error message I put in. This will also cause `npm run start` to fail if there is any existing game that might cause this error. I had to `rm db/game.db && touch db/game.db` before I can `npm run start`.

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -262,8 +262,6 @@ export class Game implements ILoadable<SerializedGame, Game> {
             }
         }
         corporationCards = customCorporationCards;
-      } else if (gameOptions.customCorporationsList && gameOptions.customCorporationsList.length < minCorpsRequired){
-        throw new Error("Not enough custom corps selected for all players");
       }
 
       corporationCards = this.dealer.shuffleCards(corporationCards);
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -59,6 +59,8 @@ import { getRandomMilestonesAndAwards } from "./MilestoneAwardSelector";
 import { CardType } from "./cards/CardType";
 import { ColonyModel } from "./models/ColonyModel";
 import { LogBuilder } from "./LogBuilder";
+import { Decks } from "./Deck";
+import { ALL_CORPORATION_DECKS } from "./cards/AllCards";
 
 export interface Score {
   corporation: String;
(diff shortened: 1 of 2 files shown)


vsrisuknimit:
I also don't understand why the old check for undefined again in the inner if statement.

https://github.com/bafolts/terraforming-mars/blob/5ad22d15074e51ae940a6cd5fcb0b126cb865a8e/src/Game.ts#L256

kberg:
> **Problem**: Currently, you cannot play any expansion corp without the expansion it comes from. For example, you cannot play Manutech without Venus Next expansion. Previously, this was possible.
> 
> **Source**: #1444 changed the code such that custom corps only filter on top of the corps available by selected expansion(s) instead of all corps as it was previously.

Good catch. I apologize!

> **Fix:** Getting CorporationCard from the custom corp list and use this instead of the corps available by selected expansion(s).
> 
> I tested it, it works, however, two problems.
> 
> 1. This is probably a bad code, there must be a way of doing this. Do I need to use Decks.findByName()?
That's the way to do it. Or you can just write what that does by yourself, but why would you?

> 2. I would like to throw an error but I don't think I can do that in a constructor. I get "Unexpected Server Response" instead of the error message I put in. This will also cause `npm run start` to fail if there is any existing game that might cause this error. I had to `rm db/game.db && touch db/game.db` before I can `npm run start`.
Does that mean that someone has to create a db from before this in order to test that your change will not break other services?

src/Game.ts
```diff
@@ -251,10 +253,15 @@ export class Game implements ILoadable<SerializedGame, Game> {
 
       const minCorpsRequired = players.length * gameOptions.startingCorporations;
       if (gameOptions.customCorporationsList && gameOptions.customCorporationsList.length >= minCorpsRequired) {
-
-        corporationCards = this.dealer.corporationCards.filter(
-          (corpCard) => gameOptions !== undefined && gameOptions.customCorporationsList.includes(corpCard.name)
-        );
+        
+        const customCorporationCards: CorporationCard[] = [];
```
kberg, line 257: The way I'm reading this is: use the custom corp list ONLY IF you supply enough to satisfy the minCorpsRequired. Is that the behavior you want?

kberg (changes requested):
I can help you with a test to prevent this issue in the future. But if you get started on one I'd be grateful.

src/Game.ts
vsrisuknimit: @kberg that is correct. It will use customized corps only if provided enough. 

I remove the error throw in case it is not enough since the app will resource to external "Unexpected Server Response" and only use my error message internally. The error also breaks many tests.

vsrisuknimit:
I have no idea how to write the test. I will leave that to other folks.

kberg:
Sent you a PR with a test.

>

vsrisuknimit:
Closing. #1635 is the same but cleaner commits.

PR #1625, opened 2020-10-16
Fixes #1624 missing VP count for self

### LINKED ISSUE #1624: Show player's VP without showing all VPs?
Once upon a time, the default was to see your own total victory points during the game. The current releases don't seem to do that, although the realtime VP does show everyone's VPs. 

I think this is a bug, and as near as I can find the issue is with getIsActivePlayer in playerInfo.ts -- this.player.id seems to return the player color rather than the player ID, although I can't figure out how that's happening.

puuj:


src/components/overview/PlayerInfo.ts
```diff
@@ -28,7 +28,7 @@ export const PlayerInfo = Vue.component("player-info", {
             return classes.join(" ");
         },
         getIsActivePlayer: function (): boolean {
-            return this.player.id === this.activePlayer.id;
+            return this.player.id === this.activePlayer.color;
```
kberg, line 31: How is this possibly correct? (I'm not doubting you @puuj but this is surprising.)

chosta:
@puuj you were correct to look for the problem in getActivePlayer but I found one more so I've made a separate PR #1630

We can close this one. Thanks for pointing it out!

src/components/overview/PlayerInfo.ts
chosta: @kberg maybe some unit tests would have caught this but they are such a hassle for frontend. Especially clicking on buttons and similar.. This one has been active for a month at least.

PR #1633, opened 2020-10-16
expose player id at end of game

bafolts:
Exposes player id at end of game

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -747,7 +747,7 @@ function getPlayers(players: Array<Player>, game: Game): Array<PlayerModel> {
             energyProduction: player.getProduction(Resources.ENERGY),
             heat: player.heat,
             heatProduction: player.getProduction(Resources.HEAT),
-            id: player.color,
+            id: game.phase === Phase.END ? player.id : player.color,
             megaCredits: player.megaCredits,
             megaCreditProduction: player.getProduction(Resources.MEGACREDITS),
             name: player.name,
--- a/src/components/GameEnd.ts
+++ b/src/components/GameEnd.ts
@@ -94,7 +94,7 @@ export const GameEnd = Vue.component("game-end", {
                         </thead>
                         <tbody>
                             <tr v-for="p in getSortedPlayers()">
-                                <td><log-player :class="getEndGamePlayerColorClass(p)">{{ p.name }}</log-player></td>
+                                <td><a :href="'/player?id='+p.id+'&noredirect'" :class="getEndGamePlayerColorClass(p)">{{ p.name }}</a></td>
                                 <td v-i18n>{{ p.corporationCard.name }}</td>
                                 <td>{{ p.victoryPointsBreakdown.terraformRating }}</td>
                                 <td>{{ p.victoryPointsBreakdown.milestones }}</td>


chosta:
This would make the specator view easier to implement. Good change.

bafolts:
Made more progress on this. Using `{} as PlayerModel` seems to be different than `const model: PlayerModel = {}`. This has uncovered some properties we weren't providing and others I can't tell if we needed. Since the vue templates don't enforce our types I will need to go through each one to double check it wont break anything.

kberg:
Are you ready for review? I haven't started looking at this, let me know if you want me to.

bafolts:
I will move this out of a draft pull request when it is ready. I wish vue.js had better type checking on the templates. As a back story I just happened to pick vue.js while I was working on this project in my spare time. I have no strong attachment to it. The early UI was largely text based. Without the type checking in the templates I am going to sanity check any properties moved or re-named. If we are still checking for player.id some what I miss this will cause bugs. Also need to write some mechanism to run client side tests at some point so client side changes aren't as much of a gamble.

kberg:
I've been thinking about writing a client-side sanity test just to break
the ice. But I know Dani would prefer another UI framework. If you're going
to make that move though, well, that'll be interesting. Perhaps create a
tracking bug to discuss?

bafolts:
Vue.js not having type validation in the templates makes things very difficult even with tests. With type validation in the template this refactor would be straight forward as I would have certainty I didn't miss any usage. I can create a separate tracking bug to discuss.

bafolts:
I went a simpler way with this for the time being. I will start working on upgrading vue to give us compile time checking. Before we have that most client side changes are very risky.

bafolts:
@kberg I went with a simpler approach here for time being. I plan to put in script during build process to add type checking to the vue components. After that making larger changes to what we expose through models will be more straight forward.

server.ts
```diff
@@ -747,7 +747,7 @@ function getPlayers(players: Array<Player>, game: Game): Array<PlayerModel> {
             energyProduction: player.getProduction(Resources.ENERGY),
             heat: player.heat,
             heatProduction: player.getProduction(Resources.HEAT),
-            id: player.color,
+            id: game.phase === Phase.END ? player.id : player.color,
```
kberg, line 750: This comment is for a separate PR, probably, and I think not particularly controversial, but how do we get to this?

```
id: game.phase === Phase.END ? player.id : undefined,
color: player.color,
```

kberg (approved):
Approve because id is misused for color and this is a net improvement.

server.ts
bafolts: Once the type checking is in place for the components I want to create a `PlayerModel` which represents the person playing the game and would have private information such as cards in hand and the player id. There will also be a `PublicPlayerModel` or information about a player that you could see if you were playing with them around a table. The will not have the player id and would use color. All of this today is stored on `PlayerModel` which makes it hard to distinguish between the two. We migrated to using `player.color` to avoid displaying the `player.id` but I don't think we updated all the places on the client and server to always check color. With vue type checking in place the compiler will find all the usages for us on the client which will make the change easy.

PR #1634, opened 2020-10-16
Settings upload download prototype

alrusdi:
Ability to upload and download (no server side involved) game settings on game creation form
upload_download

### DIFF
--- a/src/components/CreateGameForm.ts
+++ b/src/components/CreateGameForm.ts
@@ -134,7 +134,38 @@ export const CreateGameForm = Vue.component("create-game-form", {
             .then(onSucces)
             .catch(_ => alert("Unexpected server response"));
     },
-    methods: { 
+    methods: {
+        downloadCurrentSettings: function () {
+            const serializedData = this.serializeSettings();
+            if (serializedData) {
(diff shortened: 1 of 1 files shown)


kberg:
As a feature this is something many people would like. Would you be interested creating a different behavior? Particularly, I think not doing upload/download. For example this:

image

where you can provide the JSON right there -- paste in your own, or copy it out. Also provide a URL with the same behavior that a user can copy.

Just thinking. What you have is great.

alrusdi:
> Would you be interested creating a different behavior? 
I think this "file based" approach is good enough as a starting point. JSON with really useful template can be huge and not good to edit/copy/paste it

nwai90 (approved):

nwai90:
👍 A file that users can easily edit to provide their own settings is a nice starting point. I think validating the uploaded JSON file or returning an error to the user would be a good next step to handle any invalid input that was provided. In most cases the invalid input is simply ignored but there may be some unintended effects (either visually or pertaining to actual game setup)

Screenshot 2020-10-20 at 1 27 15 AM

Issue #1621, opened 2020-10-15
Possible issue with gardener milestone


chosta:
From discord `@smallman`
`minor bug:  After about 30 games of random milestones, I have NEVER seen the gardener milestone.  Some devs mentioned they already suspected this wasn't truely random.  Atleast half have been on Hellas map.  Would also love an option to have no milestones at all, they are way too luck dependant for my liking.`

Can we somehow make a dry run of lets say 10k randomizations and get some statistics on the frequency of each M or A?

bafolts:
Using the `tests/MilestoneAwardSelector.ts` one could call to `getRandomMilestonesAndAwards` multiple times. I am not sure if the logic can guarantee that 3% of the time one sees the gardener milestone. I would assume after 10k games we would expect to see it.

chosta:
Well, the real question is whether we have a milestone (or more) that is heavily underutilized.

bafolts:
> I am not sure if the logic can guarantee that 3%

I had to say 3%.....LOVE this chart :-)

Lynesth:
Well, Gardener Cultivator and Tycoon all have 7 or more pairs of synergy higher than 1.
Estate Dealer, Desert Settler, Landlord have 6.
Planner, Tactitian, Generalist and Legend have none.

This explains that.

kberg:
This graph shows that each milestone/award has a nearly equal chance of occurring: 
![image](https://user-images.githubusercontent.com/413481/96198096-d3ff6680-0f21-11eb-9c82-fc03eeb60b3e.png)

(Minor code hack just to test that assumption.)

kberg:
I did synergies 1-9. Unsurprisingly they converge: 
![image](https://user-images.githubusercontent.com/413481/96200243-7110ce00-0f27-11eb-83e5-c2c4044c965c.png)

Hard to see -- I tried making a 3d graph but no luck. Anyway, the raw data is here https://docs.google.com/spreadsheets/d/1TwB1yF9USTlZt4egbqHx0uE4Fs8luuTNnh0R3y31NC0/edit#gid=323909500

Lynesth:
We pick 10 Milestones and Awards out of a total of 30.

If Gardener is in there, there's 32% chance that there's also another MA that has a score of 2+ with it. So when you add the chances that the others end up with a bad pair as well, I'm not surprised that we get a ~3% chance of getting Gardener.

vsrisuknimit:
[...]

Issue #1311, opened 2020-09-16
Feature request: limited and unlimited randomized milestones and awards


vsrisuknimit:
Some users voiced their opinions that they are fine with unlimited synergy milestones and awards.

Perhaps offer that as a choice.

(toggle) randomized milestones and awards 
  (radio box) limited synergy - [default]
  (radio box) no-limit synergy

For the limited synergy, I propose the following rules:
- no pair that is 7+ in synergy score (prohibit landlord & desert settler or terraformer & benefactor)
- not more than one pair that is 4+ in synergy score (allow one mayor & landlord or gardener & polar explorer)
- total less than or equal to 15 in synergy score

I think this is more flexible than the current max 1 synergy for all pairing rule.

samyelias:
Are you referring to the synergy scores from https://ssimeonoff.github.io/ma-randomizer ? (Basically from https://boardgamegeek.com/thread/2023541/quantifying-milestone-and-awards-synergy-randomiza )

I forked my own version of ssimeonoff's randomizer, found here: https://samyelias.github.io/ma-randomizer . There are a few main differences: 

- I added a filter for max pair synergies; I personally never change the default value from 3 (excluding 4+ synergy pairs)
- I added weights to the randomization process to ensure a more equal probability of getting a particular goal. Otherwise, because of the synergy scores, certain goals were much more likely (like Tactician) and others much less likely (like Landlord
- Getting Desert Settler and Estate Dealer together is not good for any map except the one it first appeared on, so I made it randomly choose a map as well (shown in the readout).
- I tweaked the synergy scores compared to what's on those other sources: https://docs.google.com/spreadsheets/d/1dRpa9eUXjvPWg9-xyvaWrnD24I8ROXntFxRnswNBg40/edit#gid=330638583

Basically these are the excluded interactions I play with, that the weights for my randomizer are currently optimized for: 

TERRAFORMER and BENEFACTOR, 9
MAYOR and POLAR EXPLORER, 8
GARDENER and ECOLOGIST, 4
GARDENER and POLAR EXPLORER, 8
GARDENER and CULTIVATOR, 9
BUILDER and TYCOON, 4
BUILDER and CONTRACTOR, 9
SPECIALIST and ENERGIZER, 4
ECOLOGIST and DIVERSIFIER, 4
TYCOON and MAGNATE, 9
POLAR EXPLORER and LANDLORD, 4
POLAR EXPLORER and DESERT SETTLER, 8
[...]

PR #1637, opened 2020-10-16
add options for limited and full random milestones and awards

vsrisuknimit:
Before clicking random Ms/As
image
After turning on random Ms/As, two options are provided.
image

Options are:
- Limited synergy as proposed in #1311 (slight change, max total = 20 instead of 15, number of high synergy = 2 instead of 1)
- Full random (does not care at all about synergy)

The distribution of the limited synergy option is as follow:
image


vsrisuknimit:
@Lynesth I rewrote the algorithm, I believe it is much cleaner now.

@chosta I probably screw up some vue things. If you could take a look, it would be nice. The function seems to randomize properly. I also want the flex items ("Limited Synergy" and "Full Random") to span the whole row. Not sure how to do that.

vsrisuknimit:
@nwai90 I would like to add the following to the wiki. https://github.com/bafolts/terraforming-mars/wiki/Variants#random-milestones-and-awards (limited synergy subsection)

**Limited Synergy Milestones and Awards Randomization**
Only allows combination of milestones and awards that follow the conditions listed below. Synergy scores are based on this [synergy matrix](https://docs.google.com/spreadsheets/d/1PIUzO9XLF-MHSrTDP9QU2VW3ce7hJNciAR-G7N-RBUA/edit#gid=330638583).
1) Total synergy score is 20 or lower. (Tharsis Board has total synergy score around 20. The other two boards have much lower scores.)
2) No pair with synergy score 7 or above. (Prohibit landlord & desert settler or terraformer & benefactor.)
3) No more than two pairs that are 4+ in synergy score. (Allow two mayor & landlord or gardener & polar explorer.)

nwai90:
Updated the wiki page

chosta:
@vsrisuknimit Vue is fine. I'd suggest, when you have double option on a single line to use the full width of the column. Have the two elements with same width and center the content. Let me know if that makes sense. This is done best with `flex` and `justify-content` etc. I can help.

vsrisuknimit:
The box stretch to full column width now.

image

vsrisuknimit:
With help from @Lynesth 

image

vsrisuknimit:
This PR can close #1621 and #1311 .

Tagging @alrusdi @nwai90 for review.

src/Game.ts
```diff
@@ -55,10 +55,12 @@ import { IParty } from "./turmoil/parties/IParty";
 import { OrOptions } from "./inputs/OrOptions";
 import { LogHelper } from "./components/LogHelper";
 import { ColonyName } from "./colonies/ColonyName";
-import { getRandomMilestonesAndAwards } from "./MilestoneAwardSelector";
+import { getRandomMA_OneByOne } from "./MilestoneAwardSelector";
```
kberg, line 58: Please rename this -- In the PR where getRandomMilestonesAndAwards was introduced, it was to get away from unclear terms like MASynergy. Not to mention that _ isn't a standard methodName.

server.ts
```diff
@@ -443,6 +443,7 @@ function createGame(req: http.IncomingMessage, res: http.ServerResponse): void {
                 startingCorporations: gameReq.startingCorporations,
                 shuffleMapOption: gameReq.shuffleMapOption,
                 randomMA: gameReq.randomMA,
+                randomMAOption: gameReq.randomMAOption,
```
kberg, line 446: There's no need for randomMA and randomMAOption. Also, having randomMAOption as a string means it's hard to figure out how to track down the valid values. Please replace it with an enum. In fact, consider this

```
enum RandomMilestoneAward {
  NONE = "none,
  LIMITED = "limited,
  UNLIMITED = "unlimited"
}

randomMA: RandomMilestoneAward

File: src/MilestoneAwardSelector.ts [hunk 3]
```diff
@@ -206,3 +206,106 @@ export function computeSynergy(indexes: Array<number>) : number {
...
+
+// Selects |numberMARequested| milestones and |numberMARequested| awards from all available awards and milestones (optionally including
+// Venusian.) It does this by following these rules:
+// 1) No pair with synergy above |maxSynergyAllowed|.
+// 2) Total synergy is |totalSynergyAllowed| or below.
+// 3) Limited a number of synergy pair above |highThreshold| to |numberOfHighAllowed| or below.
+export function getRandomMA_OneByOne(withVenusian: boolean = true, 
+    numberMARequested: number, 
+    maxSynergyAllowed: number = 6, 
+    totalSynergyAllowed: number = 20,
```
kberg, line 218: Thanks for including this after my comment, but I don't think we need both this and maxSynergyAllowed. No need to overcomplicate it.

kberg: In fact, remove most of this -- focus on one single meaningful improvement and end it there. Less to test. We can make more changes later.

kberg (changes requested):
Definitely needs tests once the general API and structure are settled.

src/MilestoneAwardSelector.ts
vsrisuknimit: Disagree. I believe this is an improvement over simple max synergy for all pairs.

src/Game.ts
vsrisuknimit: renamed

src/MilestoneAwardSelector.ts
kberg: I believe you that it is -- an improvement at the cost of complexity is my concern. I see two ares where this will be a problem.

1. The test cases to ensure correctness.
2. What controls to hand off to the end-user. I was suggesting somewhere (discord?) a slider from 0-9 or 0-50 (those numbers aren't quite perfect but close) and presenting all four of these as options are too much. Someone shouldn't have to know how the innards work to make a smart selection.

none/limited/unlimited is a good start, but given that everyone's aware of the analysis _right now_ I'd rather settle on the end-game first.

server.ts
vsrisuknimit: changed to enum

src/MilestoneAwardSelector.ts
vsrisuknimit: The algorithm is not all complex. It should actually be faster than the previous method which generate a whole set at once and keep rejecting them until one pass.
1. We do need test cases. Not sure what we are testing though. Randomize correctly without duplicate? Randomize correctly with the parameters given?
2. Slider can be implemented on top of this if need be. I think it is better to give user limited options here and then a filter option for fully customizing which milestones and awards to choose. No need for users to play around with sliders and understanding how the algorithm works.

vsrisuknimit:
I want to have Limited selected when Randomizer is on. The value of v-model is set to limited right now. Just not sure how to check the box.
image

src/MilestoneAwardSelector.ts
nwai90: A finite set of predefined choices presented to the user would be best here imo. At the first iteration users don't have to know every detail that happens under the hood to decide whether to randomize or not.

About the test cases, I would pick one or two specific combinations/pairs that are guaranteed to never show up with synergy restriction applied. Given a large enough number of X runs, we should expect:

(1) to never see these combinations if synergy option is selected, and
(2) to see them occurring at least once in the case of unlimited synergy

vsrisuknimit:
If the randomizer requires more than 5 attempts to get a satisfying set of Ms and As that follow the rules, an error is thrown to prevent infinite recursion loop.

Below is a histogram of the number of attempt required to generate a set of Ms and As that follow the rules. [6,20,2,4] for [max, total, numberOfPair, highThreshold].

image

vsrisuknimit:
Tagging @chosta again for help with pre-checking the limited synergy radio button as soon as Random MA checkbox is checked.

vsrisuknimit:
- Added constants
- Added function to verify the synergy rules
- Added tests to MilestonesAwardsSelector.spec.ts.
- Added checks to make sure no infinite loop can happen.

vsrisuknimit:
I believe this is good to go.

Tagging @nwai90 again for review.

src/MilestoneAwardSelector.ts
kberg: > The algorithm is not all complex.  It should actually be faster than the previous method which generate a whole set at once and keep rejecting them until one pass.

I assert that you and I think it is not complex because we've bothered to read the code and figure it out. Don't add additional complexity, please. Gamers aren't going to need them, what's the meaningful extra value? I'm glad they'll be faster though speed isn't a concern.

nwai90 (approved):

kberg:
I chatted with @vsrisuknimit and cleared up my misunderstanding. Looks great. Thank you!

kberg (approved):

samyelias:
> If the randomizer requires more than 5 attempts to get a satisfying set of Ms and As that follow the rules, an error is thrown to prevent infinite recursion loop.

I just noticed this. If you remove ineligible goals as you pick them, you'll never get an infinite loop and the randomizer will run a little faster. This is how I implemented it on my randomizer (I zero out the weights as it picks, since I use weights).

kberg:
No matter what, I think 5 tries is a little too short. It was never that
slow to begin with -- why make such a small count?

>

vsrisuknimit:
> No matter what, I think 5 tries is a little too short. It was never that slow to begin with -- why make such a small count?
> […](#)

because of this: https://github.com/bafolts/terraforming-mars/pull/1637#issuecomment-711411034

but we could increase to 10 if you want.

kberg:
I stand corrected! You have DATA! A comment in the code pointing to that
would be awesome.

>

PR #1636, opened 2020-10-16
Add card names and associated HTML for Ares.

kberg:

PR #1638, opened 2020-10-16
Ares data structures, with modifications to ISpace, ITile and TileType.

kberg:
Depends on #1636 1636


src/CardName.ts
```diff
@@ -457,4 +457,29 @@ export enum CardName {
     TRADE_ADVANCE = "Trade Advance",
     POLITICAL_UPRISING = "Political Uprising",
     BY_ELECTION = "By-Election",
-}
\ No newline at end of file
+
+    // Ares expansion.
+    BIOENGINEERING_ENCLOSURE = "Bioengineering Enclosure",
+    BIOFERTILIZER_FACILITY = "Bio-Fertilizer Facility",
+    BUTTERFLY_EFFECT = "ButterflyEffect",
```
bafolts, line 464: This will want to merge with master branch to use `Butterfly Effect`.

src/ares/AresSpaceBonus.ts
```diff
@@ -0,0 +1,7 @@
+export enum AresSpaceBonus {
+  // Numbering ensures that these bonuses don't conflict with SpaceBonus's indexes.
+  MEGACREDITS = 100,
```
bafolts, line 3: It may be simpler to put these in `SpaceBonus` . It doesn't look like there is any overlap between the types.

src/CardName.ts
kberg: Done.

src/ares/AresSpaceBonus.ts
kberg: There is no overlap between them. I was warned perhaps to not do that, so I went this route. But I would be thrilled to do that. It will probably churn my code a bit, but it's worth it.

bafolts: I would just merge them. It should be simpler for all involved.

bafolts (approved):
Putting every space bonus in `SpaceBonus` should be less to maintain. If we ever add a bonus type to `SpaceBonus` we don't have to check if `AresSpaceBonus` has it or inadvertently risk a collision with `AresSpaceBonus` in the future.

kberg:
Thanks. I'm removing it now, a couple of tests are being difficult, but it won't take long to remove them.

PR #1639, opened 2020-10-16
Half the ares cards -- specifically those that replace other cards in the set.

kberg:
Depends on #1638


src/cards/ares/AresCardHelper.ts
```diff
@@ -0,0 +1,21 @@
+import { ISpace } from "../../ISpace";
+import { SpaceBonus } from "../../SpaceBonus";
+import { TileType } from "../../TileType";
+
+export class AresCardHelper {
+    public static setAdjacencyBonus(space: ISpace, bonusType: SpaceBonus) {
+      space.adjacency = { bonus: [bonusType] };
+      switch(bonusType) {
+      case SpaceBonus.STEEL:
+          space.tile!.tileType = TileType.MINING_STEEL_BONUS;
```
bafolts, line 10: The `space.tile` may be `undefined`. The function that calls `setAdjacencyBonus` is creating the tile which this is assuming exists. Passing `ITile` in as an additional argument will avoid needing to make this assumption.

```
public static setAdjacencyBonus(space: ISpace, tile: ITile, bonusType: SpaceBonus)
```

bafolts (changes requested):
The `AresCardHelper::setAdjacencyBonus` can benefit from typescript with the two comments I left.

src/cards/ares/AresCardHelper.ts
kberg: Good catch, and, that's what I'll do.

kberg:
Question: what would you think about a mixin?

kberg:
> You can use `bonusType: SpaceBonus.STEEL | SpaceBonus.TITANIUM` and avoid the `default` case below.  

Better still: this is very little code. It doesn't need a whole class. Ignore my comment about mixins -- though I'm still curious what you think.

kberg:
> > You can use `bonusType: SpaceBonus.STEEL | SpaceBonus.TITANIUM` and avoid the `default` case below.
> 
> Better still: this is very little code. It doesn't need a whole class. Ignore my comment about mixins -- though I'm still curious what you think.

Nope, I'm wrong and silly. :D

kberg:
Hopefully this is better. Thanks for the feedback!

src/cards/ares/MiningAreaAres.ts
```diff
@@ -0,0 +1,19 @@
...
+import { MiningArea } from "../MiningArea";
+import { AresMiningCards } from "./AresMiningCards";
+
+export class MiningAreaAres extends MiningArea {
+  public name: CardName = CardName.MINING_AREA_ARES;
+
+  protected setAdjacencyBonus(space: ISpace, bonusType: SpaceBonus.STEEL | SpaceBonus.TITANIUM) {
+    space.adjacency = { bonus: [bonusType] };
+      // TODO(kberg): this pattern will repeat. Move something to AresHandler.
+      if (space.tile) {
```
bafolts, line 13: The spacing in this change doesn't line up. fwiw there is `console.assert` available. This could then be:

```
console.assert(space.tile !== undefined, `${space.id} is expected to have a tile.");
```

bafolts:
The `MiningArea` and `MiningRights` duplication was already an annoyance as the cards are mostly identical except a placement restriction difference. Instead of the TODO to come back and consolidate this logic what do you think about?

https://github.com/kberg/terraforming-mars/pull/17

src/cards/ares/MiningAreaAres.ts
kberg: For some reason I thought console.assert disappears at runtime.

kberg:
> The `MiningArea` and `MiningRights` duplication was already an annoyance as the cards are mostly identical except a placement restriction difference. Instead of the TODO to come back and consolidate this logic what do you think about?
> 
> [kberg#17](https://github.com/kberg/terraforming-mars/pull/17)

Better -- I'd rather not have card-name-specificities in MiningCard -- pass them as parameters. `(isAres: boolean, location: ??)` where location is "anywhere" or "next to your tiles"? Fine as is, though.

kberg:
I _think_ this should cut it.

bafolts (approved):

PR #1641, opened 2020-10-17
Fix some cards shown twice in SelectCard component 

### LINKED ISSUE #1640: Duplicate cards appearing on certain floater cards (Atmo Collectors, Jovian Lanterns)
Appears to happen on cards with Floaters on Face that go to any card.  Only a visual bug,  When selecting one of the cards to place the 2 floaters, both are selected at the same time.  Selecting either one adds the Floaters to it appropriately.

Seed: c9b643627efd - to reproduce, Aridor, removes floater colony, places Luna, picks atmo collectors and plays it turn 1 after preludes.

image

I don't know the seed for this one

Happened for Jovian Lanterns.  Have not witnessed this on other floater cards besides this.

Apologies if this isn't the appropriate format to post an Issue.

Lynesth:
Interrupts are now processed when they are displayed so there's no need for cards to add themselves to the selection as they are already in playedCards when the interrupt fires.

Fixes #1640


nwai90:
@Lynesth This looks right to me. Do you think we can also remove the code block in `src/interrupts/SelectResourceCard.ts:26`? It seems that this logic is no longer used. We can probably deprecate `optionalCard` at some point as it is no longer necessary to pass in this argument.

```
// Played card is not into playedCards array yet
if (this.optionalCard !== undefined) {
    resourceCards.push(this.optionalCard);
}
```

Lynesth:
@nwai90 You are most correct. I removed it.
Fixed double quotes in files I modified and reordered imports in Game.ts since I had to remove the ICard one.

kberg:
Looks great. I'm glad this wasn't fixed by removing duplicates from the array.

kberg (approved):

PR #1643, opened 2020-10-17
Fix: Leavitt should trigger Aridor's effect

nwai90:
The Science tag gained needs to trigger effects on all relevant cards in play. Easiest way is probably to add a "dummy" card for it that has nothing but a science tag, with a different `CardType` so that it does not count for Tycoon/Magnate, show up in the game log, etc.

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1309,7 +1309,9 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
           }  
         } 
 
-        if (selectedCard.cardType !== CardType.SPECIAL) this.addPlayedCard(game, selectedCard);
+        if (selectedCard.cardType !== CardType.PROXY) {
+            this.addPlayedCard(game, selectedCard);
+        }
 
         for (const playedCard of this.playedCards) {
             if (playedCard.onCardPlayed !== undefined) {
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1309,7 +1309,7 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
           }  
         } 
 
-        this.addPlayedCard(game, selectedCard);
+        if (selectedCard.cardType !== CardType.SPECIAL) this.addPlayedCard(game, selectedCard);
 
         for (const playedCard of this.playedCards) {
             if (playedCard.onCardPlayed !== undefined) {
--- a/src/cards/promo/PharmacyUnion.ts
+++ b/src/cards/promo/PharmacyUnion.ts
@@ -71,6 +71,8 @@ export class PharmacyUnion implements CorporationCard {
                 availableOptions.options.push(
                     new SelectOption("Turn this card face down and gain 3 TR", 
                     "Gain TR", () => {
+                        if (this.isDisabled) return undefined;
+                        
                         this.isDisabled = true;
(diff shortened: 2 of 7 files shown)


nwai90:
@chosta What we were discussing. I found an additional bug with an edge case of Leavitt + Research Colony (which has a science tag). In this scenario PU should only get to flip and gain 3 TR once even though there are two interrupts in the queue.

src/Player.ts
```diff
@@ -1309,7 +1309,7 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
           }  
         } 
 
-        this.addPlayedCard(game, selectedCard);
+        if (selectedCard.cardType !== CardType.SPECIAL) this.addPlayedCard(game, selectedCard);
```
kberg, line 1312: Instead of SPECIAL how about PROXY? Also, since this.addPlayedCard is the important part of this and this special condition is not, can you make this.addPlayedCard more visible? I think enclosing the statement in braces and sticking it on a newline ought to do.

src/cards/promo/PharmacyUnion.ts
```diff
@@ -71,6 +71,8 @@ export class PharmacyUnion implements CorporationCard {
                 availableOptions.options.push(
                     new SelectOption("Turn this card face down and gain 3 TR", 
                     "Gain TR", () => {
+                        if (this.isDisabled) return undefined;
```
Lynesth, line 74: Shouldn't that be at the top of the function, just before or after 
`        if (scienceTags <= 0) return;`
?

nwai90: @Lynesth That wouldn't work (tried it initially haha). The interrupts seem to know nothing about the current game state.

Lynesth: Of course! Sent a suggestion over Discord to make it work properly while waiting for the next big change to interrupts.

src/Player.ts
nwai90: `PROXY` is a much better name. Updated!

src/cards/promo/PharmacyUnion.ts
nwai90: Done! Thanks for your help on this one

Lynesth:
👍

src/Player.ts
chosta: Did you also check if the CardType enum was used incrementally anywhere? What I learned the hard way is that enums like `Tags` should not be changed because of constrains. @kberg Is there a way in TS to point out that a data structure should not change like `final` or something?

nwai90: @chosta It should be fine - all logic specifically references one or more `CardTypes`. There is no code that iterates through all values in the `CardType` enum.

kberg (approved):

kberg:
MB my approval only goes as far as the general card structure, and the use of the proxy card. Could someone else focus on the colony behavior itself?

kberg:
@alrusdi can you look at this one for the Colonies behavior?

alrusdi:
@kberg unfortunately i'm not into colonies :-( Let's merge it and see what happens?

nwai90:
@alrusdi It's fine to merge. The colony behaviour is unchanged and has been tested quite a bit

PR #1648, opened 2020-10-18
fixing utopia ui, moving in-line styles to css classes

vsrisuknimit:
image

@nwai90


nwai90:
@vsrisuknimit There are two of the expansion icon (orange circle with triangle). It is probably incorrect; can we remove the bottom one?

vsrisuknimit:
@nwai90 all corps have duplicated icon. It came from the project cards.

image

nwai90 (approved):

Issue #1650, opened 2020-10-18
Bug  : turmoil +1 influence bonus 


jijiqw:
![image](https://user-images.githubusercontent.com/16096558/96358555-ff10d280-113a-11eb-8bbd-1151826b9fb1.png)
A player only get +1 influence bonus if he plays `incite` an `Event Analysts` both

nwai90:
Fixed and merged

PR #1651, opened 2020-10-18
Fix: turmoil.addInfluence should set new influence value

nwai90:
Ref: https://github.com/bafolts/terraforming-mars/issues/1650


kberg (approved):

kberg:
One warning: I don't have the Turmoil expertise to know whether this is right from a gaming perspective, but code LGTM

PR #1652, opened 2020-10-18
Remove AresSpaceBonus

kberg:
Depends on #1639


bafolts:
This change can also remove the file:

https://github.com/kberg/terraforming-mars/blob/am4/src/ares/AresSpaceBonus.ts

kberg:
> This change can also remove the file:
> 
> https://github.com/kberg/terraforming-mars/blob/am4/src/ares/AresSpaceBonus.ts

Isn't it? How did that

> This change can also remove the file:
> 
> https://github.com/kberg/terraforming-mars/blob/am4/src/ares/AresSpaceBonus.ts

Ugh, it got re-added during a merge.

bafolts (approved):

PR #1653, opened 2020-10-18
Ares images, icons, css.

kberg:
Does not depend on any code currently under review (though perhaps some components not yet released.)


chosta (approved):
Seems ok to me. Just make sure that the player cubes are rendered properly everywhere. I remember vaguely that colonies and cubes had an interesting css + javascript interaction.

bafolts (approved):

kberg:
Thanks both. @chosta I removed the change to board.less -- it's not worth doing in this PR, and I don't have the confidence to verify such a change.

PR #1654, opened 2020-10-18
Ares merge 5: Add the Ares milestones and awards -- first minor addition to Game.ts

kberg:
Depends on #1652


src/awards/Entrepeneur.ts
```diff
@@ -0,0 +1,17 @@
...
+
+// TODO(kberg): write a test.
+export class Entrepeneur implements IAward {
+    public name: string = "Entrepeneur";
+    public description: string = "Most tiles that grant adjacency bonuses"
+    public getScore(player: Player, game: Game): number {
+        return game.board.spaces
+          .filter((space) => (
+              space.player !== undefined
+              && space.player === player
```
bafolts, line 13: This may need to also check that `space.tile !== undefined` because of `LandClaim`.

bafolts (approved):

src/awards/Entrepeneur.ts
kberg: It won't at the moment because, at the moment, untiled spaces don't grant adjacency bonuses.

But this possibly underscores that the adjacency bonus might be better off tied to the tile. I'd like that to be a separate change, though -- once all of this is in.

PR #1658, opened 2020-10-18
Clarify and standardize raw_string

kberg:


kberg:
Is it possible to rename it to UNTRANSLATED / untranslated? The comment in LogMessageDataType is probably enough.

alrusdi (approved):

PR #1659, opened 2020-10-18
Ares merge 6: Adding some Ares-specific cards.

kberg:
Requires #1654


src/Game.ts
```diff
@@ -59,6 +59,7 @@ import { Tags } from "./cards/Tags";
 import { TileType } from "./TileType";
 import { Turmoil } from "./turmoil/Turmoil";
 import { getRandomMilestonesAndAwards } from "./MilestoneAwardSelector";
+import { IAresData } from "./ares/IAresData";
```
bafolts, line 62: Unrelated to this change but I think you could use `PlayerId` as the type for `IMilestoneCount`. This could help with documenting as I believe that `id` is the player's id.

```
export interface IMilestoneCount {
    PlayerId: string;
    count: number;
}
```

src/ares/AresHandler.ts
```diff
@@ -0,0 +1,33 @@
...
+    SEVERE
+};
+
+export class AresHandler {
+    public static hasHazardTile(space: ISpace): boolean {
+        return AresHandler.hazardSeverity(space) !== HazardSeverity.NONE;
+    }
+
+    public static hazardSeverity(space: ISpace): HazardSeverity {
+        const type = space.tile?.tileType;
```
bafolts, line 18: If `space.tile` is `undefined` this function will return `HazardSererity.NONE`. I am not familiar with the ares rules. As long as `HazardSeverity.NONE` of none makes sense if a space doesn't have a tile then this is fine if intended.

src/cards/ares/BioengineeringEnclosure.ts
```diff
@@ -0,0 +1,45 @@
...
+export class BioengineeringEnclosure implements IProjectCard, IActionCard, IResourceCard {
+  public cost: number = 7;
+  public tags: Array<Tags> = [Tags.ANIMAL];
+  public cardType: CardType = CardType.ACTIVE;
+  public name: CardName = CardName.BIOENGINEERING_ENCLOSURE;
+  public resourceType = ResourceType.ANIMAL;
+  public resourceCount = 0;
+
+  public canPlay(player: Player, _game: Game): boolean {
+    return player.getTagCount(Tags.SCIENCE) >= 1;
```
bafolts, line 20: Will this need another todo? This card also claims that "OTHERS MAY NOT REMOVE ANIMALS FROM THIS CARD". That will need to be enforced.

src/interrupts/SelectResourceCard.ts
```diff
@@ -16,15 +17,19 @@ export class SelectResourceCard implements PlayerInterrupt {
...
 
     public generatePlayerInput() {
-        let resourceCards = this.player.getResourceCards(this.resourceType);
+        let resourceCards: Array<ICard> = this.player.getResourceCards(this.resourceType);
 
         if (this.restrictedTag !== undefined) {
             resourceCards = resourceCards.filter(card => card.tags.indexOf(this.restrictedTag!) !== -1);
         }
+        if (this.invalidCards.length > 0) {
+            resourceCards = resourceCards.filter(card => this.invalidCards.indexOf(card.name) === -1);
```
bafolts, line 31: fwiw this can use `this.invalidCards.includes(card.name) === false`.

bafolts (approved):
So cool how cards fall into place! It may be coming up in a PR but at some point it needs to be enforced that others can't remove animals from `BioengineeringEnclosure`.

src/Game.ts
kberg: Thanks. Adding TODO to Networker to do this, once I write a test.

src/ares/AresHandler.ts
kberg: That's exactly correct. Hazards are very specific tiles.

src/cards/ares/BioengineeringEnclosure.ts
kberg: Good spot.

Here's code in `Predator.ts` that I haven't sent out yet.

```
    private getPossibleTargetCards(player: Player, game: Game): Array<ICard> {
        const result: Array<ICard> = [];
        game.getPlayers().forEach((p) => {
            if (p.hasProtectedHabitats() && player.id !== p.id) return;

            // TODO(kberg): the list of cards that can't have animals removed is now 2.
            // This logic is duplicated in Virus.ts.
            result.push(...p.getCardsWithResources().filter(card => card.resourceType === ResourceType.ANIMAL 
                                                                && card.name !== CardName.PETS
                                                                && card.name !== CardName.BIOENGINEERING_ENCLOSURE
                                                                && card.name !== this.name));
        });
        return result;
    }
```

src/interrupts/SelectResourceCard.ts
kberg: Thank you.

kberg: Actually I'd prefer to leave it because a) the same line is above, and b) this class has no tests.

kberg:
> So cool how cards fall into place! It may be coming up in a PR but at some point it needs to be enforced that others can't remove animals from `BioengineeringEnclosure`.

You know what? I'll add them into this PR. Just means another pass.

PR #1680, opened 2020-10-19
add space between gen and number

### LINKED ISSUE #1673: Missing whitespace after generation number
whitespace

Simple thing to fix -- should be a space between GENERATION and 9.

vsrisuknimit:
This can close #1673 .

image

### DIFF
--- a/src/components/PlayerHome.ts
+++ b/src/components/PlayerHome.ts
@@ -142,7 +142,7 @@ export const PlayerHome = Vue.component("player-home", {
                 <players-overview class="player_home_block player_home_block--players nofloat:" :player="player" v-trim-whitespace />
 
                 <div class="player_home_block player_home_block--log player_home_block--hide_log nofloat">
-                    <dynamic-title v-if="player.players.length > 1" title="Game log" :color="player.color" :withAdditional="true" :additional="'generation' + player.generation" />
+                    <dynamic-title v-if="player.players.length > 1" title="Game log" :color="player.color" :withAdditional="true" :additional="'generation ' + player.generation" />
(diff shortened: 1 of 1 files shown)


src/components/PlayerHome.ts
```diff
@@ -142,7 +142,7 @@ export const PlayerHome = Vue.component("player-home", {
                 <players-overview class="player_home_block player_home_block--players nofloat:" :player="player" v-trim-whitespace />
 
                 <div class="player_home_block player_home_block--log player_home_block--hide_log nofloat">
-                    <dynamic-title v-if="player.players.length > 1" title="Game log" :color="player.color" :withAdditional="true" :additional="'generation' + player.generation" />
+                    <dynamic-title v-if="player.players.length > 1" title="Game log" :color="player.color" :withAdditional="true" :additional="'generation ' + player.generation" />
```
alrusdi, line 145: Can you please make the "generation" word translatable? It may require to move inline code to separate function call like:
`:additional=getGameLogTitle()`

vsrisuknimit: @alrusdi Is there any reason why it need v-if/v-else there to begin with? getGenerationText() can already have player count.

chosta:
Yes, there is a reason that I don't remember atm. Just keep it as it is. It
should be refactored sooner rather than  later

kberg (approved):

src/components/PlayerHome.ts
kberg: Do that in another PR OK?

kberg (approved):

PR #1687, opened 2020-10-20
Am8: Implement Adjacency bonuses in the game.

kberg:


src/ares/AresHandler.ts
```diff
@@ -10,6 +18,111 @@ export enum HazardSeverity {
...
+        this.assertHasAres(game);
+
+        // let incrementMilestone = false;
+
+        game.board.getAdjacentSpaces(space).forEach((adjacentSpace) => {
+            if (this.earnAdacencyBonus(game, adjacentSpace, player)) {
+        //     incrementMilestone = true;
+            }
+        });
+        // if (incrementMilestone) {
```
bafolts, line 38: Probably fine to remove this commented out code. Is it needed?

kberg: It will be when I bring in milestones and awards  -- so much of the code is commingled that sometimes I don't want to have to figure out exactly where to put things back as I incorporate them, so I hope you don't mind if this stays where it is.

src/ares/AresHandler.ts
```diff
@@ -10,6 +18,111 @@ export enum HazardSeverity {
...
+                });
+            }
+        };
+    
+        const bonuses = new Multiset<SpaceBonus>();
+
+        adjacentSpace.adjacency.bonus.forEach(bonus => {
+            bonuses.add(bonus);
+            switch(bonus) {
+            case SpaceBonus.ANIMAL:
```
Lynesth, line 87: Are we indenting those or not ? I'm not sure.
Also, isn't it strange that this is out of the `game.grantSpaceBonus` method and have it called on the default case ?
Shouldn't you simply merge those 2 switch/cases ?

src/ares/AresHandler.ts
```diff
@@ -10,6 +18,111 @@ export enum HazardSeverity {
...
+                break;
+            }
+        });
+
+        const bonusText = bonuses.entries().map((elem) => `${elem[1]} ${bonusAsString(elem[0])}`).join(", ");
+        const tileText = tileTypeAsString(adjacentSpace.tile?.tileType);
+        game.log("${0} gains ${1} for placing next to ${2}", b => b.player(player).string(bonusText).string(tileText));
+
+        let ownerBonus = 1;
+        if (adjacentSpace.player !== undefined) {
```
Lynesth, line 114: Isn't that always true since you throw an error line 57 if it's undefined ?

src/ares/AresHandler.ts
```diff
@@ -10,6 +18,111 @@ export enum HazardSeverity {
...
+            if (adjacentSpace.player.playedCards.find(card => card.name === CardName.MARKETING_EXPERTS)) {
```
Lynesth, line 115: can be shortened to
```ts
if (adjacentSpace.player.cardIsInEffect(CardName.MARKETING_EXPERTS)) {`
```

src/ares/AresHandler.ts
bafolts: Putting in the ares code behind the expansion flag makes sense to me.

src/utils/Multiset.ts
```diff
@@ -0,0 +1,28 @@
...
+        const val = (this.get(key) || 0) + count;
+        this.map.set(key, val);
+        return val;
+    }
+
+    get(key: T): number | undefined {
+        return this.map.get(key);
+    }
+
+    entries(): Array<[T, number]> {
```
bafolts, line 17: I believe this can be simplified to:

```typescript
entries(): Array<[T, number]> {
   return Array.from(this.map);
}
```

src/ares/AresHandler.ts
bafolts: There is a linting rule that we don't have enabled that checks for unnecessary conditionals.

https://palantir.github.io/tslint/rules/strict-boolean-expressions/

I don't know if this was needed here or not but the linting rule is helpful for spotting these as they show up over time as properties become optional or not.

bafolts (approved):

src/ares/AresHandler.ts
kberg: Add that! It sounds great.

src/ares/AresHandler.ts
kberg: I can't parse your sentence in the context of this discussion.

src/ares/AresHandler.ts
kberg: We are indenting them -- there's a budding discussion about a common formatter.

kberg: It's not in the base one because originally the Ares-specific space bonuses weren't in Ares, and I was asked to keep as much code separate as possible. I'm still on board with that idea. I mean, unless someone wants to create a board with animal space bonuses?

src/ares/AresHandler.ts
kberg: Removed the conditional -- good catch.

src/ares/AresHandler.ts
Lynesth: As I see it, it would just add a few cases to the `grantSpaceBonus` method.
So if someone then creates a new board with animals as space bonus, it should work out of the box.
You could do it the same way you did in `SpaceBonus.ts`: add a little comment above the cases that are specific to Ares.

Anway, I don't really mind either way.

src/utils/Multiset.ts
kberg: Fixed. Added a Multiset test to this PR.

src/ares/AresHandler.ts
kberg: Good fix. I tested this against the code that uses Marketing Experts to confirm it does what I expect.

src/ares/AresHandler.ts
bafolts: All of this code will only be executed if `aresExpansion` were set to true in a game option. That is why I don't mind this staying where it is. The code being merged is helpful versus all coming in at once.

PR #1690, opened 2020-10-20
Am9: Implement non-hazard Ares costs.

kberg:


src/Game.ts
```diff
@@ -1409,8 +1414,17 @@ export class Game implements ILoadable<SerializedGame, Game> {
         );
       }
 
+      if (this.gameOptions.aresExtension) {
+        AresHandler.assertCanPay(this, player, space);
```
Lynesth, line 1418: You are calling it again in `payAdjacencyAndHazardCosts` just below, is it planned to remove one of the two ?

src/ares/AresHandler.ts
```diff
@@ -10,6 +25,111 @@ export enum HazardSeverity {
...
+        game.log("${0} gains ${1} for placing next to ${2}", b => b.player(player).string(bonusText).string(tileText));
+
+        let ownerBonus = 1;
+        if (adjacentSpace.player !== undefined) {
+            if (adjacentSpace.player.playedCards.find(card => card.name === CardName.MARKETING_EXPERTS)) {
+                ownerBonus = 2;
+            };
+            
+            adjacentSpace.player.megaCredits += ownerBonus;
+            game.log("${0} gains ${1} M€ for a tile placed next to ${2}", b => b.player(adjacentSpace.player!).number(ownerBonus).string(tileText));
```
bafolts, line 127: I _think_ this is listed in the wonderful "Effective TypeScript" book but `adjacentSpace.player` being `undefined` I assume is giving trouble here since typescript doesn't know _when_ these callbacks are called. Typescript has to assume the callback can be called whenever. Since `adjacentSpace` is provided as a closure argument it could change by the time the callback is called. I would create a `const` and store `adjacentSpace.player` to avoid having to constantly check it's state or assuming it's state.


```typescript
const adjacentPlayer = adjacentSpace.player;
if (adjacentPlayer === undefined) {
   throw new Error("must have adjacent player!");
}

// now adjacentPlayer can be used even inside callbacks and typescript knows it can't change!
```

bafolts (approved):

src/Game.ts
kberg: I wasn't planning it -- though the method could do with a rename. It both computes the cost, and also fails if the cost is too expensive. So it's useful both ways. I'll improve the structure though.

src/ares/AresHandler.ts
kberg: I like it.

PR #1698, opened 2020-10-21
Vue type checking for GameEnd component.

bafolts:
Starting to put in rudimentary type checking. Starting with the `GameEnd` component. Have to create a typescript file to run through the compiler. This change includes fixes needed for `PlayerModel` and for the template to pass the new build process.


.gitignore
```diff
@@ -17,3 +17,5 @@ pid
 
 # Local Packages
 node_modules/
+
+src/components/*Vue.ts
```
bafolts, line 21: Ignoring the generated typescript file. Want to put this in `dist/` so it is out of the way.

src/components/GameEnd.ts
```diff
@@ -95,7 +99,7 @@ export const GameEnd = Vue.component("game-end", {
                         <tbody>
                             <tr v-for="p in getSortedPlayers()">
                                 <td><log-player :class="getEndGamePlayerColorClass(p)">{{ p.name }}</log-player></td>
-                                <td v-i18n>{{ p.corporationCard.name }}</td>
+                                <td v-i18n>{{ p.corporationCard !== undefined ? p.corporationCard.name : "" }}</td>
```
bafolts, line 102: This threw a typescript compilation error that `corporationCard` can be `undefined` :-)

src/models/PlayerModel.ts
```diff
@@ -5,16 +5,25 @@ import { VictoryPointsBreakdown } from "../VictoryPointsBreakdown";
...
+import { FundedAwardModel } from "./FundedAwardModel";
+import { Phase } from "../Phase";
+import { PlayerInputModel } from "./PlayerInputModel";
+import { SpaceModel } from "./SpaceModel";
 
 export interface PlayerModel {
+    awards: Array<FundedAwardModel>;
     boardName: string;
-    corporationCard: CardModel;
+    corporationCard: CardModel | undefined;
```
bafolts, line 18: These properties were being included on `server.ts` but were not typed.

compile-vue-templates.js
```diff
@@ -0,0 +1,76 @@
+
+const templates = [{
+    file: "./src/components/GameEnd",
+    scope: [
```
bafolts, line 4: Plan to clean up the implementation below. The script needs to know these scope variables to apply to the generated template.

compile-vue-templates.js
```diff
@@ -0,0 +1,76 @@
...
+lines.unshift("declare const _v: any;");
+lines.unshift("declare function _l<T>(arg1: Array<T>, arg2: (item2: T) => any): any;");
+
+file = lines.join("\n");
+
+templateString = templateString.trim();
+templateString = templateString.substring("template: ".length);
+
+const foo = eval(templateString);
+let result = compiler.compile(foo).render;
```
bafolts, line 58: This will also ensure that our templates compile without error and throw.

src/components/GameEnd.ts
```diff
@@ -15,6 +21,19 @@ export const GameEnd = Vue.component("game-end", {
         "Button": Button
     },
     methods: {
+        fillMeIn: function () {
```
chosta: How did we come to this method? 
Is it what the vue-type-checker generates as code on compile?

bafolts: Ah, this was a mistake as I was ironing things out. This `fillMeIn` or some equivalent is what is generated on `GameEndVue.ts` class to run the type checking. I had manually entered it onto `GameEnd.ts` while working on this tool. The generated class has this function (or any function) where the compiled template is placed so typescript can run on the class.

kberg:
This is still a work in progress?

bafolts:
Yes. The #1707 took most of my non day job time. Want to clean this up. Plan to have it ready by Thursday or Friday. I will add one component at a time. Will start with `GameEnd` component.

tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
     "compilerOptions": {
+        "declaration": true,
```
bafolts, line 3: The type checking tool needs these declaration files so it can run from `/dist`.

bafolts:
@kberg and @chosta this should be ready now. This will start to give us sanity that our templates are going to work on the client. If while the vue template compiler is running it spots any errors or tips the new checker fails the build. The trick used to provide a mock `vue` to `require` from `node` we should be able to re-use to start writing unit tests on the `methods` provided with our components. This turned out smoother than I thought it would! After this I will add the remaining components one at a time.

alrusdi (approved):

chosta:
> @kberg and @chosta this should be ready now. This will start to give us sanity that our templates are going to work on the client. If while the vue template compiler is running it spots any errors or tips the new checker fails the build. The trick used to provide a mock `vue` to `require` from `node` we should be able to re-use to start writing unit tests on the `methods` provided with our components. This turned out smoother than I thought it would! After this I will add the remaining components one at a time.

Great! I'll check how it's done and try to implement it in my ongoing tasks too.

PR #1700, opened 2020-10-21
Am12: Add Ares global parameter effects.

kberg:


src/ares/AresHandler.ts
```diff
@@ -46,13 +46,15 @@ export class AresHandler {
         };
     }
 
-    public static assertHasAres(game: Game) {
+    public static assertHasAres(game: Game) : boolean {
         console.assert(game.gameOptions.aresExtension, "Assertion failure: game.gameOptions.aresExtension is not true");
         console.assert(game.aresData !== undefined, "Assertion failure: game.aresData is undefined");
+        // I wish console.assert returned a boolean value.
```
bafolts, line 52: I still think this function is cool either way! Shows we assume it exists here and we will see logs when it doesn't.

bafolts: Every function in `AresHandler` is most likely going to need to run this check. I think we will be better off if we use a callback function that we call if these conditions are true. Then at the top of every function in this class we would call.

```
public static use(game: Game, cb: (aresData: IAresData) => void): void {
    if (game.gameOptions.aresExtension === false || game.aresData === undefined) {
        console.warn("Ares is not enabled");
        return;
    }
    cb(game.aresData);
}

public static doSomething(game: Game): void {
    AresHandler.use(game, (aresData) => {
        // now aresData has to be defined
    });
}
```

src/ares/AresHandler.ts
```diff
@@ -310,6 +335,67 @@ function randomlyPlaceHazard(game: Game, tileType: TileType, direction: 1 | -1)
...
+    }
+    if (testValue >= constraint.threshold) {
+        cb();
+        constraint.available = false;
+    }
+}
+
+function testToRemoveDustStorms(game: Game, player: Player) {
+    testConstraint(
+        game.aresData!.hazardData.removeDustStormsOceanCount,
```
bafolts, line 361: As you have done the work to know `game.aresData` is defined prior to calling this I would pass in `aresData` as an argument. At that point no `!` is needed.

bafolts (approved):
I like how ares is separated into `AresHandler`. The `!` is being used again which is always risky. I think we should introduce some form of callback that gives us the `aresData` we know is defined and pass that around along with `game` to the internal functions of `AresHandler`.

src/ares/AresHandler.ts
kberg: The compiler needs the assurance, unfortunately.

What would be wrong with an assert that throws an error instead of logging it? And only under those conditions, use !? Probably not, but I thought I'd ask.

kberg:
> I like how ares is separated into `AresHandler`. The `!` is being used again which is always risky. I think we should introduce some form of callback that gives us the `aresData` we know is defined and pass that around along with `game` to the internal functions of `AresHandler`.

I'll try your callback idea.

src/ares/AresHandler.ts
bafolts: Throwing an error makes sense. As we saw yesterday with the outage if we don't end up catching the error it takes down the node process. We _should_ be catching these errors when thrown through the server though. Can probably look into a better handler on uncaught errors.

kberg:
No question we need a global catch -- and possibly other catches at
tactically appropriate levels.

In the meantime I'm going to proceed with this weird new idea (I like the
callback) and avoid !.

Submitting as-is (you gave approval) but working on a good replacement now.

>

PR #1710, opened 2020-10-22
Fix: Remove circular dependency from Playwrights

nwai90:


bafolts (approved):

bafolts:
Well that is simpler than #1711 . Still should be helpful to not have that self-executing code in the module. This solved the underlying problem though and community cards should be found again.

nwai90:
Agree. @Lynesth spotted this issue in her PR as well. We should be able to do `player.getCardCost` even on a played card and thus avoid the need to call `getProjectCardByName`

kberg:
WOW.

Lynesth:
Yep, did mention it in my PR as the test failed.

@bafolts I actually used this to find the CD: https://www.npmjs.com/package/detect-circular-deps
It's maybe not the best tool for the job (and I couldn't get its glob matching to work) but that's the only one I found that actually tells you where your CDs create problems and not just that you have some CDs. You might want to take a look at that.

Here's what I called:
```bash
find dist/ -name '*.js' -print0 | xargs -0 npx detect-circular-deps | grep -A 1 problem
```

bafolts:
I will check this out. That problem was a first for me and hard to track down. A build step preventing it would be ideal.

Lynesth:
Yeah, I spent a few hours tracking this one down as well -_-

Issue #1707, opened 2020-10-21
Heroku down, application error, last merge was 2 hours ago


vsrisuknimit:
Heroku down, application error, last merge was 2 hours ago

@chosta @bafolts

bafolts:
```
2020-10-21T19:23:01.980140+00:00 app[web.1]: /app/dist/src/Dealer.js:27
2020-10-21T19:23:01.980199+00:00 app[web.1]: found = deck.projectCards.findByCardName(cardName);
2020-10-21T19:23:01.980200+00:00 app[web.1]: ^
2020-10-21T19:23:01.980201+00:00 app[web.1]: 
2020-10-21T19:23:01.980201+00:00 app[web.1]: TypeError: Cannot read property 'projectCards' of undefined
2020-10-21T19:23:01.980202+00:00 app[web.1]: at /app/dist/src/Dealer.js:27:22
2020-10-21T19:23:01.980203+00:00 app[web.1]: at Array.forEach (<anonymous>)
2020-10-21T19:23:01.980203+00:00 app[web.1]: at getProjectCardByName (/app/dist/src/Dealer.js:23:19)
2020-10-21T19:23:01.980203+00:00 app[web.1]: at /app/dist/src/Dealer.js:157:20
2020-10-21T19:23:01.980204+00:00 app[web.1]: at Array.map (<anonymous>)
2020-10-21T19:23:01.980204+00:00 app[web.1]: at Dealer.loadFromJSON (/app/dist/src/Dealer.js:156:28)
2020-10-21T19:23:01.980204+00:00 app[web.1]: at Game.loadFromJSON (/app/dist/src/Game.js:1363:30)
2020-10-21T19:23:01.980205+00:00 app[web.1]: at Query.<anonymous> (/app/dist/src/database/PostgreSQL.js:85:18)
2020-10-21T19:23:01.980205+00:00 app[web.1]: at Query.handleReadyForQuery (/app/node_modules/pg/lib/query.js:129:12)
2020-10-21T19:23:01.980206+00:00 app[web.1]: at Connection.<anonymous> (/app/node_modules/pg/lib/client.js:266:19)
```

kberg:
Grr. This was a problem that danmir pointed out a couple of days ago with
his local db but I hadn't taken time to debug. I have his DB. This could be
related to his most recent PRs? Perhaps?

bafolts:
[...]

PR #1711, opened 2020-10-22
create singleton to get decks

bafolts:
Proper fix for #1707 . Self executing javascript complicates module loading. This moves the self executing `decks` into a `CardFinder` class. Unit tests added to find a prelude card and a project card. I also added a warning for when cards are not found. On older games seeing:

```
card not found Magnetic Field Generators Promo
card not found Great Dam Promo
card not found Deimos Down Promo
```

which I believe we expect from the suffix change to `:promo`.

### DIFF
--- a/src/CardFinder.ts
+++ b/src/CardFinder.ts
@@ -1,0 +1,52 @@
+import { ICardFactory } from "./cards/ICardFactory";
+import { IProjectCard } from "./cards/IProjectCard";
+import { CardManifest } from "./cards/CardManifest";
+import { COLONIES_CARD_MANIFEST } from "./cards/colonies/ColoniesCardManifest";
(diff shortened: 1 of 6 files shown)


src/cards/community/CommunityCardManifest.ts
```diff
@@ -3,7 +3,7 @@ import { GameModule } from "../../GameModule";
 import { CardManifest } from "../CardManifest";
 import { AgricolaInc } from "./AgricolaInc";
 import { Incite } from "./Incite";
-import { Playwrights } from "./Playwrights";
+// import { Playwrights } from "./Playwrights";
```
bafolts, line 6: grrr, have to add this back with fix

kberg (approved):
Thanks, I wondered about the singletonness of it. I am sorry it didn't wind up the way I expected.

PR #1712, opened 2020-10-22
Am13: Marketing Experts, Geological Survey, Ecological Survey. Add some hazard tests

kberg:


src/Game.ts
```diff
@@ -1510,7 +1517,7 @@ export class Game implements ILoadable<SerializedGame, Game> {
         space.bonus = [];
 
         // Must occur after all other onTilePlaced operations.
-        // AresHandler.afterTilePlacement(this, player, startingResources);
```
bafolts, line 1513: Remove before flight. Seems to be duplicated line commented out.

bafolts (approved):

src/Game.ts
kberg: I don't follow what you mean -- it's uncommented here, so any issue?

PR #1713, opened 2020-10-22
Am14:  Trigger milestones and awards.

kberg:


src/Game.ts
```diff
@@ -386,34 +386,28 @@ export class Game implements ILoadable<SerializedGame, Game> {
     }
 
     // Function to construct the board and milestones/awards list
-    public boardConstructor(boardName: BoardName, randomMA: RandomMAOptionType, hasVenus: boolean): Board {
-      const requiredQty = 5;
-      if (boardName === BoardName.ELYSIUM) {
+    public boardConstructor(boardName: BoardName, randomMA: RandomMAOptionType, hasVenus: boolean, hasAres: boolean): Board {
+      const chooseMilestonesAndAwards = function(game: Game, milestones: Array<IMilestone>, awards: Array<IAward>) {
```
bafolts, line 390: I would use a fat arrow function if possible. The fat arrow binds the this argument. I will enable another lint rule with PR to fix the places we aren't using it.

https://eslint.org/docs/rules/prefer-arrow-callback

bafolts (approved):

src/Game.ts
kberg: Will do prior to submit - thank you!

kberg: Actually, I need some help here -- can you jump in and tweak this one? I'll watch and learn.

kberg: (Committing in the meantime.)

PR #1715, opened 2020-10-22
Fix: SolarFarm test and playability check

nwai90:
- Solar Farm should not be playable if there is no spot to place tile (mandatory effect)
- `space` and `citySpace` in the test are not always equivalent if `space` happens to have a hazard tile on it


kberg (approved):
Thanks. I was thinking about going to take a shot at this today. I'm grateful.

PR #1719, opened 2020-10-22
Fix: External links on start screen should open in new tab

nwai90:


alrusdi:
Isn't it bad practice especially for mobile devices?

nwai90:
Not sure about that. These links aren't part of the game itself though and should follow same behaviour as "Join us on Discord" imo

alrusdi (approved):

alrusdi:
> Not sure about that. These links aren't part of the game itself though and should follow same behaviour as "Join us on Discord" imo

Ok, let's try it

PR #1728, opened 2020-10-23
Remove all uses of aresData! with a callback.

kberg:


src/cards/ares/DesperateMeasures.ts
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public canPlay(_player: Player, game: Game): boolean {
+      // You can't play desperate measures if there isn't a hazard marker in play.
+      return this.getHazardTiles(game).length > 0;
+    }
+
+    public play(player: Player, game: Game) {
+      return new SelectSpace("Select a hazard space to protect", this.getHazardTiles(game), (space: ISpace) => {
+        space.tile!.protectedHazard = true;
```
bafolts, line 29: Unrelated to the `aresData!` change but this is another `!` usage. May need to add a linting rule for this! My concern is more that this gets copied and pasted somewhere else in the code where we aren't certain the space will have a tile or some refactor introduces a situation where this can be `undefined`. If the space doesn't have a tile could throw an error. If a space doesn't have a tile this will throw `can't set `undefined` to true` or something similar. Seeing a manual error that there was no tile when expected would help debugging.

src/components/Ares.ts
```diff
@@ -0,0 +1,25 @@
+import Vue from "vue";
+
+export const Ares = Vue.component("ares", {
+    props: ["aresData"],
```
bafolts, line 4: To provide typing for this property the linting rule will require soon:

```
props: {
    aresData: {
        type: Object as () => IAresData
    }
}
```

src/cards/ares/DesperateMeasures.ts
kberg: That's fixed later. :D

src/components/Ares.ts
kberg: Hooray! This file is deleted in the upcoming `ares-viz` branch, listed as "am18"

src/components/SelectProductionToLose.ts
```diff
@@ -0,0 +1,158 @@
...
+        canDeductEnergy: function() {
+            return this.playerinput.payProduction.units.energy > 0;
+        },
+        canDeductHeat: function() {
+            return this.playerinput.payProduction.units.heat > 0;
+        },
+        hasWarning: function () {
+            return this.$data.warning !== undefined;
+        },
+        delta: function(type: string, direction: number) {
```
bafolts, line 54: You _may_ be able to use `type: keyof SelectProductionToLoseModel` here instead of `string`.

src/cards/ares/DesperateMeasures.ts
kberg: Also please add the linting rule.

src/inputs/SelectProductionToLose.ts
```diff
@@ -0,0 +1,19 @@
...
+    public inputType = PlayerInputTypes.SELECT_PRODUCTION_TO_LOSE;
+  
+    constructor(
+        public title: string,
+        public unitsToLose: number,
+        public player: Player,
+        public cb: (units: IProductionUnits) => PlayerInput | undefined,
+        public buttonLabel: string = "Save",
+        ) {
+            this.buttonLabel = buttonLabel;
```
bafolts, line 16: fwiw since `public buttonLabel` was used this line is unnecessary. If you look at the generated javascript in `/dist/src/inputs/SelectProductionToLose.js` you will see this duplicated.

tests/ares/AresHandler.spec.ts
```diff
@@ -88,7 +88,12 @@ describe("AresHandler", function () {
 
         const adjacentSpace = game.board.getAdjacentSpaces(firstSpace)[0];
         game.addTile(player, adjacentSpace.spaceType, adjacentSpace, {tileType: TileType.GREENERY});
+<<<<<<< HEAD
```
bafolts, line 91: Remove before flight.

src/cards/ares/DesperateMeasures.ts
kberg: Actually  I see now that this could just be ?.

kberg:
There's something wrong with this PR. I made a merge mess. Let's leave this one alone for now, and stick to the ordered PRs, sorry about that.

bafolts (approved):
Minor suggestions which aren't important. Overall looks great. I plan to eventually add each component to the vue component type checker script. That will require `props` use the object for types. The components look like they will work though.

src/cards/ares/DesperateMeasures.ts
bafolts: You are just that excited to be writing typescript! Can I blame you? What a great language! Can't help but use exclamation mark!

src/components/Ares.ts
kberg: \`~`

src/cards/ares/DesperateMeasures.ts
kberg: \`~`

kberg:
This PR is outdated and the merges are not pretty. Closing.

PR #1729, opened 2020-10-23
A much simpler approach to managing space bonuses when covering tiles.

kberg:


nwai90 (approved):

Lynesth:
It seems this also fixes the Dry Desert event. 👍

kberg:
Yes, it does.

PR #1735, opened 2020-10-23
end game ui improvement

### LINKED ISSUE #1298: Feature request: who won
We just played a game -- really fun! -- and we were sorta curious to know who won. It seems like useful information that other people might want to have.

https://terraforming-mars.herokuapp.com/the-end?id=af0cbb3c6fd

It knows our names so perhaps it could include a message like "CR won!" or (my preference) "KB won!".

### LINKED ISSUE #1732: Improve the End game screen
Untitled

- table row has the player colour from the game
- table with is auto with total column a bit bigger
- Player points are in multiple columns
  - colums are 400px wide, so the text "Victory points details for" and the player name are on 2 rows 
  - vertically align to top
  - width is auto with column total with bigger/fixed width
- backlog is removed (I find it obsolete for the purpose of the screenshots)

vsrisuknimit:
This PR can close #1298 and close #1732 .

- Change color of row each to match player color.
- Report score in columns that wrap around
- Move log to be in the same row as board
- Announce who is the winner

image

Wrap around if all columns don't fit in one row.
image

If two winners, say both. Rare to happen.
image

### DIFF
--- a/src/components/GameEnd.ts
+++ b/src/components/GameEnd.ts
@@ -3,6 +3,7 @@ import { PlayerModel } from "../models/PlayerModel";
 import { Board } from "./Board";
 import { LogPanel } from "./LogPanel";
 import { Button } from "../components/common/Button";
+import { playerColorClass } from "../utils/utils";
 
 export const GameEnd = Vue.component("game-end", {
     props: {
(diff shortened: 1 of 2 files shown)


alrusdi (approved):

vsrisuknimit:
@alrusdi any suggestion on how to put / between winners if there is more than one?

Also I want Total Point column to be wider but I added style to the element instead of <th>. Somehow I could not influence that specific <th>, any suggestion?

alrusdi:
> @alrusdi any suggestion on how to put / between winners if there is more than one?
> 

Just generate all html block in JS function. Use v-html directive to insert it into template.

> Also I want Total Point column to be wider but I added style to the element instead of . Somehow I could not influence that specific , any suggestion?

Your current approach is correct and straightforward. There no reasons to use any tricks. Simple is better than complex (c) :-)

PR #1737, opened 2020-10-24
Add defer() argument to unshift deferred actions

### LINKED ISSUE #848: Bug: Mars University interrupt triggers before Olympus Conference

Lynesth:
Fixes #848.

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -526,8 +526,12 @@ export class Game implements ILoadable<SerializedGame, Game> {
         });
     }
 
-    public defer(action: DeferredAction): void {
-        this.deferredActions.push(action);
+    public defer(action: DeferredAction, unshift: boolean = false): void {
+        if (unshift) {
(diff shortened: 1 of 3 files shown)


src/Game.ts
```diff
@@ -526,8 +526,12 @@ export class Game implements ILoadable<SerializedGame, Game> {
         });
     }
 
-    public defer(action: DeferredAction): void {
-        this.deferredActions.push(action);
+    public defer(action: DeferredAction, unshift: boolean = false): void {
```
kberg, line 529: Rename `unshift` to `priority` and you got it.

kberg: Or something like that.

Lynesth: Done!

kberg (approved):

PR #1741, opened 2020-10-24
Fix: Don't decrease own production in solo mode

nwai90:
For 3 consortium cards, player's own production got decreased in solo mode


nwai90:
@Lynesth This should address the solo mode bug

alrusdi (approved):

Lynesth:
Sorry for that oversight and thanks for the fix !

PR #1745, opened 2020-10-24
Update win message and hide it if solo loss

nwai90:
- Don't display winner message if solo game was lost
- Updated text copy

### DIFF
--- a/src/components/GameEnd.ts
+++ b/src/components/GameEnd.ts
@@ -94,7 +94,7 @@ export const GameEnd = Vue.component("game-end", {
                         Go to main page
                     </a>
                 </div>
-                <div v-if="!isSoloGame() || (isSoloGame() && player.isSoloModeWin)" class="game-end-winer-announcement"> 
+                <div v-if="!isSoloGame() || player.isSoloModeWin" class="game-end-winer-announcement">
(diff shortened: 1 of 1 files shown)


src/components/GameEnd.ts
```diff
@@ -94,8 +94,8 @@ export const GameEnd = Vue.component("game-end", {
                         Go to main page
                     </a>
                 </div>
-                <div class="game-end-winer-announcement"> 
-                    <span v-for="p in getWinners()"><log-player :class="getEndGamePlayerHighlightColorClass(p.color)">{{ p.name }}</log-player></span> is the winner!
+                <div v-if="!isSoloGame() || (isSoloGame() && player.isSoloModeWin)" class="game-end-winer-announcement">
```
bafolts, line 97: This can be simplified to:

`v-if="isSoloGame() === false || player.isSoloModeWin`

I am excited the vue type compiler is checking this typing for us!

kberg, line 97: Seems this can be simlified to `!isSoloGame() || player.isSoloModeWin`

kberg (approved):

bafolts (approved):

src/components/GameEnd.ts
kberg: Heh

PR #1746, opened 2020-10-24
Vue type checking and unit tests for Award component.

bafolts:
This moves the Award component into the vue type checking system. Also added unit tests to test the `methods` and `data` of the component.

### DIFF
--- a/compile-vue-templates.js
+++ b/compile-vue-templates.js
@@ -1,5 +1,4 @@
 
-
 /**
  * Provide a mock vue so we can interact with
  * the components from node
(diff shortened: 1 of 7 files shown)


kberg (approved):
LGTM but consider another set of more experienced eyes.

bafolts:
@chosta only change of note is moving any usage of `data` out of templates and into `methods` as those properties are difficult to run through the type checker. Manipulating `data` in `methods` allows us to unit test everything that isn't vue related.

chosta:
@bafolts I'll take a look tomorrow and try it myself on some of the Card components

Issue #1743, opened 2020-10-24
Basic corporations off by default when selecting Custom Corp List


vsrisuknimit:
They should be checked. Do not turned them off by default when selecting custom corp list.

![image](https://user-images.githubusercontent.com/14239220/97081628-9e1d4a80-15d1-11eb-8771-2b937d3725b7.png)

PR #1749, opened 2020-10-25
Fix: Select base corps by default in custom corps list

nwai90:
This PR will close https://github.com/bafolts/terraforming-mars/issues/1743


src/components/CorporationsFilter.ts
```diff
@@ -28,6 +28,7 @@ export const CorporationsFilter = Vue.component("corporations-filter", {
             cardsByModuleMap: cardsByModuleMap,
             customCorporationsList: false,
             selectedCorporations: [
+                ...cardsByModuleMap.get(GameModule.Base)!,
```
kberg, line 31: @bafolts has been pushing back on using !. I think I'm going to have to do something other than a Map to remove the unnknown-ness.

kberg (approved):

PR #1750, opened 2020-10-25
Fix: Expansion icons appear twice on corporation cards

nwai90:
Screenshot 2020-10-25 at 3 05 13 PM

### DIFF
--- a/src/styles/cards.less
+++ b/src/styles/cards.less
@@ -1493,6 +1493,13 @@ input[type="radio"]:checked + .filterDiv::after {
     text-align:center;
 }
 
+.card-utopia-invest {
+    .production.wild {
+        vertical-align: middle;
+        margin-left: 5px;
+    }
+}
+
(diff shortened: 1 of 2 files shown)


src/styles/cards.less
```diff
@@ -1493,6 +1493,13 @@ input[type="radio"]:checked + .filterDiv::after {
     text-align:center;
 }
 
+.card-utopia-invest {
```
kberg, line 1496: Is this related to expansion icons? If not, is it intentional?  Just asking for you to remove it if it was accidental.

kberg: Wait, sorry I can read.

nwai90: @kberg Intentional. The current layout is broken

Screenshot 2020-10-25 at 10 46 30 PM

kberg (approved):

PR #1775, opened 2020-10-26
Make trade bonus simultaneous for all players

### LINKED ISSUE #1647: Pluto massively slows game down
Any chance we can have players simultaneously discard to pluto for the colony bonus? As it is, having pluto in the game slows a game down by ~15-20 minutes because of the sequential discards. 

Thanks!

Lynesth:
This puts the trade bonuses in their own deferred action.
It allows for all players to pick at the same time the card they want to discard (Pluto), or the one they want to add an resource to (Enceladus, Titan, Miranda).
Fixes #1647


chosta:
This looks legit to me, thanks @Lynesth! Maybe @bafolts can take a look just in case. It is a critical component of the game, we should be certain it doesn't break anything :)

src/deferredActions/GiveTradeBonus.ts
```diff
@@ -0,0 +1,48 @@
+import { Player, PlayerId } from "../Player";
+import { Game } from "../Game";
+import { IColony } from "../colonies/Colony";
+import { DeferredAction } from "./DeferredAction";
+
+export class GiveTradeBonus implements DeferredAction {
+    public cb: () => void = () => {};
+    private waitingFor: Record<PlayerId,number> = {};
```
bafolts, line 8: You should use `Multiset` here. If you add a `remove(key: T): void` that decrements and removes the key as you have here you can avoid duplicating this logic. To check when there are no items you can then use `multiSet.entries().length === 0`.

bafolts:
Does something need to change for https://github.com/bafolts/terraforming-mars/blob/master/src/cards/colonies/ProductiveOutpost.ts#L17 ? That calls to `giveTradeBonus`.

bafolts (changes requested):
Leveraging `Multiset` in `GiveTradeBonus` should go a long way. You should add unit tests for `GiveTradeBonus`. Then that deferred is action if there is a bug and we never call the callback the game will get stuck.

Lynesth:
> Does something need to change for https://github.com/bafolts/terraforming-mars/blob/master/src/cards/colonies/ProductiveOutpost.ts#L17 ? That calls to `giveTradeBonus`.

Good catch, thanks. Yes it needs to be modified, same goes for Trade Advance.

src/deferredActions/GiveTradeBonus.ts
Lynesth: Done. Let me know if that's good for you.

Lynesth:
@bafolts I made the change to use Multiset and I fixed the 2 cards using giveTradeBonus.
I also added a failsafe in `GiveTradeBonus` to check that there's colonies to work with.

I'm not sure what kind of test I should add for it though. Shouldn't adding tests for all colony stuff be enough (which needs to be done at some point) ?

bafolts (approved):
The test I would add is for `GiveTradeBonus`. A `GiveTradeBonus.spec.ts` file would verify mostly that we always end up calling to `cb`. If for whatever reason the local `cb` is not called the game will not progress and appear stuck. Setting up a `GiveTradeBonus` and having every player take their `setWaitingFor` and then verifying that `cb` was called would go a long way.

alrusdi:
@Lynesth please resolve the conflicts

Lynesth:
@alrusdi I was making a bunch of colonies tests first, will get to it in a couple of hours :)

Lynesth:
@alrusdi It's done :)
@bafolts With the tests from #1801 the GiveTradeBonus deferred action is run 11 times. I'll add more tests later to make sure it works with every configuration imaginable.

Lynesth:
I wrote a big test for colonies which includes a `GiveTradeBonus` check. It's pretty verbose and some tests repeat previous ones but I prefered separating things properly so that it's easier to debug if the need comes.

@bafolts The test actually revealed that I broke the functionality when switching to Multiset (it was calling cb after 1 player finished their waitingFor instead of all of them). This is now fixed.

Also turned out that Luna's trade income was bugged as well. Fixed it at the same time.

I think this is now really ready to be merged :)

PR #1779, opened 2020-10-27
Priority to PU effects to lose MC before getting any

### LINKED ISSUE #1179: GMO Contract + Pharmacy Union ordering issue

Lynesth:
As stated in the title, makes **Pharmacy Union** losing MC a priority so that subsequent effects that can give you MC (GMO, Topsoil, getting money from Mons) happens after that.

Closes #1179

### DIFF
--- a/src/cards/promo/PharmacyUnion.ts
+++ b/src/cards/promo/PharmacyUnion.ts
@@ -70,20 +70,11 @@ export class PharmacyUnion implements CorporationCard {
                         orOptions.title = "Choose the order of tag resolution for Pharmacy Union";
                         return orOptions;
                     }
-                ));
+                ), true); // Make it a priority
                 return undefined;
             }
         }
 
-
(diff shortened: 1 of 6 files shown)


src/cards/promo/PharmacyUnion.ts
```diff
@@ -123,10 +114,26 @@ export class PharmacyUnion implements CorporationCard {
                             })
                         );
                     }
-                ));
+                ), true); // Make it a priority
```
kberg, line 117: You know `priorityDefer` sounds pretty good.

nwai90 (approved):

Issue #1733, opened 2020-10-23
Use seedable RNG for creating duplicate games


SliceOfBread:
It looks to me like the game uses the built-in JS random number generator (RNG). As this is not seedable, it requires all the random things (mainly the card deck) to be saved/passed to recreate a game. I mentioned this in #1723 but thought it should get its own topic. 

Currently, you cannot (certainly not easily) play a game on one server and play the same game on a different server. This would be simple if we had a seedable RNG.

A far-fetched potential issue would be where a multi-player game is hosted by one of the players. This player could pre-deal any number of games to ensure they got at least the best starting cards. This is fixable by getting part of the seed from each player.

kberg:
A seed wouldn't be enough -- you need to have uniquely-compatible code-bases. If someone changes dealing order such that cards are dealt one at a time to each individual player rather than deal all n cards at the same time, there's no guarantee of repeatability, and I would be reluctant to provide such a guarantee.

There are plenty of parts of the system that do use a seeded random number generator, by the way.

Lynesth:
We could of course add it since that's what's happening under the hood when you start a predefined game (aside from the fact that it also sets all the game options). But as with predefined games, a change in the code can make the game different.

I wouldn't mind if we added a "Enter a game seed" input but, as @kberg said, we won't guarantee you it'll work properly between versions.

vsrisuknimit:
+1 for enter a game seed.

SliceOfBread:
#1785 has the start of this. It uses the game.id as the seed for the project, prelude and corporate decks. It also uses the game.id to choose a board if the user selects the Random (Tharsis/Hellas/Elysium) option.

I just realized I probably need to add the Turmoil deck and colony selection. Anything else?

vsrisuknimit:
@SliceOfBread random first player, randomize milestone and award, shuffle board tiles.

SliceOfBread:
> What's the end goal of this change? It's possible that whatever is built would work in the short term, but puts a reliability burden on ongoing maintenance and change. For example, someone reading this project later on could read this code and say "Oh! Why are there three RNGs? It doesn't make sense." And then consolidate them into one. And then someone else comes along and reorders shuffling (just like I did with tile behavior, and even with the way decks are structured.)
> 
> Is this something that this project should continue to support ongoing? If so, then there should be enough comments to describe the feature it is supporting and tests to validate the use cases and support it when we organizationally refactor.

[...]

PR #1785, opened 2020-10-27
Seedable rng 1733

SliceOfBread:
The shuffling of corporate, prelude and project decks use the game.id as the seed for the random number generator (rng).
Also the when using the random board selection (1 of the 3 main boards), this is done using the game.id.

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -420,7 +420,7 @@ function createGame(req: http.IncomingMessage, res: http.ServerResponse): void {
 
             if (gameReq.board === "random") {
                 const boards = Object.values(BoardName);
-                boards[Math.floor(parseInt(gameId, 16) / Math.pow(16, 12) * boards.length)];
+                gameReq.board = boards[Math.floor(parseInt(gameId, 16) / Math.pow(16, 12) * boards.length)];
             }
 
             const gameOptions = {
--- a/server.ts
+++ b/server.ts
@@ -420,7 +420,7 @@ function createGame(req: http.IncomingMessage, res: http.ServerResponse): void {
 
             if (gameReq.board === "random") {
                 const boards = Object.values(BoardName);
-                gameReq.board = boards[Math.floor(parseInt(gameId, 16) / Math.pow(16, 12) * boards.length)];
+                boards[Math.floor(parseInt(gameId, 16) / Math.pow(16, 12) * boards.length)];
             }
 
             const gameOptions = {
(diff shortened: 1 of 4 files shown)


server.ts
```diff
@@ -420,7 +420,7 @@ function createGame(req: http.IncomingMessage, res: http.ServerResponse): void {
 
             if (gameReq.board === "random") {
                 const boards = Object.values(BoardName);
-                gameReq.board = boards[Math.floor(Math.random() * boards.length)];
+                gameReq.board = boards[Math.floor(parseInt(gameId, 16) / Math.pow(16, 12) * boards.length)];
```
kberg, line 423: Why does this use the game id rather than the a seeded rng?

src/Dealer.ts
```diff
@@ -110,18 +113,20 @@ export class Dealer implements ILoadable<SerializedDealer, Dealer>{
             projectCardsToRemove.push(...cardsBlackList);
         }
         const filteredDeck = deck.filter((card) => !projectCardsToRemove.includes(card.name));
-        this.deck = this.shuffleCards(filteredDeck);
+        const shuffleSeed = parseInt(gameId, 16) / Math.pow(16, 12);
+        this.deckRng = new Random( shuffleSeed);
+        this.deck = this.shuffleCards(filteredDeck, this.deckRng);
         if (this.usePreludeExtension) {
-            this.preludeDeck = this.shuffleCards(preludeDeck);
+            this.preludeDeck = this.shuffleCards(preludeDeck, new Random(shuffleSeed));
```
kberg, line 120: Looking  at a copy of your repo, there are now three different places that create a new instance of Random. Dealer is created by Game -- how about supplying an instance of Random here?

kberg (changes requested):
Looks like there are some broken tests below?

server.ts
SliceOfBread: gameId is already generated randomly.

src/Dealer.ts
SliceOfBread: Not sure I understand the question.
I used separate instances of Random so it would not matter what order the decks got shuffled in the future. The instance for the project cards is kept around for the case where the discards need to be reshuffled.

kberg: What's the end goal of this change? It's possible that whatever is built would work in the short term, but puts a reliability burden on ongoing maintenance and change. For example, someone reading this project later on could read this code and say "Oh! Why are there three RNGs? It doesn't make sense." And then consolidate them into one. And then someone else comes along and reorders shuffling (just like I did with tile behavior, and even with the way decks are structured.)

Is this something that this project should continue to support ongoing? If so, then there should be enough comments to describe the feature it is supporting and tests to validate the use cases and support it when we organizationally refactor.

SliceOfBread:
cancelling this and the associated issue #1733

PR #1790, opened 2020-10-28
Turn game.deferredActions into a DeferredActionsQueue DS 

Lynesth:
It was asked by @kberg a few times that this should be its own class.
I agree with that idea so I tried to make something that, I think, turned out okay.

It moves the logic of dealing with deferred actions in the `DeferredActionsQueue` class and simplifies the code in Game.ts, Player.ts and some the tests.

First commit is the logic, second only deals with tests.

Note: I'm not even sure why we're making a distinction between taking care of _all_ the deferred actions and taking care of those for a specific player only. Are we ever deferring actions without setting the active player as `player` ? Needs more tests but could probably remove `runAllForPlayer` and use `runAll` everywhere.

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -40,6 +40,7 @@ import { ResourceType } from "./ResourceType";
 import { Resources } from "./Resources";
 import { SelectCard } from "./inputs/SelectCard";
 import { DeferredAction } from "./deferredActions/DeferredAction";
+import { DeferredActionsQueue } from "./deferredActions/DeferredActionsQueue";
 import { SelectHowToPayDeferred } from "./deferredActions/SelectHowToPayDeferred";
 import { PlaceOceanTile } from "./deferredActions/PlaceOceanTile";
 import { RemoveColonyFromGame } from "./deferredActions/RemoveColonyFromGame";
@@ -104,7 +105,7 @@ export class Game implements ILoadable<SerializedGame, Game> {
     public lastSaveId: number = 0;
     private clonedGamedId: string | undefined;
     public seed: number = Math.random();
-    public deferredActions: Array<DeferredAction> = [];
+    public deferredActions: DeferredActionsQueue = new DeferredActionsQueue();
     public gameAge: number = 0; // Each log event increases it
     public gameLog: Array<LogMessage> = [];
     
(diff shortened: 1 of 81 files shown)


src/Game.ts
```diff
@@ -1736,6 +1698,9 @@ export class Game implements ILoadable<SerializedGame, Game> {
       // Assign each attributes
       const o = Object.assign(this, d);
 
+      // Brand new deferred actions queue
```
kberg, line 1701: Necessary comment?

src/deferredActions/DeferredActionsQueue.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    public unshift(action: DeferredAction): void {
+        this.queue.unshift(action);
+    }
+
+    public push(action: DeferredAction): void {
+        this.queue.push(action);
+    }
+
+    public next(): DeferredAction | undefined {
```
kberg, line 19: Suggestion: have next() always return DeferredAction, or throw.

src/deferredActions/DeferredActionsQueue.ts
```diff
@@ -0,0 +1,77 @@
...
+            return;
+        }
+
+        this.run(action, () => this.runAll(cb));
+    }
+
+    public runAllForPlayer(playerId: PlayerId, cb: () => void): void {
+        const action = this.nextForPlayer(playerId);
+        if (action === undefined) {
+            cb();
```
kberg, line 57: Can you explain something to me? When is cb supposed to run? I assumed it was run after the queue was done. Is that right?

kberg (commented):
There is nothing better than high-speed scrolling through so many easy-to-read tests. Thank you!

This looks great.

src/deferredActions/DeferredActionsQueue.ts
Lynesth: That's the case in `runAll` but in `runAllForPlayer` it's when there's no deferred actions for that player in the queue anymore.

src/Game.ts
Lynesth: Probably not, no. Or it could be elaborated to explain that it's because deferred actions aren't stored that we can't reload them ?

kberg (approved):

PR #1801, opened 2020-10-30
Add tests for official colonies

Lynesth:
Will add more generic tests in a futur PR.

### DIFF
--- a/tests/colonies/Europa.spec.ts
+++ b/tests/colonies/Europa.spec.ts
@@ -1,0 +1,45 @@
+import { expect } from "chai";
+import { Europa } from "../../src/colonies/Europa";
+import { Color } from "../../src/Color";
+import { Player } from "../../src/Player";
+import { Game } from "../../src/Game";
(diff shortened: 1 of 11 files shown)


tests/colonies/Europa.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+        player = new Player("test", Color.BLUE, false);
+        player2 = new Player("test2", Color.RED, false);
+        game = new Game("foobar", [player, player2], player);
+        game.gameOptions.coloniesExtension = true;
+        game.colonies.push(europa);
+    });
+
+    it("Should build", function() {
+        europa.onColonyPlaced(player, game);
+        expect(game.deferredActions).has.lengthOf(1);
```
kberg, line 23: Just thinking: make `DeferredActions` a first-class type that hides the deferred actions array. Then write a chai matcher on DeferredActions with method soleAction(). It would make things neater and smaller. A minor thing of course.

kberg (approved):

Issue #1802, opened 2020-10-30
Important Gameplay Bug: Energy Market and Power Infrastructure always read 0 as input


vsrisuknimit:
If you adjust any of these two cards with using the up and down button in the input box, the outcome is always 0.

![image](https://user-images.githubusercontent.com/14239220/97692750-884ad200-1a76-11eb-9aaf-abe9e3b0c2e5.png)

![image](https://user-images.githubusercontent.com/14239220/97692773-8f71e000-1a76-11eb-8126-e7e9b4e56ad6.png)

![image](https://user-images.githubusercontent.com/14239220/97692802-9993de80-1a76-11eb-856c-8e1c83089182.png)

vsrisuknimit:
MAX button however works.

![image](https://user-images.githubusercontent.com/14239220/97693066-f8595800-1a76-11eb-9c3c-9619f869597f.png)

vsrisuknimit:
Problem is with SelectAmount function, likely.

Same problem with anything that spend x amount. Here is what I tested so far.
- Titan Shuttles
- Insulation
- Sulphur Eating Bacteria.

PR #1795, opened 2020-10-29
More type checking on vue components

bafolts:
More type safety on vue components. Adds `SelectAmount` and `SelectHowToPayForCard`.

PR #1803, opened 2020-10-30
Fix SelectAmount not working

Lynesth:
@bafolts Reverting that change from https://github.com/bafolts/terraforming-mars/pull/1795 that broke SelectAmount.
It's probably just a matter of removing the `.data` from the methods but I don't know that part of the code and I don't have enough time to check that everything will still work afterwards, hence the simple revert.

Fixes https://github.com/bafolts/terraforming-mars/issues/1802

### DIFF
--- a/compile-vue-templates.js
+++ b/compile-vue-templates.js
@@ -38,10 +38,6 @@ checkComponent(
     require("./dist/src/components/OtherPlayer").OtherPlayer
 );
 checkComponent(
-    "src/components/SelectAmount",
-    require("./dist/src/components/SelectAmount").SelectAmount
-);
-checkComponent(
     "src/components/SelectOption",
     require("./dist/src/components/SelectOption").SelectOption
 );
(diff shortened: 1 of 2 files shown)


bafolts (approved):

bafolts:
The vue template compilers uses `with (this)`. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/with

```
Use of the with statement is not recommended, as it may be the source of confusing bugs and compatibility issues. See the "Ambiguity Contra" paragraph in the "Description" section below for details.
```

This is a confusing bug to me. I am assuming that within `with (this)` my usage of `this.amount` somehow is the wrong `thisArg`. If anything I am going to add to the template checker to not allow using `this` in templates. In the template checker I strip `with` as it doesn't work with typescript. Here I had to use `this.amount` so typescript could find the `amount`. In an upcoming PR I had this reverted anyway with a fix to not rely on `this`.

PR #1805, opened 2020-10-30
More components, even stricter type checking for v-model

bafolts:
Ever evolving process to wrangle in vue components and adhere them to types. As there have been growing pains with earlier PRs  I have put in checks to prevent those from happening in the future.

Disallow `this` inside templates. Using `this` inside templates will most likely not work as the eventual vue template is wrapped in a `with (this)`. An error is now thrown if these are spotted.

This would have caught my issue in `SelectAmount` as follows.

```
Error: don't use this inside template string for src/components/SelectAmount
    at checkComponent (/home/algorithm101/Projects/terraforming-mars/compile-vue-templates.js:134:15)
    at Object.<anonymous> (/home/algorithm101/Projects/terraforming-mars/compile-vue-templates.js:64:1)
```

Messing up `v-model` types are also now spotted. This would have caught the place I forgot to update in `SelectHowToPayForCard` as follows.

```
dist/src/components/SelectHowToPayForCardVue.ts:134:501 - error TS2322: Type 'CardModel' is not assignable to type 'string'.
```

With this change I avoided modifying the template or changing data structure. I did remove props which I couldn't see being used.


src/components/Board.ts
```diff
@@ -149,7 +181,7 @@ export const Board = Vue.component("board", {
             <div class="global-numbers-oceans" v-html="oceansValue()">
             </div>
 
-            <div v-if="aresExtension">
+            <div v-if="aresExtension && aresData !== undefined">
```
bafolts, line 184: Modified template since aresData can be undefined. Error was thrown that it could be undefined with checker in place.

src/components/CardsFilter.ts
```diff
@@ -7,17 +7,15 @@ import { ALL_PROJECT_CARD_NAMES } from "../cards/AllCards";
...
 
 interface CardsFilterModel {
-    customCorporationsList: boolean;
     selectedCardNames: Array<CardName>;
     foundCardNames: Array<CardName>;
     searchTerm: string;
 }
 
 export const CardsFilter = Vue.component("cards-filter", {
-    props: ["corporateEra", "prelude", "venusNext", "colonies", "turmoil", "promoCardsOption", "communityCardsOption"],
```
bafolts, line 17: I couldn't see where we were using any of these props internally. I assume they were copied from `CorporationsFilter`. Appears to still work without them and they are not referenced in this component.

src/components/ColoniesFilter.ts
```diff
@@ -51,7 +56,7 @@ export const ColoniesFilter = Vue.component("colonies-filter", {
             selectedColonies: [
                 ...officialColonies,
                 ...this.communityCardsOption ? communityColonies: []
-            ]
+            ] as Array<IColony> | boolean
```
bafolts, line 59: With `v-model` type checking in place this can be set to `true` or `false` if for some reason `selectedColonies` were not an array at runtime.

src/components/ColoniesFilter.ts
```diff
@@ -66,6 +71,7 @@ export const ColoniesFilter = Vue.component("colonies-filter", {
     },
     watch: {
         selectedColonies: function (value) {
+            console.log(this.selectedColonies);
```
bafolts, line 74: Will remove. Was debugging why vue thought `selectedColonies` could be `boolean`.

compile-vue-templates.js
```diff
@@ -126,15 +202,21 @@ function checkComponent(name, component) {
 
     // append types for minified functions
     lines.unshift("declare const _c: any;");
-    lines.unshift("declare const _e: any;");
+    lines.unshift("declare function _e(): void;");
```
kberg, line 205: This part is starting to get complicated -- what base knowledge do you expect from people trying to debug how templates are compiled?

src/components/Board.ts
kberg: :+1:

src/components/CardsFilter.ts
kberg: Wow.

kberg (approved):

compile-vue-templates.js
bafolts: I am hoping once all of these are in place the only typescript warnings the end user sees are related to their templates and this is hidden away. I assume the vue template compiler makes these functions short in name to avoid colliding with a template function. If we were to use a function named `_c` in our methods I assume this would cause a collision. The generated function is on one line but other than that reading through the generated template is straight forward. I will add pretty print to the compiled function so that it isn't on one line. After that the errors a user should see should be related to their use of types and not the typing here. In the hopefully rare scenario where there is a typing error here we would have to update the type. For the time being we could mark that failure as only a warning.

PR #1809, opened 2020-10-31
Type safety vue game model

bafolts:
More work adding type safety to the vue components. Trying to do this piece mail to ease reviewing and since most of this I am testing manually as the change goes out. I added beautify to the compiled vue template to ease with debugging any type errors that may arise. We didn't get any model for `getGame` and were returning serialized `Player` to the client. We will only pass back the data we need from the server now as well as the included type safety.

### DIFF
--- a/index.html
+++ b/index.html
@@ -30,7 +30,6 @@
                 <game-end
                     v-else-if="screen === 'the-end'"
                     :player="player"
-                    :game="game"
                 ></game-end>
                 <games-overview
                     v-else-if="screen === 'games-overview'"
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
   },
   "scripts": {
     "clean": "rm -r ./dist/ ./styles.css || true",
-    "build": "node make_static_json.js && eslint src/*.ts src/database/**/*.ts && lessc src/styles/common.less styles.css && tsc --build tsconfig.json && npm run templates && webpack --config webpack.config.js",
+    "build": "node make_static_json.js && eslint src/*.ts src/components/**/*.ts src/database/**/*.ts && lessc src/styles/common.less styles.css && tsc --build tsconfig.json && npm run templates && webpack --config webpack.config.js",
     "start": "node make_static_json.js && lessc src/styles/common.less styles.css && tsc --build tsconfig.json && webpack --config webpack.config.js && node dist/server.js",
     "pretest": "tsc --build tsconfig-test.json",
     "templates": "node compile-vue-templates.js && tsc --build tsconfig-vue.json",
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "chai": "^4.2.0",
     "eslint": "^6.6.0",
     "eslint-plugin-vue": "^6.2.2",
+    "js-beautify": "^1.13.0",
     "mocha": "^6.2.3",
     "nyc": "^14.1.1"
   },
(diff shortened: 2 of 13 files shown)


index.html
```diff
@@ -30,7 +30,6 @@
                 <game-end
                     v-else-if="screen === 'the-end'"
                     :player="player"
-                    :game="game"
```
bafolts, line 33: This was not being used. I played a game until the end screen to verify it ran without error.

package.json
```diff
@@ -29,6 +29,7 @@
     "chai": "^4.2.0",
     "eslint": "^6.6.0",
     "eslint-plugin-vue": "^6.2.2",
+    "js-beautify": "^1.13.0",
```
bafolts, line 32: The compiled template is fairly easy to read through when it isn't on one line. This should help with debugging any issues that arise with the compiled templates.

server.ts
```diff
@@ -984,11 +985,15 @@ function getSpaces(spaces: Array<ISpace>): Array<SpaceModel> {
 }
 
 function getGame(game: Game): string {
-    const output = {
+    const output: GameModel = {
         activePlayer: game.getPlayerById(game.activePlayer).color,
         id: game.id,
         phase: game.phase,
-        players: game.getPlayers(),
+        players: game.getPlayers().map((player) => ({
```
bafolts, line 992: We weren't using all of `Player`. This only sends to the server the properties we need.

index.html
kberg: I'm going to need some parameter to fix an Ares bug, because as of now the end-screen board doesn't show ares-specific versions of special tiles. But that doesn't need all of game, just the AresExpansion option. So, no action on your part. Just FYI.

server.ts
kberg: Should there be a concrete type for this?

src/components/App.ts
```diff
@@ -27,11 +54,9 @@ export const mainAppSettings = {
             "pinned_player_3": false,
             "pinned_player_4": false,
             "turmoil_parties": false,
-        },
-        game: {
-            players: [],
-        },
-    },
+        } as {[x: string]: boolean},
```
kberg, line 57: I was trying to figure out how to express something like this elsewhere. Thank you.

src/models/GameModel.ts
```diff
@@ -0,0 +1,16 @@
...
+import { Phase } from "../Phase";
+
+export interface GameModel {
+    activePlayer: Color;
+    id: string;
+    phase: Phase;
+    players: Array<GamePlayerModel>;
+}
+
+export interface GamePlayerModel {
```
kberg, line 12: Since there's already a PlayerModel can you add a note explaining why this is different / and/or necessary?

kberg (approved):
Minor comments - some ignorable.

server.ts
bafolts: There is a concrete type that typescript is enforcing. 

https://github.com/bafolts/terraforming-mars/blob/type-safety-vue-game/src/models/GameModel.ts#L12

src/models/GameModel.ts
bafolts: I will leave a comment. This should be used with the `GameHome` component. Probably makes sense to change this file to `GameHomeModel` to line up. The existing `PlayerModel` should be `PlayerHomeModel`.

bafolts:
Almost done with all the components! Some of these are riskier than others especially when we weren't relying on types. I will be keeping an eye out for any issues that come up to fix quickly. I played through the `GameHome` component through multiple routes and didn't spot an issue.

PR #1814, opened 2020-11-02
Reduce code duplication in cards that remove resources from other cards and allow self removal on Ants and Predators.

Lynesth:
Adds a new deferred action called **RemoveResourcesFromCard** that is used by cards or actions that can remove resources from other cards.
It is now used by **Ants**, **Predators**, **Virus**, **Stratospheric Birds** and **Air Raid**.

Also added an argument to `player.getCardsWithResources()` to allow filtering what kind of cards we're looking for.

First commit is functional, second is for tests.

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -498,7 +498,7 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
           if (removingPlayer !== this) this.resolveMonsInsurance(game);
 
           if (shouldLogAction) {
-              game.log("${0} removed ${1} resource" + (count > 1 ? "(s)" : "") + " from ${2}'s ${3}", b =>
+              game.log("${0} removed ${1} resource" + (count > 1 ? "s" : "") + " from ${2}'s ${3}", b =>
               b.player(removingPlayer)
               .number(count)
               .player(this)
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -498,11 +498,11 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
           if (removingPlayer !== this) this.resolveMonsInsurance(game);
 
           if (shouldLogAction) {
-            game.log("${0} loses ${1} resource(s) on ${2} by ${3}", b =>
-                  b.player(this)
-                  .number(count)
-                  .card(card)
-                  .player(removingPlayer));
+              game.log("${0} removed ${1} resource" + (count > 1 ? "(s)" : "") + " from ${2}'s ${3}", b =>
+              b.player(removingPlayer)
+              .number(count)
+              .player(this)
+              .card(card));
           }
         }
         // Lawsuit hook
@@ -527,23 +527,20 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
       }
     }
 
-    public getCardsWithResources(): Array<ICard> {
-      const result: Array<ICard> = [];
-
-      this.playedCards.forEach((card) => {
-        if (card.resourceType !== undefined && card.resourceCount && card.resourceCount > 0) {
-          result.push(card);
+    public getCardsWithResources(resource?: ResourceType): Array<ICard> {
+        let result: Array<ICard> = this.playedCards.filter((card) => card.resourceType !== undefined && card.resourceCount && card.resourceCount > 0);
+        if (this.corporationCard !== undefined 
+            && this.corporationCard.resourceType !== undefined 
+            && this.corporationCard.resourceCount !== undefined 
+            && this.corporationCard.resourceCount > 0) {
+                result.push(this.corporationCard);
         }
-      });
 
-      if (this.corporationCard !== undefined 
-          && this.corporationCard.resourceType !== undefined 
-          && this.corporationCard.resourceCount !== undefined 
-          && this.corporationCard.resourceCount > 0) {
-        result.push(this.corporationCard);
-      }
-      
-      return result;      
+        if (resource !== undefined) {
+            result = result.filter((card) => card.resourceType === resource)
+        }
+
+        return result;      
     }
 
     public getResourceCards(resource: ResourceType | undefined): Array<ICard> {
@@ -561,11 +558,11 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
     }  
 
     public getResourceCount(resource: ResourceType): number {
-      let count: number = 0;
-      this.getCardsWithResources().filter(card => card.resourceType === resource).forEach((card) => {
-        count += this.getResourcesOnCard(card)!;
-      });
-      return count;
+        let count: number = 0;
+        this.getCardsWithResources(resource).forEach((card) => {
+            count += this.getResourcesOnCard(card)!;
+        });
+        return count;
     }
 
     public getCardsByCardType(cardType: CardType) {
@@ -1221,7 +1218,7 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
 
         if (howToPay.floaters !== undefined && howToPay.floaters > 0) {
           if (selectedCard.name === CardName.STRATOSPHERIC_BIRDS && howToPay.floaters === this.getFloatersCanSpend()) {
-            const cardsWithFloater = this.getCardsWithResources().filter(card => card.resourceType === ResourceType.FLOATER);
+            const cardsWithFloater = this.getCardsWithResources(ResourceType.FLOATER);
             if (cardsWithFloater.length === 1) {
               throw new Error("Cannot spend all floaters to play Stratospheric Birds");
             }
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -498,7 +498,7 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
           if (removingPlayer !== this) this.resolveMonsInsurance(game);
 
           if (shouldLogAction) {
-              game.log("${0} removed ${1} resource" + (count > 1 ? "s" : "") + " from ${2}'s ${3}", b =>
+              game.log("${0} removed ${1} resource(s) from ${2}'s ${3}", b =>
               b.player(removingPlayer)
               .number(count)
               .player(this)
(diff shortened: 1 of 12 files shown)


src/Player.ts
```diff
@@ -498,11 +498,11 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
           if (removingPlayer !== this) this.resolveMonsInsurance(game);
 
           if (shouldLogAction) {
-            game.log("${0} loses ${1} resource(s) on ${2} by ${3}", b =>
-                  b.player(this)
-                  .number(count)
-                  .card(card)
-                  .player(removingPlayer));
+              game.log("${0} removed ${1} resource" + (count > 1 ? "s" : "") + " from ${2}'s ${3}", b =>
```
kberg, line 501: I hear this could cause trouble for translations.

Plus, fwiw, ths is not a refactoring, as the PR suggestions, but a functional change.

Lynesth: This specific line was only used with the card Virus in the entire app. I agree that translations will have to adapt to the change but that's the case each time we change any string. And, by the way, this one makes more sense to me in English than the previous one.

Re "refactoring" vs "functional change" I agree that I the title of this PR might be misleading, and I'll fix it.

nwai90: Leaving it as `(s)` is also a viable option (it was initially done that way for ease of localisation)

Lynesth: It's going to be 2 lines to translate instead of one:
`${0} removed ${1} resources from ${2}'s ${3}` and `${0} removed ${1} resource from ${2}'s ${3}`
instead of `${0} removed ${1} resource(s) from ${2}'s ${3}`

If that's an issue I don't mind putting back the parenthesis of course.

kberg: IMO it's a good idea to restore it. Messages become consistent that way.

src/Player.ts
```diff
@@ -527,23 +527,20 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
       }
     }
 
-    public getCardsWithResources(): Array<ICard> {
-      const result: Array<ICard> = [];
+    public getCardsWithResources(resource?: ResourceType): Array<ICard> {
+        let result: Array<ICard> = this.playedCards.filter((card) => card.resourceType !== undefined && card.resourceCount && card.resourceCount > 0);
```
kberg, line 531: ```let result = [...this.playedCards, this.corporationCard].filter(...);``` perhaps.

Does `resourceType === undefined` have a meaningful purpose? It's not obvious to me.

src/Player.ts
```diff
@@ -527,23 +527,20 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
       }
     }
 
-    public getCardsWithResources(): Array<ICard> {
-      const result: Array<ICard> = [];
+    public getCardsWithResources(resource?: ResourceType): Array<ICard> {
```
kberg, line 530: consider `resource?: ResourceType = undefined` to preserve original signature?

src/cards/Virus.ts
```diff
@@ -22,70 +21,36 @@ export class Virus implements IProjectCard {
...
-        };
+        const orOptionsPlants = (new RemoveAnyPlants(player, game, 5)).execute() as OrOptions;
+        const removePlants = orOptionsPlants !== undefined
+            ? orOptionsPlants.options.slice(0, -1)
+            : undefined;
 
         // If no other player has resources to remove
         // assume player will remove nothing from themselves
-        if (cards.length === 0 && playersWithPlants === 0) {
+        if (removeAnimals === undefined && removePlants === undefined) {
```
kberg, line 41: I imagine a future where some project card or event that says "gain 2 plants of you have zero plants" so you play this card when you have only one plant. I don't think there's anything to do here.

src/deferredActions/RemoveResourcesFromCard.ts
```diff
@@ -0,0 +1,89 @@
...
+import { Player } from "../Player";
+import { ResourceType } from "../ResourceType";
+import { OrOptions } from "../inputs/OrOptions";
+import { SelectCard } from "../inputs/SelectCard";
+import { SelectOption } from "../inputs/SelectOption";
+import { CardName } from "../CardName";
+import { ICard } from "../cards/ICard";
+import { DeferredAction } from "./DeferredAction";
+
+const animalsProtectedCards = [ CardName.PETS, CardName.BIOENGINEERING_ENCLOSURE ];
```
kberg, line 11: Mind adding a TODO to make this a card-specific attribute? That way, for instance @chosta can add it to the card's description automatically..

src/deferredActions/RemoveResourcesFromCard.ts
```diff
@@ -0,0 +1,89 @@
...
+        public player: Player,
+        public game: Game,
+        public resourceType: ResourceType | undefined,
+        public count: number = 1,
+        public ownCardsOnly: boolean = false,
+        public mandatory: boolean = true,
+        public title: string = "Select card to remove " + count + " " + resourceType + "(s)"
+    ){}
+
+    public static getAvailableTargetCards(player: Player, game: Game, resourceType: ResourceType | undefined, ownCardsOnly: boolean = false): Array<ICard> {
```
kberg, line 24: nit: Can you move this below everything else? Reading this class, I have to go from class-level context, underestanding the attributes to static context for what is essentially a (useful) utility method.

src/deferredActions/RemoveResourcesFromCard.ts
```diff
@@ -0,0 +1,89 @@
...
+import { ICard } from "../cards/ICard";
+import { DeferredAction } from "./DeferredAction";
+
+const animalsProtectedCards = [ CardName.PETS, CardName.BIOENGINEERING_ENCLOSURE ];
+
+export class RemoveResourcesFromCard implements DeferredAction {
+    constructor(
+        public player: Player,
+        public game: Game,
+        public resourceType: ResourceType | undefined,
```
kberg, line 17: What does resourceType undefined mean here?

src/deferredActions/RemoveResourcesFromCard.ts
```diff
@@ -0,0 +1,89 @@
...
+            resourceCards,
+            (foundCards: Array<ICard>) => {
+                const card = foundCards[0];
+                const owner = this.game.getCardPlayer(card.name);
+                owner.removeResourceFrom(card, 1, this.game, this.player);
+                return undefined;
+            }
+        );
+
+        if (this.mandatory) {
```
kberg, line 73: This is interesting. Can you add a comment to the constructor explaining what mandatory means? I'm still not quite sure what it means.

kberg (commented):
This is a good bit of work. All minor stuff. I'll pull in the request later today (whatever your time zone is.)

src/Player.ts
Lynesth: Given the use of the question mark, the argument is optional and defaults to undefined already or I did not understand what you said.

kberg: Sorry, /me ignorant

src/Player.ts
Lynesth: > Does `resourceType === undefined` have a meaningful purpose? It's not obvious to me.

Which equality are you refering to ?

src/cards/Virus.ts
Lynesth: It's true I suppose. I just kept the original behavior of the card here. I'll remove it.

src/deferredActions/RemoveResourcesFromCard.ts
Lynesth: Sure, I can do that !
It is when the removal is mandatory (Air Raid or Stratospheric Birds for example) in opposition to when you can choose not to do it (Virus).

src/deferredActions/RemoveResourcesFromCard.ts
Lynesth: No problem, will do.

src/deferredActions/RemoveResourcesFromCard.ts
Lynesth: Setting it to undefined would allow removing any type of resource (e.g. "Remove a resource from any card")
It is not used by any card at the moment but it could one day and since underlying functions (`player.getCardsWithResources`) support it, I thought I might as well allow it.

But your comment made me rethink about it and I'll remove it, forcing a proper ResourceType to be passed.

src/Player.ts
Lynesth: Also regarding 
> `let result = [...this.playedCards, this.corporationCard].filter(...);` perhaps.

It's a bit tricky since this.corporationCard can be `undefined` so we'd need to do:
```ts
let results: Array<ICard>;
if (this.corporationCard !== undefined) {
    result = [this.corporationCard, ...this.playedCards].filter(...);
} else {
    result = this.playedCards.filter(...);
}
```
I suppose I can change it to that though.

src/cards/Virus.ts
Lynesth: Actually this is how the whole app works at the moment. You don't get the option to remove your own plants, even when using asteroids and such.
Since there's no incentive to remove your own plants at the moment I suppose it can stay like this ? It'll be very easy to add back in as soon as it's needed.
Or maybe we can ask people if they'd like that option.

src/Player.ts
kberg: Nah you do what you were doing -- it's fine.

I just don't know what it means if resourceType is undefined. Is it just "return all the cards with resources on it?

src/cards/Virus.ts
kberg: Nah. Thanks.

src/Player.ts
Lynesth: > I just don't know what it means if resourceType is undefined. Is it just "return all the cards with resources on it?

Yes. It was the original behavior and we had to filter once more sometimes, leading to duplication. That's why I added the argument here directly.

kberg (approved):

kberg:
Great!

Issue #1576, opened 2020-10-11
With a corp-mandated first action, let the player play it voluntarily


kevinb9n:
Better if the mandated first action doesn't just happen automatically while I'm not looking.

Then I'll be more aware that it happened, and more aware that I have either 0 or 1 actions left, not 1 or 2 as usual.

For Vitor it's the same "Fund an award" option we already have, just for no money. Possible verbiage for the others:

* Draw 3 cards (Inventrix)
* Draw 3 Venus cards (Morning Star)
* Draw 2 floater cards (Celestic)
* Draw a microbe card (Splice)
* Choose and play a Prelude card (Valley Trust)
* Choose an additional colony tile (Aridor)
* Place a city tile (Tharsis Republic)
* Place a greenery tile (Philares)
* Place a colony (Poseidon)
* Place a community (Arcadian)

In fact, the current behavior is actually a rules error. A player with one of these corps is allowed to pass. If so, in gen 2 they remain in the same state: only two choices, special action or pass. When you have no other plans for gen 1, having this ability to defer can sometimes be helpful. A corp like Vitor or even Tharsis might prefer to wait and see what other players do during the first generation before deciding.

kberg:
How are the the corp-mandated first actions for Tharsis Republic, Philares, et cetera not played voluntarily? I seem to recall that you are told to choose a space.

kevinb9n:
Sure, it's only the first four in the list that completely run themselves, but that's not really the point.

I think if the player can be in control of whether to initiate it, it's more correct rules-wise and also a less confusing experience.

kberg:
Just to make sure I get it -- you're saying that in those cases the first
action shouldn't happen automatically, but just be a single option
available "execute this starter corp action?"

>

kevinb9n:
Yes, that's how I think it should be in all ten cases - that option, and "pass this generation".

Lynesth:
@kevinb9n I wanted to take a look at this today and add relevant buttons, but was not able to find an official comment saying you can pass your first generation and not take your first action. While it does make sense to me, do you have a source for that please ?

Lynesth:
Nevermind, I found it just after commenting: https://boardgamegeek.com/thread/1784175/ocean-adjacency-and-first-actions#25942959

kevinb9n:
Should have included that the first time, sorry!

PR #1815, opened 2020-11-02
Prompt player for corporation's first action

Lynesth:
This change prompts the player to actually play the first action of their corporation if they have one (Tharsis, Inventrix, etc).
It also gives them the right to pass the generation instead if they prefer. They will still have to take that action as their first before they are allowed to do anything else than pass.

explorer_2020-11-03_01-54-16

While it is a correct implementation of the rules since delaying that first action is [a legal move](https://boardgamegeek.com/thread/1784175/article/25942959#25942959), some might say that it adds 1 extra click per game (if you have one of those corps) while you would have near 0 reason to ever not take it.
I'll let you guys be the judge of that.

Closes https://github.com/bafolts/terraforming-mars/issues/1576

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -101,6 +101,7 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
     public actionsTakenThisRound: number = 0;
     private actionsThisGeneration: Set<string> = new Set<string>();
     public lastCardPlayed: IProjectCard | undefined;
+    private corporationInitialActionDone: boolean = false;
 
     // Cards
     public dealtCorporationCards: Array<CorporationCard> = [];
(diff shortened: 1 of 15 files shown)


alrusdi (approved):

src/Player.ts
```diff
@@ -2019,31 +2020,34 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
...
-      if (game.hasPassedThisActionPhase(this) || this.actionsTakenThisRound >= 2) {
-        this.actionsTakenThisRound = 0;
-        game.playerIsFinishedTakingActions();
-        return;
-      }         
+        const corporationCard = this.corporationCard;
+        if (corporationCard !== undefined &&
+            corporationCard.initialAction !== undefined &&
+            corporationCard.initialActionText !== undefined &&
+            this.corporationInitialActionDone === false
```
bafolts, line 2033: For games that are currently serialized in the database this value will come in as `undefined`. Games in that state would never run this action as `undefined !== false`. If a game were in the first round a user would never take this first action. That is probably fine as we make no guarantees as to games played on the demo server.

bafolts (approved):
This change has linked rules to the game's creator. Who are we to judge?

src/Player.ts
Lynesth: That's for you to decide. I agree that there's no guarantee with the demo server and it probably won't even impact anyone since the context is very specific (PR needs to be merged and server restarted while someone is on the 1st round of the 1st gen with an initial action corp and has not yet taken its action).

kevinb9n:
Thanks!!
I feel like a scenario where it makes sense to delay is not *completely* outlandish and will happen sooner or later. But if nothing else, I think it will just be less confusing now. More than a few times in my family I have heard "what the hell is happening, why does it think I'm placing a tile?" etc.   Long story short, THANKS again!

PR #1819, opened 2020-11-03
Load game can now delete x turns to be able to restore a game

pierrehilbert:
Will be useful in case of troubles with an action.


Lynesth:
I had no idea we had a load game page. I tried it but it doesn't seem to work properly. Anyway.
Doesn't your change allow modifications of games that could be running though ?
Shouldn't the whole load game thing duplicate a game instead and so you can select the save you want ?

chosta:
> I had no idea we had a load game page. I tried it but it doesn't seem to work properly. Anyway.
> Doesn't your change allow modifications of games that could be running though ?
> Shouldn't the whole load game thing duplicate a game instead and so you can select the save you want ?

We wanted this as a last ditch effort to save a lost game. It seems a bit intrusive, yes, but it is a start. You are correct that it could break a game and is an api that could do harm in the wrong hands. While it's good to have we need some protection I do agree.

compile-vue-templates.js
```diff
@@ -82,7 +82,7 @@ checkComponent(
 checkComponent(
     "src/components/LoadGameForm",
     require("./dist/src/components/LoadGameForm").LoadGameForm,
-    ["gameId"]
+    ["gameId", "nbrSavesToDelete"]
```
bafolts, line 85: Want to find a way to get these values through reflection of some kind. Glad you figured this out time being!

server.ts
```diff
@@ -259,6 +259,11 @@ function loadGame(req: http.IncomingMessage, res: http.ServerResponse): void {
             const gameReq = JSON.parse(body);
 
             const game_id = gameReq.game_id;
+            const nbrSavesToDelete = gameReq.nbrSavesToDelete;
+
+            if (nbrSavesToDelete > 0) {
+                Database.getInstance().deleteGameNbrSaves(game_id, nbrSavesToDelete);
```
bafolts, line 265: Making this a separate API route may make sense. It doesn't necessarily need a user interface. If the route were provided the `serverId` this would ensure that the caller of the route had the secret to the server.

compile-vue-templates.js
kberg: nbrSavesToDelete is a bit cumbersome - `rollbackCount`? Despite that, this is great!

src/database/PostgreSQL.ts
```diff
@@ -197,4 +197,14 @@ export class PostgreSQL implements IDatabase {
             }
         });
     }
+
+    deleteGameNbrSaves(game_id: string, NbrSavesToDelete: number): void {
+        if (NbrSavesToDelete > 0) {
+            this.client.query("DELETE FROM games WHERE ctid IN (SELECT ctid FROM games WHERE game_id = $1 ORDER BY save_id DESCLIMIT $2)", [game_id, NbrSavesToDelete], (err) => {
```
bafolts, line 203: I am assuming `DESCLIMIT` should be `DESC LIMIT`.

Lynesth:
I still think this shouldn't delete saves but instead make a copy of the game. What do you think of doing it like this ?

In `loadGame`, use the `restoreGame` method that accepts a `save_id` argument to retrieve the specific save you want, assign a new id to the game using `generateRandomGameId` and save it in the database. Then redirect to the `/game?id=` url.

https://github.com/bafolts/terraforming-mars/blob/972020dd4d3c47370a79234ab27de37dff80a1be/src/database/SQLite.ts#L122

pierrehilbert:
> I still think this shouldn't delete saves but instead make a copy of the game. What do you think of doing it like this ?
> 
> In `loadGame`, use the `restoreGame` method that accepts a `save_id` argument to retrieve the specific save you want, assign a new id to the game using `generateRandomGameId` and save it in the database. Then redirect to the `/game?id=` url.
> 
> https://github.com/bafolts/terraforming-mars/blob/972020dd4d3c47370a79234ab27de37dff80a1be/src/database/SQLite.ts#L122

Why should we make a copy of the game? 
The main objective is to fix the current game, I think we will multiply useless saves in the database if we just copy it.

Lynesth:
> Why should we make a copy of the game?
The main objective is to fix the current game, I think we will multiply useless saves in the database if we just copy it.

It's just to prevent tampering with other people games, nothing more. Also I don't think that we live in a time where a few rows in a database are an issue, neither space wise or performance wise.

Or as @bafolts suggested, provide the serverId to the route to ensure that it's an admin calling that page ?

pierrehilbert:
We could add this but it will only prevent someone who has the game id but don't play the game to delete some saves.
I think that the probability to find a current game is quite similar with the one to find the serverId.
In addition, it will add the need to have an admin to do it so we will have to wait instead of just being able to do it self .
Again, this feature isn't for me, I can already do it manually on my server

bafolts (approved):

PR #1820, opened 2020-11-03
Colony rewrite

Lynesth:
This PR does a few things, I'll try to be exhaustive.

#### Logic from all colonies has been moved to the Colony class
Each colony now only "describes" itself and the `Colony` class holds the logic needed to interact with them.
e.g. Europa.ts:
```ts
export class Europa extends Colony {
    public name = ColonyName.EUROPA;
    public description = "Production";
    public buildType = ColonyBenefitType.PLACE_OCEAN_TILE;
    public tradeType = ColonyBenefitType.GAIN_PRODUCTION;
    public tradeResource = [
        Resources.MEGACREDITS, Resources.MEGACREDITS,
        Resources.ENERGY, Resources.ENERGY,
        Resources.PLANTS, Resources.PLANTS, Resources.PLANTS
    ];
    public colonyBonusType = ColonyBenefitType.GAIN_RESOURCES;
    public colonyBonusResource = Resources.MEGACREDITS;
}
```
It streamlines the way trade income is calculated instead of somewhat strange methods here and there (turned out Pluto and Luna did not give the proper benefits if the marker was on the first space). Speaking of benefits, the `ColonyBenefitType` enum holds the different type of things a player can get from interacting with a colony, whether it is from building a colony, trading with it or simply collecting its colony bonus.

This is the main change of this PR and allowed to remove a lot of duplicated code (~250 less lines of code).

#### "Increase colony track before trade" improvement
This has been streamlined for every colonies/cards as well. It will only ask if it would actually end with a different outcome (gaining a different resource/production) and the player will be able to choose how much it wants to increase the track if applicable.

#### Rename "Trade Bonus" to "Colony Bonus"
This is the official terms for what players get when they have colonies on a colony and someone trades with it.

#### Removes IColony
Everything now uses directly the `Colony` interface. It was a bit of a pain to have methods that expected `Colony` while others expected `IColony` and having to work around that.

-----

This makes for a bulky `Colony` class but, overall, I think it makes code easier to read and manage in the futur.

Since @kberg mentioned he's not good with reviewing colonies stuff and since I cannot request reviews, could @bafolts @nwai90 and @alrusdi give that PR a look please ? Thanks !

If you want to test a live server of this, I've posted one in the dev channel on discord.

_Note: This PR has been split in 3 commits._
_First one is the functional changes._
_Second on is for files with only changes from IColony to Colony._
_Last one is for the tests._


chosta:
Love it!

kberg:
Anyone looking to pick this one up? I'll review it over the weekend if there are no other takers, but I'd rather someone more comfortable with Colonies jumped in.

bafolts:
I am not familiar with the colonies rules. This fell off my radar so I didn't take a look. Sounds good to me if @kberg takes a look this weekend.

Lynesth:
How about you @nwai90 ? :)

src/colonies/Colony.ts
```diff
@@ -1,91 +1,299 @@
...
+
+        if (steps === 0 || this.tradeType === ColonyBenefitType.GAIN_VP) {
+            // Don't increase
+            this.handleTrade(player, game, usesTradeFleet, decreaseTrackAfterTrade);
+            return;
         }
-        steps = Math.min(steps, MAX_COLONY_TRACK_POSITION - colony.trackPosition);
-        if (steps <= 0) {
+
+        if ((this.tradeType !== ColonyBenefitType.GAIN_PRODUCTION && this.tradeType !== ColonyBenefitType.STEAL_RESOURCES)
```
nwai90, line 99: This logic seems a bit convoluted. There are currently only two cases where the player should be prompted to choose whether to increase colony track.

It would likely be better to keep that logic and explicitly list the colonies where the player needs to make a choice imo. Tying it to `ColonyBenefitType` is error-prone if future colonies share the same `ColonyBenefitType` but do not need to prompt the player to choose.

Lynesth: 3 cases actually (Europa, Mercury and Hygiea) but I get your point and you're right, tying it to the colony name is a better idea.

src/colonies/Colony.ts
```diff
@@ -1,91 +1,299 @@
...
+                                game.defer(new DiscardCards(selectedPlayer, game, 1, "Hygiea colony effect. Select a card to discard"));
+                                return undefined;
+                            }
+                        );
+                    }
+                );
+                break;
+
+            case ColonyBenefitType.PLACE_OCEAN_TILE:
+                action = new PlaceOceanTile(player, game, 'Select ocean space for colony');
```
nwai90, line 277: Minor: Title here is changed from the original

src/colonies/Colony.ts
```diff
@@ -1,91 +1,299 @@
...
+                break
+
+            case ColonyBenefitType.STEAL_RESOURCES:
+                action = new StealResources(player, game, resource!, quantity);
```
nwai90, line 281: Using bang operator for `resource!` may not be ideal

src/colonies/Colony.ts
Lynesth: Changed it to use a new colony property. I think it's better has we don't have to maintain a list or something.

src/colonies/Colony.ts
Lynesth: Fixed.

src/colonies/Colony.ts
Lynesth: You are correct. I removed them and instead threw an error if resource in undefined at that point since it has no right to be.
I could have just went with
```ts
if (resource !== undefined) {
    ....
}
```
But it would allow the game to keep going, while this will at least stop everything and let us know something went wrong.

kberg: The problem @bafolts has been reminding me regularly is that ! unmanaged can shut down the server. While I think this can be mitigated by good use of try/catch (and could be wrong) I defer to him.

Lynesth: I changed that anyway so it should be fine :)

kberg:
Huh, this looks neglected. I'll do a review once you resolve merges.

src/cards/community/Hygiea.ts
```diff
@@ -1,79 +1,25 @@
...
-    }
-
-    public giveTradeBonus(player: Player): undefined | PlayerInput {
-      player.megaCredits += 3;
-      return undefined;
-    }
+    public description = 'Attack';
+    public buildType = ColonyBenefitType.OPPONENT_DISCARD;
+    public tradeType = ColonyBenefitType.STEAL_RESOURCES;
+    public tradeQuantity = [3, 3, 3, 3, 3, 3, 3];
```
kberg, line 11: This will do for now, of course. Thinking about how to combined the quantity and resource into a single unit:

```
class Trades {
    public trades: Array<ITrade> = [];

    public static builder(): Builder {
        return new Builder();
    }
}

class Builder {
    public trades: Trades = new Trades();

    public of(resource: Resources, qty: number): Builder {
        const t = new Trades();
        t.trades.push({ resource: resource, qty: qty });
        return this;
    }

    public build() : Trades {
        return this.trades;
    }
}
export const tds: Trades = Trades.builder()
    .of(Resources.MEGACREDITS, 3)
    .of(Resources.MEGACREDITS, 3)
    .build();
```

kberg: Or perhaps a more general term: `ColonyReward`?

kberg: Update: Now that I see what you're doing in `Colony`, that sometimes the resource type is constant throughout, my example is insufficient. But I still find it slightly disconcerting that the track resource type and resource value are stored separately.

src/colonies/Colony.ts
```diff
@@ -1,91 +1,310 @@
...
+    public giveColonyBonus(player: Player, game: Game, isGiveColonyBonus: boolean = false): undefined | PlayerInput {
+      return this.giveBonus(player, game, this.colonyBonusType!, this.colonyBonusQuantity!, this.colonyBonusResource, isGiveColonyBonus);
+    }
 
-      // Poseidon hook
-      const poseidon = game.getPlayers().filter((player) => player.isCorporation(CardName.POSEIDON));
-      if (poseidon.length > 0) {
-        poseidon[0].addProduction(Resources.MEGACREDITS);
+
+    private giveBonus(player: Player, game: Game, bonusType: ColonyBenefitType, quantity: number, resource: Resources | undefined, isGiveColonyBonus: boolean = false): undefined | PlayerInput {
```
kberg, line 144: What does isGiveColonyBonus do? Why is it only used to determine whether to defer the action or execute it now?

kberg (approved):
As best as I can tell this is OK to merge. I like the work you're doing here to eliminate repetition. But please address my question about that particular field inside `Colony`

src/cards/community/Hygiea.ts
Lynesth: Yes I get what you mean. But since all colony tiles have consistent resource except for 3 of them I thought it would be better that way to prevent making things unnecessarily too complex.

src/colonies/Colony.ts
Lynesth: It's used when that method is called from the `GiveColonyBonus` deferredAction (that manages simultaneous colony bonus for all players, like everyone discarding at the same time for pluto).
If that's the case, we want the colony bonus be executed directly, since the `GiveColonyBonus` deferredAction works in its own little loop and doesn't rely on the deferred actions queue.
Want me to add a comment about this ?

src/colonies/ColonyBenefitType.ts
```diff
@@ -0,0 +1,20 @@
+export enum ColonyBenefitType {
```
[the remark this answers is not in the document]
Lynesth: Yeah, I think I'll change it.

src/cards/community/Hygiea.ts
kberg: Gotcha.

src/colonies/Colony.ts
kberg: If you think this is obvious to people expert with Colonies, then it's probably fine. Then again, you didn't get a Colonies expert, you got me. So, yeah, something small, please? :D

kberg:
I'll merge this tonight with a small comment pls.

kberg (approved):

PR #1821, opened 2020-11-03
Vue unit tests

bafolts:
This change will use actual vue unit test tools instead of the rudimentary ones I had in place.

https://vue-test-utils.vuejs.org/

This will allow us to examine what happens with the components during events and state changes.


tests/components/TagCount.spec.ts
```diff
@@ -1,35 +1,32 @@
...
+        return localVue;
+    }
+    it("renders with no count", function () {
+        const tagCount = mount(TagCount, {
+            localVue: getLocalVue(),
+            propsData: {
+                count: 0
+            }
+        });
+        expect(tagCount.find("div[class=\"tag-display tag-no-show\"]").exists()).to.eq(true);
```
kberg, line 19: s/to.eq(true)/is.true

kberg, line 19: Good for now, let's consider https://www.chaijs.com/plugins/chai-dom/ later? Perhaps that's one I can dop.

kberg (approved):

kberg:
This is great!

tests/components/TagCount.spec.ts
bafolts: The heavy lifting with getting jsdom in place and the vue-test-utils running is in place with this change. Should be easy for others to drop in updates later. I am not sure `chai-dom` would work. The `Wrapper` exposes these `find` methods.

https://vue-test-utils.vuejs.org/api/wrapper/#find

As `find` is deprecated I will need to update these in some way to leverage `findComponent`.

PR #1825, opened 2020-11-04
Fix Research Outpost and Hazard interaction

### LINKED ISSUE #1823: [Ares] Bug - placing tile with no placement possibilities
Hi,
during my game I played Research Outpost card. Only places without tiles next to it were Dust Storms, which require 8 / 16 $ to place there. However I do not have that much money. Currently I'm stuck inside this action.
Here is this my turn: https://terraforming-mars.herokuapp.com/player?id=a862fa2596d2

Here are some screenshots (im not sure if link above will work):
Screenshot from 2020-11-04 00-25-24
Screenshot from 2020-11-04 00-25-34

Thanks for your awesome work!

Lynesth:
@kberg you probably want to take a look at this. It is a quick fix and could probably be better handled but at least is should do it for now.
Maybe all cards with specific conditions like this should be checked with hazards in mind ?

Fixes #1823

### DIFF
--- a/src/cards/ResearchOutpost.ts
+++ b/src/cards/ResearchOutpost.ts
@@ -8,6 +8,7 @@ import { SelectSpace } from "../inputs/SelectSpace";
 import { ISpace } from "../ISpace";
 import { PlayerInput } from "../PlayerInput";
 import { CardName } from '../CardName';
+import { AresHandler, HazardSeverity } from "../ares/AresHandler";
 
 export class ResearchOutpost implements IProjectCard {
     public cost = 18;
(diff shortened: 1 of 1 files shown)


kberg:
So, two things:
1. This could be fixed with `getAvailableSpacesOnLand()`. But keep in mind similar costs apply for adjacency. If `getAvailableSpacesOnLand` only applies to placing tiles, that's great, but if not, I would instead suggest creatiing `getAffordableLandSpaces()` which lays bare the placement costs.

This gets more complicated when trying to deal with, for instance, placing next to `Nuclear Zone` (which has a 2MC cost) when you have 0 MC.

If you place next to an ocean, it's obvious one offsets the others' costs. 
What if instead you also placed next to Natural Preserve, which only grants one MC? But if you own the Natural Preserve, then you get an additional MC!
What if you weren't adjacent to a MC-earning space, but, say, `Mining Area`, but you own the space, *and* have `Marketing Experts`. Then you get 2MC for the placement.

Those things all would need to be taken into account.

I wonder if instead it makes more sense to make it easy to 1) cancel the entire action, or 2) make Undo the default.

Lynesth:
@kberg I agree that this isn't a global fix and that doing something that works properly is going to be hard anyway as we've previously discussed. It just fixes Research Outpost, which city cannot be placed near other tiles so adjacency doesn't matter in this case, only the fact that it can be placed on a hazard.

kberg:
> @kberg I agree that this isn't a global fix and that doing something that works properly is going to be hard anyway as we've previously discussed. It just fixes Research Outpost, which city cannot be placed near other tiles so adjacency doesn't matter in this case, only the fact that it can be placed on a hazard.

I have a more global fix coming.

Lynesth:
I'm working on something as well. It'll be good to discuss it later then :)

kberg:
How do you feel about at least abandoning this change?

Lynesth:
I don't mind at all, feel free to close it :)

Issue #1830, opened 2020-11-04
Bug: Turmoil event `AQUIFER RELEASED BY PUBLIC COUNCIL`


vsrisuknimit:
AQUIFER RELEASED BY PUBLIC COUNCIL asks first player place an ocean.

This ocean is currently placed as if done by the player. So the player gains both TR and the placement bonus. This is wrong. The placement should be like WGT ocean placement.

![image](https://user-images.githubusercontent.com/14239220/98172066-b38e5080-1ebe-11eb-8c63-cb0d41f50973.png)

<img width="1037" alt="Screen Shot 2020-11-04 at 3 53 20 PM" src="https://user-images.githubusercontent.com/14239220/98172253-05cf7180-1ebf-11eb-9441-1e6bddc0ab1c.png">

<img width="491" alt="Screen Shot 2020-11-04 at 3 53 40 PM" src="https://user-images.githubusercontent.com/14239220/98172261-08ca6200-1ebf-11eb-817d-49c6337ed387.png">

kevinb9n:
It's weird that I can't find anything that officially states this rule.
But it seems like it just *has* to be what the designer intended.

vsrisuknimit:
@kevinb9n Turmoil rulebook.

![image](https://user-images.githubusercontent.com/14239220/98192369-e64d3e80-1ee8-11eb-919b-df4ae5cb56e7.png)

kevinb9n:
Ah jeez, I was just one section down from there, where they put all the other special cases.
I suppose they might have intended this to apply to more than one card, originally...

PR #1831, opened 2020-11-05
Fix: Ocean from global event should not give TR or placement bonus

nwai90:
Fix from @Lynesth that closes https://github.com/bafolts/terraforming-mars/issues/1830. Game phase should always be set to `Phase.SOLAR` even if solar phase option is not selected or global parameters are maxed, as a lot of core logic checks for this specific phase.

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -822,9 +822,10 @@ export class Game implements ILoadable<SerializedGame, Game> {
         }
         return;
       }
+      
       // solar Phase Option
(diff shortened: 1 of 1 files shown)


alrusdi (approved):

Lynesth:
As I mentioned, we should probably change `gameOptions.solarPhaseOption` to something like `gameOptions.useWGT` since that's what it's actually used for. It would most likely break all ongoing games though.

Issue #1240, opened 2020-08-31
Confirm Tile Placements


B0Bftl:
When Placing a Tile, a click or tap on the Map instantly places the Tile. In Combination with the automatic selection of "Place Greenery" if you have sufficient plants, this is a constant source for misclicks. Happened to me multiple Times :( 

It would be a lot user friendlier, if you had to confirm your Placement. It would also be more consistent with other actions, as you have to confirm everything else.

My proposed UI: After Selecting a space, the placed tile flickers on the selected spot. Below the map, you can press "Confirm". Until Confirm is pressed, you can reselect any other (valid) space.

alrusdi:
If someone will do this - please make it optional (via Preferences). It can slowdown the game.

K-Hartmann:
The Problem ist paticulary frequent wenn using a mobile device, scince scrolling and zooming requires clicks on the map, but without scrolling it is hard to precisely click on the rather small map.

As an additional idea, if the placement of greenerys would not be the default option whenever you have sufficient plants, it would at least be required to actively choose to place a greenery. This would remove the annoying situation where you scroll down from the top of the page, and accidently place a greenery bevor even realising that the option is active.

kevinb9n:
My wife got hit with that yesterday (the auto-selection of the greenery option). She had, or believed she had, chosen to place a city, but then inadvertently clicked somewhere else (?) before clicking the map and suddenly a greenery was there.

VanderLeal:
I would suggest that instead of making this "mandatory OR optional" to use a processual roll-back to be available. 
This would cover both bases.

Example: 
1. You pick the action.
2. You're asked to place a tile.
3. You place the tile (by choice or by accident). 
4. An "Undo Tile Placement" button shows up (as the undo turn). 
5a. If you choose any other action then you move on (go to step 6). 
5b. If you click Undo Tile Placement, you're asked to place the tile again (go to step 2).
6. End

B0Bftl:
An 'Undo' Method is impracticle, as there are Card draw bonuses. Once you've drawn the card, undoing is quite like cheating.

I think a simple confirmation is the cleanest solution. Its only downside is one additional click, which imo does no harm. It rather increases intuitivity and robustness

kberg:
I agree, optional confirmation seems the way to go.

VanderLeal:
[...]

Issue #953, opened 2020-07-08
UX: Actions menus rework


ssimeonoff:
This is a combination of some old ideas and the recent comment: 
https://github.com/bafolts/terraforming-mars/issues/943#issuecomment-654776183 , made me revise them. 

# Idea 1 - Separate the options and sub-options
Currently when a option is selected (e.g. play a card or sell a card), the sub-options are displayed right under the selected option. 

# Idea 2 - Display the options inline
Example: 
![top-menu](https://user-images.githubusercontent.com/6917565/86918431-fe71a480-c126-11ea-9384-b73c0fc51014.png)

# Idea 3 - Define the options and sort them by popularity
See the example above. 

a. Player Project Card - same as no

b. Activate Card - same as now 

c. Spend Resources - this is the interesting one. It aggregates all special moves. 
When clicked, the displayed sub-options can look like: 
![resources](https://user-images.githubusercontent.com/6917565/86918927-abe4b800-c127-11ea-8aa4-5cb5cd410129.png)
The sub-options are displayed inline, grouped by type  and sorted by popularity. 
The sub-options can also be styled slightly different (e.g. forest is green, temp is red, milestones and awards are in their lightblue colour, etc)
side-note: maybe on the row with the Trade buttons, at the end a label "X trading fleet remaining" can be displayed as well

d. Standard Projects - same as now, but displayed inline

e. End Turn or Pass - "End Turn" is 1st option and preselected (if available). Next to "Pass" a label can be displayed explaining that this is the end of the generation for the player. (maybe even make the button red)

f. Sell Cards - same as now

Top options and sub-options can have different active state design. 
Example: 
![button-active](https://user-images.githubusercontent.com/6917565/86919290-4a711900-c128-11ea-9b7e-70af8c562cbf.png)
(sub-options are styled similar to the cards)

#Idea 4 - After clicking "Save", open the last selected option.
Currently, when "Save" is clicked, the page always shows "Play project cards" selected.
This is bad UX, but with the ideas above it will get worse (more clicks will be needed), so it will be important to remember the selected top option upon submit.

Basically the top options to behave as a browser's tabs and aligned as such (maybe even can be styled as such).
When a top option has no available sub-options, currently it is not displayed at all, but also it can be displayed in a disabled state, if needed. (no option here)

jijiqw:
Wonderful!  I love these change so much , can't wait to see it !

alrusdi:
[...]

Issue #497, opened 2020-03-29
Feature Request: Cancel before placing a tile.


FreeLikeGNU:
Choosing an action that places a tile does not allow cancelling that action before the tile is placed.

kberg:
The undo feature ought to handle this. Close?

FreeLikeGNU:
> The undo feature ought to handle this. Close?
If I choose an action that places a tile can I undo if its the second action in a multiplayer game?

kberg:
I'm OK with undo being the sole mechanism to address this, both effectively achieve the same thing. But I think it's probably hard to do, given that some tile placement produces cards and other events that can't be unseen. But if undo is good enough, then there's no need for both -- people who want to take a cautious approach to their tiles being correctly placed can enable it.

kevinb9n:
@kberg I addressed the "unseen" bit in the new issue just linked, but setting that aside, I think undo is an unsatisfactory way to handle this. First because it's often unavailable, as mentioned. But also because it forces the player to continue making random choices to complete the action - and those choices could trigger things like an opponent's Philares effect.

A "cancel" you can do before anything committal ever happened would be a great thing.

kberg:
Yeah I'm beginning to come around to this cancel before placing a tile idea. But there are some die-hard players who care about minimum clicks.

nwai90:
@kberg Do you think we can close this one? As you have added a tile confirmation dialog now

kberg:
Yep, I was just coming here to do that.

kberg:
@FreeLikeGNU FYI, this is implemented.

Lynesth:
No it's not. This issue is about being able to cancel the action selected if you still haven't placed the tile.

kberg:
Good catch.

bafolts:
I am going through old issues assigned to me. Hoping to not have any open for a year! If the `SelectSpace` has an `OrOptions` or `AndOptions` parent then this seems plausible. If the `SelectSpace` has no parent from a deferred action than it wouldn't make sense to be able to cancel as you have to select a space. Good opportunity to upgrade `SelectSpace` again. Plan to pick this up next.

Lynesth:
[...]

PR #1839, opened 2020-11-06
Confirm plants conversion before placing Greenery

Lynesth:
This makes the behavior more consistent. You have to validate before placing a tile when you select the Aquifer/Greenery/City standard projects. It also alleviates the issue described here: https://github.com/bafolts/terraforming-mars/issues/1240


nwai90:
@Lynesth Can you make it an option that is off by default? This slows down the game

Lynesth:
There is no communication between in-browser player preferences and the server, so that is not an option at the moment, no.

kberg:
It's such a weird case, that's for sure.

Lynesth:
I get that different people, depending on their experience with the app, would most likely prefer one thing or the other.
It's just that it's the only action that actually validates when clicking the map, instead of having to click a button like every other.

I often find myself pondering, then I pick "Convert plants" and click "Save". The app tells me I need to select a space so I scroll up and click the map. The save button is actually useless but if you removed it then it wouldn't make sense when selecting that option.

Anyway, just my opinion on this. I'm sure if we take a bit of time we can use the preferences server side as well but that's something I guess needs more discussion.

kberg:
A data point: if this was something that impacted several minutes per game, I'd be more concerend. But given that this happens at ~20 times in a game around 95% of the time, compared to the many, many more clicks, I'm not quite suire this is the the hill to die on for performance over consistency.

But I can live with it. Also, https://xkcd.com/1172/

alrusdi:
I think we need more general solution to confirm all the tile placements and it MUST be optional

Lynesth:
Just to make it clear, it is not a confirmation of tile placement, like it was proposed in the linked issue.
It's just 1 click to say: "The action I want to take is `Convert plants to greenery`" and then place the tile.

kberg:
What's blocking this from proceeding?

alrusdi:
> What's blocking this from proceeding?

At least two of us (me and @nwai90) want it to be optional. If it can't be optional - we should reject this PR

Lynesth:
> If it can't be optional - we should reject this PR

Well there's currently no way to make this optional as I mentioned.

But I don't understand, it is the only action that doesn't require a click. Everything else does.
Are you two saying that the fact that you have to click to confirm you want to use an Aquifer/Greenery/City standard project instead of just placing the tile directly slows the game ? (genuine question)

alrusdi:
I think a lot of confirmations really slows the game and they are very annoying. 
What a problem to pass user preferences to the server btw? It can be a great start to persist them in DB later.

Lynesth:
> I think a lot of confirmations really slows the game and they are very annoying.

If actions were buttons from the start (like what @ssimeonoff proposed here https://github.com/bafolts/terraforming-mars/issues/953) then there wouldn't be a problem since eveything would be consistent.
The issue at the moment is that we have a radio menu with submenus and that the only one that doesn't behave like the others is the plant conversion. I mean, when it's your turn, you don't even have to go down to the actions, you can click on the map to put a greenery (issue raised in #1240) and that's seems pretty inconsistent to me.

> What a problem to pass user preferences to the server btw?

There's no problem with that, and it should be done at some point, I agree. It's just that it isn't possible at the moment and that's why it's not a possibility for the current PR.

alrusdi:
If we have an option to fix that issue by action button redesign (and we will do it at some point) then why so urgent to fix the greenery placement right now?

kberg:
> I think a lot of confirmations really slows the game and they are very annoying.

So @alrusdi I'm not actually going to try to say that this change should go in, but I also think your reasons for keeping this out are not that hard to dispute. Given that the number of greeneries placed in a game, as a ration of all the clicks and actions, are so small that this isn't a meanignful time change.

Greenery placement is the only specialty behavior that optimizes out the step where I have to tell the app that I'm placing a greenery tile. As an outlier, I personally found it difficult to understand, only until after about 20-30 games. (I played a LOT of ground solo games with Ares, as you can imagine.)

I don't think anyone can disagree with "a lot of conformations really slows the game and they are very annoying" as a general complaint, but relatively speaking I just don't think this is that. So I assume you really mean that you want to avoid death-by-a-thousand-cuts.

But this specialty behavior, well, it's _distinct_ and _surprising._ And as UX goes, distinct and surprising isn't good.

But I am not here to set a new direction for the game's UX.

alrusdi:
I'm not against it if it will be optional

kberg:
I'm pleased to read that. Also, I'm unhappy because now I have to go read the details about why it can't be optional. Baby steps, I suppose. :D

kberg:
OK, so: what would it take to actually support this?  Let's start with an awful hack and call it the worst suggestion that works.

### The worst suggestion so far
* Add `GetUserPreferences extends PlayerInput`, which returns a dictionary of the player's preferences without actually requesting input.
* Use `GetUserPreferences` in `convertPlantsIntoGreenery`.
* The callback will evaluate whether to return a `SelectSpace` or `SelectOption` `PlayerInput` instance.

alrusdi:
We can add an endpoint like /sync-preferences and save current user preferences in the Player instance. It can be done at the start of the game. That preferences can be used to check if confirmation of the greenery placement needed. Later we can persist preferences in DB with player account

kberg:
And by syncing if decide part-way through the game to toggle that preference it'll update it? Yeah, perhaps.

What if it didn't actually persist it? Just keep it in memory, and if the instance restarts it asks for a new ... no, that won't work. ... Hmm... would the server-side be the definitive place for user preferences?

alrusdi:
> Hmm... would the server-side be the definitive place for user preferences?
It should. But we don't have an account/profile entity yet to store preferences with it.

kberg:
> > Hmm... would the server-side be the definitive place for user preferences?
> 
> It should. But we don't have an account/profile entity yet to store preferences with it.
> Client can switch devices/browser, but preferences should remain the same (some of them at least :-) )

Since there's no plans for user login, let's go back to the original suggestion which fetches options upon request. Any strong objections to making it work that way?

bafolts:
I think the suggestion in #1240 makes more sense than having to click twice prior to placing the tile altogether. The issue in #1240 would still remain but the confirmation would happen prior to `SelectSpace`. Updating `SelectSpace` to require clicking a button prior to selecting the space should be possible. We could have double click work as a stop gap to place the tile without require finding the button to click to confirm. Upon clicking to "Convert Plants" we could scroll up and bring the board into focus as you will need to click there. I would consider how `SelectSpace` would work best versus attempting to make the current `SelectSpace` less confusing. Scrolling the user up to the board could go a long way for the time being. It should lead to a faster game and be same number of clicks while avoiding confusion around needing to select a space on the board.

bafolts:
How about, when `SelectSpace` is triggered we bring the board into focus for mobile users or users on smaller screens. At this point you have to select a space. Upon clicking once the tile is placed on the board. Next to the tile is a tooltip like "Save". If you click again you can move the tile location and the "Save" button is still right by the tile. If you double click instead of click the tile is placed as before with no confirmation.

kberg:
To me, @bafolts the issue is that in order to Do Anything, you have to start with the OrOptions widget, _except_ place greenery. Eliminating that outlier, in my opinon, is preferable. I wouldn't have any issue with *two* confirmations. But that's me - I'm not playing Speed Terraforming Mars. :D

nwai90:
> How about, when `SelectSpace` is triggered we bring the board into focus for mobile users or users on smaller screens. At this point you have to select a space. Upon clicking once the tile is placed on the board. Next to the tile is a tooltip like "Save". If you click again you can move the tile location and the "Save" button is still right by the tile. If you double click instead of click the tile is placed as before with no confirmation.

The current implementation favours convenience in that `SelectSpace` is immediately triggered if the player had enough plants to place a greenery, without the need for any additional clicks beyond placing the greenery tile on the board. The above approach could work and is the gist of https://github.com/bafolts/terraforming-mars/issues/497. Ideal scenario would be to eventually introduce an option to enable/disable these in-game confirmations so that they are optional for the user as @alrusdi has mentioned.

bafolts:
Ah, I now understand what this solution is going for within `OrOptions`. For what it's worth. The current `SelectSpace` happened to be the easiest to implement when I was writing this way back when. If I assume the speed players are playing single player games could consider only enabling this for games with more than one player. This wouldn't interrupt the single player games and would avoid requiring an option.

Lynesth:
I can understand that some of you prefer the way it is currently implemented but it seems you don't think this is bad UX.
Let's try to be impartial here and realize that:
- That action is the only one that doesn't behave like the others
- It presents a `Save` button that does nothing
- It gives the player the risk to make an error by simply clicking/tapping on the map if that option is selected

Also I want to make clear that I am not in favour of a "confirm all tile placement" system, except if that's optional.
I am just talking about reducing errors while making available actions consistent.

Why not make a poll on discord to collect players' opinion on this ?

chosta:
> * It presents a `Save` button that does nothing

This is very confusing I must admit.
I feel that this conversation is not making progress as it's obvious there are different opinions and not a single correct solution to the problem.
My take on this is -> keep the current implementation -> get rid of the `Save` button (if possible) -> think of a better general UX solution (me and @ssimeonoff can work on that)

kberg:
The error message that accompanies the Save button failure is the very
thing that allowed me to know how to place a greenery in the first place.
Please keep it until a better change is in place.

chosta:
I was thinking about this as I got burned the other day by not placing on the corect hex the other day. It was on my mobile phone and you to be very nifty to perform this operation. I failed ofc and probably lost the game. I also understand why we would not want it when playing on desktop and using timers etc. So for me the solution should be: Add an option to have a warning / or not and let the players decide. On mobile I would 100% use it and on desktop 100% not.

PR #1840, opened 2020-11-06
Reformat imports for src

kberg:


kberg:
This is split in two because it seems the files from both src/ and tests/ were too much for the diff view.

I did see a single quotes change in the source, I hope that's not a problem.

```
find src -name "*.ts" | xargs sed -i "s/from '\(.*\)'/from \"\1\"/"
find src -name "*.ts" | xargs sed -i "s/import {\([^ ]*\)}/import { \1 }/"
```

alrusdi (approved):

alrusdi:
Love. May be we need go further with gts? https://github.com/google/gts

kberg:
Thanks for the quick review.

> gts is Google's TypeScript style guide, and the configuration for our formatter, linter, and automatic code fixer. No lint rules to edit, no configuration to update, no more bike shedding over syntax.

That sounds good to me. It comes down to doing a global reformat as a completely separate PR. There's an open PR about reformatting IIRC, add that reference there?

PR #1842, opened 2020-11-06
Fix interesting outlier where the only available land spaces are hazards. And you're broke.

### LINKED ISSUE #1823: [Ares] Bug - placing tile with no placement possibilities

kberg:
Is this enough to c.l.o.s.e #1823?

### DIFF
--- a/src/Board.ts
+++ b/src/Board.ts
@@ -143,16 +143,53 @@ export abstract class Board {
             );
     }
 
-    public getAvailableSpacesOnLand(player?: Player): Array<ISpace> {
+    public getAvailableSpacesOnLand(player?: Player, checkAffordability = true): Array<ISpace> {
         const landSpaces = this.getSpaces(SpaceType.LAND, player).filter(space => {
             const hasPlayerMarker = space.player !== undefined;
             // A space is available if it doesn't have a player marker on it or it belongs to |player|
             const safeForPlayer = !hasPlayerMarker || space.player === player;
-            // And also, if it doesn't have a tile. Unless it's a hazard tile. 
-            const playableSpace = space.tile === undefined || AresHandler.hasHazardTile(space) 
-            // If it does have a hazard tile, make sure it's not a protected one.
-            const blockedByDesperateMeasures = space.tile?.protectedHazard === true;
-            return safeForPlayer && playableSpace && !blockedByDesperateMeasures;
+            if (safeForPlayer === false) {
+                return false;
+            }
+
+            // The space either a) has no tile, b) has a regualar tile, or c) has a hazard tile.
+
+            // a) the space has no tile.
+            if (space.tile === undefined) {
+                return true;
+            }
+
+            // b) the space has a regular tile.
+            if (!AresHandler.hasHazardTile(space)) {
+                return false;
+            }
+
+            // c) the space has a hazard tile.
+            if (AresHandler.hasHazardTile(space)) {
+                // Hazard tiles can be covered if they're not protected ... 
+                const blockedByDesperateMeasures = space.tile?.protectedHazard === true;
+                // and the player can pay the coverage cost
+                const canAfford = (checkAffordability && 
+                    (player === undefined || player.canAfford(AresHandler.hazardCost(space))));
+                return canAfford && !blockedByDesperateMeasures;
+            } else {
+                return false;
+            }
+
+            // A note about paying additional costs to place a tile. The test for canAfford could actually get much
+            // more complicated.
+            //
+            // As an example, having 0 MC and placing a tile next to Nuclear Zone (which has a 2MC adjacency cost).
+            // Things you would have to consider:
+            // * Are you also placing next to an ocean?
+            // * Are you placing next to Natural Preserve, which only grants one MC? But if you own
+            //   the Natural Preserve, then you get an additional MC.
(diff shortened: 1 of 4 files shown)


src/Board.ts
```diff
@@ -143,16 +143,53 @@ export abstract class Board {
...
+                return true;
+            }
+
+            // b) the space has a regular tile.
+            if (!AresHandler.hasHazardTile(space)) {
+                return false;
+            }
+
+            // c) the space has a hazard tile.
+            if (AresHandler.hasHazardTile(space)) {
```
Lynesth, line 168: Can it be anything else at this point ?

src/Board.ts
```diff
@@ -143,16 +143,53 @@ export abstract class Board {
...
+                // Hazard tiles can be covered if they're not protected ... 
+                const blockedByDesperateMeasures = space.tile?.protectedHazard === true;
```
Lynesth, line 170: You can probably leave the `?` out since `space.tile` cannot be undefined at this point.

src/Board.ts
```diff
@@ -143,16 +143,53 @@ export abstract class Board {
             );
     }
 
-    public getAvailableSpacesOnLand(player?: Player): Array<ISpace> {
+    public getAvailableSpacesOnLand(player?: Player, checkAffordability = true): Array<ISpace> {
```
Lynesth, line 146: Set `checkAffordability` type to boolean.
Also, when is `getAvailableSpacesOnLand` ever called with `false` as a second argument ? Does this need to be an argument ?

src/Board.ts
```diff
@@ -143,16 +143,53 @@ export abstract class Board {
...
+            if (!AresHandler.hasHazardTile(space)) {
+                return false;
+            }
+
+            // c) the space has a hazard tile.
+            if (AresHandler.hasHazardTile(space)) {
+                // Hazard tiles can be covered if they're not protected ... 
+                const blockedByDesperateMeasures = space.tile?.protectedHazard === true;
+                // and the player can pay the coverage cost
+                const canAfford = (checkAffordability &&
```
Lynesth, line 172: So if `checkAffordability` is set to false, we can't afford it ?

Lynesth:
Should have grouped those comments into a review, sorry about that.

src/Board.ts
kberg: Could be an ocean on a land space.

src/Board.ts
kberg: Oh good! Fixed.

src/Board.ts
kberg: Whoops! Changed that. I'll add a test. :D Not that I honestly think anyone will call that will checkAffordability === false.

kberg: Perhaps if it's the neutral player?

src/Board.ts
Lynesth: That would be caught by the previous if block
```ts
             // b) the space has a regular tile. 
             if (!AresHandler.hasHazardTile(space)) { 
```

src/Board.ts
```diff
@@ -167,9 +167,9 @@ export abstract class Board {
             // c) the space has a hazard tile.
             if (AresHandler.hasHazardTile(space)) {
                 // Hazard tiles can be covered if they're not protected ... 
-                const blockedByDesperateMeasures = space.tile?.protectedHazard === true;
+                const blockedByDesperateMeasures = space.tile.protectedHazard === true;
                 // and the player can pay the coverage cost
-                const canAfford = (checkAffordability && 
+                const canAfford = (!checkAffordability ||
                     (player === undefined || player.canAfford(AresHandler.hazardCost(space))));
```
Lynesth, line 173: You can remove extra parenthesis not needed anymore here.

Lynesth:
Actually, this PR does not take into account what the player must pay for the card/action as well and will return false positives and errors.

Example using Research Outpost (costs 18 MC). Let's say the player has 20 MC and all available spaces for the city are hazards.
Here's the card's canPlay method:
```ts
        return this.getAvailableSpaces(player, game).length > 0;
    }
```
This will return true but once the card is played (and the player now only has 2 MC), it will throw an error has it will pass an empty spaces list to the SelectSpace input.

kberg:
> Actually, this PR does not take into account what the player must pay for the card/action as well and will return false positives and errors.
> 
> Example using Research Outpost (costs 18 MC). Let's say the player has 20 MC and all available spaces for the city are hazards.
> Here's the card's canPlay method:
> 
> ```ts
>     public canPlay(player: Player, game: Game): boolean {
>         return this.getAvailableSpaces(player, game).length > 0;
>     }
> ```
> 
> This will return true but once the card is played (and the player now only has 2 MC), it will throw an error has it will pass an empty spaces list to the SelectSpace input.

That's a good observation. Damn.

You said you had a larger fix?

PR #1850, opened 2020-11-06
More type checking on vue components

bafolts:
This adds more type checking to the vue components. This time I had to make changes as we were passing in `undefined` instead of `string` when calling `onSave`. It seems we attempted to work around this on the server but I modified this to not pass in `undefined`. Also spotted two places we were still relying on `PlayerId` which needs to check `Color`.

### DIFF
--- a/src/components/SelectCard.ts
+++ b/src/components/SelectCard.ts
@@ -1,9 +1,10 @@
 
 import Vue from "vue";
 import { Button } from "../components/common/Button";
+import { VueModelCheckbox, VueModelRadio } from "./VueTypes";
 
 interface SelectCardModel {
-    cards: Array<CardModel> | CardModel | boolean;
+    cards: VueModelRadio<CardModel> | VueModelCheckbox<Array<CardModel>>;
 }
 
 import { Card } from "./card/Card";
--- a/src/components/SelectCard.ts
+++ b/src/components/SelectCard.ts
@@ -3,14 +3,28 @@ import Vue from "vue";
 import { Button } from "../components/common/Button";
 
 interface SelectCardModel {
-    cards: Array<CardModel>;
+    cards: Array<CardModel> | CardModel | boolean;
 }
 
 import { Card } from "./card/Card";
 import { CardModel } from "../models/CardModel";
+import { PlayerInputModel } from "../models/PlayerInputModel";
 
 export const SelectCard = Vue.component("select-card", { 
-    props: ["playerinput", "onsave", "showsave", "showtitle"],
+    props: {
+        playerinput: {
+            type: Object as () => PlayerInputModel
+        },
+        onsave: {
+            type: Object as () => (out: Array<Array<string>>) => void
+        },
+        showsave: {
+            type: Boolean
+        },
+        showtitle: {
+            type: Boolean
+        }
+    },
     data: function () {
         return {
             cards: []
@@ -18,7 +32,7 @@ export const SelectCard = Vue.component("select-card", {
     },
     components: {
         Card,
-        "Button": Button
+        Button
     },
     methods: {
         saveData: function () {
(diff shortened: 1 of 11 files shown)


src/components/SelectPlayer.ts
```diff
@@ -1,30 +1,54 @@
...
+            if (this.selectedPlayer !== undefined) {
+                result[0].push(this.selectedPlayer);
+            }
+            this.onsave(result);
         }
     },
     template: `<div>
   <div v-if="showtitle === true">{{playerinput.title}}</div>
-  <label v-for="player in playerinput.players" :key="player" class="form-radio form-inline">
+  <label v-for="player in (playerinput.players === undefined ? [] : playerinput.players)" :key="player" class="form-radio form-inline">
```
kberg, line 48: A noob question, why is this construct better than, say, playerinput.players || []?

src/components/SelectPlayerRow.ts
```diff
@@ -3,13 +3,16 @@ import Vue from "vue";
...
+    methods: {
+    },
     data: function () {
         return {};
     },
-    render: function (createElement) {
-        const player: PlayerModel = this.player as PlayerModel;
-        return createElement("span", { domProps: { innerHTML: player.name } });
-    }
+    template: `<span>{{player === undefined ? "" : player.name}}</span>`
```
kberg, line 16: I think about that || comment again here, where this could (possibly?) be more conclsely written as `player?.name || ""`. Unless I'm missing something.

kberg (approved):

src/components/SelectCard.ts
```diff
@@ -3,22 +3,36 @@ import Vue from "vue";
 import { Button } from "../components/common/Button";
 
 interface SelectCardModel {
-    cards: Array<CardModel>;
+    cards: Array<CardModel> | CardModel | boolean;
```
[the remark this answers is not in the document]
bafolts: When using `v-model` with checkboxes vue generates a function which checks at runtime if the property is an array with `Array.isArray`. When it is an array it uses `Array<CardModel>` and updates the array. If it isn't an array vue passes in `true` or `false` when the property is checked. Here we use `CardModel` when we are using the radio button when only one card can be selected. It would be nice to assure `v-model` that since we have `Array.isArray` we don't need to consider `boolean`. Will have to see if it is doable. To answer the question yes this suggests there is a refactoring necessary. With the typing explicitly written out it is now showing off how complex the type is, can simplify in next pass.

src/components/SelectPlayer.ts
bafolts: It is not better or worse but different. I can use `playerinput.players || []` here as that is more succinct.

src/components/SelectPlayer.ts
```diff
@@ -1,30 +1,54 @@
...
+            type: Boolean
+        },
+        showtitle: {
+            type: Boolean
+        }
+    },
     data: function () {
         return {
-            selectedPlayer: undefined
+            selectedPlayer: undefined as Color | "NEUTRAL" | undefined
```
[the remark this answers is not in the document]
bafolts: I will add this type now. This was one of the reasons I wanted to add typing so we could see where we were using `Color | "NEUTRAL"` as well as other types on the client.

src/components/SelectPlayerRow.ts
bafolts: The optional chaining operator is still moving out to browsers.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

When we use `?` in typescript it is compiled for us into syntax which will run in ecmascript 6. If using it here in a browser that doesn't support optional chaining it could very well prevent the entire script from running due to syntax error.

kberg: So earlier when you called it "different" versus better or worse, this is actually a way in which || is worse. I'll be mindful of both of these, thanks.

PR #1852, opened 2020-11-07
remaining vue components typed

bafolts:
This adds type checking to the remaining vue components! Between this and the unit tests in place for vue components now in great shape.


kberg (approved):

kberg:
Huzzah!

Issue #1794, opened 2020-10-28
Stormcraft - 'Max' function for temp raise is not bounded to 8 total heat


andystager:
For temp raising using Stormcraft, when 'Max' is pressed for either floaters or heat, it defaults to the total available resource rather than a limit to sum of 8 total heat.

Maybe related to #842 ?

PR #1853, opened 2020-11-07
set maximum values for stormcraft heat to energy

bafolts:
For #1794 . Sets the maximum value to a maximum of 8 or 4. Also added error checking for over paying.


kberg (commented):
I would like to say approved, but I'm not quite sure I follow the logic changes here. A test would help a lot.

src/Player.ts
```diff
@@ -1726,19 +1726,19 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
...
-      let heatAmount: number;
-      let floaterAmount: number;
       if (this.isCorporation(CardName.STORMCRAFT_INCORPORATED) && this.getResourcesOnCorporation() > 0 ) {
+        let heatAmount: number;
+        let floaterAmount: number;
         const raiseTempOptions = new AndOptions (
           () => {
-            if (heatAmount + (floaterAmount * 2) < 8) {
+            const total = heatAmount + (floaterAmount * 2);
+            if (total < 8) {
```
nwai90, line 1738: 👍 We could also use `constants.HEAT_FOR_TEMPERATURE` here and in all places in this PR that have the hardcoded value of 8

nwai90 (approved):

src/Player.ts
bafolts: OK. I will also add test. The logic change here is that one can only spend 8 heat or 4 floaters to raise temperature. The maximum amount they can spend is then 8 or 4, regardless of how much heat or floaters on card they have.

kberg: BTW, sorry -- I should have said, I'm not dying for a test here, but clarity. I rescind any request for a test, implied or explicit.

bafolts: Ask and ye shall receive. I moved the conversion function onto the card to ease writing tests.

Issue #1851, opened 2020-11-07
A problem about PROJECT WORKSHOP


jijiqw:
![image](https://user-images.githubusercontent.com/16096558/98447926-a1810c00-2163-11eb-976f-5ffabcbd20bd.png)
![image](https://user-images.githubusercontent.com/16096558/98447927-a5149300-2163-11eb-9082-5530cf69681b.png)
I still can increase conlony track after  discarded the Trade Envoys

PR #1854, opened 2020-11-08
Fix: Remove blue card effects upon discard

nwai90:
Closes https://github.com/bafolts/terraforming-mars/issues/1851


src/Player.ts
```diff
@@ -162,10 +162,26 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
...
-    public increaseTitaniumValue() {
+    public increaseTitaniumValue(): void {
       this.titaniumValue++;
     }
 
+    public decreaseTitaniumValue(): void {
+        this.titaniumValue--;
+      }
+
+    public getSteelValue(): number {
```
kberg, line 173: How much would you mind renaming these `steelMegaCreditValue` ? Your call.

src/cards/community/ProjectWorkshop.ts
```diff
@@ -103,4 +105,33 @@ export class ProjectWorkshop implements CorporationCard {
     private logCardDraw(game: Game, player: Player, drawnCard: IProjectCard) {
         game.log("${0} drew ${1}", b => b.player(player).card(drawnCard));
     }
+
+    private checkForCardEffectsToRemove(player: Player, card: ICard): void {
```
kberg, line 109: This separates the card's effect from its removal. Consider instead, an interface such as

```
export interface IDiscardable {
  discard(game: Game, player: Player);
}
```

and for example, make these changes to `AdvancedAlloys`

```
export class AdvancedAlloys implements IProjectCard, **IDiscardable** {
    public cost = 9;
    public tags = [Tags.SCIENCE];
    public name = CardName.ADVANCED_ALLOYS;
    public cardType = CardType.ACTIVE;

    public play(player: Player) {
      player.increaseTitaniumValue();
      player.increaseSteelValue();
      return undefined;
    }
    **public discard(_game: Game, player: Player) {
      player.decreaseTitaniumValue();
      player.decreaseSteelValue();
      return undefined;
    }**
}
```

src/Player.ts
```diff
@@ -162,10 +162,26 @@ export class Player implements ILoadable<SerializedPlayer, Player>{
...
+        return this.steelValue;
+    }
+  
+    public increaseSteelValue(): void {
+        this.steelValue++;
+    }
+
+    public decreaseSteelValue(): void {
+        this.steelValue--;
```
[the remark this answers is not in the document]
nwai90: Agree. Will add constants for it. It should not be zero-boundary but rather that the value cannot fall below the default value

src/Player.ts
nwai90: We would have to do the same for all `getTitaniumValue` functions as well. I personally think the names are quite clear as they stand though

src/cards/community/ProjectWorkshop.ts
nwai90: This is a possibility 👍 Implementing a new interface for it would work for sure. It might be better to leave this custom behaviour on the corporation card though as it is highly specific.

kberg: The work done for Geological Survey and Ecological Survey is a bit outside the normal structure of things, and adding a new interface is the way I would update this change -- in fact, it's something I would very much like to do.

nwai90: I think adding a property `onDiscard?: (player: Player) => void;` to `ICard` is a better fit for this scenario. WDYT?

Lynesth: I like the onDiscard member on cards.

kberg: Agreed.

Do you want to add that in a follow-up PR? I'm fine with that.

nwai90: @kberg I will add it in this PR 👍

kberg (approved):

kberg:
Thanks!

PR #1857, opened 2020-11-08
Remove superfluous uses of `as Gameoptions`

kberg:


tests/TestingUtils.ts
```diff
@@ -53,11 +53,11 @@ export const setCustomGameOptions = function(options: object = {}): GameOptions
         clonedGamedId: undefined,
         cardsBlackList: [],
         aresExtension: false,
-        aresHazards: undefined,
+        aresHazards: false,
```
kberg, line 56: Yes, this is in fact necessary as `aresHazards` couldn't be undefined.

kberg:
Eep! Should be good now.

bafolts (approved):

PR #1860, opened 2020-11-08
Mandatory Venus completion as an option for a new game

### LINKED ISSUE #1411: Feature request: Venus TR track mandatory
We play with a house rule that the Venus terraforming track is mandatory to complete to end the game.

Would it be possible to add that as a starting option?

Starcalc:
Closes #1411


src/components/CreateGameForm.ts
```diff
@@ -656,6 +660,10 @@ export const CreateGameForm = Vue.component("create-game-form", {
                                 <label for="venusMA-checkbox">
                                     <span v-i18n>Venus Milestone/Award</span>
                                 </label>
+                                <input type="checkbox" v-model="venusTR" id="venusTR-checkbox">
+                                <label for="venusTR-checkbox">
+                                    <span v-i18n>Venus at 30% for win</span>
```
nwai90, line 665: Minor: This label is quite long and it's also "for game end" instead of "for win". Perhaps just `"Venus Terraforming"` will suffice?

Starcalc, line 665:
```suggestion
                                    <span v-i18n>Venus Terraforming</span>
```

Starcalc: Sorry, I hit the wrong button to change the code. I accept your suggestion and inserted it my "own review" below. I'm not quite familiar what would be the correct way to accept your idea. Should I have created a new pull request with the suggested changes?

src/Game.ts
```diff
@@ -574,6 +576,12 @@ export class Game implements ILoadable<SerializedGame, Game> {
                 this.board.getOceansOnBoard() === constants.MAX_OCEAN_TILES &&
                 this.getVenusScaleLevel() === constants.MAX_VENUS_SCALE;
       }
+      if (this.gameOptions.venusTR) {
```
kberg, line 579: Remove the duplication with

```
return this.oxygenLevel >= constants.MAX_OXYGEN_LEVEL &&
               this.temperature >= constants.MAX_TEMPERATURE &&
               this.board.getOceansOnBoard() === constants.MAX_OCEAN_TILES &&
               (this.getVenusScaleLevel() === constants.MAX_VENUS_SCALE || this.gameOptions.venusTR !== true);
```

src/components/CreateGameForm.ts
```diff
@@ -114,7 +115,8 @@ export const CreateGameForm = Vue.component("create-game-form", {
             soloTR: false,
             clonedGameData: undefined,
             cloneGameData: [],
-            allOfficialExpansions: false
+            allOfficialExpansions: false,
+            venusTR: false
```
kberg, line 119: I'm not going to argue strongly for a better name, but I think there's a better name than `venusTR` since this isn't about a **Terraforming Rating**

src/components/CreateGameForm.ts
```diff
@@ -656,6 +660,10 @@ export const CreateGameForm = Vue.component("create-game-form", {
                                 <label for="venusMA-checkbox">
                                     <span v-i18n>Venus Milestone/Award</span>
                                 </label>
+                                <input type="checkbox" v-model="venusTR" id="venusTR-checkbox">
```
kberg, line 663: Should not be visible if Venus is off. Could you add a `v-if`?

src/Game.ts
kberg, line 579: So here's a use case that will cause a problem:

Player enables Venus in the UI. Then they enable this option. And then they turn off Venus in the UI.

This game will never end. So you'll need an additional comparison for the expansion itself being initialized. Which is fairly standard, so that's not a big deal. But it does make my suggested fix more complicated -- but you can still remove the duplication.

kberg (commented):
Can you add a test in Game.spec.ts, please, to ensure this doesn't break existing behavior if the option is off?

src/components/CreateGameForm.ts
nwai90: True. Perhaps `requiresVenusTrackCompletion` or similar

src/components/CreateGameForm.ts
Starcalc: This is included in the same part where the Milestone is checked, so the <template v-if="venusNext"> already works for this one, too.

src/components/CreateGameForm.ts
Starcalc: I tried to blend in with the other variable names like venusMA and soloTR. I don't mind though, I've already changed it locally and will update soon.

src/Game.ts
Starcalc: I've added a small "deselect"-function in CreateGameForm.ts to make sure that requiresVenusTrackCompletion is set to false when venusNext is deactivated.

src/components/CreateGameForm.ts
kberg: > I tried to blend in with the other variable names like venusMA and soloTR. I don't mind though, I've already changed it locally and will update soon.

Ah, thanks for the context. Still, thank you!

src/Game.ts
kberg: That's definitely good, but I'd still add it here -- there's a UI that lets someone upload parameters via JSON, so it's going to need a second change _someplace._

Starcalc: Added the failsafe. And correct the last entry of the old variable name, sorry, I've overlooked that one.

alrusdi:
@Starcalc please resolve the conflicts (merge upstream master branch into your)

src/Game.ts
kberg: > @Starcalc please resolve the conflicts (merge upstream master branch into your)
Yeah, when that's done we can take another look at this.

src/Game.ts
```diff
@@ -98,6 +98,7 @@ export interface GameOptions {
   customCorporationsList: Array<CardName>;
   cardsBlackList: Array<CardName>;
   customColoniesList: Array<ColonyName>;
+  requiresVenusTrackCompletion: boolean; // Venus must be completed for win
```
kberg, line 101: Minor: Not for the win, but to end the game.

src/Game.ts
```diff
@@ -572,6 +574,12 @@ export class Game implements ILoadable<SerializedGame, Game> {
                 this.board.getOceansOnBoard() === constants.MAX_OCEAN_TILES &&
                 this.getVenusScaleLevel() === constants.MAX_VENUS_SCALE;
       }
+      if (this.gameOptions.venusNextExtension && this.gameOptions.requiresVenusTrackCompletion) {
```
kberg, line 577: Consider

```
private marsIsTerraformed(): boolean {
  const oxygenMaxed = this.oxygenLevel >= constants.MAX_OXYGEN_LEVEL;
  const temperatureMaxed = this.temperature >= constants.MAX_TEMPERATURE;
  const oceansMaxed = this.board.getOceansOnBoard() === constants.MAX_OCEAN_TILES;
  const globalParametersMaxed = oxygenMaxed && temperatureMaxed && oceansMaxed;

  const venusMaxed = this.getVenusScaleLevel() === constants.MAX_VENUS_SCALE;

  // Solo games with Venus need venux maxed to end the game.
  if (this.players.length === 1 && this.gameOptions.venusNextExtension) {
    return globalParametersMaxed && venusMaxed;
  }
  // new option "requiresVenusTrackCompletion" also makes maximizing Venus a game-end requirement
  if (this.gameOptions.venusNextExtension && this.gameOptions.requiresVenusTrackCompletion) {
    return globalParametersMaxed && venusMaxed;
  }
  return globalParametersMaxed;
}
```

Reduces repetition.

kberg (commented):
Thanks! This is a nice change. Two minor comments.

src/Game.ts
Starcalc: I've included this in the latest commit, thank you for the suggestion

kberg (approved):
Thanks for the work!

Issue #1642, opened 2020-10-17
Show warnings when doing an action or playing a card that may not be "useful"


Lynesth:
## The idea
That title is probably a bit vague, so let me try to explain better.
The idea is to show a warning if the card/action you are playing is probably not going to do what you expect. Here are some examples of such moves:

- Using the standard project **Asteroid** when temperature is already maxed.
- Activating **Martian Rails** without a city on mars.
- Playing **Energy Tapping** when none of your opponents have energy production.
- Playing **Virus** against a opponent that has **Protected Habitats**.

Not saying that there's no value at all in playing those, but it's most likely not going to do what you expected if you're not focused enough.

## What exactly
Here's the proposed list of things to track and show warnings for:
- Global Parameter (O2, Temp, Venus, Ocean) already maxed (or not getting full effect from card)
- Removing plants/microbes/animals while your only valid target has Protected Habitats (or it's animals on Pets)
- Removing resources without any valid target, like playing Virus but noone has microbes/plants (not counting yourself)
- Decreasing anyone's production where the only valid target is yourself
- Gaining resources/production for every city without any city
- Gaining resources/production/bonus for every colony without any colony
- Gaining resources/production for each specific tag without said tag (or without enough tags to get at least 1)
- On cards that place Oceans + Remove plants, if one of the available targets has Arctic Algae
- Adding resources while having no cards able to receive them
- Look at top card to buy it without enough money to buy the card

I think we should only track what the card/action does and not subsequent effects that derives from the play. Example: If a card places a Greenery, we should **not** check if oxygen is already maxed.

## How (technically)
There's currently two proposals that need inputs and opinions:

#### Card by card
Add a `checkWarnings: (Game, Player) => void` method to cards that will populate a `warningText?: string | undefined` property.
Call `checkWarning` after checking if the card/action is playable. Use `warningText` to print the warning message.

#### A class to check them all
Create a new class that will be called to check for warnings, using `checkWarnings: (Game, Player, ICard) => string | undefined`.
Add a few properties to cards (non exhaustive list): `addsOxygen?: number`, `addsTemperature?: number`, `addsOcean?: number`, `addsVenus?: number`, `removesPlants?: number`, etc. Use string returned`checkWarnings to print the warning message.

## How (visually)
The proposal is to have a message that doesn't is simple, visible and doesn't require you to take any additional action.
It would just be a small warning above the button that validates the action you want to take.
[...]

PR #1864, opened 2020-11-09
How to afford Reds policy and maybe more

Lynesth:
**_Disclaimer: this is a draft, a proof of concept. It is mostly working but is not intended to be merged as is. I hope that it can instead start a meaningful discussion about the issue this PR tries to resolve._**

One of the biggest issue we have at the moment is the lack of proper handling of the interactions between cards/actions and Reds policy. We have checks here and there to cover very basic affordability tests but that's not enough (I can expand on that if you want).

So here I made a new function (`RedsPolicy.canAffordRedsPolicy`), whose goal is to replace `player.canAfford` when Reds are ruling.
Here's a summary of how it checks if you can afford Reds _(everytime I mention MC, it means MC+heat if you're playing Helion)_:

1. It first calculate how much TR you will get from the card you play, taking into account that temperature increase can lead to an extra ocean being placed, and so on. That gives us the total amount you'll have to pay Reds.
2. It calculates, given all your already played cards and your corporation, how much "cashback" you are sure to get from simply playing the card (from things like Media Group, Advertising, Meat Industries + Martian Zoo, etc).
    - If you can afford everything (card + Reds) with your current MC + the "cashback", then it **_returns_** that you can afford the card.
3. It sets aside the amount owed to Reds, since it can only be paid with MC, and check if you can pay the card adding other resources (steel, titanium, etc).
    - If you can afford it that way, then it **_returns_** that you can afford the card but that you cannot spend more than X MC on the card itself (since the rest will be used for Reds)
    - If you still can't afford the card, the only way to get more cash would be by placing tiles. If the card doesn't place any (including temp ocean) then it simply **_returns_** that you cannot afford the card.
4. It simulates adding the tiles on the map one by one, starting with oceans, and for each permutation it stores how much MC you would get (from ocean adjacency or heat space bonus) and that would allow you to pay for it.
    - If you can afford it by placing tiles a certain way, then it **_returns_** that you can afford the card but that you cannot spend more than X MC on the card itself and that you must place the tiles in certain spaces.
    - If you still can't, then it **_returns_** that you cannot afford the card.

Then we hack into the `SelectSpace` and `SelectHowToPayForCard` inputs to enforce the rules if applicable.

The goal isn't to be 100% accurate but to approach that without going into very complicated stuff (like I don't think Philares effects will be taken into account ever in that function). Also, it shouldn't never return a false positive.

-----

#### Notes:
I tried to be pretty verbose in the comments but here are some notes about the implementation:
- The function expect an object with details about the card you are playing, like how much temperature increase it does, how much oceans it places, etc. This kind of things could probably be set on the card directly and retrieved that way.
- It's designed to work with standard projects, plants/heat conversion and card actions as well, but at the moment, the max MC you can spend on the action is only enforced for cards, not those.
- Speaking about that, the way I hacked into `SelectHowToPayForCard` is a bit sluggish but I just wanted something that worked for this poc and I'm not really familiar with the front end of the app.
- When you can afford everything by placing your tiles a certain way, it returns a tree of spaces which, when placing the last tile of the branch, will give you enough to pay for it. This effectively gives the player all available options.
- It does not work with Helias special tile for now as I totally forgot about the fact that it adds an ocean. 

I only modified Lava Flows and Big Asteroid to work with it for now. It should however work properly with any card in the game by simply following the example set in those two.

-----

It is currently tailored to work for Reds Policy only but my hope is that the way I'm trying to solve the issue gets approval and that it could then be extended to be a full replacement of `player.canAfford`, while also taking care of adjacency bonus/malus from Ares extension.

Let me know what you think :)


Lynesth:
@bafolts @kberg @alrusdi @chosta @nwai90

src/turmoil/RedsPolicy.ts
```diff
@@ -0,0 +1,355 @@
...
+import { Player } from "../Player";
+import { ResourceType } from "../ResourceType";
+import { SpaceBonus } from "../SpaceBonus";
+import { SpaceName } from "../SpaceName";
+import { StandardProjectType } from "../StandardProjectType";
+import { Tags } from "../cards/Tags";
+import { TileType } from "../TileType";
+
+/*
+ * TODO: Most of the members of that class could be inferred from card metadata once it's usable
```
chosta, line 16: as @kberg advised it would be nice to have TODO (author) mentioned so we know who to harass for it :)

chosta (commented):
Fist of all - great work! This is indeed something that causes a lot of trouble and we all knew it would take a more holistic approach to the proble to get it solved for good. I like the approach you chose but can't be really helpful with the implementation as others know this part of the codebase better. I have 2 suggestions though:
* It would be nice to have a property for every card that needs to be checked against the RedPolicy. This way every newly added card should make the developer think about Reds (terrorists!)
* If we have the property I mentioned it would be really easy to also provide a visual warning that "The card you are about to use, may have special Reds requirements" or smth. I'm willing to help with the frontend for that.

Lynesth:
Thanks for your comments @chosta !
The Reds of Mars are really just the Greens of Earth :)

> It would be nice to have a property for every card that needs to be checked against the RedPolicy. This way every newly added card should make the developer think about Reds (terrorists!)

I think that would be nice with the current state of the app yes. If we're really moving forward to using card metadata server side as well though, it would simply be inferred automatically.

> If we have the property I mentioned it would be really easy to also provide a visual warning that "The card you are about to use, may have special Reds requirements" or smth. I'm willing to help with the frontend for that.

Yes. I was thinking of using the same idea I already proposed since it kinda falls in that category: https://github.com/bafolts/terraforming-mars/issues/1642

Lynesth:
I would love if this could receive more feedback as I don't think I'm going to keep working on it unless it gets some approval :)

nwai90:
> I would love if this could receive more feedback as I don't think I'm going to keep working on it unless it gets some approval :)
> I know it's not a small function and my opening comment is kind of a big wall of text as well but I'm here (or discord) if you have any question.

I like the approach. It is a complex chunk of logic and likely can be simplified further but this should suffice as a first cut. Might also be helpful to setup a preset game seed on a separate server that allows the 2 modified cards to be manually tested against various scenarios.

Lynesth:
> Might also be helpful to setup a preset game seed on a separate server that allows the 2 modified cards to be manually tested against various scenarios.

This is a good idea. I've sent the link in the dev chan on discord.

kberg:
Keep Open.

kberg:
Several months later and I'm finally getting to this, because I can understand it now (where I couldn't earlier.)

This general approach makes a lot of sense. My criticism is not about the idea but about a meta-problem with this app, which is that it intends to predict everything. This does a great job there.

If however, we moved to a framework that made undo cheap and simple, a player could play a card, and discover they don't have the money to pay additional costs, and roll back their choice. Which is another discussion altogether.

kberg:
I'm gonna close this out. There hasn't been any activity since December 2020, except for my comment in May, four months ago. Since I'm on the verge of taking on a different approach, laid out in #3595, this doesn't need to be around, though I edited #3595 to reference it as prior art.

PR #1605, opened 2020-10-13
Cards metadata to vue

chosta:
Continuing the battle with HTML_data.ts:
+ added `CardMetadata` interface (still more attributes to come)
+ added CardRequirements to metadata
+ abstracting the `<div class="content">` by checking if metadata exists. If yes => use Vue components , no => use the HTML_data.ts content
+ first card to be set free is also the best one in the game ->>> `Dust Seals` !

This is WIP so, please just comment on what you think needs to change in the existing code and not what is still missing :)

PR #1867, opened 2020-11-09
Cards metadata items + symbols to Vue

chosta:
Another batch from #1605 [which I closed now]

Considering the comments from @Lynesth and @kberg I did some refactoring and changed the approach of how to "draw" on the cards. I chose the "render" wording this time. A couple of notes on this one:
* Had to do initial formatting for the cards I wanted to change but this happened in-between the new `gts` rules so I have 2 commits dealing with that only (at the start and at the end)
* Using one class CardRenderer to render data on top of the card canvass. Fully committed to the builder pattern now
* Refactored the vue render components by reducing boilerplate and repeating elements
* CardSpecial is now CardSymbol (which should make more sense)
* Added a new enum that should describe all possible item types that can exist on the cards with the caveat that it will duplicate some of the existing enums but it will reduce boilerplate code and help keep things tight. It is an enum that handles rendered items only so I guess we can have it as a separate entity.


src/components/card/CardRenderItemComponent.ts
```diff
@@ -0,0 +1,82 @@
...
+        isSubtract: function (): boolean {
+            return this.item.amount < 0;
+        },
+        itemsToShow: function (): number {
+            if (this.item.showDigit) return 1;
+            return this.getAmountAbs();
+        },
+        itemHtmlContent: function (): string {
+            // in case of symbols inside
+            if (this.item instanceof CardRenderItem && this.item.amountInside())
```
chosta: This one has been fixed in 13fc068

.prettierrc.yaml
```diff
@@ -1,3 +1,6 @@
-printWidth: 100
-tabWidth: 4
-quoteProps: "preserve"
+tabWidth: 2
```
bafolts, line 1: I didn't realise we had this `.prettierrc.yaml` file. When trying to use `gts` directly I kept getting a strange error where this was changing single quotes back to double quotes and I wasn't sure why. This would explain what I was seeing.

chosta: I slipped this one in a recent PR, maybe I should have made a separate one just so everyone is aware.

bafolts (approved):

kberg (approved):

Lynesth:
One thing I find a bit strange is that `any()` comes after the production/resource it modifies.
Would that be possible to make it come before so that it reads better ? Also can you make it not a function if it's never supposed to receive any argument, just for the sake of readability again ?
`pb.plants(-1).any().br().energy(2)` vs `pb.any.plants(-1).br().energy(2)`

.prettierrc.yaml
chosta: It's 2 now with `gts`. The linter forces it.

src/cards/Algae.ts
```diff
@@ -5,18 +5,29 @@ import {Player} from '../Player';
...
-    public cardType = CardType.AUTOMATED;
-    public canPlay(player: Player, game: Game): boolean {
-      return game.board.getOceansOnBoard() >= 5 - player.getRequirementsBonus(game);
-    }
-    public play(player: Player) {
-      player.plants++;
-      player.addProduction(Resources.PLANTS, 2);
-      return undefined;
-    }
+  public cost = 10;
```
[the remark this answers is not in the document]
chosta: I did try, the problem is that the `gts` change came in the middle of my PR work and @bafolts pushed a change with enforced formatting for all files. I'm sure we have our formatting issues in the past now as my formatter works according to all (almost) the `gts` rueles.

src/cards/render/CardRenderItem.ts
```diff
@@ -0,0 +1,42 @@
+import {CardRenderItemType} from './CardRenderItemType';
+
+export class CardRenderItem {
+  private _anyPlayer: boolean = false;
+  private _showDigit: boolean = false;
+  private _amountInside: boolean = false;
+  constructor(protected _cardRenderItemType: CardRenderItemType, protected _amount: number = -1) {
```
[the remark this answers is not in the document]
chosta: I think I had a reason to do that but I agree that `_type` makes more sense. Will change it

chosta:
> `pb.plants(-1).any().br().energy(2)` vs `pb.any.plants(-1).br.energy(2)`

Well to me it reads better when the main item is at the beginning but I have some ideas how to improve it.

I was actually thinking about `.plants({amount: -1, anyPlayer: true})` or the short `.plants(-1, true)` as we will probably have more stuff as a part of the item like `.microbes({amount:2, asterix: true})`. I'll make adjustments on the fly with new elements coming but your suggestion is an improvement.

PR #1875, opened 2020-11-11
feat: sort cards in hand

bafolts:
This change allows sorting cards in hand. We can expand upon this feature. For now it is used for when selecting how to pay for a card as that is usually when you sort cards while playing. The drag and drop sorting is simple enough to implement we can probably scale this to use other places. I want to add unit tests prior to merging. Putting up draft pr for early excitement.


tests/components/SortableCards.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+        expect(JSON.parse(expectedStorage["cardOrderfoo"])).to.deep.eq({
+            [CardName.ANTS]: 2,
+            [CardName.CARTEL]: 1
+        });
+    });
+    it("puts new cards at end of order and removes old", async function () {
+        expectedStorage["cardOrderfoo"] = JSON.stringify({
+            [CardName.ANTS]: 2,
+            [CardName.CARTEL]: 1,
+            [CardName.DECOMPOSERS]: 3
```
nwai90, line 73: Should this be `CardName.BIRDS`?

bafolts: This test is using `CardName.DECOMPOSERS` which would represent a card the user had in their hand but through some action no longer has the card in their hand. The `propsData` no longer contains the decomposers card. This test is making sure in this situation we put the new cards `CardName.BIRDS` on the end and don't consider the removed card `CardName.DECOMPOSERS`.

src/components/CardOrderMixin.ts
```diff
@@ -0,0 +1,38 @@
+
+import {CardModel} from '../models/CardModel';
+
+const STORAGE_PREFIX = 'cardOrder';
+
+export const CardOrderMixin = {
+  name: 'CardOrderMixin',
+  methods: {
+    getCardOrder: function(playerId: string) {
```
kberg, line 9: And this will be playerId and not color?

src/components/CardOrderMixin.ts
```diff
@@ -0,0 +1,38 @@
...
+        return {};
+      }
+    },
+    getOrdered: function(playerId: string, cards: Array<CardModel>): Array<CardModel> {
+      const order = this.getCardOrder(playerId);
+      const copy = cards.slice();
+      copy.sort((a, b) => {
+        return order[a.name] - order[b.name];
+      });
+      return copy;
```
kberg, line 27: Will all of this work in thew face of local storage being cleared out?

src/components/SelectHowToPayForCard.ts
```diff
@@ -41,17 +44,20 @@ export const SelectHowToPayForCard = Vue.component('select-how-to-pay-for-card',
     },
   },
   data: function(): SelectHowToPayForCardModel {
+    let cards: Array<CardModel> = [];
     let card: string | undefined;
     if (this.playerinput !== undefined &&
             this.playerinput.cards !== undefined &&
             this.playerinput.cards.length > 0) {
-      card = this.playerinput.cards[0].name;
+      cards = (this as unknown as typeof CardOrderMixin.methods).getOrdered(this.player.id, this.playerinput.cards);
```
kberg, line 52: Is there any way to reduce this -- may I say -- gross visual bit of code?

tests/components/SortableCards.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+        let cards = sortable.findAllComponents({
+            name: "Card"
+        });
+        expect(cards.length).to.eq(2);
+        expect(cards.at(0).props().card.name).to.eq(CardName.ANTS);
+        expect(cards.at(1).props().card.name).to.eq(CardName.CARTEL);
+        const draggers = sortable.findAllComponents({
+            ref: "cardbox"
+        });
+        await draggers.at(0).trigger("dragstart");
```
kberg, line 55: Oh my goodness, this test.

kberg: ... in a good way.

kberg (approved):
Wow, this is amazing.

src/components/CardOrderMixin.ts
bafolts: This will be `playerId`. The card order is specific to the player on the client.

src/components/CardOrderMixin.ts
bafolts: I think this would work by coincidence but I should make this clearer. If either `order[a.name]` is `undefined` or `order[b.name]` is `undefined` those values should be put on the end. If as a user you never order cards you will always end up returning `NaN` which seems to sort fine, but this should be explicit. Will address prior to merging as this is only part of change that still concerns me.

bafolts: The items that are not found need to be taken out and placed on the end in their original order. Only items that have a hit can be sorted.

src/components/SelectHowToPayForCard.ts
bafolts: I have to say as the comeback is just there, but `unknown`. Been putting this off if vue supports typing with the mixins but need to look it up so that it is now `unknown`.

kberg: One of these days.

kberg:
I'm going to leave this to you to merge -- looks like you might make the bit about the system surviving a cache clear more obvious, but otherwise, merge!

PR #1877, opened 2020-11-12
separate build and start

bafolts:
The `build` and `start` commands have been synonymous for some time. The original idea was that `build` was for building without starting the server. In the original commits `build` was only `tsc` so it wasn't too important. As more has been added to `build` a lot has been duplicated in `start`. Looking at heroku builds I am fairly sure it runs both `build` and `start` duplicating these efforts. I deployed this branch manually to verify this change still works with heroku. 

This change is more in line with heroku node best practices to separate `build` and `start`.

https://devcenter.heroku.com/articles/node-best-practices


bafolts:
@Lynesth using `postinstall` I am not sure how heroku wouldn't then subsequently run `build` anyway. This approach no matter what will be less duplication between `build` and `start` in our `package.json`. If the time limit is only for the `npm run start` command this should help us on heroku as well.

Lynesth:
LGTM 👍

alrusdi (approved):

PR #1881, opened 2020-11-12
Fixing renamed parameter...

pierrehilbert:
I forgot to change the parameter name in the server.ts ❌


vincentneko (approved):
Looking good to me

PR #1882, opened 2020-11-13
reduce repetition and simplify logic

otolabqu:


src/Player.ts
```diff
@@ -190,16 +189,16 @@ export class Player implements ILoadable<SerializedPlayer, Player> {
...
-        this.hasIncreasedTerraformRatingThisGeneration = true;
-        return;
       }
 
-      this.terraformRating++;
-      this.hasIncreasedTerraformRatingThisGeneration = true;
+      this.increaseTerraformRatingExecute();
     }
 
+    private increaseTerraformRatingExecute() {
```
kberg, line 197: Drop the `execute`. It will make more sense.

kberg (changes requested):
Nice improvement, you have to fix the formatting failures, and I have recommended a minor change.

Lynesth:
Could probably just do 
```ts
public increaseTerraformRating(game: Game) {
  if (game.phase === Phase.ACTION && game.gameOptions.turmoilExtension && PartyHooks.shouldApplyPolicy(game, PartyName.REDS)) {
    if (this.canAfford(REDS_RULING_POLICY_COST)) {
      game.defer(new SelectHowToPayDeferred(this, REDS_RULING_POLICY_COST, false, false, 'Select how to pay for TR increase'));
    } else {
      // Cannot pay Reds, will not increase TR
      return;
    }
  }
  this.terraformRating++;
  this.hasIncreasedTerraformRatingThisGeneration = true;
  return;
}
```

src/Player.ts
Lynesth: It's already the name of the function just above, that's probably why `Execute` was added.

nwai90: Note: The current logic for `else` block ("Cannot pay Reds, will not increase TR") was a legacy stopgap measure. Ideally this code should not be reachable at all if the player could not pay Reds tax while the Reds are ruling (validated through `canPlay` for cards and separate logical checks for SPs).

The top-level function `public increaseTerraformRating(game: Game) {}` could be rewritten as follows. This might break some stuff in the short-term but it would also help us to discover any cards that have missing `canPlay` logic.

```
if (PartyHooks.shouldApplyPolicy(game, PartyName.REDS) && game.phase === Phase.ACTION) {
    game.defer(new SelectHowToPayDeferred(this, REDS_RULING_POLICY_COST, false, false, 'Select how to pay for TR increase'));
}

this.terraformRating++;
this.hasIncreasedTerraformRatingThisGeneration = true;
```

nwai90:
> Could probably just refactor the function like this instead:
> 
> ```ts
> public increaseTerraformRating(game: Game) {
>   if (game.phase === Phase.ACTION && PartyHooks.shouldApplyPolicy(game, PartyName.REDS)) {
>     if (this.canAfford(REDS_RULING_POLICY_COST)) {
>       game.defer(new SelectHowToPayDeferred(this, REDS_RULING_POLICY_COST, false, false, 'Select how to pay for TR increase'));
>     } else {
>       // Cannot pay Reds, will not increase TR
>       return;
>     }
>   }
>   this.terraformRating++;
>   this.hasIncreasedTerraformRatingThisGeneration = true;
> }
> ```
> 
> _Edit: Made it a bit shorter_

Yep. We can remove the `else` block and the check for `this.canAfford(REDS_RULING_POLICY_COST)` after https://github.com/bafolts/terraforming-mars/pull/1864 is merged

kberg:
@otolabqu when do you think you'll plan to get back to this? I'd love to have this resovled this weekend. Thanks!

otolabqu:
> @otolabqu when do you think you'll plan to get back to this? I'd love to have this resovled this weekend. Thanks!

I read the comments and thought we were waiting for `after #1864 is merged` but see now that that was to make it even better afterwards, and I now made the suggested change by Lynesth for now.

otolabqu:
@kberg I think this is waiting on you now, Github says you requested changes.

kberg:
Oh, that's much better. Thanks!

kberg (approved):

PR #1889, opened 2020-11-14
separate webpack command and faster builds for local

bafolts:
In a quest to speed up the local build process. This will cut down the time webpack takes for local builds.

Before: 

```
Hash: c773998b9e687a6242f8
Version: webpack 4.44.2
Time: 7700ms
```

After:

```
Hash: a4c68bf5f2073b038919
Version: webpack 4.44.2
Time: 1874ms
```

This also enables the vue warnings in console. Seems I have the typing information wrong on `props`. Will address in subsequent PR. Those are littering vue debug logs.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -5,7 +5,7 @@ const CompressionPlugin = require("compression-webpack-plugin");
 
 module.exports = {
   devtool: "source-map",
-  mode: process.env.NODE_ENV === "production" ? "production" : "none",
+  mode: process.env.NODE_ENV === "production" ? "production" : "development",
   entry: [
     './build/script.js'
   ],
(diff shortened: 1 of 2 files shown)


webpack.config.js
```diff
@@ -1,10 +1,11 @@
 'use strict'
 
+const process = require("process");
 const CompressionPlugin = require("compression-webpack-plugin");
 
 module.exports = {
   devtool: "source-map",
-  mode: 'production',
+  mode: process.env.NODE_ENV === "production" ? "production" : "none",
```
chosta, line 8: I'm used to "development" mode. Just curious why "none" is better in our case?

chosta (approved):
Great!

webpack.config.js
bafolts: I can use "development". When webpack told me the options were "production", "development", or "none" I just picked "none".

PR #1891, opened 2020-11-14
vue runtime warning fixes

bafolts:
With the `development` flag set on webpack warnings are logged to the console during runtime for vue typing errors. Fixes included here.

### DIFF
--- a/src/components/CorporationsFilter.ts
+++ b/src/components/CorporationsFilter.ts
@@ -150,7 +150,6 @@ export const CorporationsFilter = Vue.component('corporations-filter', {
         </div>
         <div class="corporations-filter-group" v-for="entry in corpsByModule">
             <div class="corporations-filter-toolbox-cont">
-                <h2>{{ name }}</h2>
                 <div class="corporations-filter-toolbox">
                     <a href="#" v-on:click.prevent="selectAll(entry[0])">All</a> | 
                     <a href="#" v-on:click.prevent="selectNone(entry[0])">None</a> | 
--- a/src/components/GameEnd.ts
+++ b/src/components/GameEnd.ts
@@ -96,7 +96,7 @@ export const GameEnd = Vue.component('game-end', {
                     </a>
                 </div>
                 <div v-if="!isSoloGame() || player.isSoloModeWin" class="game-end-winer-announcement">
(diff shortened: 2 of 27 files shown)


src/components/CorporationsFilter.ts
```diff
@@ -150,7 +150,6 @@ export const CorporationsFilter = Vue.component('corporations-filter', {
         </div>
         <div class="corporations-filter-group" v-for="entry in corpsByModule">
             <div class="corporations-filter-toolbox-cont">
-                <h2>{{ name }}</h2>
```
bafolts, line 153: Nothing is using this as far as I can tell. Vue logs warning that `name` is not a provided prop or data.

src/components/GameEnd.ts
```diff
@@ -96,7 +96,7 @@ export const GameEnd = Vue.component('game-end', {
                     </a>
                 </div>
                 <div v-if="!isSoloGame() || player.isSoloModeWin" class="game-end-winer-announcement">
-                    <span v-for="p in getWinners()"><log-player :class="getEndGamePlayerHighlightColorClass(p.color)">{{ p.name }}</log-player></span> won!
+                    <span v-for="p in getWinners()"><span :class="'log-player ' + getEndGamePlayerHighlightColorClass(p.color)">{{ p.name }}</span></span> won!
```
bafolts, line 99: The usage of `log-player` and `log-card` was creating a custom HTML element that we never defined. Vue was warning against this.

src/components/card/CardRowComponent.ts
```diff
@@ -26,7 +26,7 @@ export const CardRowComponent = Vue.component('CardRowComponent', {
   },
   template: ` 
         <CardRenderItemComponent v-if="isItem()" :item="data"/>
-        <CardProductionBoxComponent v-else-if="isProduction()" :data="data.rows" />
+        <CardProductionBoxComponent v-else-if="isProduction()" :rows="data.rows" />
```
bafolts, line 29: I changed this name from `data` to `rows` as vue has it's own `data` property on the component which could become confusing.

src/styles/log.less
```diff
@@ -47,6 +47,3 @@ log-card {
     text-transform: uppercase;
 }
 
-log-generation {
```
bafolts, line 50: Nothing using this as far as I can tell.

kberg (approved):

PR #1893, opened 2020-11-15
PoliticalAgendas fan expansion

nwai90:
Groundwork from me and @kberg for https://boardgamegeek.com/thread/2411719/turmoil-20-political-agendas. The expansion is originally for GIGA map and has been adjusted for base maps.

The main idea: Each party has its own unique agenda with 2 possible ruling bonuses and 4 possible ruling policies.

Three game modes will be introduced with this expansion:
1) STANDARD: Use official ruling policy and bonus for all parties. This will be the default mode.
2) RANDOM: Randomize ruling policy and bonus for every party.
3) CHAIRMAN: Incoming chairman can set ruling policy and bonus for the new ruling party.


server.ts
```diff
@@ -795,7 +797,7 @@ function getPlayers(players: Array<Player>, game: Game): Array<PlayerModel> {
       coloniesExtension: game.gameOptions.coloniesExtension,
       steel: player.steel,
       steelProduction: player.getProduction(Resources.STEEL),
-      steelValue: player.getSteelValue(),
+      steelValue: player.getSteelValue(game),
```
kberg, line 800: Huh. Is there a way to get this right without  having to have player call PartyHooks for each invocation?

src/Game.ts
```diff
@@ -82,6 +82,7 @@ export interface GameOptions {
   turmoilExtension: boolean;
   promoCardsOption: boolean;
   communityCardsOption: boolean;
+  politicalAgendasExtension: boolean;
```
kberg, line 85: Move this two lines down to be in line with Game.ts?

src/Player.ts
```diff
@@ -173,7 +173,8 @@ export class Player implements ILoadable<SerializedPlayer, Player> {
       }
     }
 
-    public getSteelValue(): number {
+    public getSteelValue(game: Game): number {
+      if (PartyHooks.shouldApplyPolicy(game, PartyName.MARS, 'mfp03')) return this.steelValue + 1;
```
kberg, line 177: Replace the strings for policy card ids with an enum. Can be a TODO.

src/turmoil/PoliticalAgendasData.ts
```diff
@@ -0,0 +1,19 @@
+import {Bonus} from './Bonus';
+import {PartyName} from './parties/PartyName';
+import {Policy} from './Policy';
+
+export enum AgendaStyle {
+  STANDARD,
+  RANDOM,
+  CHAIRMAN
```
kberg, line 8: Add `// To be implemented`?

kberg (commented):
So by and large I think this is good, but a) I'd recommend having a place for localized testing and feedback so as to ensure we didn't break anything, or anybody's brains. Perhaps merge this not into master, but a nearby branch?

src/turmoil/Turmoil.ts
```diff
@@ -40,8 +42,9 @@ export class Turmoil implements ILoadable<SerializedTurmoil, Turmoil> {
     public distantGlobalEvent: IGlobalEvent | undefined;
     public commingGlobalEvent: IGlobalEvent | undefined;
     public currentGlobalEvent: IGlobalEvent | undefined;
+    public politicalAgendasData: PoliticalAgendasData;
 
-    constructor(game: Game) {
+    constructor(game: Game, politicalAgendas: boolean) {
```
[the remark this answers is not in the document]
nwai90: Done

server.ts
nwai90: It's the same for `player.getTitaniumValue(game)` as the new ruling party needs to be checked every generation

kberg: This might not be the right approach, but imagine if game had an attribute, `steelValueAdjustmentThisGeneration.`

The name sucks, but, well, /shruggie

src/turmoil/PoliticalAgendasData.ts
```diff
@@ -0,0 +1,19 @@
+import {Bonus} from './Bonus';
+import {PartyName} from './parties/PartyName';
+import {Policy} from './Policy';
+
+export enum AgendaStyle {
+  STANDARD,
+  RANDOM, // TODO: Implement
```
kberg, line 7: Random isn't implemented?

src/turmoil/parties/PartyHooks.ts
```diff
@@ -20,7 +20,7 @@ export class PartyHooks {
     }
   }
 
-  static shouldApplyPolicy(game: Game, partyName: PartyName) {
+  static shouldApplyPolicy(game: Game, partyName: PartyName, policyID?: string) {
```
kberg, line 23: By the way, I think we call this `policyId`. I'm not going to ask you to change it -- your call.

kberg:
As a reminder, perhaps a sanity playtest with the feature off would help, and/or a playtest running somewhere else.

Tests for the new policies? They can be added one-by-one, I don't mind.

src/turmoil/PoliticalAgendasData.ts
nwai90: Not yet. I will finish defining the alt policies for the remaining two parties first (Scientists and Reds). We will get there at some point

nwai90:
> As a reminder, perhaps a sanity playtest with the feature off would help, and/or a playtest running somewhere else.
> 
> Tests for the new policies? They can be added one-by-one, I don't mind.

Will add tests after finishing the alt policies. We can deploy this to a separate app for beta testing when the first version is ready

src/turmoil/parties/PartyHooks.ts
nwai90: Done

src/Player.ts
```diff
@@ -633,6 +635,11 @@ export class Player implements ILoadable<SerializedPlayer, Player> {
         tagCount += this.scienceTagCount;
       }
 
+      // PoliticalAgendas Scientists P4 hook
+      if (tag === Tags.SCIENCE && this.hasTurmoilScienceTagBonus) {
```
nwai90, line 639: The alternative was to do:

```
// PoliticalAgendas Scientists P4 hook
if (tag === Tags.SCIENCE && PartyHooks.shouldApplyPolicy(game, PartyName.SCIENTISTS, 'sp04')) {
    tagCount += 1;
}
```

This would however involve changing hundreds of files as the function signatures for `getTagCount`, `getMultipleTagsCount`, `checkMultipleTagsPresence` etc. would all have to be modified to include `Game`, which is not desirable.

nwai90:
@bafolts This is deployed at https://tm-beta.lemur.nc:8000/ (select Turmoil >> Agendas >> Random / Chairman).

Functionally this is ready except for a UI issue with Chairman mode (component does not re-render) that @kberg is looking into. I will also be adding unit tests to cover the individual policies and bonuses.

bafolts:
I am not familiar with this expansion but I will review for anything I see that would impact the server or games played with the other expansions.

kberg:
I'm supposed to be looking into this today, right? "Ready by Christmas"
right? OK. Tonight.

src/Game.ts
```diff
@@ -1173,15 +1181,31 @@ export class Game implements ISerializable<SerializedGame> {
...
+      this.oxygenLevel = Math.max(constants.MIN_OXYGEN_LEVEL, this.oxygenLevel + increments);
+      return undefined;
+    }
+
     // Literal typing makes |increments| a const
     const steps = Math.min(increments, constants.MAX_OXYGEN_LEVEL - this.oxygenLevel);
 
     if (this.phase !== Phase.SOLAR) {
+      // PoliticalAgendas Reds P4 hook
+      if (PartyHooks.shouldApplyPolicy(this, PartyName.REDS, 'rp04')) {
```
bafolts, line 1200: I would throw these `rp04`, `sp03`, etc in a `TurmoilPolicy` or similarly named enum. Will make this easier to understand what this `rp04` is and it's significance.

src/Player.ts
```diff
@@ -2118,16 +2126,124 @@ export class Player implements ISerializable<SerializedPlayer> {
...
+          ),
+        );
+      }
     }
 
-    // Turmoil Kelvinists capacity
-    if (this.canAfford(10) && PartyHooks.shouldApplyPolicy(game, PartyName.KELVINISTS)) {
-      action.options.push(this.turmoilKelvinistsAction(game));
+    // Turmoil Kelvinists action
+    if (PartyHooks.shouldApplyPolicy(game, PartyName.KELVINISTS)) {
```
bafolts, line 2145: I like what the ares expansion did where it his this logic inside `src/ares/AresHandler`. I would create a `TurmoilHandler` or use `Turmoil` and add these actions from there. This will cut down on how much code goes directly into `Game`. Here if a function were provided `action.options`, `game`, and `player` it could generate all of this from within `Turmoil` and outside of `Game`.

src/SerializedPlayer.ts
```diff
@@ -59,7 +61,7 @@ export interface SerializedPlayer {
     titaniumProduction: number;
     titaniumValue: number;
     tradesThisTurn: number;
+    turmoilPolicyActionUsed: boolean;
     usedUndo: boolean;
-    turmoilScientistsActionUsed: boolean;
```
bafolts, line 63: This rename will need to be considered when deserializing or accept that when this is first merged players will lose this data during the transition. Losing the data during the transition is fine with me but otherwise will need the comment indicating we can remove this by some date or that it is deprecated.

src/Player.ts
kberg: I'm doing the same thing with MoonHandler, with a function isMoon(game, (moonData) => void) fwiw

src/components/CreateGameForm.ts
```diff
@@ -246,6 +249,28 @@ export const CreateGameForm = Vue.component('create-game-form', {
         return RandomMAOptionType.NONE;
       }
     },
+    isPoliticalAgendasExtensionEnabled: function(): Boolean {
+      return this.politicalAgendasExtension !== AgendaStyle.STANDARD;
+    },
+    politicalAgendasExtensionToggle: function() {
+      const component = (this as any) as CreateGameModel;
```
bafolts, line 256: This conversion is unnecessary. Should be fine to use `this` directly and avoid creating `component` and double setting both `component` and `this`.

src/components/CreateGameForm.ts
```diff
@@ -246,6 +249,28 @@ export const CreateGameForm = Vue.component('create-game-form', {
...
+        this.politicalAgendasExtension = AgendaStyle.STANDARD;
+      }
+    },
+    getPoliticalAgendasExtensionAgendaStyle: function(type: 'random' | 'chairman'): AgendaStyle {
+      if (type === 'random') {
+        return AgendaStyle.RANDOM;
+      } else if (type === 'chairman') {
+        return AgendaStyle.CHAIRMAN;
+      } else {
+        return AgendaStyle.STANDARD;
```
bafolts, line 271: Given that `type` can be only `random` or `chairman` this remaining `else` should never get hit. I would remove it or `console.warn` that we aren't expecting to reach that code block.

src/components/Turmoil.ts
```diff
@@ -17,68 +17,186 @@ export const Turmoil = Vue.component('turmoil', {
       }
       return party.toLowerCase().split(' ').join('_');
     },
-    getBonus: function(party: PartyName) {
+    getBonus: function(party: PartyName, staticAgendas: PoliticalAgendasModel | undefined) {
       if (party === PartyName.MARS) {
-        return `<div class="resource money party-resource">1</div> / 
+        if (staticAgendas !== undefined && staticAgendas.marsFirstBonus === 'mb02') {
```
bafolts, line 22: This should all be ported into the vue template string at some point. Using `v-html` loses almost all the benefits that vue offers. These should be vue components instead of functions on this component. A typo in this HTML will go unnoticed but would be caught by the vue type checker if it were not using `v-html`. With `v-html` as far as vue cares this string will be placed as the html of the parent element regardless of what it contains.

src/components/Turmoil.ts
```diff
@@ -17,68 +17,186 @@ export const Turmoil = Vue.component('turmoil', {
...
+          <div class="tile oxygen-tile req-tile-small" style="margin: 10px -5px;"></div>
+          <div class="tile ocean-tile req-tile-small"></div>
+          <div class="tile temperature-tile req-tile-small"></div>
+          : <div class="resource card card-with-border"></div></span>`;
+        } else if (staticAgendas !== undefined && staticAgendas.scientistsPolicy === 'sp04') {
+          return `<div class="scientists-requisite"><div class="resource-tag tag-science party-resource-tag"></div></div>`;
+        } else {
+          return `<span class="money resource">10</span>
+            <span class="red-arrow"></span>
+            <span class="card card-with-border resource party-resource"></span>
```
bafolts, line 117: This string is repeated three times. I would generate it with a for loop.

src/deferredActions/ChoosePoliticaAgenda.ts
```diff
@@ -0,0 +1,47 @@
...
+import {SelectOption} from '../inputs/SelectOption';
+import {Game} from '../Game';
+import {Player} from '../Player';
+import {PlayerInput} from '../PlayerInput';
+import {DeferredAction} from './DeferredAction';
+import {IParty} from '../turmoil/parties/IParty';
+import {Agenda} from '../turmoil/PoliticalAgendas';
+import {Turmoil} from '../turmoil/Turmoil';
+
+export class ChoosePoliticalAgenda implements DeferredAction {
```
bafolts, line 12: I would move this file to `src/turmoil/deferredActions` as it is a deferred action for turmoil

src/server/ServerModel.ts
```diff
@@ -494,6 +497,25 @@ function getTurmoil(game: Game): TurmoilModel | undefined {
       };
     }
 
+    const staticAgendas = game.turmoil.politicalAgendasData.staticAgendas;
+    let staticAgendasData;
+    if (staticAgendas) {
+      staticAgendasData = {
+        marsFirstPolicy: staticAgendas.get(PartyName.MARS)!.policyId,
```
bafolts, line 504: I don't think this is a good use case for `Map`. This is assuming that the map is going to have all of these entries. I would instead create an interface with `mars`, `scientists`, `unity`, `greens`, `reds`, and `kelvinists` as properties that have to exist. Then the usage of the `!` operator will be unnecessary.

src/turmoil/Bonus.ts
```diff
@@ -0,0 +1,10 @@
+import {Game} from '../Game';
+
+export type BonusId = string;
+
+export interface Bonus {
+  id: BonusId;
+  description: string;
+  isDefault?: boolean;
```
bafolts, line 8: I would use `isDefault` and make this property required. A value of `true` indicates it is the default and `false` indicates that it is not. Otherwise what does the property `isDefault` not existing on `Bonus` represent?

src/turmoil/Policy.ts
```diff
@@ -0,0 +1,18 @@
...
+import {OrOptions} from '../inputs/OrOptions';
+import {ISpace} from '../ISpace';
+import {Player} from '../Player';
+
+export type PolicyId = string;
+
+export interface Policy {
+  id: PolicyId;
+  description: string;
+  isDefault?: boolean;
```
bafolts, line 12: I would make this property required and provide `false` when it isn't default.

src/deferredActions/ChoosePoliticaAgenda.ts
nwai90: We could. This would place deferredActions in several places though, making them harder to locate. Perhaps it is better  to have them all in the same folder with subfolders? e.g. `src/deferredActions/turmoil`, `src/deferredActions/colonies`, etc.

src/turmoil/Bonus.ts
nwai90: Agree. Will make it a required property

src/turmoil/Policy.ts
nwai90: Agree. Will make it a required property

src/SerializedPlayer.ts
nwai90: The property has been removed entirely and replaced with `turmoilPolicyActionUsed`. Some data will be lost during transition which is probably fine with advance communication unless we maintain both properties with the intent to deprecate `turmoilScientistsActionUsed` in future instead

src/Player.ts
nwai90: Will move to TurmoilHandler 👍

src/Game.ts
nwai90: 👍 Moved identifiers to `TurmoilPolicy`

nwai90:
@bafolts This is ready for re-review, could you have a look at it again?

bafolts:
I will look at this tomorrow. Seems to have a conflict.

src/models/TurmoilModel.ts
```diff
@@ -1,36 +1,195 @@
...
+      ruling = game.turmoil.rulingParty.name;
+    }
+
+    const lobby = Array.from(
+      game.turmoil.lobby,
+      (player) => game.getPlayerById(player).color,
+    );
+
+    const reserve = game.turmoil.getPresentPlayers().map((player) => {
+      const number = game.turmoil!.getDelegates(player);
```
bafolts, line 78: To avoid using `game.turmoil` I would create a `const turmoil = game.turmoil;` in the outer function. At that point you don't have to assume `game.turmoil` will be defined. This has to assume `game.turmoil` exists because the typescript compiler doesn't know when this callback will be called.

src/models/TurmoilModel.ts
```diff
@@ -1,36 +1,195 @@
...
+        scientists: getStaticAgenda(PartyName.SCIENTISTS),
+        unity: getStaticAgenda(PartyName.UNITY),
+        greens: getStaticAgenda(PartyName.GREENS),
+        reds: getStaticAgenda(PartyName.REDS),
+        kelvinists: getStaticAgenda(PartyName.KELVINISTS),
+      };
+    }
+
+    return {
+      chairman: chairman,
```
bafolts, line 144: Minor but this entire return can be simplified leading to less typescript we have to write:

```
return {
   chairman,
   ruling,
   dominant,
   parties,
   lobby,
   reserve,
   distant,
   ```

**Review by bafolts — APPROVED** [review]
As a follow up, as far as I can tell none of the `Policy` need state so we don't really need to be calling `new` to create new instances. This is using more memory and CPU than we need. One instance would work. Instead of exporting the class we could export one instance of the class. Also related to the issue filed with getting some of the expansion logic out of `Player` and `Game` there is an idea I will start to tinker with.

Would be better for memory to only export the constant instances of the classes. For example with `src/turmoil/parties/Scientists` we would only export one instance of each bonus, policy, and party.

```
export SCIENTISTS_BONUS_1 = new ScientistsBonus01();
export SCIENTISTS_PARTY = new Scientists();
```

Unless I am missing something and these aren't constants, then disregard.

kberg:
WOO HOO!

src/models/TurmoilModel.ts
nwai90: @bafolts You can have a relook now. All comments are addressed and conflicts are resolved except this one (not sure what it means)

chosta:
@nwai90 Is this ready to roll?

nwai90:
> @nwai90 Is this ready to roll?

Yes all good, conflicts and PR comments are addressed

chosta:
Congratulations! Amazing work!

kberg:
\o/

PR #1900, opened 2020-11-15
allow land claim to place on south pole

bafolts:
This allows `LandClaim` to reserve the south pole at all times.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,18 +40,19 @@
     "extends": "@istanbuljs/nyc-config-typescript"
   },
   "scripts": {
+    "build": "npm run compile && npm run lint && npm run templates && npm run webpack",
     "clean": "rm -r ./build/ ./styles.css || true",
     "compile": "tsc",
-    "prep": "node make_static_json.js && lessc src/styles/common.less styles.css",
-    "build": "npm run prep && npm run compile && npm run lint && npm run templates && npm run webpack",
-    "start": "node build/server.js",
+    "cover": "nyc mocha --file build/tests/utils/Vue.js --recursive build/tests",
+    "fix": "eslint --fix src server.ts script.ts tests --ext ts",
     "lint": "eslint src server.ts script.ts tests --ext ts",
+    "precover": "tsc --build tsconfig-test.json",
+    "prebuild": "node make_static_json.js && lessc src/styles/common.less styles.css",
     "pretest": "tsc --build tsconfig-test.json",
+    "start": "node build/server.js",
     "templates": "node compile-vue-templates.js && tsc --build tsconfig-vue.json",
     "test": "mocha --file build/tests/utils/Vue.js --recursive build/tests",
     "testgiven": "tsc --build tsconfig.json && npm run webpack && tsc --build tsconfig-test.json && mocha",
-    "precover": "tsc --build tsconfig-test.json",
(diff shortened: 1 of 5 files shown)


tests/utils/Vue.ts
```diff
@@ -29,3 +30,18 @@ const Vue = require('vue');
 // setup for default exports
 Vue.default = Vue;
 
+// don't save to database during tests
+Database.getInstance = function () {
```
bafolts, line 34: While also looking at how we serialise data and communicate with database I realised that during a unit test run the SQLite database is storing the test data! Adding a mock database here to avoid that overhead.

package.json
```diff
@@ -40,18 +40,19 @@
     "extends": "@istanbuljs/nyc-config-typescript"
   },
   "scripts": {
+    "build": "npm run compile && npm run lint && npm run templates && npm run webpack",
     "clean": "rm -r ./build/ ./styles.css || true",
     "compile": "tsc",
-    "prep": "node make_static_json.js && lessc src/styles/common.less styles.css",
-    "build": "npm run prep && npm run compile && npm run lint && npm run templates && npm run webpack",
-    "start": "node build/server.js",
+    "cover": "nyc mocha --file build/tests/utils/Vue.js --recursive build/tests",
```
kberg, line 46: Why is this file change related to land claim?

src/cards/LandClaim.ts
```diff
@@ -17,10 +20,19 @@ export class LandClaim implements IProjectCard {
     public canPlay(player: Player, game: Game): boolean {
       return game.board.getAvailableSpacesOnLand(player).length > 0;
```
kberg, line 21: Should this also call the new function?

src/cards/LandClaim.ts
```diff
@@ -17,10 +20,19 @@ export class LandClaim implements IProjectCard {
...
     }
+    private getNonReservedLandSpaces(game: Game): Array<ISpace> {
+      return game.board.spaces.filter((space) => {
```
kberg, line 24: Under what circumstance is this algorithm only appropriate for this local use case? I wonder why this isn't centralized in Board.ts for example

package.json
bafolts: I had linting errors while working on this solution so I added `fix` which is a script `gts` has which fixes linting errors. While in this file I put these in alphabetical order. They only changed order.

src/cards/LandClaim.ts
bafolts: I moved this to `Board.ts`. Since `HellasBoard` always calls to filter `SpaceName.HELLAS_OCEAN_TILE` this separate function made the most sense to me. Land claim is a fairly unique project card. Moving the function to `Board` lessens the number of imports `LandClaim` has to take on though.

tests/utils/Vue.ts
kberg: 😂😂😂😆😁😄😃🙂🤔😐🙁😦😦😢😢😭😭😭

kberg (approved):

bafolts:
Randomly placed tiles for single player game are sometimes place a tile on `SpaceName.HELLAS_OCEAN_TILE` and the test is failing, putting up fix.

PR #1901, opened 2020-11-15
chore: upgrade sqlite3 to 5.0.0

bafolts:
pre-built binaries for sqlite3 seem to be missing for some build environments. Upgrading to latest hoping for better support.


kberg (approved):

bafolts:
This change seems to sometimes cause `npm run cover` to fail during builds. I think since the mock database was added with unit tests this change will stop failing intermittently. Will need to rollback from `5.0.0` if this continues to fail builds.

PR #1907, opened 2020-11-16
Delay messages about server start

alrusdi:
Change order of messages appearing when server starts.
First show all the lines like "load game ..."
and then "Starting server on port..." to make game overview links more available to copy

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -1084,19 +1084,19 @@ function serveResource(res: http.ServerResponse, s: Buffer): void {
   res.end();
 }
 
-gameLoader.start();
+gameLoader.start(() => {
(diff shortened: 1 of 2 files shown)


kberg (approved):
Smart.

bafolts:
This will break heroku if it takes longer than 30 seconds to load all games. The server needs to start listening as soon as possible. The `GameLoader` is designed to collect requests until a game is loaded. Waiting to listen until every game has loaded defeats the purpose. While heroku is switching over users will get some 4xx or 5xx response instead of the open connection until we have the game loaded. I would roll this back or at least only log that the server has started later versus listening later as well.

PR #1908, opened 2020-11-16
Compiller flag to enable incremental builds

alrusdi:
It improves compilation time for consequential builds. I believe it can speed up "cover" step on CI too

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -16,7 +16,8 @@
         "target": "es5",
         "types": ["node"],
         "moduleResolution": "node",
-        "resolveJsonModule": true
+        "resolveJsonModule": true,
+        "incremental": true
     },
     "include": [
         "src/**/*.ts",


kberg (approved):
Incremental builds?? Yum!

bafolts:
As far as I know this was already on as `true` is the default.

```
--incremental | boolean | true if composite is on, false otherwise | Enable incremental compilation by reading/writing information from prior compilations to a file on disk. This file is controlled by the --tsBuildInfoFile flag.
```

alrusdi:
Strange. I've found much faster compilation times after this change. Revert it then

bafolts:
I don't think it does any harm being set to `true`. Since `composite` defaults to `true` I am pretty sure it is irrelevant.

alrusdi:
@bafolts have you seen this project https://github.com/pinterest/esprint 
It may speed up eslint little bit

PR #1911, opened 2020-11-17
manually write out game toJSON for serializing

bafolts:
We have been relying on whatever `JSON.stringify` happens to return when serializing data for database. This leads to larger payloads than needed and in some cases isn't even working. This first step will define all the properties we send to database and get back from database. From there plan to fix use cases where we are losing data during the process. This will help with restoring games from database.


bafolts:
Marking this as draft while I sanity check it. I will also run every game from production database through it to make sure they don't log warnings.

kberg:
Yay!

kberg:
I'll give this a look tomorrow unless someone else grabs it first.

bafolts:
I gave this another pass through tonight. Getting more confident in it. It loads games from database without logging a warning. Will do some undo operation trial runs tomorrow.

bafolts:
Added unit tests to verify each property of a class ended up serialized.

src/ILoadable.ts
```diff
@@ -1,3 +1,4 @@
 export interface ILoadable<T, T2> {
-    loadFromJSON(d: T): T2;
+  loadFromJSON(d: T): T2;
+  serialize(): T;
```
kberg, line 3: s/ILoadable/ISerializable? Another day.

src/SerializedPlayer.ts
```diff
@@ -10,11 +12,11 @@ export interface SerializedPlayer {
     color: Color;
```
kberg, line 12: If you're reordering SerializedGame's attributes to be in alphabetical order, can you do that here, too? Next time is OK.

kberg (commented):
Hard to tell how correct everything is, but I don't see anything meaningfully peculiar.

bafolts:
> Hard to tell how correct everything is, but I don't see anything meaningfully peculiar.

Yes, this stuff is going to be painful to update. Feeling comfortable this will not break game. We call to serialize during most actions. Not sure how many people even use undo. Hoping to get the undo operation and re-loading of games from database much more reliable. This will be first step of many.

I made the other suggestions.

bafolts:
Gave this a few manual sanity checks. See games load from production database fine. This will end up storing smaller payloads per each save. Old games I believe will pull over fine. That transition from the old saved format to the new format is where I see potential problems coming up. The added unit tests will ensure I didn't miss any keys. When adding keys in the future the tests will fail to remind developers they need to consider the serialized equivalent of a class. Plan to add deferred actions and the "waiting for" to the serialized data.

kberg:
>  When adding keys in the future the tests will fail to remind developers they need to consider the serialized equivalent of a class.

I believe you mean that the tests will fail, thereby reminding developers. Not "fail to remind" :D

src/cards/IProjectCard.ts
```diff
@@ -12,4 +13,5 @@ export interface IProjectCard extends ICard {
     resourceType?: ResourceType;
     hasRequirements?: boolean;
     bonusResource?: Resources | undefined;
+    targetCards?: Array<RobotCard>;
```
kberg, line 16: IProjectCard containing a reference to RobotCard, which subclasses IProjectCard? Seems ... you sure?

tests/Dealer.spec.ts
```diff
@@ -0,0 +1,15 @@
...
+
+describe('Dealer', function() {
+  it('serializes every property', function() {
+    const dealer = new Dealer(false, false, false, false, false, false, false, false);
+    const serialized = dealer.serialize();
+    const serializedKeys = Object.keys(serialized);
+    const dealerKeys = Object.keys(dealer);
+    serializedKeys.sort();
+    dealerKeys.sort();
+    expect(serializedKeys).to.deep.eq(dealerKeys);
```
kberg, line 12: A crazy idea I have for tests: After every step in a game, serialize and deserialize. After. Every. Step. See which tests fail.

kberg (approved):
Just tell me that the replicating robots thing is ok.

src/cards/IProjectCard.ts
bafolts: I moved this off of `IProjectCard` and referenced `SelfReplicatingRobots` when serializing as it is only used for that card.

tests/Dealer.spec.ts
bafolts: This would catch when the serialization or deserialization were going to fail and would have value. Would need a way to introduce it which probably wouldn't be that bad given in javascript you can override anything you want. The other tedious part of the serialized data is the old data. When migrating to a new data format we have to modify things slowly to not break older games. Then after some time delete the older games. If there is some each way to modify every `ISerializable` to run this via `afterEach` it would be well worth it so developers keep it in mind.

PR #1912, opened 2020-11-17
Card effect to Vue

chosta:
* CardRenderEffect implementation following the builder pattern. A card effect has 4 parts `cause - delimiter - effect - description`
* Implemented some effect cards and added some new builder functions

I tried to reduce boilerplate and focused on clean API, just not sure about the `startEffect` function. I wanted to announce end of cause and start of effect part with a function that also generates two new arrays which then can be fetched via getters in the Vue component. It does do the job and hopefully it's intuitive.
```
  public get startEffect(): Builder {
    this.br;
    this._addSymbol(CardRenderSymbol.colon());
    this.br;
    return this;
  }
```

```
...
public get cause(): Array<ItemType> | undefined {
    this._validate();
    return this.rows[0];
  }

  public get delimiter(): ItemType {
    this._validate();
    if (this.cause?.length === 0) {
      return undefined;
    }
    return this.rows[1][0];
  }
...
```

### DIFF
--- a/src/cards/render/CardRenderer.ts
+++ b/src/cards/render/CardRenderer.ts
@@ -78,7 +78,7 @@ export class CardRenderEffect extends CardRenderer {
   public get description(): ItemType {
     this._validate();
     // TODO (chosta): validate builder method to make sure it's the last element
-    return `Effect: ${this.rows[2].pop()}`;
+    return `Effect: ${this.rows[2].slice(-1)[0]}`;
   }
 }
 
@@ -268,16 +268,16 @@ class Builder {
 
     const row = this._getCurrentRow();
     if (row !== undefined) {
-      const item = row?.pop();
+      const item = row.pop();
+      if (item === undefined) {
+        throw new Error('Called "any" without a CardRenderItem.');
+      }
       if (!(item instanceof CardRenderItem)) {
         throw new Error('"any" could be called on CardRenderItem only');
       }
 
-      if (item === undefined) {
-        throw new Error('Called "any" without a CardRenderItem.');
-      }
       item.anyPlayer = true;
-      row?.push(item);
+      row.push(item);
       this._data.push(row);
     }
 
@@ -293,7 +293,7 @@ class Builder {
 
     const row = this._getCurrentRow();
     if (row !== undefined) {
-      const item = row?.pop();
+      const item = row.pop();
       if (!(item instanceof CardRenderItem)) {
         throw new Error('"played" could be called on CardRenderItem only');
       }
@@ -302,7 +302,7 @@ class Builder {
         throw new Error('Called "played" without a CardRenderItem.');
       }
       item.isPlayed = true;
-      row?.push(item);
+      row.push(item);
       this._data.push(row);
     }
 
--- a/src/cards/render/CardRenderer.ts
+++ b/src/cards/render/CardRenderer.ts
@@ -3,7 +3,7 @@ import {CardRenderSymbol} from './CardRenderSymbol';
 import {CardRenderItemSize} from './CardRenderItemSize';
 import {CardRenderItemType} from './CardRenderItemType';
 
-type ItemType = CardRenderItem | CardRenderProductionBox | CardRenderSymbol | undefined;
+type ItemType = CardRenderItem | CardRenderProductionBox | CardRenderSymbol | CardRenderEffect | string | undefined;
 
 export class CardRenderer {
   constructor(protected _rows: Array<Array<ItemType>> = [[]]) {}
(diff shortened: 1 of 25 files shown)


src/cards/render/CardRenderer.ts
```diff
@@ -23,142 +23,258 @@ export class CardRenderProductionBox extends CardRenderer {
...
+
+  public get effect(): Array<ItemType> {
+    this._validate();
+    return this.rows[2];
+  }
+
+  public get description(): ItemType {
+    this._validate();
+    // TODO (chosta): validate builder method to make sure it's the last element
+    return `Effect: ${this.rows[2].slice(-1)[0]}`;
```
bafolts, line 81: This seems equivalent to using `this.rows[2].pop()`.

src/cards/render/CardRenderer.ts
```diff
@@ -167,6 +283,41 @@ class Builder {
...
+  /**
+   * Mark the last item in the queue as 'isPlayed'
+   * e.g. titanium().played will result in a resource circle instead of square
+   */
+  public get played(): Builder {
+    this._checkExistingItem();
+
+    const row = this._getCurrentRow();
+    if (row !== undefined) {
+      const item = row?.pop();
```
bafolts, line 296: Since this is inside a block which has checked that `row !== undefined` the `row?` should be unnecessary.

src/cards/render/CardRenderer.ts
```diff
@@ -167,6 +283,41 @@ class Builder {
...
+      if (!(item instanceof CardRenderItem)) {
```
bafolts, line 297: The check that `item === undefined` should be placed first. When item is `undefined` this will return `true`. The check for `item` being `undefined` will never be reached.

src/cards/turmoil/EventAnalysts.ts
```diff
@@ -5,25 +5,34 @@ import {CardType} from '../CardType';
...
+      game.turmoil.addInfluenceBonus(player);
     }
+    return undefined;
+  }
+
+  public metadata: CardMetadata = {
+    description: 'Requires that Scientists are ruling or that you have 2 delegates there',
+    cardNumber: 'T05',
+    requirements: CardRequirements.builder((b) => b.party(PartyName.SCIENTISTS)),
+    renderData: CardRenderer.builder((b) => b.effectBox((be) => be.startEffect.influence(1).description('You have  +1 influence'))),
```
bafolts, line 36: Should probably only be one space character between `have` and `+1`.

src/cards/render/CardRenderer.ts
chosta: Funny thing is that they seem to be equal with the difference that `.pop` actually removes the item from the array so I decided to not close that door but it's probably better to have a `.pop` here I do agree.

src/cards/render/CardRenderer.ts
```diff
@@ -23,142 +23,258 @@ export class CardRenderProductionBox extends CardRenderer {
...
+
+  public get effect(): Array<ItemType> {
+    this._validate();
+    return this.rows[2];
+  }
+
+  public get description(): ItemType {
+    this._validate();
+    // TODO (chosta): validate builder method to make sure it's the last element
+    return `Effect: ${this.rows[2].pop()}`;
```
bafolts, line 81: I didn't realise this was trying to pull the last value without modifying state. I would be surprised if `pop` could be used as now calling to `description` is going to modify the state of `rows[2]`.

src/components/card/CardRenderEffectBoxComponent.ts
```diff
@@ -0,0 +1,60 @@
...
+import {CardRenderSymbolComponent} from './CardRenderSymbolComponent';
+import {CardRenderEffect, CardRenderProductionBox} from '../../cards/render/CardRenderer';
+import {CardProductionBoxComponent} from './CardProductionBoxComponent';
+import {CardRenderSymbol} from '../../cards/render/CardRenderSymbol';
+import {CardRenderItem} from '../../cards/render/CardRenderItem';
+import {CardDescription} from './CardDescription';
+
+export const CardRenderEffectBoxComponent = Vue.component('CardRenderEffectBoxComponent', {
+  props: {
+    data: {
```
bafolts, line 12: I would use another name besides `data`. Vue has it's own `data` function to pull the `data` of a component. As it ends up being of type `CardRenderEffect` using `effect` would avoid this naming collision.

bafolts (changes requested):
I was confused by a `slice(-1)[0]` not realizing that was performing a peek to see the last item of the array versus modifying the state of the array. I am not sure `pop` can be used in place as that will modify the state of the array.

I had minor suggestions other than that but we need to avoid using `data` as the name of a property within vue as it is confusing since vue has it's own `data` property which is supposed to be a function to return the data of the component. I would change the name `data` on the component to some name that doesn't look like a part of vue.

src/cards/render/CardRenderer.ts
chosta: VScode does it automatically for me. Didn't think about that. Thanks

src/cards/render/CardRenderer.ts
chosta: Interesting. TS didn't infer that this was a problem. I guess it's not almighty after all.

src/components/card/CardRenderEffectBoxComponent.ts
chosta: Agree

bafolts (approved):

PR #1919, opened 2020-11-19
Refactor global parameter increase methods

Lynesth:
Each method is refatored in its own commit.


nwai90 (approved):

src/Game.ts
```diff
@@ -1172,122 +1172,96 @@ export class Game implements ILoadable<SerializedGame, Game> {
       player.takeAction(this);
     }
 
-    public increaseOxygenLevel(
-        player: Player, steps: 1 | 2): undefined {
+    public increaseOxygenLevel(player: Player, _steps: 1 | 2): undefined {
```
kberg, line 1175: Changing `steps` to `_steps` is probably a bad idea: as far as I can tell it's a standard that function parameters that start with `_` imply they are unused, which is not the case here.

kberg (changes requested):

src/Game.ts
Lynesth: I changed it.

kberg (approved):

PR #1923, opened 2020-11-19
Translation de

Starcalc:
This commit translates all as-of-now missing titles for all the cards. The description will follow another time. :-)

I'm not sure how to remove the wrong commits that were queued, but this really shouldn't mess up anything.

I hope that "main" is the correct target, as it is marked as "default".


src/components/CreateGameForm.ts
```diff
@@ -664,7 +664,7 @@ export const CreateGameForm = Vue.component('create-game-form', {
                                 </label>
                                 <input type="checkbox" v-model="requiresVenusTrackCompletion" id="requiresVenusTrackCompletion-checkbox">
                                 <label for="requiresVenusTrackCompletion-checkbox">
-                                    <span v-i18n>Venus Terraforming</span>&nbsp;<a href="https://github.com/bafolts/terraforming-mars/wiki/Variants#venus-terraforming" class="tooltip" target="_blank">&#9432;</a>
```
chosta: `main` replaces the `master` branch as of today
It seems that you reverted the change that added the tooltip for the Venus Terraforming after the merge if I'm reading this correct. Please, make sure we have it shown.

chosta (changes requested):

src/components/CreateGameForm.ts
Starcalc: I've added the tooltip in "my" repository and reverted "my" entry. #1920 isn't merged yet, so this is correct.

chosta: Ok

chosta (approved):

PR #1924, opened 2020-11-20
Reformat compile-vue-templates so the VSC editor doesn't complain.

kberg:


bafolts (approved):

bafolts:
Let me change this setting where it must be up to date. Too cumbersome.

PR #1927, opened 2020-11-20
Cards to Vue - all starting with letter A

chosta:
The first PR handling all common cards starting with A. I'm doing it in this order so reviewers can easily pull and compare with the current cards (the best way to spot inaccuracies). Also, I want to finish this task until Christmas, so we have 25 more letters to go + Preludes and Corporations.
* Added CardsVictoryPoints component handling special points per amount cases
* Added secondary tag functionality
* Added text and special text functionality
* Several improvements, refactoring
* ~10 new cards

### DIFF
--- a/src/cards/AdaptationTechnology.ts
+++ b/src/cards/AdaptationTechnology.ts
@@ -2,6 +2,8 @@ import {IProjectCard} from './IProjectCard';
 import {Tags} from './Tags';
 import {CardType} from './CardType';
 import {CardName} from '../CardName';
+import {CardMetadata} from './../cards/CardMetadata';
+import {CardRenderer} from './../cards/render/CardRenderer';
 
 export class AdaptationTechnology implements IProjectCard {
   public cost = 12;
@@ -17,5 +19,15 @@ export class AdaptationTechnology implements IProjectCard {
   }
   public getVictoryPoints() {
     return 1;
+  }
+  public metadata: CardMetadata = {
+    cardNumber: '153',
+    renderData: CardRenderer.builder((b) => {
+      b.effectBox((eb) => {
+        eb.plate('Global requirements').startEffect.text('+/- 2');
+        eb.description('Effect: Your global requirements are +2 or -2 steps, your choice in each case.');
+      });
+    }),
+    victoryPoints: 1,
   }
 }
--- a/src/cards/AdaptationTechnology.ts
+++ b/src/cards/AdaptationTechnology.ts
@@ -2,8 +2,8 @@ import {IProjectCard} from './IProjectCard';
 import {Tags} from './Tags';
 import {CardType} from './CardType';
 import {CardName} from '../CardName';
-import {CardMetadata} from './../cards/CardMetadata';
-import {CardRenderer} from './../cards/render/CardRenderer';
+import {CardMetadata} from './CardMetadata';
+import {CardRenderer} from './render/CardRenderer';
 
 export class AdaptationTechnology implements IProjectCard {
   public cost = 12;
--- a/src/cards/AdaptationTechnology.ts
+++ b/src/cards/AdaptationTechnology.ts
@@ -4,18 +4,18 @@ import {CardType} from './CardType';
 import {CardName} from '../CardName';
 
 export class AdaptationTechnology implements IProjectCard {
-    public cost = 12;
-    public tags = [Tags.SCIENCE];
-    public name = CardName.ADAPTATION_TECHNOLOGY;
-    public cardType = CardType.ACTIVE;
+  public cost = 12;
+  public tags = [Tags.SCIENCE];
+  public name = CardName.ADAPTATION_TECHNOLOGY;
+  public cardType = CardType.ACTIVE;
 
-    public getRequirementBonus(): number {
-      return 2;
-    }
-    public play() {
-      return undefined;
-    }
-    public getVictoryPoints() {
-      return 1;
-    }
+  public getRequirementBonus(): number {
+    return 2;
+  }
+  public play() {
+    return undefined;
+  }
+  public getVictoryPoints() {
+    return 1;
+  }
 }
--- a/src/cards/promo/AsteroidDeflectionSystem.ts
+++ b/src/cards/promo/AsteroidDeflectionSystem.ts
@@ -9,36 +9,36 @@ import {Resources} from '../../Resources';
 import {Game} from '../../Game';
 
 export class AsteroidDeflectionSystem implements IActionCard, IProjectCard, IResourceCard {
-    public name = CardName.ASTEROID_DEFLECTION_SYSTEM;
-    public cost = 13;
-    public tags = [Tags.SPACE, Tags.EARTH, Tags.STEEL];
-    public resourceType = ResourceType.ASTEROID;
-    public resourceCount: number = 0;
-    public cardType = CardType.ACTIVE;
-    public hasRequirements = false;
+  public name = CardName.ASTEROID_DEFLECTION_SYSTEM;
+  public cost = 13;
+  public tags = [Tags.SPACE, Tags.EARTH, Tags.STEEL];
+  public resourceType = ResourceType.ASTEROID;
+  public resourceCount: number = 0;
+  public cardType = CardType.ACTIVE;
+  public hasRequirements = false;
 
-    public canPlay(player: Player): boolean {
-      return player.getProduction(Resources.ENERGY) >= 1;
-    }
+  public canPlay(player: Player): boolean {
+    return player.getProduction(Resources.ENERGY) >= 1;
+  }
 
-    public play(player: Player) {
-      player.addProduction(Resources.ENERGY, -1);
-      return undefined;
-    }
+  public play(player: Player) {
+    player.addProduction(Resources.ENERGY, -1);
+    return undefined;
+  }
 
-    public canAct(): boolean {
-      return true;
-    }
+  public canAct(): boolean {
+    return true;
+  }
 
-    public action(player: Player, game: Game) {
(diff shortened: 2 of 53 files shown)


src/cards/promo/AsteroidRights.ts
```diff
@@ -12,88 +12,92 @@ import {SelectCard} from '../../inputs/SelectCard';
 import {OrOptions} from '../../inputs/OrOptions';
 import {SelectOption} from '../../inputs/SelectOption';
 import {SelectHowToPayDeferred} from '../../deferredActions/SelectHowToPayDeferred';
+import {CardMetadata} from '../../cards/CardMetadata';
+import {CardRenderer} from '../../cards/render/CardRenderer';
 
 export class AsteroidRights implements IActionCard, IProjectCard, IResourceCard {
-    public name = CardName.ASTEROID_RIGHTS;
```
kberg, line 17: What generates these reformats? I'd like to apply all of these, now, before edoing more PRs like this, _please._

chosta:
I'm not doing it on purpose I promise :/
I was left with the impression that gts enforces 2 space tabs...That's why
I'm doing it. It's not consistent otherwise. Feel free to reformat all
cards without Metadata to use 2 spaces.

src/cards/promo/AsteroidRights.ts
kberg: > I'm not doing it on purpose I promise :/ I was left with the impression that gts enforces 2 space tabs...That's why I'm doing it. It's not consistent otherwise. Feel free to reformat all cards without Metadata to use 2 spaces.
I get it, for sure. If you can do it, I'll approve really fast. But going through and reformatting the files to use 2-space-indents, I probably wont' get to for some time.

chosta:
Why don't you go through commits then (in order)?This is how I've been
setting it up. First formatting and then all other changes. I tried to make
them meaningful too.

chosta:
I know it's a bit inconvenient to review because of the reformatting but I think it does not need much to merge this one. I'd suggest looking at the non iCard changes as they are more important. It's easier to go through each commit in https://github.com/bafolts/terraforming-mars/pull/1927/commits. Much more readable. I'm putting a bit of pressure here because I have some spare hours and intend to work on the task. Thanks @kberg @bafolts

src/cards/AdaptationTechnology.ts
```diff
@@ -2,20 +2,32 @@ import {IProjectCard} from './IProjectCard';
 import {Tags} from './Tags';
 import {CardType} from './CardType';
 import {CardName} from '../CardName';
+import {CardMetadata} from './../cards/CardMetadata';
```
bafolts, line 5: Since these are already in the `src/cards` directory should be able to use `./CardMetadata` and `./render/CardRenderer` here instead of going one directory back to cards.

src/cards/promo/AsteroidDeflectionSystem.ts
```diff
@@ -7,38 +7,54 @@ import {Tags} from '../Tags';
 import {Player} from '../../Player';
 import {Resources} from '../../Resources';
 import {Game} from '../../Game';
+import {CardMetadata} from '../../cards/CardMetadata';
+import {CardRenderer} from '../../cards/render/CardRenderer';
```
bafolts, line 11: These can be simplified to `../render` or `../CardMetadata`.

src/cards/promo/AsteroidDeflectionSystem.ts
```diff
@@ -7,38 +7,54 @@ import {Tags} from '../Tags';
...
 
-    public getVictoryPoints(): number {
-      return this.resourceCount;
-    }
+  public getVictoryPoints(): number {
+    return this.resourceCount;
+  }
+  public metadata: CardMetadata = {
+    cardNumber: 'X27',
+    // description: 'Add 2 asteroids to this card',
```
bafolts, line 50: Remove before flight.

src/cards/promo/AsteroidHollowing.ts
```diff
@@ -8,33 +8,48 @@ import {Player} from '../../Player';
...
 
-    public getVictoryPoints(): number {
-      return Math.floor(this.resourceCount / 2);
-    }
+  public getVictoryPoints(): number {
+    return Math.floor(this.resourceCount / 2);
+  }
+
+  public metadata: CardMetadata = {
+    cardNumber: 'x13',
```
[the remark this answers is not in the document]
bafolts: Should eventually make this `cardNumber` and enum. Within the enum can document what this `cardNumber` is.

src/cards/promo/AsteroidHollowing.ts
```diff
@@ -8,33 +8,48 @@ import {Player} from '../../Player';
...
+    description: '1VP for each 2 asteroids on this card',
```
bafolts, line 46: I am assuming `CardRenderVictoryPoints` can generate this description. Can most likely remove this description.

src/cards/promo/AsteroidRights.ts
bafolts: I will get to this today.

src/cards/render/CardRenderVictoryPoints.ts
```diff
@@ -0,0 +1,24 @@
+import {CardRenderItemType} from './CardRenderItemType';
+import {CardRenderItem} from './CardRenderItem';
+
+export class CardRenderVictoryPoints {
+  public targetOneOrMore: boolean = false; // marking target to be one or more res (Search for Life)
+  constructor(public item: CardRenderItem, public points: number = 1, public target: number = 1) {
+    if (this.points > this.target) {
```
bafolts, line 7: The `PhysicsComplex` card receives 2 victory points per science resource on the card. If I am understanding this class that card would provide `2` for points and `1` for target. It would trip this error. I don't think this error is providing any value and should be removed.

src/cards/render/CardRenderVictoryPoints.ts
```diff
@@ -0,0 +1,24 @@
...
+      throw new Error(`Illegal victory points setup. Target points have to be greater or equal points and not: ${this.points}/${this.target}`);
+    }
+  }
+
+  public getPointsHtml(): string {
+    if (this.target === this.points) return `${this.target}/`;
```
bafolts, line 13: If a card provides `2` victory points regardless of any resources on the card would it provide `2` for points and `1` for target? Is the `/` suffix a typo?

src/components/card/CardRenderItemComponent.ts
```diff
@@ -105,6 +125,12 @@ export const CardRenderItemComponent = Vue.component('CardRenderItemComponent',
     itemHtmlContent: function(): string {
       // in case of symbols inside
       if (this.item instanceof CardRenderItem && this.item.amountInside) return this.item.amount.toString();
+      if (this.item.isPlate || this.item.text !== undefined) return this.item.text || 'n/a';
+      if (this.item.secondaryTag !== undefined) {
+        const classes: string[] = ['card-icon'];
+        classes.push(`tag-${this.item.secondaryTag}`);
+        return '<div class="' + generateClassString(classes) + '"/>';
```
bafolts, line 132: I would use `<div></div>` instead of `<div />`

https://stackoverflow.com/questions/3558119/are-non-void-self-closing-tags-valid-in-html5

bafolts (approved):
Overall seemed fine. I had minor suggestions regarding the relative imports. Most were going all the way up to the root and then down to cards which is unnecessary.

src/cards/render/CardRenderVictoryPoints.ts
bafolts: Ah, I now see that there is `number | CardRenderVictoryPoints` where the goal is to use `number` when the victory points are static. Could consider always requiring both `points` and `target` to be provided to this class instead of defaulting to `1`. This could help clarify that `number` should be used when the victory points are static. Could also name this `CardRenderDynamicVictoryPoints`.

src/cards/promo/AsteroidHollowing.ts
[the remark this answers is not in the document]
kberg: Actually I'm fairly certain these are the card numbers on the printed cards themselves.

src/cards/promo/AsteroidRights.ts
kberg: How can I help?

bafolts: Will need to:

* Run `npx prettier --write src` and most likely add a `npm run format` script which formats code for developers not using vscode with prettier.
* Solve the parenthesis spacing discrepancy between `eslint` and `prettier`
* Solve `eslint` enforcing two spaces differently than `prettier`.

src/components/card/CardRenderItemComponent.ts
chosta: I wasn't aware. I'll try to not do that although it's kind of a habit.

src/cards/promo/AsteroidRights.ts
chosta: It worked. I still have some autoformatter adjustments to make but it's certainly an improvement. Thanks!

src/cards/promo/AsteroidHollowing.ts
chosta: It can but I'll leave this as a TODO as we want to have a search in description in the `debug-ui` and I want to have the metadata field wherever possible.

src/cards/render/CardRenderVictoryPoints.ts
chosta: Yes, defaulting to 1 seems to be the wrong approach.

PR #1929, opened 2020-11-21
Search for cards in the debug-ui using the description as a filter.

kberg:
To make this work I pre-fetch all the cards, and store data in memory.
This turns out to make searching MUCH faster!

Incidentally, most of the cards do not have text, so while this will seem broken, it surely isn't.

### DIFF
--- a/src/components/DebugUI.ts
+++ b/src/components/DebugUI.ts
@@ -1,14 +1,31 @@
 import Vue from 'vue';
-import {Card, getCardExpansionByName} from './card/Card';
+import {Card} from './card/Card';
 import {
+  ALL_CARD_MANIFESTS,
   ALL_CORPORATION_CARD_NAMES,
   ALL_PRELUDE_CARD_NAMES,
   ALL_PROJECT_CARD_NAMES,
 } from '../cards/AllCards';
 import {GameModule} from '../GameModule';
+import {ICard} from '../cards/ICard';
+
+interface Entry {
+  card: ICard,
+  module: GameModule
+}
+const cards: Map<string, Entry> = new Map();
+ALL_CARD_MANIFESTS.forEach((manifest) => {
+  manifest.projectCards.cards.forEach((card) =>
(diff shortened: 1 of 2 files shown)


src/components/DebugUI.ts
```diff
@@ -1,14 +1,31 @@
...
 } from '../cards/AllCards';
 import {GameModule} from '../GameModule';
+import {ICard} from '../cards/ICard';
+
+interface Entry {
+  card: ICard,
+  module: GameModule
+}
+const cards: Map<string, Entry> = new Map();
+ALL_CARD_MANIFESTS.forEach((manifest) => {
```
chosta, line 17: If I'm reading this correct, there are three identical `.forEach` loops doing the same thing? Is there any hidden purpose to it?

src/components/DebugUI.ts
```diff
@@ -48,15 +65,22 @@ export const DebugUI = Vue.component('debug-ui', {
...
-      if (!nameMatches) {
-        return false;
+      const card = cards.get(cardName);
+      const filterText = this.$data.filterText.toUpperCase();
+      if (this.$data.filterText.length > 0) {
+        if (cardName.toUpperCase().indexOf(filterText) === -1) {
+          return false;
+        }
+        if (this.$data.filterDescription) {
+          const desc = card?.card.metadata?.description;
```
chosta, line 75: I think we can also search through all text inside the card too. I'll write a function for CardRenderer for it.

chosta (approved):
Looks good, just not sure about the identical `forEach` loops and I'm adding a TODO for myself to have a CardRender.getAllText to be used for the global search.

src/components/DebugUI.ts
kberg: I tried joining them as one, but it was a typescript mess.

src/components/DebugUI.ts
kberg: Isn't that what this is?

kberg: (What did I get wrong?)

PR #1930, opened 2020-11-21
Add check boxes for individual expansions in debug-ui.

kberg:
This was content accidentally merged into `main`. :/


alrusdi:
Main branch is main nowadays :-) @bafolts isn't it?

kberg:
Ah are we adopting main? Good.

PR #1932, opened 2020-11-21
Move BiofertilizerFacility to CardRender. Create CardRenderTileComponent, which I'm sure isn't right yet.

kberg:
No question this isn't lining up right, yet. I'm sure the use of the special-tile background isn't going to work, and the sizing is also off. Of course at some point these tiles deserve a refresh.

### DIFF
--- a/src/cards/ares/BiofertilizerFacility.ts
+++ b/src/cards/ares/BiofertilizerFacility.ts
@@ -12,6 +12,9 @@ import {CardType} from '../CardType';
 import {IProjectCard} from '../IProjectCard';
 import {Tags} from '../Tags';
 import {AddResourcesToCard} from '../../deferredActions/AddResourcesToCard';
+import {CardMetadata} from '../CardMetadata';
+import {CardRequirements} from '../CardRequirements';
+import {CardRenderer} from '../render/CardRenderer';
 
 export class BiofertilizerFacility implements IProjectCard {
     public cost = 12;
@@ -19,6 +22,19 @@ export class BiofertilizerFacility implements IProjectCard {
     public cardType = CardType.AUTOMATED;
     public name = CardName.BIOFERTILIZER_FACILITY;
 
+    public metadata: CardMetadata = {
+      description: 'Requires 1 science tag. Increase your plant production 1 step. ' +
+          'Add up to 2 microbes to any card. ' +
+          'Place this tile which grants an <b>adjacency bonus</b> of 1 plant and 1 microbe.',
+      cardNumber: 'A02',
+      requirements: CardRequirements.builder((b) => b.tag(Tags.SCIENCE)),
+      renderData: CardRenderer.builder((b) => {
+        b.productionBox((pb) => pb.plants(1));
+        b.microbes(2);
+        b.br;
+        b.tile('board-space-tile--biofertilizer_facility');
+      }),
+    };
     public canPlay(player: Player, _game: Game): boolean {
       return player.getTagCount(Tags.SCIENCE) >= 1;
     }
--- a/src/cards/render/CardRenderer.ts
+++ b/src/cards/render/CardRenderer.ts
@@ -288,7 +288,6 @@ class Builder {
   }
 
   public tile(selector: string): Builder {
-    this._checkExistingItem();
     this._addTile(selector);
     return this;
   }
--- a/src/cards/render/CardRenderer.ts
+++ b/src/cards/render/CardRenderer.ts
@@ -3,7 +3,7 @@ import {CardRenderSymbol} from './CardRenderSymbol';
 import {CardRenderItemSize} from './CardRenderItemSize';
 import {CardRenderItemType} from './CardRenderItemType';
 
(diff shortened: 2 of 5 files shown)


src/cards/render/CardRenderer.ts
```diff
@@ -107,14 +111,23 @@ class Builder {
     }
   }
 
-  protected _addSymbol(Symbol: CardRenderSymbol): void {
+  protected _addSymbol(symbol: CardRenderSymbol): void {
     const row = this._getCurrentRow();
     if (row !== undefined) {
-      row.push(Symbol);
+      row.push(symbol);
```
chosta, line 117: well spotted

src/cards/render/CardRenderer.ts
```diff
@@ -274,6 +287,12 @@ class Builder {
     return this;
   }
 
+  public tile(selector: string): Builder {
+    this._checkExistingItem();
```
chosta, line 291: I think we can drop `this._checkExistingItem()` here. The purpose of this function is to verify existence of a item that the current function needs to modify in some way. E.g. `plants(-1).any`. In the case of `tile()` we will have an issue with a card like mining area, where the first element of the array should be a tile.

src/components/card/CardRenderTileComponent.ts
```diff
@@ -0,0 +1,20 @@
...
+    },
+  },
+  methods: {
+    getClasses: function(): string {
+      return this.item.selector + ' tile';
+    },
+  },
+  template: `
+  <div class="tile special-tile">
+    <div :class="getClasses()></div>
```
chosta, line 17: I prefer the shorthand `<div :class="getClasses()" />`
Also this one is probably broken as there is a missing second "

src/cards/ares/BiofertilizerFacility.ts
```diff
@@ -12,13 +12,29 @@ import {CardType} from '../CardType';
...
+      description: 'Requires 1 science tag. Increase your plant production 1 step. ' +
+          'Add up to 2 microbes to any card. ' +
+          'Place this tile which grants an <b>adjacency bonus</b> of 1 plant and 1 microbe.',
+      cardNumber: 'A02',
+      requirements: CardRequirements.builder((b) => b.tag(Tags.SCIENCE)),
+      renderData: CardRenderer.builder((b) => {
+        b.productionBox((pb) => pb.plants(1));
+        b.microbes(2);
+        b.br;
+        b.tile('board-space-tile--biofertilizer_facility');
```
chosta, line 35: This could be improved to not use a direct selector but I can do it myself as I wanted to implement the `tile()` in my next PR.

chosta (changes requested):
It's a good step forward. I might expand on this PR after merging to handle all tiles. Will probably have an enum instead of direct selector.
Check my _existingItem comment though

src/components/card/CardRenderTileComponent.ts
kberg: image

I knew that comment looked familiar! :D

src/cards/ares/BiofertilizerFacility.ts
kberg: I was thinking about this. While separating client from server is always good, does that add anything meaningful to this project?

src/components/card/CardRenderTileComponent.ts
chosta: well, it turns out I was wrong all along...as @bafolts pointed out his code review HTML5 is not on friendly terms with this shorthand.
https://stackoverflow.com/questions/3558119/are-non-void-self-closing-tags-valid-in-html5

So yeah, `<div></div>` should be preferred and I need to fight the habit of doing the <div/>

src/cards/ares/BiofertilizerFacility.ts
chosta: Yes. If we change the class-name in the .less file, this would fail. Also, tests.

chosta: You got me thinking about this.To be fair, we ALSO need to set the className inside the Vue component if we didn't in the backend. So changing the `.less` class name would have to be followed by changing it in the Vue component too. It just a frontend pattern (in React, Vue, Angular) to deal with classNames in the component logic. The value is that devs would most likely expect it there.

src/cards/render/CardRenderer.ts
kberg: :+1:

src/cards/render/CardRenderer.ts
kberg: Done.

src/components/card/CardRenderTileComponent.ts
kberg: ack.

src/cards/ares/BiofertilizerFacility.ts
kberg: Whatever you prefer!

kberg:
We're gonna need some tests for these things, too. :/

chosta (approved):

PR #1933, opened 2020-11-22
initialize players in turn order

bafolts:
Initializes players in turn order.


src/Game.ts
```diff
@@ -156,46 +155,35 @@ export class Game implements ISerializable<SerializedGame, Game> {
       public id: string,
       private players: Array<Player>,
       private first: Player,
-      gameOptions?: GameOptions,
```
bafolts, line 159: Easier than dealing with an `undefined` for `gameOptions` I defined it in constructor. The generated javascript writes what we had for us.

src/Game.ts
```diff
@@ -156,46 +155,35 @@ export class Game implements ISerializable<SerializedGame, Game> {
...
-    ) {
-      Database.getInstance();
```
bafolts, line 161: This shouldn't be needed as it returns a value we don't use.

src/Game.ts
```diff
@@ -280,13 +268,15 @@ export class Game implements ISerializable<SerializedGame, Game> {
 
       corporationCards = this.dealer.shuffleCards(corporationCards);
 
+      // Failsafe for exceding corporation pool
```
chosta, line 271: you probably meant `exceeding`

src/Game.ts
```diff
@@ -280,13 +268,15 @@ export class Game implements ISerializable<SerializedGame, Game> {
...
+      if (gameOptions.startingCorporations * players.length > corporationCards.length) {
```
chosta, line 272: I would probably throw an error here and made sure the frontend handles validation. But this doesn't hurt anyway.

chosta (approved):

PR #1936, opened 2020-11-22
in production cache file reads

bafolts:
The https://terraforming-mars.herokuapp.com site seems to be getting more traffic. As we don't cache any resources we get continual requests for file reads. If enough of these come in I believe they are slowing down requests for some users. With this change in production the reads for `fs.readFile` are cached. Also added compression to `styles.css` which takes it down to 40kb from 242kb.

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -5,6 +5,7 @@ import * as http from 'http';
 import * as fs from 'fs';
 import * as path from 'path';
 import * as querystring from 'querystring';
+import * as zlib from 'zlib';
 
 import {AndOptions} from './src/inputs/AndOptions';
 import {BoardName} from './src/BoardName';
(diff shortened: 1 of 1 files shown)


Lynesth:
Why limit the use of the caching to production env ?
Do devs really have the server running and modify assets on the fly and just reload the web page ?

Apart from that, it looks good to me.

bafolts:
Everything used to be cached but was changed to read each time in:

https://github.com/bafolts/terraforming-mars/commit/8acbcb44f5d64b5bf013bf0847be3a9783666114

Assuming people developing want the flexibility to be able to change files and not have to re-start server.

alrusdi (approved):

PR #1944, opened 2020-11-24
display number of cards selected when no cards can be selected

### LINKED ISSUE #1770: Prevent single click "no cards buy" action
Untitled

Very basic confirmation to prevent misclicking and buying 0 cards. 
Even with the extra padding around the button, there is s high chance of misclick on mobile when aiming for the first card.

- if no cards are selected and "buy" is clicked, show a 2nd button (button "buy" become disabled)
- if the 2nd button is displayed, selecting a card removes it

bafolts:
This will display `Buy 0 cards` through `Buy 4 cards` when selecting card to buy. Should resolve #1770.

image
image

May need to update translations.


funkybaboon:
While this is better than nothing, I lost my game by trying to select the far left card and accidentally hitting the "Buy" button. I was fully aware that I need to select cards first. How about an additional Popup that asks if you really want to buy 0 cards?

nwai90:
> While this is better than nothing, I lost my game by trying to select the far left card and accidentally hitting the "Buy" button. I was fully aware that I need to select cards first. How about an additional Popup that asks if you really want to buy 0 cards?

@funkybaboon Popup confirmations are fine if they are optional

kberg:
I like the idea, but this sort of goes against the rest of the selection UI where the button's text is determined by the caller.

kberg:
Reading #1770, consider two buttons: one that says "Skip this action" and the other says "buy these cards" and only one button is active at a ti me?

bafolts:
Updated to use two buttons with one disabled.

2020-11-24-075204_747x351_scrot
2020-11-24-075223_740x345_scrot

Lynesth:
Could you add a skipButtonLabel to be able to customize the text ?

src/components/SelectCard.ts
```diff
@@ -59,6 +58,11 @@ export const SelectCard = Vue.component('select-card', {
         this.playerinput.cards,
       );
     },
+    noneOrMany: function(): boolean {
```
alrusdi, line 61: Strange function name. Can it be renamed to something what it actually does? isManyCardsSelectionRequired? inputNeedsManyCards? multiselectionRequired?

alrusdi (approved):

kevinb9n:
hey! I'm seeing the new "skip this action" button sometimes, but for example am not seeing it when using Business Network. FYI.

bafolts:
Currently the skip this action button is only displayed if more than one card can be selected. Probably makes sense to remove this restriction.

https://github.com/bafolts/terraforming-mars/blob/main/src/components/SelectCard.ts#L63

There was also a suggestion to add what the text should be when the action in skipped. Here for Business Network this would be helpful as the text for skipping should be `Discard`. For situations where only one card can be selected at most it may make sense to make the UI even simpler. Could show a button for "Buy" or "Discard" and not require the user to click on the card at all.

PR #1948, opened 2020-11-24
Cards vue b

chosta:
All cards starting with the letter B
* Added animals to VP class
* Added .tr()
* Improved effect description to accept undefined
* [WIP] Improved the CardRenderTile to use the TileType enum and started separating logic and abstacting css classes (should be done for all tiles in the next PR)

### DIFF
--- a/src/cards/BeamFromAThoriumAsteroid.ts
+++ b/src/cards/BeamFromAThoriumAsteroid.ts
@@ -5,6 +5,9 @@ import {CardType} from './CardType';
 import {Player} from '../Player';
 import {Resources} from '../Resources';
 import {CardName} from '../CardName';
+import {CardMetadata} from './CardMetadata';
+import {CardRequirements} from './CardRequirements';
+import {CardRenderer} from './render/CardRenderer';
 
 export class BeamFromAThoriumAsteroid implements IProjectCard {
     public cost = 32;
@@ -22,5 +25,17 @@ export class BeamFromAThoriumAsteroid implements IProjectCard {
     public getVictoryPoints() {
       return 1;
     }
+    public metadata: CardMetadata = {
+      cardNumber: '058',
+      description: 'Requires a Jovian tag. Increase your heat production and Energy production 3 steps each',
+      requirements: CardRequirements.builder((b) => b.tag(Tags.JOVIAN)),
+      renderData: CardRenderer.builder((b) => {
+        b.productionBox((pb) => {
+          pb.heat(3).br;
+          pb.energy(3);
+        });
+      }),
+      victoryPoints: 1,
+    };
 }
 
--- a/src/cards/Birds.ts
+++ b/src/cards/Birds.ts
@@ -8,6 +8,10 @@ import {ResourceType} from '../ResourceType';
 import {Resources} from '../Resources';
 import {CardName} from '../CardName';
 import {DecreaseAnyProduction} from '../deferredActions/DecreaseAnyProduction';
+import {CardMetadata} from './CardMetadata';
+import {CardRequirements} from './CardRequirements';
+import {CardRenderer} from './render/CardRenderer';
+import {CardRenderDynamicVictoryPoints} from './render/CardRenderDynamicVictoryPoints';
 
 export class Birds implements IActionCard, IProjectCard, IResourceCard {
     public cost = 10;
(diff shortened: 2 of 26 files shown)


src/cards/BeamFromAThoriumAsteroid.ts
```diff
@@ -22,5 +25,17 @@ export class BeamFromAThoriumAsteroid implements IProjectCard {
     public getVictoryPoints() {
       return 1;
     }
+    public metadata: CardMetadata = {
+      cardNumber: '058',
```
kberg, line 29: I've taken to having two copies of this PR open at once -- so I can look at what's deleted from HTML_data.ts while reading this. And what I want to know is -- where are you getting the '058' from?

kberg (commented):
I'll get back to this later, but I thought I'd send this one comment so far. Also, **thank you so so much for making this PR easy to read.**

src/cards/BeamFromAThoriumAsteroid.ts
chosta: https://ssimeonoff.github.io/cards-list

src/cards/render/CardRenderer.ts
```diff
@@ -120,7 +121,7 @@ class Builder {
     }
   }
 
-  protected _addTile(selector: string): void {
+  protected _addTile(selector: TileType): void {
```
chosta, line 124: @kberg as I mentioned in discord I want to use TileType enum instead of string selectors (or any other attributes) and add classes in the Vue component to handle the markup.

chosta:
@kberg I've added two changes:
+ rework on your solution for the ares tiles (as described in discord)
+ rework for description to be able to also accept parameters like `align`. This is needed for some exceptions like AICentral, wheer there is not enough space for everything because of the [points] element. We have some other cards that need these adjustments too, so a `string | Interface` as type for description seemed the correct approach at the time.

src/cards/Birds.ts
```diff
@@ -34,4 +38,19 @@ export class Birds implements IActionCard, IProjectCard, IResourceCard {
...
+      renderData: CardRenderer.builder((b) => {
+        b.effectBox((eb) => {
+          eb.empty().startAction.animals(1);
+          eb.description('Action: Add an animal to this card');
+        }).br;
+        b.productionBox((pb) => {
+          pb.minus().plants(-2).any;
+        });
+      }),
+      victoryPoints: CardRenderDynamicVictoryPoints.animals(1, 1),
```
kberg, line 54: 1,1  must mean one point per one animal (or the other way, but i get the idea. Nice.)

src/cards/BuildingIndustries.ts
```diff
@@ -20,4 +21,14 @@ export class BuildingIndustries implements IProjectCard {
       player.addProduction(Resources.STEEL, 2);
       return undefined;
     }
+    public metadata: CardMetadata = {
+      cardNumber: '065',
+      description: 'Decrease your Energy production 1 step and increase your steel production 2 steps',
+      renderData: CardRenderer.builder((b) => {
+        b.productionBox((pb) => {
+          pb.minus().energy(1).br;
```
kberg, line 29: Is there a value in the plus() or minus() at all? I mean, what if this line read `pb.energy(-1)`?

src/cards/ares/BiofertilizerFacility.ts
```diff
@@ -25,14 +25,14 @@ export class BiofertilizerFacility implements IProjectCard {
     public metadata: CardMetadata = {
       description: 'Requires 1 science tag. Increase your plant production 1 step. ' +
           'Add up to 2 microbes to any card. ' +
-          'Place this tile which grants an <b>adjacency bonus</b> of 1 plant and 1 microbe.',
+          'Place this tile which grants an ADJACENCY BONUS of 1 plant and 1 microbe.',
```
kberg, line 28: Well, that sure is smart.

src/styles/cards_v2.less
```diff
@@ -473,6 +482,17 @@
...
             .card-tile {
                 margin-left: 4px;
+                margin-right: 4px;
+            }
+            .card-special-tile--M {
+                background-size: 40px;
+                width: 40px;
+                height: 40px;
+            }
+            .card-biofertilizer_facility--M {
```
kberg, line 492: Can you tell me what the M suffix is for? Also, perhaps this belongs in  `ares.less`.

kberg (approved):
Nothing stopping this from going in. Nice work!

src/cards/BuildingIndustries.ts
chosta: The problem is that sometimes we have a + income but no + sign on the card. Same goes for -
We have to force it somehow so that is why I just add it manually. 
After I'm done with all the cards, I'll reiterate on things like that and look for patterns to abstract. Something like "all negative production should have minus symbol" etc.

src/cards/ares/BiofertilizerFacility.ts
chosta: It's just consistent with the text markup we have so far. I think we can introduce formatting in a future update.

src/styles/cards_v2.less
chosta: M = Medium (S, M, L, XL)
There are items on cards that come in different sizes. Ares tiles are probably all the same size, so you have a point here. I'll try to address this in my next PR which will handle the `.tile()` functionality.

PR #1950, opened 2020-11-24
WIP merge nwai90/ux_imporvements  about add community cards/corp/preludes

Ronnoc:
@nwai90 
still a largePR

merge some code from nwai90/ux_imporvements


kberg:
This is a pretty huge PR. @nwai90 any thoughts?

nwai90:
I would prefer this to be split into many smaller PRs if it is to be done at all. That will make reviews easier too as some core logic was extended/modified along the way. A PR of this size is very difficult to review accurately and there already seem to be several merge errors introduced e.g. `525cef13aee065e5675c5e5f6604d9eb9e941735` and `ff915a0cc0e62731fb629bb2ed038f7ea9c1786a`.

Ronnoc:
Ok I will try to split it

PR #1952, opened 2020-11-25
Atmoscoop: Heat/Venus increase before adding floaters

Lynesth:
Couldn't reproduce the issue mentioned on Discord but this will hopefully reduce user mistake to have Heat/Venus increase uncoupled from the floaters addition (and may help debugging said bugs).

Also made it so that when a OrOptions has a callback, it will be run after the selected option's callback.


Lynesth:
It actually fixes the issue encountered on discord.

nwai90 (approved):

PR #1954, opened 2020-11-25
Rewrite Imported Nitrogen

Lynesth:
Reduce code duplication and split the action in two. Add microbes to card, then add animals.
Here's the reason for that choice:

Because of how `AndOptions` is implemented, a failsafe (using microbesAdded and animalsAdded) had to be implemented to prevent adding multiple microbes if an error occured (if you didn't select an animal card for example).
But because of that failsafe, it also means that, in case of an error, you cannot change your mind and select another microbe card, as microbes were already added to the card you previously selected.


nwai90 (approved):

kberg:
I mean, simpler is good!

PR #1957, opened 2020-11-26
move src/cards to src/cards/base

bafolts:
Self-explanatory. Moves all base cards to the `src/cards/base` directory. No other changes included.


bafolts:
This is hard to review with the github UI. Below is diff file from CLI.

[easier.txt](https://github.com/bafolts/terraforming-mars/files/5599824/easier.txt)

kberg:
This is a great idea, but since Dani is in the middle of moving all those other files, how can these be done without merge conflict issues?

bafolts:
git mv was used to move the files so there should be no conflicts unless
the same parts of files were changed.

kberg:
> git mv was used to move the files so there should be no conflicts unless the same parts of files were changed.
Fantastic. Looks great. Thank you!

kberg (approved):

Issue #1127, opened 2020-08-11
Strange card selecting bug on initial hand


mloballer:
So I was playing a game, selected all my cards and hit save. There was an orange outline around them, so I know they were all selected. I got the error message not enough cards selected, so I thought I forgot to select a prelude or corp, but I saw that it was selected, so I decided to unselect and reselect my cards, and then I decided to change my mind and select capital ( didn't chose this one in the first try). After I selected, it somehow bought 5 cards, and doubled local shading and research outpost (the two cards I originally chose), not sure how.

![image](https://user-images.githubusercontent.com/10591526/89851654-d7e9c380-db41-11ea-86f3-7706c4fb968f.png)

chosta:
Thanks for the nice description. I'll try to recreate this.
We have to add a backend exception to not allow card duplication in the initial selection IMO anyway.

Lynesth:
Was able to reproduce. Will submit a fix.

PR #1959, opened 2020-11-26
Fix first cards pick when an error occur

Lynesth:
The issue it fixes has a couple of ways to be exploited. I'll give one example below.

**Steps to reproduce the issue:**
 - Create a 2 players game
 - Make sure one of the player has Phobolog
 - Select 8+ cards with Phobolog player and confirm the selection [_No error_]
 - Confirm the other player selection, you get an error saying you took too many cards [_But that's really for Phobolog_]
 - Change the other player selection (or not) and validate again

**Consequences:**
 - The other player has more/duplicated cards [_You get cards from the first and the second selections you made_]
 - Phobolog player has no cards [_Yay_]
 - If the other player is the first in turn order, its corporation is played twice
 - All cards not selected during other player both elections were added to the discard pile, so there's most likely duplicates there as well

**Solution:**
 - Move initial cards buying checks into the `AndOptions` callback to be returned directly to the correct player
 - Discard cards after the checks passed instead of when the selection is made

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -312,7 +312,7 @@ export class Game implements ISerializable<SerializedGame, Game> {
           }
         } else {
           this.setStartingProductions(player);
-          this.playerHasPickedCorporationCard(player, new BeginnerCorporation() );
+          this.playerHasPickedCorporationCard(player, new BeginnerCorporation());
         }
       }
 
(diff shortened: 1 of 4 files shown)


Lynesth:
Fixes #1127

src/Game.ts
```diff
@@ -743,6 +733,18 @@ export class Game implements ISerializable<SerializedGame, Game> {
...
+        // Check for negative M€
+        const cardCost = corporation.cardCost !== undefined ? corporation.cardCost : player.cardCost;
+        if (corporation.name !== CardName.BEGINNER_CORPORATION && player.cardsInHand.length * cardCost > corporation.startingMegaCredits) {
+          player.cardsInHand = [];
+          player.preludeCardsInHand = [];
+          throw new Error('Too many cards selected');
+        }
+        // discard all unpurchased cards
+        player.dealtProjectCards
+          .filter((card) => player.cardsInHand.includes(card))
```
nwai90, line 745: 😛

Lynesth: Yeah yeah, fixing 😝

alrusdi (approved):

PR #1961, opened 2020-11-26
merge pallas colony in nwai90/ux_improvements

Ronnoc:


src/components/Colony.ts
```diff
@@ -76,25 +76,25 @@ export const Colony = Vue.component('colony', {
       if (colony.name === ColonyName.PLUTO || colony.name === ColonyName.MIRANDA) return -12;
       return 0;
     },
-    getIapetus:(): string => {
+    getIapetus: (): string => {
```
alrusdi, line 79: Strange syntax. Why do we need ":" on the left side of () ?

Ronnoc: I'm not fammilar with vue at all.
code in main is same like this? 
https://github.com/bafolts/terraforming-mars/blob/main/src/components/Colony.ts

Ronnoc:
@nwai90

nwai90:
This looks ok, I think. Can't really tell if anything is missing but someone else should be reviewing it, otherwise it would just be me checking my own old code 😛

src/components/Colony.ts
Lynesth: It's actually
```ts
getIapetus: () => { ... }
```
But since we're also typing the return value, that's why we have another colon on the right side of the parenthesis for `: string`.

I think we could write it this way (but it should really be the same in the end):
```ts
getIapetus(): string { ... }
```

src/colonies/Colony.ts
```diff
@@ -239,6 +240,34 @@ export abstract class Colony implements SerializedColony {
...
+        }
+        break;
+
+      case ColonyBenefit.GIVE_MC_PER_DELEGATE:
+        if (game.turmoil) {
+          let partyDelegateCount = PLAYER_DELEGATES_COUNT - game.turmoil.getDelegates(player.id);
+          if (game.turmoil.lobby.has(player.id)) partyDelegateCount--;
+          if (game.turmoil.chairman === player.id) partyDelegateCount--;
+
+          player.megaCredits += partyDelegateCount;
```
Lynesth, line 266: Should probably log it and use `setResource` method (that said, @nwai90, when do you change the name ? :p).
```ts
player.setResource(Resources.MEGACREDITS, partyDelegateCount);
LogHelper.logGainStandardResource(game, player, Resources.MEGACREDITS, partyDelegateCount);
```

Lynesth (commented):
Quickly checked front end code, which seems ok. Left a few comments regarding server side code.

src/colonies/Colony.ts
Ronnoc: If logGain will future merge into setResource func?

Ronnoc:
> Quickly checked front end code, which seems ok. Left a few comments regarding server side code.

fixed

Ronnoc:
What else do i need to do

kberg:
> What else do i need to do

My mistake, sorry. I'll remove that tag.

Lynesth:
@Ronnoc 
I took the liberty to modify the colonies filter component (committed directly to your branch).
Please merge `main` in, resolve any conflicts and then it should be good to go !

Ronnoc:
I merged `main` . And it's fast forward .

Lynesth (approved):
This looks fine to me.

PR #1964, opened 2020-11-27
Refactor Player's runInput method

Lynesth:
Removed unnecessary variables.
Removed duplicated code by adding a `checkInputLength` method.
Removed unnecessary checks from `SelectHowToPayForCard` since those are performed in the callback.
Text for errors that should not happen (like incorrect input length) was kept "simple".
Didn't touch Ares inputs @kberg, but you can probably just add `this.checkInputLength(input, 1, 1);` for both of them.

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -699,7 +699,7 @@ export class Player implements ISerializable<SerializedPlayer, Player> {
       }
     }
 
-    private checkInputLength(input: Array<Array<string>>, length: number, firstOptionLength?: number) {
+    private checkInputLength(input: ReadonlyArray<ReadonlyArray<string>>, length: number, firstOptionLength?: number) {
       if (input.length !== length) {
         throw new Error('Incorrect options provided');
       }
@@ -708,7 +708,28 @@ export class Player implements ISerializable<SerializedPlayer, Player> {
       }
     }
 
-    private runInput(game: Game, input: Array<Array<string>>, pi: PlayerInput): void {
+    private parseHowToPayJSON(json: string): HowToPay {
+      const defaults: HowToPay = {
+        steel: 0,
+        heat: 0,
+        titanium: 0,
+        megaCredits: 0,
+        microbes: 0,
+        floaters: 0,
+        isResearchPhase: false,
+      };
+      try {
+        const howToPay: HowToPay = JSON.parse(json);
+        if (Object.keys(howToPay).every((key) => key in defaults) === false) {
+          throw new Error('Input contains unauthorized keys');
+        }
+        return howToPay;
+      } catch (err) {
+        throw new Error('Unable to parse HowToPay input ' + err);
+      }
+    }
+
+    private runInput(game: Game, input: ReadonlyArray<ReadonlyArray<string>>, pi: PlayerInput): void {
       if (pi instanceof AndOptions) {
         this.checkInputLength(input, pi.options.length);
         for (let i = 0; i < input.length; i++) {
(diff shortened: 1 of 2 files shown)


src/Player.ts
```diff
@@ -686,46 +686,42 @@ export class Player implements ISerializable<SerializedPlayer, Player> {
     }
 
     public getCard(cards: Array<IProjectCard>, cardName: string): IProjectCard {
-      const foundCards = cards.filter((card) => card.name === cardName);
```
bafolts, line 689: This is a good change. As the terraforming-mars.herokuapp.com server is getting more traffic every little bit will help.

src/Player.ts
```diff
@@ -686,46 +686,42 @@ export class Player implements ISerializable<SerializedPlayer, Player> {
...
-    private runInput(
-      game: Game,
-      input: Array<Array<string>>,
-      pi: PlayerInput): void {
+
+    private checkInputLength(input: Array<Array<string>>, length: number, firstOptionLength?: number) {
+      if (input.length !== length) {
+        throw new Error('Incorrect options provided');
+      }
+      if (firstOptionLength !== undefined && input[0].length !== firstOptionLength) {
```
bafolts, line 706: This needs to check that `input.length > 0`. If `input.length` were `0` this would throw a runtime error that `length` is not a property of `undefined`.

src/Player.ts
```diff
@@ -735,24 +731,23 @@ export class Player implements ISerializable<SerializedPlayer, Player> {
...
-        const remainingInput = [];
-        for (let i = 1; i < input.length; i++) {
-          remainingInput.push(input[i].slice());
+        // input length is variable, can't test it with checkInputLength
+        if (input.length === 0 || input[0].length !== 1) {
+          throw new Error('Incorrect options provided');
         }
-        this.runInput(game, remainingInput, waiting.options[optionIndex]);
+        const optionIndex = parseInt(input[0][0]);
+        const selectedOptionInput = input.slice(1);
```
bafolts, line 746: This _should_ be fine to only call `slice` on the top level array. The name `selectedOptionInput` is clearer than `remainingInput`. Prior to this change we would have had a copy of the input arrays as well. I don't think we ever mutate the input so this shouldn't matter. While here lets change the type of `input` to `ReadyonlyArray<ReadyonlyArray<string>>`. This will ensure that nothing is modifying these arrays.

src/Player.ts
```diff
@@ -889,39 +827,12 @@ export class Player implements ISerializable<SerializedPlayer, Player> {
           floaters: 0,
           isResearchPhase: false,
         };
-        if (!this.canUseHeatAsMegaCredits) payMethod.heat = 0;
```
bafolts, line 892: Lets make a private function for `parseHowToPay(string: json): HowToPay` that can be used in both places this is used. This should further remove code duplication as the `const payMethod: HowToPay = {` wont be declared in both places.

There shouldn't be any need to use `Object.assign`. Should be able to use:

```typescript
private parseHowToPay(string: json): HowToPay {
  try {
    const howToPay: HowToPay = JSON.parse(string);
    // do run-time type checking to verify
    return howToPay;
  } catch (err) {
    // unable to parse input, log warning to be helpful with debugging
  }
}
```

Typescript will take our word for it so that runtime type checking you have verifying the properties makes sense. This is removing the checks that heat can be used as well as steel and titanium but I do think we should be verifying these each place they are used so this seems fine to me to remove. Putting in one function will be helpful.

bafolts (changes requested):

src/Player.ts
Lynesth: It is expected that the `length` argument to that function is greater than or equal to 1, so with the check 3 lines above I it should be fine, wouldn't it ?

src/Player.ts
Lynesth: Done it.
Had to keep the exhaustive "default" HowToPay variable to check against since interfaces are not available at runtime and simply using an array of keys would prevent any futur changes to the HowToPay interface to be caught here.

Lynesth: Also if we want to be thorough, we could change the properties check to:
```ts
        if (Object.keys(howToPay).every((key) =>
          key in defaults &&
          typeof (howToPay as any)[key] === typeof (defaults as any)[key],
        ) === false) {
```
Let me know how you feel about it.

bafolts: As long was we are checking that a player isn't cheating that should be of more value than this runtime type checking. When we send this data from client we are using `HowToPay` so assuming the user isn't malicious we are only verifying here the `JSON.stringify` worked on the client. The check in place right now should be a better than nothing sanity check that the client and server haven't gotten out of sync.

bafolts (approved):

PR #1966, opened 2020-11-27
changed 'colony' to 'colony tile', where appropriate

### LINKED ISSUE #1870: minor: "trade with a colony tile"
I only just noticed this! Should say "trade with a colony tile" and "select a colony tile for trade".

No colonies exist yet in this game.

image

Starcalc:
Fixes  #1870, but also for other places where it was called. Fixed for the translations so that the translations still work. (Did not change the translations itself, except for german).


Lynesth:
LGTM.

nwai90 (approved):

PR #1968, opened 2020-11-28
Add Amazonis board + red highlight for volcanic areas from nwai90/ux_improvements

Ronnoc:
@nwai90


Lynesth:
Please split this PR in 2. One for the Amazonis board and one for the red highlights.

Also, there's a couple of issues with the Amazonis board, like Land Claim or the solo setup being able to select the center space but I'll make a proper review later.

Issue #1969, opened 2020-11-28
Can Robotic Workforce work on Ares' Capital card?


kberg:
A bug report on Discord suggested it could not be done. https://discord.com/channels/737945098695999559/742721510376210583/782125546213081108

andylatto:
Very simple fix to that one; the builderCardNames list in RoboticWorkforce.ts includes CAPITAL, but doesn't include CAPITAL_ARES. There may be other Ares cards missing from this list; I'll check.

andylatto:
Yep. BIOFERTILIZER_FACILITY, OCEAN_CITY, OCEAN_FARM, COMMERCIAL_DISTRICT_ARES, SOLAR_FARM, MINING_AREA_ARES, MINING_RIGHTS_ARES, MOHOLE_AREA_ARES, and NATURAL_PRESERVE_ARES. Several of these also need to get added to lists in the canPlay code to check that the player has enough energy production to use, the mining cards need to have the same special-casing as their non-ares counterparts.SOLAR_FARM is a little trickier, since the production this card produces is variable.

kberg:
Oh, man, Solar Farm. >:-(

Lynesth:
Fixed by #1974.

PR #1970, opened 2020-11-28
Added two test cases for RoboticWorkforce and Ares

Starcalc:
This is to test issue #1969. I've added two test cases, but both of the fail on ares cards. Maybe I just wrote the tests wrong or this indicates a deeper problem. This is work-in-progress, so do-not-merge as it would break the tests as of now.


Starcalc:
#1974 incorporates this

PR #1971, opened 2020-11-28
add hex volcanic pic

Ronnoc:


Ronnoc:
@Lynesth

src/SpaceBonus.ts
```diff
@@ -11,5 +11,8 @@ export enum SpaceBonus {
     MEGACREDITS,
     ANIMAL,
     MICROBE,
-    POWER
+    POWER,
+
+    // Custom maps / special tiles
+    VOLCANIC,
```
kberg, line 17: Is there a better way to do this? This isn't a space bonus.

kberg: For instance, consider adding code directly to Board.ts to use this, or provide a visualization overlay that _uses_ similar code to BoardSpace.

kberg:
Image copied here for reference:

image

kberg:
General visualization comments:
1. The yellow border conflicts with Ares' border, which implies that a tile is already there, and that it provides a bonus.
2. What are peoples' thoughts of what this looks like on a board?

src/SpaceBonus.ts
kberg: > General visualization comments:
> 
> 1. The yellow border conflicts with Ares' border, which implies that a tile is already there, and that it provides a bonus.
> 2. What are peoples' thoughts of what this looks like on a board?

Ah, also I just noticed in another PR that there's some use of a red border, too? Red borders are also used in Ares, to indicate adjacency costs. That conflict will also need to be addressed, if not differentiated.

Ronnoc: volcanic is a space info, maybe another tile image (which not similar with ares ux) is better
emm, I think the motivation is to pay less effort when creating new map with volcanic places

VOLCANIC is not a tile too, it is a subtype of land

kberg: I had hoped someone else would make opinions about the image itself, but they haven't. So, I'm not sure I think it's a good addition. It's visually striking, and so if it's going to be a visual attention getter, I ask what its value might be. Given that volcanoes are limited, in the base game, to just a couple of cards being impacted by them, it imposes a lot of visual demand for little return.

alrusdi: It is not obvious from image if it is volcanic. If it is covered by other tiles it will not help much to get if you will be able to play card with volcanic area placement requirement.

Lynesth: I think that it shouldn't use a SpaceBonus either but should be defined based on the names of the Spaces, much like it already is for the original board.
Also regarding the image, I understand that the point of the highlight is to show where those volcanic spaces are on the new Amazonis board, since it has its own, but I don't like it either. If it ever was to happen, it should be something light, like the oceans spaces.

kberg: Are the text and pointers to those spaces insufficient? I guess they are because I typically just rely on the app highlighting which spaces are valid.

Ronnoc: I thought just because draw a line is much more difficulty than set a value

kberg: So it doesn't sound like this has the support to go in. Sorry about that.

PR #1973, opened 2020-11-28
add LeadershipSummit from nwai90/ux_improvements

Ronnoc:


kberg (changes requested):
Can this become part of the CardManifest instead? In the same way there are two sources of Preludes, this would fit naturally into there.

kberg:
@Ronnoc do you want some help with this?

Ronnoc:
> @Ronnoc do you want some help with this?

Sorry, but I didn't find globalEvents manifest

kberg:
> > @Ronnoc do you want some help with this?
> 
> Sorry, but I didn't find globalEvents manifest

There is no GlobalEventsManifest. I am saying that we would add a new card type to the CardManiffest represnting global events.

Also, it seems every time you force push our prior discussions disappear. Can you find a way to not force push?

Ronnoc:
> > > @Ronnoc do you want some help with this?
> > 
> > 
> > Sorry, but I didn't find globalEvents manifest
> 
> There is no GlobalEventsManifest. I am saying that we would add a new card type to the CardManiffest represnting global events.
> 
> Also, it seems every time you force push our prior discussions disappear. Can you find a way to not force push?

Sorry, I will not force push.

kberg (approved):
We can always revamp another time.

PR #1974, opened 2020-11-28
Fix Robotic Workforce to work with Ares cards

Lynesth:
Please check the PR commit by commit.
The first one contains the actual changes.
The second one is simply formatting and reordering things around.
Tests inspired by @Starcalc.

### DIFF
--- a/src/cards/base/RoboticWorkforce.ts
+++ b/src/cards/base/RoboticWorkforce.ts
@@ -11,368 +11,368 @@ import {DecreaseAnyProduction} from '../../deferredActions/DecreaseAnyProduction
 import {SpaceBonus} from '../../SpaceBonus';
 
 export class RoboticWorkforce implements IProjectCard {
-    public cost = 9;
-    public tags = [Tags.SCIENCE];
-    public name = CardName.ROBOTIC_WORKFORCE;
-    public cardType = CardType.AUTOMATED;
-    public hasRequirements = false;
-    public canPlay(player: Player, game: Game): boolean {
-      return this.getAvailableCards(player, game).length > 0;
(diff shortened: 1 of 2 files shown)


src/cards/base/RoboticWorkforce.ts
```diff
@@ -196,27 +211,40 @@ export class RoboticWorkforce implements IProjectCard {
...
           if (bonusResource !== undefined && bonusResource === Resources.STEEL) {
             this.miningSteelProduction++;
           }
           if (bonusResource !== undefined && bonusResource === Resources.TITANIUM) {
             this.miningTitaniumProduction++;
           }
+          break;
+
+        case CardName.SOLAR_FARM:
+          const solarFarmSpace = game.board.getSpaceByTileCard(CardName.SOLAR_FARM);
```
kberg: This is a clever idea. I would like to add this to the Ares wiki page for how this game deals with outliers. Am I reading it right that what you're doing here is adding the number of energy based on the number of plants the tile was placed on?

kberg: Same comment applies to Mining Area and Mining Rights.

kberg (approved):
Thanks for splitting out the reformat.

src/cards/base/RoboticWorkforce.ts
Lynesth: Yes that's what it does for Solar Farm.

For Mining Area and Rights something was already in for those two, that saves the production you get directly on the card (in the `bonusResource` member). See the `play` method of `src/cards/base/MiningCard.ts` for details.

PR #1977, opened 2020-11-28
Add Test players with well-known IDs.  Apply in two places.

kberg:
This makes the id unmodifiable post construction.

I'd like to know if people think this is a bad idea, or that it might disrupt deserialization, for instance.

I tested this by changing `Capital` to force an ungainly failure. This is what it output

```
  1) Capital
       Should play:
     Error: the string "Cant find player de450352d26b" was thrown, throw an Error :)
  2) CapitalAres
       Should play:
     Error: the string "Cant find player blue-id" was thrown, throw an Error :)
  3) OceanCity
       Ocean City counts as VP for Capital:
     Error: the string "Cant find player 1e89d3cb8660" was thrown, throw an Error :)
```

That's because `CapitalAres` has the new player with a custom ID.


src/Player.ts
```diff
@@ -148,8 +148,9 @@ export class Player implements ISerializable<SerializedPlayer, Player> {
         public name: string,
         public color: Color,
         public beginner: boolean,
-        public handicap: number = 0) {
-      this.id = this.generateId();
+        public handicap: number = 0,
+        id: PlayerId | undefined = undefined) {
+      this.id = id === undefined ? this.generateId() : id;
```
bafolts, line 153: This change makes sense to me and should be fine even with `ISerializable`. The `id` of player gets set through `Object.assign`. I wish we didn't use `Object.assign` as it makes it difficult to know which properties are being set.

https://github.com/bafolts/terraforming-mars/blob/main/src/Player.ts#L2308

It could be provided though which would avoid having to generate the id unnecessarily.

https://github.com/bafolts/terraforming-mars/blob/main/src/Game.ts#L1717

Would also make sense to me to always have `id` provided to `Player` and move `generateId` to `Game`. The player doesn't care what it's id as whereas `Game` needs them to be unique and difficult to reproduce for security.

kberg: > I wish we didn't use `Object.assign` as it makes it difficult to know which properties are being set.

Added #1979

kberg: > Would also make sense to me to always have id provided to Player and move generateId to Game. The player doesn't care what it's id is whereas Game needs them to be unique and difficult to reproduce for security.

I started peeking at that, and realized it would make this PR explode. So, making this a TODO.

kberg:
PTAL, if this goes in I can start moving callers over.

bafolts (approved):
This is good change. If the player api were to change in some way we would only need to modify the test players versus the player instances creating within each spec file.

kberg:
> If the player api were to change in some way we would only need to modify the test players versus the player instances creating within each spec file.

I can't parse this sentence.

PR #1978, opened 2020-11-28
Remove IParty.agenda stuff, rename Chairman to Dynamic (though I don't love that, either.)

kberg:
This looks as though it the PA branch needs a pull from main first.


nwai90:
@kberg Should we close this one as you are going to be starting from a fresh slate?

kberg:
Yes. I'll pick this up tomorrow. SLEEEP.

PR #1980, opened 2020-11-29
Move webpack output to the build folder

kberg:
This PR has two commits: the first is the functional one, the second one reformats.


bafolts (changes requested):
This needs to update where the file is read in `server.ts`.

https://github.com/bafolts/terraforming-mars/blob/main/server.ts#L1070

This line should change to:

```diff
index 6547460f..206c4756 100644
--- a/server.ts
+++ b/server.ts
@@ -1067,7 +1067,7 @@ function serveAsset(req: http.IncomingMessage, res: http.ServerResponse): void {
       res.setHeader('Content-Encoding', 'gzip');
       suffix = '.gz';
     }
-    file = `dist${req.url}${suffix}`;
+    file = `build${req.url}${suffix}`;
```

kberg:
> This needs to update where the file is read in `server.ts`.
> 
> https://github.com/bafolts/terraforming-mars/blob/main/server.ts#L1070
> 
> This line should change to:

Thanks. I tested this manually once and it worked, so what did I do wrong the first time? (Did I forget to clean the dist/ directory? :D)

kberg:
> > This needs to update where the file is read in `server.ts`.
> > https://github.com/bafolts/terraforming-mars/blob/main/server.ts#L1070
> > This line should change to:
> 
> Thanks. I tested this manually once and it worked, so what did I do wrong the first time? (Did I forget to clean the dist/ directory? :D)

Yep, turns out that was it. Thanks. This should do it.

bafolts (approved):

PR #1981, opened 2020-11-30
dynamic translations for numbers

bafolts:
Start to allow for translations that involve a number. Can expand upon this logic as needed. For now one number in the string is supported.

### DIFF
--- a/src/directives/i18n.ts
+++ b/src/directives/i18n.ts
@@ -4,6 +4,21 @@ import * as raw_translations from '../../assets/translations.json';
 
 const TM_translations: {[x: string]: {[x: string]: string}} = raw_translations;
 
+function translateThroughRegExp(englishText: string, language: string): string | undefined {
+  for (const translation of Object.keys(TM_translations)) {
+    if (translation.indexOf('{{d}}') !== -1) {
+      const re = new RegExp(translation.replace('{{d}}', '(\\d+)'));
+      if (re.test(englishText)) {
+        const languages = TM_translations[translation];
+        if (languages !== undefined && languages[language] !== undefined) {
+          return englishText.replace(re, languages[language]);
+        }
+      }
+    }
+  }
+  return undefined;
+}
+
 export function translateText(englishText: string): string {
   let translatedText = englishText;
   const lang = PreferencesManager.loadValue('lang') || 'en';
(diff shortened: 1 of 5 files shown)


src/directives/i18n.ts
```diff
@@ -4,27 +4,48 @@ import * as raw_translations from '../../assets/translations.json';
...
-      if (stripedText !== englishText) {
-        translatedText = '(' + stripedText + ')';
-      }
-    } else if (stripedText && stripedText.length > 3) {
-      console.log('Please translate "' + stripedText + '"');
+    return languages[lang];
+  }
+
+  // match through RegExp
+  const regMatch = translateThroughRegExp(englishText, lang);
```
kberg, line 36: It's not _exactly_ a regex since `{{d}}` isn't more than really a regex. At first I had hoped this meant other regex replacement were possible, though I imagine that's a terrible way to internationalize!

This is minor, of course, but consider using `{{d}}` in the replacement string, and not using the term regex in the method description. But if you did neither of these things I wouldn't bother.

kberg (approved):

src/directives/i18n.ts
alrusdi: How to translate a string "'Select a card to keep and pass the rest to **Bafolts**" this way? Bafolts in this string is dynamig value. I've suggested to write in code something like this:
"'Select a card to keep and pass the rest to {Bafolts}"
Translation string for russian lang will be:
"'Select a card to keep and pass the rest to {}"
And replacement code like this:

```
const translatableString = "Select a card to keep and pass the rest to {Bafolts}";
let translation = "Выберите карту и передайте ее игроку: {1}";
let matchesRegEx = new RegExp(/{[^}]+}/g);
let allMatches = translatableString.match(matchesRegEx) || [];

for (let key in allMatches) {
    let idx = parseInt(key); // we want 1-based index 
    let match = allMatches[key].slice(1, -1); // strip { and }
    translation = translation.replace("{" + (idx+1) + "}", match)
}
return translation;
```

bafolts: For strings I was envisioning we would use `{{s}}` similar to printf. The dynamic string would then become `Select a card to keep and pass the rest to {{s}}`

bafolts: While this will work fine for numbers which are easy to match, strings are going to pose a challenge as we don't know when to end the match.

```
"Select Brian Folts to keep".replace(/Select (.+) Folts to keep/, "Select not to $1")
"Select not to Brian"
```

I will close this PR while I re-consider an approach that will work for both numbers and strings.

bafolts:
Closing to get out of the open list with plans to re-open when ready.

bafolts:
@alrusdi I have work in place to pass out a `DynamicTranslatableStringModel`. Will finish up mapping the values tomorrow. Will work like the log messages after this.

PR #1982, opened 2020-11-30
Diaspora movement now counts own wild tags

Lynesth:


kberg:
Perhaps I'm ignorant, why would the player's own wild tags count but not others?

Lynesth:
Wild tags can be anything only for the owner of the tag, when he is the one doing an action.
> when **you** perform an action, the wild tag counts as any tag of your choice.

kberg:
Thanks.

>

kberg (approved):

PR #1983, opened 2020-11-30
All test Player instances go through TestPlayers.

kberg:
Today I learned that Organize Imports reorders imports, and not just removes the unused imports. But by and large, this is a lot of files. LMK how you want to review this.

### DIFF
--- a/tests/Player.spec.ts
+++ b/tests/Player.spec.ts
@@ -10,14 +10,15 @@ import {PowerSupplyConsortium} from '../src/cards/base/PowerSupplyConsortium';
 import {SaturnSystems} from '../src/cards/corporation/SaturnSystems';
 import {SelectOption} from '../src/inputs/SelectOption';
 import {Resources} from '../src/Resources';
+import {TestPlayers} from './TestingUtils';
 
 describe('Player', function() {
   it('should initialize with right defaults', function() {
-    const player = new Player('test', Color.BLUE, false);
+    const player = TestPlayers.BLUE.newPlayer();
     expect(player.corporationCard).is.undefined;
   });
   it('Should throw error if nothing to process', function() {
-    const player = new Player('test', Color.BLUE, false);
+    const player = TestPlayers.BLUE.newPlayer();
     const game = new Game('foobar', [player], player);
     (player as any).setWaitingFor(undefined, undefined);
     expect(function() {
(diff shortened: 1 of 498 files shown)


tests/Player.spec.ts
```diff
@@ -10,14 +10,15 @@ import {PowerSupplyConsortium} from '../src/cards/base/PowerSupplyConsortium';
 import {SaturnSystems} from '../src/cards/corporation/SaturnSystems';
 import {SelectOption} from '../src/inputs/SelectOption';
 import {Resources} from '../src/Resources';
+import {TestPlayers} from './TestingUtils';
 
 describe('Player', function() {
   it('should initialize with right defaults', function() {
-    const player = new Player('test', Color.BLUE, false);
+    const player = TestPlayers.BLUE.newPlayer();
```
Lynesth, line 17: Can at least one test in there still use an actual player so that it still tests generateId() ?

Lynesth (commented):
It's a huge PR, not possible to review all files. Not sure how to review it. Split it by folder ?
A couple of comments:
- You missed a few `new Player( ... )` and I don't think that's intended
- To be honest, I don't like that the imports are sorted by path but it seems most files were already sorted that way

tests/Player.spec.ts
kberg: The whole task took me about an hour, I'm happy to redo all of it so it doesn't sort the imports.

I did miss a few "new Player" calls, but that's because (as you may see) I used a broad brush.

kberg: (There has GOT to be an easier way to do this.)

bafolts:
I can review. There is no test for `generateId` today. That function is calling to generate random ids. This should eventually move so that `Game` provides the id for each `Player`. Once game is providing the id then game can ensure that the player ids are unique as is needed by `Game`. As written today we don't necessarily get a unique id but rely on probability to hopefully generate a random id we haven't used yet so I see no harm in not generating this id time being.

bafolts:
The github ui is very slow, looking at this through the CLI diff is simpler:

```
git diff e0b877a648c00e142d77f2cbf95bfb305d3bcefd~ e0b877a648c00e142d77f2cbf95bfb305d3bcefd
```

bafolts (approved):
Going through the diff file for this commit it does what it claims and updates player instances to go through TestPlayers. Some may have been missed but updating those in separate PR should lead to a smaller PR at that time.

Lynesth:
@bafolts 

What I meant was to have at least one test that actually creates a Player the same way the app does.

Currently the only difference would be that the constructor would call `generateId()` (hence my example) but the point was that any futur changes to the Player class/constructor could not in someway be shadowed in the tests by the specific way they are created in TestPlayers.

Either that, or, as you mentioned, change the way players are created in the app by making the `id` argument mandatory and moving `generateId()` outside the constructor.

kberg:
@Lynesth moving generateId to the constructor is my next step.
@bafolts thanks for ingesting that whole thing at once. You're like one of those little snakes eating mouse-sized PRs.

PR #1984, opened 2020-11-30
Cards Vue M

sienmich:
My first commit here. I'll be grateful If you could tell me if I'm doing the right thing. Thanks!

### DIFF
--- a/src/components/card/CardRenderItemComponent.ts
+++ b/src/components/card/CardRenderItemComponent.ts
@@ -115,6 +115,8 @@ export const CardRenderItemComponent = Vue.component('CardRenderItemComponent', 
           classes.push('card-tag-event');
         } else if (type === CardRenderItemType.SPACE) {
           classes.push('card-tag-space');
+        } else if (type === CardRenderItemType.SCIENCE) {
+          classes.push('card-tag-science');
         } else if (type === CardRenderItemType.JOVIAN) {
           classes.push('card-tag-jovian');
         } else if (type === CardRenderItemType.VENUS) {
--- a/src/components/card/CardRenderItemComponent.ts
+++ b/src/components/card/CardRenderItemComponent.ts
@@ -33,6 +33,12 @@ export const CardRenderItemComponent = Vue.component('CardRenderItemComponent', 
         if (this.item.size !== undefined) {
           classes.push(`card-ocean--${this.item.size}`);
         }
+      } else if (type === CardRenderItemType.GREENERY) {
+        classes.push('card-global-requirement');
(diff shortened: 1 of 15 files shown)


chosta (approved):
Great work!
I think putting description on the bottom of the metadata interface makes more sense. I'll start doing it myself.
Feel free to work on the cards. You can skip the ones you are unsure about and tag me.

src/styles/cards_v2.less
```diff
@@ -337,8 +337,6 @@
             }
             .card-tag-science {
                 background-image: url("./assets/tags/science.png");
-                background-size: 46px;
```
kberg, line 340: Just making sure this is intentional.

kberg:
> Great work!
> I think putting description at the bottom of the metadata interface makes more sense. I'll start doing it myself.
> Feel free to work on the cards. You can skip the ones you are unsure about and tag me.

I agree. This is terrific. Thanks for jumping in!

src/styles/cards_v2.less
chosta: Yeah, a bit funny but correct I think. I intend to add some variables and mixins to the less file once we have all card items described. I wish we had js-in-css :)

sienmich: All other card tags don't have `background-size` nor `background-position`, so I think the science tag also shouldn't. After this change the Mars Uni looks good.

src/components/card/CardRenderItemComponent.ts
```diff
@@ -33,6 +33,12 @@ export const CardRenderItemComponent = Vue.component('CardRenderItemComponent',
         if (this.item.size !== undefined) {
           classes.push(`card-ocean--${this.item.size}`);
         }
+      } else if (type === CardRenderItemType.GREENERY) {
+        classes.push('card-global-requirement');
+        classes.push('card-greenery-global-requirement');
+        if (this.item.size !== undefined) {
+          classes.push(`card-greenery--${this.item.size}`);
+        }
```
sienmich, line 41: I'm not sure about this one though. I copied it from above (oceans). But the greeneries are not really a `card-global-requirement`. Any suggestions?

chosta: They can be the same as `CardRenderer.city()`. It's not super important as long as the css is correct but for the sake of consistancy, they can be a `card-tile` with `card-greenery--${size}`

chosta:
@sienmich Let's remove the "Draft" and merge this. I need the greenery icon elsewhere :)

sienmich:
@chosta Ok, go ahead :)

PR #1988, opened 2020-12-01
Last set of TestPlayer migrations.

kberg:
Please feel free to opine about the use of TestPlayers in Players.spec.ts


tests/Game.spec.ts
```diff
@@ -393,11 +391,11 @@ describe('Game', function() {
   });
 
   it('Removes Hellas bonus ocean space if player cannot pay', function() {
+    const player = TestPlayers.BLUE.newPlayer();
```
kberg, line 394: I moved this up above the comment as the comment really relates to `secondPlayer`. This repeats for the test method below.

tests/Player.spec.ts
```diff
@@ -1,7 +1,5 @@
 
 import {expect} from 'chai';
-import {Color} from '../src/Color';
-import {Player} from '../src/Player';
```
kberg, line 4: I'll say this: there's nothing at all about the player constructor that is interesting, other than generating the player ID, and at this point I wouldn't bother writing a test for it just so Player was imported. Thoughts? Do you want one of those sanity tests where you call the constructor just to verify its values?

kberg:
Whoops! Will fix.

nwai90 (approved):

tests/Player.spec.ts
Lynesth: I think one test in "real" conditions would be nice yes. Can be done after you moved the id generation outside of the constructor though.

kberg: OK. Fascinating that the change I made to support this is actually breaking tests in surprising places. Thankfully they're database tests. (THANK YOU @bafolts)

PR #1989, opened 2020-12-01
Top bar

alrusdi:
Optional sticky top bar. I don't like visuals, but it is probably ok as a start...
sticky

Demo: http://37.230.117.156:9000/


kberg:
I'll look tomorrow evening. It's a great start.

src/components/Preferences.ts
```diff
@@ -247,6 +248,12 @@ export const Preferences = Vue.component('preferences', {
                             <i class="form-icon"></i> <span v-i18n>Hide corporation names for players</span>
                         </label>
                     </div>
+                    <div class="preferences_panel_item">
```
kberg, line 251: Instead, I'd make it dismissable and redisplayable right on the bar. You've seen this type of thing, of course. And that toggle can be stored as the preference.

kberg: ... or both, now that I think about it.

kberg (commented):
Basically LG, but someone else should opine about CSS and such.

chosta (commented):
Thanks for making that optional!
I have no issues with vue and logic. Lets see how the visuals will turn out to be and let me know if you need help.

alrusdi:
> Thanks for making that optional!
> I have no issues with vue and logic. Lets see how the visuals will turn out to be and let me know if you need help.

If we will make decision to make top bar removable in place (as @kberg suggested above) then i need some visual solution for it. Just "close button" on the right side of the bar looks odd

chosta:
How about an indicator that slides the bar left / right? Example - Evernote web interface:

sidebar-1

src/components/Preferences.ts
alrusdi: > ... or both, now that I think about it.

I've implemented both

src/components/TopBar.ts
```diff
@@ -0,0 +1,45 @@
...
+  },
+  data: function() {
+    return {};
+  },
+  methods: {
+    toggleBar() {
+      const newVal = this.isExpanded() ? '1': '';
+      PreferencesManager.saveValue('hide_top_bar', newVal);
+      PreferencesManager.preferencesValues.set('hide_top_bar', this.isExpanded());
+      this.$forceUpdate();
```
bafolts, line 23: You can avoid using `$forceUpdate` if a data property is created for `hideTopBar`. When that property were to change the template would update without needing to call to force the update. The `hideTopBar` would default to `PreferencesManager.loadValue('hide_top_bar')`.

bafolts (changes requested):
I would avoid using `$forceUpdate` and use data properties to let vue handle updating the component naturally without forcing it.

src/components/TopBar.ts
alrusdi: Your review lead me to read this article https://michaelnthiessen.com/force-re-render/
Author's point seems to be reasonable, so I've used his technique to force the component update.

src/components/TopBar.ts
```diff
@@ -0,0 +1,50 @@
...
+    PlayerInfo,
+  },
+  data: function() {
+    return {
+      componentKey: 0,
+    };
+  },
+  methods: {
+    forceRerender() {
+      this.componentKey += 1;
```
bafolts, line 22: This is at least not using $forceUpdate but still seems to be working against vue. With vue we shouldn't have to be concerned with when the renderer is going to update and it should more or less happen naturally as the data is updated.  If this used:

```
data: function() {
  return {
    hideTopBar: PreferencesManager.preferencesValue.loadValue('hide_top_bar') !== '1'
  }
},
methods: {
   toggleBar(): {
     this.componentKey = this.componentKey === '1' ? '' : '1'
     PreferencesManager.saveValue('hide_top_bar', this.componentKey);
  }
}
```
The template will be updated as we are naturally updating `hideTopBar`. Anyone updating `hideTopBar` then doesn't need to call to `forceRenderer`. As this is at least not using $forceUpdate though it is good enough.

bafolts (approved):

PR #1990, opened 2020-12-02
add translatable Message for other places besides LogMessage

bafolts:
This re-uses `LogMessage` message parsing logic. For now only `RAW_STRING` is supported. Can expand upon as needed.

### DIFF
--- a/src/components/CreateGameForm.ts
+++ b/src/components/CreateGameForm.ts
@@ -3,13 +3,14 @@ import {Color} from '../Color';
 import {BoardName} from '../BoardName';
 import {CardName} from '../CardName';
 import {CorporationsFilter} from './CorporationsFilter';
-import {$t} from '../directives/i18n';
+import {translateMessage} from '../directives/i18n';
 import {IGameData} from '../database/IDatabase';
 import {ColoniesFilter} from './ColoniesFilter';
 import {ColonyName} from '../colonies/ColonyName';
 import {CardsFilter} from './CardsFilter';
 import {Button} from '../components/common/Button';
 import {playerColorClass} from '../utils/utils';
+import {LogMessageDataType} from '../LogMessageDataType';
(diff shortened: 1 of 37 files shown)


bafolts:
Should have this done by tomorrow. Finishing up the rest of the type conversions for the update of `playerinput.title` using `string | DynamicTranslatableStringModel` and then will add unit test again.

src/components/CreateGameForm.ts
```diff
@@ -196,7 +196,10 @@ export const CreateGameForm = Vue.component('create-game-form', {
       }
     },
     getPlayerNamePlaceholder: function(player: NewPlayerModel): string {
-      return $t('Player ' + player.index + ' name');
+      return $t({
+        message: 'Player {0} name',
+        values: [String(player.index)],
```
alrusdi, line 201: Shouldn't we use LogMessage approach here? May be even reuse it? Action titles with interactive elements can be useful. For example we can show action origin (card, corporation, turmoil event, Player widget etc.)

bafolts: What stands out to me here as code smell is that `$t` has to consider the `DynamicTranslatableStringModel` instead of only being concerned with taking in strings and providing the translation. Re-using `LogMessage` would come with overhead as I can't think of why we would need `timestamp` or `type`. The `message` and `data` is nearly identical though. Could most like have `LogMessage` extend this class here. In this place here only a string is needed. For the titles it probably makes sense to create a `PlayerInputTitle` component as I have to update every place we are displaying the title on a `PlayerInput` anyway. Then we can update all the titles at once as well.

bafolts:
@alrusdi I re-used the same pattern as `LogMessage` and created a `Message` for the shared logic.

alrusdi (approved):
It is great!

PR #1991, opened 2020-12-02
Rewrite Turmoil serialization.

kberg:
This is a series of small changes that grew into something a bit larger, but I think cleaner. It removes Object.assign, for instance, which means serialization and deserialization are far more explicit.

While making this change I also changed the way Turmoil instances are constructed. If you're going to make a Turmoil from a Game, you can use loadFromGame(), which  works great for first-time set-up. But when you want to load from JSON, you no longer need a Game instance, which makes sense, doesn't it? Why create a vestigial game just to create a new instance? I hope that we can propagate this up to the Game and Player instances, and get rid of the need for vestigial players just to, load a Game from the database, or to clone one.

While at it, I made a backwards-compatible change to the save format with the intent to remove the older ways.

This removes references to (most) non-serializable objects in Turmoil - instead storing well-known indexes. It also makes a breaking change away from a couple of typos in field names that can at least be removed now.

Testing was great at finding the bugs in my serialization, but I would like to hear feedback on serialization test requests before you felt comfortable making a change like this, especially because I would like to propagate this to Player and Game.

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -917,12 +917,12 @@ function getTurmoil(game: Game): TurmoilModel | undefined {
     }
 
     let coming;
-    if (game.turmoil.commingGlobalEvent) {
+    if (game.turmoil.comingGlobalEvent) {
       coming = {
-        name: game.turmoil.commingGlobalEvent.name,
-        description: game.turmoil.commingGlobalEvent.description,
-        revealed: game.turmoil.commingGlobalEvent.revealedDelegate,
-        current: game.turmoil.commingGlobalEvent.currentDelegate,
+        name: game.turmoil.comingGlobalEvent.name,
+        description: game.turmoil.comingGlobalEvent.description,
+        revealed: game.turmoil.comingGlobalEvent.revealedDelegate,
+        current: game.turmoil.comingGlobalEvent.currentDelegate,
       };
     }
 
--- a/server.ts
+++ b/server.ts
@@ -916,9 +916,9 @@ function getTurmoil(game: Game): TurmoilModel | undefined {
       };
     }
 
-    let comming;
+    let coming;
     if (game.turmoil.commingGlobalEvent) {
-      comming = {
+      coming = {
         name: game.turmoil.commingGlobalEvent.name,
         description: game.turmoil.commingGlobalEvent.description,
         revealed: game.turmoil.commingGlobalEvent.revealedDelegate,
@@ -944,7 +944,7 @@ function getTurmoil(game: Game): TurmoilModel | undefined {
       lobby: lobby,
       reserve: reserve,
       distant: distant,
-      comming: comming,
+      coming: coming,
       current: current,
     };
   } else {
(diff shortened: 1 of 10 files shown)


src/turmoil/globalEvents/GlobalEventDealer.ts
```diff
@@ -102,19 +102,19 @@ export const NEGATIVE_GLOBAL_EVENTS: Array<IGlobalEventFactory<IGlobalEvent>> =
   {globalEventName: GlobalEventName.SOLAR_FLARE, Factory: SolarFlare},
 ];
 
+const ALL_EVENTS = [
```
kberg, line 105: An extra change which moves allEvents out, it's unnecessary. I did this while trying to change getGlobalEventByName, which creates a new instance of the global event on every call.

kberg:
Man, did I do a sloppy job on the client-side. That's what I get for running `npm run clean && npm run test`

server.ts
```diff
@@ -916,13 +916,13 @@ function getTurmoil(game: Game): TurmoilModel | undefined {
       };
     }
 
-    let comming;
```
bafolts, line 919: I am not familiar with turmoil. I always thought `comming` was the name of something specific to the game :-)

kberg: What do you think of the construction mechanism, Brian? I hate that we have to make fake players just to load a game from the database. I thought I'd start with something less -- risky.

src/styles/turmoil.less
```diff
@@ -158,7 +158,7 @@
     }
 }
 
-.global-event--comming .event-party--bottom {
+.global-event--comin .event-party--bottom {
```
bafolts, line 161: Should be `coming`.

src/turmoil/Turmoil.ts
```diff
@@ -33,15 +33,21 @@ export class Turmoil implements ISerializable<SerializedTurmoil, Turmoil> {
...
     public playersInfluenceBonus: Map<string, number> = new Map<string, number>();
-    public globalEventDealer: GlobalEventDealer;
+    public globalEventDealer: GlobalEventDealer = new GlobalEventDealer();
     public distantGlobalEvent: IGlobalEvent | undefined;
-    public commingGlobalEvent: IGlobalEvent | undefined;
+    public comingGlobalEvent: IGlobalEvent | undefined;
     public currentGlobalEvent: IGlobalEvent | undefined;
 
-    constructor(game: Game) {
+    constructor(game?: Game) {
```
bafolts, line 44: This is a good change and we need to see it across the board. It is difficult to deserialize anything when the constructor automatically changes the state of the class.

bafolts:
When I test a serialization change I usually run through every game currently in the "production" POSTGRES database to ensure those all deserialize without error. Usually the largest risk is how the API change may impact database that is currently in the database. I can provide read only access to the POSTGRES host if you want to test in this way. Could most likely jimmy rig up some github action that also used this read only access come to think of it.

kberg:
Abandoning this for a rewrite.

PR #1995, opened 2020-12-02
Cards vue G+H

chosta:
More cards


src/cards/base/GHGProducingBacteria.ts
```diff
@@ -56,4 +59,20 @@ export class GHGProducingBacteria implements IActionCard, IProjectCard, IResourc
...
       return orOptions;
     }
+    public metadata: CardMetadata = {
+      description: 'Requires 4% oxygen.',
+      cardNumber: '034',
+      requirements: CardRequirements.builder((b) => b.oxygen(4)),
+      renderData: CardRenderer.builder((b) => {
+        b.effectBox((eb) => {
+          eb.empty().startAction.microbes(1);
+          eb.description('Action: Add 1 Microbe to this card.');
```
kberg, line 69: A thought: if every Effect Box has a decsription, you don't really need eb.description. This could be written instead as

```
b.effectBox('Action: Add 1 Microbe to this card.', (eb) => {
  eb.empty().startAction.microbes(1);
});
```

src/cards/base/GreatEscarpmentConsortium.ts
```diff
@@ -19,4 +23,15 @@ export class GreatEscarpmentConsortium implements IProjectCard {
...
+    public metadata: CardMetadata = {
+      cardNumber: '061',
+      requirements: CardRequirements.builder((b) => b.production(Resources.STEEL)),
+      renderData: CardRenderer.builder((b) => {
+        b.productionBox((pb) => {
+          pb.minus().steel(-1).any.br;
+          pb.plus().steel(1);
+        });
+      }),
+      description: 'Requires that you have steel production. Decrease any steel production 1 step and increase your own 1 step.',
```
kberg, line 35: Should steel be capitalized?

src/cards/prelude/HousePrinting.ts
```diff
@@ -18,4 +20,12 @@ export class HousePrinting implements IProjectCard {
     public getVictoryPoints() {
       return 1;
     }
+    public metadata: CardMetadata = {
+      cardNumber: 'P36',
+      renderData: CardRenderer.builder((b) => {
+        b.productionBox((pb) => pb.steel(1));
+      }),
+      description: 'Increase your steel production 1 step.',
```
kberg, line 28: Same?

src/components/card/CardRenderTileComponent.ts
```diff
@@ -47,6 +47,8 @@ export const CardRenderTileComponent = Vue.component('CardRenderTileComponent',
           classes.push('card-tile-symbol-commercial-district');
         } else if (type === TileType.DEIMOS_DOWN) {
           classes.push('card-tile-symbol-deimos-down');
+        } else if (type === TileType.GREAT_DAM) {
```
kberg, line 50: Replace this growing list with a `Map<TileType, string>` Now or next time. In fact, do it on another PR on its own?

kberg (approved):
Minor comments only.

kberg:
Feel free to merge when you are ready. I love it.

src/cards/base/GHGProducingBacteria.ts
chosta: Yes, I was considering this as a further refactoring. Probably doing it soon.

src/cards/base/GreatEscarpmentConsortium.ts
chosta: I'm just copy/pasting from the existing cards. Not sure capital letters for resources is a good idea but we need to be consistent. I'll change it and will try to watch out for this in the future.

src/components/card/CardRenderTileComponent.ts
chosta: I'm adding this to a list of minor refactoring TODOs on the render. I'll probably implement it as soon as I hit around 50% of all cards. At that point most of the logic + exceptions would be known so it might be a good time for an optimization iteration.

PR #2005, opened 2020-12-04
Cards vue I

chosta:
All I cards and a couple from previous letters that needed some special attention.


src/cards/base/Herbivores.ts
```diff
@@ -43,16 +44,17 @@ export class Herbivores implements IProjectCard, IResourceCard {
...
           eb.greenery().startEffect.animals(1);
           eb.description('Effect: When you place a greenery tile, add an Animal to this card.');
         }).br;
-        b.animals(1).productionBox((pb) => pb.minus().plants(1).any).br;
         b.text('1 VP per 2 Animals on this card', CardRenderItemSize.TINY, true);
+        b.animals(1).productionBox((pb) => pb.minus().plants(1).any);
       }),
       description: {
-        text: 'Requires 8% oxygen. Add 1 Animal to this card. Decrease any Plant production 1 step.',
+        text: 'Requires 8% oxygen. +1 animal to this card. -1 plant production',
```
nwai90, line 57: Is the text change intentional? It looks a little strange

src/cards/base/ImportedNitrogen.ts
```diff
@@ -32,4 +34,14 @@ export class ImportedNitrogen implements IProjectCard {
       game.defer(new AddResourcesToCard(player, game, ResourceType.ANIMAL, 2));
       return undefined;
     }
+    public metadata: CardMetadata = {
+      cardNumber: '163',
+      renderData: CardRenderer.builder((b) => {
+        b.tr(1).br;
+        b.plants(4).br;
+        b.microbes(3).digit.asterix().nbsp.or().nbsp;
```
nwai90, line 42: We should remove the `.or()` here as it is `AND` for this card

src/cards/base/Herbivores.ts
chosta: I'm fighting for every pixel here. It's that or no description as I don't want to make custom css for separate cards. (at least not yet)

src/cards/base/ImportedNitrogen.ts
chosta: oops

src/cards/base/Herbivores.ts
kberg: How about restoring the original text for the moment, and // TODO to indicate this is too long? I caution that omitting the "any" needs to be restored regardless.

src/cards/base/InventionContest.ts
```diff
@@ -27,4 +29,10 @@ export class InventionContest implements IProjectCard {
         return undefined;
       });
     }
+    public metadata: CardMetadata = {
+      cardNumber: '192',
+      renderData: CardRenderer.builder((b) => {
+        b.text('look at the top 3 cards from the deck. take 1 of them into hand and discard the other two', CardRenderItemSize.SMALL, true);
```
kberg, line 35: These sentences aren't capitalized.

src/cards/base/Herbivores.ts
kberg: How many cards have you rewritten from their original text?

kberg (approved):

src/cards/base/InventionContest.ts
chosta: No need to as the text is capital letters anyway. But will do, for the sake of punctuation.

src/cards/base/Herbivores.ts
chosta: `Any` needs to be there, I agree. I've changed 2 cards so far. Will figure something out. Maybe a text-overflow propery or smth. Don't want a custom `css` positioning for it but if we have to do it, I'll do it.

PR #2006, opened 2020-12-04
Minor CSS fix for long player names

nwai90:
Current behaviour: Long player names will spill over into 2nd row or more, resulting in uneven heights for player bars.
Screenshot 2020-12-04 at 10 46 33 PM

Change: Add missing property so that `text-overflow: ellipsis` works correctly and keeps all heights constant
Screenshot 2020-12-04 at 10 46 15 PM

### DIFF
--- a/src/styles/players_overview.less
+++ b/src/styles/players_overview.less
@@ -83,6 +83,7 @@
                         max-width: 90px;
                         overflow: hidden;
                         text-overflow: ellipsis;
+                        white-space: nowrap;
                     }
                     .player-name--inactive {
                         cursor: default !important;


kberg:
Not sure I think this is the right approach. white-space: pre will do its best to preserve the full name. But I don't know the details of how THAT would change.

kberg:
> Not sure I think this is the right approach. white-space: pre will do its best to preserve the full name. But I don't know the details of how THAT would change.

Wow, I didn't mean to kill the conversation. Sorry!@

nwai90:
> Not sure I think this is the right approach. white-space: pre will do its best to preserve the full name. But I don't know the details of how THAT would change.

Had a look at the documentation. I believe both `nowrap` and `pre` are equivalent for this particular case where we never want it to break to the next line

alrusdi (approved):

PR #2007, opened 2020-12-04
Added timers

sienmich:
Hi, I'm working on timers.

Right now it looks like this (they are in real time :smile: ) :
Screenshot from 2020-12-04 17-54-56

I'm wondering about two things:

1. How and where to display them? Right now they are below the corp name, but this is temporary place, they make whole player-info bar too fat. They can replace the `active` status (active player is the one whose timer is rising), but they do not replace the `passed` status.

2. Currently they count time only during your main turn (while making actions), not during draft and card buying. Would you like to count time spend on drafting? More like only when everybody else is waiting or just always when somebody is waiting or even always when you can perform some action?

Cheers!


nwai90:
@sienmich Nice work! IMO it suffices to count time only during the active player's turn (Action phase). Placing the timer below corporation name looks like a reasonable trade off.

A couple of points to consider:
1. Could add an option in Preferences to hide the timer for those who don't use it
2. Perhaps make the timers count down instead of up?

ltdstudio:
good desin, maybe we can make a time award for game.
time awards

kberg:
I do not recommend using the timer as a VP source.

>

alrusdi:
>     Would you like to count time spend on drafting

In our "real life" playgroup it was most time consuming part of the game. So i believe we must take into count draft time.

chosta:
This makes the UI box too long. Let me think about where to put the timesrs and don't break the design. Otherwise it's a great improvement, as many experienced players enjoy faster play.

kevinb9n:
Having the timer count downward would imply some kind of consequence for when it reaches zero.  But applying any consequences to this timer will be *extremely* difficult to do fairly without warping incentives. Decisions-per-generation vary widely in number and complexity depending on play style and what cards I manage to get.

I don't see it working. I think counting upward, just so that people have some *awareness* of where the time is going, will make a big difference I think. But the player whose timer is the highest is not necessarily the slowest.

kevinb9n:
(Ideally, it should be ticking when and only when the game is awaiting any kind of decision from me, which I am currently able to give, irrespective of whose turn it is, or whether it is also waiting on other players too.)

sienmich:
Ok, now timers are counting always when you can perform some action (so in at the beginning of the research phase all timers are rising).
The first action is skipped (choosing corporation or first card in initial drafting), so it's not counting time before you open the game.

sienmich:
@chosta, any ideas where to put them? I think they should also be visible in initial drafting phase.

sienmich:
I'm not a big fan of time award or timers counting down, I agree with @kevinb9n. I want the timers to just count the time, that was the best solution in my board terraforming experience.

I'm planning to:
1. Add option to hide them
2. Include them in game results screen
3. Improve UI, put them somewhere else
4. Fix drafting timers (Right now a player who is choosing a draft card does not know if other players are done or are still choosing. So theirs timers are rising, but later when server synchronizes they drop back).

Any more ideas? :smile:

kberg:
For what it's worth, I *do* strongly want countdown timers, but that can be dealt with in another PR.

chosta:
Good work so far!
A couple of comments
> 1. Add option to hide them

IMO, timers should be hidden by default. We have to reduce clutter and as a new feature it would need time to settle. Also, not everyone would be comfortable with the pressure the timer gives.

> 3. Improve UI, put them somewhere else
Let me try and figure it out this week. It's a very tight space in the playerInfo component at the moment but I'll figure it out.

Timers in js are a bit tricky so this would need to be thoroughly tested. I'd suggest we go to a separate branch at some point and put it on a beta server. Please check how timers behave in multiple open tabs too.

kberg:
Is this feature compatible with undo? If I wait 5 minutes and undo an action, what will my timer show?

kberg:
I suspect to make this work well, it might be time to introduce a new table.

sienmich:
Good point, it does not work with undo. I'll think about it.

vsrisuknimit:
> Is this feature compatible with undo? If I wait 5 minutes and undo an action, what will my timer show?

It should count all the time until undo as well.

kberg:
> > Is this feature compatible with undo? If I wait 5 minutes and undo an action, what will my timer show?
> 
> It should count all the time until undo as well.

Agreed. Having stared at the database code for the last couple of days, I have a suggestion for addressing this.

My advice, is to not worry about it for this PR. This will be a great addition.

sienmich:
Yeah. I think simple cloning timer before the undo action and restoring it after could work. Since the undo action is being changed right now I'll leave it for next PR.

sienmich:
Ok, I've done all I wanted in this PR. Ready for review and testing!

sienmich:
2 things that should have been done in near future, but I don't know how:
 - Fixing timer with undo
 - Improving drafting timers (Right now a player who is choosing a draft card does not know if other players are done or are still choosing. So theirs timers are rising, but later when server synchronizes they drop back)

kberg:
> Yeah. I think simple cloning timer before the undo action and restoring it after could work. Since the undo action is being changed right now I'll leave it for next PR.

I was thinking just creating a separate table for the timer. No undo there! :D

sienmich:
So what's next? Somebody wanna review?

vsrisuknimit:
@sienmich conflicts.

tagging @bafolts @kberg for review.

kberg:
I was just going to ask. Ha! Since there's a merge conflict, can you resolve that, and THEN I'll take a first look?

kberg (commented):
Looks like you have another conflict to resolve. Which gives me time to go watch Mandalorian with Wifeberg. @sienmich if you are on the discord channel, drop me a DM so I can ask a couple of questions if I have them. sry. thanks. I appreciate it.

sienmich:
Ya, that merge didn't go as planned. I will look at it tomorrow. Thanks!

src/Timer.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  public start() {
+    if (this.running === 0) {
+      this.started_at = new Date().getTime();
+    }
+    this.running++;
+  }
+
+  public stop() {
```
kberg, line 28: Not sure I understand everything this method is doing -- part of that is of course that I'm not following what's going on with running, or why a player woul dhave mutliple runnings?

src/components/overview/PlayerTimer.ts
```diff
@@ -0,0 +1,33 @@
+import Vue from 'vue';
+import {Timer} from '../../Timer';
+
+export const PlayerTimer = Vue.component('player-timer', {
```
kberg, line 4: I can't tell by reading, but how does this timer stop counting when it is not the player's turn?

kberg (changes requested):
Take a look at the build issue with `ServerModel`.

I like where this is going, but I have some questions about what it's doing. Before making any significant changes, perhaps help me understand `Timer`, and the `running` count.

Also, I'm sorry if this has already been asked and answered: How does this work during a draft phase? Do all players timers count?

kyshantry:
So I'd say that it should definitely count during drafting/buying. 

One idea for a count down is that you get a certain amount of time for each turn plus a stored bank

sienmich:
@kberg, so my main idea of this implementation was to make timers work like this (added just those two lines):
```.ts
class Player {
...
    public setWaitingFor(input: PlayerInput, cb: () => void): void {
      this.timer.start();
      this.waitingFor = input;
      this.waitingForCb = () => {
        this.timer.stop();
        cb();
      };
    }
```
This automatically starts a timer whenever there is an action to perform and stops it when the action is finished. Works with draft, buying cards, choosing prelude or standard projects.
However sometimes after the player finishes action there is a new `timer.start()` called just before (like 0.001 s) the last `timer.stop()`.
For example (chronological order):

1. `timer.start()` called
2. `timer.stop()` called
3. `timer.start()` called
4. `timer.start()` called
5. `timer.stop()` called
6. `timer.stop()` called

To resolve this the Timer has `running` attribute which tells how many more `start()`s there were than `stop()`s. Does it make sense? :)

Should I paste this into the code as documentation?

src/Timer.ts
```diff
@@ -0,0 +1,54 @@
...
+  public start() {
+    if (this.running === 0) {
+      this.startedAt = new Date().getTime();
+    }
+    this.running++;
+  }
+
+  public stop() {
+    this.running--;
+    if (!this.afterFirstAction) {
```
Lynesth, line 30: Why don't we count the time elapsed for the first action ?
Did you mean the time we spend picking corporation/preludes/project cards ?
If so, what's the reason behind this and how does it work out with the initial draft option ?

src/components/overview/PlayerTimer.ts
sienmich: The PlayerTimer (Vue) does not stop, it just refreshes every second. Every second it displays the `Timer.toString()` which changes or not based on the `running` attribute which depends on `start()` and `stop()` calls.

src/Timer.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  public toString() {
+    const date: Date = new Date(this.sumElapsed + (this.running ? new Date().getTime() - this.startedAt:0));
+    const res = [
+      Math.floor(date.getTime() / (1000*60*60)),
+      date.getUTCMinutes(),
+      date.getUTCSeconds()];
+
+    return res.map((s) => String(s).padStart(2, '0')).join(':');
```
Lynesth, line 47: This can simply be:
```ts
    const elapsedSeconds = this.sumElapsed + (this.running ? Date.now() - this.startedAt : 0);
    return new Date(elapsedSeconds * 1000).toISOString().substr(11, 8);
```

kberg:
> @kberg, so my main idea of this implementation was to make timers work like this (added just those two lines):
> 
> ```ts
> class Player {
> ...
>     public setWaitingFor(input: PlayerInput, cb: () => void): void {
>       this.timer.start();
>       this.waitingFor = input;
>       this.waitingForCb = () => {
>         this.timer.stop();
>         cb();
>       };
>     }
> ```
> 
> This automatically starts a timer whenever there is an action to perform and stops it when the action is finished. Works with draft, buying cards, choosing prelude or standard projects.
> However sometimes after the player finishes action there is a new `timer.start()` called just before (like 0.001 s) the last `timer.stop()`.
> For example (chronological order):
> 
> 1. `timer.start()` called
> 2. `timer.stop()` called
> 3. `timer.start()` called
> 4. `timer.start()` called
> 5. `timer.stop()` called
> 6. `timer.stop()` called
> 
> To resolve this the Timer has `running` attribute which tells how many more `start()`s there were than `stop()`s. Does it make sense? :)
> 
> Should I paste this into the code as documentation?

Doesn't have to be this detailed. Don't worry about it for now -- now that I understand that I can look at this more closely.

src/Timer.ts
sienmich: Using the `setWaitingFor` method described above the first `timer.start()` is called with game creation, before players even get their links to play. Best thing I came up with is not counting the first action. Without the initial drafting it does not count choosing initial cards, with initial drafting it does not count drafting first card.
Or should I just ignore that, because at game creation all players' timers are running anyway?

src/Timer.ts
sienmich: Nice, thanks, fixed

src/components/overview/PlayerTimer.ts
```diff
@@ -0,0 +1,33 @@
...
+      handler() {
+        setTimeout(() => {
+          this.updateTimer();
+        }, 1000);
+      },
+    },
+  },
+  methods: {
+    updateTimer: function() {
+      this.timerText = Timer.fromJSON(this.timer).toString();
```
Lynesth, line 29: Creating a new Timer object every second to get elapsed time doesn't seem like a very efficient way to do it.
You could maybe make that `toString` method static and call it with the timer as argument to make it a bit lighter:
```ts
this.timerText = Timer.toString(this.timer);
```

Also this is somehow called twice each time for the active player. Noticed it by adding a `console.log()` after the assignment. Didn't dig to find out why.

firefox_2020-12-14_02-34-26

src/Timer.ts
Lynesth: You can probably ignore it since it's equivalent for all players and it makes the code a bit "complicated" as a side effect. You can get rid of `afterFirstAction` that way. But feel free to gather others opinion on this matter.

Lynesth:
If the server restarts after the corporation cards have been played but before the first player finished his turn, then all timers are active. It's a pretty specific case but can you look into that as well ?

src/components/overview/PlayerTimer.ts
sienmich: Ok, will change `toString`.

Interesting, the `mounted()` hook is called twice, so there are two parallel refreshers going.
Adding
```.ts
  mounted() {
    console.log('mounted PlayerStatus');
  },
```
to PlayerStatus.ts displays two `mounted PlayerStatus` logs, which means the problem is somewhere higher.
Any idea why? It's my first time in Vue.

sienmich: Ok, there is this new Top bar feature with second timer, which is hidden

src/Game.ts
```diff
@@ -314,6 +315,9 @@ export class Game implements ISerializable<SerializedGame> {
           this.setStartingProductions(player);
           this.playerHasPickedCorporationCard(player, new BeginnerCorporation());
         }
+
+        // Setup timers
+        player.timer.visible = gameOptions.showTimers;
```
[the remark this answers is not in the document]
sienmich: `visible` does not change per player. I agree, it should belong to some `GameOptions` object rather than `Player`.
However right now every similar feature works that way. The `PlayerModel` class has attributes:

- `coloniesExtension`
- `boardName`
- `showOtherPlayersVP`
- `initialDraft`

and many more. They are all constructed with `getPlayerModel()` from `ServerModel.ts`. So making it the other way with Timers would be inconsistent.

Maybe lets create new issue with this?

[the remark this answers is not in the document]
sienmich: @Lynesth sugested (https://github.com/bafolts/terraforming-mars/pull/2007/#issuecomment-744025822) to move `visible` to `PlayerModel`. I think it's the right choice.

kevinb9n:
Those of you who want to continue from this to a sudden-death countdown timer, I'm *asking* you to please read #2108, but it's so long that I can't blame you if you don't.

src/Timer.ts
sienmich: Well, I agree now. @kberg, what you think?

src/Timer.ts
```diff
@@ -0,0 +1,54 @@
+export class Timer {
+  public sum_elapsed: number;
+  public started_at: number;
+  public running: number;
+  public visible: boolean;
+  public after_first_action: boolean;
+
+  constructor(
```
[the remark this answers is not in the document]
sienmich: Ok, I just serialized it. Is this the way it should be now?

src/components/overview/PlayerTimer.ts
sienmich: Ok, fixed that `toString`. I think I will not change the top bar, 1 ping every second is not that much. If not then let's create new issue about the top bar?

src/Game.ts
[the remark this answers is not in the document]
sienmich: And I moved it.

sienmich:
> If the server restarts after the corporation cards have been played but before the first player finished his turn, then all timers are active. It's a pretty specific case but can you look into that as well ?

That's was a hard one, 3 hours learning how saving the game works. Fixed! Additionally now `running` is boolean, double win.

src/Player.ts
```diff
@@ -2252,6 +2255,7 @@ export class Player implements ISerializable<SerializedPlayer> {
     }
 
     public setWaitingFor(input: PlayerInput, cb: () => void): void {
+      this.timer.start();
```
kberg, line 2258: So what this means is now setting that you're waiting for something is code for starting the timer. OK. This seems reasonable.

src/SerializedTimer.ts
```diff
@@ -0,0 +1,7 @@
+export interface SerializedTimer {
```
kberg, line 1: Oh, huh. So, it's possible I made you do far more work than I expected. If I asked you to write a Serialized instance, I'm very sorry. If I didn't ask you to do it but you _inferred it_ I'm still sorry. A lot has happened in the last few days. But leave this for now. I still will ask you to make a change in Timer.deserialize.

src/Timer.ts
```diff
@@ -0,0 +1,58 @@
...
+      startedAt: this.startedAt,
+      running: this.running,
+      afterFirstAction: this.afterFirstAction,
+      lastStoppedAt: Timer.lastStoppedAt,
+    };
+  }
+
+  public static deserialize(d: SerializedTimer): Timer {
+    const timer = new Timer();
+    Object.assign(timer, d);
```
kberg, line 29: We are actively removing calls to Object.assign during deseriailzation, and instead, assigning each individual value. Could you please add those five assignments here?

src/Timer.ts
```diff
@@ -0,0 +1,58 @@
...
+    Timer.lastStoppedAt = d.lastStoppedAt;
+    return timer;
+  }
+
+  // start() is always called when player can perform new input action.
```
kberg, line 34: "... when the game is waiting for a player to supply input"?

Lynesth:
Thanks for the changes, I'll take a new look at it in a few hours.

src/Timer.ts
```diff
@@ -0,0 +1,58 @@
...
+  public start() {
+    this.running = true;
```
kberg, line 36: This is much simpler! Wow!

src/Timer.ts
```diff
@@ -0,0 +1,58 @@
...
+    Timer.lastStoppedAt = Date.now();
+    if (!this.afterFirstAction) { // Skipping timer for first move in game
+      this.afterFirstAction = true;
+      return;
+    }
+    this.sumElapsed += Timer.lastStoppedAt - this.startedAt;
+  }
+
+  public static toString(d: SerializedTimer) {
+    const elapsed = d.sumElapsed + (d.running ? Date.now() - d.startedAt : 0);
```
kberg, line 54: What's the toString used for? Is it worth adding whether the timer is enabled?

kberg (changes requested):
I'm marking as "Request changes" only because there are a couple. Plus you probably have merge conflicts. But this doesn't look confusing any more, and should go in fairly soon.

src/SerializedTimer.ts
sienmich: No problem :)

src/Timer.ts
sienmich: Yep, fixed

sienmich:
Currenly testable at https://terraforming-mars-sienmich.herokuapp.com/

src/components/GameEnd.ts
```diff
@@ -26,6 +27,9 @@ export const GameEnd = Vue.component('game-end', {
     getEndGamePlayerRowColorClass: function(color: string): string {
       return playerColorClass(color.toLowerCase(), 'bg_transparent');
     },
+    getTimer: function(p: PlayerModel): string {
+      return Timer.toString(p.timer.serialize());
```
Lynesth, line 31: This should be `p.timer`, it's already serialized.

Lynesth (changes requested):

src/components/GameEnd.ts
sienmich: Thanks, fixed

Lynesth (approved):

chosta:
Can I take a look before merge. Will do in a couple of hours.

kberg:
I'll look again tonight. I took yesterday off to make some art for a change.

>

src/Player.ts
```diff
@@ -2478,6 +2483,8 @@ export class Player implements ISerializable<SerializedPlayer> {
       // Rebuild each drafted cards
       player.draftedCards = cardFinder.cardsFromJSON(d.draftedCards);
 
+      player.timer = Timer.deserialize(d.timer);
```
bafolts, line 2486: OK. Chiming in. This all looks great and the only thing that stands out is that this `timer` will be `undefined` for every game currently in the database. When first rolling this out the type of `timer` needs to be `SerializedTimer | undefined`.

bafolts (changes requested):

src/Player.ts
bafolts: Which begs the question....for games in the database when this timer shows up we will most likely have to just have those users deal with some timer that starts at default value. Eventually wont be a problem.

Lynesth: I always forget about ongoing games, we should probably just wipe the db in those cases :p

I think that instead of having timer `SerializedTimer | undefined`, shouldn't we simply create a new timer when deserializing if `d.timer` is undefined ?

Also regarding new games, they shouldn't see the timers anyway since the option didn't exist.

bafolts: I would put `SerializedTimer | undefined` on the type for now so that typescript can check that we are covering our tracks every place we use the property. It will most likely end up using a new timer when `d.timer` is `undefined` but we will get typescript working for us for free. Putting :robot: s to work!

bafolts: @kberg has been putting comments above places like this with `// remove by [some date]` so that we know the `| undefined` is supposed to be short lived. Future visitors to the code then know they can safely remove the `| undefined`.

Lynesth: Nevermind me, I thought you wanted the whole `timer` to be `| undefined`, not just the argument to the `Timer.deserialize` method. Read your comment a bit too fast.

sienmich: Ok, same thing for `timer.start()` and `.stop()`. They are called even when times ate off. Or do we care only about end game screens?

bafolts: Calling `timer.start()` and `timer.stop()` should be fine even if the values are not used. If the property is `undefined` every old game would fail to deserialize. Only looking to address that situation here with my requested change.

kberg: > OK. Chiming in. This all looks great and the only thing that stands out is that this `timer` will be `undefined` for every game currently in the database. When first rolling this out the type of `timer` needs to be `SerializedTimer | undefined`.

How did I miss that.

kberg (commented):
Yeah, serialization. Otherwise this looks pretty good.

kberg (commented):
Also please add a test in Player.spec.ts like `backward compatible deserialization for actionsThisGeneration` for the timer, a la Brian's concern.

src/Player.ts
sienmich: Ok, fixed

kberg:
I think this is ready to go!

kberg (approved):

vsrisuknimit:
We have playtested this many times. It seems to be working great. Glad it will be ready soon.

sienmich:
Still waiting for @bafolts approve, am I right?

bafolts (approved):

Issue #2010, opened 2020-12-05
vue render error on game setup screen.


bafolts:
If you start a solo game with default settings with console logs open an error is logged to the console.

```
vue.esm.js:628 [Vue warn]: Error in render: "TypeError: Cannot read property 'name' of undefined"

found in

---> <PlayerStatus>
       <PlayerInfo>
         <TopBar>
           <PlayerHome>
             <Root>
vue.esm.js:1897 TypeError: Cannot read property 'name' of undefined
    at Proxy.eval (eval at createFunction (vue.esm.js:11658), <anonymous>:3:565)
    at VueComponent.Vue._render (vue.esm.js:3551)
    at VueComponent.updateComponent (vue.esm.js:4069)
    at Watcher.get (vue.esm.js:4482)
    at new Watcher (vue.esm.js:4471)
    at mountComponent (vue.esm.js:4076)
    at VueComponent../node_modules/vue/dist/vue.esm.js.Vue.$mount (vue.esm.js:9057)
    at VueComponent../node_modules/vue/dist/vue.esm.js.Vue.$mount (vue.esm.js:11953)
    at init (vue.esm.js:3127)
    at createComponent (vue.esm.js:5983)
```

bafolts:
In this situation `player.corporationCard` is `undefined`. The fix is easy but the vue template type checker should have caught this. Will see where it went wrong and fix.

PR #2015, opened 2020-12-05
update vue type checking for all directories

bafolts:
The vue type checking was not checking directories two deep. Updated the blob to catch more of the components and updated issues. Fixing the `player.corporationCard` which can be `undefined` which ended up being an error from the vue type checker fixed the bug in #2010

### DIFF
--- a/compile-vue-templates.js
+++ b/compile-vue-templates.js
@@ -342,6 +342,7 @@ function checkComponent(name, component, dataProperties) {
   lines.unshift('declare function $set(arg1: any, key: string, value: string): void;');
   // seems to be array looper iterating function needs to pass along type information
   lines.unshift('declare function _l(arg1: number, arg2: (item2: number, idx: number) => any): any;');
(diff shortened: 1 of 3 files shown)


compile-vue-templates.js
```diff
@@ -342,6 +342,7 @@ function checkComponent(name, component, dataProperties) {
   lines.unshift('declare function $set(arg1: any, key: string, value: string): void;');
   // seems to be array looper iterating function needs to pass along type information
   lines.unshift('declare function _l(arg1: number, arg2: (item2: number, idx: number) => any): any;');
+  lines.unshift('declare function _l<T>(arg1: {[x: string]: T}, args: (item2: T, idx: number) => any): any;');
```
bafolts, line 345: The `v-for` can work on objects. This adds the typing needed for a usage used in `PlayerStatus` that was erroring. Allows us to iterate on the values of object in template.

kberg (approved):

PR #2017, opened 2020-12-05
show card duplication option first

### LINKED ISSUE #1965: Change default action for Self Replicating Robots
Maybe it is just me but when activating Self Replicating Robots, don't people almost always want to 'Double resources' as opposed to selecting a new card?

I think this can be accomplished with a one line change (but perhaps this screws up something else):
from
orOptions.options.push(new SelectCard(
  'Select card to double robots resource', 'Double resource', robotCards,
to
orOptions.options.**unshift**(new SelectCard(
  'Select card to double robots resource', 'Double resource', robotCards,

Screenshot from 2020-11-27 14-20-48
Screenshot from 2020-11-27 14-28-11

bafolts:
Shows option to duplicate cards first. Changed some `.filter.length > 0` to leverage `.some` while in file. Closes #1965

### DIFF
--- a/src/cards/promo/SelfReplicatingRobots.ts
+++ b/src/cards/promo/SelfReplicatingRobots.ts
@@ -37,18 +37,33 @@ export class SelfReplicatingRobots implements IProjectCard {
     }
 
     public canAct(player: Player): boolean {
-      if (this.targetCards.length > 0) {
-        return true;
-      }
-      if (player.cardsInHand.filter((card) => card.tags.filter((tag) => tag === Tags.SPACE || tag === Tags.STEEL).length > 0).length > 0) {
-        return true;
-      }
-      return false;
+      return this.targetCards.length > 0 ||
+             player.cardsInHand.some((card) => card.tags.some((tag) => tag === Tags.SPACE || tag === Tags.STEEL));
     }
 
     public action(player: Player, game: Game) {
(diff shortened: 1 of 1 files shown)


nwai90 (approved):

src/cards/promo/SelfReplicatingRobots.ts
```diff
@@ -37,18 +37,33 @@ export class SelfReplicatingRobots implements IProjectCard {
...
+      const selectableCards = player.cardsInHand.filter((card) => card.tags.some((tag) => tag === Tags.SPACE || tag === Tags.STEEL));
+
+      if (this.targetCards.length > 0) {
+        const robotCards = this.targetCards.map((targetCard) => targetCard.card);
+        orOptions.options.push(new SelectCard(
+          'Select card to double robots resource', 'Double resource', robotCards,
+          (foundCards: Array<IProjectCard>) => {
+            let resourceCount = 0;
+            for (const targetCard of this.targetCards) {
+              if (targetCard.card.name === foundCards[0].name) {
```
Lynesth, line 55: I noticed this in a few places but I don't think it's necessary since that check is already made before the callback is called. You should be able to directly work on `foundCards[0]` and get rid of the `for` loop.

bafolts: Good point

Issue #2016, opened 2020-12-05
The X icon represents a *positive* integer of the player's choice


kevinb9n:
My interpretation of this post from @Fryxen is that the X icon represents a *positive* integer of the player's choosing.
https://boardgamegeek.com/thread/1958633/article/28564463#28564463

Applies to: Insulation, Power Infrastructure, Energy Market, Hi-Tech Lab, Sulphur-Eating Bacteria, Titan Shuttles.

This raises the question of what the default value in the widget should be. I would suggest perhaps defaulting to the max, and there isn't really a strong need for a "MAX" button anymore then.

Lynesth:
Should be fixed with latest merge.

kevinb9n:
Fine to leave this closed, just mentioning that I'm not 100.0% sure I've guessed right about Insulation, but the answer ought to hopefully appear here at some point.
https://boardgamegeek.com/thread/2559952/questions-about-what-can-be-zero-spending-zero-etc

kevinb9n:
confirmed, I guessed right, X is defined != 0.

PR #2023, opened 2020-12-06
X (any number of) means greater than or equal to 1

Lynesth:
This adds a `min` property to `SelectAmount` and makes sure that the following cards can only be played/acted upon if you have resource/production to spend (no more 0 for 0):
- Energy Market
- Insulation
- Hi-Tech Lab
- Power Infrastructure
- Sulphur-eating Bacteria
- Titan Shuttles

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -689,6 +689,7 @@ function getWaitingFor(
     canUseHeat: undefined,
     players: undefined,
     availableSpaces: undefined,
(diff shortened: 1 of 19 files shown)


Lynesth:
Closes https://github.com/bafolts/terraforming-mars/issues/2016

nwai90 (approved):

PR #2025, opened 2020-12-06
Restructure Political Agendas, remove structural duplication.

kberg:


src/turmoil/PoliticalAgendas.ts
```diff
@@ -0,0 +1,73 @@
...
+
+      const bonus: Bonus = agendaStyle === AgendaStyle.STANDARD ? bonuses[0] : bonuses[Math.floor(Math.random() * bonuses.length)];
+      const policy: Policy = agendaStyle === AgendaStyle.STANDARD ? policies[0] : policies[Math.floor(Math.random() * policies.length)];
+      staticAgendas.set(p.name, {bonusId: bonus.id, policyId: policy.id} as Agenda);
+    });
+    return staticAgendas;
+  }
+
+  private static getAgenda(rulingParty: IParty, staticAgendas: Map<PartyName, Agenda> | undefined): Agenda {
+    if (staticAgendas === undefined) {
```
nwai90, line 55: `staticAgendas === undefined` occurs when `AgendaStyle` is `CHAIRMAN`. Is it correct to return `rulingParty.bonuses[0].id` and `rulingParty.policies[0].id` here? That would always return the default value

src/turmoil/PoliticalAgendas.ts
```diff
@@ -0,0 +1,73 @@
...
+  // The agenda for this generation.
+  thisAgenda: Agenda;
+  // Standard and Random agendas are static through the game. CHAIRMAN changes
+  // by the will of the chairperson.
+  staticAgendas: Map<PartyName, Agenda> | undefined;
+}
+
+export class PoliticalAgendas {
+  public static newInstance(agendaStyle: AgendaStyle, parties: Array<IParty>, firstRulingParty: IParty | undefined): PoliticalAgendasData {
+    if (firstRulingParty === undefined) {
```
nwai90, line 27: This `if` block might be redundant here as the initial ruling party should always be one of the six parties

nwai90 (approved):
Left a couple of small comments/questions

src/turmoil/PoliticalAgendas.ts
kberg: I, uh, I forget. Once we write chairman, this will become much more clear in the meantime.

src/turmoil/PoliticalAgendas.ts
kberg: It isn't. This is called by Turmoil's constructor. `rulingParty` is an attribute on Turmoil of this definitoin:  `    public rulingParty: undefined | IParty = undefined;`

Let this be for now, I'm rewriting a lot of how serialization is going to work, such that this will be less of a problem ongoing.

PR #2027, opened 2020-12-06
protected valley word fix

otolabqu:
Is there any plan to change the translation system? This is not really optimal.


kberg (approved):

kberg:
Thank you. This matches the card, it seems.

PR #2038, opened 2020-12-07
Turmoil serialization updated.

kberg:
This is the first in a series of updates that will adopt the same behavior: no direct access to constructors allowed, it will be private. Deserialization will be renamed to, hey, `deserialize.`

This works just fine with `Turmoil`, `Dealer` and `Player` but once you apply the same thing to `Game` you see just how tightly coupled the current structure for serialization/deserialization is tied to existing behavior. It won't be hard to fix, but it's clear that the whole thing has to be done in a series of small PRs. Hence, the easy one first.

### DIFF
--- a/src/turmoil/Turmoil.ts
+++ b/src/turmoil/Turmoil.ts
@@ -34,41 +34,43 @@ export class Turmoil implements ISerializable<SerializedTurmoil> {
     public dominantParty: undefined | IParty = undefined;
     public lobby: Set<PlayerId> = new Set<PlayerId>();
     public delegate_reserve: Array<PlayerId | 'NEUTRAL'> = []; // eslint-disable-line camelcase
-    public parties: Array<IParty> = [];
+    public parties: Array<IParty> = ALL_PARTIES.map((cf) => new cf.Factory());
     public playersInfluenceBonus: Map<string, number> = new Map<string, number>();
-    public globalEventDealer: GlobalEventDealer;
+    public readonly globalEventDealer: GlobalEventDealer = new GlobalEventDealer();
     public distantGlobalEvent: IGlobalEvent | undefined;
     public commingGlobalEvent: IGlobalEvent | undefined;
     public currentGlobalEvent: IGlobalEvent | undefined;
 
-    constructor(game: Game) {
-      // Init parties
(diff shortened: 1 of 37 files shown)


kberg:
I did not include the that removes Object.assign, or renaming fields. One thing at a time.

src/turmoil/Turmoil.ts
```diff
@@ -364,41 +366,39 @@ export class Turmoil implements ISerializable<SerializedTurmoil> {
     }
 
     // Function used to rebuild each objects
```
chosta, line 368: just 2 typos from previous author I think
`..each objects` -> `..each object`
`..each attributes` -> `..each attribute`
I would argue that we don't even need the comments

chosta (approved):
This looks good to me. @kberg Just remind me what was the benefit of having the static method approach here? Curious to learn about design patterns.

kberg:
> This looks good to me. @kberg Just remind me what was the benefit of having the static method approach here? Curious to learn about design patterns.

It's not going to be particularly obvious from this PR, but the others will make more sense -- particularly when it comes to `Game.`

The benefit making a _static factory method_ outside of the constructor is that it lets you do additional set-up without having to make it part of the constructor. In Java, dependency injection is the pattern you try to follow: make constructs themselves as simple as possible (eg only responsible for setting attributes) and not executing any additional set-up. (Game does this, as does Player -- their constructors are more complicated than they should be.)

If you look at my prototype for Game (which is a big, big mess) https://github.com/kberg/terraforming-mars/commit/a45a8cd7ecc42a50646b6468d989d9a128d2fd62 Board, Dealer and seed are now built outside of the constructor, because there's actually no reason to do it inside the constructor.

Now, how this pertains to the deserialize / loadFromJson methods is this: the deserialization pattern until now was: create a new instance using vestigial short-term values that loadFromJson will replace. That's kind of ugly. So instead, try to reduce Game as much as possible to low-level values. And the deserialization method will be responsible for construction.

The reason that PR is such a mess is because the database integration APIs are particularly designed around the way the current solution is built. But more on that when I get to Game, which will be later today.

bafolts (approved):

chosta:
@kberg Thanks!

PR #2039, opened 2020-12-07
Adjust Dealer serialization using the same approach used for Turmoil.

kberg:
Two additional changes:
1. Removed a few really unnecsesary fields in the serialization protocol. So funny.
2. Moved a couple of Dealer-only tests from `Game.spec.ts` to `Dealer.spec.tMoved a couple of Dealer-only tests from `Game.spec.ts` to `Dealer.spec.ts`.

### DIFF
--- a/src/Dealer.ts
+++ b/src/Dealer.ts
@@ -22,32 +22,21 @@ export class Dealer implements ISerializable<SerializedDealer> {
     public preludeDeck: Array<IProjectCard> = [];
     public discarded: Array<IProjectCard> = [];
     public corporationCards: Array<CorporationCard> = [];
-    private useCorporateEra: boolean = true;
-    private usePreludeExtension: boolean = false;
-    private useVenusNextExtension: boolean = false;
-    private useColoniesNextExtension: boolean = false;
-    private usePromoCards: boolean = false;
-    private useTurmoilExtension: boolean = false;
-    private useAresExtension: boolean = false;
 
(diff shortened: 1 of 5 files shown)


src/Dealer.ts
```diff
@@ -22,32 +22,21 @@ export class Dealer implements ISerializable<SerializedDealer> {
     public preludeDeck: Array<IProjectCard> = [];
     public discarded: Array<IProjectCard> = [];
     public corporationCards: Array<CorporationCard> = [];
-    private useCorporateEra: boolean = true;
```
bafolts, line 25: This is a great change. I didn't realize we were storing these properties even though they were only used once. In a quest to save memory usage on a heroku instance this will do a tiny part!

src/SerializedDealer.ts
```diff
@@ -7,11 +7,4 @@ export interface SerializedDealer {
     deck: Array<IProjectCard | CardName>;
     discarded: Array<IProjectCard | CardName>;
     preludeDeck: Array<IProjectCard | CardName>;
-    useAresExtension: boolean;
```
bafolts, line 10: I appreciate how cool this change is! Typescript has us covered now as well with type safety here.

kberg: I was also surprised at how much neater this particular object is now that it's been dusted and dusted.

bafolts (approved):

PR #2043, opened 2020-12-07
More translatable strings

alrusdi:


src/Player.ts
```diff
@@ -2004,17 +2004,26 @@ export class Player implements ISerializable<SerializedPlayer> {
...
-            } else {
-              return undefined;
-            }
-          }));
-          this.corporationInitialActionDone = true;
-          return undefined;
-        });
+        const initialActionOption = new SelectOption(
+          {
+            message: 'Take first action of ${0} corporation',
```
nwai90, line 2009: Is the change of log message intentional? `Take ${0} first action` might be more succinct

nwai90 (approved):

src/Player.ts
alrusdi: Original was Take ${0}\'s first action. That **'s** meaningless in other-than-english languages.
'Take first action of ${0} corporation' is incorrect?

nwai90: It's correct, I approved (personally prefer shorter message of `Take ${0} first action`)

PR #2044, opened 2020-12-07
Simplified some code

sienmich:
There is a lot of code that is not that good looking even when u just scroll through it.
Does it make sense to improve those little things?

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -206,9 +206,7 @@ export class Game implements ISerializable<SerializedGame, Game> {
       );
 
       // Clone game
-      if (gameOptions !== undefined &&
-        gameOptions.clonedGamedId !== undefined &&
-        !gameOptions.clonedGamedId.startsWith('#')) {
+      if (gameOptions.clonedGamedId !== undefined && !gameOptions.clonedGamedId.startsWith('#')) {
         this.cloneGame(gameOptions.clonedGamedId, this);
         this.clonedGamedId = '#' + gameOptions.clonedGamedId;
         return;
@@ -249,7 +247,7 @@ export class Game implements ISerializable<SerializedGame, Game> {
       }
 
       // Setup Ares hazards
-      if (gameOptions.aresExtension && gameOptions.aresHazards !== false) {
+      if (gameOptions.aresExtension && gameOptions.aresHazards) {
         AresHandler.setupHazards(this, players.length);
       }
 
@@ -402,8 +400,7 @@ export class Game implements ISerializable<SerializedGame, Game> {
     }
 
     public isSoloMode() :boolean {
-      if (this.players.length === 1) return true;
-      return false;
+      return this.players.length === 1;
     }
 
     private setStartingProductions(player: Player) {
@@ -838,10 +835,8 @@ export class Game implements ISerializable<SerializedGame, Game> {
     public gameIsOver(): boolean {
       // Single player game is done after generation 14 or 12 with prelude
       if (this.isSoloMode()) {
-        if (this.generation === 14 || (this.generation === 12 && this.gameOptions.preludeExtension)) {
-          return true;
-        }
-        return false; // Solo mode must go on untill 14 or 12 generation even if Mars is already terraformed
+        // Solo mode must go on until 14 or 12 generation even if Mars is already terraformed
+        return this.generation === 14 || (this.generation === 12 && this.gameOptions.preludeExtension);
       }
       return this.marsIsTerraformed();
     }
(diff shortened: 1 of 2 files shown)


chosta:
It makes sense as long as they pass all the tests. There is a lot of legacy code that we are all slowly improving and all positive changes are welcome.

src/Player.ts
```diff
@@ -1996,7 +1994,7 @@ export class Player implements ISerializable<SerializedPlayer, Player> {
       if (corporationCard !== undefined &&
             corporationCard.initialAction !== undefined &&
             corporationCard.initialActionText !== undefined &&
-            this.corporationInitialActionDone === false
+            !this.corporationInitialActionDone
```
Lynesth, line 1997: I won't ask to change when the code is already using an exclamation mark but in cases it isn't, I'd prefer if you left it as `=== false` as it is very easy to miss the exclamation mark when reading code.

src/Player.ts
```diff
@@ -2146,7 +2144,7 @@ export class Player implements ISerializable<SerializedPlayer, Player> {
         remainingAwards.title = 'Fund an award';
         remainingAwards.buttonLabel = 'Confirm';
         remainingAwards.options = game.awards
-          .filter((award: IAward) => game.hasBeenFunded(award) === false)
+          .filter((award: IAward) => !game.hasBeenFunded(award))
```
Lynesth, line 2147: Same as above.

Lynesth (commented):
Pretty much ok with the changes appart from one little remark.

src/Player.ts
kberg: I'll go one step further -- please restore this. I'm fairly certain I've been asked not to use that mechanism.

sienmich: Ok, that's a good point.
```suggestion
            this.corporationInitialActionDone === false
```

src/Game.ts
```diff
@@ -249,7 +247,7 @@ export class Game implements ISerializable<SerializedGame, Game> {
       }
 
       // Setup Ares hazards
-      if (gameOptions.aresExtension && gameOptions.aresHazards !== false) {
+      if (gameOptions.aresExtension && gameOptions.aresHazards) {
```
kberg, line 250: Please restore this. The !== false was intentional.

src/Player.ts
sienmich:
```suggestion
          .filter((award: IAward) => game.hasBeenFunded(award) === false)
```

src/Game.ts
```diff
@@ -989,7 +982,7 @@ export class Game implements ISerializable<SerializedGame, Game> {
         return;
       }
 
-      if ( ! this.isLastActiveRoundOfDraft(initialDraft, this.initialDraftIteration === 3 ? true : false)) {
+      if ( ! this.isLastActiveRoundOfDraft(initialDraft, this.initialDraftIteration === 3)) {
```
kberg, line 985: If this stays as is, remove the whitespace between ! and its subexpression.

src/Game.ts
```diff
@@ -1384,7 +1377,7 @@ export class Game implements ISerializable<SerializedGame, Game> {
       });
 
       // Oceans are not subject to Ares adjacency production penalties.
-      const subjectToHazardAdjacency = (tile.tileType === TileType.OCEAN) ? false : true;
```
kberg, line 1387: Just noting this one -- not sure it shouldn't change, but perhaps @bafolts can share some thoughts.

kberg (commented):
Nobody submit this until @bafolts has a look. I know some constructs (eg !.) present a risk of killing server, and so he has issued an edict outlawing them.

src/Game.ts
bafolts: `aresData` can be `undefined` but `aresHazards` is only of type `boolean` and not optional so using either should be fine. As far as I know using either here should be fine.

src/Game.ts
bafolts: The change seems simpler than using a ternary. Makes sense to me. The checks seem identical to me.

bafolts (approved):

src/Game.ts
```diff
@@ -1774,8 +1767,7 @@ export class Game implements ISerializable<SerializedGame, Game> {
           };
         } else if (element.player) {
           // Correct Land Claim
-          const player = this.players.find((player) => player.id === element.player!.id);
-          space.player = player;
+          space.player = this.players.find((player) => player.id === element.player!.id);
```
bafolts, line 1770: Want to avoid using `element.player!` as in the situation where `element.player` happened to be `undefined` this will throw a runtime exception. In this cases it was only copied and not introduced. You may be thinking, but element.player has to exist, we checked in the statement above! That is true but within this anonymous function typescript doesn't know  when this callback is going to be called. As far as I know there is no typing to tell typescript this function is going to be called synchronously. That would avoid needing to use `element.player!` in this situation.

Issue #746, opened 2020-05-30
Suggestion: Auto reload/refresh view when "Undo turn" action is taken


nwai90:
**Context:** Currently, if "Undo turn" is done, a player needs to refresh the page to see the correct undone state. This can create some strange state errors

PR #2045, opened 2020-12-07
Allow undoing actions 1 by 1 when last in gen or solo

### LINKED ISSUE #1309: Undo after all other players pass
It seems like the undo function does not work after the first action taken by the last active player in a gen.  As example, in a 2-player game if one player passes and the other player has 8 actions left, that player cannot undo any actions after the first of those 8.

Lynesth:
Allow undoing every single action instead of every odd ones when playing solo or when all other players have passed for the generation. You can actually undo multiple actions in a row with this if you want to.

Also automatically refreshes for the active player. It doesn't for other players until it's their turn to do something or they need to manually reload the page. I have an idea to have it work properly after this is merged.

Changed action text from `Undo Turn` to `Undo last action`. Modified translations files accordingly but didn't modify the translations themselves.

Fixes #1309
Fixes half of #746


kberg (commented):
I haven't looked closely. This looks as though it would double a game's storage. Hmm.,

Lynesth:
> I haven't looked closely. This looks as though it would double a game's storage. Hmm.,

Ok so in the PR, it saves before doing an action (at the beginning or `takeAction`).
When undoing it will reload the penultimate save and delete the last save from the database.
https://github.com/Lynesth/terraforming-mars/blob/9fe9b4c2ace53bd3cb73a14efe1025b540094ecf/src/Player.ts#L1749-L1751

Example:
- Save n° 3 is made
- Player does a Power Plant standard project
- Save n° 4 is made
- Player hits undo
- Save n° 3 is reloaded from database
- Save n° 4 is deleted from database
- Game tries to save n° 3 again but it doesn't since it's already in the database
- Player does an Asteroid standard project instead
- Save n° 4 is made
- ...

alrusdi (approved):

bafolts (changes requested):
I am going to review this prior to merging. Most times I check there are over 10k rows in the database for games. This change could have large side effects we may not be considering.

bafolts:
> I misunderstood your concern. You are correct that it will double the amount of saves for a game. But I don't think this is an issue. I mean, it is 2020, databases are pretty fast and disk space for such data shouldn't be an issue.

The demo site is using a `hobby-basic` plan. Seems to have 10,000,000 rows available and 10GB available. Looking right now we are using 59k rows and 1.2GB. Doubling this we would be at 118k rows and 2.4GB storage which is still within these limits. We store a lot of data we don't have to and @kberg is working on updating the serialization process. Some of those changes will cut down in the amount of storage. We are nowhere near the 10,000,000 rows so this change seems reasonable.

Lynesth:
> Looking right now we are using 59k rows and 1.2GB.

Games are supposed to remove all but the first and last save when they are finished.
How long do you keep unfinished games to end up with 59k rows ?
Assuming (high end of the spectrum) a 4p game that ends in 10 generations, they would probably have around 400 rows just before finishing the game.
Since you have data, can you extract the number of rows and player count grouping by game_id ?

src/Player.ts
```diff
@@ -1743,9 +1744,11 @@ export class Player implements ISerializable<SerializedPlayer> {
 
     // Propose a new action to undo last action
     private undoTurnOption(game: Game): PlayerInput {
-      return new SelectOption('Undo Turn', 'Undo', () => {
+      return new SelectOption('Undo last action', 'Undo', () => {
         try {
-          Database.getInstance().restoreGame(game.id, game.lastSaveId, game);
+          Database.getInstance().restoreGame(game.id, game.lastSaveId - 2, game);
```
bafolts, line 1749: These queries don't necessarily run synchronously. We need to introduce callbacks which are called once the queries finish. It shouldn't be an issue I suppose if these didn't run synchronously unless someone clicked undo and took an action fast enough before `deleteGameNbrSaves` had finished. 

This is unrelated but we need to update all these database calls to take a callback for when they have finished.

src/Player.ts
```diff
@@ -1962,13 +1965,32 @@ export class Player implements ISerializable<SerializedPlayer> {
...
+      const allOtherPlayersHavePassed: boolean = passedPlayers.length === players.length - 1 && passedPlayers.includes(this.color) === false;
+
+      if (game.lastSaveId === 0) {
+        /*
+         * This is needed to create the initial save, just before the first player
+         * takes its first action of the game, since saves don't overwrite themselves
+         */
+        game.lastSaveId = 1;
+      }
+      Database.getInstance().saveGameState(game.id, game.lastSaveId, game.toJSON(), players.length);
```
bafolts, line 1991: This seems to duplicate https://github.com/bafolts/terraforming-mars/blob/main/src/Game.ts#L319 . Are both of these needed?

src/Player.ts
```diff
@@ -1962,13 +1965,32 @@ export class Player implements ISerializable<SerializedPlayer> {
...
       if (game.deferredActions.nextForPlayer(this.id) !== undefined) {
         game.deferredActions.runAllForPlayer(this.id, () => {
           this.takeAction(game);
         });
         return;
       }
 
+      const players = game.getPlayers();
+      const passedPlayers = game.getPassedPlayers();
+      const allOtherPlayersHavePassed: boolean = passedPlayers.length === players.length - 1 && passedPlayers.includes(this.color) === false;
```
bafolts, line 1982: This seems useful enough to move into a separate function. Could also unit test it easier in a separate function.

src/Player.ts
```diff
@@ -1992,7 +2014,7 @@ export class Player implements ISerializable<SerializedPlayer> {
         game.phase = Phase.ACTION;
       }
 
-      if (game.hasPassedThisActionPhase(this) || this.actionsTakenThisRound >= 2) {
+      if (game.hasPassedThisActionPhase(this) || (allOtherPlayersHavePassed === false && this.actionsTakenThisRound >= 2)) {
```
bafolts, line 2017: Is this change for `allOtherPlayersHavePassed` here to allow a player to continue taking actions to avoid going through the `playerIsFinishedTakingActions` function? Did this change introduce a problem with going through `playerIsFinishedTakingActions`?

src/Player.ts
```diff
@@ -1743,9 +1744,11 @@ export class Player implements ISerializable<SerializedPlayer> {
 
     // Propose a new action to undo last action
     private undoTurnOption(game: Game): PlayerInput {
-      return new SelectOption('Undo Turn', 'Undo', () => {
+      return new SelectOption('Undo last action', 'Undo', () => {
         try {
-          Database.getInstance().restoreGame(game.id, game.lastSaveId, game);
+          Database.getInstance().restoreGame(game.id, game.lastSaveId - 2, game);
+          Database.getInstance().deleteGameNbrSaves(game.id, 1);
+          this.usedUndo = true; // To prevent going back into takeAction()
```
bafolts, line 1751: This instance of `Player` is going away. Once `restoreGame` is called playback will resume on a new instance of `Player`. I don't think this `usedUndo` is needed. I believe you can instead set `waitingForCb` to `undefined`. Once `restoreGame` is called on `game` this instance of `Player` will eventually be garbage collected. I would set `this.waitingForCb` to `undefined` and remove the `usedUndo`.

bafolts (changes requested):
OK. I have a few questions but overall this looks like it will work and the added storage complexities shouldn't push us over the demo server limit and we are continuing to improve upon how much data we store in postgres. I see that eventually decreasing by a lot as we store a lot of complete JSON objects versus look-ups, for example we will store an entire `Player` versus `PlayerId`.

The one thing I would change is to get rid of `usedUndo`, by setting the `waitingForCb` of that `Player` to `undefined` I believe the introduced state can be avoided. Once the game is restored we end up on a new instance of player. The less state we have on `Player` should be better as it is less we have to manage and keep updated.

src/Player.ts
Lynesth: I didn't check in details how replaying a game from the seed works. Does it need the initial save before players even pick their corp and all ?

src/Player.ts
Lynesth: Yes that was the point when I first started working on this, but I'm not sure it's still needed.
And I think it's probably inducing an error regarding deferred actions now that I look at it. I'll check in more details later today.

src/Player.ts
Lynesth: That was my plan at first, but you can't tamper with waitingForCb anymore as soon as `process` is called.
https://github.com/bafolts/terraforming-mars/blob/f96db26c3bea7b4cf43db419748d402cd27d75e8/src/Player.ts#L2206-L2213
That's why I had to came up with this to prevent reentering the whole `takeAction` logic.

src/Player.ts
kberg: I'm a little surprised at the seed thing, too. I was staring at it this evening and it makes no sense upon restore.  But please don't address that without talking with me.

kberg:
A question: it doesn't look like it, but are you adding or changing IDatabase behavior? I'm actually looking -- just this minute -- at changing the behavior of undo. Which I think understands how a server deals with games and game state.

src/Player.ts
bafolts: If this isn't needed I would remove it for now. There is enough changing with this change that the less we can change will be better.

src/Player.ts
bafolts: I would put this block into some `game.saveState()`. I don't know how many games are using undo but if undo is not enabled we don't have to save every game state and we could continually override one `lastSaveId`. This should help alleviate memory usage concerns as I am assuming not every game is using the undo option. If `undoOption` is `false` the `saveState` function can override the one instance we have in the database. If `undoOption` is `true` the `saveState` can do what you have here with keeping track of every state.

bafolts: I will check out what the `Game.ts#L319` is trying to do and if it would still be needed with this change.

Lynesth:
Ok so taking the reviews into account + several discussions on discord:
- We're back to 1 save at the start of a player's turn if undo option is not selected. It will keep the number of rows on the low side.
- Behavior when undo option is selected is unchanged compared to my first comment on the subject in this PR.
- Didn't touch the first save thing. Will make a separate PR for it if needed but it seems @kberg is working on games loading, so he will probably take care of this.
- Added `player.allOtherPlayersHavePassed()` method.

-----

> A question: it doesn't look like it, but are you adding or changing IDatabase behavior? I'm actually looking -- just this minute -- at changing the behavior of undo. Which I think understands how a server deals with games and game state.

Nope didn't touch that at all.

src/Player.ts
```diff
@@ -2203,6 +2225,13 @@ export class Player implements ISerializable<SerializedPlayer> {
       });
     }
 
+    public allOtherPlayersHavePassed(game: Game): boolean {
+      if (game.isSoloMode()) return true;
+      const players = game.getPlayers();
+      const passedPlayers = game.getPassedPlayers();
+      return passedPlayers.length === players.length - 1 && passedPlayers.includes(this.color) === false;
```
bafolts, line 2232: This has to be `passedPlayers.includes(this.id)`. The `passedPlayers` is an array of `PlayerId`.

src/Player.ts
```diff
@@ -1970,13 +1973,32 @@ export class Player implements ISerializable<SerializedPlayer> {
...
+      if (game.lastSaveId === 0) {
+        /*
+         * This is needed to create the initial save, just before the first player
+         * takes its first action of the game, since saves don't overwrite themselves
+         */
+        game.lastSaveId = 1;
+      }
+
+      if (this.actionsTakenThisRound === 0 || game.gameOptions.undoOption) {
+        Database.getInstance().saveGameState(game.id, game.lastSaveId, game.toJSON(), players.length);
```
bafolts, line 1998: If `game.lastSaveId++` happened before this call to `saveGameState` is there any need to have the separate block above separately setting to `1` when `lastSaveId` is `0`? Seems unnecessary if this added first. I could be missing something.

src/Player.ts
Lynesth: https://github.com/bafolts/terraforming-mars/blob/3c8b349be32274f130b3ec8e5182df74146072b5/src/Game.ts#L1315

src/Player.ts
Lynesth: It didn't get call on the very first turn of the game, right after everyone has chosen their initial cards.
I could also simply call it after the initial save here. I suppose it'd make more sense:
https://github.com/bafolts/terraforming-mars/blob/3c8b349be32274f130b3ec8e5182df74146072b5/src/Game.ts#L321-L323

bafolts (approved):

PR #2046, opened 2020-12-07
Update README.md

ltdstudio:
update docker image server in Readme


kberg:
This is a minor change but can you add more meaningful descriptions in the future? At least move from describing what file changed to what meaningful difference changed.

kberg (approved):

ltdstudio:
ok

PR #2054, opened 2020-12-08
Make GlobalEventDealer serializable on its own.

kberg:
This lets me put its behavior in one place. This is going to make it easier as I continue to adjust serialization behavior. Also added a sanity serialization/deserialization test, which will come in handle once I add a SerializedGlobalDealer.

@chstoa you asked why this style is better, and here's another example of why: first, by adding a fully-built dealer into Turmoil, Turmoil is no longer responsible for figuring out anything about how to make a deck. Second, Turmoil makes GlobalEventDealer readonly. Which meanse that I can't just do

turmoil.globalEventDealer = blah

once the object is constructed. So, fine, instead, just feed it into its constructor.


kberg:
By the way, this depends on #2053, or, you can ignore #2053 and deal with the indentation here.

PR #2056, opened 2020-12-08
There is always a ruling party and a dominant party.

kberg:
Is this safe to do? I had a discussion on discord about whether there is a time where there is no ruling party, and the answer I got was

> There is always a ruling party during the gen. Even in gen 1, greens are ruling, although I don't think the green's ability is in play during the prelude play. During preludes and the solar phase, ruling party effects do not apply.

I know already that the solar phase won't take ruling effects into account, but what about preludes? Does this change cause breaking behavior during an, um, prelude phase?

### DIFF
--- a/src/turmoil/SerializedTurmoil.ts
+++ b/src/turmoil/SerializedTurmoil.ts
@@ -6,7 +6,7 @@ import {PlayerId} from '../Player';
 export interface SerializedTurmoil {
     chairman: undefined | PlayerId | 'NEUTRAL';
     rulingParty: IParty;
-    dominantParty: undefined | IParty;
+    dominantParty: IParty;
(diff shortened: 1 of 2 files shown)


Lynesth:
> I know already that the solar phase won't take ruling effects into account, but what about preludes? Does this change cause breaking behavior during an, um, prelude phase?
No it's fine. This is properly checked and a distinction is made between Phase.ACTION and Phase.PRELUDES.

kberg:
I added `dominantParty` to this PR based on a comment by @Lynesth.

bafolts (approved):

bafolts:
> there is always a ruling party

ain't that the truth

Issue #2062, opened 2020-12-08
"node build/server.js" can't find module "terraforming-mars\build\server.js


JeremyDietrich:
Hi there, I don't know much about how the files and stuff are supposed to work, but last time I downloaded this I followed the simple directions and it worked perfectly. I decided I wanted to get the newest stuff so I deleted my terraforming-mars directory for a fresh set of files, but now when I run the simple directions it breaks on "npm start" with the error message below. Any ideas what to do? Thanks, and sorry for the unhelpful comment!

> terraforming-mars@1.0.0 start C:\Users\me\terraforming-mars
> node build/server.js

internal/modules/cjs/loader.js:969
  throw err;
  ^

Error: Cannot find module 'C:\Users\me\terraforming-mars\build\server.js'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:966:15)
    at Function.Module._load (internal/modules/cjs/loader.js:842:27)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
    at internal/main/run_main_module.js:17:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []

kberg:
Just a guess: did you run npm install?

JeremyDietrich:
Yes, I ran git clone, npm install, npm start. The npm install had a couple of warnings that some optional dependencies were skipped, but no errors or vulnerabilities

bafolts:
You need to run `npm run build` now prior to `npm start`.

bafolts:
Instructions from README.md

https://github.com/bafolts/terraforming-mars#windows-or-linux-game-server-running

JeremyDietrich:
Ahhh, thanks, right missed that at the beginning. But now it says "cannot find module 'dotenv' "

JeremyDietrich:
OK I forgot to run "npm install" first this time, sorry about that. Re-trying now

JeremyDietrich:
Now there are almost 70k errors that seem to be linebreak-style errors...

bafolts:
Can you post what some of them are? I am not sure how many people build on windows machines. Nothing is OS specific as far as I know so this should build on any OS running node. Could be forgetting something though.

JeremyDietrich:
Every single one as far back as I can scroll is "error  Expected linebreaks to be 'LF' but found 'CRLF'  linebreak-style". And it built fine on Windows one or two months ago, when I last updated it

[...]

PR #2063, opened 2020-12-09
add gitattributes file for windows users

bafolts:
I don't use windows but I believe this will address issue reported in #2062 . New linting rule is checking for lines ending in LF.


kberg (approved):
Wish this was here when I worked those first three weeks on the project! :D

PR #2057, opened 2020-12-08
Significantly rework Turmoil serialization.

kberg:
Remove syntactic errors in field names, reduce the amount of data stored, and write both backward- and forward- compatible tests.

Also removed Object.assign, which treats class instances as just data stores, in the extralinguistic way we want to avoid ongoing.

It's late and I myself haven't re-read this. But I'm sending it out for review anyway.

PR #2069, opened 2020-12-10
set properties without Object.assign

bafolts:
Added regression test. when deserializing a `Turmoil` the usage of `Object.assign` we setting `players` to a type we aren't enforcing with typescript!

### DIFF
--- a/src/turmoil/Turmoil.ts
+++ b/src/turmoil/Turmoil.ts
@@ -368,10 +368,12 @@ export class Turmoil implements ISerializable<SerializedTurmoil> {
     // Function used to rebuild each objects
     public static deserialize(d: SerializedTurmoil): Turmoil {
       const turmoil = new Turmoil(d.rulingParty.name, d.dominantParty.name);
-      // Assign each attributes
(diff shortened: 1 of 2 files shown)


kberg:
I haven't looked at your PR yet but I'm doing some of this in #2057 I'm in the middle of a precarious game of Targi so I'll get to this in a bit.

bafolts:
This will fix build for deployment tomorrow. Shouldn't step on the work in #2057 too much.

kberg (approved):

kberg:
Sorry about the issue.

bafolts:
I am excited every other game didn't break. Historically during these issues we would inadvertently take down every other game!

vsrisuknimit:
> This will fix build for deployment tomorrow. Shouldn't step on the work in #2057 too much.

@bafolts could you let me know when it is deployed? Yesterday I told folks on Discord to avoid the main demo server. I will let them know it is okay to use after deployment.

bafolts:
This is deployed. The demo server is at record memory usage so I assume people have gotten word.

PR #2070, opened 2020-12-10
Fix test flakiness by determining the list of parties that are serialized and deserialized.

kberg:


nwai90 (approved):

bafolts:
Thanks! Saw this while I was developing but I was in too much of a panic and rush when I at first thought server may be down for everyone. Forgot to go back and address it. Needed to remember my 2020 new years resolution of patience.

kberg:
No problem -- I only noticed it after my own PR went in, and so I thought
it was my fault.

PR #2072, opened 2020-12-10
santa and xmas tree

vsrisuknimit:
- Added Santa and Christmas tree
- Moved boards to a different folder

@bafolts @chosta do we need v-if to check that the Santa and tree are displayed only in December/January? Or just remove them manually afterward.

image


chosta:
Ho-Ho-Ho!
Just remove the code after the celebrations.

chosta (approved):
A note for the future. 
It's better to try and reduce the size of the images before adding them to the app instead of using background size. Background-size only scales the image but the size remains the same.

bafolts:
Can remove martian santa in January manually -- or sooner if people seem to hate it. I find it hard to believe that if after many generations while terraforming mars that santa would be lost to time. From what I understand of the man their naughty and nice list is not limited by space nor time.

PR #2073, opened 2020-12-10
Move all the server model code out of server.ts

kberg:
... letting server.ts focus on, well, serving.

This will also make it easier to write unit tests as needed.


src/server/ServerModel.ts
```diff
@@ -0,0 +1,572 @@
+require('dotenv').config();
```
bafolts, line 1: This shouldn't be needed here in `ServerModel`. Someone added this at one point to read configuration values from some environment file. Assuming this was a copy and paste deal from `server.ts`. I would remove before flight.

bafolts (approved):
I would remove the `require('dotenv').config()` being included in `ServerModel` before merging.

PR #2092, opened 2020-12-12
load games when needed from database

bafolts:
Whenever the server starts up we currently load every game from the database into javascript memory. With the increasing number of games being played on the demo site we are running out of javascript memory. This change loads only the game id and player ids into javascript memory when the server is starting. Unless a game is requested we then don't bother wasting javascript memory on the game. The storing of game id and player id in memory is used to avoid constantly querying the database for situations which may results in 404 as this cheaper heroku instance doesn't have the greatest postgres database either.

Will add unit tests and then move this to a real PR.


server.ts
```diff
@@ -293,22 +293,15 @@ function loadGame(req: http.IncomingMessage, res: http.ServerResponse): void {
...
-          }
-          gameLoader.addGame(gameToRebuild);
-        },
-      );
-      res.setHeader('Content-Type', 'application/json');
-      res.write(getGame(gameToRebuild));
-      res.end();
+      gameLoader.getGameByGameId(game_id, (game) => {
+        if (game === undefined) {
+          // game was not found!
```
bafolts, line 298: Need to have this return 404. Looking at this now this current version must not have worked if game wasn't found and would have returned the `gameToRebuild` prior to it being restored.

src/database/GameLoader.ts
```diff
@@ -36,66 +36,98 @@ export class GameLoader {
...
-      const pendingGames = this.pendingGame.get(gameId);
-      if (pendingGames !== undefined) {
-        for (const pendingGame of pendingGames) {
-          pendingGame(this.games.get(gameId));
-        }
-        this.pendingGame.delete(gameId);
+    private loadPlayerIntoMemory(playerId: string, cb: (game: Game | undefined) => void): void {
+      const gameId = this.knownPlayerIds.get(playerId);
+      if (gameId !== undefined) {
+        this.loadGameIntoMemory(gameId, cb);
```
kberg, line 86: This is brittle. it assumes that calling loadPlayerIntoMemory occurs when loadGmeIntoMemory hasn't been called. Some of that is a naming issue, some of that is that we could solve a lot of this with Promises -- I think.

In the meantime, this probably works. I say probably because to be certain I have to read about when and how lGIM and lPIM are called, hence, context makes this brittle. A little flexibility here would go a long way. Or at minimum, clear documentation.

src/database/GameLoader.ts
```diff
@@ -36,66 +36,98 @@ export class GameLoader {
...
-      for (const pendingGame of Array.from(this.pendingGame.values())) {
-        for (const cb of pendingGame) {
-          cb(undefined);
-        }
-      }
+      // are waiting for player or games, since
+      // we now have every game id and player id
+      // in database load any that have been requested
+      this.pendingGame.forEach((cbs, gameId) => {
+        this.loadGameIntoMemory(gameId, (game) => {
```
kberg, line 117: If this is happening in `onAllGamesLoaded` then all games are not loaded. Consider renaming this method to `onAllKeysLoaded`.

kberg: Also, if this APi is doing what it's doing, how does the entire PR fulfill its promise to only load games when needed?

src/database/IDatabase.ts
```diff
@@ -37,6 +38,14 @@ export interface IGameData {
  * in the game. Why, I have no idea, says kberg.
  */
 export interface IDatabase {
+
+    /**
+     * Pulls most recent version of game
+     * @param game_id the game id to load
+     * @param cb called with game if exists
```
kberg, line 45: "called with game if exists, else err" ?

kberg (changes requested):
There's a lot I like about this, but I'm putting on my extra-pedantic lenses for this one.

src/database/GameLoader.ts
bafolts: In the updated version I tried to make this more idempotent. This became `loadPlayer` and always loads the player checking every map to see if it needs to. I haven't spotted it yet but it feels like this is another class hiding in here which if pulled out would make this instantly clearer. I added a queue of games and players that need to be loaded that get worked through one at a time. It gives preference to loading games as one game contains multiple players that would become available. I am attempting to avoid the situation where 5 players are waiting to take a turn while the server is restarting. These 5 players would all end up requesting their player id simultaneously and if we don't act on the requests in a queue we could end up re-loading the same game 5 times. Most of this comment I am typing I will put in class documentation as it may not be clear why all of this complexity is here.

bafolts: Once a game is up and running we never call out to the database except for saving state and undo operations. When heroku is restarting, the `server.ts` starts up and any "waiting for" request will query for their player id. When the data isn't available they have to wait until all of the game id and player ids have been loaded from the database. Once those are loaded we have to work through any request that happened to be waiting. Anyone starting a new game during this time shouldn't be impacted and will avoid all of this logic.

src/database/GameLoader.ts
bafolts: I changed this in my recent push and got rid of `start`. Until any request for player or game is made we never hit the database.

kberg:
OK, I read your responses and it looks like you're mid-repair. Ping me when you feel like this is ready for another look.

server.ts
bafolts: Ugh, still need to do this. It wasn't here before but makes sense to add.

server.ts
```diff
@@ -258,22 +256,16 @@ function loadGame(req: http.IncomingMessage, res: http.ServerResponse): void {
...
-      const player2 = new Player('test2', Color.RED, false, 0);
-      const gameToRebuild = new Game(game_id, [player, player2], player);
-      Database.getInstance().restoreGameLastSave(
-        game_id,
-        gameToRebuild,
-        function(err) {
-          if (err) {
-            return;
-          }
-          gameLoader.addGame(gameToRebuild);
```
bafolts, line 271: Ugh. I don't know if anyone even uses this feature, but my change wont work. At some point later in the UI this game ends up being created and the new version needs to be added to the map at that time. Very frustrating. I will have to add a `GameLoader::delete` to remove a game so that this can then add the game back with this new version.

bafolts: I need to address this prior to merging. This needs to update the copy of `Game` we have in the database. Will need a `GameLoader::delete` which removes our local copy in javascript so that on subsequent calls to `getByGameId` we pull the most recent version from the database into javascript memory.

bafolts:
OK. @kberg I think this is now ready. I added a `GameLoader::remove` so when we "reload" a game we can remove the copy in javascript and get fresh data from database.

server.ts
```diff
@@ -8,7 +8,6 @@ import * as querystring from 'querystring';
 import * as zlib from 'zlib';
 
 import {BoardName} from './src/BoardName';
-import {Color} from './src/Color';
```
kberg, line 11: Already a massive improvement.

server.ts
```diff
@@ -21,9 +20,8 @@ import {Server} from './src/server/ServerModel';
 const serverId = process.env.SERVER_ID || generateRandomServerId();
 const styles = fs.readFileSync('styles.css');
 let compressedStyles: undefined | Buffer = undefined;
-const gameLoader = new GameLoader();
 const route = new Route();
-const gameLogs = new GameLogs(gameLoader);
+const gameLogs = new GameLogs();
```
kberg, line 24: Oh my goodness, seriously, this is getting better.

server.ts
```diff
@@ -257,23 +255,19 @@ function loadGame(req: http.IncomingMessage, res: http.ServerResponse): void {
...
-            return;
-          }
-          gameLoader.addGame(gameToRebuild);
-        },
-      );
-      res.setHeader('Content-Type', 'application/json');
-      res.write(getGameModelJSON(gameToRebuild));
-      res.end();
+      // remove copy in javascript
+      GameLoader.getInstance().remove(game_id);
```
kberg, line 259: I haven't read the new `GameLoader` yet, but my instinct is that clearing the copy probably belongs as part of `getByGameId`, and possibly doesn't happen if the meat of `getByGameId` fails.

server.ts
```diff
@@ -569,19 +563,18 @@ function serveStyles(req: http.IncomingMessage, res: http.ServerResponse): void
   res.end(buffer);
 }
 
-gameLoader.start(() => {
```
kberg, line 572: Huh. I'm very curious now.

kberg: I'm reading this like watching a murder mystery unfold.

No Color import? GameLogs doesn't need a GameLoader? A donut hole with a little donut inside it? With its own little donut hole?

src/database/GameLoader.ts
```diff
@@ -2,149 +2,228 @@
...
+  private state = State.WAITING;
+
+  // indicates if game is being loaded from database
+  private loadingGame = false;
+
+  // games available in javascript memory, undefined represents game
+  // in database not yet loaded
+  private readonly games = new Map<GameId, Game | undefined>();
+
+  // player ids which we know exist mapped to game id
```
kberg, line 38: oh my gosh this documentation is fantastic.

src/database/GameLoader.ts
```diff
@@ -2,149 +2,228 @@
...
+  private readonly gameRequests: Array<[GameId, LoadCallback]> = [];
+
+  // requests for player that are waiting to load
+  private readonly playerRequests: Array<[PlayerId, LoadCallback]> = [];
+
+  private static instance?: GameLoader;
+
+  public static getInstance(): GameLoader {
+    if (GameLoader.instance === undefined) {
+      GameLoader.instance = new GameLoader();
```
kberg, line 51: Add `private constructor() { }` I think?

src/database/GameLoader.ts
```diff
@@ -2,149 +2,228 @@
...
-      }
+  public add(game: Game): void {
+    this.games.set(game.id, game);
+    for (const player of game.getPlayers()) {
+      this.playerIds.set(player.id, game.id);
     }
+  }
+
+  /**
+   * Determines if game is available in javascript
```
kberg, line 64: Why the short line length? /minor

src/database/GameLoader.ts
```diff
@@ -2,149 +2,228 @@
...
-                console.error(`unable to load game ${game_id}`, err);
-              } else {
-                console.log(`load game ${game_id}`);
-                this.games.set(gameToRebuild.id, gameToRebuild);
-                for (const player of gameToRebuild.getPlayers()) {
-                  this.playerToGame.set(player.id, gameToRebuild);
-                  this.onGameLoaded(gameToRebuild.id, player.id);
+      let loaded = 0;
+      allGames.forEach((gameId) => {
+        Database.getInstance().getGame(
```
kberg, line 207: For readability, how about sticking this in a private `loadGameFromDatabase` method?

kberg (commented):
I'll need another pass to internalize most of the rest of this, but this looks like quite a good improvement.

src/database/GameLoader.ts
bafolts: I seem to be a steam punk tech luddite. I work on a raspberry pi 4 model b with an old tiny monitor. This was a long line to me!

server.ts
bafolts: This is similar to an onion, but instead of increasing tears while peeling deeper there is an ever increasing smile and feelings of joy.

server.ts
bafolts: A second argument to `getByGameId` would work here. The argument would be `true` if you wanted to force a new copy from database.

kberg: > A second argument to `getByGameId` would work here. The argument would be `true` if you wanted to force a new copy from database.

Agree

alrusdi:
@kberg do you satisfied with current state of PR? Your change request is blocking it from merge.

bafolts:
I am working on his changes now.

bafolts:
@kberg the heroku server is running out of memory while starting up and trying to load games from memory. Is this good enough to approve?

src/database/GameLoader.ts
```diff
@@ -2,149 +2,223 @@
...
-      } else {
-        this.pendingGame.set(gameId, [cb]);
-      }
+  /**
+   * Determines if game is available in javascript
+   * memory
+   * @param {GameId} gameId id to check
+   * @return {boolean} true iff game is available in memory
+   */
+  public has(gameId: GameId): boolean {
```
Lynesth, line 69: I'd rename this method to `isLoaded()`. It currently makes it pretty confusing with `this.games.has()`.
And then maybe add a `isPreloaded()` one that simply does `this.games.has()` (see comment below for reference)

src/database/GameLoader.ts
```diff
@@ -2,149 +2,223 @@
...
+  }
 
-      Database.getInstance().getGames((err, allGames) => {
-        if (err) {
-          console.error('error loading all games', err);
-          this.onAllGamesLoaded();
-          cb();
-          return;
-        }
+  private loadAllGames(): void {
```
Lynesth, line 183: Could we rename this as well ?
It's not doing the same thing as `loadGame` at all, so probably something like `preloadAllGames()` would be better ?

src/database/GameLoader.ts
```diff
@@ -39,7 +39,7 @@ export class GameLoader {
   private readonly playerIds = new Map<PlayerId, GameId>();
 
   // requests for game that are waiting to load
-  private readonly gameRequests: Array<[GameId, LoadCallback]> = [];
+  private readonly gameRequests: Array<[GameId, boolean, LoadCallback]> = [];
```
kberg, line 42:
```
// requests for games that are waiting to load
// [idToLoad, bypassCache, callback]
```

kberg (approved):
One comment to clarify the boolean. But that's optional. This looks good to go. Good work!

src/database/GameLoader.ts
bafolts: I meant to remove this. Was needed before refactor. Gone now.

PR #2093, opened 2020-12-12
Add LoadCallback type for readability comprehension for new people re…

kberg:
…ading GameLoader.


bafolts (approved):
This is really good! The typescript force is strong in this one!

PR #2099, opened 2020-12-13
Game may no longer have multiple players with the same player ID, and `firstPlayer` must be in the list of players.

kberg:
This only happens in tests, and as you can see by this change, it's *a lot* of tests, which casually use `[player, player]` as their list of players in a game. This causes some super-weird test failures when a player receives both ends of an action. We shouldn't be doing it anyway.

There were surprisingly few tests that didn't like this: MiningGuild and GameLoader. Both were easy enough to fix, ... no actually MiningGuild was a pain in the butt. Mining Guild's issue was that that it used maxOutOceans, which by default stops when it places 9 oceans -- but there are 12 ocean spaces! So I switched to a more direct space filter.

GameLoader just needed dedicated Player IDs. Also a bit of a pain trying to manage the sheer volume of callbacks between GameLoader and tests.

**What do you need, reviewer?**

It's a lot of files. Do you want me to split it out? No problem. Almost all the changes are dead simple to understand (except MiningGuild and GameLoader, but I've discussed that.) Also, do you want me to rename `redPlayer`? I am not terribly fond of `player2` - you tell me.

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -189,7 +189,7 @@ export class Game implements ISerializable<SerializedGame> {
       public gameOptions: GameOptions = {...DEFAULT_GAME_OPTIONS}) {
       {
         const _playerIds = players.map((p) => p.id);
-        if (_playerIds.find((pid) => pid === first.id) === undefined) {
+        if (_playerIds.includes(first.id) === false) {
           throw new Error('Cannot find first player ' + first.id + ' in ' + _playerIds);
         }
         if (new Set(_playerIds).size !== players.length) {
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -190,10 +190,10 @@ export class Game implements ISerializable<SerializedGame> {
       {
         const _playerIds = players.map((p) => p.id);
         if (_playerIds.find((pid) => pid === first.id) === undefined) {
-          throw new Error('Cannot find first player ' + first + ' in ' + _playerIds);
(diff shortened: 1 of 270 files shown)


tests/database/GameLoader.spec.ts
```diff
@@ -71,13 +72,14 @@ describe('GameLoader', function() {
     });
     expect(actual).to.be.undefined;
     loader.start();
+    console.log(actual!.getPlayers());
```
Lynesth, line 75: Is this supposed to stay there ?

src/Game.ts
```diff
@@ -186,8 +186,17 @@ export class Game implements ISerializable<SerializedGame> {
       public id: GameId,
       private players: Array<Player>,
       private first: Player,
-      // ... creates a shallow copy.
       public gameOptions: GameOptions = {...DEFAULT_GAME_OPTIONS}) {
+      {
+        const _playerIds = players.map((p) => p.id);
+        if (_playerIds.find((pid) => pid === first.id) === undefined) {
```
Lynesth, line 192:
```ts
        if (_playerIds.includes(first.id) === false) {
```

Lynesth (approved):

tests/database/GameLoader.spec.ts
kberg: Nope! Sorry. Thanks.

src/Game.ts
kberg: Yesss... thank you.

PR #2100, opened 2020-12-13
Games can't have the same color apply to two different players.

kberg:
Do not review until after reviewing #2099


Lynesth (approved):
LGTM

PR #2107, opened 2020-12-13
Add meaningful tests to IndenturedWorkers and Conscription.

kberg:
This meant having to pull the hard work of passing out of a callback and into a first-class method. No biggie.

### DIFF
--- a/tests/cards/base/IndenturedWorkers.spec.ts
+++ b/tests/cards/base/IndenturedWorkers.spec.ts
@@ -1,18 +1,44 @@
 import {expect} from 'chai';
 import {IndenturedWorkers} from '../../../src/cards/base/IndenturedWorkers';
+import {MicroMills} from '../../../src/cards/base/MicroMills';
 import {Game} from '../../../src/Game';
+import {Player} from '../../../src/Player';
 import {TestPlayers} from '../../TestingUtils';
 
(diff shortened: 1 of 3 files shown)


tests/cards/base/IndenturedWorkers.spec.ts
```diff
@@ -1,18 +1,44 @@
...
     player.victoryPointsBreakdown.setVictoryPoints('victoryPoints', card.getVictoryPoints());
     expect(player.victoryPointsBreakdown.victoryPoints).to.eq(-1);
     expect(card.getCardDiscount(player, game)).to.eq(0);
   });
+
+  it('Should apply card discount until next card played', () => {
+    player.playCard(game, card);
+    expect(card.getCardDiscount(player, game)).to.eq(8);
+
+    player.playCard(game, new MicroMills());
```
chosta, line 31: Maybe we should also add a card of 8+ cost to check for the correct discount (cost - 8)?

chosta (approved):

tests/cards/base/IndenturedWorkers.spec.ts
kberg: A test for calculating card discounts is probably sufficient. I'm going to merge this as it's wholly better and I've moved on to a serialization test.

PR #2111, opened 2020-12-14
Remove unused field Player.generationPlayed.

kberg:
This seems like an acceptable change. I cannot find a single use for it.


Lynesth:
Not used indeed.
From https://github.com/bafolts/terraforming-mars/pull/324 : ` Switch generationPlayed to private, not used anymore but could be used for statistics in the future`

kberg:
OK, let's nix it.

>

Lynesth (approved):

PR #2115, opened 2020-12-14
All the P

chosta:
Note that I've also removed the HTML_DATA for N, O to avoid conflicts in merging. I'll make sure N + O comes before P when merging, so don't worry about blank cards.

### DIFF
--- a/src/cards/base/Pets.ts
+++ b/src/cards/base/Pets.ts
@@ -7,6 +7,10 @@ import {ResourceType} from '../../ResourceType';
 import {CardName} from '../../CardName';
 import {IResourceCard} from '../ICard';
 import {Board} from '../../Board';
+import {CardMetadata} from '../CardMetadata';
+import {CardRenderer} from '../render/CardRenderer';
+import {CardRenderDynamicVictoryPoints} from '../render/CardRenderDynamicVictoryPoints';
+import {CardRenderItemSize} from '../render/CardRenderItemSize';
 
 export class Pets implements IProjectCard, IResourceCard {
     public cost = 10;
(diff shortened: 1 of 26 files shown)


src/cards/base/Pets.ts
```diff
@@ -28,4 +32,18 @@ export class Pets implements IProjectCard, IResourceCard {
...
     }
+    public metadata: CardMetadata = {
+      cardNumber: '172',
+      renderData: CardRenderer.builder((b) => {
+        b.effectBox((eb) => {
+          eb.city(CardRenderItemSize.SMALL).any.startEffect.animals(1);
+          eb.description('Effect: When any City tile is placed, add an Animal to this card.');
+        }).br;
+        b.animals(1).br;
+        b.text('Animals may not be removed from this card', CardRenderItemSize.SMALL, true).br;
```
kberg, line 43: Isn't this meant to be capitalized?

src/cards/base/PowerGrid.ts
```diff
@@ -16,4 +17,12 @@ export class PowerGrid implements IProjectCard {
       player.addProduction(Resources.ENERGY, 1 + player.getTagCount(Tags.ENERGY));
       return undefined;
     }
+
+    public metadata: CardMetadata = {
+      cardNumber: '102',
+      renderData: CardRenderer.builder((b) => {
+        b.productionBox((pb) => pb.energy(1).slash().energy(1).played);
```
kberg, line 24: Another case of resource vs tag, and here, the second energy actually ought to be Power. Your thoughts?

src/cards/turmoil/ParliamentHall.ts
```diff
@@ -29,4 +32,16 @@ export class ParliamentHall implements IProjectCard {
...
+
+    public metadata: CardMetadata = {
+      cardNumber: 'T08',
+      requirements: CardRequirements.builder((b) => b.party(PartyName.MARS)),
+      renderData: CardRenderer.builder((b) => {
+        b.productionBox((pb) => {
+          pb.megacredits(1).slash().building(3).played;
+        });
+      }),
+      description: 'Requires that Mars First are ruling or that you have 2 delegates there. Increase your MC production 1 step for every 3 Building tag you have, including this.',
```
kberg, line 44: "every 3 Building tag" should be "tags".

kberg (approved):
Looks great.

src/cards/base/Pets.ts
chosta: It is, by the `true` parameter at the end. I wish TS had something like python's parameter naming
`text(isUppercase=True)`
In ES6 you can do `text({isUppercase=True})` but it doesn't appeal to me.

src/cards/base/PowerGrid.ts
chosta: It's on my refactoring list. Certainly better semantics could be used in these cases. 
E.g.
CardRequirements.builder((b) => b.tag(Tags.VENUS).tag(Tags.JOVIAN).tag(Tags.EARTH))
can become:
`CardRequirements.builder((b) => b.unity())` [From turmoil, unity is the bonus for jov, earth, venus tags] its conceptually appropriate and it is repeated multiple times.

Issue #2109, opened 2020-12-13
Bug: Games at setup don't survive server sleep on a DB connected server


lomky:
Server Setup: 
Heroku app with postgres dyno spun up & connected via env var

Steps to recreate:
 - Setup and start a multiplayer game through player link creation, but no further
 - Allow the server to idle and go to sleep
 - Attempt to access either the `game` or `player` link
 - Receive `Unexpected server response`

![image](https://user-images.githubusercontent.com/6129479/102018501-8d6c9380-3d3b-11eb-979d-7835f6e555f4.png) |

I confirmed in the database that the game has been saved
![image](https://user-images.githubusercontent.com/6129479/102018529-be4cc880-3d3b-11eb-8cfc-1cd368b4029a.png)

I set up test games at various other points (just finished corp & card selection, first turns all taken) - they do not encounter this issue. They also have multiple save entries in the database, as you would expect.

bafolts:
I will label this as an enhancement only because I believe this was the original intentions. Since we have the game in the database though there is no reason we can't salvage it in this situation.

https://github.com/bafolts/terraforming-mars/blob/main/src/database/PostgreSQL.ts#L70

If I remember this was done because of how the code happened to work probably while de-serializing a game from the database. It makes sense to me to get this to work though.

lomky:
Noted! For now we'll just make sure to finish setup before the server goes idle.

For understanding the actual use case, my friend is playing with another friend in Europe, so they are playing (very) asynchronously.

---

Also, I confirmed that this happens at any point during the setup turns:
  - Setup and start a multiplayer game through player link creation
  - One player finishes their setup
  - Allow the server to idle and go to sleep
  - Attempt to access either the game or player link
  - Receive Unexpected server response

In this situation, no additional database save was created when the one player did their setup.

Issue #2098, opened 2020-12-13
Feature/System: Save game state during the initial draft


vsrisuknimit:
Currently if a server rolls back/hiccup/crash during the initial draft, the game is lost, because the game state is not saved until the end of the draft.

It would be good to save the game state during the initial draft. Once after the first pack. A second time after the second pack.

kberg:
This seems like a Good Idea.

PR #2116, opened 2020-12-14
Save/Reload games during draft or research phases

Lynesth:
This properly saves the game each time all players have drafted 1 card and passed the others to their neighbour.
So in case of the server going awol it will go back to the point it was last save in the draft.

It also saves when the draft is done (or if playing without the draft variant) while everyone is deciding which cards to buy (aka research phase). If server goes down, it'll go back to the beginning of the research phase, for all players.

Again, this will add a few saves per game - around 50 rows for a 4p game with all draft variants - but it seems negligible.

Quick summary of what the code consists of:
- A new method, `game.save()`, which saves the game \o/
- A refactor of `playerIsFinishedWithDraftingPhase` because it was a mess to understand. Still not 100% happy with it but I think it's better.
- A removal of `isLastActiveRoundOfDraft` (linked to the refactor above) and instead use the numbers of cards left to decide if it was the last round.
- A small reordering of things in `getDraftCardsFrom`, nothing important.
- A bunch of `this.save()` added here and there to save the game at appropriate times.
- A better handling of what to do when the game is loaded from json, so that it can restart during draft or research phase (the whole point of this PR).

-----

This was obviously ~~troughly~~ ~~throughouly~~ ~~thoroug...~~ very well tested by yours truly.


Lynesth:
Closes https://github.com/bafolts/terraforming-mars/issues/2098
Fixes https://github.com/bafolts/terraforming-mars/issues/2109

src/Game.ts
```diff
@@ -319,22 +320,11 @@ export class Game implements ISerializable<SerializedGame> {
...
-          }
         } else {
-          this.setStartingProductions(player);
           this.playerHasPickedCorporationCard(player, new BeginnerCorporation());
         }
       }
 
-      // Save initial game state
-      Database.getInstance().saveGameState(this.id, this.lastSaveId, this.toJSON(), this.players.length);
-      this.lastSaveId = 1;
```
Lynesth, line 336: Game is now saved when entering either `runDraftRound` or `gotoInitialResearchPhase` a few lines below this.

bafolts (commented):
A quick read through this looks right. As this is dealing with some of the lower level stuff again I am going to take time to ensure we aren't missing something. Will look again tomorrow.

bafolts (approved):

alrusdi:
@Lynesth feel free to merge it to main!

lomky:
Much appreciated! :rocket:

Issue #2134, opened 2020-12-17
Solar farm iconography


sienmich:
Currently looks like this:
![Screenshot from 2020-12-17 15-51-14](https://user-images.githubusercontent.com/19472604/102503969-8f5a8d80-4080-11eb-87bd-a417964039df.png)
It's bad, because it's about plants resources not tags.
But how should it look like?

1. ![Screenshot from 2020-12-17 15-50-43](https://user-images.githubusercontent.com/19472604/102503971-8f5a8d80-4080-11eb-8ca9-0807e0e3d502.png)
2. ![Screenshot from 2020-12-17 15-53-39](https://user-images.githubusercontent.com/19472604/102503967-8ec1f700-4080-11eb-9584-e9128d9c3f2f.png)
3. ![Screenshot from 2020-12-17 15-56-28](https://user-images.githubusercontent.com/19472604/102503964-8e296080-4080-11eb-8738-41927e40b405.png)

The 1. Looks too much like it involves plants production imo. 2 is great. 3. is kinda clever, nah?
Which one should it be? I can make PR.
Summon @chosta

chosta:
You can haz PR :)
I think @Lynesth also had a comment about the Solar farm Tile icon and the way it represented energy.
2 is just fine
3 might be confusing

kberg:
Preference order 1, 2, 3. I defer to Dani.

>

kevinb9n:
It would be nice to try sticking with TM's (kinda weird) "greedily expanding production boxes" convention (and thus not 2 or 3), but part of how it works is that being surrounded doesn't change the meaning of almost any symbol in the game *except* the six standard resources and the unspecified-resource icon. So I think 1 doesn't work.

If you want a 4... sure, it's still a kinda inexplicable, but so much in the game is.

<img width=300 src="https://user-images.githubusercontent.com/934551/102511470-ac21a180-403d-11eb-9085-ecbe07943b35.png" alt="another option">

chosta:
@sienmich we can try @kevinb9n's suggestion
`emptyTile().secondaryTag(Tags.PLANTS) maybe?`
let's see how it looks like

SliceOfBread:
I think it should look like this as it is power production per plant on empty tile (not empty tile production).
![powerperplant](https://user-images.githubusercontent.com/13003705/102528704-e48ca400-406c-11eb-99f3-57f020ea1476.png)

kevinb9n:
[...]

PR #2136, opened 2020-12-17
Updated SolarFarm.ts

sienmich:
Option 2. from https://github.com/bafolts/terraforming-mars/issues/2134


chosta:
please check my comment in #2134

chosta (changes requested):
Sorry to go back and forth on this but I went out and tried out all the options myself and now feel that the option should be the best in this case. We have an asterix that makes the player read the text and decide what to do.
please, go for 
`pb.energy(1).slash().plants(1)`

sienmich:
No problem :)
Done

chosta (approved):

PR #2140, opened 2020-12-18
Restore Natural Preserve card iconography.

kberg:
image

### DIFF
--- a/src/components/card/CardRenderTileComponent.ts
+++ b/src/components/card/CardRenderTileComponent.ts
@@ -70,7 +70,9 @@ export const CardRenderTileComponent = Vue.component('CardRenderTileComponent', 
       } else if (type === TileType.NATURAL_PRESERVE) {
         if (this.item.isAres) {
           classes.push('card-tile-ares');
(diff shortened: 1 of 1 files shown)


src/components/card/CardRenderTileComponent.ts
```diff
@@ -70,7 +70,9 @@ export const CardRenderTileComponent = Vue.component('CardRenderTileComponent',
       } else if (type === TileType.NATURAL_PRESERVE) {
         if (this.item.isAres) {
           classes.push('card-tile-ares');
-          classes.push('board-space-tile--natural_preserve');
+          classes.push('board-space-tile--natural_preserve_ares');
+        // } else {
```
alrusdi, line 74: Do we really need these comment?

alrusdi (approved):

src/components/card/CardRenderTileComponent.ts
kberg: Huh, this could actually be an error.

kberg: Yep, deleting that was the thing.

PR #2141, opened 2020-12-18
Add custom type NeutralPlayer

kberg:
I'm not sure this is worthwhile, someone else tell me if I'm wrong -- but the overzealous use of the literal `'NEUTRAL'` seems a bit of an invitation for typos.

I tried making a `const NEUTRAL = 'NEUTRAL'` but that causes all other types of problems. This may be the best we can do, in which case perhaps it's not even worthwhile. Let me know.

### DIFF
--- a/src/cards/turmoil/BannedDelegate.ts
+++ b/src/cards/turmoil/BannedDelegate.ts
@@ -1,7 +1,7 @@
 import {IProjectCard} from '../IProjectCard';
 import {CardName} from '../../CardName';
 import {CardType} from '../CardType';
-import {Player} from '../../Player';
+import {Player, PlayerId} from '../../Player';
 import {Game} from '../../Game';
 import {OrOptions} from '../../inputs/OrOptions';
 import {SelectDelegate} from '../../inputs/SelectDelegate';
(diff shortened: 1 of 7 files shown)


Lynesth:
To be honest, I'm not found of PlayerIdOrNeutral. I think we could either:

- Drop the `'NEUTRAL'` type altogether and everything would still work since PlayerId is a string. But I get that it gives some kind of info as to what to expect from the argument.
- Create an `export type NeutralPlayer = 'NEUTRAL';` (still in Turmoil.ts) and use the type `NeutralPlayer` instead of `'NEUTRAL'` type everywhere possible. It should also allow replacing it in places where `'NEUTRAL'` is used in conjunction with other types, like `Player`.

kberg:
> To be honest, I'm not found of PlayerIdOrNeutral.

Me either.

I prefere your second option, let me give it a shot.

bafolts:
This will be good change one way or another.

kberg:
PTAL, I updated according to Lynesth's suggestions and updated the Pr description accordingly.

chosta (approved):
Since you've already discussed the approach, I checked and didn't find any issues.

PR #2143, opened 2020-12-18
Restore flaky test. Let's see why it fails in CI.

kberg:
I've run this a few times, and it passes. I could run it a hundred times, but this will be much quicker.


bafolts (approved):

kberg:
NO IT FAILED ALREADY!

kberg:
Comment is that it has to do with when the global event prefers a given political party. I'm addressing this by clearing all delegates before the test starts. Then testing flakiness with 

``` (for i in {1..100}; do npx mocha --recursive build/tests/Turmoil.spec.js ; done) > /tmp/out.txt ```

Without fix, 95% pass rate. With fix, 100% pass rate.

PR #2144, opened 2020-12-18
Part of of implementing chairperson agenda selection

kberg:
It removes the synchronous nature of agenda selection occurring with the generation change. Doesn't work fully yet, so don't use it!


kberg:
Wow, this turned into a bit of a complexity explosion, but  I promise it's better.

PR #2147, opened 2020-12-19
Move board-related classes to src/boards and tests/boards.

kberg:


kberg:
As I start to look at The Moon and how to deal with what is effectively a second board, I thought I'd start by reducing the vast number of classes in the src/ root.

bafolts (approved):

PR #2160, opened 2020-12-22
Switch ISpace.player from Player to Playerid.

kberg:
By example, game storage for a game with 40 tiles dropped from 1.96MiB to 0.32 Mib.
This adds a little work in code, but surprisingly not very often at all.

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -446,7 +446,7 @@ export class Game implements ISerializable<SerializedGame> {
     }
 
     // Function to retrieve a player by it's id
-    public getPlayerById(id: string): Player {
+    public getPlayerById(id: PlayerId): Player {
       const player = this.players.find((p) => p.id === id);
       if (player === undefined) {
         throw new Error(`player ${id} does not exist on game ${this.id}`);
@@ -455,7 +455,7 @@ export class Game implements ISerializable<SerializedGame> {
     }
 
     // Function to return an array of players from an array of player ids
-    public getPlayersById(ids: Array<string>): Array<Player> {
+    public getPlayersById(ids: Array<PlayerId>): Array<Player> {
       return ids.map((id) => this.getPlayerById(id));
     }
 
@@ -1320,8 +1320,7 @@ export class Game implements ISerializable<SerializedGame> {
       return this.board.spaces.filter(
         (space) => space.tile !== undefined &&
                    space.tile.tileType === tileType &&
-                   space.player !== undefined &&
-                   space.player === player,
+                   space.player === player.id,
       ).length;
     }
     public addTile(
@@ -1334,8 +1333,8 @@ export class Game implements ISerializable<SerializedGame> {
       }
 
       // Land claim a player can claim land for themselves
-      if (space.player !== undefined && space.player !== player) {
-        throw new Error('This space is land claimed by ' + space.player.name);
+      if (space.player !== undefined && space.player !== player.id) {
+        throw new Error('This space is land claimed by ' + this.getPlayerById(space.player)?.name);
       }
 
       if (space.spaceType !== spaceType) {
@@ -1374,7 +1373,7 @@ export class Game implements ISerializable<SerializedGame> {
 
 
       // Part 3. Setup for bonuses
-      const arcadianCommunityBonus = space.player === player && player.isCorporation(CardName.ARCADIAN_COMMUNITIES);
+      const arcadianCommunityBonus = space.player === player.id && player.isCorporation(CardName.ARCADIAN_COMMUNITIES);
       let startingResources: Multiset<Resources | ResourceType> | undefined = undefined;
       AresHandler.ifAres(this, () => {
         startingResources = AresHandler.beforeTilePlacement(player);
@@ -1384,7 +1383,7 @@ export class Game implements ISerializable<SerializedGame> {
 
       // Part 4. Place the tile
       space.tile = tile;
-      space.player = player;
+      space.player = player.id;
       LogHelper.logTilePlacement(this, player, space, tile.tileType);
 
       // Part 5. Collect the bonuses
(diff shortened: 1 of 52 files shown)


src/Game.ts
```diff
@@ -1720,25 +1719,29 @@ export class Game implements ISerializable<SerializedGame> {
...
+        }
+      };
+
       d.board.spaces.forEach((element: ISpace) => {
         const space = this.getSpace(element.id);
         if (element.tile) {
           const tileType = element.tile.tileType;
           const tileCard = element.tile.card;
           const protectedHazard = element.tile.protectedHazard;
-          if (element.player) {
```
kberg, line 1729: In reading this code I see this block is unnecessary if the block below is changed from an else block to a full-standing if block.

src/awards/DesertSettler.ts
```diff
@@ -7,8 +7,7 @@ export class DesertSettler implements IAward {
     public description: string = 'Most tiles south of the equator (the four bottom rows)'
     public getScore(player: Player, game: Game): number {
       return game.board.spaces
-        .filter((space) => space.player !== undefined &&
```
kberg, line 10: In these changes you will find cases where I remove tests of `if (space.player !== undefined)`. These become unnecessary when `space.player` is now either the player id -- the expression that replaces it takes care of the whole thing.

But keep an eye out in case I got one of them wrong.

src/cards/base/MiningArea.ts
```diff
@@ -11,7 +11,9 @@ export class MiningArea extends MiningCard {
     public name = CardName.MINING_AREA;
     protected getAvailableSpaces(player: Player, game: Game) {
       return super.getAvailableSpaces(player, game)
-        .filter((space) => game.board.getAdjacentSpaces(space).filter((adjacentSpace) => adjacentSpace.tile !== undefined && adjacentSpace.tile.tileType !== TileType.OCEAN && adjacentSpace.player === player).length > 0);
+        .filter((space) =>
```
kberg, line 14: Sorry reviewer, this was just too much on one-line.

src/cards/promo/Philares.ts
```diff
@@ -88,15 +88,21 @@ export class Philares implements CorporationCard {
     }
 
     public onTilePlaced(player: Player, space: ISpace, game: Game): void {
+      function isPhilares(playerId: PlayerId | undefined): boolean {
```
kberg, line 91: This is going to demonstrate a little of the CPU ineffeciency of my change -- having to look up a player's corpration by ID all the time. Fortunately player lookup by Id is not that big -- I mean it's searching 5-6 elements at a time.

kberg:
**hello reviewer**

I can see this going in two directions:
1. Abandon this particular change in favor of keeping Player on ISpace, and creating a mechanism for serializing and deserializing an ISpace by PlayerId. I could do that, but it's a bit more w.o.r.k -- specifically I'd want to extract the entire board serialization into its own thing. That might be better, let me know.
2. Renaming ISpace.player to ISpace.playerId

Either way, this needs a test. Your thoughts?

src/boards/Board.ts
```diff
@@ -179,7 +179,7 @@ export abstract class Board {
       player: Player | undefined = undefined,
       predicate: (value: ISpace) => boolean = (_x) => true) {
       const spaces = this.spaces.filter((space) => {
-        return this.canPlaceTile(space) && (space.player === undefined || space.player === player);
+        return this.canPlaceTile(space) && (space.player === undefined || space.player === player?.id);
```
Lynesth, line 182:
```ts
        return this.canPlaceTile(space) && (
                 space.player === undefined || (player !== undefined && space.player === player.id)
               );
```

src/cards/base/MiningArea.ts
```diff
@@ -11,7 +11,9 @@ export class MiningArea extends MiningCard {
...
+          game.board.getAdjacentSpaces(space).filter((adjacentSpace) =>
+            adjacentSpace.tile !== undefined && adjacentSpace.tile.tileType !== TileType.OCEAN && adjacentSpace.player === player.id).length > 0);
```
Lynesth, line 16:
```ts
          game.board.getAdjacentSpaces(space).some((adjacentSpace) =>
            adjacentSpace.tile !== undefined && adjacentSpace.tile.tileType !== TileType.OCEAN && adjacentSpace.player === player.id));
```

src/cards/promo/Philares.ts
Lynesth: It's called a few times in the method though. Here's an idea to call it only once.
```ts
    public onTilePlaced(player: Player, space: ISpace, game: Game): void {
      // Using `-1` so that it doesn't match with `undefined`
      const philaresPlayerId = game.getPlayers().find((player) => player.isCorporation(CardName.PHILARES).id || -1;

      if (space.tile !== undefined && space.tile.tileType !== TileType.OCEAN) {
        let bonusResource: number = 0;
        if (space.player === philaresPlayerId) {
          bonusResource = game.board.getAdjacentSpaces(space)
            .filter((space) => space.tile !== undefined && space.player !== undefined && space.player !== player.id)
            .length;
        } else if (space.player !== undefined) {
          bonusResource = game.board.getAdjacentSpaces(space)
            .filter((space) => space.tile !== undefined && space.player === philaresPlayerId)
            .length;
        }
        // ...
```

To make it even more efficient, we could add a property to that card that saves the Philares player id once and for all since it's not gonna change but it doesn't seem necessary.

src/Game.ts
```diff
@@ -1334,8 +1333,8 @@ export class Game implements ISerializable<SerializedGame> {
       }
 
       // Land claim a player can claim land for themselves
-      if (space.player !== undefined && space.player !== player) {
-        throw new Error('This space is land claimed by ' + space.player.name);
+      if (space.player !== undefined && space.player !== player.id) {
+        throw new Error('This space is land claimed by ' + this.getPlayerById(space.player)?.name);
```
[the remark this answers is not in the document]
kberg: I forget.

kberg:
So good news: this has been entirely rewritten! If you still think removing Player as a first-class object in ISpace is a good idea, let me know.

kberg:
@Lynesth I'm closing this PR. If you still think the Player to ID is worth changing permanently, let's chat separately.

PR #2163, opened 2020-12-22
Flaky timer test

bafolts:
Noticed today a new unit test for `Timer` was taking a second to run and relying on `setTimeout` to take a second. I had this test fail on me. The `setTimeout` function can not guarantee that the callback will be called at exactly 1000ms. When it is called too late this unit tests fails. I modified the unit test to manually provide a time and run without using `setTimeout`.


kberg (approved):
CLEVER.

PR #2164, opened 2020-12-22
Only store CardName with Dealer.

bafolts:
We create `IProjectCard` when making the decks for the dealer. The dealer doesn't need a complete `IProjectCard`. A game should never exhaust the deck so this should lead to less overall memory used by the server. Right now when a game is created we complete an instance of `IProjectCard` for every card in the deck which end up sitting in javascript memory.

### DIFF
--- a/src/SerializedDealer.ts
+++ b/src/SerializedDealer.ts
@@ -1,10 +1,8 @@
 import {CardName} from './CardName';
-import {CorporationCard} from './cards/corporation/CorporationCard';
-import {IProjectCard} from './cards/IProjectCard';
 
 export interface SerializedDealer {
-    corporationCards: Array<CorporationCard | CardName>;
-    deck: Array<IProjectCard | CardName>;
-    discarded: Array<IProjectCard | CardName>;
-    preludeDeck: Array<IProjectCard | CardName>;
+    corporationCards: Array<CardName>;
+    deck: Array<CardName>;
+    discarded: Array<CardName>;
+    preludeDeck: Array<CardName>;
 }
--- a/tests/cards/base/SearchForLife.spec.ts
+++ b/tests/cards/base/SearchForLife.spec.ts
@@ -39,14 +39,14 @@ describe('SearchForLife', function() {
   it('Should act', function() {
     player.playedCards.push(card);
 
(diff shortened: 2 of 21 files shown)


bafolts:
@kberg I still need to finish up some of the unit tests. This shouldn't change much from here so probably good time to leave feedback if you see any concerns.

src/SerializedDealer.ts
```diff
@@ -1,10 +1,8 @@
 import {CardName} from './CardName';
-import {CorporationCard} from './cards/corporation/CorporationCard';
-import {IProjectCard} from './cards/IProjectCard';
 
 export interface SerializedDealer {
-    corporationCards: Array<CorporationCard | CardName>;
```
bafolts, line 6: Enough time has passed this can be changed to the simple format and doesn't need to expect both types anymore.

tests/cards/base/SearchForLife.spec.ts
```diff
@@ -39,14 +39,14 @@ describe('SearchForLife', function() {
...
     player.playedCards.push(card);
 
-    while (game.dealer.discarded.find((c) => c.tags.length === 1 && c.tags[0] === Tags.MICROBES) === undefined ||
-               game.dealer.discarded.find((c) => c.tags.length === 1 && c.tags[0] !== Tags.MICROBES) === undefined) {
-      player.megaCredits = 1;
-      card.action(player, game);
-      game.deferredActions.runNext();
-      expect(player.megaCredits).to.eq(0);
-    }
+    //    while (game.dealer.discarded.find((c) => c.tags.length === 1 && c.tags[0] === Tags.MICROBES) === undefined ||
```
bafolts, line 42: Going to update to put cards on the deck manually so we know which get drawn instead of relying on discarded having `tags`.

kberg:
> a game should never exhaust a deck

Oh I've done that on at least three or four occasions. Maybe with all the expansions, but yeah.

bafolts:
By exhaust I mean every project card was played leaving no cards left to draw from the deck and no cards in the discard pile. The case where you run out of cards and reshuffle should be common. As long as the card doesn't end up in a players hand with this change it shouldn't take up memory.

bafolts:
May re-visit later. A garbage collection process that removes games that are no longer being used should be less impactful than this change. Attempting not to early optimize and make it impossible to work with `Dealer` in quest to spare memory.

PR #2161, opened 2020-12-22
Reduce spacing in Game, Player and Board.

kberg:
Sometimes minor refactors cause the compiler to dislike the 4-character
indent for the first indentation.

PR #2089, opened 2020-12-12
add limited random board option

Ronnoc:
https://github.com/bafolts/terraforming-mars/issues/2048

Try it on 

http://192.144.184.221:8080/

|  board   | Exmaple  | Exmaple  | Exmaple  |
|  ----  | ----  | ----  | ----  |
| before  | bef1 | bef2 | bef3 |
| Origin  |org1 | org2 | org3 |
| Hellas  | hel1 | hel2 | hel3 |
| Ely  | ely1 | ely2 | ely3 |

PR #2167, opened 2020-12-22
Rewrite board serialization

kberg:
Depends on #2161 and invalidates #2160 

This is a lot of commits one after the other. Best to review each commit individually.
Test is pending. Setting do-not-merge until I write a test.


kberg:
Test added, this is ready for review.

src/boards/Board.ts
```diff
@@ -202,7 +170,7 @@ export abstract class Board {
   }
 
   public canPlaceTile(space: ISpace): boolean {
-    return space !== undefined && space.tile === undefined && space instanceof Land;
```
bafolts, line 205: This is a good change to not check for `undefined` when the type for `space` does not indicate it can be `undefined` 👍 . Not relying on `instanceof` also always better as `instanceof` is a javascript operator and not typescript enforced.

kberg: Thanks. And it enabled me to remove those classes - which was a bit of a pleasant surprise.

src/boards/Board.ts
```diff
@@ -222,4 +190,69 @@ export abstract class Board {
...
+    return {
+      spaces: this.spaces.map((space) => {
+        return {
+          id: space.id,
+          spaceType: space.spaceType,
+          tile: space.tile,
+          player: space.player?.id,
+          bonus: space.bonus,
+          adjacency: space.adjacency,
+          x: space.x,
```
bafolts, line 204: There was a change made in #2089 which I think is helpful as it saves space. It assumes each board will have the same number of tiles in each row and then works through the array in that way knowing `x` and `y` based on the row index. This is helpful as it avoids needing to store `x` and `y` in javascript memory and also while serializing. In another issue we should remove this `x` and `y` as they are unnecessary and taking up space.

src/boards/BoardBuilder.ts
```diff
@@ -101,9 +110,33 @@ export class BoardBuilder {
...
+        return Space.ocean(idx, pos_x, pos_y, bonus);
       } else {
-        return new Land(idx, pos_x, pos_y, bonus);
+        return Space.land(idx, pos_x, pos_y, bonus);
       }
     }
 }
+
+class Space implements ISpace {
+  constructor(public id: string, public spaceType: SpaceType, public bonus: Array<SpaceBonus>, public x: number, public y: number ) {
```
bafolts, line 121: Should this become the `private constructor` as it is intended these are made through the `static` functions?

src/boards/Board.ts
kberg: You know, I would agree with you, but I suggest keeping these. First, they make Desert Settler (and others) much easier to deal with.

But also it's going to make The Moon harder to work with, as it's going to use Board, with [3,4,5,4,3.] (Not to mention two more colonies! :D) There's alo a Mercury map and then super giga map. :D

src/boards/BoardBuilder.ts
kberg: Good catch, fixed.

src/boards/Board.ts
bafolts: The `x` and `y` could become calculated properties on space that are calculated based on their position within the ordered `spaces` array. No code should get hurt then but space should be saved. The added CPU overhead should be minimal.

src/boards/Board.ts
```diff
@@ -222,4 +190,69 @@ export abstract class Board {
...
+      player: player,
+      bonus: space.bonus,
+      adjacency: space.adjacency,
+      x: space.x,
+      y: space.y,
+    } as ISpace;
+  }
+
+  /*
+    d.board.spaces.forEach((element: ISpace) => {
```
bafolts, line 230: Remove before flight.

src/boards/Board.ts
kberg: Agreed, but I'm just asking to wait on removing those until I get a sense of how Moon is going to work.

src/boards/SerializedBoard.ts
```diff
@@ -0,0 +1,21 @@
+import {IAdjacencyBonus} from '../ares/IAdjacencyBonus';
+import {ITile} from '../ITile';
+import {Player, PlayerId} from '../Player';
+import {SpaceBonus} from '../SpaceBonus';
+import {SpaceType} from '../SpaceType';
+
+export interface SerializedBoard {
+    spaces: Array<SerializedSpace>;
```
bafolts, line 8: Should be two spaces.

bafolts (approved):
Had two minor things to remove before merging.

src/boards/Board.ts
bafolts: Yes, this PR didn't add the `x` and `y`. Those values do make most operations simpler. At some point the quest to lower server memory would impact readability and then start to impact CPU. With an added garbage collection process on `GameLoader` it shouldn't really matter. If someone starts a single player game and finishes it the garbage collector could remove it from javascript memory when it ran next.

PR #2170, opened 2020-12-23
Add warning property to cards

Lynesth:
_Disclaimer: I don't like Vue. This is working and doesn't seem_ too much _but please @chosta or @alrusdi, take a look and see if there's a better way to take care of the frontend part._

This is the first part of the implementation to improve Reds taxes calculation which will then lead to a full cards rewrite (hopefully).
This PR just adds a new property called `warning` to ICard that can be displayed below the card when selected.

The goal is to assign some meaningful message to that property during a card's `canAct` or `canPlay` to give some info to the user. No card is using it at the moment. It looks like this (don't pay attention to the text itself, it was just to test it):

firefox_2020-12-23_14-05-45

It is related to #1642.


kberg:
As far as I'm concerned this is a net improvement, and so ought to go in. It can most definitely be improved.

kberg:
That said, there's a discussion about this topic -- and there I raised my concern about a slippery slope. But this is good.

Lynesth:
> I raised my concern about a slippery slope.

That's why this PR only gives the tool but doesn't use it yet.
It will be up to us to approve the uses we made of it to prevent sliding down that slope ;)

src/cards/ICard.ts
```diff
@@ -46,4 +46,5 @@ export interface ICard {
     cost?: number;
     cardType: CardType;
     metadata?: CardMetadata;
+    warning?: string;
```
alrusdi, line 49: This warning must implement Message interface to be translatable to other-than-english languages

src/components/SelectHowToPayForCard.ts
```diff
@@ -372,6 +387,7 @@ export const SelectHowToPayForCard = Vue.component('select-how-to-pay-for-card',
   </label>
 
   <section v-trim-whitespace>
+    <div v-if="hasCardWarning()" class="card-warning">{{ cardWarning }}</div>
```
alrusdi: Please add v-i18n property

alrusdi (changes requested):

alrusdi:
@Lynesth what exact points against Vue?

Lynesth:
> @Lynesth what exact points against Vue?

I guess simply the fact that I don't know it and usually spend way too much times doing what sounds like simple things to do.

src/cards/ICard.ts
kberg: Thanks, I was just coming here to make that comment.

bafolts (changes requested):
I would send the warning through `PlayerInputModel` with the `warning` property and not associate it with one card. It makes sense that there would be a warning regarding input from a player but not a warning regarding one card. The warning only makes sense through the context of player input. The error we throw today "Not enough cards selected" should also use this `PlayerInputModel` warning property.

kevinb9n:
Two of us in the family play a LOT, and I asked the other today if there was one thing she wished would change the most, and this happened to be the one she brought up.

I think you can hold out against the slippery slope IF you have a clear criterion that you try to converge toward. Something like: "there is at least one instruction on this card that will provably cause NO change to the game state". Not "a state change that might not matter", etc.

Lynesth:
@bafolts
Since these warnings are card dependent, how exactly do you want this to be done, I don't get it.
They are printed before you validated the input and they don't prevent you from validating the input.

src/cards/ICard.ts
Lynesth: Did it and used `$t` to translate it, like we do for the title.

alrusdi (approved):
Now it is seems to be correct to me.

src/styles/cards.less
```diff
@@ -43,6 +43,12 @@
     margin-bottom: 15px;
 }
 
+.card-warning {
+    font-size: 18px;
+    color: #d73a49;
```
chosta, line 48: Can you, please, add a variable in variables.less called `@color_error` and use it as a variable in the class .card-warning. I'll probably do a PR soon to add some more such variables and refactor the existing by using variables instead of hardcoded values.

PR #2172, opened 2020-12-23
Make Game constructor private, and game deserialization doesn't rely …

kberg:
…on full game set-up.
This is DEFINITELY still broken, and a work-in-progress.

I haven't done a single bit of testing, other than to verify that existing tests pass. However,
1. Cloning is fully disabled.
2. `deserialize` doesn't restore any of the fields that would have been set with Object.assign. (I could just restore that field for now.)
3. I need feedback on the details of the database operations, including rewriting how undo is managed. (`Player` shouldn't be doing squat with the database.)


kberg:
I can see now that this PR is probalby still too big. I'm going to make a change that ONLY deals with the tests as an interim.

kberg:
(if you have any suggestions for splitting th is out., please tell me.)

kberg:
In fact, #2173 will fix all the tests at once. Merge that first.

src/Game.ts
```diff
@@ -454,87 +447,84 @@ export class Game implements ISerializable<SerializedGame> {
...
-
-      // Initial Draft
-      if (game.gameOptions.initialDraftVariant) {
-        game.runDraftRound(true);
-      } else {
-        game.gotoInitialResearchPhase();
-      }
-    });
-  }
+  // private cloneGame(gameId: GameId): void {
```
bafolts, line 450: I would remove this implementation for the time being. The clone game operation needs to generate a new game id and new player ids and then update any places which may have a reference to a player id for the new player id. Most of this function will probably change when re-implemented anyway.

src/Game.ts
```diff
@@ -1633,166 +1623,125 @@ export class Game implements ISerializable<SerializedGame> {
...
-    return result;
-  }
-
-  // Function used to rebuild each objects
-  public loadFromJSON(d: SerializedGame): Game {
-    // Assign each attributes
-    const o = Object.assign(this, d);
-    const cardFinder = new CardFinder();
+  public static deserialize(d: SerializedGame): Game {
+    // const o = Object.assign(this, d);
```
bafolts, line 1627: Do the honors. Remove this line altogether. Save us all from headaches and strange bugs.

kberg: Not yet -- I haven't fixed it yet.

src/Game.ts
kberg: Is deleting it like this the right approach?

src/GameSetup.ts
```diff
@@ -66,4 +69,26 @@ export class GameSetup {
...
+    player.addProduction(Resources.STEEL);
+    player.addProduction(Resources.TITANIUM);
+    player.addProduction(Resources.PLANTS);
+    player.addProduction(Resources.ENERGY);
+    player.addProduction(Resources.HEAT);
+  }
+
+  public static includesCommunityColonies(gameOptions: GameOptions) : boolean {
+    if (!gameOptions.customColoniesList) return false;
+    if (gameOptions.customColoniesList.includes(ColonyName.IAPETUS)) return true;
```
bafolts, line 84: This is only moving and perhaps we need to create an issue to follow up on this but at some point someone is going to add a community colony and forget to add it here. This can also use `gameOptions.customColoniesList.some`.

src/Game.ts
bafolts: The word "right" is a trigger word for me. Either keeping it in or removing it is fine. This isn't a blocker for me. If the commented function remained in place and we never get around to updating it then this will remain in the code. If someone wanted to see what the old `cloneGame` implementation was they would see it if they went back in git history.

src/Player.ts
```diff
@@ -1756,13 +1756,11 @@ export class Player implements ISerializable<SerializedPlayer> {
...
-        Database.getInstance().restoreGame(game.id, game.lastSaveId - 2, game);
-        Database.getInstance().deleteGameNbrSaves(game.id, 1);
-        this.usedUndo = true; // To prevent going back into takeAction()
-      } catch (error) {
-        console.log(error);
-      }
+      GameLoader.getInstance().restoreGameAt(game.id, game.lastSaveId - 2, (game) => {
+        if (game !== undefined) {
+          this.usedUndo = true; // To prevent going back into takeAction()
+        }
```
bafolts, line 1762: The code before this was assuming `restoreGame` and `deleteGameNbrSaves` ran synchronously when they don't. This change is making the same assumption but setting `usedUndo` after the asynchronous operation now. May need to move the `this.usedUndo` outside of this callback as I don't know what impact this will have on the undo operation. Having the `GameLoader` do this is great is it moves direct calls to `Database` from `Player` or `Game`.

src/database/GameLoader.ts
```diff
@@ -98,32 +96,47 @@ export class GameLoader {
...
 
+  public restoreGameAt(gameId: GameId, saveId: number, cb: LoadCallback): void {
+    try {
+      Database.getInstance().restoreGame(gameId, saveId, (err, game) => {
+        if (game !== undefined) {
+          Database.getInstance().deleteGameNbrSaves(gameId, 1);
+          this.add(game);
+          cb(game);
+        } else {
+          console.log(err);
```
bafolts, line 107: This can use `console.error`.

src/database/GameLoader.ts
```diff
@@ -98,32 +96,47 @@ export class GameLoader {
...
+          cb(undefined);
+        }
+      });
+    } catch (error) {
```
bafolts, line 111: I don't think `Database.getInstance().restoreGame` will throw an error but if there is an error the `err` argument will be defined. Instead of catching error here I would check that `err` is not defined.

bafolts (approved):

kberg:
I've done manual verification, and I think this is good to go. I realize there's no new unit tests, ... gah.

PR #2186, opened 2020-12-25
Report correct name when CardFinder can't find a card.

kberg:


alrusdi (approved):

kberg:
Thank you! For this and the other reviews.

Issue #2185, opened 2020-12-25
Standard projects as cards


sienmich:
When I'm playing with new players, they don't know what standard projects are doing. Aquifer is not the most common English word. You have to know, that City gives you 1 MC prod. Also you can not check projects prices if you can't afford them.
  
I would like to have standard projects as cards.
Why?
1. They look better than list
2. Better info for new players
3. Cleaner code (`Player.ts` is too big file, it should be split anyway)

Like so:
![Screenshot from 2020-12-25 01-52-48](https://user-images.githubusercontent.com/19472604/103112196-07fab480-4654-11eb-82aa-082ae75e3569.png)
Maybe they can be smaller to better distinguish?
![Screenshot from 2020-12-25 01-55-05](https://user-images.githubusercontent.com/19472604/103112445-bbb07400-4655-11eb-84e1-cb838c8e96d0.png)

Is it a good idea? What do you think?

kberg:
Love it. Riffing on your example, I'd further distinguish it from a card like so.

![image](https://user-images.githubusercontent.com/413481/103112747-5c3a7000-4625-11eb-8a43-99067903b37e.png)

kberg:
Ain't I A Artist?

(Possibly a different color banner, too?)

Lynesth:
https://github.com/bafolts/terraforming-mars/issues/953#issuecomment-711088298

sienmich:
Nice! I like those layouts.
@kberg, all of the standard projects except selling patents have cost in MC, so putting that cost in corner is reasonable imo.
Ok, I'll work on this :)

kberg:
You know, I wrote why I didn't think it belongs in the corner for a reason: it makes it look as if it's a pay once use once card. If your goal is to make it clear to new users, then I think it's worth this extra step. In fact, I'd remove the entire top part of the card up to but not including, the blue line. (Don't make me draw again.)

chosta:
This again :)? I propose a discord discussion before implementing. I
vaguely remember it was a bit controversial. I am all for "card" look but
let's make sure it's the right thing to do.

kevinb9n:
I definitely still favor doing something like this.

I'm aware of two options (that have precedent) for how to depict them visually:

1. as if they are cards – that you don't have to actually have first and that don't go anywhere when played
2. as if they are card-actions (with red arrow) – which magically aren't on a card (i.e., there is nothing to mark as used for the rest of the gen)

some advantages of 1:

[...]

PR #2188, opened 2020-12-25
Standard projects as cards

sienmich:
Idea from https://github.com/bafolts/terraforming-mars/issues/2185

Currently they look like this (edited):
Screenshot from 2020-12-27 01-45-22


kevinb9n:
Looking good!
It looks like it's using greenery_no_o2 instead of greenery.png?

Not sure about a banner color, but yellow is sorta associated with corporation cards already. I thought of using gray because it also indicates neutral-ness (like gray delegates, sorta).

kevinb9n:
Only asking if it happens to be easy:
Can *some* symbol like maybe an asterisk go in the expansion symbol location for Buffer Gas, just as a reminder that it won't always be there? Feel free to ignore.

src/CardFinder.ts
```diff
@@ -114,4 +131,24 @@ export class CardFinder {
       });
       return result;
     }
+
+    public standardProjectsFromJSON(cards: Array<ICard | CardName>): Array<StandardProjectCard> {
```
kberg, line 135: Since standard projects are fully computable by gameOptions, I would not use serialization / deserialziation at all. Just rebuild the standard projects from the game options.

src/CardName.ts
```diff
@@ -1,4 +1,16 @@
 export enum CardName {
+    // Standard projects:
+    STANDARD_GREENERY = 'Greenery',
```
kberg, line 3: Keep the same order that the rest of the cards have, which I believe is mostly by expansion. Unless I'm mistaken.

src/Dealer.ts
```diff
@@ -168,6 +174,7 @@ export class Dealer implements ISerializable<SerializedDealer> {
         deck: this.deck.map((c) => c.name),
         discarded: this.discarded.map((c) => c.name),
         preludeDeck: this.preludeDeck.map((c) => c.name),
+        standardProjects: this.standardProjects.map((c) => c.name),
```
kberg, line 177: I like THIS.

src/SerializedDealer.ts
```diff
@@ -7,4 +7,5 @@ export interface SerializedDealer {
     deck: Array<IProjectCard | CardName>;
     discarded: Array<IProjectCard | CardName>;
     preludeDeck: Array<IProjectCard | CardName>;
+    standardProjects: Array<IProjectCard | CardName>;
```
kberg, line 10: Same - this can entirely be removed.

src/SerializedPlayer.ts
```diff
@@ -44,7 +44,6 @@ export interface SerializedPlayer {
     plants: number;
     plantsNeededForGreenery: number;
     playedCards: Array<SerializedCard>;
-    powerPlantCost: number;
```
kberg, line 47: Why is that?

src/cards/StandardCardManifests.ts
```diff
@@ -221,6 +221,13 @@ import {WavePower} from './base/WavePower';
 import {Windmills} from './base/Windmills';
 import {Worms} from './base/Worms';
 import {Zeppelins} from './base/Zeppelins';
+import {Aquifer} from './standardProjects/Aquifer';
+import {City} from './standardProjects/City';
+import {PowerPlantStandard} from './standardProjects/PowerPlant';
+import {PowerPlantThorgate} from './standardProjects/PowerPlantThorgate';
+import {Greenery} from './standardProjects/Greenery';
+import {AsteroidStandard} from './standardProjects/Asteroid';
+import {BufferGas} from './standardProjects/BufferGas';
```
kberg, line 230: This belongs in the expansion the standard project comes from.

src/cards/StandardCardManifests.ts
```diff
@@ -221,6 +221,13 @@ import {WavePower} from './base/WavePower';
 import {Windmills} from './base/Windmills';
 import {Worms} from './base/Worms';
 import {Zeppelins} from './base/Zeppelins';
+import {Aquifer} from './standardProjects/Aquifer';
+import {City} from './standardProjects/City';
+import {PowerPlantStandard} from './standardProjects/PowerPlant';
```
kberg, line 226: I'm pretty confident these two projects can be conjoined into one.

kberg (commented):
I'll review more later. Good first step.

src/CardFinder.ts
sienmich: Agree, fixed

src/cards/StandardCardManifests.ts
sienmich: You mean the BufferGas? Not really. It was introduced in Prelude, but it is meant to be in game no mater if you are using the prelude cards or not (if you don't use them, then solo mode is 14 turns, but still has the Buffer Gas). Right?

src/cards/StandardCardManifests.ts
sienmich: You mean the two power plants? I guess you are right, normal cards don't change their costs yet with discounts. I'll delete the second one

sienmich: Done

src/SerializedPlayer.ts
sienmich: Right now the condition if a player is thorgate is checked only in `PowerPlant.ts` (with `player.corporationCard?.name === CardName.THORGATE`), so there is no point to keep for every player this `powerPlantCost` variable imo

src/CardName.ts
sienmich: At the beginning of this file it is sorted by expansion and alphabetically, from the middle it's going more and more random order. I think it's nice to have them together, as they differ from normal cards more than normal cards differ between expansions. I ordered them by the price. Ok?

sienmich:
Ok, finished :smile:
Reviewers are welcome. Would you like some testing server?

About those colors: I'm far from a graphic. I tried some grays but they were really ugly. For me those goldy yellows are quite interesting. Those Standard Projects seem very powerful now :) I'm very open to some color propositions, best with two color hexes. Maybe even `#dev-topics` pool?

kberg:
> Ok, finished 
> Reviewers are welcome. Would you like some testing server?
> 
> About those colors: I'm far from a graphic. I tried some grays but they were really ugly. For me those goldy yellows are quite interesting. Those Standard Projects seem very powerful now :) I'm very open to some color propositions, best with two color hexes. Maybe even `#dev-topics` pool?

The colors are two things:
1. Great
2. Good enough. Launch as is.

src/Dealer.ts
```diff
@@ -16,44 +15,41 @@ import {CardTypes, Deck} from './Deck';
...
 import {ARES_CARD_MANIFEST} from './cards/ares/AresCardManifest';
+import {StandardProjectCard} from './cards/standardProjects/StandardProjectCard';
+import {GameOptions} from './Game';
 
 export class Dealer implements ISerializable<SerializedDealer> {
     public deck: Array<IProjectCard> = [];
     public preludeDeck: Array<IProjectCard> = [];
     public discarded: Array<IProjectCard> = [];
     public corporationCards: Array<CorporationCard> = [];
+    public standardProjects: Array<StandardProjectCard> = [];
```
kberg, line 26: Why exactly would one need to _deal_ from a deck of standard projects? I bet the answer to that question is "it doesn't, actually" and this implementation can be further simplified.

src/Dealer.ts
```diff
@@ -64,46 +60,42 @@ export class Dealer implements ISerializable<SerializedDealer> {
           .map((cf) => new cf.Factory());
         deck.push(...cardInstances);
       }
-      function addToDecks(manifest: CardManifest) {
+      function addToDecks(manifest: CardManifest, standardProjectsOnly = false) {
```
kberg, line 63: Why is a standardProjectsOnly parameter necessary?

src/Player.ts
```diff
@@ -2503,4 +2194,18 @@ export class Player implements ISerializable<SerializedPlayer> {
 
     return colonyTilesAlreadyBuiltOn < game.colonies.length;
   }
+
+  public getAvailableStandardProjects(game: Game) {
+    let projects = game.dealer.standardProjects;
+    if (game.getPlayers().length !== 1) {
```
kberg, line 2200: This probably should be done at game set-up.

src/Player.ts
```diff
@@ -2503,4 +2194,18 @@ export class Player implements ISerializable<SerializedPlayer> {
...
+      projects = projects.filter((card) => card.name !== CardName.STANDARD_BUFFER_GAS);
+    }
+
+    return new SelectCard(
+      'Standard projects',
+      'Confirm',
+      projects.sort((a, b) => a.cost - b.cost).filter((card) => card.canAct(this, game)),
```
kberg, line 2207: This sort is probably debateable. I'd rather see a defined order. Since Standard Projects are small and really won't change much, we can define a hard-coded order. That's kind of my preference.

kberg (commented):
Wondering just how much this round of comments are helpful or not.

I'd like to try to remove the complexity out of Dealer, which I think is possible, and I think really will simplify what you're doing. Perhaps this weekend I can take a crack at it, if you think what I'm suggesting is either unclear or bonkers.

Leave the cards in the Manifests, that's a good thing (I've secretly wanted CardManifest to really be an `ExpansionManifest` which might contain things like Colonies, eg, so whether or not these are cards, I see them as belonging there.

Lynesth:
Hmm, I'm a bit like @kberg here.
I understand wanting to make them look like cards in the front end, but why are you making them cards in the back end ? That's not what they are and it's really confusing me.

sienmich:
> why are you making them cards in the back end ?
@Lynesth, my point of view:
For example, there may come a time (maybe soon :smile:) when somebody would like to rewrite how `canPlay()` and  `canAct()` are working to include all "ocean placement bonus vs reds policy" scenarios. Having more common interfaces in our repo makes everything easier to change and adapt, coding becomes simpler. `ICard` fits perfectly for SP,  they have their layout generated from `metadata`, they have their `name`, `play/action` and `canAct/canPlay` methods. I think that having them as cards is much cleaner than 400 lines of `if` instructions in `Player.ts`.
I know that SP are not cards in the board game, but the board version is managed by humans. For us it will be easier in future to have them as cards.
Would you agree? :)

kberg:
For the record, I like that they look and act like cards. I'm just saying
the similarity ends there.

>

src/Dealer.ts
sienmich: "it doesn't, actually" :smile:

However I see this that way:
We need some structure / object which loads all SPs and stores them. There are two options:
1. I write new structure for SP only
2. I use the Dealer.
3. Eventually I could split the Dealer into CardLoader and Dealer, then SP could only use the CardLoader.
I think 2. is the simpler one - less redundant code like loading the cards, or choosing which expansion to include.
Would you agree? If not I can change, no problem :)

src/Player.ts
sienmich: I agree, fixed

src/Player.ts
sienmich: I think sort is not that bad, that's the order in original game. Hard coding sounds like additional unnecessary code for me. However we don't need to sort them every time, moved that to game creation.

src/Dealer.ts
sienmich: Maybe 3 is not that bad after all, i'm willing to choose 3 if u want :)

src/Dealer.ts
sienmich: Because SPs are not being serialized I need to load them with every game creation from card manifests.
Like so:
```
    public static deserialize(d: SerializedDealer, gameOptions: GameOptions): Dealer {
      const dealer = new Dealer();
   ...
      dealer.discarded = cardFinder.cardsFromJSON(d.discarded);
      dealer.preludeDeck = cardFinder.cardsFromJSON(d.preludeDeck);
      dealer.standardProjects = this.newInstance(gameOptions, true).standardProjects;
      return dealer;
    }
```
To speed things up I'm loading only SPs, not every card in game.
I'm not sure about that one. Is there a better way?

src/CardFinder.ts
bafolts: If this is fixed is this function still needed?

src/Dealer.ts
bafolts: I would not wedge standard projects in through `Dealer`. The standard projects most likely make more sense associated with `Board` or `Game` than the `Dealer`.

sienmich: Ok. I will split the Dealer into Dealer and CardLoader. I will use the CardLoader part but I'll store SPs in Game.

src/Player.ts
```diff
@@ -1321,211 +1317,6 @@ export class Player implements ISerializable<SerializedPlayer> {
...
-            this.increaseTerraformRatingSteps(1, game);
-            this.onStandardProject(StandardProjectType.BUFFER_GAS);
-          },
-        ));
-        return undefined;
-      },
-    );
-  }
-
-  private buildPowerPlant(game: Game): PlayerInput {
```
bafolts, line 1418: Regardless of any other comments I leave on this pull request this change is great. Moving these into another file was a long time coming.

sienmich: Thanks! :D

src/Dealer.ts
bafolts: The `ALL_CARDS_MANIFESTS` should provide access to the standard projects. A `CardLoader` may be overkill but I can never complain if there is less logic in `Game`.

src/Player.ts
```diff
@@ -129,7 +126,6 @@ export class Player implements ISerializable<SerializedPlayer> {
   // Turmoil
   public turmoilScientistsActionUsed: boolean = false;
 
-  public powerPlantCost: number = 11;
```
bafolts, line 132: I can't complain with this going away. When first writing all the project cards way back when I often tried to avoid explicitly calling out cards for game logic. With where the project is now I will always be in favor of using less memory in this way when we can calculate the value from another stored value.

src/Dealer.ts
```diff
@@ -64,46 +60,42 @@ export class Dealer implements ISerializable<SerializedDealer> {
...
-      if (ares) {
-        addToDecks(ARES_CARD_MANIFEST);
-      }
-      if (promoCards) {
-        addToDecks(PROMO_CARD_MANIFEST);
-      }
-      if (communityCards) {
-        addToDecks(COMMUNITY_CARD_MANIFEST);
+
+      const manifests = [
```
bafolts, line 72: You shouldn't need to type cast if you use `const manifests: Array<[boolean, CardManifest>] = []`

sienmich: Thanks, I was just googling this one :)

src/cards/standardProjects/BufferGas.ts
```diff
@@ -0,0 +1,35 @@
...
+import {StandardProjectCard} from './StandardProjectCard';
+import {PartyHooks} from '../../turmoil/parties/PartyHooks';
+import {PartyName} from '../../turmoil/parties/PartyName';
+
+export class BufferGas extends StandardProjectCard {
+  public name = CardName.STANDARD_BUFFER_GAS;
+  public cost = 16;
+
+  public canAct(player: Player, game: Game): boolean {
+    let cost = this.cost;
```
bafolts, line 16: Instead of removing this manually I would just return false here if not a single player game. This should remove complexity in game setup with having to remove this project.

sienmich: Cool, done

src/cards/standardProjects/PowerPlant.ts
```diff
@@ -0,0 +1,39 @@
...
+    let cost = this.cost;
+    if (player.corporationCard?.name === CardName.THORGATE) {
+      cost -= 3;
+    }
+    return player.canAfford(cost, game);
+  }
+
+  actionEssence(player: Player): void {
+    if (player.corporationCard?.name === CardName.THORGATE) {
+      player.megaCredits += 3;
```
bafolts, line 23: Giving 3 mega credits after paying for the project is different than the cost being 3 less. If you only have 8 mega credits the `canAct` will return true and this could very well happen to work where we subtract megacredits to a negative value only to then add 3 here to return a player to 0. Instead of changing in this way I would make the `getPowerPlantCost` a public getter on `Player`.

```
public Player {
   get powerPlantCost(): number {
     return this.corporationCard?.name === CardName.THORGATE ? 8 : 11;
   }
}
```

src/cards/standardProjects/StandardProjectCard.ts
```diff
@@ -0,0 +1,47 @@
...
+          this.actionEssence(player, game);
+        },
+      ));
+
+      game.log('${0} used ${1} standard project', (b) => b.player(player).card(this));
+      player.onStandardProject(this);
+
+      return undefined;
+    }
+    abstract canAct(player: Player, game: Game): boolean
```
bafolts, line 46: I am not familiar with the reds ruling party but if using this base class could consider adding the cost there. Is it specific to certain projects or based on a property like `cost` that could be calculated? If it is based on the `name` then it wouldn't make sense to put here.

src/cards/standardProjects/StandardProjectCard.ts
```diff
@@ -0,0 +1,47 @@
...
+    public cardType = CardType.STANDARD_PROJECT;
+    public hasRequirements = false;
+    public tags = [];
+    public abstract name: CardName;
+    abstract cost: number;
+
+    public play() {
+      return undefined;
+    }
+    abstract actionEssence(player: Player, game: Game): void
```
bafolts, line 27: This can be `protected` as far as I can tell.

src/components/card/Card.ts
```diff
@@ -109,8 +115,8 @@ export const Card = Vue.component('card', {
       if (this.actionUsed) {
         classes.push('cards-action-was-used');
       }
-      if (this.isCorporationCard()) {
```
bafolts, line 112: Did something change where the `isCorporationCard` block can now be removed?

src/styles/cards.less
```diff
@@ -1212,28 +1216,6 @@ input[type="radio"]:checked + .filterDiv::after {
     box-shadow: 0 2px 2px 1px rgba(0, 0, 0, 0.8);
 }
 
-.card-colony {
```
bafolts, line 1215: Did this move somewhere? Surprised to see this being removed.

src/styles/cards.less
```diff
@@ -89,10 +89,10 @@ input[type="checkbox"]:checked + .filterDiv::after,
 input[type="radio"]:checked + .filterDiv::after {
     content: "✓";
     position: absolute;
-    width: 252px;
+    width: calc(100% + 12px);
```
bafolts, line 92: Did this `width` and `top` need to change? This will impact more than only the standard project cards being added. What did this change visually? With enough changing in this pull request may be better to make this change separately.

bafolts (changes requested):
I had a few comments and questions about stuff which seemed like they may have been removed by mistake. Similar to @kberg I also think we should move the complexity out of `Dealer` as the list of standard projects is static for the most part. This seems very close. We didn't have unit tests prior to this. With these standard projects moving into their own classes it should be much easier to unit test them. If unit tests not added before merging we should make an issue to write unit tests for these.

Lynesth:
> Having more common interfaces in our repo makes everything easier to change and adapt, coding becomes simpler.

I won't disagree with that, but the only thing standard projects would have from `ICard` are `name`, `canAct` and `action`.
My point is that a `Bike` is not a `Car` and wouldn't be lumped with them or be made part of a `CarDealer` just because it can be use as a transportation mean and has wheels. An underlying interface/class that both implement/extend from would most likely be made in that case.

Nevertheless, I would keep each of them as their own class, like you did, since this is a good thing to remove them from Player. 👍

src/styles/cards.less
sienmich: It was unused for a while.
There was also a conflict. SP Colony automatically is getting `.card-colony` class, so removing it was a fix.

src/styles/cards.less
sienmich: I changed that in a way, that this orange selection tool works both on normal cards and SPs (which are smaller).
Without this change the selection box was always normal size, so was really ugly with SPs.

src/components/card/Card.ts
sienmich: This class (`card-corp`) was unused

src/cards/standardProjects/StandardProjectCard.ts
sienmich: Agree, I'll change that

src/cards/standardProjects/StandardProjectCard.ts
sienmich: You mean to automatize that paying additional 3 MC? I agree so much. However that's how every other card in game is currently coded, it's a thing for another PR.

src/cards/standardProjects/StandardProjectCard.ts
sienmich: Done

sienmich:
Allright, the Dealer has been split into Dealer and CardLoader. Now the Dealer is created from a CardLoader. SPs are stored in Game (just `public standardProjects: Array<StandardProjectCard>;`) and are loaded also using the CardLoader. I agree this is much cleaner than with the Dealer only.
@kberg, @bafolts

src/cards/standardProjects/PowerPlant.ts
sienmich: Fixed

src/CardLoader.ts
```diff
@@ -0,0 +1,85 @@
...
+    return manifest.corporationCards;
+  }
+  public static getPreludeCards(manifest: CardManifest) {
+    return manifest.preludeCards;
+  }
+
+  public getCards<T extends CardTypes>(getter: (arg0: CardManifest) => Deck<T>) : Array<T> {
+    let deck: Array<T> = [];
+    for (const manifest of this.manifests) {
+      this.addToDeck(deck, getter(manifest));
```
sienmich, line 78: I'm not js native speaker. `getCards` is meant to be called like `getCards(CardLoader.getCorporationCards)`. Is there a better way to do this?

src/Game.ts
```diff
@@ -219,22 +222,15 @@ export class Game implements ISerializable<SerializedGame> {
...
-      gameOptions.cardsBlackList,
-    );
+    const cardLoader = new CardLoader(gameOptions);
+    const dealer = Dealer.newInstance(cardLoader);
 
     const activePlayer = first.id;
 
     const game: Game = new Game(id, players, first, activePlayer, gameOptions, seed, board, dealer);
 
+    game.standardProjects = cardLoader.getCards(CardLoader.getStandardProjects).sort((a, b) => a.cost - b.cost);
```
bafolts, line 232: The `standardProjects` property can be set within the constructor. The `gameOptions` is provided as an argument to the `Game` constructor.

src/Game.ts
```diff
@@ -147,6 +149,7 @@ export class Game implements ISerializable<SerializedGame> {
   public phase: Phase = Phase.RESEARCH;
   public dealer: Dealer;
   public board: Board;
+  public standardProjects: Array<StandardProjectCard> = [];
```
bafolts, line 152: If this is set from within the constructor you shouldn't need to set it to an empty array. It will be populated once the constructor has finished executing.

src/CardLoader.ts
```diff
@@ -0,0 +1,85 @@
...
+      if (this.gameOptions.cardsBlackList) {
+        deck = deck.filter((card) => ![...this.gameOptions.cardsBlackList].includes(card.name));
```
bafolts, line 80: `cardsBlackList` is an array and according to it's typedef it will be defined. The check for it being a truthy value is unnecessary. As it is already an array pushing it to another array is also unnecessary. This can be simplified to:

```
deck = deck.filter((card) => this.gameOptions.cardsBlackList.includes(card.name) === false);
```

src/Game.ts
sienmich: Right, fixed

src/Game.ts
sienmich: Agree

src/CardLoader.ts
sienmich: True, thanks

bafolts:
As 45 files have changed here I ran out of steam reviewing today. I will finish looking at this tomorrow. I seem to be only spotting last minute minor changes at this point.

bafolts (changes requested):
OK. This time I attempted to leave all my comments at once. I am memory obsessed and this change will be adding a property to `Game` so runs the risk of using more memory per game. We sometimes run a thousand games on the terraforming-mars.herokuapp.com per day. On any given day I would say 100 - 1000 games run. We are already using too much memory.

With that in mind I would only change two more things with this PR:

1. Use singletons for the standard project classes. This will ensure we only have one copy of these in memory and we don't intend these to be used in any other way than this. Since `ICardFactory` is assuming that `Factory` always provides a new copy of the card we will need to modify `ICardFactory`. I would add an optional `Instance` that if not undefined returns that instance.

2. Avoid storing `standardProjects` on `Game`. This list can be calculated each time it is needed from `game.gameOptions` using `CardLoader`.

src/CardLoader.ts
sienmich: Thanks!

src/cards/venusNext/VenusCardManifest.ts
```diff
@@ -109,6 +110,9 @@ export const VENUS_CARD_MANIFEST = new CardManifest({
     {cardName: CardName.OMNICOURT, Factory: Omnicourt},
     {cardName: CardName.MINING_QUOTA, Factory: MiningQuota},
   ],
+  standardProjects: [
+    {cardName: CardName.STANDARD_AIR_SCRAPPING, Factory: AirScrapping},
```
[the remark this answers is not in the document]
sienmich: @bafolts 
Ok. From what I understand if you want to have `Factory: AirScrapping.getInstance`, then I need to change  
`Factory: new () => T;` into `Factory: new () => T | () => T;`. Am I right?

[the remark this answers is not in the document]
sienmich: Putting that `.getInstance()` in every card would create a lot of unnecessary code. I don't see a way to make it once in some `Card` class.
However I came up with this solution:
```ts
export function sharedObject<T>(Obj: new () => T): () => T {
  class sharedObject {
    public constructor() { }

    private static instance?: T = new Obj();

    public static getInstance(): T {
      if (sharedObject.instance === undefined) {
        sharedObject.instance = new Obj();
      }
      return sharedObject.instance;
    }
  }
  return sharedObject.getInstance;
}
```
Does it make sense to you? Then we can do `Factory: sharedObject(AirScrapping)`. Still requires changing Factory type, as above. And that change is big. Maybe I'll open next PR?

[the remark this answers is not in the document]
bafolts: Doing that in a separate PR should be fine. The memory difference should be equivalent to less than 10 cards added. I could work on that change. First thought of it today while reviewing this PR and it would go a long way into reducing memory usage as most cards don't need to have separate instances created.

bafolts (approved):

src/cards/venusNext/VenusCardManifest.ts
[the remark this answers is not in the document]
bafolts: As the standard projects are not stored in memory on `Game` this shouldn't be an issue anymore anyway. They will get created and then garbage collected after used.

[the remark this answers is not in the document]
sienmich: Agree, all green cards for sure.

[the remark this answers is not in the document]
bafolts: Will need an elegant way to implement it. Most likely some thing on the card manifest or card factory can store it if we see the card doesn't have a resource or some other property that would be tracking state.

PR #2190, opened 2020-12-26
s/Tags.STEEL/Tags.BUILDING and s/Tags.MICROBES/Tags.MICROBE.

kberg:
Since each Tag has a string value, this should have no impact on running games.


nwai90 (approved):

bafolts (approved):
Thank you for this.

PR #2194, opened 2020-12-27
Include `Player.spendableMegacredits`, which takes heat into account.

kberg:
Is this a good idea? It reduces some use, and I'd use it at least twice.

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -981,6 +981,14 @@ export class Player implements ISerializable<SerializedPlayer> {
     );
   }
 
+  /**
+   * @return {number} the number of avaialble megacredits. Which is just a shorthand for megacredits,
+   * plus any units of heat available thanks to Thorgate.
+   */
+  public spendableMegacredits(): number {
+    return (this.canUseHeatAsMegaCredits) ? (this.heat + this.megaCredits) : this.megaCredits;
+  }
+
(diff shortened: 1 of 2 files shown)


src/Player.ts
```diff
@@ -981,6 +981,14 @@ export class Player implements ISerializable<SerializedPlayer> {
     );
   }
 
+  /**
+   * @return {number} the number of avaialble megacredits. Which is just a shorthand for megacredits,
+   * plus any units of heat available thanks to Thorgate.
```
bafolts, line 986: I believe this is thanks to Helion.

src/Player.ts
```diff
@@ -1824,9 +1827,6 @@ export class Player implements ISerializable<SerializedPlayer> {
       const canUseSteel = card.tags.indexOf(Tags.BUILDING) !== -1;
       const canUseTitanium = card.tags.indexOf(Tags.SPACE) !== -1;
       let maxPay = 0;
-      if (this.canUseHeatAsMegaCredits) {
```
bafolts, line 1827: I never argue with a PR removing branches.

bafolts (approved):

src/Player.ts
kberg: Ugh, I thought I fixed it beforehand. /shamecube

PR #2201, opened 2020-12-28
separate teplate for v-else

bafolts:
I moved the `v-else` onto an element with `v-for` and this seems to have indirectly attempted to run the `v-else` on each item in the array. I moved this to a template and that seems to have solved the problem.


Lynesth (approved):

bafolts:
I will deploy this to the main site after merging.

PR #2202, opened 2020-12-28
Move player ID generation outside of the Player constructor.

kberg:
Once this is done, I think cloning will be easier.


bafolts (approved):

bafolts:
Cloning will definitely be easier after this.

PR #2203, opened 2020-12-28
santa leaving mars

bafolts:
One must assume that even many generations into life on mars society will frown upon domiciles or business establishments which show a failure to remove their decorations after the holidays.


kberg:
I would not be opposed to having those decorations on a) a timer and b) controlled by environment variables for someone who doesn't want its inclusion. But for now...

kberg (approved):

PR #2204, opened 2020-12-28
Highlight volcanic spaces and Noctis City

illicitonion:
In playing, I found that Noctis City was easy to overlook as a reserved
space because the line to its label is easy to miss. This draws extra
attention to spaces which have special conditions.


kberg:
Would you mind pasting a screencap of the change?

illicitonion:
Of course!

Tharsis:
tharsis

Elysium:
elysium

There is no change for Hellas, as it has no special spaces.

src/SpaceType.ts
```diff
@@ -2,5 +2,6 @@
 export enum SpaceType {
     LAND = 'land',
     OCEAN = 'ocean',
-    COLONY = 'colony'
+    COLONY = 'colony',
+    LAND_HIGHLIGHTED = 'land-highlighted',
```
bafolts, line 6: A `LAND_HIGHLIGHTED` is still a `LAND` so I think this may cause more harm than good. Any place we are checking if something is `SpaceType.LAND` we would also most likely need to check that it is also `LAND_HIGHLIGHTED`. On the front end the `SpaceModel` is used to provide information about a space. I would add a property to `SpaceModel` for `highlighted?: boolean` which indicates if the space is a `LAND_HIGHLIGHTED` instead of using this enum.

Instead of then passing `highlightedSpaces` to `BoardBuilder` you can create a `getHighlightedLandSpaces()` similar to `getNonReservedLandSpaces` which each board can override as necessary. Then if you modify `getSpaces` and provide `game.board` a highlighted space can be calculated when needed by the front end.

https://github.com/bafolts/terraforming-mars/blob/main/src/server/ServerModel.ts#L565
https://github.com/bafolts/terraforming-mars/blob/main/src/boards/Board.ts#L161

With the above suggestions this change shouldn't be a big deal for the back end. I am curious what @kevinb9n thinks of this change as they are among the top 10 people who give most thought to the user experience. The change makes sense to me to indicate these tiles are not only land spaces but tied to cards in some unique way.

illicitonion: Good call - I didn't realise how much `SpaceType` was used - switched over to just be a render-time concern when producing a `SpaceModel`. Thanks for the pointers!

src/server/ServerModel.ts
```diff
@@ -572,6 +576,7 @@ function getSpaces(spaces: Array<ISpace>): Array<SpaceModel> {
       spaceType: space.spaceType,
       tileType: space.tile && space.tile.tileType,
       color: getColor(space),
+      isHighlighted: highlightedSpaces.includes(space.id as SpaceName),
```
kberg, line 579: Pedantic: I'd prefer `highlight` but not pushing hard.

Suggest switching from SpaceName to string, which are mostly the same thing, but removes the implication that a highlighted spaces needs a custom name. (Which will be the truth soon as I move toward including The Moon, which has its own space entirely.)

I wouldn't object to adding a SpaceId type, doesn't have to be in this PR.

Clearly neither of these are urgent.

kberg (commented):
With the exception of discussing the UX, this is a nice simple addition.

But I still have a concern about the border. Not only does red have a connotation in Ares, see this PR #1971 which tried to solve the same problem. Here's my comment from there.

> I had hoped someone else would make opinions about the image itself, but they haven't. So, I'm not sure I think it's a good addition. It's visually striking, and so if it's going to be a visual attention getter, I ask what its value might be. Given that volcanoes are limited, in the base game, to just a couple of cards being impacted by them, it imposes a lot of visual demand for little return.

While this change is _less_ visually striking I have the same objection here. You don't _always_ need to know where the reserved spaces are.

kberg (changes requested):
Changing my review to 'request changes'

src/styles/board.less
```diff
@@ -253,6 +253,10 @@
 	background: url(assets/hex_white.png) no-repeat;
 }
 
+.board-space-type-land-highlighted {
+	background: url(assets/hex_red.png) no-repeat;
```
chosta, line 257: First of all, good job! I left a comment last night but it seems to be lost somehow. It's was about the visuals. In my humble opinion we should look for a colour sublter than 'cursed blood'. Also the border is does not need to be more than 1px, actually I would insist on that :) and I'm sure @ssimeonoff would agree.
Let me know if you need specific color suggestions, we can provide a couple.
Also, I'm pretty sure we can just use a border propery instead of an image.

chosta (changes requested):

src/styles/board.less
ssimeonoff: I really want those hexes to be thinner (the grey and blue ones as well) or removed all together, replaced by some light border/shade.

illicitonion:
I switched everything to 1px borders, and the special spaces to be black rather than red bordered. I think I preferred the thicker borders, but I'll leave it up to y'all.

My CSS isn't up for replacing background images with CSS borders, but it'd be great to see if someone else's is!

New screenshot:

Screenshot 2020-12-29 at 12 17 08

src/server/ServerModel.ts
illicitonion: Renamed to `highlight` and switched to be a `string`, updated names accordingly.

chosta:
Ah, my bad. By 1px I meant the width to be the same as the original tiles which is more than 1px. Please, revert to the normal width and keep it universal for all tiles. Black border looks better for volcanic areas for the moment.

illicitonion:
Screenshot 2020-12-29 at 13 38 33

src/styles/board.less
ssimeonoff: e.g only css
Untitled

chosta: I like this one better. @ssimeonoff please provide the code if possible so @illicitonion can do the change in this pr.

chosta:
Good, but black still sticks out too much. Please, check Simon's comment above. If we add a different hue to the tile itself instead of the border it does look like a mountain or something on a higher ground. Much more immersive.

kberg:
If a permanent border is what is going to happen, then I suggest this, or lighter:

image

kberg:
image

kberg:
image
I don't like this one, but I'm riffing on ideas.

kberg:
OR just decrease the opacity of the legend: 
image

This is my current favorite.

src/styles/board.less
ssimeonoff: It is pretty basic.

Add clip-path to all tiles for simplicity (or add it per different tiles class if you prefer)
``` 
.board-space {
    width: 46px;
    height: 51px;
    position: absolute;
    clip-path: polygon(50% 0, 100% 25%, 100% 75%, 50% 100%, 0 75%, 0 25%);
 }
```
And then add some background-color to the volcanic and Noctis tiles. For example:
- volcanic - #8408 (if you prefer it less transparent, e.g. try #840b)
- noctis - #fff4 (if you want it less transparent, try #fff6 - #fff8 feels way too white)

kberg:
Sorry to spam all of you.

That last one is my favorite, and also the simplest. Just remove this line

```
.board-caption {
	font-family: Tahoma,Verdana,Segoe,sans-serif;
	font-size: 10px;
	font-weight: bold;
	fill: white;
--->	fill-opacity: 0.5;
	pointer-events: none;
}
```

It doesn't have to be solely the default value, but it's a lot simpler than passing the new data around.

illicitonion:
I think we still need _something_ in the space itself, because right now the Monses at the edge of Mars are pretty easy to overlook.

I also think it needs to be in the border, not just the background - if Noctis City is surrounded by greeneries and oceans, a subtle background doesn't really show up as significant/different. If it's going to be a background change, it's going to need to be pretty major.

Maybe:

Screenshot 2020-12-29 at 17 31 12

Alternatively, I don't dislike the two-tone border in:

> image

chosta:
How about:
1. Increase legend opacity (as @kberg suggested)
2. Use volcanic color for the tiles (as @ssimeonoff suggested in the comment)
3. Don't change border (@chosta my personal preference)

Go for this and we can move on. Later someone can jump in and review if needed.

chosta:
Alternative:
Just use #840b for border color. And let's see the board with some greeneries if possible, next or on top of the volcanic areas (or mixed)

illicitonion:
Latest iteration:

Empty:
Screenshot 2020-12-29 at 18 08 27

Surrounded:
Screenshot 2020-12-29 at 18 11 14

chosta (approved):

kberg:
> I think we still need _something_ in the space itself, because right now the Monses at the edge of Mars are pretty easy to overlook.

That can be solved with the same SVG that draws lines to the spots.

kberg:
Roughly: 

image

I realize you have done a bunch of work, but we already have a legend, and I think that's the approach to take.

illicitonion:
> I realize you have done a bunch of work, but we already have a legend, and I think that's the approach to take.

I have no qualms with throwing things away if there's a better way! :)

> Roughly:
> 
> image

I think I'd be on board with doing _both_ of these things, but I think the shared (and distinguished) background colour is still a useful distinction to make, e.g. to see what spaces are grouped together as being common.

Being able to see "There is one volcanic spot left" because there's exactly one space of a colour seems a lot handier at a glance than looking at the the things pointed to by legends, distinguishing between the types (e.g. Noctis vs volcanics), and potentially filtering out the ones which already have tiles on...

kberg:
> Being able to see "There is one volcanic spot left" because there's exactly one space of a colour seems a lot handier at a glance than looking at the the things pointed to by legends, distinguishing between the types (e.g. Noctis vs volcanics), and potentially filtering out the ones which already have tiles on...

THAT convinces me. I agree. I'd suggest retaining a fair degree of opacity allowing the map beneath to show through.

Do both? The SVG and the tile?

kberg:
Daniel, I just realized I know who you are. Hello!

kevinb9n:
Hey! Your name also looks familiar to me (we might all work at the same company?).

I have a small humble request, which I'm happy to file separately if preferred:

I think that there should be something visual that all reserved spaces clearly have in common, which is not shared by the volcanic spaces, because I think it's a common misunderstanding among new players that volcanic spaces are reserved.

Ideally (but don't fall prey to unwanted scope-creep!), I would want this "reserved" style (whatever it is) to apply it to the spaces off Mars as well, and I'd even consider that when your *opponent* has played Land Claim or Arcadian Communities that those spaces appear this way too. Then, the concept of reservedness would have no mystery to it.....

illicitonion:
> Hey! Your name also looks familiar to me (we might all work at the same company?).

I used to, quite close to @kberg, but alas I have since moved on :)

> I have a small humble request, which I'm happy to file separately if preferred:
> 
> I think that there should be something visual that all reserved spaces clearly have in common, which is not shared by the volcanic spaces, because I think it's a common misunderstanding among new players that volcanic spaces are reserved.
> 
> Ideally (but don't fall prey to unwanted scope-creep!), I would want this "reserved" style (whatever it is) to apply it to the spaces off Mars as well, and I'd even consider that when your _opponent_ has played Land Claim or Arcadian Communities that those spaces appear this way too. Then, the concept of reservedness would have no mystery to it.....

I think that's generally a good idea, but definitely lean towards it being a separate request :)

kevinb9n:
> I think that's generally a good idea, but definitely lean towards it being a separate request :)

Filed, so you can forget it :-)

illicitonion:
> > I think that's generally a good idea, but definitely lean towards it being a separate request :)
> 
> Filed, so you can forget it :-)

Thanks!

@kberg @bafolts I think we're good to go now that #2233 is in:

Screenshot 2020-12-30 at 23 46 27

kberg:
Daniel, let's take a look at the image opacity, please. Matching the ocean,
at least. One could argue that Oceans are more important, yet those spaces
have opacity underneath. I'll give it a shot locally.

>

illicitonion:
> Daniel, let's take a look at the image opacity, please. Matching the ocean, at least. One could argue that Oceans are more important, yet those spaces have opacity underneath.

I consciously made them _more_ opaque than the oceans, because the changed border on the oceans makes them really easy to spot, whereas when the volcanic spaces are less opaque they're hard to distinguish from other spaces (particularly the ones around row 6 column 5 on Tharsis). I think anything less opaque than `#840a` is definitely too subtle, but I could live with anything in the `#840a` - `#840c` range... Open to other suggestions, though!

kberg:
Gotcha. Here's what I was thinking::

image


That LHS is #8834005c. Part of the problem is that you're using a different rendering mechanism from the oceans, so by creating transparency, it also turns the border white.

Here's how oceans are rendered.
```
.board-space-type-ocean {
	background: url(assets/hex_blue.png) no-repeat;
	clip-path: polygon(50% 0, 100% 25%, 100% 75%, 50% 100%, 0 75%, 0 25%);
    background-color: #0af4;
}
```

versus what you've done

```
.board-space-type-land-volcanic {
	background-color: #840c;
}
```

I think you'll be happier if you match that pattern. And here's what I think you can do. Gimp's color picker says that your border is #b68f6b (Your 840c mixed with the border.)

Make the `--volcanic` selector look like this

```
.board-space-type-land-volcanic {
	background: url(assets/hex_840c.png) no-repeat;
	clip-path: polygon(50% 0, 100% 25%, 100% 75%, 50% 100%, 0 75%, 0 25%);
	background-color: #8305;
}
```

In other words, do like the ocean and separate the border from the body. Your thoughts?

kberg:
I mean:

image

illicitonion:
> Make the `--volcanic` selector look like this
> 
> ```
> .board-space-type-land-volcanic {
> 	background: url(assets/hex_840c.png) no-repeat;
> 	clip-path: polygon(50% 0, 100% 25%, 100% 75%, 50% 100%, 0 75%, 0 25%);
> 	background-color: #8305;
> }
> ```
> 
> In other words, do like the ocean and separate the border from the body. Your thoughts?

I'm on-board with using a non-white hex background. I'll note that the `clip-path` is already in use, because of the style applied to `.board-space`, so it's really just the background image / colour split we're talking about.

> Left is a proposal, right is yours.
> 
> image

I think that left is still a bunch too subtle, and pretty strongly prefer the right there... But also, in your left, it looks like the hex borders are the same colour as other tiles, or very close to them. It feels like having a much more differentiated border allows for a much more subtle background (which is what the oceans do). How about something like this:

Screenshot 2020-12-31 at 01 02 44

kberg (approved):
Although I prefer lighter, this is more than good enough. In time I could see a good lighter solution.

kberg:
Thank you!

PR #2206, opened 2020-12-28
have static and dynamic factories for cards

bafolts:
When we create factories from the card manifest we always use factories which create new instances of a card. For most cards this is unnecessary as we don't track any state with the card. This should cut down on the amount of memory we are using while games are created as we will re-use one instance of a card between every dealer. For this first change I included every event card as those should not have any state as they are played as a one time deal. After this can slowly look for other cards which can be ported this way.


bafolts:
As a history lesson here. At first one instance of the card was always shared. There was a bug early on where resources on one card were being tracked between every instance of the card between games. At that point the factory was added and it was assumed that creating a new instance of the card was no harm done. Now with the number of games seen running on one server I don't think the overhead is worth it to not use one instance for most cards.

src/cards/StandardCardManifests.ts
```diff
@@ -232,245 +232,245 @@ import {SellPatents} from './standardProjects/SellPatents';
...
-    {cardName: CardName.UNDERGROUND_CITY, Factory: UndergroundCity},
-    {cardName: CardName.UNDERGROUND_DETONATIONS, Factory: UndergroundDetonations},
-    {cardName: CardName.URBANIZED_AREA, Factory: UrbanizedArea},
-    {cardName: CardName.WATER_IMPORT_FROM_EUROPA, Factory: WaterImportFromEuropa},
-    {cardName: CardName.WATER_SPLITTING_PLANT, Factory: WaterSplittingPlant},
-    {cardName: CardName.WAVE_POWER, Factory: WavePower},
-    {cardName: CardName.WINDMILLS, Factory: Windmills},
-    {cardName: CardName.WORMS, Factory: Worms},
-    {cardName: CardName.ZEPPELINS, Factory: Zeppelins},
+    CardManifest.dynamicFactory(CardName.ADAPTATION_TECHNOLOGY, AdaptationTechnology),
```
kberg, line 235: Consider renaming dynamicFactory to dynamic, and staticFactory to static, and importing those directly.

kberg:
This is a good change, even without my suggestions it's a net improvement. But just let me know your thoughts.

Lynesth:
This will break the latest `warning` property introduced (though could be reworked so that it's provided to the PlayerInput differently, with a map for example).

Instead of having dynamic and static factories, why not only have static ones and have a wrapper, a "sleeve", that holds the card once it's in a player's hand (or play area, etc) and that would have those dynamic properties, like `resourceCount` or `warning` ?

bafolts:
I will close this until it is ready. Always using static cards and creating some other interface for when the card is played is a better solution but will take time. Will look at other strategies to achieve the underlying goal here of reducing memory footprint now that `warning` has showed up. As long as the properties that weren't changing like `name`, `cost`, `tag`, etc were only stored once per class that would also work as then naturally the only properties that we copied per class would be the ones that could be modified. As they say, back to the drawing board!

bafolts:
Well ok, I feel dumb now. I think this can be solved even simpler.

Instead of using a property on the class

```ts
public cost = 9;
```

If we instead use a getter for these properties

```ts
public get cost(): number {
  return 9;
}
```

This change is small but the generated javascript is telling. In case 1 the generated javascript is:

```
function Ants() {
    this.cost = 9;
}
```
Whenever we call to `new Ants()` we are getting a `cost` property added with a new value of `9` stored in memory. In case 2 the situation is how we want it and the generated javascript is:

```
Object.defineProperty(Ants.prototype, "cost", {
      get: function () {
          return 9;
      },
      enumerable: false,
      configurable: true
});
```

Here _every_ instance of Ants will share this same function which will always return a new value of `9` but this shouldn't get stuck in memory.

The only _risk_ here is that since `enumerable` seems to get set to `false` if we were relying on walking the properties of a card somewhere that would fail, but I don't think we walk the properties in this way, especially with the update to avoid `Object.assign`.

@kberg and @Lynesth I will slowly start to port cards to this format. Can start with a few and if we don't see smoke keep adding the rest. This should achieve the goal of not constantly duplicating in memory the same values while avoiding larger structural changes.

bafolts:
The `metadata` propertly which recently rolled out I am assuming correlates with the recent increase in memory we see on heroku instances. This should remedy the situation.

kberg:
How much would using `readonly` mitigate this? I happened to start using
that with parameters to document that they don't change.

bafolts:
I don't think `readonly` will help AFAIK that is only enforced at typescript level. The value would still be duplicated and added as a new property on each instance. We want the properties on `ICard` that aren't changing to be on the `prototype` so they are shared between instances.

https://www.w3schools.com/js/js_object_prototypes.asp

This wont impact the interfaces as either way it is a property.

PR #2207, opened 2020-12-29
move adjacent map away from property and remove undefined return values

bafolts:
Some suggestions I had during code review of #2089 were lost in translation or in the mix of the larger pull request as a lot was changing. This is first pass of changes I would like to make before merging to `main`. I see a few more things that may need to be updated prior to merging.

As an upcoming change, I am not sure what order https://github.com/bafolts/terraforming-mars/blob/main/src/boards/BoardBuilder.ts#L75 is trying to sort. If it is supposed to be ascending or descending order I am not sure it is achieving that.


chosta:
I don't see an issue here. Probably good to go.

PR #2210, opened 2020-12-29
Improve some loops in Player.ts

Lynesth:


src/Player.ts
```diff
@@ -1212,8 +1209,10 @@ export class Player implements ISerializable<SerializedPlayer> {
 
     // Activate some colonies
     if (game.gameOptions.coloniesExtension && selectedCard.resourceType !== undefined) {
-      game.colonies.filter((colony) => colony.resourceType !== undefined && colony.resourceType === selectedCard.resourceType).forEach((colony) => {
-        colony.isActive = true;
+      game.colonies.forEach((colony) => {
```
kberg, line 1212: This is the exact same code that you changed on line 662 in Game.ts. Looks like there is room for duplication removal.

kberg (approved):

src/Player.ts
Lynesth: Yeah I noticed that, maybe later :)

PR #2211, opened 2020-12-29
Make CardLoader.get...Cards() less convoluted

Lynesth:
Unless I missed something I think it makes more sense to keep the "complication" of the method inside the class itself.


alrusdi (approved):

kberg (approved):

sienmich:
Yes, I knew there has to be a better way than what I did. Approved! :smile:

PR #2212, opened 2020-12-29
More improvements to some loops logic

Lynesth:


src/cards/ares/EcologicalSurvey.ts
```diff
@@ -17,7 +17,7 @@ export class EcologicalSurvey implements IProjectCard {
 
   private countGreeneryTiles(game: Game): number {
     return game.board.spaces.filter(
-      (space) => space.tile?.tileType === TileType.GREENERY).length;
+      (space) => space.tile !== undefined && space.tile.tileType === TileType.GREENERY).length;
```
chosta, line 20: What is wrong with `space.tile?.tileType`?

src/cards/promo/Philares.ts
```diff
@@ -91,19 +91,23 @@ export class Philares implements CorporationCard {
     }
 
     public onTilePlaced(player: Player, space: ISpace, game: Game): void {
+      const philaresPlayer = game.getPlayers().find((player) => player.isCorporation(CardName.PHILARES));
+      if (philaresPlayer === undefined) {
+        console.error('Could not find Philares player');
```
chosta, line 96: Should we add a test for Philares player not found?

src/cards/promo/Philares.ts
```diff
@@ -91,19 +91,23 @@ export class Philares implements CorporationCard {
...
+        return;
+      }
       if (space.tile !== undefined && space.tile.tileType !== TileType.OCEAN) {
         let bonusResource: number = 0;
-        if (space.player !== undefined && space.player.isCorporation(CardName.PHILARES)) {
+        if (space.player !== undefined && space.player.id === philaresPlayer.id) {
```
chosta, line 101: I have a feeling that this conditional logic could be even further simplified but not sure how atm :)

chosta (approved):

src/cards/ares/EcologicalSurvey.ts
Lynesth: In this case, nothing really, but it seems that we're going the "old" way of not using it, which I don't mind to be honest.
But there's other applications of the optional chaining operator that can lead to some issues, so I guess it's fine not using it.

src/cards/promo/Philares.ts
Lynesth: It should not happen though. I mean, if Philares is in play, it should have a player.
I mostly added the check so that TS doesn't complain about using `philaresPlayer.id` afterwards.

src/cards/promo/Philares.ts
Lynesth: Here's an idea. Not sure that's simplified though.
```ts
    public onTilePlaced(player: Player, space: ISpace, game: Game): void {
      if (space.tile === undefined || space.player === undefined) {
        return;
      }

      const philaresPlayer = game.getPlayers().find((player) => player.isCorporation(CardName.PHILARES));
      if (philaresPlayer === undefined) {
        console.error('Could not find Philares player');
        return;
      }

      const playerIsPhilares = player.id === philaresPlayer.id;
      let bonusResource: number = 0;
      bonusResource = game.board.getAdjacentSpaces(space)
        .filter((space) =>
          space.tile !== undefined &&
          space.player !== undefined &&
          (space.player.id === player.id) !== playerIsPhilares)
        .length;
      if (bonusResource > 0) {
        this.selectResources(philaresPlayer, game, bonusResource);
      }
    }
```

src/cards/ares/EcologicalSurvey.ts
bafolts: On the client side the optional chaining operator isn't always available. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

On most new browsers and node it is available so it shouldn't be a problem. I don't think it matters one way or the other. We are going the "old" way as I have been writing javascript too long before this optional operator was a thing and it's use is still growing on me. Since this code only runs server side I don't think it matters any way. I would prefer to avoid it client side as the browser support is relatively new. Avoiding it on both client and server is simpler as you don't have to consider where your code would run.

src/cards/promo/Philares.ts
bafolts: We would have failed silently prior to this change as all of the checks below would have never had a match. So the logging to `console.error` should be a net improvement.

bafolts (approved):
On the "main" site I have seen as much traffic as 20 requests per second for hours on end. During this time the load on the dyno will grow. Changes like this will go a long way. Thank you!

PR #2213, opened 2020-12-29
Replace saveGameState with saveGame. This moves lastSaveId management into the datbase implementation.

kberg:
The variations between the two database implementations is minor, but there is starting to look like enough that I might suggest a bridge design pattern to further isolate shared state that exists between the game and the database.


src/database/SQLite.ts
```diff
@@ -128,14 +128,18 @@ export class SQLite implements IDatabase {
...
+    this.db.run(
+      'INSERT INTO games(game_id, save_id, game, players) VALUES(?, ?, ?, ?)',
+      [game.id, game.lastSaveId, game.toJSON(), game.getPlayers().length],
+      (err: { message: any; }) => {
+        if (err) {
+          // Should be a duplicate, does not matter
+          return;
+        }
+      },
+    );
```
Lynesth, line 142: Need to increment game.lastSaveId here as well.

src/Game.ts
```diff
@@ -362,15 +362,15 @@ export class Game implements ISerializable<SerializedGame> {
...
-      * Because we save at the start of a player's takeAction, we need to
-      * save the game in the database before increasing lastSaveId so that
-      * reloading it doesn't create another new save on top of it, like this:
-      *
-      * increment -> save -> reload -> increment -> save
-      *
-      */
-    Database.getInstance().saveGameState(this.id, this.lastSaveId, this.toJSON(), this.players.length);
-    this.lastSaveId++;
+     * TODO(kberg): not sure what to do with this comment.
```
Lynesth, line 365: You can remove it from here and make a not in the databases' saveGame methods that the lastSaveId increment must happen after the save so that it is not moved. from there.

Lynesth (changes requested):

src/database/SQLite.ts
kberg: Good catch.

src/Game.ts
kberg: > You can remove it from here and make a note in the databases' `saveGame` methods that the `lastSaveId` increment must happen after the save so that no one inadvertently moves it.

Done, thanks.

Lynesth (approved):

PR #2214, opened 2020-12-29
Add saveGameResults to sqlite implementation.

kberg:
Resolve #2213 first.

```
> select * from game_results ;
c745a0dcfd69||1|14|{"boardName":"tharsis","undoOption":false,"showTimers":false,"fastModeOption":false,"showOtherPlayersVP":false,"corporateEra":true,"venusNextExtension":false,"coloniesExtension":false,"preludeExtension":false,"turmoilExtension":false,"aresExtension":false,"aresHazards":true,"promoCardsOption":false,"communityCardsOption":false,"solarPhaseOption":false,"removeNegativeGlobalEventsOption":false,"includeVenusMA":true,"draftVariant":false,"initialDraftVariant":false,"startingCorporations":2,"shuffleMapOption":false,"randomMA":"No randomization","soloTR":false,"customCorporationsList":[],"cardsBlackList":[],"customColoniesList":[],"requiresVenusTrackCompletion":false}|[{"corporation":"United Nations Mars Initiative","playerScore":14}]
```


src/database/SQLite.ts
```diff
@@ -73,8 +74,16 @@ export class SQLite implements IDatabase {
     });
   }
 
-  saveGameResults(_game_id: GameId, _players: number, _generations: number, _gameOptions: GameOptions, _scores: Array<Score>): void {
-    return;
+  saveGameResults(game_id: GameId, players: number, generations: number, gameOptions: GameOptions, scores: Array<Score>): void {
+    this.db.run(
+      'INSERT INTO game_results (game_id, seed_game_id, players, generations, game_options, scores) VALUES($1, $2, $3, $4, $5, $6)',
```
bafolts, line 79: I had no idea this was a thing. We could start displaying metrics on some page in regards to games played and list of scores. On the postgres server for the "main" site this list most likely has a lot of results.

bafolts: WOAH

```
terraforming-mars::DATABASE=> SELECT count(*) FROM game_results;
 count 
-------
 51439
(1 row)
```

bafolts: I want to avoid getting into the business of tracking user data but for single player games I think we could ask for initials or some name similar to arcade machines. This would allow a user to provide their own unique name to represent their score. Their score should last for quite some time on whatever server they were running on as we don't purge this table.

bafolts (approved):

PR #2216, opened 2020-12-29
Added checkMinRequirements and checkMaxRequirements

Lynesth:
It's pretty easy to forget to add/substract `player.getRequirementsBonus()` when checking for global requirements or to forget that you need to multiply Venus and Temperature by 2.

Anyway, less random comparisons, 2 new methods to take care of that.

All tests pass properly but, when reviewing, please make sure once more that I didn't invert a Min/Max somewhere (should be fine, but who knows).

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -1269,9 +1269,12 @@ export class Game implements ISerializable<SerializedGame> {
     } else if (parameter === GlobalParameter.TEMPERATURE) {
       currentLevel = this.getTemperature();
       playerRequirementsBonus *= 2;
-    } else { // parameter === GlobalParameter.VENUS
+    } else if (parameter === GlobalParameter.VENUS) {
       currentLevel = this.getVenusScaleLevel();
       playerRequirementsBonus *= 2;
+    } else {
+      console.warn(`Unknown GlobalParameter provided: ${parameter}`);
+      return false;
     }
 
     if (max) {
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -59,6 +59,7 @@ import {IAresData} from './ares/IAresData';
 import {Multiset} from './utils/Multiset';
 import {GameSetup} from './GameSetup';
 import {CardLoader} from './CardLoader';
+import {GlobalParameter} from './GlobalParameter';
 
 export type GameId = string;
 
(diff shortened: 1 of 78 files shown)


nwai90 (approved):

src/Game.ts
```diff
@@ -1249,6 +1250,37 @@ export class Game implements ISerializable<SerializedGame> {
...
+    let playerRequirementsBonus: number = player.getRequirementsBonus(this, parameter === GlobalParameter.VENUS);
+
+    if (parameter === GlobalParameter.OCEANS) {
+      currentLevel = this.board.getOceansOnBoard();
+    } else if (parameter === GlobalParameter.OXYGEN) {
+      currentLevel = this.getOxygenLevel();
+    } else if (parameter === GlobalParameter.TEMPERATURE) {
+      currentLevel = this.getTemperature();
+      playerRequirementsBonus *= 2;
+    } else { // parameter === GlobalParameter.VENUS
```
bafolts, line 1272: Instead of the comment I would check that `else if (parameter === GlobalParameter.VENUS)`. In the case we were to hit the final `else` this could log a warning and `return false`. Logging that we were provided an unsupported `GlobalParameter`.

bafolts (changes requested):

src/Game.ts
Lynesth: Good call, I'll do that.

src/cards/base/CloudSeeding.ts
```diff
@@ -17,7 +18,7 @@ export class CloudSeeding implements IProjectCard {
 
     public canPlay(player: Player, game: Game): boolean {
       return player.getProduction(Resources.MEGACREDITS) > -5 &&
-        game.board.getOceansOnBoard() >= 3 - player.getRequirementsBonus(game) &&
+        game.checkMinRequirements(player, GlobalParameter.OCEANS, 3) &&
```
kberg, line 21: I suggest `meetsMinRequirements`.

kberg (approved):
A good change.

bafolts (approved):

src/cards/base/CloudSeeding.ts
Lynesth: Feel free to sed it everywhere :p
I prefer merging this now before @bafolts forces me to fix hundreds of merge conflicts.

PR #2218, opened 2020-12-29
Remove card not found warning for now

Lynesth:


kberg:
Why bother doing this for now?

>

chosta:
Because it's slowing the app because of a great amount of console.warn. Example:
Searching in debug-ui shows 30k warnings since it tries to rerender every element of every card (this is a Vue / React thing actually but console.warn is doing a lot of damage)

kberg:
Well that's a good reason.

>

kberg (approved):
Approve modulo the change I just suggested.

PR #2219, opened 2020-12-29
Heat spending improved

sienmich:


sienmich:
Somethings wrong, I can feel it

PR #2223, opened 2020-12-29
Reimplement Clone

kberg:
What do people think of this kind of object traversal to clone a game? What do you see as major areas to look out for?


server.ts
```diff
@@ -449,11 +450,26 @@ function createGame(req: http.IncomingMessage, res: http.ServerResponse): void {
...
-      res.setHeader('Content-Type', 'application/json');
-      res.write(getGameModelJSON(game));
-      res.end();
+      if (gameOptions.clonedGamedId !== undefined && !gameOptions.clonedGamedId.startsWith('#')) {
+        Cloner.clone(gameId, gameOptions.clonedGamedId, players, firstPlayerIdx, (err, game) => {
+          if (err) {
+            throw err;
+          }
+          if (game === undefined) {
+            throw new Error(`game ${gameOptions.clonedGamedId} not cloned`); // how to nest errs in the way Java nests exceptions?
```
bafolts, line 459: This error will end up being thrown on the callstack from where this callback is called within `Cloner`. If it went uncaught there it would end up as an uncaught exception which would stop the node process. Since we are in `server` it may make more sense to return a 404 status code since there wasn't an error but we most likely didn't find the game. As we didn't get an error some 400 level response code would make sense.

src/database/Cloner.ts
```diff
@@ -0,0 +1,52 @@
...
+    players: Array<Player>,
+    firstPlayerIndex: number,
+    cb: DbLoadCallback<Game>): void {
+    Database.getInstance().loadCloneableGame(sourceGameId, (err, serialized) => {
+      if (err !== undefined) {
+        cb(err, undefined);
+        return;
+      }
+      if (serialized === undefined) {
+        cb(err, undefined);
```
bafolts, line 18: If we are here `err` will be `undefined`. May be easier to read to pass `cb(undefined, undefined)`. This represents that we had no error pulling the data but were unable to find the game.

src/database/Cloner.ts
```diff
@@ -0,0 +1,52 @@
...
+      serialized.players = players.map((player) => player.serialize());
+      serialized.first = serialized.players[firstPlayerIndex];
+
+      const game: Game = Game.deserialize(serialized);
+      cb(undefined, game);
+    });
+  }
+
+  private static replacePlayers(obj: any, ids: Array<string>) {
+    if (obj === undefined || obj === null || typeof obj !== 'object') {
```
bafolts, line 34: I don't think this needs to check that `typeof obj !== 'object'`. The error from node I read claims that only `undefined` and `null` can't be converted.

```
> Object.entries({})
[]
> Object.entries(undefined)
Uncaught TypeError: Cannot convert undefined or null to object
    at Function.entries (<anonymous>)
```

src/database/Cloner.ts
```diff
@@ -0,0 +1,52 @@
...
+      return;
+    }
+    const keys = Object.entries(obj);
+    keys.forEach(([key, val]) => {
+      if (obj.hasOwnProperty(key)) {
+        if (typeof val === 'string') {
```
bafolts, line 40: I think this will work great and it avoids requiring developers to consider how we clone a property when the `PlayerId` is used. I was concerned this wouldn't work with `Array` since an array has number indices and this could attempts to write a string value but it seems to work fine with arrays as well.

```
> Object.entries(["a", "b", "c"])
[ [ '0', 'a' ], [ '1', 'b' ], [ '2', 'c' ] ]
> var a = ["a", "b", "c"]; a["0"] = "z";
'z'
> a
[ 'z', 'b', 'c' ]
```

src/database/Cloner.ts
```diff
@@ -0,0 +1,52 @@
...
+      serialized.id = id;
+      serialized.players = players.map((player) => player.serialize());
+      serialized.first = serialized.players[firstPlayerIndex];
+
+      const game: Game = Game.deserialize(serialized);
+      cb(undefined, game);
+    });
+  }
+
+  private static replacePlayers(obj: any, ids: Array<string>) {
```
bafolts, line 33: This can be `Array<PlayerId>` to help with self documenting. It is possible to use `object` instead of `any` as the type here if the updated property gets written with `Object.defineProperty(obj, key, { value: ids[idx] })`. If the type were `object` then we would have to protect against `null` since `typeof null` is `object` but we wouldn't have to protect against `undefined`. It would also avoid using `any`.

src/database/PostgreSQL.ts
```diff
@@ -180,14 +179,19 @@ export class PostgreSQL implements IDatabase {
...
-  saveGameState(game_id: GameId, save_id: number, game: string, players: number): void {
-    // Insert
-    this.client.query('INSERT INTO games(game_id, save_id, game, players) VALUES($1, $2, $3, $4)', [game_id, save_id, game, players], (err) => {
-      if (err) {
-        // Should be a duplicate, does not matter
-        return;
-      }
-    });
+  saveGame(game: Game): void {
+    // TODO(kberg): why is player size a useful first-class piece of data?
```
bafolts, line 183: I am pretty sure this is only used by the haphazard garbage collection process to determine number of players.

https://github.com/bafolts/terraforming-mars/blob/main/src/database/PostgreSQL.ts#L146

Ideally we would have a separate table of `game_to_player` which mapped game id to player id. This would allow us to calculate this value and it would also allow the `GameLoader` to pull the player and game ids without needing to deserializing the entire game.

server.ts
kberg: There's a catch 17 lines down processing the error. Isn't that what you want?

Also, I'm a little surprised there isn't a global catch around `processRequest()` -- wouldn't that prevent that entire class of issues?

kberg: Rather, why doesn't the catch in `requestHandler` address this problem in its entirety?

src/database/Cloner.ts
kberg: It's mostly a lazy way of dealing with proving to the Typescript compiler that serialized is defined.

src/database/Cloner.ts
kberg: Nice, but is that slightly too clever? Let me know, happy to defer.

src/database/Cloner.ts
kberg: Good check. :+1:

src/database/Cloner.ts
kberg: ... and remove `|| typeof obj !== 'object'` Nice!

kberg: Actually, that doesn't work. By removing 'any' and saying it's an object, the compiler complains with 
```
Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '{}'.
  No index signature with a parameter of type 'string' was found on type '{}'.
```

So in order to perform this extralinguistic behavior, it's got to be an `any`.

src/database/PostgreSQL.ts
kberg: I realize now, it's shown in the CreateGameForm when asking to clone a game.

src/database/Cloner.ts
```diff
@@ -0,0 +1,85 @@
...
+    if (serialized === undefined) {
+      cb(err, undefined);
+      return;
+    }
+
+    const sourceGameId: GameId = serialized.id;
+    const oldPlayerIds: Array<PlayerId> = serialized.players.map((player) => player.id);
+    const newPlayerIds: Array<PlayerId> = players.map((player) => player.id);
+    if (oldPlayerIds.length !== newPlayerIds.length) {
+      throw new Error(`Failing to clone from a ${oldPlayerIds.length} game ${sourceGameId} to a ${newPlayerIds.length} game.`);
```
bafolts, line 29: This needs to call to `cb` with this error. We don't write much asynchronous javascript in the project but in this case the error has to propagate out through `cb`.

bafolts (changes requested):
Spotted once place where error was thrown but `cb` needs to be called. Other comments were minor.

src/Game.ts
```diff
@@ -214,26 +214,23 @@ export class Game implements ISerializable<SerializedGame> {
...
-    const game: Game = new Game(id, players, first, activePlayer, gameOptions, seed, board, dealer);
+    const game: Game = new Game(id, players, firstPlayer, activePlayer, gameOptions, seed, board, dealer);
 
-    // Clone game
-    if (gameOptions.clonedGamedId !== undefined && !gameOptions.clonedGamedId.startsWith('#')) {
-      throw new Error('Clone game disabled temporarily. Should be restored in January.');
-    //   game.cloneGame(gameOptions.clonedGamedId);
-    //   game.clonedGamedId = '#' + gameOptions.clonedGamedId;
-    //   return game;
+    if (gameOptions.clonedGamedId !== undefined) {
```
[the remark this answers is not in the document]
kberg: Yep, good idea.

src/database/Cloner.ts
kberg: I think the rewrite is much more sensible.

bafolts (approved):

Issue #2215, opened 2020-12-29
Card not found in tests


Lynesth:
@bafolts There's 2 vuejs tests that are not able to find cards somehow. It's been like that from the beginning as far as I can tell, and tests are still passing properly but I realized the issue when I added a warning if the card couldn't be retrieved in #2208.

![firefox_2020-12-29_20-36-18](https://user-images.githubusercontent.com/5318258/103274324-84571400-4a15-11eb-81ff-af2bada4e08f.png)

Lynesth:
Ok so we have 2 culprits, both in `src/components/card/Card.ts`.
The first one is its `getCard` method that searches through project cards, then corps, then standard projects.
Second one is the `getCardExpansionByName` function that searches through each type of cards of each manifests (that's at most 36 searches).

Lynesth:
Ok so, last post from me on this matter.
- I removed the warning for now, as it really slows down browsers when there's a lot of cards.
- The component calls `getCard` way too many times, iterating over all cards again and again while this should be done once and saved (when mounted for example ?)
- I'm not sure how to "fix" the warning thing. How I see it (without changing too many things), it's either we simply keep it removed or we make sure that each call to `Deck.findByName` from the component searches through all cards at once.

bafolts:
In regards to the performance, I am going to assume without having measured it that when you take an action on the UI and it seems to respond slowly this is due to us traversing through the countless cards looking for your card. We have plentiful cards now that iterating through each card would be noticeable if you were doing it often. Since the `cardName` is a unique key we should be able to use a map to find the card without needing to iterate through every card. This change should help us on both the server and the client. I will change this to use a `Map` with card name as key for `Deck` to make these look-ups far less CPU intensive.

In regards to the warning. I do think it was helpful as it spotted this area for improvement and it wasn't an error. Searching through every deck at once will address the issue. I will put up a PR with these two changes. It should go a long way for CPU improvement. I will turn the warning back on with the change.

PR #2224, opened 2020-12-29
consolidate card finding to card component

bafolts:
First part of a two part series for #2215 . This part removes `Decks` and the find card operations to the card component where they are used. It simplifies some of the `Deck` by leveraging `ICard` directly as well. Also moves `BeginnerCorporation` to the base manifest so that it doesn't always need to be considered separately. It is filtered out from the dealer as the dealer should never deal it.

### DIFF
--- a/src/Deck.ts
+++ b/src/Deck.ts
@@ -7,15 +7,4 @@ export class Deck<T extends ICard> {
   public findByCardName(name: string): ICardFactory<T> | undefined {
     return this.cards.find((cf) => cf.cardName === name);
   }
-
-  public shuffled(): Deck<T> {
-    const shuffled: Array<ICardFactory<T>> = [];
-    const copy = this.cards.slice();
-    while (copy.length) {
-      shuffled.push(
-        copy.splice(Math.floor(Math.random() * copy.length), 1)[0],
-      );
-    }
-    return new Deck(shuffled);
-  }
 }
--- a/src/Deck.ts
+++ b/src/Deck.ts
@@ -1,51 +1,21 @@
-import {CardName} from './CardName';
+import {ICard} from './cards/ICard';
 import {ICardFactory} from './cards/ICardFactory';
-import {CorporationCard} from './cards/corporation/CorporationCard';
-import {IProjectCard} from './cards/IProjectCard';
-import {PreludeCard} from './cards/prelude/PreludeCard';
-import {StandardProjectCard} from './cards/standardProjects/StandardProjectCard';
 
-export type CardTypes = IProjectCard | CorporationCard | PreludeCard | StandardProjectCard;
+export class Deck<T extends ICard> {
+  constructor(public cards: Array<ICardFactory<T>>) {}
 
-export class Deck<T extends CardTypes> {
-    cards: Array<ICardFactory<T>>;
-
-    constructor(cards: Array<ICardFactory<T>>) {
-      this.cards = cards;
-    }
-
-    public findByCardName(name: string): ICardFactory<T> | undefined {
-      return this.cards.find((cf) => cf.cardName === name);
-    }
-
-    public shuffled(): Deck<T> {
-      const shuffled: Array<ICardFactory<T>> = [];
-      const copy = this.cards.slice();
-      while (copy.length) {
-        shuffled.push(
-          copy.splice(Math.floor(Math.random() * copy.length), 1)[0],
-        );
-      }
-      return new Deck(shuffled);
-    }
-}
-
-export class Decks {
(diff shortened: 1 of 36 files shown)


src/cards/IProjectCard.ts
```diff
@@ -1,15 +1,11 @@
-import {CardType} from './CardType';
 import {ICard} from './ICard';
 import {Player} from '../Player';
 import {Game} from '../Game';
-import {ResourceType} from '../ResourceType';
 import {Resources} from '../Resources';
 
 export interface IProjectCard extends ICard {
     canPlay?: (player: Player, game: Game) => boolean;
-    cardType: CardType;
```
bafolts, line 10: These are defined in `ICard` and redundant.

src/cards/StandardCardManifests.ts
```diff
@@ -371,6 +372,7 @@ export const BASE_CARD_MANIFEST = new CardManifest({
     {cardName: CardName.ZEPPELINS, Factory: Zeppelins},
   ],
   corporationCards: [
+    {cardName: CardName.BEGINNER_CORPORATION, Factory: BeginnerCorporation},
```
bafolts, line 375: Would not be surprised if someone playing as beginner corporation was ending up in a situation where their card was not found. This simplified the look-up logic as this doesn't have to be called out separately.

src/cards/community/AerospaceMission.ts
```diff
@@ -1,14 +1,13 @@
...
 import {Player} from '../../Player';
 import {PreludeCard} from '../prelude/PreludeCard';
-import {IProjectCard} from '../IProjectCard';
 import {CardName} from '../../CardName';
 import {Game} from '../../Game';
 import {BuildColony} from '../../deferredActions/BuildColony';
 import {CardMetadata} from '../CardMetadata';
 import {CardRenderer} from '../render/CardRenderer';
 
-export class AerospaceMission extends PreludeCard implements IProjectCard {
```
bafolts, line 11: These are all redundant. The `PreludeCard` always extends `IProjectCard`.

src/cards/corporation/CorporationCard.ts
```diff
@@ -3,38 +3,17 @@ import {ICard} from '../ICard';
...
-import {SelectOption} from '../../inputs/SelectOption';
-import {SelectSpace} from '../../inputs/SelectSpace';
-import {CardType} from '../CardType';
 
 export interface CorporationCard extends ICard {
     initialActionText?: string;
     initialAction?: (player: Player, game: Game) => PlayerInput | undefined;
     startingMegaCredits: number;
     cardCost?: number;
-    play: (
```
bafolts, line 18: These are redundant and defined on `ICard`.

src/components/card/Card.ts
```diff
@@ -77,12 +37,39 @@ export const Card = Vue.component('card', {
       type: Boolean,
     },
   },
+  data: function() {
+    let cardInstance: ICard | undefined;
```
bafolts, line 41: @Lynesth this will store the card once when the data function runs which should only run once. Should avoid constantly having to find the value. Consolidated finding the expansion and card to this one loop.

src/Deck.ts
```diff
@@ -1,51 +1,21 @@
...
-    // currently flooding player's console. console.warn(`card not found ${cardName}`);
-    return undefined;
+  public findByCardName(name: string): ICardFactory<T> | undefined {
+    return this.cards.find((cf) => cf.cardName === name);
   }
 
-  public static allCardNames(decks: Array<Deck<CardTypes>>): Array<CardName> {
-    const arrays: Array<Array<CardName>> = decks.map((deck) => deck.cards.map((cf) => cf.cardName));
-    return ([] as Array<CardName>).concat(...arrays);
+  public shuffled(): Deck<T> {
```
Lynesth, line 11: This does not seem to be used anywhere anymore.

Lynesth:
You mentioned earlier using a map instead to search for the card, which is a great idea as it would be much faster.
It'd be even better now that we are only using the card name almost everywhere.

Could we change `Deck.cards` from
`Array<{cardName: CardName, Factory: new () => T, compatibility ?: GameModule}>`
to
`Map<CardName, {Factory: new () => T, compatibility ?: GameModule}>`
?

bafolts:
@Lynesth I still plan to do the change to use a `Map` but want to use a separate PR. I removed the `shuffled` which wasn't being used.

src/Deck.ts
bafolts: Sweet. One less shuffle function.

kberg (approved):
This looks great. The card rework I started on back in Septebmer has many flaws, I'm glad to see you making some of it better.

PR #2226, opened 2020-12-30
use getters for corporation cards

bafolts:
Been trying to work around a problem a few other ways in regards to all of our cards creating new values for their properties. With a moment of inspiration when @Lynesth mentioned creating a separate interface for dynamic properties I remembered that when using a `get` for a property we don't have to store these values. With this change the values which are static such as `name`, `tags`, `cost`, and `metadata` wont store a new value on each instance but will instead provide the value at runtime.

The trade off here is the generated javascript will be slightly larger per card and some operations will run marginally slower as the value will be calculated when it is needed versus stored in memory. As a dealer creates an instance of most cards per game I think this trade off is well worth it.

I will do this roll out in batches. This starts with the corporation cards.

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -2085,7 +2085,7 @@ export class Player implements ISerializable<SerializedPlayer> {
     player.actionsThisGeneration = new Set<CardName>(d.actionsThisGeneration);
 
     if (d.pickedCorporationCard !== undefined) {
-      player.pickedCorporationCard = cardFinder.getCorporationCardByName(typeof d.pickedCorporationCard === 'string' ? d.pickedCorporationCard : d.pickedCorporationCard.name);
+      player.pickedCorporationCard = cardFinder.getCorporationCardByName(d.pickedCorporationCard);
     }
 
     // Rebuild corporation card
--- a/src/SerializedPlayer.ts
+++ b/src/SerializedPlayer.ts
@@ -4,9 +4,7 @@ import {Color} from './Color';
 import {SerializedCard} from './SerializedCard';
 import {VictoryPointsBreakdown} from './VictoryPointsBreakdown';
 import {SerializedTimer} from './SerializedTimer';
-import {CorporationCard} from './cards/corporation/CorporationCard';
 
-// TODO(kberg): remove reference to CorporationCard by 2021-01-15
 export interface SerializedPlayer {
     actionsTakenThisRound: number;
     actionsThisGeneration: Array<CardName>;
@@ -39,7 +37,7 @@ export interface SerializedPlayer {
     name: string;
     needsToDraft: boolean | undefined;
     oceanBonus: number;
-    pickedCorporationCard: CardName | CorporationCard | undefined;
+    pickedCorporationCard: CardName | undefined;
     plantProduction: number;
     plants: number;
     plantsNeededForGreenery: number;
--- a/src/cards/Card.ts
+++ b/src/cards/Card.ts
@@ -1,0 +1,51 @@
+
+import {CardMetadata} from './CardMetadata';
+import {CardName} from '../CardName';
+import {CardType} from './CardType';
+import {Tags} from './Tags';
+
+/**
+ * Utility function to pull the static instance
+ * of a class.
+ * @param {Card} instance of a card
+ * @return {Object} static instance of card
+ */
+function staticInstance(instance: Card) {
+  return Object.getPrototypeOf(instance).constructor;
+}
+
+interface StaticCardProperties {
+  cardType: CardType;
+  initialActionText?: string;
+  metadata: CardMetadata;
+  name: CardName;
+  startingMegaCredits?: number;
+  tags?: Array<Tags>;
+}
+
+export abstract class Card {
+  constructor(properties: StaticCardProperties) {
+    const instance = staticInstance(this);
+    if (instance.properties === undefined) {
+      instance.properties = properties;
+    }
+  }
+  public get cardType() {
+    return staticInstance(this).properties.cardType;
+  }
+  public get initialActionText() {
+    return staticInstance(this).properties.initialActionText;
+  }
+  public get metadata() {
+    return staticInstance(this).properties.metadata;
+  }
+  public get name() {
+    return staticInstance(this).properties.name;
+  }
+  public get startingMegaCredits() {
+    return staticInstance(this).properties.startingMegaCredits;
+  }
+  public get tags() {
+    return staticInstance(this).properties.tags;
+  }
+}
--- a/src/cards/Card.ts
+++ b/src/cards/Card.ts
@@ -38,9 +38,9 @@ export abstract class Card {
     return this.properties.name;
   }
   public get startingMegaCredits() {
-    return this.properties.startingMegaCredits || 0;
+    return this.properties.startingMegaCredits === undefined ? 0 : this.properties.startingMegaCredits;
   }
   public get tags() {
-    return this.properties.tags || [];
+    return this.properties.tags === undefined ? [] : this.properties.tags;
   }
 }
--- a/src/cards/Card.ts
+++ b/src/cards/Card.ts
@@ -4,16 +4,6 @@ import {CardName} from '../CardName';
 import {CardType} from './CardType';
 import {Tags} from './Tags';
 
-/**
- * Utility function to pull the static instance
- * of a class.
- * @param {Card} instance of a card
- * @return {Object} static instance of card
- */
-function staticInstance(instance: Card) {
-  return Object.getPrototypeOf(instance).constructor;
-}
-
 interface StaticCardProperties {
   cardType: CardType;
   initialActionText?: string;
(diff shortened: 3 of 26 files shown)


Lynesth:
This is a great idea to fix the memory issue.
The overhead from using functions instead of properties should be negligible though.

src/Player.ts
```diff
@@ -2085,7 +2085,7 @@ export class Player implements ISerializable<SerializedPlayer> {
     player.actionsThisGeneration = new Set<CardName>(d.actionsThisGeneration);
 
     if (d.pickedCorporationCard !== undefined) {
-      player.pickedCorporationCard = cardFinder.getCorporationCardByName(typeof d.pickedCorporationCard === 'string' ? d.pickedCorporationCard : d.pickedCorporationCard.name);
+      player.pickedCorporationCard = cardFinder.getCorporationCardByName(d.pickedCorporationCard);
```
bafolts, line 2088: Had to update this. The JSON version no longer includes the getter properties so the backwards compatible unit test was failing. No longer needs to be backwards compatible.

src/cards/ICard.ts
```diff
@@ -46,6 +46,6 @@ export interface ICard {
     resourceCount?: number;
     cost?: number;
     cardType: CardType;
-    metadata?: CardMetadata;
+    metadata: CardMetadata;
```
bafolts, line 49: I changed this to be required. While making these changes if I were to have a typo and not spell `metadata` properly it may go missed.

tests/cards/ares/BioengineeringEnclosure.spec.ts
```diff
@@ -1,37 +1,19 @@
...
-import {CardType} from '../../../src/cards/CardType';
-import {ResourceType} from '../../../src/ResourceType';
 import {expect} from 'chai';
 import {ARES_OPTIONS_NO_HAZARDS} from '../../ares/AresTestHelper';
 import {TestPlayers} from '../../TestingUtils';
 
 describe('BioengineeringEnclosure', function() {
   let card : BioengineeringEnclosure; let player : Player; let game : Game;
-
-  const scienceTagCard: IProjectCard = {
```
bafolts, line 16: These cards being created were missing `metadata` property. For simplicity I put in a card which should still pass the tests.

src/SerializedPlayer.ts
```diff
@@ -4,9 +4,7 @@ import {Color} from './Color';
 import {SerializedCard} from './SerializedCard';
 import {VictoryPointsBreakdown} from './VictoryPointsBreakdown';
 import {SerializedTimer} from './SerializedTimer';
-import {CorporationCard} from './cards/corporation/CorporationCard';
 
-// TODO(kberg): remove reference to CorporationCard by 2021-01-15
```
kberg, line 9: /me nods

src/cards/ICard.ts
kberg: Agree. It's the right time for it.

src/cards/community/ScienceTagCard.ts
```diff
@@ -2,14 +2,25 @@ import {IProjectCard} from '../IProjectCard';
...
 export class ScienceTagCard implements IProjectCard {
-    public cost = 0;
-    public tags = [Tags.SCIENCE];
-    public name = CardName.SCIENCE_TAG_BLANK_CARD;
-    public cardType = CardType.PROXY;
-
-    public play() {
-      return undefined;
-    }
+  public get cost() {
```
kberg, line 8: You're going to do this en masse with sed or something, yes? If you want I can help with that.

src/cards/corporation/BeginnerCorporation.ts
```diff
@@ -4,26 +4,35 @@ import {Player} from '../../Player';
...
-    public tags = [];
-    public startingMegaCredits: number = 42;
-    public name = CardName.BEGINNER_CORPORATION;
-    public cardType = CardType.CORPORATION;
-    public play(player: Player, game: Game) {
-      for (let i = 0; i < 10; i++) {
-        player.cardsInHand.push(game.dealer.dealCard());
-      }
-      return undefined;
+  public get tags() {
```
kberg, line 10: What if this was all done in the constructor?

eg

constructor() {
  super(CardName.BEGINNER_CORPORATION, CardType.CORPORATION, 42, [])
}

kberg (approved):

src/cards/community/ScienceTagCard.ts
bafolts: I can do en masse or a chunk at a time. I live in VIM and my terminal so it would be something of that nature. Or my hands will get their workout for the day. I may do a chunk at a time for ease of reviewing. Considering the unrelated changes I had to change for this I may run into other minor changes along the way. I will probably start with a letter a day and if I don't hit road blocks then do the rest.

src/cards/corporation/BeginnerCorporation.ts
bafolts: This is all done in the constructor now. It has to rely on some javascript to access the static instance of a card but this should allow for easier to read and write code on our end. This is now done with:

```
constructor() {
  super({
    name: CardName.BEGINNER_CORPORATION,
    startingMegaCredits: 42
  });
}
```

The underlying `Card` base class is orchestrating how this works behind the scenes. It is creating the dynamic static properties and using those for the getters so it is always responsible for them. This will be even less memory as only `Card.prototype` will have the getters. The object being passed in to `super` should get pushed onto the stack since it is an argument and avoid needing the garbage collector to clean it up.

src/cards/Card.ts
```diff
@@ -0,0 +1,51 @@
...
+import {Tags} from './Tags';
+
+/**
+ * Utility function to pull the static instance
+ * of a class.
+ * @param {Card} instance of a card
+ * @return {Object} static instance of card
+ */
+function staticInstance(instance: Card) {
+  return Object.getPrototypeOf(instance).constructor;
```
kberg, line 14: Call once and store?

kberg (changes requested):
Pulling back approval for the moment since this is in flux. But this is better.

src/cards/Card.ts
```diff
@@ -0,0 +1,46 @@
...
+  startingMegaCredits?: number;
+  tags?: Array<Tags>;
+}
+
+export abstract class Card {
+  private readonly properties: StaticCardProperties;
+  constructor(properties: StaticCardProperties) {
+    const staticInstance = Object.getPrototypeOf(this).constructor;
+    if (staticInstance.properties === undefined) {
+      if (properties.cardType === CardType.CORPORATION && properties.startingMegaCredits === undefined) {
```
bafolts, line 21: @kberg the properties are now stored and re-used between each instance. I added this runtime type checker as with this change typescript had a new compilation error that `startingMegaCredits` being `number | undefined` does not match `number`. This should always be provided for a corporation card so this checks here at runtime the first time the constructor is used.

kberg: Excellent.

src/cards/Card.ts
```diff
@@ -0,0 +1,46 @@
...
+    return this.properties.metadata;
+  }
+  public get name() {
+    return this.properties.name;
+  }
+  public get startingMegaCredits() {
+    return this.properties.startingMegaCredits || 0;
+  }
+  public get tags() {
+    return this.properties.tags || [];
```
kberg, line 44: I'm not clear whether || is accepted or not. I thought it wasn't.

src/cards/corporation/BeginnerCorporation.ts
```diff
@@ -2,28 +2,30 @@
...
+      cardType: CardType.CORPORATION,
+      name: CardName.BEGINNER_CORPORATION,
+      metadata: {
+        cardNumber: 'R00',
+        description: 'You start with 42 MC. Instead of choosing from 10 cards during setup, you get 10 cards for free.',
+        renderData: CardRenderer.builder((b) => {
+          b.megacredits(42).nbsp.cards(10).digit;
+        }),
+      },
+      startingMegaCredits: 42,
```
kberg, line 22: Move this up one line above metadata?

src/cards/corporation/CrediCor.ts
```diff
@@ -1,44 +1,46 @@
...
+        }),
+      },
+    });
+  }
+  public onCardPlayed(player: Player, _game: Game, card: IProjectCard) {
+    if (player.corporationCard !== undefined && player.corporationCard.name === this.name && card.cost >= 20) {
+      player.megaCredits += 4;
+    }
+  }
+  public onStandardProject(player: Player, project: StandardProjectCard) {
```
kberg, line 38: Will the new standard project card behavior call onCardPlayed?

kberg: Why does onCardPlayed check the corporation card and this does not.

src/cards/corporation/EcoLine.ts
```diff
@@ -1,36 +1,39 @@
...
-        b.corpBox('effect', (ce) => {
-          ce.effectBox((eb) => {
-            eb.plants(7).digit.startAction.greenery();
-            eb.description('Effect: You may always pay 7 plants, instead of 8, to place greenery.');
+export class EcoLine extends Card implements CorporationCard {
+  constructor() {
+    super({
+      name: CardName.ECOLINE,
+      tags: [Tags.PLANT],
+      startingMegaCredits: 36,
```
kberg, line 15: Ongoing, can you endeavor to keep the fields in the same order?

I would even ask for a blank line between everything and metadata. If the compiler dislikes that, add // ?

tests/cards/ares/BioengineeringEnclosure.spec.ts
kberg: Aw. I liked having mocks.

kberg (approved):

src/cards/corporation/CrediCor.ts
bafolts: The `player` argument provided should be the player that played the card.

https://github.com/bafolts/terraforming-mars/blob/main/src/cards/standardProjects/StandardProjectCard.ts#L49

This should not call to `onCardPlayed` as the card is not played and added to the player's hand, the action on the card is only taken.

As we are _assuming_ that `player` is going to be the player that played the standard project and it is never good to assume I will update `onStandardProject` to call `this.onCardPlayed` so that we don't have to duplicate the check and we no longer have to hope the caller only calls us when `player` is what we want.

src/cards/Card.ts
bafolts: It is _accepted_ but I can change to use ternary as that is probably easier for everyone to read regardless of what language they are coming from.

tests/cards/ares/BioengineeringEnclosure.spec.ts
bafolts: Can still have mocks....just have to provide some value for `metadata` now which I was too lazy to come up with.

PR #2228, opened 2020-12-30
Fixes bug that loaded both ares and normal cards to the game.

sienmich:


bafolts (approved):

src/CardLoader.ts
```diff
@@ -52,11 +52,11 @@ export class CardLoader {
     };
   }
 
-  private addToDeck<T extends CardTypes>(deck: Array<T>, cards: Deck<T>): void {
-    const cardInstances = cards.cards
+  private addDeck<T extends CardTypes>(cards: Array<T>, deck: Deck<T>): void {
```
kberg, line 55: Thanks for fixing this.

src/cards/ares/AresCardManifest.ts
```diff
@@ -52,7 +52,7 @@ export const ARES_CARD_MANIFEST = new CardManifest({
     {cardName: CardName.RESTRICTED_AREA_ARES, Factory: RestrictedAreaAres},
     {cardName: CardName.SOLAR_FARM, Factory: SolarFarm},
   ],
-  projectCardsToRemove: [
+  cardsToRemove: [
```
kberg, line 55: Is this also going to subtitute corporation cards? If not, this should preserve its original name.

sienmich: It is, that's why I changed the name :)

PR #2230, opened 2020-12-30
Remove Player from serialized milestomes and awards.

kberg:
Remove Player from serialized milestomes and awards.

This reduces the example game I have been using from its original size
of 1.950M bytes to 340K to 27.5K

This is database storage, mind you.


kberg:
This ends my serialization work.

src/awards/FundedAward.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export function deserializeFundedAwards(
+  fundedAwards: Array<SerializedFundedAward>,
+  players: Array<Player>,
+  awards: Array<IAward>): Array<FundedAward> {
+  return fundedAwards.map((element: SerializedFundedAward) => {
+    const awardName = element.award?.name !== undefined ? element.award.name : element.name;
+    if (awardName === undefined) {
+      throw new Error('Milestone name not found');
```
nwai90, line 33: This should probably be `Award name not found`

src/awards/FundedAward.ts
```diff
@@ -0,0 +1,54 @@
...
+    }
+    const award: IAward | undefined = awards.find((award) => award.name === awardName);
+    if (award === undefined) {
+      throw new Error(`Award ${awardName} not found when rebuilding Claimed Milestone`);
```
nwai90, line 37: Similarly for these, should they be `Award ${awardName} not found when rebuilding Funded Award`?

kberg: Ah, yes, thanks.

src/awards/FundedAward.ts
```diff
@@ -0,0 +1,54 @@
...
+import {Player, PlayerId} from '../Player';
+
+export interface FundedAward {
+  award: IAward;
+  player: Player;
+}
+
+// TODO(kberg): remove references to award and player by 2020-02-01
+export interface SerializedFundedAward {
+  name?: string;
```
bafolts, line 11: With all the enums we have in place we don't have `AwardName`. Good to know.

src/awards/FundedAward.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (playerId === undefined) {
+      throw new Error(`Player ID not found when rebuilding Funded Award ${awardName}`);
+    }
+    const player = players.find((player) => player.id === playerId);
+    if (player === undefined) {
+      throw new Error(`Player ${playerId} not found when rebuilding Funded Award ${awardName}`);
+    }
+
+    return {
+      award: award,
```
bafolts, line 50: Fun fact, typescript will allow you to do `return { award, player }` here.

kberg: OK made the change.

bafolts (approved):
Looks good.

PR #2248, opened 2021-01-01
More cards modified to use less memory.

bafolts:
This introduces the change to extend `Card` for less memory consumption per instance of a card. This is for base cards starting with letter A through P.

### DIFF
--- a/src/cards/Card.ts
+++ b/src/cards/Card.ts
@@ -2,31 +2,52 @@
 import {CardMetadata} from './CardMetadata';
 import {CardName} from '../CardName';
 import {CardType} from './CardType';
+import {IAdjacencyBonus} from '../ares/IAdjacencyBonus';
+import {ResourceType} from '../ResourceType';
 import {Tags} from './Tags';
 
 interface StaticCardProperties {
+  adjacencyBonus?: IAdjacencyBonus;
   cardType: CardType;
+  cost?: number;
+  hasRequirements?: boolean;
   initialActionText?: string;
   metadata: CardMetadata;
   name: CardName;
+  resourceType?: ResourceType;
   startingMegaCredits?: number;
   tags?: Array<Tags>;
 }
 
+const staticCardProperties = new Map<CardName, StaticCardProperties>();
+
 export abstract class Card {
   private readonly properties: StaticCardProperties;
   constructor(properties: StaticCardProperties) {
-    const staticInstance = Object.getPrototypeOf(this).constructor;
-    if (staticInstance.properties === undefined) {
+    let staticInstance = staticCardProperties.get(properties.name);
+    if (staticInstance === undefined) {
       if (properties.cardType === CardType.CORPORATION && properties.startingMegaCredits === undefined) {
         throw new Error('must define startingMegaCredits for corporation cards');
       }
-      staticInstance.properties = properties;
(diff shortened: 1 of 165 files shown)


kberg:
That is A LOT of cards, my friend, with non-trivial differences in the diff. Any tips on how to make the UI do what I need? Or any chance you can break that up?

bafolts:
Most of the diffs for the cards are only moving properties in to the call to super. Going to have to update hundreds of cards. Would need the UI to be smart enough to spot that the diff only involved moving the data. In lieu of the UI being able to spot that the code only moved how many cards in a batch would be preferred?

kberg:
Actually hang out I think I have an idea.

src/cards/Card.ts
```diff
@@ -2,32 +2,53 @@
...
+    this.properties = staticInstance;
+  }
+  public get adjacencyBonus() {
+    return this.properties.adjacencyBonus;
   }
   public get cardType() {
     return this.properties.cardType;
   }
+  public get cost() {
+    return this.properties.cost === undefined ? 0 : this.properties.cost;
```
kberg, line 47: See, this seems like a good place for `this.properties.cost || 0`. What do you think?

kberg (commented):
56 files reviewed, 109 to go.

src/cards/Card.ts
bafolts: Works for me. I am not too caught up on ternary or using `||` here. Six of one half a dozen of the other. Not having to check for `undefined` always better in my book.

kberg (commented):
110 down, 55 to go.

tests/ares/AresHandler.spec.ts
```diff
@@ -57,8 +59,8 @@ describe('AresHandler', function() {
     // Even though there's already a game, with a board, that laid out hazards, this is going to use a clean set-up.
 
     const deck = game.dealer.deck;
-    deck[deck.length - 1].cost = 5;
-    deck[deck.length - 2].cost = 3;
+    deck.push(new EnergyTapping());
```
kberg, line 62: The comment up top talks about how these cards are chosen for their predetermined costs. That was forced in the earlier version of the test, and now it's happenstance, and not well documented.

Can you please add assertions that those cards have cost 5 and 3? If not, I'd really prefer to restore something hand-crafted.

kberg (changes requested):
That last one!

kberg (approved):
I can do that last one in another PR. I'd rather unblock you.

bafolts:
This went out with the deployment today. I usually look at the memory charts every few hours to see if the server is still up. It has been improving over the past few weeks. With this change I expect the slow of the chart to decrease. Yesterday it took 12 hours before the memory climbed enough to get over 60% which seems to be where the heroku dyno crashes and the dyno runs `npm run start` again and we are back and running. If this change stretches that out to a full 24 hours to line up with when I normally deploy we will be in great shape. Besides the log messages I don't know what else would be climbing in memory with each game after this change. From what I saw with the log messages the node javascript should be using string interning so those messages should mostly end up as pointers to objects which shouldn't be more than a few bytes large. Thanks for approving! I will keep the subsequent cards to come to less than 26 at a time.

PR #2252, opened 2021-01-02
Add a local filesystem for storing game results. Meant as a new development option.

kberg:
Need to restore to an old version? Copy the history entry!
Need to know what happened in between saves? Run `diff`!
Want to query details about a savepoint? Use `jq`!

### DIFF
--- a/src/database/LocalFilesystem.ts
+++ b/src/database/LocalFilesystem.ts
@@ -55,6 +55,7 @@ export class Localfilesystem implements IDatabase {
   getGames(cb: (err: any, allGames: Array<GameId>) => void) {
     const gameIds: Array<GameId> = [];
 
+    // TODO(kberg): use readdir since this is expected to be async anyway.
     fs.readdirSync(dbFolder, {withFileTypes: true}).forEach((dirent: Dirent) => {
       if (!dirent.isFile()) {
         return;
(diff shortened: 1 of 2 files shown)


src/database/LocalFilesystem.ts
```diff
@@ -0,0 +1,91 @@
...
+    }
+  }
+  getClonableGames(cb: (err: any, allGames: Array<IGameData>) => void) {
+    cb(undefined, []);
+  }
+
+  getGames(cb: (err: any, allGames: Array<GameId>) => void) {
+    const gameIds: Array<GameId> = [];
+
+    fs.readdirSync(dbFolder, {withFileTypes: true}).forEach((dirent: Dirent) => {
```
bafolts, line 58: I would modify this to use `readdir` instead of `readdirSync`. Using `Sync` will block thread. As there is already a callback being provided using the non sync method shouldn't be too difficult.

bafolts (approved):
I would recommend avoiding `Sync` calls when possible or it is easy. The postgres and sqlite database implementations are already making asynchronous calls places assuming they are synchronous. As this is only meant for development it shouldn't be a huge deal. I think node uses `Sync` in the name trying to scare people away from using it. They need to go further and name it `readdirSYNC`. Javascript is single threaded so anytime `Sync` is used no other code will be executing.

src/database/LocalFilesystem.ts
kberg: I'll follow up with a TODO. This can go through a couple of rounds of sophistication. But as-is, this would have been so useful while debugging serialization issues in the past.

PR #2253, opened 2021-01-02
Pad save ids in the localfilesystem store.

kberg:


src/database/LocalFilesystem.ts
```diff
@@ -20,12 +20,18 @@ export class Localfilesystem implements IDatabase {
     }
   }
 
+  _pad(saveId: number, z = '0') {
```
bafolts, line 23: Probably simpler to use built-in function https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart

bafolts (changes requested):
Using the built-in `String` method will avoid having to maintain our own implementation.

bafolts (approved):

PR #2255, opened 2021-01-02
Start the database as part of server startup, not waiting for the first operation.

kberg:
This will catch database initialization errors earlier, plus. in theory, once the port is open for traffic, it's ready to serve data.

I could see adding a vestigial call to preload game IDs / initialize the GameLoader. Your thoughts?

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -564,6 +564,7 @@ function serveStyles(req: http.IncomingMessage, res: http.ServerResponse): void 
 }
 
 console.log('Starting server on port ' + (process.env.PORT || 8080));
+Database.getInstance();
 console.log('version 0.X');
 
 server.listen(process.env.PORT || 8080);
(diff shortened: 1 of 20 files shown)


server.ts
```diff
@@ -564,6 +564,7 @@ function serveStyles(req: http.IncomingMessage, res: http.ServerResponse): void
 }
 
 console.log('Starting server on port ' + (process.env.PORT || 8080));
+Database.getInstance();
```
bafolts, line 567: I would change this to something that shows the intentions of catching errors with connecting to database. As written a year from now one could wonder why we had this call to get the instance that we never used. With the try...catch we would throw an error and future reader also knows what intentions were.

```
try {
  Database.getInstance():
} catch (exception) {
  throw new Error('unable to connect to database');
}
```

kberg: Done.

bafolts (approved):

PR #2256, opened 2021-01-02
Add the ability for a card to need to be compatible with more than one other module.

kberg:
I had hoped not to start putting behavior specific to the moon yet, but this has been overwritten once already (as code moved from Dealer to CardLoader so I thought I'd get it in now.

### DIFF
--- a/src/CardLoader.ts
+++ b/src/CardLoader.ts
@@ -38,19 +38,27 @@ export class CardLoader {
 
   private static include(gameOptions: GameOptions) {
     return function(cf: ICardFactory<ICard>): boolean {
-      const expansion = cf.compatibility;
-      switch (expansion) {
-      case undefined:
+      if (cf.compatibility === undefined) {
         return true;
-      case GameModule.Venus:
-        return gameOptions.venusNextExtension;
-      case GameModule.Colonies:
-        return gameOptions.coloniesExtension;
-      case GameModule.Turmoil:
-        return gameOptions.turmoilExtension;
-      default:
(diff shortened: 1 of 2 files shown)


src/CardLoader.ts
```diff
@@ -38,19 +38,27 @@ export class CardLoader {
...
-      case GameModule.Colonies:
-        return gameOptions.coloniesExtension;
-      case GameModule.Turmoil:
-        return gameOptions.turmoilExtension;
-      default:
-        throw ('Unhandled expansion type: ' + expansion);
       }
+      const expansions: Array<GameModule> = Array.isArray(cf.compatibility) ? cf.compatibility : [cf.compatibility];
+      let meets = true;
+      expansions.forEach((expansion) => {
```
bafolts, line 46: Instead of using `meets` I think this can make use of `every`. In this situation we need every game option to provide the expansion.

```
expansions.every((expansion) => {
  switch (expansion) {
    case GameModule.Venus:
      return gameOptions.venusNextExtension;
 // you know the rest
 ```

bafolts (changes requested):
I would leverage `Array.prototype.every`. There are a lot of useful functions on `Array.prototype`.

src/CardLoader.ts
kberg: Nice! Done.

bafolts (approved):

bafolts:
Your hand-built every was impressive! Thought we would get to use the `&=` operator before I saw it was `every`.

PR #2266, opened 2021-01-04
use Map for efficient card finding

bafolts:
Finish the work for #2215 . Uses a `Map` and `Set` within `CardManifest` and `Deck` to make the look-up more efficient.

### DIFF
--- a/src/cards/AllCards.ts
+++ b/src/cards/AllCards.ts
@@ -28,8 +28,11 @@ export const ALL_CARD_MANIFESTS: Array<CardManifest> = [
 ];
 
 function allCardNames(decks: Array<Deck<ICard>>): Array<CardName> {
-  const arrays: Array<Array<CardName>> = decks.map((deck) => deck.cards.map((cf) => cf.cardName));
-  return ([] as Array<CardName>).concat(...arrays);
+  const cardNames: Array<CardName> = [];
+  for (const deck of decks) {
+    deck.factories.forEach((cf) => cardNames.push(cf.cardName));
+  }
+  return cardNames;
 }
 
 export const MANIFEST_BY_MODULE: Map<GameModule, CardManifest> =
@@ -38,7 +41,7 @@ export const MANIFEST_BY_MODULE: Map<GameModule, CardManifest> =
 export const ALL_PROJECT_DECKS = ALL_CARD_MANIFESTS.map(
   (deck) => deck.projectCards,
 );
-export const ALL_CORPORATION_DECKS = ALL_CARD_MANIFESTS.map(
+const ALL_CORPORATION_DECKS = ALL_CARD_MANIFESTS.map(
   (deck) => deck.corporationCards,
 );
 export const ALL_PRELUDE_DECKS = ALL_CARD_MANIFESTS.map(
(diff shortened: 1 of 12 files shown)


src/cards/AllCards.ts
```diff
@@ -28,8 +28,11 @@ export const ALL_CARD_MANIFESTS: Array<CardManifest> = [
 ];
 
 function allCardNames(decks: Array<Deck<ICard>>): Array<CardName> {
-  const arrays: Array<Array<CardName>> = decks.map((deck) => deck.cards.map((cf) => cf.cardName));
-  return ([] as Array<CardName>).concat(...arrays);
+  const cardNames: Array<CardName> = [];
+  for (const deck of decks) {
+    deck.factories.forEach((cf) => cardNames.push(cf.cardName));
+  }
```
Lynesth, line 34: How about making it a `Set` ?
```ts
return new Set([decks.flatMap((deck) => deck.map((cf) => cf.cardName))]);
```

Or even a `Map<CardName, ICardFactory>` so that when searching you directly use that one instead of going through the other 4 ?

src/cards/AllCards.ts
```diff
@@ -38,7 +41,7 @@ export const MANIFEST_BY_MODULE: Map<GameModule, CardManifest> =
 export const ALL_PROJECT_DECKS = ALL_CARD_MANIFESTS.map(
   (deck) => deck.projectCards,
 );
-export const ALL_CORPORATION_DECKS = ALL_CARD_MANIFESTS.map(
```
kberg, line 41: since the rest are exported, why not keep it available? Very minor, ignorable.

src/components/CorporationsFilter.ts
```diff
@@ -10,15 +10,22 @@ function corpCardNames(module: GameModule): Array<CardName> {
     console.log('manifest %s not found', manifest);
     return [];
   } else {
-    return manifest.corporationCards.cards.map((cf) => cf.cardName)
-      .filter((cardName) => cardName !== CardName.BEGINNER_CORPORATION);
+    const cards: Array<CardName> = [];
+    manifest.corporationCards.factories.forEach((cf) => {
```
kberg, line 14: Why is this better than filter/map?

kberg (approved):
LG, handing off to Lynseth as he had some questions.

src/components/CorporationsFilter.ts
Lynesth: We were doing 2 loops before, there's only 1 one.

Lynesth: It still could be "one-lined" to
```ts
const cards: Array<CardName> = manifest.corporationCards.factories.flatMap((cf) =>
  cf.cardName !== CardName.BEGINNER_CORPORATION ? cf.cardName : []
);
```
But it's find as it is now.

bafolts: Since `factories` is now of type `Map` and not `Array` all of the `Array` functions are not available. As far as I know there is only `forEach` with `Map`.

src/cards/AllCards.ts
bafolts: These lists are mostly used by `DebugUI` and `CorporationsFilter` where they are iterated through as `Array`. With `factories` now being `Map` the major benefit should be that calls to `Deck::findCardByName` should go from `O(n)` to `O(1)`. If you have 20 cards in your hand prior to this change we would have searched through the deck until we found your card twenty times. Now the map will find the card in constant time.

PR #2272, opened 2021-01-04
Reorganizing how solo play sets up. Dropping some reliance on game during set-up.

kberg:


src/Game.ts
```diff
@@ -231,24 +231,28 @@ export class Game implements ISerializable<SerializedGame> {
 
     const activePlayer = firstPlayer.id;
 
+    // Single player game player starts with 14TR
+    if (players.length === 1) {
```
nwai90, line 235: Suggestion: We could replace all instances of `if (players.length === 1)` and `if (this.players.length === 1)` with `if (this.isSoloMode())`

src/Game.ts
```diff
@@ -267,6 +271,12 @@ export class Game implements ISerializable<SerializedGame> {
       game.turmoil = Turmoil.newInstance(game);
     }
 
+    // and 2 neutral cities and forests on board
+    if (players.length === 1) {
```
nwai90, line 275: As above

src/Game.ts
kberg: NB this occurs prior to Game's construction . I did that on purpose, FWIW.

src/Game.ts
```diff
@@ -231,24 +231,28 @@ export class Game implements ISerializable<SerializedGame> {
 
     const activePlayer = firstPlayer.id;
 
+    // Single player game player starts with 14TR
+    if (players.length === 1) {
+      gameOptions.draftVariant = false;
```
nwai90, line 236: This logic could be moved to `GameSetup.initializeSoloGame` or equivalent

kberg:
I like that idea.

>

src/Game.ts
```diff
@@ -1349,6 +1357,12 @@ export class Game implements ISerializable<SerializedGame> {
     });
   }
 
+  public simpleAddTile(player: Player, space: ISpace, tile: ITile) {
```
Lynesth, line 1360: Why move that out of the above method ?

kberg:
I'm trying to remove everything that doesn't need a fully-formed game above the actual create - I'd like to find a way to add more common data fully-formed to the constructor, rather than setting them post-construction.

bafolts (approved):
This seems good enough to merge. There were comments but no one has any requested changes. I will merge this. As the amount of code moving with this project is growing I would like to set standard that if there is something with the pull request that seems enough of an issue to not merge people use the "Request changes" versus only "Comment". I take "Comment" as general comments not worth blocking merging.

PR #2273, opened 2021-01-05
First pass at embedding Game in Player.

kberg:
Relies on #2272


src/Player.ts
```diff
@@ -66,6 +66,7 @@ export class Player implements ISerializable<SerializedPlayer> {
   private usedUndo: boolean = false;
   private waitingFor?: PlayerInput;
   private waitingForCb?: () => void;
+  private _game: Game | undefined = undefined;
```
bafolts, line 69: A `Player` should always have a `Game` so the `undefined` here may be confusing. May be worth noting here in a comment that `undefined` is only expected due to how our code is written where we may be creating a `Player` preparing to put it on a `Game`. That is what is implied below by "too soon" but may not be clear.

bafolts (approved):
This should be a good overall change and rolling it out slowly like this will show us if it is going to introduce anything we are forgetting. Often times APIs have expanded because a `Game` would be needed when only `Player` was available. Sometimes it would involve having to pass `Game` through multiple places to get it where it was needed. This will allow more work to be done on only `Player`. There is some OO principal to pass as an argument only the part that you need but here we are in a grey area as we need both `player` and `game` in most situations. This is clearer though since as written today we assume that `game` is the game associated with the player where that isn't necessarily the case. With this change we would also always be ensuring we have the `game` associated with the `player`. I don't think we can avoid the issue where `game` may get left `undefined` unless we eventually create the `Player` within `Game`. Heading that direction would be helpful as it would get around the need to use `Game | undefined` which is exposing us to risk if we ever failed to set `game`. If it couldn't be `undefined` we wouldn't have that issue and we should never really have a player without a game.

kberg:
> This should be a good overall change and rolling it out slowly like this will show us if it is going to introduce anything we are forgetting. Often times APIs have expanded because a `Game` would be needed when only `Player` was available. Sometimes it would involve having to pass `Game` through multiple places to get it where it was needed. This will allow more work to be done on only `Player`. There is some OO principal to pass as an argument only the part that you need but here we are in a grey area as we need both `player` and `game` in most situations. This is clearer though since as written today we assume that `game` is the game associated with the player where that isn't necessarily the case. With this change we would also always be ensuring we have the `game` associated with the `player`. I don't think we can avoid the issue where `game` may get left `undefined` unless we eventually create the `Player` within `Game`. Heading that direction would be helpful as it would get around the need to use `Game | undefined` which is exposing us to risk if we ever failed to set `game`. If it couldn't be `undefined` we wouldn't have that issue and we should never really have a player without a game.

Actually, with this setup we won't have a case where Game is undefined when we use Player, with the exception of during setup *before* calling the Game constructor. Because the game and player constructors themselves are hidden we can make certain assurances of what is defined before the rest of the API is used.

If you wanted to take this just one step further to make the assurance even more solid, during game setup, don't create Players, but something that looks like a player but has no behavior.

PR #2277, opened 2021-01-05
Move Ares' mechanism for logging tile and space bonus enums into the enums themsleves using a shared namespace.

kberg:
Tested by manually running this and generating messages. The log messages are a little ugly, and can use some cleaning.


src/LogHelper.ts
```diff
@@ -43,27 +43,7 @@ export class LogHelper {
...
-      type = 'ocean tile';
-      break;
-
-    default:
-      type = 'special tile';
-      break;
-    }
-
-    this.logBoardTileAction(game, player, space, type);
+    this.logBoardTileAction(game, player, space, TileType.toString(tileType) + ' tile');
```
bafolts, line 46: The build size thanks you.

src/SpaceBonus.ts
```diff
@@ -1,15 +1,37 @@
...
+// for now.
 
 export enum SpaceBonus {
-    TITANIUM,
-    STEEL,
-    PLANT,
-    DRAW_CARD,
-    HEAT,
-    OCEAN,
+    TITANIUM, // 0
```
bafolts, line 6: Future developers considering changing the order here thank you.

bafolts (approved):

PR #2279, opened 2021-01-05
Cards vue refactor - description(1)

chosta:
Adding `effect` and `action` instead of effectBox + streamlining description for these elements. Also a couple of cards after the change.

### DIFF
--- a/src/cards/base/AquiferPumping.ts
+++ b/src/cards/base/AquiferPumping.ts
@@ -24,7 +24,7 @@ export class AquiferPumping extends Card implements IActionCard, IProjectCard {
       metadata: {
         cardNumber: '187',
         renderData: CardRenderer.builder((b) => {
-          b.effectBox((eb) => eb.megacredits(8).steel(1).brackets.startAction.oceans(1).description('Action: Spend 8 MC to place 1 ocean tile. STEEL MAY BE USED as if you were playing a Building card.'));
+          b.action('Spend 8 MC to place 1 ocean tile. STEEL MAY BE USED as if you were playing a Building card.', (eb) => eb.megacredits(8).steel(1).brackets.startAction.oceans(1));
         }),
       },
     });
--- a/src/cards/base/BusinessNetwork.ts
+++ b/src/cards/base/BusinessNetwork.ts
@@ -26,7 +26,7 @@ export class BusinessNetwork extends Card implements IActionCard, IProjectCard {
         cardNumber: '110',
         description: 'Decrease your MC production 1 step.',
         renderData: CardRenderer.builder((b) => {
-          b.effectBox((eb) => eb.empty().startAction.empty().description()).text('Action: Look at the top card and either buy it or discard it', CardRenderItemSize.SMALL, true).br;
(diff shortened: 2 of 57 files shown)


src/cards/render/CardRenderer.ts
```diff
@@ -83,9 +83,12 @@ export class CardRenderEffect extends CardRenderer {
 
   public get description(): ItemType {
     this._validate();
-    // TODO (chosta): validate builder method to make sure it's the last element
```
chosta, line 86: This is kind of not needed now as setting a description is pushing the string to the last row (see `public set description`) and there is already a check if effectBox has a cause, effect and description

kberg:
Note conflict on four files.

src/cards/base/AquiferPumping.ts
```diff
@@ -24,7 +24,7 @@ export class AquiferPumping extends Card implements IActionCard, IProjectCard {
       metadata: {
         cardNumber: '187',
         renderData: CardRenderer.builder((b) => {
-          b.effectBox((eb) => eb.megacredits(8).steel(1).brackets.startAction.oceans(1).description('Action: Spend 8 MC to place 1 ocean tile. STEEL MAY BE USED as if you were playing a Building card.'));
+          b.action('Spend 8 MC to place 1 ocean tile. STEEL MAY BE USED as if you were playing a Building card.', (eb) => eb.megacredits(8).steel(1).brackets.startAction.oceans(1));
```
kberg, line 27: nit: linebreak? Very minor.

src/cards/base/BusinessNetwork.ts
```diff
@@ -26,7 +26,7 @@ export class BusinessNetwork extends Card implements IActionCard, IProjectCard {
         cardNumber: '110',
         description: 'Decrease your MC production 1 step.',
         renderData: CardRenderer.builder((b) => {
-          b.effectBox((eb) => eb.empty().startAction.empty().description()).text('Action: Look at the top card and either buy it or discard it', CardRenderItemSize.SMALL, true).br;
+          b.action(undefined, (eb) => eb.empty().startAction.empty()).text('Action: Look at the top card and either buy it or discard it', CardRenderItemSize.SMALL, true).br;
```
kberg, line 29: Why is text necessary if description is the first parameter to action()?

src/cards/render/CardRenderer.ts
```diff
@@ -441,6 +444,27 @@ class Builder {
     return this;
   }
 
+  public standartProject(description: string, eb: (builder: EffectBuilder) => void): Builder {
```
kberg, line 447: "standart"

kberg (approved):

kberg:
I love it.

src/cards/base/BusinessNetwork.ts
chosta: It's a hack. The original iconography requires bolded uppercase text, while description anywhere else is teeny-tiny non-formatted text. I'd just roll with it for now.

src/cards/render/CardRenderer.ts
chosta: I'll probably keep making this mistake as this word ends in `t` in my language. I need to find a way to hack my brain to stop it!

PR #2281, opened 2021-01-06
Convert less common parameters to AddResourceToCard to an options intererface

kberg:


src/deferredActions/AddResourcesToCard.ts
```diff
@@ -7,40 +7,53 @@ import {LogHelper} from '../LogHelper';
 import {Tags} from '../cards/Tags';
 import {DeferredAction} from './DeferredAction';
 
+export namespace AddResourcesToCard {
+  export interface Options {
```
Lynesth, line 11: Why namespace and export it ?

src/deferredActions/AddResourcesToCard.ts
```diff
@@ -7,40 +7,53 @@ import {LogHelper} from '../LogHelper';
...
+    count?: number;
+    restrictedTag?: Tags;
+    title?: string;
+    filter?: (card: ICard) => boolean;
```
Lynesth, line 15: What is this used for ?

src/deferredActions/AddResourcesToCard.ts
```diff
@@ -7,40 +7,53 @@ import {LogHelper} from '../LogHelper';
...
-        public count: number = 1,
-        public restrictedTag?: Tags,
-        public title: string = 'Select card to add ' + count + ' ' + resourceType + '(s)',
+        public options: AddResourcesToCard.Options = {},
   ) {}
 
   public execute() {
+    const count = this.options.count || 1;
+    const title = this.options.title ||
+      'Select card to add ' + count + ' ' + this.resourceType + '(s)';
```
Lynesth, line 30: To prevent any problem in case the title isn't properly set when resourceType is undefined:
```ts
'Select card to add ' + count + ' ' + (this.resourceType !== undefined ? this.resourceType : 'resource') + '(s)';
```

src/deferredActions/AddResourcesToCard.ts
kberg: The Moon. It allows a second pass at cards, as it restricts some cards from being chosen.

src/deferredActions/AddResourcesToCard.ts
kberg: It allows you to write things like this:

```game.defer(
  new AddResourcesToCard(
    player,
    game,
    ResourceType.MICROBE,
    {count: 2} as AddResourcesToCard.Options));
```

src/deferredActions/AddResourcesToCard.ts
kberg: I wonder why resourceType would ever be optional. Let me see if it ever comes in undefined.

But if it does, then yes.

kberg: There's a community card that does something weird and different. Huh.

`action = new AddResourcesToCard(player, game, undefined, {count: quantity, restrictedTag: Tags.VENUS, title: 'Select Venus card to add ' + quantity + ' resource(s)'});`

Specifically, it's not filtering on resource type, but any venus card with a resource.

Lynesth: If you mean Maxwell Base, it's a Venus Next card, but yeah it can add a resource to any of your Venus card, whatever the resource type. Same for Corroder Suits and maybe others.

kberg: Corroder Suites does it different: 
```
      return new SelectCard(
        'Select card to add 1 resource',
        'Add resource',
        CorroderSuits.getVenusResCards(player),
        (foundCards: Array<ICard>) => {
          player.addResourceTo(foundCards[0], 1);
          LogHelper.logAddResource(game, player, foundCards[0]);
          return undefined;
        },
      );
```

I specifically mean the Venus colony.

src/deferredActions/AddResourcesToCard.ts
```diff
@@ -7,40 +7,53 @@ import {LogHelper} from '../LogHelper';
...
   public execute() {
+    const count = this.options.count || 1;
+    const title = this.options.title ||
+      'Select card to add ' + count + ' ' + (this.resourceType || 'resources') + '(s)';
     let resourceCards = this.player.getResourceCards(this.resourceType);
 
-    if (this.restrictedTag !== undefined) {
-      resourceCards = resourceCards.filter((card) => card.tags.indexOf(this.restrictedTag!) !== -1);
+    if (this.options.restrictedTag !== undefined) {
+      resourceCards = resourceCards.filter((card) => card.tags.indexOf(this.options.restrictedTag!) !== -1);
```
sienmich, line 34:
```suggestion
      resourceCards = resourceCards.filter((card) => card.tags.includes(this.options.restrictedTag!));
```

src/deferredActions/AddResourcesToCard.ts
```diff
@@ -7,40 +7,53 @@ import {LogHelper} from '../LogHelper';
...
 
     if (resourceCards.length === 0) {
       return undefined;
     }
 
     if (resourceCards.length === 1) {
-      this.player.addResourceTo(resourceCards[0], this.count);
-      LogHelper.logAddResource(this.game, this.player, resourceCards[0], this.count);
+      this.player.addResourceTo(resourceCards[0], count);
+      LogHelper.logAddResource(this.game, this.player, resourceCards[0], count);
```
sienmich, line 46: Maybe this log should be moved to `player.addResourceTo`? Then it would be only in 1 place, not 2

src/deferredActions/AddResourcesToCard.ts
```diff
@@ -7,40 +7,53 @@ import {LogHelper} from '../LogHelper';
...
-        public count: number = 1,
-        public restrictedTag?: Tags,
-        public title: string = 'Select card to add ' + count + ' ' + resourceType + '(s)',
+        public options: AddResourcesToCard.Options = {},
   ) {}
 
   public execute() {
+    const count = this.options.count || 1;
+    const title = this.options.title ||
+      'Select card to add ' + count + ' ' + (this.resourceType || 'resources') + '(s)';
```
sienmich, line 30: Maybe add here the `restrictedTag` info, sth like `'Select ' + (this.restrictedTag || '') + ' card' `? And remove title from most of calls

sienmich (commented):
Ok, my first review :smile:. My comments are not really about your changes, more about the code, so not that important. Cheers!

src/deferredActions/AddResourcesToCard.ts
kberg: Perhaps later. Player is already 2200 lines of code. I'd like to see it start to get restructured, and the payback for this isn't quite so much.

src/deferredActions/AddResourcesToCard.ts
kberg: Good idea, but in another PR. Any change I make about logging text impacts the translations, so I'd rather wait on that.

kberg:
Any other concerns?

kberg:
@Lynesth anything preventing this from being submitted?

Lynesth (approved):
Looks ok to me

PR #2282, opened 2021-01-06
Cards to vue - refactor description (2)

chosta:
All effectBoxes -> action() or effect()

A lot of files but I see other ongoing card refactoring I decided to push this a bit so we get it done and avoid conflicts.


kberg (approved):

kberg:
Love it.

PR #2284, opened 2021-01-06
Use an options object for SelectHowToPay.

kberg:


src/deferredActions/SelectHowToPayDeferred.ts
```diff
@@ -7,35 +7,46 @@ export class SelectHowToPayDeferred implements DeferredAction {
...
       },
     );
   }
 }
+
+export namespace SelectHowToPayDeferred {
+  export interface Options {
+    canUseSteel?: boolean;
+    canUseTitanium?: boolean;
+    title?: string; // = ;
```
sienmich, line 48:
```suggestion
    title?: string;
```

sienmich (approved):

PR #2286, opened 2021-01-06
Converted all `player.cardsInHand.push(game.dealer.dealCard())` to `p…

sienmich:
…layer.drawCard(game)`.

This seems simpler to me.

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -58,6 +58,7 @@ import {IAresGlobalParametersResponse, ShiftAresGlobalParameters} from './inputs
 import {Timer} from './Timer';
 import {GameLoader} from './database/GameLoader';
 import {CardLoader} from './CardLoader';
+import {range} from './utils/utils';
 
 export type PlayerId = string;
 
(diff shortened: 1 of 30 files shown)


src/cards/promo/HiTechLab.ts
```diff
@@ -35,7 +35,7 @@ export class HiTechLab implements IProjectCard {
           const cardsDrawn: Array<IProjectCard> = [];
           for (let counter = 0; counter < amount; counter++) {
             cardsDrawn.push(game.dealer.dealCard());
```
kberg, line 37: Did you skip this?

kberg (commented):
LG, just checking one thing before approving.

src/cards/promo/HiTechLab.ts
sienmich: Yep, here we select one card to draw from many, so i skipped. However this pattern also repeats many times, so I'll add this as an option. Thanks!

kberg: Do it separately, this can be pulled in./

kberg (approved):

src/Player.ts
```diff
@@ -1303,6 +1304,12 @@ export class Player implements ISerializable<SerializedPlayer> {
     );
   }
 
+  public drawCard(game: Game, n = 1): void {
+    range(n).forEach(() => {
```
Lynesth, line 1308: Why create an array instead of a simple `for` loop ?

sienmich: Changed in https://github.com/bafolts/terraforming-mars/pull/2292

Issue #1391, opened 2020-09-26
Feature request: indicator for protected plant resource and metal bonus


vsrisuknimit:
- Tiny shield icon next to plant resource box for Protected Habitat and Asteroid Deflection System.
- One white dot for each metal resource bonus. For example, two dots next to titanium if a player has both Advanced Alloys and Mercurian Alloys. Don't think it should take Unity into account here.

![Untitled-1-01](https://user-images.githubusercontent.com/14239220/94340696-c2393c00-ffd1-11ea-9daf-49b0d50b1037.png)

vsrisuknimit:
![Untitled-1-01](https://user-images.githubusercontent.com/14239220/94342731-0122be00-ffe1-11ea-97e7-7360b9e89d76.png)

Alternative with + symbol instead of white dot.

kevinb9n:
I don't think I get any votes here :-) but I like it!

I think the titanium bonus dots will be confusing/misleading if they don't factor Unity in, though. You'll end up getting bugs filed about it.

Might as well also show the shield next to "RES: N" on a card? Pets would always have it. Pretty useful.

vsrisuknimit:
![image](https://user-images.githubusercontent.com/14239220/94351157-1cfb8380-0024-11eb-9084-51c197928f13.png)

Something like this?

If you are already looking at the opponent's animal cards, would you not already see their protected habitat?

kyshantry:
Yes, excellent! I like the pluses or the value of the resource; I feel like the value of the resources will be a bit noisy though. I agree don't include unity. And per the example with strat birds: yes, just like that. Sometimes folks have a billion blue cards and if you're just checking for animals you might miss protected habs :)

kevinb9n:
I can think of two arguments against accounting for Unity:

a) it's not necessary for some purposes, like deciding which player to decrease the titanium production of
b) the user who sees it might accidentally think it's permanent, remember that, then be surprised later

Are there others?

(a) doesn't seem like an actual disadvantage, and for (b), the whole point of this feature is users not having to rely on their faulty memories but being able to see the information in black and white.

Here's an idea: you could show permanent bonus as solid dots and temporary bonuses as hollow dots.

My personal feeling is that not showing the Unity bonus at all may be actively misleading.

kevinb9n:
Oh, and re: "would you not already see their protected habitat?"

[...]

Issue #2150, opened 2020-12-20
Feature: Player boards minor rework


ssimeonoff:
![1](https://user-images.githubusercontent.com/6917565/102722118-45cba600-42ff-11eb-8459-89c540c7eca7.png)

1. The main idea is to add a header to the player boards.
It will add more height, I know...

2.  The secondary idea is to combine to timers and player state labels in a single element
When a player is "active" the timer is ticking. 

3. Minor UI changes as colours and reordering 
More black and less white text - corporations and player names (easier for the eyes)

4. As a future feature, the toggle button TOGGLE TAGS VIEW to have a 3rd state showing the discounts (displayed in that header space)
![2](https://user-images.githubusercontent.com/6917565/102722119-46643c80-42ff-11eb-8274-7ff6b2416c67.png)

5. (optional) More info in the header
Since there is free space (between the corporation an the discounts bar), more info items can be added there. 
Example: The played cards count can be displayed there instead with a more detail overview.
![3](https://user-images.githubusercontent.com/6917565/102722657-dd7ec380-4302-11eb-91d9-8bb187c6544e.png)
Red has played 1 of each type of project cards and has 2 played card that can be activated (active action from corporations corporations are included).

chosta:
Good stuff. I'm doing it next week!

kevinb9n:
> 4. As a future feature, the toggle button TOGGLE TAGS VIEW to have a 3rd state showing the discounts (displayed in that header space)
>    ![2](https://user-images.githubusercontent.com/6917565/102722119-46643c80-42ff-11eb-8274-7ff6b2416c67.png)

This is very good.
It's fine that it's discounts only. I'd love to have a tracker for when-play effects one day, but enough of them aren't strictly tag-based that it can't be shoehorned into this model anyway. We can forget about them for now.

In addition to per-tag discounts there are of course per-card discounts, from four cards which can total down to -6. I think it would work to show that value in the same way as these, just to the left of them, right above the "cards in hand" count.

The one remaining type of card discount is also per-card, but conditional (Cutting Edge), so meh, it'll just be a stealth discount, seems fine to me.

Btw #1426 talks about drawing a separation between the tags (up to the city tag) and everything else (after that). I do think that would help players of Aridor and Diversity Support (and the Diversity global event) to understand them correctly.


>    Since there is free space (between the corporation an the discounts bar), more info items can be added there.
>    Example: The played cards count can be displayed there instead with a more detail overview.
[...]

PR #2287, opened 2021-01-06
dot for alloys, shield for protected plants

vsrisuknimit:
image

image

Dots to indicate whether steel and Ti has bonus from alloys (Advanced Alloys, Mercurian Alloys, Regoplastics). It does not care about Phobolog (user can already see the corp) or Unity (the policy affects everyone).

Shield to indicate whether plants are protected.

All work done by @nwai90 . I am just porting this over from his experimental branch.


vsrisuknimit:
This can close #1391 .

vsrisuknimit:
Tagging @chosta for review.

kberg:
I love it, I really do. Two thoughts: 1) consider a modest + instead of dots and 2) the Moon expansion will provide similar protection to steel and titanium. I'd love to be able to fit in there.

I guess a third thought? Can that shield be reused for plants, animals and microbes?

(By modest, I mean something that isn't very thin, even if it's closer to a square than a plus? Like 
image


(Like the swiss flag's cross.)

kevinb9n:
Hello! Just want to say in a friendly way that I believe this decision is a mistake, which makes the interface unfriendly to novice players of the game.

You're asking for mental gymnastics to be done, which experienced players like us have no trouble with. I personally really don't want to see the app go in that direction.

A number in the interface should have a clear and simple meaning. The clearest and simplest, by far, is to portray the value that each unit of the resource has. Yes, I am fully aware that there are disadvantages to having to show yet another numeral near all the existing ones, or to show a number of dots which can in the rarest of circumstances get as high as 7. But I feel that one must be over-weighting those disadvantages by a wide margin in order to feel that the current solution is acceptable.

If this goes through, I'm not saying it will be a disaster, only that to *me* it will be the clearest design mistake I've seen in this particular app, that's all.

vsrisuknimit:
@kberg 

image

(1) An example of what + looks like. We could put it up for a vote. (2) There are very few cards that take away Ti or Steel. Don't see a need for a shield there. It adds too much cluster to the overview tab for too little benefit. (3) Yes, the shield is in asset and can be reused for protected animals. That would be a different PR though.

chosta:
I'm not sure about this exact solution although it is needed.
Let me gather my thoughts in the morning and with the help of some coffeee I'll give my suggestions. I am not against it but I have concerns similar to @kevinb9n. I know this has been long overdue but lets do it the right way. We will have an indicator one way or another.

chosta:
Just to be clear, the shield for plants works pretty well. The dot / plus is what concerns me.

kberg:
> (2) There are very few cards that take away Ti or Steel. Don't see a need for a shield there.

Right, I agree -- what I'm saying, though, is that there's this card coming:

image

along with others that do. I'd like for there to be room.

vsrisuknimit:
> You're asking for mental gymnastics to be done, which experienced players like us have no trouble with. I personally really don't want to see the app go in that direction.

There is no mental gymnastic requires. There should be no calculation at all. The point of the alloy indicator is NOT tell exactly how much Ti and steel cost. The point of the indicator is to tell who benefit more from Steel and Ti than other players, so that they can attack those players or hate-draft accordingly.

vsrisuknimit:
> along with others that do. I'd like for there to be room.

Yes, there should be room for a shield. Whether we need it or not is up to debate. There are many more asteroid / plant-burning cards than Ti/Steel decreasing cards.

kberg:
> > along with others that do. I'd like for there to be room.
> 
> Yes, there should be room for a shield. Whether we need it or not is up to debate. There are many more asteroid / plant-burning cards than Ti/Steel decreasing cards.

That's fine. When I resolve this ehnahcement with the Moon, I might have to do something such as move the shield up, to the right, and reduced size. And that's fine - we can improve from there.

kberg:
> > You're asking for mental gymnastics to be done, which experienced players like us have no trouble with. I personally really don't want to see the app go in that direction.
> 
> There is no mental gymnastic requires. There should be no calculation at all. The point of the alloy indicator is NOT tell exactly how much Ti and steel are worth. The point of the indicator is to tell who benefit more from Steel and Ti than other players, so that they can attack those players or hate-draft accordingly.
> 
> The overview table is such a central element to the game. It should be as clean as possible.

"mental gymnastics" is a charged phrase meant to imply unfairness, when in fact this feature, as-is, already makes calculating costs easier for a player than the current state, which is to have to look up cards.

IMO, this PR, as a net improvement, is worth submitting, and iterating upon.

src/server/ServerModel.ts
```diff
@@ -375,6 +376,7 @@ function getPlayers(players: Array<Player>, game: Game): Array<PlayerModel> {
       name: player.name,
       plants: player.plants,
       plantProduction: player.getProduction(Resources.PLANTS),
+      plantsAreProtected: player.plantsAreProtected(),
```
kberg, line 379: @nwai90 here's a good use for the general purpose object representing resources. Let's expand on that idea a moment, and put a type on it:

```
interface Stuff<T> {
  plants: T;
  heat: T;
  titanium: T;
};

etc.

Then ServerModel can deliver

units: Stuff<Number>;
protected: Stuff<Boolean>

kberg (approved):

vsrisuknimit:
Let's not merge until @chosta got time to sleep and have coffee.

chosta:
I guess this is going through.
@vsrisuknimit can you check what the maximum possible amount of Ti or Steel discounts per resource? Is there going to be enough space for all of them. Also, should we show a dot / plus when unity is in charge?

chosta:
> Let's not merge until @chosta got time to sleep and have coffee.

Oh it could be merged. I don't see any harm. If I don't like something or I have a better suggestion, I'll add it in a PR.

vsrisuknimit:
@chosta oops. We commented at the same time. 

As I mentioned above, it only counts alloys. The point of the dot is not to help calculate the exact worth of Steel or Ti. It is to help other players know who care more about steel and Ti. So no dot for Unity, since everyone benefits. Not Phobolog since you can already read the corp name. The maximum number is 2 for both metals.

chosta:
> As I mentioned above, it only counts alloys.

Hmm, in this case we need a legend. This got me confused as I was thinking about discounts. I still need my coffee, let's continue tomorrow.

vsrisuknimit:
> This got me confused as I was thinking about discounts.

@chosta There is no discount here. You are getting confused with space discount from card like Space Station and Quantum Extractor.

chosta:
> @chosta There is no discount here.

I understood that when you told me. So that is the issue here. It's not self-explanatory and raises more questions than it answers.

vsrisuknimit:
> If it's alloys only, players need to be able to check what the symbols stand for easily.

Tooltip? On hover? But that is not good for mobile right?

kberg:
Consider replacing dots with a MC icon. eg ([2], [3], [4]) or
[2]+1(subscript)

>

chosta:
I think just a popup [legend] 
Could be next to the Concise / Full buttons.

We can let this pass and I'll improve once I start working on. #2150

vsrisuknimit:
> Consider replacing dots with a MC icon. eg ([2], [3], [4]) or [2]+1(subscript)
> […](#)

@kberg That might be too distracting. You will have to show 2 and 3 for everyone otherwise it can be confusing why some players say 4 or 5.

vsrisuknimit:
Tagging @ssimeonoff for this topic.

kevinb9n:
Sorry, I try to avoid using charged language, so I want to revise that now. I just meant that it takes more mental effort, which I think is just factual.

Let me keep thinking about how a general-purpose Effects Tracker might possibly work, which would have the room (somehow!) to show the full values, because if it's at all possible it would take some of the pressure off here. That is, then it's easier (at least for me) to understand how this feature really only needs to serve the sort of "which player cares most about this resource" use case.

And under that model, sure, okay, it can leave Unity out, and it can skip showing the first 2/3 representing the base value. But my lingering question is: really, why not include Phobolog? Yes, you *can* see the corp name there, and *most* players know what that name means once they've played several games. But it's still going to look just plain wrong when I play Phobolog and you play Advanced Alloys and yet somehow our titanium displays are different. Also, when I see a Phobolog opponent and I see a dot there, I'm going to assume it IS the Phobolog dot and I'm not going to guess that it really means you *also* played Mercurian.

vsrisuknimit:
> And under that model, sure, okay, it can leave Unity out, and it can skip showing the first 2/3 representing the base value. But my lingering question is: really, why not include Phobolog? Yes, you _can_ see the corp name there, and _most_ players know what that name means once they've played several games. But it's still going to look just plain wrong when I play Phobolog and you play Advanced Alloys and yet somehow our titanium displays are different. Also, when I see a Phobolog opponent and I see a dot there, I'm going to assume it IS the Phobolog dot and I'm not going to guess that it really means you _also_ played Mercurian.

I am open to adding it for Phobolog. So up to 3 dots for Ti.

kevinb9n:
> > Consider replacing dots with a MC icon. eg ([2], [3], [4]) or [2]+1(subscript)
> > […](#)
> 
> @kberg That might be too distracting. You will have to show 2 and 3 for everyone otherwise it can be confusing why some players say 4 or 5.

(I would agree it should do it consistently like that. Note that for new players even just the 2s and 3s are a helpful reminder. But in the heat column, only Helion would have a 1, because in that case the rest are zero.)

vsrisuknimit:
image

Some changes:
- simplify the vue methods.
- vertical arrangement instead of horizontal
- allow a dot from Phobolog. So max two dots for steels, three for Ti. Still no dot for Unity.

vsrisuknimit:
Do not merge yet. Ongoing discussion on discord.

kberg:
> Do not merge yet. Ongoing discussion on discord.

Sure thing. I'll leave it to other people to merge.

bafolts:
I was confused as to what the dots represent and I am still confused as to what they represent while reading through this so the user interface may not be intuitive. The shield is intuitive to me and I assume it means you can not take those resources. The dots are the confusing part. I would still be fine merging this as we could always change the dots to something else later if while playing players arrived upon a better idea once seeing it in practice.

bafolts:
@vsrisuknimit I think the shield seems reasonable enough. Could consider a separate PR for the protected plants while figuring out strategy for the dots.

vsrisuknimit:
New commit allows for Unity as well.

@bafolts the dot represent metal bonus. One dot for each extra value for the metal. If you play Advanced Alloys, you get one dot each for Steel and Ti. Phobolog would have one dot for Ti from the beginning.

This allow players to tell who benefit more from the metal in case you want to attack someone's Ti or hate draft space cards from someone.

vsrisuknimit:
image

@bafolts in this case, we can quickly tell that Green's steel is worth one more than normal while Yellow (Phobolog)'s Ti is worth 3 extra than normal.

Ready to merge I believe unless someone has a strong objection. Allowing Phobolog and Unity now after the poll result on discord. Plus symbol is not as popular as the dot. A fair number of votes for number representing the worth of the metal (lower than dots), but I dislike the clutter so I am not going to do it. 

image

kevinb9n:
> but I dislike the clutter so I am not going to do it.

Respect!

Thanks so much for being openminded to the concerns and for taking the questions to discord. And thanks for tackling this in the first place and seeing it through. This is a great improvement.

kberg:
I should add: there's a card in The Moon called Lunar Steel which grants an additional steel megacredit bonus.

chosta:
We have to add a legend for these symbols in an upcoming PR but I see this as an important improvement for intermediate to experienced players. We will take care of the novice soon enough.

Issue #2192, opened 2020-12-27
Standard actions as cards


sienmich:
Once #2185 is done, I'm thinking about making standard actions also as cards.

I mean those 4:
1. Convert plants to greenery
2. Convert heat to temp
3. Trade with colony
4. Send a delegate

And those 2 (they would require some nice graphic imo, sb have an idea / can help?)
5. Claim a milestone
6. Fund an award

They all work just like standard project, except the fact they are not. So I think having them in front and back end as some StandardAction which is parent class to the new StandardProject would be nice. Any thoughts? :smile:

sienmich:
And possibly we can do sth with kelvinist / scientists action too?

bafolts:
The code separation will be great for this as it will take logic out of `Player` and `Game` again. Displaying the standard actions as cards on the UI makes sense. On the back end I would start to separate them from cards or subclass card in a way where parts can be shared. The `IActionCard` really has nothing to do with a card. I would consider renaming this `IAction` and begin moving away from everything being named `Card` when they are not cards. Similar to the `CardManifest`. That should really be `ExpansionManifest`.

kevinb9n:
+1
Completely agree, the SP vs SA distinction is arbitrary and what's good for one is good for the other!

Does it gray out the ones I can't currently afford?

If you are willing to show three different cards for Trade, then there won't need to be a second step to choose what to pay.

I assume that Send Delegate will just have no cost before the arrow if I have someone in the lobby.

> And those 2 (they would require some nice graphic imo, sb have an idea / can help?)
> 5. Claim a milestone
> 6. Fund an award

You can just do what the Vitor card is doing.

kevinb9n:
(Whoever plays this as Ecoline first will complain that it doesn't change to 7 plants for them (and Cryo Sleep etc.) but I wouldn't suggest doing that unless we think we're gonna do cost changes more comprehensively like in #2187 / #776.)

kevinb9n:
How about I file a separate issue exploring how we deal with special party-based actions (of which there are now many, if PolitAgen becomes official)?

kevinb9n:
It sounds like our top-level "your turn" menu is reducing down to just:

* Play card
* Use card action
* Standard action
* Standard project
* (and the non-action options: end, pass, undo)

(I'm assuming for now we can fold the turmoil actions in somewhere appropriate, and I'm also not worrying about the special first action of corps like Inventrix.)

[...]

Issue #2187, opened 2020-12-25
Toggle your goggles [see your cards as modified by effects]


kevinb9n:
Sorry I don't have drawings yet, just wanted to get this idea down for a while.

Sometimes when ogling my cards my mind boggles and I wish I could toggle on some goggles. [Imagine these toggles as separate buttons in the sidebar, or if you prefer, one button to rule them all.]

* COST goggles: click it on, and all costs of unplayed cards (#776) and standard actions (e.g. #2185) are magically rewritten based on all discount effects I have in play

* REQUIREMENT goggles: click it on, and all requirements on unplayed cards are rewritten to account for at least Inventrix/AdaptTech (discussed as part of #1407). (This might weirdly rewrite to "max 9 oceans" or maybe even "-32 C" but won't hide the requirement completely, because Cutting-Edge.)

* COUNT THE THINGS goggles: click it on, and all instructions on unplayed cards that involve either the "per" icon like Medical Lab, or a conditional OR clause like Nitro-Rich, are rewritten to show the actual amount it would currently give. Also rewrites card-actions I have in play like Orbital Cleanup, and VPs like on Water Import. (To be sure, you would often like to know the raw measurement in question, or at least have a way to know whether you are on the cusp of the next bump up, but that would require actual UI of some kind instead of just rewriting using existing UI we already have. It can be considered independently.)

* TRIGGER goggles: (putting these here only to further illustrate the general goggle concept, but I have no doubt that it's too hard) click it on, and see on each card (somehow) additional instructions from your active "when-play-card/tag" triggers. Example: after playing OptAero, now see 3 MC 3 Heat on all your space events. (Warning: there is very much a slippery slope here, and it will never be feasible to really show every single thing clearly that will result.)

**Important clarifications**

* Each goggle(?) has to be *very easy* to turn on and off, without having to go into settings, or this whole thing will suck
* To keep cards simple and clean, the goggle concept is 100% about rewriting a card to look "exactly" like a different card that *could* itself have been printed - except:
* It's crucial to always have *some* visual indicator that the information has been changed. For example, change text from black to either dark green (persistent effect) or dark red (includes at least one temporary effect, like from Special Design)
* Goggles shouldn't change card text. So even with goggles on, I can still compare the raw (text) and cooked (graphic) forms.
* While a desktop user mouse-hovers on a cost/requirement/etc, it could flip to the opposite goggle mode temporarily.
[...]

PR #2288, opened 2021-01-06
Added Standard Actions

sienmich:
Closes https://github.com/bafolts/terraforming-mars/issues/2192

For now i created two standard actions:
Screenshot from 2021-01-06 20-33-30

Screenshot from 2021-01-06 20-34-26

Converting plants to greenery now requires confirmation (tagging https://github.com/bafolts/terraforming-mars/issues/1240).
I understand that's controversial. I'll be happy to allow placing greenery without confirmation, however I don't know how to implement that. Same goes with current selling patents. I would like to just replace the `Sell Patents` text with sell patents card. For example if `PlayerInput` could have `title` as text or as card would be cool. Any ideas how it should be done? I'm kinda stuck here for now
Also it would be nice to not hide those standard actions even when `Standard Actions` radio button is not selected imo. Same with standard projects. Do you agree? 

Implementation details:
I created StandardAction class. StandardProject is its subclass.


src/cards/standardActions/ConvertHeat.ts
```diff
@@ -0,0 +1,46 @@
...
+import {CardName} from '../../CardName';
+import {CardMetadata} from '../CardMetadata';
+import {CardRenderer} from '../render/CardRenderer';
+import {Game} from '../../Game';
+import {REDS_RULING_POLICY_COST} from '../../constants';
+import {PartyHooks} from '../../turmoil/parties/PartyHooks';
+import {PartyName} from '../../turmoil/parties/PartyName';
+import * as constants from '../../constants';
+import {StandardAction} from './StandardAction';
+import {SelectOption} from '../../inputs/SelectOption';
```
sienmich, line 11:
```suggestion
```

nwai90:
@sienmich Agree that `StandardAction` and `StandardProject` should be differentiated. The main issue I see with this is that every greenery placement and every temperature increase would now require an extra click each, which is not ideal. The idea is great though! 👍

bafolts:
To avoid the extra clicks we could consider some updated `OrOptions` which used something that looked like a card. Would need a back button though. With that component in place we could then avoid the issue with having to click and confirm.

bafolts:
Thinking some more. Some `UserActions` player input that supported going back once the top level option were supported should work. Would need to come up with a card for each action such as `Play Project Card` as well. When clicked the UI would go to the card selection screen but now include a back button to go back to the list of all the other actions.

kevinb9n:
Looking good!

at some point I would like to discuss where this is heading relative to #953 (I put some thoughts at https://github.com/bafolts/terraforming-mars/issues/2192#issuecomment-751511295) but I'm not trying to block you up.

Lynesth:
> Thinking some more. Some `UserActions` player input that supported going back once the top level option were supported should work. Would need to come up with a card for each action such as `Play Project Card` as well. When clicked the UI would go to the card selection screen but now include a back button to go back to the list of all the other actions.

That would lead to even more clicks then.

sienmich:
Ok, I don't know how to update those `OrOptions` so they could accept cards. Could sb maybe help with this?
For now I resolved this same as Sell Patents are working - they are looking good in logs, and their whole logic is in separate files. I'd like to merge this for now.

kberg:
> @sienmich Agree that `StandardAction` and `StandardProject` should be differentiated. The main issue I see with this is that every greenery placement and every temperature increase would now require an extra click each, which is not ideal. The idea is great though!

How many clicks does a player make in the entire game? I find it particularly interesting that greeneries (and in this case, temperature) are a focus. But OK.

sienmich:
I think less click is better. I will work on displaying cards instead of text in Player Input. Can I merge current version? Any code concerns?

kberg:
Ah, I didn't realize I was the one giving the OK on this. Any last concerns? 
@nwai90 @bafolts @Lynesth ?

bafolts:
This change is OK with me. No last concerns.

src/CardFinder.ts
```diff
@@ -33,50 +34,40 @@ export class CardFinder {
       return CardFinder.decks;
     }
 
-    public getStandardProjectCardByName(cardName: string): StandardProjectCard | undefined {
-      let found : (ICardFactory<StandardProjectCard> | undefined);
-      CardFinder.getDecks().some((deck) => {
-        found = deck.standardProjects.findByCardName(cardName as CardName);
-        return found !== undefined;
+    public getCardByName<T extends ICard>(cardName: string, decksToSearch: (man: CardManifest) => Array<Deck<T>>): T | undefined {
```
kberg, line 37: Minor comments, shorten decksToSearch to decksd, and fully spell manifest, which is that way everywhere else.

src/CardFinder.ts
```diff
@@ -33,50 +34,40 @@ export class CardFinder {
...
       });
       if (found !== undefined) {
         return new found.Factory();
       }
-      console.warn(`standard project card not found ${cardName}`);
+      console.warn(`card not found ${cardName}`);
       return undefined;
     }
 
+    public getProjectOrSAByName(cardName: string): StandardAction | IProjectCard | undefined {
```
kberg, line 53: Please don't shorten StandardAction.

src/CardLoader.ts
```diff
@@ -66,8 +66,8 @@ export class CardLoader {
   public getProjectCards() {
     return this.getCards((manifest) => manifest.projectCards);
   }
-  public getStandardProjects() {
-    return this.getCards((manifest) => manifest.standardProjects);
+  public getStandardActions() {
```
kberg, line 69: Why is this being renamed? They're standard projects, a fairly commonly-understood term.

src/CardName.ts
```diff
@@ -1,14 +1,18 @@
...
-    STANDARD_SELL_PATENTS = 'Sell Patents',
-    STANDARD_POWER_PLANT = 'Power plant', // 'Power Plant' is already taken
-    STANDARD_ASTEROID = 'asteroid', // 'Asteroid' is already taken
-    STANDARD_AIR_SCRAPPING = 'Air Scrapping',
-    STANDARD_BUFFER_GAS = 'Buffer Gas',
-    STANDARD_BUILD_COLONY = 'Colony',
-    STANDARD_AQUIFER = 'Aquifer',
-    STANDARD_GREENERY = 'Greenery',
-    STANDARD_CITY = 'City',
+    STANDARD_SELL_PATENTS = 'Sell Patents:SP',
```
kberg, line 3: Why is giving this a suffix necessary?

src/Player.ts
```diff
@@ -1754,80 +1731,61 @@ export class Player implements ISerializable<SerializedPlayer> {
...
+      if (remainingMilestones.options.length >= 1) action.options.push(remainingMilestones);
     }
 
-    if (this.getPlayableActionCards(game).length > 0) {
-      action.options.push(
-        this.playActionCard(game),
-      );
+    const convertPlants = new ConvertPlants();
+    if (convertPlants.canAct(this, game)) {
+      action.options.push(convertPlants.action(this, game));
```
kberg, line 1763: In the cards API, `action` means Do The Thing because it's time to do it. Here, action means something else: return the thing to offer the player. This name is going to be confusing.

src/cards/standardActions/StandardAction.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export abstract class StandardAction implements IActionCard, ICard {
+    public cardType = CardType.STANDARD_ACTION;
+    public hasRequirements = false;
+    public tags = [];
+    public cost = 0;
+    public abstract name: CardName;
+    public abstract metadata: CardMetadata;
+
+    abstract action(_player: Player, _game: Game): PlayerInput | undefined
```
kberg, line 17: Why does action have _player, and _game, but canAct has player and game?

kberg (changes requested):
So, what exactly is the point of this PR? What are you actually trying to achieve? I'm trying to dig in and understand it from the code changes alone, but having a little trouble.

src/CardName.ts
sienmich: There are normal cards called Asteroid and Power Plant. The names must be unique. Earlier it was solved by `asteroid` and `power plant` (no caps). But this way is much better imo

src/CardLoader.ts
sienmich: They are still standard projects - SPs.
SP is a subclass of SA. I didn't want to create separate getters for them, since they are almost the same thing.
So `getStandardActions` returns all SAs, which means SPs and some `Convert Plants`

src/CardFinder.ts
sienmich: Sure, np. However what does the `d` in `decksd` stands for?

src/cards/standardActions/StandardAction.ts
sienmich: My bad, fixed

src/Player.ts
sienmich: I agree, most of the cards in `.action()` have `game.defer(sth)`. It's not the cleanest. It's made this way because of clicks saving. If `.action()` was in fact placing the greenery, then we would need to confirm it earlier.
I'm planing to change that soon. Once I know how display cards in the middle of some `PlayerInput` objects, I would like to improve that action selecting section.

sienmich:
> So, what exactly is the point of this PR? What are you actually trying to achieve? I'm trying to dig in and understand it from the code changes alone, but having a little trouble.

I moved those two standard actions from `Player.ts` to separate files to have cleaner code. Also those two actions now display in log, they look nicer than text imo. Additionally this is initial work to some better looking action selection I'm planning to do.

src/cards/standardActions/ConvertPlants.ts
```diff
@@ -0,0 +1,51 @@
...
+        player.plants -= player.plantsNeededForGreenery;
+        return undefined;
+      },
+    );
+  }
+
+  public metadata: CardMetadata = {
+    cardNumber: 'SA1',
+    renderData: CardRenderer.builder((b) =>
+      b.standardProject('Spend 8 Plants to place a greenery tile and raise oxygen 1 step.', (eb) => {
```
nwai90, line 46: @sienmich Could the description and iconography in metadata also make use of `player.plantsNeededForGreenery`? e.g. 7 plants for Ecoline instead of 8

sienmich: It's not that easy, we have one card instance for whole server right now. It's a good idea, once we have changing card prices it should have been done, but not in this PR imo.

kevinb9n: I do think it's better that the UI not adjust any costs based on player effects until/unless there's at least a plan under way to do it more consistently. The big picture end goal might be something like https://github.com/bafolts/terraforming-mars/issues/2187 (sorry for wall of text)

src/CardName.ts
kberg: I did not know that! I agree.

src/CardFinder.ts
kberg: It means I made a typo. I meant `decks`. Sorry!

src/CardLoader.ts
kberg: I don't think Action is a first class object anywhere else in this game, and I think this is the crux of my misunderstanding.

src/Player.ts
kberg: I appreciate clicks saving, but not at the cost of being able to understand the code.

Much of what I've been doing in the last couple of months on this project is taking bits that are hard to understand and make them easier to understand -- I'd rather get that resolved now.

If the whole point of this PR is to remove a click then perhaps it's worth going back and trying something else? I don't want to ruin your hard work, so don't give up yet, because first before doing anything like that, I need to _understand_ this PR.

kberg:
Aha! I get it now. These really ARE the standard actions from your player board. OK.

src/CardFinder.ts
```diff
@@ -33,50 +34,40 @@ export class CardFinder {
...
 
-    public getStandardProjectCardByName(cardName: string): StandardProjectCard | undefined {
-      let found : (ICardFactory<StandardProjectCard> | undefined);
-      CardFinder.getDecks().some((deck) => {
-        found = deck.standardProjects.findByCardName(cardName as CardName);
-        return found !== undefined;
+    public getCardByName<T extends ICard>(cardName: string, decksd: (manifest: CardManifest) => Array<Deck<T>>): T | undefined {
+      let found : (ICardFactory<T> | undefined);
+      CardFinder.getDecks().some((man) => {
+        decksd(man).some((deck) => {
```
kberg, line 40: I'm so sorry I made you do this. :-(

src/CardLoader.ts
kberg: I suggest not creating a mental notion of standard project is a standard action. Especially not without good clear of documentation for maintainers.

src/Player.ts
```diff
@@ -1776,60 +1753,43 @@ export class Player implements ISerializable<SerializedPlayer> {
       return;
     }
 
+    this.setWaitingFor(this.getNormalActions(game), () => {
+      this.actionsTakenThisRound++;
+      this.takeAction(game);
+    });
+  }
+
+  private getNormalActions(game: Game) {
```
kberg, line 1762: Why is this 'normal' and not 'standard'? I don't understand this yet, but if 'standard' isn't going to work here, then perhaps 'common?

src/cards/standardActions/StandardAction.ts
```diff
@@ -0,0 +1,28 @@
+import {CardType} from '../CardType';
+import {CardMetadata} from '../CardMetadata';
+import {CardName} from '../../CardName';
+import {IActionCard, ICard} from '../ICard';
+import {Player} from '../../Player';
+import {Game} from '../../Game';
+import {PlayerInput} from '../../PlayerInput';
+
+export abstract class StandardAction implements IActionCard, ICard {
```
kberg, line 9: Describe with a comment, please.

src/cards/standardActions/ConvertPlants.ts
```diff
@@ -0,0 +1,51 @@
...
+import {StandardAction} from './StandardAction';
+import {SelectSpace} from '../../inputs/SelectSpace';
+import {ISpace} from '../../boards/ISpace';
+
+
+export class ConvertPlants extends StandardAction {
+  public name = CardName.CONVERT_PLANTS;
+
+  public canAct(player: Player, game: Game): boolean {
+    const hasEnoughPlants = player.plants >= player.plantsNeededForGreenery;
```
kberg, line 20: I suggest similar code to ConvertHeat.

src/CardLoader.ts
kberg: I might even suggest that StandardAction be renamed to PlayerBoardAction. Because it's only really representing the two actions on a player board.

kberg (changes requested):
I kind of don't mind that the game thinks of standard projects and standard actions as 'cards', and fully adopting that idea. After all visually the idea of a 'card' includes drawing a recatangle on a screen. (ref: https://www.nngroup.com/articles/cards-component/#:~:text=Summary%3A%20A%20%E2%80%9Ccard%E2%80%9D%20is,visually%20resembling%20a%20playing%20card.)

src/CardLoader.ts
sienmich: I'd like to convert other actions as well, like colony trading, send a delegate, I just didn't want to make too much changes in 1 PR. They would be nicer with proper graphics insted of text. In this case I think StandardAction is a better than PlayerBoardAction. I'll add some documentation then. Should I separate those SAs from SPs in `CardLoader` then?

src/Player.ts
sienmich: Next time I'll work more on easier understanding me then :)

kberg:
@sienmich let me know if you want some help with some of this.

src/Player.ts
sienmich: Changed to common, added comment

src/cards/standardActions/StandardAction.ts
sienmich: Done

src/CardFinder.ts
```diff
@@ -35,50 +36,40 @@ export class CardFinder {
...
       });
       if (found !== undefined) {
         return new found.Factory();
       }
-      console.warn(`standard project card not found ${cardName}`);
+      console.warn(`card not found ${cardName}`);
       return undefined;
     }
 
+    public getProjectOrStandardActionByName(cardName: string): StandardActionCard | IProjectCard | undefined {
```
kberg, line 55: Note to self to make sure I know why this is necessary.

src/CardFinder.ts
```diff
@@ -35,50 +36,40 @@ export class CardFinder {
...
-      CardFinder.getDecks().some((deck) => {
-        found = deck.corporationCards.findByCardName(cardName as CardName);
-        return found !== undefined;
-      });
-      if (found !== undefined) {
-        return new found.Factory();
-      }
-      console.warn(`corporation card not found ${cardName}`);
-      return undefined;
+      return this.getCardByName(cardName, (man) => [man.corporationCards]);
```
kberg, line 64: Please use `manifest`

src/Player.ts
```diff
@@ -1748,60 +1725,43 @@ export class Player implements ISerializable<SerializedPlayer> {
       return;
     }
 
+    this.setWaitingFor(this.getCommonActions(game), () => {
```
kberg, line 1728: What does this do? Paging @Lynesth whom I think might understand this best.

tests/cards/moon/MoonColonyStandardProject.spec.ts
```diff
@@ -30,7 +30,7 @@ describe('MoonColonyStandardProject', () => {
   });
 
   it('has discount', () => {
-    card.action(player, player.game);
```
kberg, line 33: Changes like this make me think irrelevant changes have leaked into your branch.

src/CardFinder.ts
sienmich: It's used in LogPanel, when we want to display some card

src/CardFinder.ts
sienmich: Sure, I think it was mentioned somewhere else earlier, sorry :)

tests/cards/moon/MoonColonyStandardProject.spec.ts
sienmich: I didn't want to add new actions using old (player, game) interface, so I updated a bit too much. Sorry, next time it will be on separate PR

src/Player.ts
sienmich: So the function `Player.takeAction()` was 250 lines and hard to understand. I've extracted half of it (new `getCommonActions()`), which returns normal turn actions as noted in comment next to declaration:
```ts
// Return possible mid-game actions like play a card and fund an award, but no play prelude card.
  private getCommonActions(game: Game) {
```
I think it's a necessary change to improve the code readability.

sienmich:
Since this PR grew a bit too big and is changing too many things I'll close it and submit those changes in split PRs

PR #2289, opened 2021-01-06
Added copy to clipboard for player links + restyle

chosta:
url-copy

Will add translations once text for this is approved.


nwai90:
This looks like a nice QOL improvement!

chosta:
> This looks like a nice QOL improvement!
Indeed. It would make game creation for tournament volunteers easier, too.

src/components/GameHome.ts
```diff
@@ -1,30 +1,87 @@
...
-                <li v-for="player in (game === undefined ? [] : game.players)">
-                    <span class="player_home_block nofloat">
-                        <span class="player_name" :class="'player_bg_color_'+ player.color"><a :href="'/player?id=' + player.id">{{player.name}}</a></span>
-                    </span>
-                </li>
-            </ul>
-        </div>
+      <div id="game-home" class="game-home-container">
+        <h1>Terraforming mars [game id: <span>{{getGameId()}}</span>]</h1>
+        <h4>Click on YOUR name to start game. Click on the button next to player name to copy the playable link for THAT player.</h4>
```
kberg, line 74: Thanks for removing the exclamation mark. That's what I came here to say. :D

src/components/GameHome.ts
```diff
@@ -1,30 +1,87 @@
...
+// 5. remove the dummy input
+function copyToClipboard(text: string): void {
+  const input = document.createElement('input');
+  input.setAttribute('value', text);
+  document.body.appendChild(input);
+  input.select();
+  document.execCommand('copy');
+  document.body.removeChild(input);
+}
+const DEFAULT_COPIED_ID = '-1';
```
kberg, line 21: I don't know what this field means without context. Can you add a comment saying what it is, please?

src/components/GameHome.ts
kberg, line 74: "Instructions: To start the game, separately copy and share the links with all players, *and then* click on your name.

Save this page in case you or one of your opponents loses a link." ? Something. A little more. I can do it in a follow-up.

src/components/GameHome.ts
```diff
@@ -1,30 +1,87 @@
...
+        return 'n/a';
+      }
+    },
+    setCopiedIdToDefault: function() {
+      this.urlCopiedPlayerId = DEFAULT_COPIED_ID;
+    },
+    getPlayerCubeColorClass: function(color: string): string {
+      return playerColorClass(color.toLowerCase(), 'bg');
+    },
+    getHref: function(id: string): string {
```
kberg, line 60: Are these `id` meant to be playerId? Can you please expand?

src/components/GameHome.ts
```diff
@@ -1,30 +1,87 @@
...
       type: Object as () => GameHomeModel | undefined,
     },
   },
+  components: {
+    Button,
+  },
   data: function() {
-    return {};
+    return {
+      urlCopiedPlayerId: DEFAULT_COPIED_ID as string,
```
kberg, line 34: Is this meant to represent the URL of the player ID most recently copied to the clipboard? Used for, ... I'm not sure after reading everything?

kberg (commented):
Requesting some comments and renames so for later maintenance and comprehension.

src/components/GameHome.ts
chosta: That is good. Thanks!

src/components/GameHome.ts
chosta: yes, it's actually player id. I'll make it more readable.

src/components/GameHome.ts
chosta: This is used to keep the currently clicked "copy" button. The reason is that there should be only one message saying what is currently copied to clipboard, so `urlCopiedPlayerId` represents that. I will try to use better variable names and add a comment.

kberg (approved):

kberg:
Thank you!

PR #2290, opened 2021-01-07
Rename SelectHowToPayForCard to SelectHowToPayForProjectCard.

kberg:
This is to help clarify that this relates to paying the cost to play a card, versus paying to keep a card.


nwai90 (approved):

kberg:
Thanks!

PR #2291, opened 2021-01-07
turmoil party indicator

vsrisuknimit:
image

Indicator on the side bar for the ruling party. Not using the game party symbols since that is not used anywhere yet and we probably should avoid those copyrighted images. I have to truncate some party names to fit in the box.


vsrisuknimit:
There is no translation for those text yet. Not sure how to integrate i18n there and what are the appropriate translation that will fit in the box.

src/components/Preferences.ts
```diff
@@ -147,6 +156,25 @@ export const Preferences = Vue.component('preferences', {
...
+        return '';
+      }
+      return this.turmoil.ruling.toLowerCase().split(' ').join('_');
+    },
+    getRulingParty: function(): string {
+      const rulingPartyName = this.turmoil.ruling;
+      if (rulingPartyName === PartyName.MARS) {
+        return `Mars`;
+      } else if (rulingPartyName === PartyName.SCIENTISTS) {
+        return `Science`;
```
chosta, line 171: I know there is not enough space for 'Scientists' and 'Kelvinists' but it wold be nice to not rename them. Consider using the icons instead of text?
The color + icon combo should be enough imo.

src/styles/turmoil.less
```diff
@@ -221,6 +221,15 @@
     text-transform: uppercase;
     margin: 0 0 6px -3px;
 
+    &.party-name-indicator{
+        height: 20px;
+        line-height: 22px;
+        width: 50px;
+        margin: 10px 0 0px 3px;
+        font-size: 11px;
+        text-align: center;
```
chosta, line 230: I'd probably add a subtle border shadow so it pops a bit. `Kelvinits` needs it the most as it's lacking a contrast with the background.

chosta (approved):
A good improvement. I'm actually instinctively looking at the top left to see who's ruling only to get disappointed. Now it should be e a warm and fuzzy feeling instead :)

src/components/Preferences.ts
vsrisuknimit: "Not using the game party symbols since that is not used anywhere yet and we probably should avoid those copyrighted images."

src/styles/turmoil.less
vsrisuknimit: Tried adding border. Not satisfied with how it looks. I like the current clean design.

src/components/Preferences.ts
kevinb9n: FYI @ssimeonoff told me once that the icons are free and clear to use, just not things like the photos on cards, and certainly the app crossed that line long ago.

PR #2292, opened 2021-01-07
Refactored Card Drawing

sienmich:
Second part from https://github.com/bafolts/terraforming-mars/pull/2286#pullrequestreview-563040171
Now usuing @kberg's `options` when drawing cards. Seems nice.


src/deferredActions/DrawCards.ts
```diff
@@ -1,35 +1,116 @@
...
+      `Select at most ${DrawCards.cardString(this.options.amount)} to buy`,
+      'Buy',
+      cards,
+      (selected: Array<IProjectCard>) => {
+        selected = selected.slice(0, this.options.amount);
+        this.selected(selected, cards);
+        if (selected.length > 0) {
+          this.game.defer(new SelectHowToPayDeferred(
+            this.player, selected.length * this.player.cardCost,
+            false, false, 'Select how to pay for action'));
```
sienmich, line 68:
```suggestion
          this.game.defer(new SelectHowToPayDeferred(
            this.player, selected.length * this.player.cardCost,
            {title: 'Select how to pay for cards'}));
```

src/deferredActions/DrawCards.ts
```diff
@@ -1,35 +1,115 @@
...
+    );
+  }
+
+  private get specialRequirements(): boolean {
+    return !!(this.options.include || this.options.tag || this.options.resource || this.options.cardType);
+  }
 
   public execute() {
-    const drawnCards: Array<IProjectCard> = [];
+    const total = this.options.from || this.options.amount;
```
Lynesth, line 83: return if `this.options.amount <= 0`

src/deferredActions/DrawCards.ts
```diff
@@ -1,35 +1,115 @@
...
+      LogHelper.logDrawnCards(this.game, this.player, cards);
+    } else {
+      this.game.log('${0} drew ${1} card(s)', (b) => b
+        .player(this.player)
+        .number(cards.length),
+      );
+    }
+  }
+
+  private selectCardsToDraw(cards: Array<IProjectCard>) {
```
Lynesth, line 43: `selectCardsToKeep` ?

src/deferredActions/DrawCards.ts
```diff
@@ -1,35 +1,115 @@
...
+      (selected: Array<IProjectCard>) => {
+        selected = selected.slice(0, this.options.amount);
+        this.selected(selected, cards);
+        this.log(cards);
+        return undefined;
+      }, this.options.amount, this.options.amount,
+    );
+  }
+
+  private selectCardsToBuy(cards: Array<IProjectCard>) {
```
Lynesth, line 57: Where do you check that the player can buy the card(s) ?

src/deferredActions/DrawCards.ts
```diff
@@ -1,35 +1,115 @@
...
-        public tag?: Tags,
-  ) {}
+        public options: DrawCards.Options = {amount: 1},
+  ) { }
+
+  public static cardString(amount: number) {
+    return `${amount} card${amount === 1 ? '' : 's'}`;
+  }
+
+  private selected(selected: Array<IProjectCard>, from: Array<IProjectCard>) {
```
Lynesth, line 23: I'm not sure what this function does by reading its name, especially since it's also the name of a variable used quite often in this class.

src/deferredActions/DrawCards.ts
```diff
@@ -1,35 +1,115 @@
...
         public player: Player,
         public game: Game,
-        public count: number = 1,
-        public tag?: Tags,
-  ) {}
+        public options: DrawCards.Options = {amount: 1},
+  ) { }
+
+  public static cardString(amount: number) {
+    return `${amount} card${amount === 1 ? '' : 's'}`;
```
Lynesth, line 20: I think there's a reason we're using `card(s)` instead but someone will probably be able to tell you why, I don't remember (something to do with translation).

src/deferredActions/DrawCards.ts
```diff
@@ -1,35 +1,115 @@
...
+        this.game.log('${0} bought ${1}', (b) => b
+          .player(this.player)
+          .string(DrawCards.cardString(this.options.amount)));
+        return undefined;
+      }, this.options.amount, 0,
+    );
+  }
+
+  private get specialRequirements(): boolean {
+    return !!(this.options.include || this.options.tag || this.options.resource || this.options.cardType);
```
Lynesth, line 79: Would you mind using `Boolean(...)` instead of double negation `!!` ?
Or checking each of them against `undefined` ?

Lynesth (changes requested):

src/deferredActions/DrawCards.ts
Lynesth: Ok so I see it is done before the call to this.
It should probably be moved here, with a different title 'You cannot afford to buy a card' (or whatever) as it was before.

Lynesth (changes requested):
I also notice you sometimes left
```ts
player.drawCard( ... );
return undefined;
```
and sometimes changed it for
```ts
return player.drawCard( ... );
```
Could you standardize this (2nd variant) ?

src/deferredActions/DrawCards.ts
nwai90: We use it for ease of localisation so that we don't have to worry about plural forms at all. It would be too onerous to have to handle every noun (card, resource, etc.)

src/cards/base/InventionContest.ts
```diff
@@ -26,18 +26,7 @@ export class InventionContest extends Card implements IProjectCard {
...
-    ];
-    return new SelectCard('Select card to take into hand', 'Take', cardsDrawn, (foundCards: Array<IProjectCard>) => {
-      player.cardsInHand.push(foundCards[0]);
-      cardsDrawn
-        .filter((c) => c !== foundCards[0])
-        .forEach((c) => game.dealer.discard(c));
-      return undefined;
-    });
+  public play(player: Player, game: Game): SelectCard<IProjectCard> {
+    return player.drawCard(game, {amount: 1, from: 3})!;
```
kberg, line 30: ! is very, very much dissuaded (except in tests).

kberg (commented):
It's too late at night, so I can't review this reasonably well, but I noticed one thing.

src/deferredActions/DrawCards.ts
sienmich: Good idea, added that inside

src/cards/venusNext/SponsoredAcademies.ts
```diff
@@ -22,10 +21,10 @@ export class SponsoredAcademies implements IProjectCard {
 
     public play(player: Player, game: Game) {
       game.defer(new DiscardCards(player, game));
-      game.defer(new DrawCards(player, game, 3));
+      player.drawCard(game, {amount: 3});
       const otherPlayers = game.getPlayers().filter((p) => p.id !== player.id);
       for (const p of otherPlayers) {
-        game.defer(new DrawCards(p, game));
+        p.drawCard(game);
```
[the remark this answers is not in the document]
sienmich: @Lynesth, I think I'm gonna make sth like this (in `Player.ts`):
```ts
  public drawCard(game: Game, options?: DrawCards.Options) {
    game.defer(new DrawCards(this, game, options));
  }
```
Because in log we want first info which card was played and later how many cards were drawn. Am i right? Does it make sense?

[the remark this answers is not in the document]
sienmich: Or maybe, doesn't really matter. What do u prefer?
```ts
  public drawCard(game: Game, options?: DrawCards.Options) {
    game.defer(new DrawCards(this, game, options));
    return undefined;
  }
```

src/deferredActions/DrawCards.ts
```diff
@@ -1,35 +1,118 @@
...
+    }
+  }
+
+  private selectCardsToKeep(cards: Array<IProjectCard>) {
+    return new SelectCard(
+      `Select ${this.options.amount} card(s) to keep`,
+      'Select',
+      cards,
+      (selected: Array<IProjectCard>) => {
+        selected = selected.slice(0, this.options.amount);
```
Lynesth, line 46: I'm not sure but, why are we slicing the array here ?

src/deferredActions/DrawCards.ts
```diff
@@ -1,35 +1,118 @@
...
+      ...range(this.options.amount + 1)
+        .filter((n) => this.player.canAfford(n * this.player.cardCost)),
+    );
+    const info = howManyCanAfford !== this.options.amount ? ' (you cannot afford more)':'';
+    return new SelectCard(
+      `Select at most ${howManyCanAfford}${info} card(s) to buy`,
+      'Buy',
+      cards,
+      (selected: Array<IProjectCard>) => {
+        selected = selected.slice(0, howManyCanAfford);
```
Lynesth, line 65: Same question as above.

src/deferredActions/DrawCards.ts
```diff
@@ -1,35 +1,118 @@
...
+        if (selected.length > 0) {
+          this.game.defer(new SelectHowToPayDeferred(
+            this.player, selected.length * this.player.cardCost,
+            {title: 'Select how to pay for cards'}));
+        }
+        this.game.log('${0} bought ${1} card(s)', (b) => b
+          .player(this.player)
+          .number(selected.length));
+        return undefined;
+      }, this.options.amount, 0,
```
Lynesth, line 76: I think the maximum number of cards should be `howManyCanAfford` here instead of `this.options.amount`.

src/deferredActions/DrawCards.ts
```diff
@@ -1,35 +1,118 @@
...
+  }
+
+  private selectCardsToBuy(cards: Array<IProjectCard>) {
+    const howManyCanAfford = Math.max(
+      ...range(this.options.amount + 1)
+        .filter((n) => this.player.canAfford(n * this.player.cardCost)),
+    );
+    const info = howManyCanAfford !== this.options.amount ? ' (you cannot afford more)':'';
+    return new SelectCard(
+      `Select at most ${howManyCanAfford}${info} card(s) to buy`,
```
Lynesth, line 61: I know you probably wanted it to be future proof somehow but since all actions that allow you to buy a card only let you buy one, you'd probably better of checking we can afford 1 card and change the message completely if we can't.

If you really want to keep it that way, then please just change the title and the button if we can't buy cards at all. "Select at most 0" with a "Buy" button is strange.

src/deferredActions/DrawCards.ts
```diff
@@ -1,35 +1,118 @@
...
   public execute() {
-    const drawnCards: Array<IProjectCard> = [];
+    const total = this.options.from || this.options.amount;
+    if (total <= 0) return undefined;
+    const include = (card: IProjectCard) => (
+      (!this.options.tag || card.tags.includes(this.options.tag)) &&
+      (!this.options.resource || this.options.resource === card.resourceType) &&
+      (!this.options.cardType || this.options.cardType === card.cardType));
+
+    const cards = this.game.dealer.drawProjectCardsByCondition(this.game, total, this.options.include || include);
```
Lynesth, line 92: Should also probably add a failsafe after this one in case no cards were drawn (Celestic in a non venus game for example) to simply return.

Lynesth (changes requested):

src/cards/venusNext/SponsoredAcademies.ts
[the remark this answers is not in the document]
Lynesth: It should be fine to do yes. But it's late so I might be missing something.

src/deferredActions/DrawCards.ts
sienmich: If somebody is cheating, he could select more cards than he should (the restriction is on client side, he could just upload custom POST to server or select more cards modifying some code). This slice is server side verification, if there are more cards selected then intended I just slice them.
I'm not sure if this is necessary. Should i remove this?

src/deferredActions/DrawCards.ts
sienmich: Ok, I'll change that.
I'm thinking of exporting this part and applying it to buying cards after draft (from 4), I'm not sure if those prices are checked there.

src/deferredActions/DrawCards.ts
sienmich: Thanks, fixed

src/deferredActions/DrawCards.ts
sienmich: Done

src/cards/venusNext/SponsoredAcademies.ts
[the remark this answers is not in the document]
sienmich: I changed all to `return player.drawCard( ... );`

src/deferredActions/DrawCards.ts
Lynesth: All checks are already performed server side in `Player.runInput()` so you can remove it from here.

sienmich: Ok, thanks!

sienmich: Done

src/cards/venusNext/SponsoredAcademies.ts
```diff
@@ -21,7 +22,7 @@ export class SponsoredAcademies implements IProjectCard {
 
     public play(player: Player, game: Game) {
       game.defer(new DiscardCards(player, game));
-      player.drawCard(game, {amount: 3});
+      game.defer(new DrawCards(player, game, {amount: 3}));
       const otherPlayers = game.getPlayers().filter((p) => p.id !== player.id);
       for (const p of otherPlayers) {
         p.drawCard(game);
```
Lynesth, line 28: Not sure it really matters but I think active player usually prefers drawing before the others, so probably leave this as a deferred action as well.

src/Player.ts
```diff
@@ -994,42 +994,12 @@ export class Player implements ISerializable<SerializedPlayer> {
...
-      ), () => {
-        payForCards();
-      },
-    );
+    game.defer(new DrawCards(this, game, {
+      amount: dealtCards.length,
+      isBuying: true,
+      cards: dealtCards,
+      cb: () => game.playerIsFinishedWithResearchPhase(this),
+    }));
```
Lynesth, line 1002: Have you checked that this is actually working ?
What part of the code runs all deferred actions after calling each player's `runResearchPhase` ?

src/Player.ts
```diff
@@ -1223,6 +1193,10 @@ export class Player implements ISerializable<SerializedPlayer> {
       }
     }
 
+    if (selectedCard.cardType !== CardType.PROXY) {
+      this.addPlayedCard(game, selectedCard);
+    }
+
```
Lynesth, line 1199: Moving this here will probably break a lot of things we don't have tests for.
If you just want to put the "Red played this card" log before playing the card, move the log here but leave adding the card to the played cards where it was before.

Lynesth (changes requested):

src/Player.ts
sienmich: Ok, right

src/Player.ts
sienmich: You are right, it's not working. I was testing it, but changed to `game.defer` instead of `player.setWaitingFor` just before the commit. Good spot, thanks!

sienmich: I don't know how to get around the fact that my `drawCard` returns `SelectCard<IProjectCard> | undefined`. In this case it's clearly the first one. Could I just cast it (`as SelectCard<IProjectCard>`)?
Or is there a way to tell TypeScript that when `options.isBuying === true` then return type is not undefined?
I could also make `setWaitingFor` accept `undefined`, but that's probably too ugly.

sienmich: Ok, I added an if, seems the best solution to me

src/Player.ts
```diff
@@ -1193,9 +1193,7 @@ export class Player implements ISerializable<SerializedPlayer> {
       }
     }
 
-    if (selectedCard.cardType !== CardType.PROXY) {
-      this.addPlayedCard(game, selectedCard);
-    }
+    game.log('${0} played ${1}', (b) => b.player(this).card(selectedCard));
```
Lynesth, line 1196: This should not be logged if the card is of type PROXY.

Lynesth (changes requested):

src/Player.ts
Lynesth: Unfortunately, you can't tell TS what you'd like. Because the `cards` you are passing to the function might be an empty array for some reason (since `this.dealCards()` can return an empty array and `this.draftedCards` could be empty as well as far as TS knows), in which case it'll return undefined anyway.

src/inputs/SelectAmount.ts
```diff
@@ -8,7 +8,7 @@ export class SelectAmount implements PlayerInput {
     constructor(
         public title: string | Message,
         public buttonLabel: string = 'Save',
-        public cb: (amount: number) => undefined,
+        public cb: (amount: number) => undefined | PlayerInput,
```
Lynesth, line 11: What are this change...

src/inputs/SelectOption.ts
```diff
@@ -1,19 +1,13 @@
...
-import {SelectCard} from './SelectCard';
-import {ICard} from '../cards/ICard';
 
 export class SelectOption implements PlayerInput {
     public inputType: PlayerInputTypes = PlayerInputTypes.SELECT_OPTION;
     constructor(
         public title: string | Message,
         public buttonLabel: string = 'Select',
-        public cb: () => SelectSpace | SelectHowToPay | AndOptions | SelectOption | SelectAmount | SelectCard<ICard> | undefined) {
+        public cb: () => PlayerInput | undefined) {
```
Lynesth, line 10: ... and this one for ?

sienmich: @Lynesth, After changing the
```ts
player.drawCard(game);
return undefined;
```
into
```ts
return player.drawCard(game);
```
a lot of functions no longer are returning undefined (because `drawCard(...): undefined | SelectCard<IProjectCard>`).

I see no harm (yet) in possibility of different PlayerInputs callbacks to return other PlayerInputs. Is there one?

`SelectOption` signature is not a problem, but that `SelectAmount` change would be helpful.

Lynesth (approved):
This look ok to me now.
I'd still like 1 more pair of eyes to take a look though if possible @bafolts @kberg

src/deferredActions/DrawCards.ts
```diff
@@ -1,35 +1,119 @@
...
 import {Tags} from '../cards/Tags';
 import {IProjectCard} from '../cards/IProjectCard';
 import {DeferredAction} from './DeferredAction';
+import {SelectCard} from '../inputs/SelectCard';
+import {ResourceType} from '../ResourceType';
+import {CardType} from '../cards/CardType';
+import {SelectHowToPayDeferred} from './SelectHowToPayDeferred';
+import {LogHelper} from '../LogHelper';
 
 export class DrawCards implements DeferredAction {
```
kberg, line 12: OK, so I'm going to give some feedback on this. This is an improvement, but there's a further improvement to consider. Please excuse me if you know about these ideas already.

There's a notion in OO called the [Single Responsibility Principle](https://en.wikipedia.org/wiki/Single-responsibility_principle) which basically says that one class should do one thing well. Now, drawCard isn't a class, it's a method, but there's a similar idea about methods, which is kind of the notion of [coupling](https://en.wikipedia.org/wiki/Coupling_(computer_programming)). This method really does several things:

1. It draws `from` cards from the project deck.
2. It lets a player select up to `amount` cards
3. It makes the player pay for them if configured.

In other words, I expect something called 'drawCards' to do kind of one thing: **Draw Cards.**

I'm not *entirely* there's a better approachin what the right approach is yet, but I want to think about it. In some ways, imagine an API a little different from this (don't go changing things, I'm just thinking.) But the fact that all the options are mixed into one object (rules for drawing, rules for keeping, rules for buying) makes me cautious.

src/deferredActions/DrawCards.ts
```diff
@@ -1,35 +1,119 @@
...
+      return this.selectCardsToKeep(cards);
     }
+  }
+}
 
-    // Add the cards to the player's hand
-    this.player.cardsInHand.push(...drawnCards);
-    return undefined;
+export namespace DrawCards {
+  export interface Options {
```
kberg, line 108: I suggest renaming a few things to disambiguate:

1. change `amount` to `count`. Amount can be mistaken for a MC amount
2. change `from` to `keep`, and invert its responsibility as a parameter: Now, 'keep' is always the lower value, never `count` nee `amount`. "Draw 2 from 4" sounds like drawing 2 cards from a deck of 4 cards, not drawing 4 cards from a deck and keeping 2.
3. `isBuying` can be easier to read as `paying`

Can you please add a comment describing what cards is, as a parameter?

src/deferredActions/DrawCards.ts
```diff
@@ -1,35 +1,119 @@
...
+
+  private log(cards: Array<IProjectCard>) {
+    if (this.specialRequirements) {
+      LogHelper.logDrawnCards(this.game, this.player, cards);
+    } else {
+      LogHelper.logCardChange(this.game, this.player, this.options.isBuying ? 'bought' : 'drew', cards.length);
+    }
+  }
+
+  private runCb() {
```
kberg, line 36: Can be simplified to this.options.cb?() and then you can remove this function?

src/deferredActions/DrawCards.ts
```diff
@@ -1,35 +1,119 @@
...
+        this.drawSelectedDiscardOthers(selected, cards);
+        this.log(selected);
+        this.runCb();
+        return undefined;
+      }, this.options.amount, this.options.amount,
+    );
+  }
+
+  private selectCardsToBuy(cards: Array<IProjectCard>) {
+    const howManyCanAfford = Math.min(this.options.amount, Math.floor(this.player.spendableMegacredits() / this.player.cardCost));
```
kberg, line 57: You can just call this `max`

src/deferredActions/DrawCards.ts
```diff
@@ -1,35 +1,119 @@
...
+    const info = howManyCanAfford !== this.options.amount ? `at most ${howManyCanAfford} (you cannot afford more) `:'';
```
kberg, line 58: I suggest keeping this a little simpler:

```
let msg = max === 0 ?
  'You cannot afford any cards' :0
  `Select up to ${max} cards to buy for ${this.player.cardCost} MC each`;

return new SelectCard(
  msg,
  ...
```

Having part of the initial message logic here, and even more down in the `SelectCard` constructor makes it harder to read.

src/deferredActions/DrawCards.ts
```diff
@@ -1,35 +1,119 @@
...
+    return new SelectCard(
+      howManyCanAfford === 0 ? `You cannot afford any cards` : `Select ${info}card(s) to buy`,
+      howManyCanAfford === 0 ? 'Ok' : 'Buy',
+      cards,
+      (selected: Array<IProjectCard>) => {
+        this.drawSelectedDiscardOthers(selected, cards);
+        if (selected.length > 0) {
```
kberg, line 65: Can the statement above be run only after the payment? Can be a TODO.

src/inputs/SelectAmount.ts
kberg: Can you respond to this question, please?

src/deferredActions/DrawCards.ts
kberg, line 12: This is now a pretty critical and meaty bit of code, particularly the varied amount of options for drawing and selectig cards. What do you think would be the right tests to make you confident this worked?

kberg (changes requested):
By and large a good PR, but i have a few thoughts and comments.

src/deferredActions/DrawCards.ts
sienmich: I mostly agree.
I think some kind of a solution would be to make more smaller classes, like `DrawCards`, `BuyCards`, `SelectCardsToKeep` etc. And make the `DrawCards` use thaose smaller ones and rename it to some kind of `AdvancedDrawCards` or `GetNewCards`.
I'd prefer to leave it as it is and aim to refactor with OO rule our giants like `Player` or `Game`

src/inputs/SelectAmount.ts
sienmich: Here: https://github.com/bafolts/terraforming-mars/pull/2292#discussion_r553979583

src/deferredActions/DrawCards.ts
sienmich: `this.options.cb?()` is not working. However, `this.options.cb?.();` is. Turns out, that `?.` is the overloaded operator. https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining

src/deferredActions/DrawCards.ts
sienmich: It's tested via all cards test, like `BusinessContacts` and `BusinessNetwork`, so it should be all right. Should I add separate test file?

kberg: I think so, given how critical this is.

src/cards/venusNext/MorningStarInc.ts
```diff
@@ -15,11 +14,8 @@ export class MorningStarInc implements CorporationCard {
     public cardType = CardType.CORPORATION;
 
     public initialActionText: string = 'Draw 3 Venus-tag cards';
-    public initialAction(player: Player, game: Game) {
-      const cards = game.drawCardsByTag(Tags.VENUS, 3);
-      player.cardsInHand.push(...cards);
-      LogHelper.logDrawnCards(player, cards);
-      return undefined;
+    public initialAction(player: Player) {
+      return player.drawCard(3, {tag: Tags.VENUS});
```
kberg, line 18: So there's something funny about this being a return value`drawCards` reads like an immediate action, so what is the return value? If it's not obvious, if you have to go look up a couple of functions in to figure it out, we need something better.

What is the return value of drawCard?

If it's a deferred action, then I suggest calling drawCard "drawCardDeferred" etc.

src/deferredActions/DrawCards.ts
```diff
@@ -1,35 +1,114 @@
...
+    discard.forEach((card) => {
+      if (preserve.find((f) => f.name === card.name) === undefined) {
+        player.game.dealer.discard(card);
+      }
+    });
+  }
+
+  export function choose(player: Player, cards: Array<IProjectCard>, options: DrawCards.ChooseOptions): SelectCard<IProjectCard> {
+    let max = options.keepMax || cards.length;
+    console.log(max, options.keepMax, cards.length);
```
kberg, line 82: Remove before flight.

src/deferredActions/DrawCards.ts
```diff
@@ -1,35 +1,114 @@
...
+    if (options.paying) {
+      max = Math.min(max, Math.floor(player.spendableMegacredits() / player.cardCost));
+      console.log(max, player.spendableMegacredits(), player.cardCost);
```
kberg, line 85: This too.

tests/defferedActions/DrawCards.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  beforeEach(function() {
+    player = TestPlayers.BLUE.newPlayer();
+    const redPlayer = TestPlayers.RED.newPlayer();
+    Game.newInstance('foobar', [player, redPlayer], player);
+  });
+
+  function check(expectedCardsInHand: number, expectedDiscardedCards: number) {
+    expect(player.cardsInHand.length).to.eq(expectedCardsInHand);
+    expect(player.game.dealer.discarded.length).to.eq(expectedDiscardedCards);
```
kberg, line 25: expect(player.game.dealer.discarded).has.length(expectedDiscardedCards)

tests/defferedActions/DrawCards.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  it('keeps cards', function() {
+    DrawCards.keep(player, [cards[0], cards[1]]);
+    check(2, 0);
```
kberg, line 30: Tell me if you think this is more readable. Up top, have 

`let dealer: Dealer` and initialize to `game.dealer`

And then replace calls to `check` with

```
expect(player.cardsInHand).have.length(2);
expect(dealer.discardedCards).have.length(0);
```

I think people will be able to understand that more than check(2, 0). Sometimes there's better options than terse, but you tell me.

kberg (commented):
Couple of things. This is looking pretty good.

tests/defferedActions/DrawCards.spec.ts
sienmich: Agree, changed

src/cards/venusNext/MorningStarInc.ts
sienmich: I just added the return types as template, now it looks like this:
```ts
  public drawCard(count?: number, options?: DrawCards.DrawOptions): undefined {
    return DrawCards.keepAll(this, count, options).execute();
  }

  public drawCardKeepSome(count?: number, options?: DrawCards.AllOptions): SelectCard<IProjectCard> {
    return DrawCards.keepSome(this, count, options).execute();
  }
```
It makes a lot of sense, because if we just draw cards, then we do not need additional input, and we are buying / choosing, then we need to Select those Cards.

sienmich: `return player.drawCard(3, {tag: Tags.VENUS});` is just a shortcut from
```
player.drawCard(3, {tag: Tags.VENUS});
return undefined;
```
, as @Lynesth suggested above

Lynesth: I suggested it since it can (maybe that changed now, haven't checked) return either undefined or a SelectCard input. So it makes sense to simply have it like that.

kberg: > `return player.drawCard(3, {tag: Tags.VENUS});` is just a shortcut from

I'll ask you to restore the original, because I think it's not clear. *IF* you're going to actually return a SelectInput instead, then I'd ask for a name more appropriate "draw cards" is a statement to draw cards, which this does most of the time.  

@Lynesth a related side question: what's the difference between returning SelectInput and calling game.defer? Is this basically two ways of doing the same thing? If so, is it right to drop one?

kberg (commented):
Modulo my one comment about return values, I think this is good to go.

I would approve the PR now, except that I think if you make a change based on my last comment , it's worth just reading it to catch issues first. Otherwise, we're basically done.

kberg (approved):

kberg:
This is really good. It's a net change of -100 lines, and that's WITH tests.

src/cards/venusNext/MorningStarInc.ts
Lynesth: @kberg If the returned value of a `play` call (or an `action` one) is not undefined, it will be deferred using `game.defer()`.

So you can return a SelectInput if that's the only thing the `play` method needs to prompt the player.

But if you have to prompt the player for multiple things (place an ocean and draw cards, discard then draw cards, etc) then you have to use game.defer directly inside the method.

PR #2295, opened 2021-01-07
another batch of cards for less memory

bafolts:
Few more cards converted to the less memory pattern.

### DIFF
--- a/src/cards/base/Tardigrades.ts
+++ b/src/cards/base/Tardigrades.ts
@@ -1,23 +1,37 @@
 import {IProjectCard} from '../IProjectCard';
 import {Tags} from '../Tags';
+import {Card} from '../Card';
 import {CardType} from '../CardType';
 import {ResourceType} from '../../ResourceType';
 import {CardName} from '../../CardName';
 import {IResourceCard} from '../ICard';
 import {Player} from '../../Player';
-import {CardMetadata} from '../CardMetadata';
 import {CardRenderer} from '../render/CardRenderer';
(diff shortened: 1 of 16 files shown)


src/cards/base/Tardigrades.ts
```diff
@@ -1,23 +1,37 @@
...
 import {Player} from '../../Player';
-import {CardMetadata} from '../CardMetadata';
 import {CardRenderer} from '../render/CardRenderer';
 import {CardRenderItemSize} from '../render/CardRenderItemSize';
 import {CardRenderDynamicVictoryPoints} from '../render/CardRenderDynamicVictoryPoints';
 
-export class Tardigrades implements IProjectCard, IResourceCard {
-    public cost = 4;
-    public resourceType = ResourceType.MICROBE;
-    public resourceCount: number = 0;
```
chosta, line 16: I'm just noticing the missing resourceCount although I'm not sure if it could be skipped.

chosta (approved):

src/cards/base/Tardigrades.ts
bafolts: The `resourceCount` needs to change per instance of the class so it is not included with the properties which are static. The `resourceCount` property is defined still:

https://github.com/bafolts/terraforming-mars/pull/2295/files/162164789651bb4da581c8be00ee87ac366a33f9#diff-7fb830981ccdfca13d906120e3b877fc9f5f62d0bfe2c7d72c33a7dde4e4a6afR34

PR #2296, opened 2021-01-07
Fixes the notice stuck to the bottom right

chosta:
`Not affiliated with FryxGames, Asmodee Digital or Steam in any way.`

is now absolutely positioned and reasonable in almost all screens. General layout could be further improved though in another PR.


kberg:
Thanks.

kberg (approved):

PR #2297, opened 2021-01-07
new version of metal bonus indicator

vsrisuknimit:
New version of the metal bonus indicator as proposed by @ssimeonoff .

Show the value of the metal as a superscript next to the metal icon. Only show the value that is above 2 for steel or above 3 for Ti

Before Unity round. 
image

During Unity round.
image


chosta (approved):
Much better.

kberg (commented):
I LOVE IT.

Issue #2299, opened 2021-01-08
The handicap field comes from the new game form as a string


kberg:
For example, this came from a sample HTTP request:

```
 "players": [
        {
            "index": 1,
            "name": "f",
            "color": "red",
            "beginner": false,
            "handicap": 0,
            "first": false
        },
        {
            "index": 2,
            "name": "g",
            "color": "green",
            "beginner": false,
            "handicap": "4",
            "first": true
        }
```

Notice how the first player has 0, a number, but the second one has "4", a string.

kberg:
@chosta you think this is something you can fix?

PR #2300, opened 2021-01-08
Fix hanidicap field, which was a string for quite some time.

kberg:
…ient as a string.

Related to #2299

### DIFF
--- a/src/components/CreateGameForm.ts
+++ b/src/components/CreateGameForm.ts
@@ -725,7 +725,7 @@ export const CreateGameForm = Vue.component('create-game-form', {
                                                 </label>
             
                                                 <label class="form-label">
-                                                    <input type="number" class="form-input form-inline player-handicap" value="0" min="0" :max="10" v-model="newPlayer.handicap" />
+                                                    <input type="number" class="form-input form-inline player-handicap" value="0" min="0" :max="10" v-model.number="newPlayer.handicap" />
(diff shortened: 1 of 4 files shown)


chosta:
Yeah, I think it's a Vue thing converting to strings. 
The Number() conversion doesn't hurt anyway.

chosta (approved):

src/components/CreateGameForm.ts
```diff
@@ -725,7 +725,7 @@ export const CreateGameForm = Vue.component('create-game-form', {
                                                 </label>
             
                                                 <label class="form-label">
-                                                    <input type="number" class="form-input form-inline player-handicap" value="0" min="0" :max="10" v-model="newPlayer.handicap" />
+                                                    <input type="number" class="form-input form-inline player-handicap" value="0" min="0" :max="10" v-model.number="newPlayer.handicap" />
```
bafolts, line 728: I am pretty sure that vue does not care that we used `type="number"` here. Eventually this value is pulled using `domElement.value` which will be a string.

This is seen by the following:

```javascript
var n = document.createElement("input");
n.type = "number";
n.value = 4
console.log(typeof n.value)
> string
```

kberg: > I am pretty sure that vue does not care that we used `type="number"` here. Eventually this value is pulled using `domElement.value` which will be a string.
> 
> This is seen by the following:
> 
> ```js
> var n = document.createElement("input");
> n.type = "number";
> n.value = 4
> console.log(typeof n.value)
> > string
> ```

I have verified by hand, by setting a breakpoint in the server itself, that by changing this the body returns a number instead of a string.

PR #2304, opened 2021-01-08
Changed productionBox to production

chosta:
It's a global name change.


kberg (approved):
That was SIMPLE. Thanks!

PR #2305, opened 2021-01-08
Requirements to unity

chosta:
Added a shorthand to requirements:
`b.tag(Tags.VENUS).tag(Tags.EARTH).tag(Tags.JOVIAN)` becomes `unity()`

### DIFF
--- a/src/cards/CardRequirements.ts
+++ b/src/cards/CardRequirements.ts
@@ -105,6 +105,13 @@ class Builder {
     return this;
   }
 
+  public unity(): Builder {
+    this.reqs.push(new TagCardRequirement(Tags.VENUS, -1));
+    this.reqs.push(new TagCardRequirement(Tags.EARTH, -1));
+    this.reqs.push(new TagCardRequirement(Tags.JOVIAN, -1));
+    return this;
+  }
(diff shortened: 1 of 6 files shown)


kberg:
Is there an historical reason for this?

chosta:
> Is there an historical reason for this?

Mostly reducing code repetition but I can give you two other reasons (more personal):
1) When adding the cards with the triple tag requirement I always wondered which tag came first / second / third so I had to check. I figured it should be automated so I don't have to think about it.
2) It's thematic. When unity is ruling you get 1 MC for each of the 3 tags.

kberg:
> > Is there an historical reason for this?
> 
> Mostly reducing code repetition but I can give you two other reasons (more personal):
> 
> 1. When adding the cards with the triple tag requirement I always wondered which tag came first / second / third so I had to check. I figured it should be automated so I don't have to think about it.
> 2. It's thematic. When unity is ruling you get 1 MC for each of the 3 tags.

Gotcha. My opinion is this is probably not worthwhile. Do you have strong feelings about this?

chosta:
No, I can live without it :)

PR #2312, opened 2021-01-09
Move game.getSpace to board.getSpace where it belongs.

kberg:
This makes it available for Board subclasses, (like MoonBoard!)

PR #2313, opened 2021-01-09
Add Moon spaces, tiles, board, basic data.

kberg:
Relies on #2312


src/moon/MoonBoard.ts
```diff
@@ -0,0 +1,95 @@
...
+  public getSpacesWithTile(tileType: TileType): Array<ISpace> {
+    return this.spaces.filter((space) => space.tile?.tileType === tileType);
+  }
+
+  public static newInstance(): MoonBoard {
+    const spaces: Array<ISpace> = [];
+
+    spaces.push(Space.colony(MoonSpaces.LUNA_TRADE_STATION));
+    // y = 0
+    spaces.push(Space.mine('m02', 2, 0));
```
bafolts, line 60: When these ids can be linked with `MoonSpaces` leveraging the `MoonSpaces` will ensure we keep track of these. I would use them for any id which is tracked by `MoonSpaces`.

kberg: I can't parse this comment.

bafolts (approved):
Minor comment to use the `MoonSpaces` wherever the corresponding string such as `m02` is used.

The moon is the first milestone on the road to the stars.

src/moon/MoonBoard.ts
bafolts: For example you can replace `m02` with `MoonSpaces.MARE_IMBRIUM`. Any string that is associated with the `MoonSpaces` enum I would leverage the enum.

kberg:
Aha, that makes sense. Thanks.

PR #2320, opened 2021-01-09
Turn valid victory point types into a string-limited type.

kberg:


bafolts:
Something strange here. Seems there is nothing to merge. Commit may have gone in another merge already.

kberg:
ta

PR #2322, opened 2021-01-09
Change CardMetadata test to not play solo mode.

kberg:
That sets up forests and cities that would be considered worth victory points according to some cards,
and makes these tests less brittle.

### DIFF
--- a/tests/cards/CardMetadata.spec.ts
+++ b/tests/cards/CardMetadata.spec.ts
@@ -10,7 +10,9 @@ describe('CardMetadata', function() {
 
   beforeEach(function() {
     player = TestPlayers.BLUE.newPlayer();
-    game = Game.newInstance('foobar', [player], player);
+    const redPlayer = TestPlayers.RED.newPlayer();
+    game = Game.newInstance('foobar', [player, redPlayer], player);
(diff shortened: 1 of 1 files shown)


tests/cards/CardMetadata.spec.ts
```diff
@@ -10,7 +10,9 @@ describe('CardMetadata', function() {
 
   beforeEach(function() {
     player = TestPlayers.BLUE.newPlayer();
-    game = Game.newInstance('foobar', [player], player);
+    const redPlayer = TestPlayers.RED.newPlayer();
+    game = Game.newInstance('foobar', [player, redPlayer], player);
+    console.log(game.getCitiesInPlay());
```
bafolts, line 15: Remove before flight. Remains of debugging code.

bafolts (approved):

tests/cards/CardMetadata.spec.ts
kberg: You can do that with the edit dropdown if you are ever so inclined.

image

kberg: image

Issue #1285, opened 2020-09-13
Feature request: show mc during initial selection


vsrisuknimit:
During initial card purchase (after initial drafting is done), once the user select a corporation. Show how much megacredit the user has from that corporation. After click on each prelude and project card, deduct or increase the mc accordingly. This will help users calculate how many cards to keep or what prelude to pick.

vsrisuknimit:
![FireShot Capture 158 - Terraforming Mars - terraforming-mars](https://user-images.githubusercontent.com/14239220/93021918-b56d1f00-f5b3-11ea-889b-06544106ab1c.png)

pierrehilbert:
I will work on it. Should it take into account preludes bonus/malus?
I think it should but would like your advice.

vsrisuknimit:
@pierrehilbert Yes, it should take into account prelude that provide and deduct cash.

vsrisuknimit:
It should also handle Polyphemos and Terralabs differently.

I don't know if it should take into account the extra mc from corp immediate reveal bonus (Vitor and Splice and Pharmacy Union).

kevinb9n:
I think this would be an excellent feature. Some thoughts:

For any MC amount that is guaranteed to be given or deducted by Corp reveal effects, I believe the value *should* account for those adjustments. If it doesn't, I think it will violate user expectations and they will file bugs. And they'd be right to: consider that whether Vitor had been written as "45 MC, effect includes itself" or "48 MC, effect excludes itself" is only a *wording* choice that has no in-game consequence.

The number you calculate might be negative. If so, I would still just display it. It will remind the player to adjust their selections until it is at least zero. (That still doesn't guarantee you can afford your project cards, which you have to buy before Prelude benefits, so you might still get an error when you try to start. That may be a little odd but I think it doesn't represent a real problem. The remedy would be to show *both* pre-Preludes and post-Preludes values, but that's just too much information 99.99% of the time.)

It *could* display a number even with no Corp selected, which would probably be negative if you select your project cards first, but I agree that there's not much value in this, and it seems cleaner to just not show it yet. Incidentally, when deciding whether to show this number, please check whether the number of Corporations is *exactly* one. That way it will function properly whether the request in bug 1400 ever happens or not. I see no need to check that Preludes==2 though.

kevinb9n:
Pharmacy Union makes a good example of the importance of taking corp reveal effects into account. That player would be unhappy to discover they have 8 less money available than they thought.

[...]

PR #2324, opened 2021-01-09
use SelectInitialCards for more specific user interface

bafolts:
Calling all user interface people, @kevinb9n , @chosta , @nwai90 , @Lynesth . I have the work in place to display the number of mega credits remaining on the card selection screen. Want to keep this PR focused on the part to display the number. 

It currently looks like:

image

I am finishing up the code related change so figured I would get the UI portion up early as I don't often dabble in the front end.

I can't find the issue for this but there was an issue at some point. Helpful if someone could link it.

This should open up more flexibility for making changes to this screen.


src/server/ServerModel.ts
```diff
@@ -258,6 +258,7 @@ function getWaitingFor(
   switch (waitingFor.inputType) {
   case PlayerInputTypes.AND_OPTIONS:
   case PlayerInputTypes.OR_OPTIONS:
+  case PlayerInputTypes.SELECT_INITIAL_CARDS:
     playerInputModel.options = [];
     for (const option of (waitingFor as AndOptions | OrOptions)
```
sienmich, line 263: Here the `waitingFor as AndOptions | OrOptions` now seems wrong to me, right?

sienmich:
I like this very much, having PlayerInput in Vue looks nice to me.

src/server/ServerModel.ts
bafolts: You are correct. I will update `PlayerInput` and add an `options?: Array<PlayerInput>`. Then this check can be updated to check that `waitingFor.options !== undefined`. The cast to `AndOptions | OrOptions` is being used since that type has the `options` property.

Lynesth: Since SelectInitialCards extends AndOptions and doesn't add anything to it, it seems acceptable to me.

src/components/SelectInitialCards.ts
```diff
@@ -0,0 +1,87 @@
...
+    },
+    corporationChanged: function(cards: Array<CardName>) {
+      this.selectedCorporation = new CardFinder().getCorporationCardByName(cards[0]);
+    },
+    preludesChanged: function(cards: Array<CardName>) {
+      this.selectedPrelude = cards;
+    },
+  },
+  template: `<div>
+    <div v-if="selectedCorporation">starting mega credits: {{getRemainingMegaCredits()}}</div>
```
Lynesth, line 80: Maybe something like "You will start the game with " and the number in the MC icon ?

nwai90: Agree with this, or `Starting Megacredits: {NUMBER} in MC box`

src/components/SelectInitialCards.ts
```diff
@@ -0,0 +1,87 @@
...
+      selectedCorporation: undefined as CorporationCard | undefined,
+      selectedPrelude: [] as Array<CardName>,
+    };
+  },
+  methods: {
+    getRemainingMegaCredits: function() {
+      if (this.selectedCorporation === undefined) {
+        return NaN;
+      }
+      return this.selectedCorporation.startingMegaCredits - (this.selectedCards.length * 3);
```
nwai90, line 52: This should ideally be `(this.selectedCards.length * player.cardCost);` to account for Polyphemos (5 MC to buy each card) and Terralabs (1 MC to buy each card).

However since corporation card has not yet been played at this point one possibility might be to check if `this.selectedCorporation.name === CardName.POLYPHEMOS` or `this.selectedCorporation.name === CardName.TERRALABS_RESEARCH` in order to make the necessary adjustments to player's remaining MC

src/components/SelectInitialCards.ts
vsrisuknimit: I would consider putting it at the bottom next to `Start` instead of the top.

src/components/SelectInitialCards.ts
vsrisuknimit: While checking the corp, please also check for certain corps that increases/decrease mc upon reveal. (Splice, Vitor, Pharma Union)
https://github.com/bafolts/terraforming-mars/issues/1285#issuecomment-756082899

chosta (commented):
Great stuff. This approach could be used elswhere too.

src/components/SelectInitialCards.ts
```diff
@@ -0,0 +1,87 @@
...
+  },
+  data: function() {
+    return {
+      selectedCards: [] as Array<CardName>,
+      selectedCorporation: undefined as CorporationCard | undefined,
+      selectedPrelude: [] as Array<CardName>,
+    };
+  },
+  methods: {
+    getRemainingMegaCredits: function() {
```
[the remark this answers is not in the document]
bafolts: No reason in particular. I _think_ typescript will calculate this for us if we have only returned `number` so it is unnecessary to manually specify.

src/components/SelectCard.ts
```diff
@@ -44,6 +45,11 @@ export const SelectCard = Vue.component('select-card', {
     Card,
     Button,
   },
+  watch: {
+    cards: function() {
+      this.$emit('cardschanged', this.getData());
```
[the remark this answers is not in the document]
bafolts: Yes. This ended up pretty great. With vue it seems you have to look at available options for the goal. Probably worth repeating. We are only left with vue here because two years ago I decided to use vue because I had used react and angular and wanted to learn vue. I have never used these events so grateful I am still learning vue after 2 years!

src/inputs/SelectInitialCards.ts
```diff
@@ -0,0 +1,54 @@
...
+      );
+
+      if (game.gameOptions.preludeExtension) {
+        this.options.push(
+          new SelectCard(
+            'Select 2 Prelude cards', undefined, player.dealtPreludeCards,
+            (preludeCards: Array<IProjectCard>) => {
+              player.preludeCardsInHand.push(...preludeCards);
+              return undefined;
+            }, 2, 2,
```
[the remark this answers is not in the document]
bafolts: I may be misinterpreting the question but with this change none of this code is duplicated in `src/inputs/AndOptions`. Which block is this duplicating? Using a `2` for the number of preludes selected would make sense as a constant but I am not sure where else it is used outside of `SelectInitialCards` now.

src/components/SelectInitialCards.ts
bafolts: Once the initial component is in place we can make improvements later. I will most likely miss at least one of these corporation cards but will try to get them all in first commit.

src/inputs/SelectInitialCards.ts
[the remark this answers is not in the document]
chosta: My bad. I was actually referring to `pickCorporationCard` but I now see that `SelectInitialCards` should simplify it and there will be no code repetition.

src/components/SelectCard.ts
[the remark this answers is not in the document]
chosta: Indeed. This project has been a great learning opportunity for me too.

src/components/SelectInitialCards.ts
alrusdi: Aaaand i18n please!

kevinb9n:
Whoops, missed the notification. 

I don't object to launch-and-iterate, but since so many people use the bleeding edge, it's ideal *if possible* if the exact meaning of the number doesn't have to change once released. Since users will get used to what they think the number means.

My comments on that were here: https://github.com/bafolts/terraforming-mars/issues/1285#issuecomment-699694482

But ignore me if it would cause any problem.

In terms of UI I think it would be great if the number were displayed nearer to the Start button.

bafolts:
@kevinb9n this project does seem to be gaining an ever increasing user base. 

> I don't object to launch-and-iterate, but since so many people use the bleeding edge, it's ideal if possible if the exact meaning of the number doesn't have to change once released. Since users will get used to what they think the number means.

The goal will be to show the total after the corporations effects were applied.

I will merge the underlying server side change first and then introduce the UI change in a separate PR. This will allow us to focus on how the adjustments are calculated in one PR and keep the behind the scenes changes separate.

bafolts:
I haven't forgotten about this! With MLK weekend in US I plan on finishing this during the weekend.

src/components/CardOrderStorage.ts
```diff
@@ -6,7 +6,7 @@ const STORAGE_PREFIX = 'cardOrder';
 export class CardOrderStorage {
   public static getCardOrder(playerId: string): {[x: string]: number} {
     try {
-      const order = localStorage.getItem(`${STORAGE_PREFIX}${playerId}`);
+      const order = typeof localStorage === "undefined" ? null : localStorage.getItem(`${STORAGE_PREFIX}${playerId}`);
```
bafolts, line 9: This makes unit tests simpler which do not have local storage.

bafolts:
This should be ready for review. Inspired by the recent client side tests added by @kberg I added tests for this new component. Once this is merged I will do the subsequent change to start listing the remaining mega-credits. This change in place will allow us to customize this important page.

chosta (approved):
Looking good to me. I've pulled locally and tried it a couple of times. Can't wait to see the discount changing on card-change. This approach could be used in the future for similar improvements.

bafolts:
I also tried this multiple times and added the unit tests for extra sanity. Given this UI will be hit during most games though if something does go wrong I am glad this change is going out first. Now on to the fun parts that this change allows!

PR #2326, opened 2021-01-10
Remove game from most method params in LogHelper.

kberg:


kberg:
One of the things to notice in this PR is how this impacts two things, all because `game` is no longer needed just for logging:
1. a few of the card instances are starting to remove a need for the second game parameter in a few of its methods. A view of things to come.
2. Some tests no longer need to hold on to game instances. Which may seem a little weird because now tests must call Game.newInstance to initialize a game, but no longer need it. This is one thing I'm slightly unsure of, not in terms of whether things still work, but whether this code smell reveals one of the unpleasant side-effects of this whole idea.

bafolts:
I don't think it is code smell. Prior to this change we could test for scenarios which weren't even feasible. We could have passed a `Player` from one `Game` and then a completely different `Game` as another argument. This led to more flexible testing but doesn't really represent how `Player` and `Game` are used in practice. We really should have been verifying before this change that every `player` argument was even a player on `Game`. We were always assuming the two were associated.

bafolts:
This should lead to a smaller `main.js` once webpack is done building everything.

bafolts (approved):

PR #2334, opened 2021-01-10
buttons to hide green and orange cards

### LINKED ISSUE #2309: Feature request: a button to collapse non-blue cards (outside preferences)
Currently the button to hide non-blue cards is in the preference. While this is good, users may want to hide and unhide the non-blue cards many times through out the games. Going back and forth to the preference settings requires too many clicks.

**Proposal**

One of the three methods to collapse the stack.
1) Click on any green / red cards to collapse either card stack.
2) A button near 'played cards' to collapse both stack.
3) A button each at the bottom of the green / red stacks to collapse either card stack.

What should be show once the stacks are collapsed
1) Total number of cards.
2) Total number of cards costing more than 20 mc. (Optional, we might not need it.)
3) Total VP from that stack.
4) Variable  VP cards: Immigration Shuttles, Space Port Colony, Jovian Multipliers. (Optional, we might not need it.)

Benefits
The long stack of green/red cards are what is taking a lot of vertical space. We will be able to save a lot of vertical space by being able to collapse and expand these stacks at will.

bitmap

Tagging @chosta for help.

vsrisuknimit:
This PR can close #2309 .

By default
image

After hiding green cards is clicked
image

After hiding orange cards is clicked
image

Things that I would like to do. In order of importance:
1) Force refresh. Right now you have to refresh the page for this to take effect.
~~2) Both buttons to be on the same row but above the cards. I tried to nested div but compiler screamed at me for having more than one nested template or something.~~
3) Fix css. I always mess up css selector. There must be a cleaner way to do this.
~~4) Show cards number of those particular types on the buttons. ~~


vsrisuknimit:
The buttons are now on the same row.

image

vsrisuknimit:
I see ForceRender() function is TopBars.ts but I don't under how it works. Why does it force render?

https://github.com/bafolts/terraforming-mars/blob/5f75aec88078be7aa019fe8fbcf729611daac846/src/components/TopBar.ts#L20

bafolts:
Force rendering should be avoided. Vue will monitor when `data` changes which should require updating the component. Here whenever a card is hidden or shown if you update a property on `data` then vue will naturally update the component.

src/components/PlayerHome.ts
```diff
@@ -90,6 +91,30 @@ export const PlayerHome = Vue.component('player-home', {
 
       return 'generation ' + this.player.generation;
     },
+    toggleAutomatedCardsHiding() {
+      const newVal = this.isAutomatedCardShown() ? '1': '';
+      PreferencesManager.saveValue('hide_automated_cards', newVal);
+      PreferencesManager.preferencesValues.set('hide_automated_cards', this.isAutomatedCardShown());
+    },
+    isAutomatedCardShown(): boolean {
+      const val = PreferencesManager.loadValue('hide_automated_cards');
```
bafolts, line 100: Storing this in `PreferencesManager` is fine but we need to leverage the tools that vue provides for this. I would make `hide_automated_cards` a property on `data`. Then in a `watch` when `hide_automated_cards` changes I would write that change to `PreferencesManager`. What is here is working against vue instead of with it and leading to needing the force update.

```
data: {
   return {
      hide_automated_cards: PreferencesManager.loadValue('hide_automated_cards')
   }
},
watch: {
   hide_automated_cards: function () {
     PreferencesManager.preferencesValues.set('hide_automated_cards', this.hide_automated_cards);
   }
}
```

With this in place as you update `this.hide_automated_cards` the `PreferencesManager` will naturally be kept up to date. We really should work to remove any place we are using force update and refactor it to work with vue.

vsrisuknimit: @bafolts I added data and watch but it still requires a refresh to update the state of the hide/show. I don't understand vue so I am probably doing something wrong here.

vsrisuknimit:
image

Example with transparency.

It still requires manual refresh to take effect after clicking the button. I need help here @bafolts , @chosta .

I also would like to show the number for each type. I have not figured out how yet.

vsrisuknimit:
Perhaps the transparency should be switched. Normally it is somewhat transparency. Then once you hide something, that button becomes more bright, showing you can click to un-hide.

vsrisuknimit:
Now I have the card type number.

All type shown (buttons are transparent)
image

Automated and event card are hidden (buttons are light up)
image

vsrisuknimit:
Switch transparency back according to some feedbacks. Lit when showing card but not too bright. Dim when hiding cards but not too dim.

image

image

vsrisuknimit:
@bafolts now that I merged your PR, do I need to do anything else? Sorry I am still very confused by Git.

src/components/PlayerHome.ts
```diff
@@ -90,6 +115,46 @@ export const PlayerHome = Vue.component('player-home', {
...
+        return this.isAutomatedCardShown() ? 'Hide automated cards' : 'Show automated cards';
+      } else if (hideType === 'EVENT') {
+        return this.isEventCardShown() ? 'Hide event cards' : 'Show event cards';
+      } else {
+        return '';
+      }
+    },
+    getHideButtonClass: function(hideType: string): string {
+      if (hideType === 'ACTIVE') {
+        return this.isActiveCardShown() ? 'hiding-card-button hiding-card-button--active' : 'hiding-card-button hiding-card-button--active-transparent';
```
bafolts, line 149: There is a lot of duplication here with `hiding-card-button hiding-card-button--`. I would put this in a `const prefix` to store it once. This takes up space in the generated build and it is also code that has to be updated if this prefix were to change. Then this can become.

```
if (hideType === 'ACTIVE') {
    return this.isActiveCardshown() ? prefix + "active" : prefix + "active-transparent";
}
```

src/components/PlayerHome.ts
```diff
@@ -90,6 +115,46 @@ export const PlayerHome = Vue.component('player-home', {
...
+    },
+    isAutomatedCardShown(): boolean {
+      return this.hide_automated_cards !== '1';
+    },
+    isEventCardShown(): boolean {
+      return this.hide_event_cards !== '1';
+    },
+    getToggleLabel: function(hideType: string): string {
+      if (hideType === 'ACTIVE') {
+        return this.isActiveCardShown() ? 'Hide active cards' : 'Show active cards';
```
bafolts, line 138: Simplify these to reduce duplication to:

`return (this.isActiveCardShown() ? 'Hide' : 'Show') + ' active cards';`

bafolts (approved):
The usage of vue looks good now. I am glad we have this in place because we should copy this pattern every place we are updating local storage. Minor string updates I would make before merging that aren't critical.

src/components/PlayerHome.ts
vsrisuknimit: Simplified this and the css.

src/components/PlayerHome.ts
vsrisuknimit: Simplified this and the css.

src/components/PlayerHome.ts
```diff
@@ -135,22 +135,23 @@ export const PlayerHome = Vue.component('player-home', {
     },
     getToggleLabel: function(hideType: string): string {
```
alrusdi, line 136: @vsrisuknimit can you please add translation support for these strings?

alrusdi (changes requested):

src/components/PlayerHome.ts
vsrisuknimit: added translation

PR #2336, opened 2021-01-11
Moon #3: introducing the card manifest but very little else.

kberg:


bafolts (approved):
> There are great ideas undiscovered, breakthroughs available to those who can remove one of truth's protective layers. There are places to go beyond belief." -Neil Armstrong

kberg:
> > There are great ideas undiscovered, breakthroughs available to those who can remove one of truth's protective layers. There are places to go beyond belief." -Neil Armstrong

:rocket:

PR #2337, opened 2021-01-11
Move some Board methods just used for randomly selecting a forest space to GameSetup.

kberg:
Move some Board methods just used for randomly selecting a forest space to GameSetup.
Also, push down rng from the newInstance, to start sharing it (so shared seeds yield the same results.)

I simplified the randomizer, although it doesn't use the card draw mechanism from the rules, yet.

Any thoughts on a good way to test this deterministically?


bafolts (approved):
As long as `Random` is providing random numbers then your function should also be random. I say that not being familiar at all with random number generators. For this solution either way as it removes another shuffle implementation.

PR #2338, opened 2021-01-11
Fix SRR (de)serializing

### LINKED ISSUE #2278: Card vanished from Self-Replicating Robots
Gorg on discord:
had a bug today, had phobos on self-replicating robots. It just dissappeared from play, inexplicably. Normally it would be listed at the very bottom. Simply removed from my hand
apologies if known bug

me on discord:
Okay I've triple checked this.  A Capital card vanished off of my SRR. (I scrolled up a bit in the log to show that it had been there)

I think it happened when I undid placing a greenery tile. I always reload right after undo and then I noticed the thing was gone.
I went ahead and finished the generation in case something kinda reset it, but no. I guess I'll leave it in this state now. https://terraforming-mars.herokuapp.com/player?id=4dcdfac088e0

Lynesth:
Fixes #2278

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1960,7 +1960,12 @@ export class Player implements ISerializable<SerializedPlayer> {
         resourceCount: c.resourceCount,
       };
       if (c instanceof SelfReplicatingRobots) {
-        result.targetCards = c.targetCards;
+        result.targetCards = c.targetCards.map((t) => {
+          return {
+            card: {name: t.card.name},
+            resourceCount: t.resourceCount,
+          };
+        });
       }
       return result;
     });


kberg (commented):
How do you know this works?

src/Player.ts
```diff
@@ -1960,7 +1960,12 @@ export class Player implements ISerializable<SerializedPlayer> {
         resourceCount: c.resourceCount,
       };
       if (c instanceof SelfReplicatingRobots) {
-        result.targetCards = c.targetCards;
+        result.targetCards = c.targetCards.map((t) => {
```
kberg, line 1963: Something to consider later on: So this whole thing becomes

this.playedCards.map((c) => c.serialize());

Card gains a serialize method:

```
Card.serialize(): SerializedCard {
  return { name: c.name, resourceCount: c.resourceCount };
}
```

and SRR subclasses it:

```
SRR.serialize(): SerializedCard {
  const sc = super.serialize();
  sc.targetCards = this.targetCards.map((c) => c.serialize());
}
```

kberg: That removes the duplicate need for the two lines below, and puts serialization where it belongs.

kberg (approved):

src/Player.ts
Lynesth: Those are different though: ```SerializedCard``` !== ```SerializedRobotCard```.

kberg:
Ah, I see. Still, I'd like to see serialization / deserialization move to the cards, it looks like a good time to do that now.

Lynesth:
> How do you know this works?

Woops, didn't see that question.
Well, deserializing the card didn't work before:
```js
[2021-01-11 03:40:36 GMT+0000] card not found undefined
[2021-01-11 03:40:36 GMT+0000] did not find card for SelfReplicatingRobots { card:
   { properties:
      { cardType: 'green',
        name: 'GHG Factories',
        tags: [Array],
        cost: 11,
        hasRequirements: false,
        metadata: [Object] } },
  resourceCount: 2 }
```
I don't get the error now.

Also I tested the following scenario which is able to reproduce the error:
- Play a game with undo enabled
- Play SRR
- Add a card to SRR
- Do another action (whatever)
- Undo <- This removes the card from SRR

bafolts:
The issue here was introduced by the static memory change. Prior to the change if you were to call `JSON.stringify` on an implementation of `ICard` you would end up with `name` as a property. After the change the `name` property is not enumerable so doesn't show up on `JSON.stringify`. This fix is the right thing to do as we should be manually building these `SerializedRobotCard`. Typescript isn't wrong since what we were providing does have a `name` property but typescript doesn't know that we needed the property to be enumerable.

PR #2341, opened 2021-01-11
Add regression test for StripMine with Reds ruling

nwai90:
Adds a regression test related to https://discord.com/channels/737945098695999559/742826825922904225/798210229597634631.

The issue is related to an incorrect check which assumes that the user can pay for Reds tax using steel. It seems that canPlay requires 2 checks, namely (1) that the user has enough MC to pay Reds tax, and (2) the user has enough MC plus steel and/or titanium where applicable to pay the card's cost plus Reds taxes.

Pre-Lynesth's Reds tax PR we will likely need to update this check for all other building and space cards that can raise TR as an interim measure.


kberg (approved):
Very good.

nwai90:
@Lynesth has pointed out that this does not handle all situations (e.g. player with 25 MC and 3 steel can still choose to spend all 25 MC to pay for the card) but it's better than nothing for now. We had a similar issue with Dirigibles and Stratospheric Birds before where it was necessary to check that the player did not spend all floaters to pay for playing the card.

PR #2344, opened 2021-01-12
keyboard shortcut

### LINKED ISSUE #1879: Keyboard Shortcuts
I think it would be helpful to have keyboard shortcuts for moving up and down the page.

Q: Game board
W: Turmoil board
E: Milestones and Awards
R: Tag Overview
T: Game log

A: Action
S: Cards in hand
D: Played cards
F: Colonies

vsrisuknimit:
Here is a draft of the keyboard shortcuts.

It does not listen yet or not redirect. I cannot tell which is which. I look into event listener on the browser but I don't see anything when a press a key. 

Tag @bafolts for help.

This can close #1879 .


src/components/PlayerHome.ts
```diff
@@ -49,6 +50,16 @@ export const PlayerHome = Vue.component('player-home', {
   },
   mixins: [PlayerMixin],
   methods: {
+    navigatePage: function(event: any) {
+      console.log('key pressed');
+      if (event.keyCode === KeyboardNavigation.COLONIES) {
+        console.log('f pressed');
```
kberg: remove.

src/KeyboardNavigation.ts
```diff
@@ -0,0 +1,12 @@
+
+export enum KeyboardNavigation {
```
kberg, line 2: What a great selection of shortcuts.

kberg (commented):
**UM.** How did I manage to push directly to your repo?

Anyway, I managed to get keyboard shortcuts sort of workingg, and somehow they're now in this PR because of it.

vsrisuknimit:
Thank @kberg and @bafolts  . I changed the getElementByClass to getElementById instead. That is more specific and redirect us to where we want. Changed the behavior of the scrollIntoView to put the element in the middle of the page instead of the very far top which is obscured by the top bar.

Do we need both mounted and v-on? I thought one of the them should do. I tried to remove v-on but it seems we need it.

One major problem is that now there is this little blue border going over the whole element with v-on. Not sure how to remove it.

image

vsrisuknimit:
I was able to remove the blue outline but we hate in-line style so that perhaps should be somewhere else. I am not sure where since :class for that element is doing something weird.

chosta:
@vsrisuknimit you can use a function for the classes based on conditional logic.
Do a global search for `getClasses` to see how this could be done.

vsrisuknimit:
@Lynesth raised the issue of non-qwerty keyboards. This mapping does not make a lot of sense there.

I think the best path forward is to have mappable keys. I will need help with that.

Otherwise we can do shift+1, shift+2, but one key is much better.

kberg:
> @Lynesth raised the issue of non-qwerty keyboards. This mapping does not make a lot of sense there.
> 
> I think the best path forward is to have mappable keys. I will need help with that.
> 
> Otherwise we can do shift+1, shift+2, but one key is much better.

No, but 1,2,3,4,5 are excellent

vsrisuknimit:
@kberg just plain number will not work well. Some users may type in a number when paying for a cost.

image

vsrisuknimit:
Tested. Typing in the input field will trigger the keyboard shortcut. So we should avoid plain number.

bafolts (changes requested):
This looks right. I would use `KeyboardEvent` instead of `any`.

src/components/PlayerHome.ts
```diff
@@ -74,6 +75,45 @@ export const PlayerHome = Vue.component('player-home', {
   },
   mixins: [PlayerMixin],
   methods: {
+    navigatePage: function(event: any) {
```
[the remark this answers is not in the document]
vsrisuknimit: @bafolts if I use KeyboardEvent, event.target.nodeName does not work. "Property 'nodeName' does not exist on type 'EventTarget'."

vsrisuknimit:
Added key mapping graphics in the help page.

image

vsrisuknimit:
These are the poll results from Discord.

image

image

I also added the ability to change the scrolling speed in preferences setting.

src/components/PlayerHome.ts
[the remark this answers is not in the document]
vsrisuknimit: Fixed now. Cast event.target as Element.

src/KeyboardNavigation.ts
```diff
@@ -0,0 +1,12 @@
...
+  GAMEBOARD = 'Digit1',
+  TURMOILBOARD = '',
```
bafolts, line 4: Many of these are missing. I can't imagine this works with these missing.

src/components/PlayerHome.ts
```diff
@@ -163,7 +208,7 @@ export const PlayerHome = Vue.component('player-home', {
     );
   },
   template: `
-        <div id="player-home" :class="player.turmoil ? 'with-turmoil': ''">
+        <div v-on:keydown="navigatePage" tabindex="0" id="player-home" :class="'shortkey-no-outline'+(player.turmoil ? ' with-turmoil': '')">
```
bafolts, line 211: I think you can use `v-on:keydown.prevent` instead of manually calling to prevent default ourselves.

src/KeyboardNavigation.ts
vsrisuknimit: 4 are what I decided would be useful in the end. I left the rest there as a place holder in the future if we want remappable keys. Should I remove them all?

src/components/PlayerHome.ts
vsrisuknimit: keydown.prevent causes an error.

image

src/KeyboardNavigation.ts
bafolts: Yes. If we decide to add them later we can add them then. I would also remove their corresponding ids from the template.

vsrisuknimit: Removed the unused keys and ids.

bafolts (approved):

PR #2345, opened 2021-01-12
Add first Moon card, Mare Nectaris Mine.

kberg:
This adds A LOT of dressing, but it's all tested. The one thing to note is that MOON_CARD_MANIFEST is now part of ALL_MANIFESTS, which does run the risk of introducing the cards into games, except the Dealer isn't going to touch it since it's not registered within the dealer.

But that's what I needed to do to get the card rendered in the UI.

image

On the positive side, many, may cards are now going to be much simpler to write because this work is done.


src/cards/moon/MareNectarisMine.ts
```diff
@@ -0,0 +1,42 @@
...
+          b.moonMine().asterix();
+        }),
+      },
+    });
+  }
+
+  public reserveUnits = Units.of({titanium: 1});
+
+  public play(player: Player) {
+    Units.deductUnits(this.reserveUnits, player);
```
bafolts, line 36: Something beautiful about how close this matches to what is written on the card on these 4 lines.

bafolts (approved):

PR #2346, opened 2021-01-12
Fix: Virus should enable playing CrashSiteCleanup in solo

nwai90:
Closes https://github.com/bafolts/terraforming-mars/issues

### DIFF
--- a/src/cards/base/Virus.ts
+++ b/src/cards/base/Virus.ts
@@ -37,7 +37,7 @@ export class Virus extends Card implements IProjectCard {
   }
   public play(player: Player, game: Game): PlayerInput | undefined {
     if (game.getPlayers().length === 1) {
-      new RemoveAnyPlants(player, game, 5).execute();
+      game.someoneHasRemovedOtherPlayersPlants = true;
       return undefined;
     }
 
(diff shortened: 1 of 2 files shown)


src/cards/base/Virus.ts
```diff
@@ -37,6 +37,7 @@ export class Virus extends Card implements IProjectCard {
   }
   public play(player: Player, game: Game): PlayerInput | undefined {
     if (game.getPlayers().length === 1) {
+      new RemoveAnyPlants(player, game, 5).execute();
```
Lynesth, line 40: One could argue that you can simply set `this.game.someoneHasRemovedOtherPlayersPlants = true;` directly instead.

Lynesth (approved):

src/cards/base/Virus.ts
nwai90: Very true :D Should I?

bafolts: Yes you should. It is far simpler to set `game.someoneHasremovedOtherPlayersPlants`.

nwai90: Done!

PR #2349, opened 2021-01-12
Fix: Flooding shouldn't suggest to remove MC if oceans are maxed

nwai90:
Bug report from Discord: https://discord.com/channels/737945098695999559/742721510376210583/798565553681596436

### DIFF
--- a/src/cards/base/Flooding.ts
+++ b/src/cards/base/Flooding.ts
@@ -50,17 +50,23 @@ export class Flooding extends Card implements IProjectCard {
       game.defer(new PlaceOceanTile(player, game));
       return undefined;
     }
+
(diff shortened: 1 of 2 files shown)


src/cards/base/Flooding.ts
```diff
@@ -50,17 +50,23 @@ export class Flooding extends Card implements IProjectCard {
       game.defer(new PlaceOceanTile(player, game));
       return undefined;
     }
+
+    const oceansMaxedBeforePlacement = game.board.getOceansOnBoard() === MAX_OCEAN_TILES;
```
kberg, line 54: The extra local variable doesn't add too much for readability, but this is a minor comment. Thanks for this fix!

kberg (approved):

PR #2353, opened 2021-01-12
Fixed helion card buying bug

sienmich:
Bug from discord where helion could pay 0 mc and 0 heat and buy some cards.

Very ugly one.

It was caused by that line of code:
```ts
      app.$data.isResearchPhase = app.playerinput.title === 'Select how to pay for cards';
```
With my card rework I changed the card selection msg to exactly this one, while on main server it was earlier like this: `Select how to pay 6 for purchasing 2 card(s)`.

I removed the unnecessary `isResearchPhase` flag. It was added earlier to allow Helion pay with heat in Research Phase, now it works without it.

### DIFF
--- a/src/deferredActions/SelectHowToPayDeferred.ts
+++ b/src/deferredActions/SelectHowToPayDeferred.ts
@@ -21,6 +21,8 @@ export class SelectHowToPayDeferred implements DeferredAction {
       return undefined;
     }
 
+    console.log('need to pay:', this.amount);
+
     return new SelectHowToPay(
       this.options.title || 'Select how to pay for ' + this.amount + ' MCs',
       this.options.canUseSteel || false,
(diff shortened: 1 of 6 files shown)


src/deferredActions/SelectHowToPayDeferred.ts
```diff
@@ -21,6 +21,8 @@ export class SelectHowToPayDeferred implements DeferredAction {
       return undefined;
     }
 
+    console.log('need to pay:', this.amount);
```
kberg, line 24: Remove before flight.

kberg (approved):
Thanks for the explanation. Yuch.

src/deferredActions/SelectHowToPayDeferred.ts
kberg: Well, remove after flight, please. :D

PR #2360, opened 2021-01-13
Moon 7: Fix Board.getAdjacentSpaces to work properly with the Moon.

kberg:
Added pretty complete unit tests to prevent someone else from thinking this was quite so trivial.


src/boards/Board.ts
```diff
@@ -5,8 +5,20 @@ import {TileType} from '../TileType';
...
+/**
+ * A representation of any hex board. This is normally Mars (Tharsis, Hellas, Elysium) but can also be The Moon.
+ *
+ * It also includes additional spaces, known as Colonies, that are not adjacent to other spaces.
+ */
 export abstract class Board {
-  public abstract spaces: Array<ISpace>;
+  private maxX: number = 0;
+  private maxY: number = 0;
+  constructor(public spaces: Array<ISpace>) {
```
bafolts, line 16: I _think_ (I tried it locally really quick as well) you can use `protected constructor` here. This will avoid needing to have `private constructor` in every class that extends `Board`.

bafolts (approved):
If the `protected constructor` works from `src/boards/Board` I would use that and remove each separate `private constructor` in the classes extending `Board` as that will remove some code. Not a big deal though as that is an optimization and semantics.

src/boards/Board.ts
kberg: Nice tip. Will do.

PR #2361, opened 2021-01-13
Update RoboticWorkforce with a Units data structure. Also,...

kberg:
Also, move some of the units to cards that have static properties. Tests still pass, is that good enough?

This will vastly simplify the work involved in enabling cards for Robotic Workforce by just annotating the cards themselves. There are still outliers, like Solar Farm, et cetera, but this will vastly reduce the amount of code in Robotic Workforce.

(Plus the fields, like those in AI Central) will reduce the logic in getAvailableCards, and can also be used in card.canPlay() and card.play().


src/cards/base/RoboticWorkforce.ts
```diff
@@ -258,129 +259,127 @@ export class RoboticWorkforce extends Card implements IProjectCard {
...
+        new Updater(CardName.TECTONIC_STRESS_POWER, {energy: 3}),
+        new Updater(CardName.TITANIUM_MINE, {titanium: 1}),
+        new Updater(CardName.TROPICAL_RESORT, {megacredits: 3, heat: -2}),
+        new Updater(CardName.UNDERGROUND_CITY, {energy: -2, steel: 2}),
+        new Updater(CardName.URBANIZED_AREA, {energy: -1, megacredits: 2}),
+        new Updater(CardName.UTOPIA_INVEST, {steel: 1, titanium: 1}),
+        new Updater(CardName.WINDMILLS, {energy: 1}),
       ];
 
       const result:Updater = updaters.filter((u) => u.name === foundCard.name)[0];
```
Lynesth, line 353: Just use `find` here.

src/cards/base/RoboticWorkforce.ts
```diff
@@ -258,129 +259,127 @@ export class RoboticWorkforce extends Card implements IProjectCard {
...
       }
-      if (player.getProduction(Resources.PLANTS) + result.plantProduction < 0) {
-        throw 'not enough plant production';
+      if (player.getProduction(Resources.PLANTS) + units.plants < 0) {
+        throw new Error('not enough plant production');
       }
-      if (player.getProduction(Resources.HEAT) + result.heatProduction < 0) {
-        throw 'not enough heat production';
+      if (player.getProduction(Resources.HEAT) + units.heat < 0) {
+        throw new Error('not enough heat production');
```
Lynesth, line 374: This is missing the check for megaCredits as well.
But I noticed you have a `hasProduction` method in Units, why not use that instead ?

Lynesth (commented):
Haven't checked that you didn't mess up the conversion from arguments to units in the `updaters` array.

src/cards/base/RoboticWorkforce.ts
```diff
@@ -258,129 +259,127 @@ export class RoboticWorkforce extends Card implements IProjectCard {
...
+        new Updater(CardName.MINING_OPERATIONS, {steel: 2}),
+        new Updater(CardName.MINING_QUOTA, {steel: 2}),
+        new Updater(CardName.MINING_RIGHTS, {steel: this.miningSteelProduction, titanium: this.miningTitaniumProduction}),
+        new Updater(CardName.MINING_RIGHTS_ARES, {steel: this.miningSteelProduction, titanium: this.miningTitaniumProduction}),
+        new Updater(CardName.MOHOLE, {heat: 3}),
+        new Updater(CardName.MOHOLE_AREA, {heat: 4}),
+        new Updater(CardName.MOHOLE_AREA_ARES, {heat: 4}),
+        new Updater(CardName.MOHOLE_EXCAVATION, {steel: 1, heat: 2}),
+        new Updater(CardName.NATURAL_PRESERVE, {energy: 1}),
+        new Updater(CardName.NATURAL_PRESERVE_ARES, {energy: 1}),
```
sienmich, line 321: MC, not energy

sienmich (changes requested):

src/cards/base/RoboticWorkforce.ts
kberg: Thanks.

src/cards/base/RoboticWorkforce.ts
kberg: I didn't know whether the error codes were meaningful.

kberg:
> Haven't checked that you didn't mess up the conversion from arguments to units in the `updaters` array.

Really, good point. I'm going to verify correctness using a spreadsheet today. You two have already found two errors, and I should have bothered to do this work first. Mea culpa. Sorry about that.

src/cards/base/RoboticWorkforce.ts
kberg: So you know, the linne to the right does list it under energy.

kberg (commented):
> Haven't checked that you didn't mess up the conversion from arguments to units in the `updaters` array.

Here is my verification:

https://docs.google.com/spreadsheets/d/1u1K3tZIdnjJYs97yaSk7JffvDAIjqxANWC06yWpAWOk/edit#gid=0

I eyeballed everything and it's correct. That @sienmich found an error is because the base data was in error. Moving this stuff over to the card itself will eliminate any future risk of error.

src/cards/base/RoboticWorkforce.ts
kberg: Fixed.

src/cards/Card.ts
```diff
@@ -67,4 +69,7 @@ export abstract class Card {
   public get tags() {
     return this.properties.tags === undefined ? [] : this.properties.tags;
   }
+  public get productionDelta() {
+    return this.properties.productionDelta;
```
bafolts, line 73: Tests still passing is good enough for me. A test written for this change would mostly be verifying that javascript sets a property which seems like overkill. Assuming that `RoboticWorkforce` has a unit test that relies on this value being set that is verifying this change works.

bafolts (approved):

src/cards/base/RoboticWorkforce.ts
kberg: @sienmich I am blocked on submitting this until you release the changes request.

sienmich (approved):

PR #2376, opened 2021-01-14
Vp text consistency

### LINKED ISSUE #2325: consistency of VP text on cards
I think the separation of vp-text from immediate-instruction text on the cards is a good move. I have some questions and review comments:

Main thing: is it possible to move the VP text to the bottom of the card to the left of the VP symbol? I feel like to new players the cards are a confusing jumble of different kinds of information and any kind of structuring like this would (I think) help.

To conserve space, perhaps the wording could be standardized on the compact form, using Ants as an example: "1 vp per 2 microbes here." When the resource is described in two words like "preservation resource" then it can just say "resource". (In general I think we could always use "here" in place of "to/from this card".)

Jovian Lanterns is missing that "here" part.

Commercial District: "place this tile" doesn't belong in the vp text?

These four cards aren't yet separating the vp text from the immediate text. If any are intentional for space reasons I am wondering if we have other solutions that are possible, because I think the separation seems good for accessibility.

Asteroid Deflection System
Asteroid Hollowing
Birds
Decomposers

Lastly Self-Rep Robots seems to be using vp-text style for its effect. Is that, again, just because it was the only way to make it fit?

@chosta

chosta:
This closes #2325

Managed to resolve all but two of the cards mentioned. `Birds` and `Asteroid deflection system` don't have enough space to work with. I'll leave them for a future iteration.

### DIFF
--- a/src/cards/colonies/JovianLanterns.ts
+++ b/src/cards/colonies/JovianLanterns.ts
@@ -62,7 +62,7 @@ export class JovianLanterns implements IProjectCard, IResourceCard {
           eb.titanium(1).startAction.floaters(2);
         }).br;
         b.tr(1).floaters(2).asterix().br;
-        b.vpText('1VP per 2 floaters.');
+        b.vpText('1 VP per 2 floaters here.');
       }),
       description: {
         text: 'Requires 1 Jovian tag. Increase your TR 1 step. Add 2 floaters to ANY card.',
--- a/src/cards/ares/CapitalAres.ts
+++ b/src/cards/ares/CapitalAres.ts
@@ -23,7 +23,7 @@ export class CapitalAres extends Capital {
             pb.minus().energy(2).br;
             pb.plus().megacredits(5);
           }).nbsp.tile(TileType.CAPITAL, false, true).br;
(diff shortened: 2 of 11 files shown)


src/cards/colonies/JovianLanterns.ts
```diff
@@ -62,7 +62,7 @@ export class JovianLanterns implements IProjectCard, IResourceCard {
           eb.titanium(1).startAction.floaters(2);
         }).br;
         b.tr(1).floaters(2).asterix().br;
-        b.vpText('1VP per 2 floaters.');
+        b.vpText('1 VP per 2 floaters here.');
```
sienmich, line 65: All other cards in this PR have `on this card` instead of `here`, maybe this also should?

kberg (approved):
A thankless task.

src/cards/colonies/JovianLanterns.ts
kevinb9n: (I was recommending going with "here" instead of "on/to/from this card" across the board as a general space-saving measure.)

sienmich: Sound good for me, we can change other cards to `here`, I just like consistency

PR #2378, opened 2021-01-14
Refactored cards requirements.

sienmich:
Right now every card has it's requirements in 3 places:

1. `metadata.requirements`
2. `metadata.description`
3. canPlay() method

This PR makes 3. (`canPlay()`) automatically from the 1. (`metadata.requirements`).

More specifically:
In the `abstract class Card` I added this method:
```ts
  public canPlay(player: Player, _game?: Game) {
    if (this.properties.metadata.requirements === undefined) {
      return true;
    }
    return this.properties.metadata.requirements.satisfies(player);
  }
```
Now every card with requirements either does not implement it's own `canPlay()` or uses `super()` inside of it.
Checking the requirements is inside the `CardRequirement` class.


sienmich:
Additionally I'd like soon to generate requirements in `metadata.description` automatically. This could help removing inconsistencies like `Requires +2 C` vs `Requires +2 C°`. It's harder because of translations. Please let me know what u think about it.

src/cards/CardRequirement.ts
```diff
@@ -75,16 +79,84 @@ export class CardRequirement {
...
+      return player.game.checkRequirements(player, GlobalParameter.TEMPERATURE, this.amount, this.isMax);
+    case RequirementType.VENUS:
+      return player.game.checkRequirements(player, GlobalParameter.VENUS, this.amount, this.isMax);
+    case RequirementType.TR:
+      return this.satisfiesInequality(player.getTerraformRating());
+    case RequirementType.REMOVED_PLANTS:
+      return player.game.someoneHasRemovedOtherPlayersPlants;
+    case RequirementType.RESOURCE_TYPES:
+      const standardResources = [Resources.MEGACREDITS, Resources.STEEL, Resources.TITANIUM, Resources.PLANTS, Resources.ENERGY, Resources.HEAT]
+        .map((res) => player.getResource(res)).filter((count) => count > 0).length;
```
Lynesth, line 138:
```ts
      const standardResources = [Resources.MEGACREDITS, Resources.STEEL, Resources.TITANIUM, Resources.PLANTS, Resources.ENERGY, Resources.HEAT]
        .filter((res) => player.getResource(res) > 0).length;
```

src/cards/CardRequirement.ts
```diff
@@ -75,16 +79,84 @@ export class CardRequirement {
...
+      const nonStandardResources = player.getCardsWithResources().map((card) => card.resourceType).filter((v, i, a) => a.indexOf(v) === i).length;
```
Lynesth, line 139: Just a thought:
```ts
      const nonStandardResources = new Set(player.getCardsWithResources().map((card) => card.resourceType)).size;
```

src/cards/CardRequirement.ts
```diff
@@ -75,16 +79,84 @@ export class CardRequirement {
...
+      return this.satisfiesInequality(coloniesCount);
+    case RequirementType.FLOATERS:
+      return this.satisfiesInequality(player.getResourceCount(ResourceType.FLOATER));
+    case RequirementType.GREENERIES:
+      const greeneries = player.game.board.spaces.filter(
+        (space) => space.tile !== undefined &&
+            space.tile.tileType === TileType.GREENERY &&
+            (space.player === player || this._isAny),
+      ).length;
+      return this.satisfiesInequality(greeneries);
```
Lynesth, line 117: Just an idea to prevent looping the whole board each time:
```ts
      let greeneries = 0;
      return player.game.board.spaces.some((space) => {
        if (space.tile?.tileType === TileType.GREENERY && (space.player.id === player.id || this._isAny)) {
          greeneries++;
          return this.satisfiesInequality(greeneries);
        } else {
          return false;
        }
      });
```

But maybe the number of cities/greeneries/oceans could already be saved on the board directly ?

src/cards/CardRequirement.ts
```diff
@@ -75,16 +79,84 @@ export class CardRequirement {
...
+    if (this.isMax) {
+      return calculated <= this.amount;
+    }
+    return calculated >= this.amount;
+  }
+
+  public satisfies(player: Player): boolean {
+    switch (this.type) {
+    case RequirementType.CHAIRMAN:
+      if (player.game.turmoil !== undefined) {
```
kberg, line 93: return player.game.turmoil?.chairman === player.id;

kberg (commented):
What concerns me about this PR is that it starts to conjoin behavior and visualization a little closer than I'm clear, yet, is a good idea. Once we go down this path, it's hard to come back.  But look -- I'm happy to hear other peoples' opinions. I won't be reviewing this tonight, though.

src/cards/CardRequirement.ts
sienmich: It's not working, because `satisfiesInequality` sometimes checks for <=, sometimes for >=.

sienmich:
> What concerns me about this PR is that it starts to conjoin behavior and visualization a little closer than I'm clear, yet, is a good idea. Once we go down this path, it's hard to come back. But look -- I'm happy to hear other peoples' opinions. I won't be reviewing this tonight, though.

Well, all I can say is that it is a good idea in my opinion. Card visualization is a representation of its behavior. Having less redundancy between them should lead to less bugs, less inconsistency, more readable code. And much nicer and faster new cards implementation.

chosta:
I like where this is going. I have one concern for the automatic description though. Printing the description for cards in an automated manner would work well for a card that has nothing else to add as a description but not so well for cards with bigger descriptions. Example:
`Advanced Ecosystems` -> `(Requires a Plant tag, a Microbe tag, and an Animal tag.)`
but
`AI Central` -> `(Requires 3 Science tags to play. Decrease your Energy production 1 step.)`

You probably intend to prepend the automated description but this mean the rest of it should be written as "Decrease your Energy Production 1 step" which would look incomplete when added to the metadata. You would then probably want to automate the second part of the description too, which could be easy for most of the cards but there are some exceptions with more customized descriptions like `Desperate Measures` where you'd need custom text describing the card. I'm just raising awareness of the 80/20 problem that we might have here and whether we should take the path to complete description automation (which I support).

src/cards/CardRequirement.ts
Lynesth: Ah yeah, we'd either have to make 2 different loops depending on isMax:
```ts
      const countGreenery = (space: ISpace) => space.tile?.tileType === TileType.GREENERY && (space.player.id === player.id || this._isAny);
      let greeneries = 0;
      if (isMax) {
        return player.game.board.spaces.every((space) => {
          if (countGreenery(space)) {
            greeneries++;
            return this.satisfiesInequality(greeneries);
          } else {
            return true;
          }
        });
      } else {
        return player.game.board.spaces.some((space) => {
          if (countGreenery(space)) {
            greeneries++;
            return this.satisfiesInequality(greeneries);
          } else {
            return false;
          }
        });
      }
```

Or you could do it like this I guess:
```ts
      let greeneries = this.isMax ? this.amount * 2 : 0;
      return player.game.board.spaces.some((space) => {
        if (space.tile?.tileType === TileType.GREENERY && (space.player.id === player.id || this._isAny)) {
          this.isMax ? greeneries-- : greeneries++;
          return this.satisfiesInequality(greeneries);
        } else {
          return false;
        }
      });
```

src/cards/ares/EcologicalSurvey.ts
```diff
@@ -20,7 +19,7 @@ export class EcologicalSurvey extends Card implements IProjectCard {
       metadata: {
         description: 'Requires 3 greeneries on Mars.',
         cardNumber: 'A07',
-        requirements: CardRequirements.builder((b) => b.greeneries(3)),
+        requirements: CardRequirements.builder((b) => b.greeneries(3).any()),
```
chosta, line 22: Not important but this got me a bit confused as it probably belongs to a different PR or at least a separate commit.

src/cards/ares/GeologicalSurvey.ts
```diff
@@ -19,7 +18,7 @@ export class GeologicalSurvey extends Card implements IProjectCard {
 
       metadata: {
         cardNumber: 'A09',
-        requirements: CardRequirements.builder((b) => b.greeneries(5).max()),
+        requirements: CardRequirements.builder((b) => b.greeneries(5).any().max()),
```
chosta, line 21: Same here

src/cards/colonies/CoronaExtractor.ts
```diff
@@ -4,28 +4,29 @@ import {CardType} from '../CardType';
...
 import {CardRenderer} from '../render/CardRenderer';
+import {Card} from '../Card';
 
-export class CoronaExtractor implements IProjectCard {
-    public cost = 10;
-    public tags = [Tags.SPACE, Tags.ENERGY];
-    public name = CardName.CORONA_EXTRACTOR;
-    public cardType = CardType.AUTOMATED;
+export class CoronaExtractor extends Card implements IProjectCard {
+  constructor() {
```
chosta, line 12: So here you are just moving things in the constructor I guess. Should be a separate PR imo.

src/cards/CardRequirement.ts
sienmich: Nice, clever cheat :). I think the original looping through all spaces is not that computational consuming (half of the times we have to loop all board spaces anyway), and is simpler. So I'd leave it as it is?

src/cards/CardRequirement.ts
```diff
@@ -75,16 +79,92 @@ export class CardRequirement {
   public get amount(): number {
     return this._amount;
   }
+
+  protected satisfiesInequality(calculated: number) : boolean {
```
chosta, line 83: This needs unit tests :) The whole `satisfies` functionality I mean.

src/cards/CardRequirement.ts
```diff
@@ -75,16 +79,92 @@ export class CardRequirement {
...
+      return player.game.turmoil?.chairman === player.id;
+
+    case RequirementType.CITIES:
+      if (this._isAny) {
+        return this.satisfiesInequality(player.game.getCitiesInPlay());
+      } else {
+        return this.satisfiesInequality(player.getCitiesCount(player.game));
+      }
+
+    case RequirementType.COLONIES:
```
chosta, line 102: Will this work for `Pioneer Settlement` where the requirements has `max()` too. (Maybe it passes the tests, not sure I've covered it properly.

chosta (changes requested):
Apart from some organizational issues which should be just advice for future PRs, I'm mostly satisfied. What I'd like to see is some good unit test coverage on the `satisfies` functionality though.

chosta:
> What concerns me about this PR is that it starts to conjoin behavior and visualization a little closer than I'm clear, yet, is a good idea. Once we go down this path, it's hard to come back. But look -- I'm happy to hear other peoples' opinions. I won't be reviewing this tonight, though.

Opionion:
I'm with you on the `behaviour + viz` part but to be honest but these days I'm more inclined to sacrifice a bit of clean and lean for handy and automated.

src/cards/ares/EcologicalSurvey.ts
sienmich: I had to change this, so for EcologicalSurvey the satisfies() is working good. Ok, next time I can separate the commits better :)

src/cards/CardRequirement.ts
sienmich: It will, the `satisfiesInequality` checks for min / max

chosta (approved):
Please, check what I did for CardRenderer tests concerning global requirements here #2375

Maybe it's a good idea to make sure requirements are inside bounds too.

PR #2385, opened 2021-01-15
Add Moon standard projects and three of the five preludes.

kberg:


src/components/moon/MoonBoard.ts
```diff
@@ -0,0 +1,115 @@
...
+  components: {
+    'moon-space': MoonSpace,
+  },
+  data: function() {
+    return {
+    };
+  },
+  methods: {
+    getAllNonColonySpaces: function(): Array<SpaceModel> {
+      const boardSpaces: Array<SpaceModel> = this.model!.spaces;
```
bafolts, line 31: Using `this.model!` should be unnecessary! The `model` property does not indicate it can be `undefined`!

src/server/ServerModel.ts
```diff
@@ -360,6 +362,9 @@ function getCards(
   }));
 }
 
+// NOTE: This doesn't return a proper PlayerModel. It returns only a partial one, because
+// many of the field's values aren't set. That's why the code needs an "as PlayerModel" at
```
bafolts, line 366: I don't know if this is by design or by accident. Often times I would use `as PlayerModel` early on. I didn't realize that typescript treated these different until after I had used it a few places. This comment is helpful though to remind people of the situation.

bafolts: Really could consider avoiding using `as Type` for casting as it is almost always dangerous.

bafolts (approved):
This is excellent work!

I only spotted two places where `!` was being used unnecessarily. I would remove those as they aren't needed.

src/components/moon/MoonBoard.ts
```diff
@@ -0,0 +1,115 @@
...
+        (space1: SpaceModel, space2: SpaceModel) => {
+          return parseInt(space1.id) - parseInt(space2.id);
+        },
+      );
+      return boardSpaces.filter((s: SpaceModel) => {
+        return s.spaceType !== SpaceType.COLONY;
+      });
+    },
+    getSpaceById: function(spaceId: string) {
+      for (const space of this.model!.spaces) {
```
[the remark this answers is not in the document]
kberg: Done.

src/server/ServerModel.ts
kberg: I'm seeing that now, also. I'll knock this one out soon enough.

src/components/moon/MoonBoard.ts
kberg: Good catch, fixing now.

PR #2389, opened 2021-01-15
Drop IProductionUnits in favor of Units.

kberg:


src/Player.ts
```diff
@@ -390,6 +390,27 @@ export class Player implements ISerializable<SerializedPlayer> {
     }
   };
 
+  public setProductionForTest(units: Partial<Units>) {
```
bafolts, line 393: The restriction of `private` is only a typescript formality. We could set these values from within `TestingUtils` with:

```typescript
export const setPlayerProduction(player: Player, units: Partial<Units>) {
    if (units.megacredits !== undefined) {
        (player as any).megaCreditProduction = units.megacredits;
    }
    // following this pattern of using (player as any) to set the private properties
}
```

This would avoid needing to have a function that we hope doesn't end up being used based on it's name. What do you think?

kberg: Yeesh  gah OK

kberg: For the record I don't love this, but that's fine. I'd rather subclass Player that comes out of TestPlayer. But right now I want to get this submitted.

bafolts: Yes subclass Player even better. Didn't consider that. The mark the properties as protected in Player.

bafolts (approved):

src/Player.ts
kberg: In another PR.

PR #2396, opened 2021-01-15
timer on by default

vsrisuknimit:
Setting timer option to be checked by default.

image

This is only for the info-only timer as its current state. If we ever have a timer with penalty once someone runs out of time, I don't think it should be on by default.


kberg (approved):

kberg:
Actually I would prefer it off during tests. I already merged, but if you don't mind removing it from there, I'd be grateful.

PR #2402, opened 2021-01-15
Separate Event

### LINKED ISSUE #1426: Small feature request: rearrange tags in summary table
- Move none closer to wild tag since it is still a tag, somewhat.
- Event to the right since it is not really an on-going tag at all.
- Small thin line between separating None and Event from the rest of tags. Good to know how many tags you actually have.

image (10)

Tagging @chosta

chosta:
closes #1426

> Move none closer to wild tag since it is still a tag, somewhat.

@vsrisuknimit I'm not sure about moving `none` closer to `wild` just for the sake of it. It will complicate things in code if I do it anyway. But the separation would be helpful for those tag counting events and cards for sure!

the-best

### DIFF
--- a/src/cards/Tags.ts
+++ b/src/cards/Tags.ts
@@ -11,6 +11,6 @@ export enum Tags {
     ANIMAL = 'animal',
     CITY = 'city',
     WILDCARD = 'wild',
-    EVENT = 'event',
     MOON = 'moon',
+    EVENT = 'event', // please, always keep event at the tail
 }
--- a/src/components/Tag.ts
+++ b/src/components/Tag.ts
@@ -1,9 +1,10 @@
 import Vue from 'vue';
+import {Tags} from '../cards/Tags';
 
 export const Tag = Vue.component('tag', {
   props: {
     tag: {
-      type: String,
+      type: String as () => Tags,
     },
     size: {
       type: String,
@@ -22,7 +23,6 @@ export const Tag = Vue.component('tag', {
       if (this.type !== undefined) {
         classes.push(`tag-type-${this.type}`);
       }
-
       return classes.join(' ');
     },
   },
--- a/src/components/TagCount.ts
+++ b/src/components/TagCount.ts
@@ -1,10 +1,11 @@
 import Vue from 'vue';
 import {Tag} from './Tag';
+import {Tags} from '../cards/Tags';
 
 export const TagCount = Vue.component('tag-count', {
   props: {
     tag: {
-      type: String,
+      type: String as () => Tags,
     },
     count: {
       type: Number,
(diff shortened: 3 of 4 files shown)


src/components/Tag.ts
```diff
@@ -1,9 +1,10 @@
 import Vue from 'vue';
+import {Tags} from '../cards/Tags';
 
 export const Tag = Vue.component('tag', {
   props: {
     tag: {
-      type: String,
+      type: String as () => Tags,
```
bafolts, line 7: Sweet.

src/components/TagCount.ts
```diff
@@ -29,6 +30,12 @@ export const TagCount = Vue.component('tag-count', {
       if (this.count === 0) {
         classes.push('tag-no-show');
       }
+      // event tag is added to the tail because it is not counted for actual tag for some cards and turmoil events (e.g. Interplanetary Trade)
+      if (this.tag === Tags.EVENT) {
+        classes.push('tag-event-separate');
+      }
+      console.log(this.type);
```
bafolts, line 37: This was probably added as debugging. Please remove before merging.

bafolts (approved):
Using `String as () => Tags` is great change! Spotted one `console.log` I would remove before merging.

vsrisuknimit:
@chosta Looks great!

kevinb9n:
@chosta Can we please please move the so-called "wild tag" to the other side of the line too? People need help understanding that it is NOT itself a tag. I was even talking to an experienced player who accidentally thought he had gotten to 9 tags for purposes of that turmoil event and really torpedoed his game.

src/cards/Tags.ts
```diff
@@ -11,6 +11,6 @@ export enum Tags {
     ANIMAL = 'animal',
     CITY = 'city',
     WILDCARD = 'wild',
-    EVENT = 'event',
     MOON = 'moon',
+    EVENT = 'event', // please, always keep event at the tail
```
kberg, line 15: Why does event need to be at the end, specifically?

chosta: Because I'm lazy. I'll do it right in the next PR.

chosta:
@kevinb9n @vsrisuknimit can we decide what needs to be before the separator?
I thought wild tag was counted towards Interplanetary Trade, if not maybe we should move it?

src/cards/Tags.ts
kberg: The best response and worst response. (A good response.) OK, thanks.

src/components/TagCount.ts
```diff
@@ -29,6 +30,11 @@ export const TagCount = Vue.component('tag-count', {
       if (this.count === 0) {
         classes.push('tag-no-show');
       }
+      // event tag is added to the tail because it is not counted for actual tag for some cards and turmoil events (e.g. Interplanetary Trade)
+      if (this.tag === Tags.EVENT) {
+        classes.push('tag-event-separate');
```
kberg, line 35: nit: consider a dedicated div as a separator.

kevinb9n:
> @kevinb9n @vsrisuknimit can we decide what needs to be before the separator?
> I thought wild tag was counted towards Interplanetary Trade, if not maybe we should move it?

If you have all 10 tag types (Venus: 11), then a "wild tag" will *not* cause Interplanetary Trade to give you 11 mc production (Venus: 12). You can only make it become some real tag type, but you already have all of them.

PR #2405, opened 2021-01-15
Simplified some cards with the new `canPlay()`

sienmich:


src/cards/base/Capital.ts
```diff
@@ -50,9 +49,9 @@ export class Capital extends Card implements IProjectCard {
     });
   }
   public canPlay(player: Player, game: Game): boolean {
-    return player.getProduction(Resources.ENERGY) >= 2 &&
-        game.checkMinRequirements(player, GlobalParameter.OCEANS, 4) &&
-        game.board.getAvailableSpacesForCity(player).length > 0;
+    return super.canPlay(player) &&
```
kberg, line 52: So, this is tricky. I would like to avoid calling super.canPlay() with EVERY call to canPlay. In the same way standard projects have their `essense` methods, the same thing would apply here. However, I don't recommend calling it `essence` because the global parameters are what you might also call _essential._

kberg: I'll add: I'm doing something mildly similar with Moon behaviors, but my long-term plan is to keep The Moon outside key code execution paths until it's more easily adopted, and also because Moon cards don't need the same degree of backward compatibility that these cards need. It is why I consider that a slightly separate case.

sienmich: Agree. Pushed. I called it `canPlayAdditionalChecks`.

src/cards/prelude/LavaTubeSettlement.ts
```diff
@@ -45,7 +45,7 @@ export class LavaTubeSettlement extends Card implements IProjectCard {
     return LavaFlows.getVolcanicSpaces(player);
   }
 
-  public canPlay(player: Player): boolean {
+  protected canPlayAdditionalChecks(player: Player, game: Game): boolean {
```
sienmich, line 48:
```suggestion
  protected canPlayAdditionalChecks(player: Player): boolean {
```

src/cards/Card.ts
```diff
@@ -74,10 +74,17 @@ export abstract class Card {
...
     return this.properties.productionDelta;
   }
-  public canPlay(player: Player, _game?: Game) {
-    if (this.properties.metadata.requirements === undefined) {
-      return true;
+
+  // Checks if requirements are satisfied and runs canPlayAdditionalChecks.
+  // Do not override, change the canPlayAdditionalChecks in cards implementation.
+  public canPlay(player: Player, game?: Game) {
+    if (this.properties.metadata.requirements?.satisfies(player) === false) {
```
kberg, line 81: This statement here is why I am concerned about having requierments in metadata. Metadata is DATA. But what this method shows is metadata having BEHAVIOR.

src/cards/Card.ts
```diff
@@ -74,10 +74,17 @@ export abstract class Card {
   public get productionDelta() {
     return this.properties.productionDelta;
   }
-  public canPlay(player: Player, _game?: Game) {
-    if (this.properties.metadata.requirements === undefined) {
-      return true;
+
+  // Checks if requirements are satisfied and runs canPlayAdditionalChecks.
+  // Do not override, change the canPlayAdditionalChecks in cards implementation.
+  public canPlay(player: Player, game?: Game) {
```
kberg, line 80: Hmm. We can do better.

As far as I can tell, canPlay is called in five places: `Player`, `ValleyTrust`, `Tactician`, `Playwright` and `CuttingEdgeTechnology`.

The most important one is in `Player.getPlayableCards` which reads like this

```
getPlayableCards() {
  player.cardsInHand.filter((card) => {
    // lots of checks for money and costs.
    if (player.canAfford(cost) && card.canPlay(player, card)) {
      return true;
    };
  }
}
```

That can be simplified by breaking out that whole method into its own. Which I'm going to sneakily call `canPlay` So now there's Player.canPlay(card) and IProjectCard.canPlay(...);

I submit that the right place to make this change is to put it in there. And then the other four classes that use card.canPlay() can adjust accordingly as they need to.

The problem with this is that it changes the way all our tests will need to work. Right now, calling `card.canPlay` is enough, but it won't be. Moving canPlay behavior into Player means cards have to go down that route, which means calling the new `Player.canPlay(card)`. Which includes just one more wrinkle: `Player.canPlay` includes the megacredit cost of a card. That's easy: these tests which are moving from calling card.canPlay to Player.canPlay() also need a set-up step that incldues megacredits.

AND THEN

Once that's done we can start thinking about how to move all the additional cost checks for cards that raise the TR rate. But that's also tricky, and worth discussion at a later point.

Once that settles down, the same change can

kberg (changes requested):

src/cards/Card.ts
kberg: Take a look at https://github.com/bafolts/terraforming-mars/pull/2444 as an example of how I'm doing similar things.

sienmich:
First I'll move the `requirements` to the card from the card metadata, then I'll create new PR similar to this one, in which the `canPlay` will be simplified.

PR #2412, opened 2021-01-16
Reorganize SelectHowToPayForProjectCard, which was needlessly overcomplicated.

kberg:
By simplifying this, it will significantly localize the changes I need to add to accommodate Moon cards with steel and titanium costs.

I am not afraid to say that those tests were a huge pain. But now I know how to do them.

SelectHowToPay is different from SelectHowToPayForProjectCard, but the enterprising coder will merge the two, I am sure.


bafolts (approved):
This PR is a work of art. :clap:

bafolts:
Those tests are great.

PR #2418, opened 2021-01-16
timer off for test

vsrisuknimit:
Timer off for test as requested by @kberg .


kberg (approved):
Thanks for the quick response!

PR #2422, opened 2021-01-16
Add a single tsconfig that represents all files, breaking out behaviors for each custom case.

kberg:
This solves a few problems:
1. It removes the duplication between configurations
2. It restores test files as first class objects within vscode (I assume without disrupting other development patterns)
3. It paves the way for further separating frontend and backend code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,9 +41,9 @@
     "extends": "@istanbuljs/nyc-config-typescript"
   },
   "scripts": {
-    "build": "npm run lint && npm run compile && npm run templates && npm run webpack",
+    "build": "npm run compile && npm run templates && npm run webpack && npm run lint",
     "clean": "rm -r ./build/ ./styles.css || true",
-    "compile": "tsc",
+    "compile": "tsc --build tsconfig-backend.json",
     "cover": "nyc mocha --file build/tests/utils/Vue.js --recursive build/tests",
     "fix": "eslint --fix src server.ts script.ts tests --ext ts",
     "lint": "eslint src server.ts script.ts tests --ext ts",
--- a/tsconfig-backend.json
+++ b/tsconfig-backend.json
@@ -1,0 +1,11 @@
+{
+    // TODO(kberg): splitting this into a frontend and backend, please. But make sure the tests split along with it.
+    "incremental": true,
+    "extends": "./tsconfig.json",
+    "compilerOptions": {
+        "types": ["node"]
+    },
+    "exclude": [
+        "tests",
+    ]
+}
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -14,14 +14,13 @@
         "skipLibCheck": true,
         "sourceMap": true,
         "target": "es5",
-        "types": ["node"],
+        "types": ["node", "mocha", "chai"],
         "moduleResolution": "node",
         "resolveJsonModule": true,
-        "incremental": true
     },
     "include": [
         "src/**/*.ts",
-        "script.ts",
-        "server.ts"
+        "tests/**/*.ts",
+        "*.ts",
     ]
 }
--- a/compile-vue-templates.js
+++ b/compile-vue-templates.js
@@ -1,4 +1,3 @@
-
 /**
  * Provide a mock vue so we can interact with
  * the components from node
--- a/script.ts
+++ b/script.ts
@@ -1,4 +1,3 @@
-
 import Vue from 'vue';
 
 import {translateTextNode} from './src/directives/i18n';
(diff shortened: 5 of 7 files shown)


package.json
```diff
@@ -41,9 +41,9 @@
     "extends": "@istanbuljs/nyc-config-typescript"
   },
   "scripts": {
-    "build": "npm run lint && npm run compile && npm run templates && npm run webpack",
+    "build": "npm run compile && npm run templates && npm run webpack && npm run lint",
```
bafolts, line 44: This change is smart. If while building there is a lint error everything needed to run the game would have built.

tsconfig-backend.json
```diff
@@ -0,0 +1,11 @@
+{
+    // TODO(kberg): splitting this into a frontend and backend, please. But make sure the tests split along with it.
+    "incremental": true,
+    "extends": "./tsconfig.json",
+    "compilerOptions": {
+        "types": ["node"]
```
bafolts, line 6: This split is helpful. It will make sure we don't ever accidentally use something specific to node on the client by mistake.

tsconfig-backend.json
```diff
@@ -0,0 +1,11 @@
+{
+    // TODO(kberg): splitting this into a frontend and backend, please. But make sure the tests split along with it.
```
bafolts, line 2: Comments in JSON files are not valid but typescript seems to be allowing it. Since typescript is working with the comments this is great.

tsconfig.json
```diff
@@ -14,14 +14,13 @@
         "skipLibCheck": true,
         "sourceMap": true,
         "target": "es5",
-        "types": ["node"],
+        "types": ["node", "mocha", "chai"],
```
bafolts, line 17: Only `tsconfig-test` needs `mocha` and `chai`. I would put these in `tsconfig-test` when overriding.

bafolts (dismissed):
I would make sure to only put `types` where they are needed. Someone running code for the server can not assume `mocha` and `chai` exist. Only the spec files can assume those are going to exist.

package.json
kberg: What I like about this is that it'll stop github but still allow me to play test. Or even run unit tests.

tsconfig-backend.json
kberg: We need one more big step, which i might ask you to do -- have 3 separate top directories: backend, common, frontend. I don't know how to move many many things around without breaking everyone's in-flight PRs, though.

tsconfig.json
kberg: That was my original plan. However, this is vscode-friendly. For quite some time since these were split, the UI experience has been pretty difficult -- getting tests up to speed. So, I hope this is an acceptable compromise.

kberg:
> I would make sure to only put `types` where they are needed. Someone running code for the server can not assume `mocha` and `chai` exist. Only the spec files can assume those are going to exist.

I agree, but can we leave this for now, or, do you have an alternative suggestion?

package.json
bafolts: Yes. This is very good. For some reason I enjoy keeping older computers running with Linux. I don't have CPU to spare when I run `npm run build`. Having `npm run lint` run last will be amazing.

tsconfig-backend.json
bafolts: As long as `git mv` is used there shouldn't be any issues. Git will know that the file was moved. While merging the file should be moved for someone that may have a local edit.

Would cards end up in `common`? Parts of a card are used on server and other parts are used on the front end. When we use webpack it isn't bundling `Player` or `Game` for the front end since it knows we only use those values as type references. Putting cards into `common` would make sense to me since they are used on both back end and front end. 

As long as no one else moves any files or adds new files I don't think we would have merge conflicts. Before the github action runs the change is merged with `main` so we should never end up breaking the build with this change.

tsconfig.json
bafolts: What is the UI experience without this change? Does vscode not think calls to `mocha` or `chai` will compile in our test files?

Was reading into what vscode recommends and stumbled upon something called project references I have never used. It looks like it may be what we want to use here.

https://www.typescriptlang.org/docs/handbook/project-references.html

kberg: The code in the tests isn't parsed for live linting, or when finding instances of things in the source tree, test entries don't appear. If I rename something in the source tree, it doesn't rename in the test tree. Those are the three major issues.

kberg:
I'll look at what you suggested, but I think that didn't seem to be a good option, more for what would be a real multi-project scenario.

>  If someone were to use node on client side code the build may pass but we would have an error at runtime on client. Since we haven't hit this already I am most likely over optimizing.

Not necessarily. But I do think splitting out the `frontend` and `common` will help a great deal with things like that.

kberg:
Going to try that approach.

PR #2425, opened 2021-01-16
throw when no spaces available

### LINKED ISSUE #2416: Ares : Game hangs when trying to place a hazard tile when there is no space
We have a game where placing an ocean tile would trigger the placement of hazard tiles on the board. However, there is no place left for a hazard tile, since all free spaces are adjacent to a city tile. This results in the game looping forever trying to find an empty space.

testToPlaceErosionTiles() -> randomlyPlaceHazard() -> getSpaceByOffset() -> getNthAvailableLandSpace() ->
...
    while (idx < 0) {
        idx += spaces.length;
...

Here spaces.length is 0, which means it never gets out of the loop. I guess there should be a check on the spaces to see whether a hazard tile can be placed at all, and if a tile cannot be placed the game should continue as normal.

bafolts:
This throws an error when there are no spaces available to avoid blocking the thread. This would take down heroku instance.

Closes #2416

### DIFF
--- a/src/boards/Board.ts
+++ b/src/boards/Board.ts
@@ -168,6 +168,9 @@ export abstract class Board {
       return this.canPlaceTile(space) && (space.player === undefined || space.player === player);
     }).filter(predicate);
     let idx = (direction === 1) ? distance : (spaces.length - (distance + 1));
+    if (spaces.length === 0) {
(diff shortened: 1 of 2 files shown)


kberg (approved):

kberg:
Thank you.

PR #2437, opened 2021-01-17
Playwrights and ProjectInspection can always play or act

bafolts:
Allows `Playwrights` and `ProjectInspection` to always take action or play. This brings this card up to date with the change to allow a card to do nothing if the user wants to and also avoids infinite loop between the two cards when played together.

### DIFF
--- a/src/cards/community/Playwrights.ts
+++ b/src/cards/community/Playwrights.ts
@@ -25,15 +25,18 @@ export class Playwrights implements CorporationCard {
       return undefined;
     }
 
-    public canAct(player: Player, game: Game): boolean {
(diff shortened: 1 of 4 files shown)


kberg (approved):
Consider a comment in both files pointing to the bug so an enterprising coder doesn't go and create antother crashed instance.

Lynesth (changes requested):
You cannot use an action to simply stall for time unless it's because of one of the 3 exceptions.
#2441 fix the infinite loop issue with Project Inspection and should be used instead of simply allowing those two cards to act all the time.

PR #2439, opened 2021-01-17
Just messing around with the way standard projects are filtered.

kberg:
This matches the way other cards are chosen and will facilitate with The Moon.


kberg:
Closing for the moment.

PR #2440, opened 2021-01-17
Fix not being able to pass anymore

Lynesth:
https://github.com/bafolts/terraforming-mars/commit/0cdd287cf584dc21cf0cae0d2355ccf44651c082#r46035374

Edit: Could need some tests that player inputs are doing what they say they do, not just testing the underlying methods. Not gonna write them though :D


nwai90 (approved):

kberg (approved):
Ugh, sorry.

PR #2441, opened 2021-01-17
Fix infinite loop issue with Playwrights + ProjectInspection

nwai90:
@bafolts Regarding https://github.com/bafolts/terraforming-mars/pull/2437 it would enable two stall actions every time which is not ideal rules-wise. Was discussing with @Lynesth and we verified that this would address the issue. Can you see if it is a better approach?

### DIFF
--- a/src/cards/community/Playwrights.ts
+++ b/src/cards/community/Playwrights.ts
@@ -45,6 +45,9 @@ export class Playwrights implements CorporationCard {
           });
 
           const cost = player.getCardCost(game, selectedCard);
+          // Special case for Law Suit which enters another player's discard pile
+          const isLawSuit = selectedCard.name === CardName.LAW_SUIT ? false : true;
+
           game.defer(new SelectHowToPayDeferred(
             player,
             cost,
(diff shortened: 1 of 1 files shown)


bafolts:
This bug is tedious to replicate. I would be surprised if this resolved the issue but could be wrong. I can probably write a unit test that replicates the issue. If this change does resolve the issue then I can add my unit test as a regression test. If it doesn't then we still have work to do. I am fine solving this another way.

bafolts:
I still see the issue with this unit test:

```
  it.only('Works with Playwrights', function() {
    const playwrights = new Playwrights();                                                                      
    player.corporationCard = playwrights;                                                                       
    player.playedCards.push(card);                                                                              
    player.setActionsThisGeneration(playwrights.name);                                                          
    player.setActionsThisGeneration(card.name);
    expect(card.canPlay(player, game)).is.true;      
  });   
```

This will reach maximum call stack size.

Lynesth:
You are correct @bafolts 
I'll propose a fix in a few minutes.

nwai90:
Closing as superseded by https://github.com/bafolts/terraforming-mars/pull/2459

PR #2444, opened 2021-01-17
Propely preventing invalid alloy deductions with Moon Cards

kberg:
This PR is probably a few too many things. Let me know and I'll simplify it / split it up. But also it depends on PR #2439, so that should get resolved first, anyhow.

How the meat of this PR works:
1. Cards with play-time resource costs have a field that lists those unit costs.
2. When getPlayableProjectCards evalutes each card in a player's hand inidividually, fitlering them in and out.
3. That filter also includes steel and titanium as needed as part of calculating building and space tag cards.
4. That filter also reads the cards play-time resource costs and removes those from the available alloys. That's because if it's a building card that costs 10MC, you COULD spend 8MC and one steel. But if the card costs 10MC, AND, you need a unit of steel when you play the card, that one steel might have to go to the card cost. You can't double-count the steel.
5. Some of this is lessened by four Moon cards which can reduce or eliminate the costs assocaited with those reserve units. Those cards behavior are activated in `MoonExpansion.adjustedReserveUnits`, which is why it's called in a couple of places.

Pause, and we'll continue with the client-side.

6. Now you have to get the (web) client to pay for the card. The card's _adjusted reserve units_ are sent down to the client along with the the rest of the card metadata.
7. When the player selects to pay for that card, the front-end once again has to prevent double-paying with steel, and so it reduces the amount of _available_ steel by the _adjusted reserve unit_ amount. Same goes for titanium. So now the player _can't_ use all the steel in their inventory to pay for the card. They just can't.
8. There's a nice little warning when their full inventory of steel or titanium is not made available.

I also made a change to the `SelectHowToPayForProjectCard` input, just reducing the size of the constructor.


src/Units.ts
```diff
@@ -26,6 +26,27 @@ export namespace Units {
   //   log?: boolean;
   // };
 
+  export const EMPTY: Units = {
+    get megacredits() {
```
bafolts, line 30: I had a large rant prepared here regarding that this would introduce situations where typescript can't help us enforce we don't set these properties.....

Then I remembered `Readonly<T>`.

I would use `export const EMPTY: Readonly<Units>` here and this will be amazing.

kberg: Wish I could have seen that rant.

src/cards/moon/FirstLunarSettlement.ts
```diff
@@ -23,6 +23,8 @@ export class FirstLunarSettlement extends PreludeCard implements IProjectCard {
     });
   }
 
+  public tilesBuilt = [TileType.MOON_COLONY];
```
bafolts, line 26: If this property is static I would use a getter or the static card property mechanism to re-use memory. Now that all I see is `Readonly` if this doesn't change I would use `tilesBuilt: ReadonlyArray<TileType>` in `IMoonCard`.

src/Units.ts
kberg: Your proposal introduces a problem: someone can change the value of the EMPTY constant for everyone else.

This test fails:

```
  it('EMPTY', () => {
    const units: Readonly<Units> = Units.EMPTY;
    const someOtherUnits = Units.EMPTY;
    const units2: Units = units;
    units2.megacredits = 1;
    expect(units.megacredits).eq(0);
    expect(someOtherUnits.megacredits).eq(0);
  });
```

kberg: OH! BOTH Readonly AND the get-only thing.

src/cards/moon/FirstLunarSettlement.ts
kberg: So this raises new questions. There are some static things I could add to the staticproperties but then it starts to get messy. What do you think?

src/models/CardModel.ts
```diff
@@ -10,4 +11,5 @@ export interface CardModel {
     cardType: CardType;
     isDisabled: boolean;
     warning?: string | Message;
+    reserveUnits: Units;
```
bafolts, line 14: I have my `Readonly` hammer and everything is looking like nails to me. This is another location where we don't plan on these `Units` being modified once set. I would use `Readonly<Units>`.

kberg: Good idea.  It sounds like the carpenter should come in and nail everything down at once.

src/cards/moon/ImprovedMoonConcrete.ts
```diff
@@ -0,0 +1,36 @@
...
+
+      metadata: {
+        description: 'Effect: When you build a mine on the Moon, you spend 1 steel less. / Spend 2 steel. Raise Mining Rate 1 step.',
+        cardNumber: 'M37',
+        renderData: CardRenderer.builder((_b) => {}),
+      },
+    });
+  }
+
+  public reserveUnits = Units.of({steel: 2});
```
bafolts, line 25: I believe this is also a static card property. Can you use a getter or set this up so we don't create new `Units` for each instance of this card?

kberg: Same question as above.

(I was also hoping this could be dealt with in a follow-up PR, but if you want to do it now, that's fine -- note that none of these cards will be accessible except in the debug UI. (unless someone uses the "secret" URL to create a game.) and so I assume memory consumption concerns for the moment are minimal.

Speaking of which, I suggest let's get the rest of the cards that can extend Card. I've come up with a fairly simple VSCode macro to simplify those.)

bafolts (approved):

PR #2445, opened 2021-01-17
Confirmation dialog box confirms tile placement, disableable in preferences.

kberg:
image
If you choose Cancel it lets you select another space.

And this can be disabled in preferences.
image

I'd weak the UX just a bit, but it turned out to be simple. What do you think?


alrusdi (approved):
Great first step in correct direction!
I personally don't like standard prompt. It probably should be some kind of inline confirmation widget near the tile you want to place. But it can be done latter.

nwai90:
@kberg This looks neat! Clean and simple 👍

src/components/SelectSpace.ts
```diff
@@ -63,13 +71,15 @@ export const SelectSpace = Vue.component('select-space', {
         elTile.classList.add('board-space--available');
 
         elTile.onclick = () => {
-          clearAllAvailableSpaces();
-          for (let j = 0; j < elTiles.length; j++) {
-            (elTiles[j] as HTMLElement).onclick = null;
+          if (confirm()) {
```
alrusdi, line 74: may be just:
if ( ! confirm()) return;
?

nwai90:
There's also a minor question of whether the default preference setting should be having tile confirmations enabled or disabled, but not much of a deal imo

kberg:
> I personally don't like standard prompt. It probably should be some kind of inline confirmation widget near the tile you want to place. But it can be done latter.

My thoughts exactly. But after spending 30 minutes looking at modal dialogs code I gave up.

Once you add modal dialogs, things get much more complicated, too (the underlying click handlers, currently unclickable because of the dialog box, have to be removed and re-enabled. Or something.)

I don't even care about the standard, I just wish there was a standard dialog "yes, no, [x] don't show this again"

src/components/SelectSpace.ts
kberg: Yes, it could. But  I'm going to fix a second bit -- which is making your selected space more obvious. That's going to make this slightly trickier.

kberg: Also, I don't know how this works with i18-n.

kberg:
**Update** i've tried to make the selected tile stand out, but I am fighting with the DOM's event queue. I tried setTimeout to avoid the microtasks. It works -- *sometimes*

src/components/SelectSpace.ts
```diff
@@ -44,32 +44,74 @@ export const SelectSpace = Vue.component('select-space', {
...
+        }
+      }
+    };
+
+    const confirm = function() {
+      const hideTileConfirmation = PreferencesManager.loadValue('hide_tile_confirmation') === '1';
+      if (hideTileConfirmation) {
+        return true;
+      }
+      return window.confirm('Place your tile here? (This dialog can be disabled in preferences).');
```
bafolts, line 78: The native `window.confirm` method is unique in that it blocks the thread until a response has been provided. If later we modify this to use our own stylized component we will not be able to block the thread. I would use a `cb` argument to `confirm` and use a callback pattern. That way if and when we change this to use a custom component we wont be relying on this blocking the thread and it should drop into place easier.

```
const confirm = function (cb: (result: boolean) => void) {
     // what you have
     const result = window.confirm('Place your tile here?');
     cb(result);
}
```

kberg: I see your point, but might as well do that later.

kberg: Or rather, why now and not later?

bafolts: Since this isn't exposed and it is private to the component should be fine. Every caller to the `confirm` function will have to be updated if we ever use a custom confirm since we can't block the thread like `window.confirm`. The more callers, the more work we will have.

src/components/SelectSpace.ts
```diff
@@ -44,32 +44,74 @@ export const SelectSpace = Vue.component('select-space', {
...
-      const elTiles = document.getElementsByClassName('board-space-selectable');
-      for (let i = 0; i < elTiles.length; i++) {
-        const elTile = elTiles[i] as HTMLElement;
-        const el_id = elTile.getAttribute('data_space_id');
-        if ( ! el_id || ! setOfSpaces[el_id]) continue;
+    const animateSpace = function(tile: Element, activate: boolean) {
+      if (activate) {
+        tile.classList.add('board-space--available');
+      } else {
+        tile.classList.add('board-space--available');
```
bafolts, line 59: I mentioned this to you in a private chat but putting note here to keep track. This is setting the same class name and is most likely a typo introducing a bug.

src/components/SelectSpace.ts
kberg: A-OK.

src/components/SelectSpace.ts
kberg: That fixed so many things. Thanks for spotting it.

kberg:
I actually think this is ready. You can test it by using my Heroku instance: http://terraforming-mars-kberg.herokuapp.com

src/components/SelectSpace.ts
```diff
@@ -75,7 +75,7 @@ export const SelectSpace = Vue.component('select-space', {
       if (hideTileConfirmation) {
         return true;
       }
-      return window.confirm('Place your tile here? (This dialog can be disabled in preferences).');
+      return window.confirm('Place your tile here?\n\n(This confirmation can be disabled in preferences).');
```
alrusdi, line 78: Can you please make this text translatable?

chosta:
> Once you add modal dialogs, things get much more complicated, too (the underlying click handlers, currently unclickable because of the dialog box, have to be removed and re-enabled. Or something.)

Ahh. My archenemy - the custom modal dialog. Even that the native confirm dialog brings us back to 1995, I'll take it until we get proper modals (that might be never) :)

alrusdi:
> Ahh. My archenemy - the custom modal dialog. Even that the native confirm dialog brings us back to 1995, I'll take it until we get proper modals (that might be never) :)

We already have ready-to-use modals in our css framework 
https://picturepan2.github.io/spectre/components/modals.html#modals

src/components/SelectSpace.ts
```diff
@@ -44,32 +44,73 @@ export const SelectSpace = Vue.component('select-space', {
...
+              this.$data.spaceId = tile.getAttribute('data_space_id');
+              tile.classList.add('board-space--selected');
+              this.saveData();
+            } else {
+              animateSpace(tile, false);
+              // without the timeout, this flashing animation gets out of sync with
+              // the other tiles, like the blinker of the car in front of you.
+              setTimeout(() => animateAvailableSpaces(tiles), 0);
+            }
+          }, 15); // 15ms gives the entire DOM a chance to catch up.
```
chosta, line 113: 15 ms might not be enough depending on OS, Browser version etc.
I say we roll with it and see how it goes.

chosta (approved):
Good old DOM intervention. Never goes wrong :)
I'm giving a thumbs up although would like to have the setting disabled by default.
Also for the future:
1) Proper modals (I wish)

chosta:
> We already have ready-to-use modals in our css framework
> https://picturepan2.github.io/spectre/components/modals.html#modals
> 
> But modals is bad UX in general. We need to consider better approaches.

Yes, we have these. Haven't checked allt he stuff they are capable of. 
Also agree that modals are being overused and should be avoided if possible.

Issue #2443, opened 2021-01-17
Tactician milestone miscounting cards with requirements


puuj:
Happening in latest build and seems to be an issue as far as a couple days back. 
<img width="145" alt="image" src="https://user-images.githubusercontent.com/58322/104833159-814c8800-584b-11eb-9296-202c82a52670.png">
<img width="574" alt="image" src="https://user-images.githubusercontent.com/58322/104833172-8e697700-584b-11eb-97fd-473972075b53.png">

kberg:
What five cards do you have? What is this screenshot supposed to show as it pertains to the award?

vsrisuknimit:
@sienmich #2378 and #2404 might have something to do with this. The bug is confirmed by other players.

sienmich:
Yep, sorry, will fix

PR #2452, opened 2021-01-17
Fixes Tactician milestone

sienmich:
Fixes Tactician milestone and removes unnecesarry `hasRequirements` attribute from IProjectCard

### DIFF
--- a/src/cards/Card.ts
+++ b/src/cards/Card.ts
@@ -13,7 +13,6 @@ interface StaticCardProperties {
   adjacencyBonus?: IAdjacencyBonus;
   cardType: CardType;
   cost?: number;
-  hasRequirements?: boolean;
   initialActionText?: string;
   metadata: CardMetadata;
   name: CardName;
@@ -49,9 +48,6 @@ export abstract class Card {
   }
   public get cost() {
     return this.properties.cost === undefined ? 0 : this.properties.cost;
-  }
-  public get hasRequirements() {
-    return this.properties.hasRequirements;
   }
   public get initialActionText() {
     return this.properties.initialActionText;
--- a/src/cards/IProjectCard.ts
+++ b/src/cards/IProjectCard.ts
@@ -6,6 +6,5 @@ import {Resources} from '../Resources';
 export interface IProjectCard extends ICard {
     canPlay?: (player: Player, game: Game) => boolean;
     cost: number;
-    hasRequirements?: boolean;
     bonusResource?: Resources | undefined;
 }
--- a/src/cards/base/Asteroid.ts
+++ b/src/cards/base/Asteroid.ts
@@ -18,7 +18,6 @@ export class Asteroid extends Card implements IProjectCard {
       name: CardName.ASTEROID,
       tags: [Tags.SPACE],
       cost: 14,
-      hasRequirements: false,
 
       metadata: {
         description: 'Raise temperature 1 step and gain 2 titanium. Remove up to 3 Plants from any player.',
--- a/src/cards/base/BigAsteroid.ts
+++ b/src/cards/base/BigAsteroid.ts
@@ -18,7 +18,6 @@ export class BigAsteroid extends Card implements IProjectCard {
       name: CardName.BIG_ASTEROID,
       tags: [Tags.SPACE],
       cost: 27,
-      hasRequirements: false,
 
       metadata: {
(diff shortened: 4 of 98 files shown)


sienmich:
Closes https://github.com/bafolts/terraforming-mars/issues/2443

kberg:
I'm going to ask Brian to review this, since static card properties are his area of expertise, and I've already expressed concern about the way card properties and metadata are commingling, so I'd rather just stay out of this one

nwai90 (approved):
Functionally I believe this works fine. I have similar concerns to @kberg in that basing Tactician and CuttingEdgeTechnology's checks on having requirements in a card's metadata conflates business logic with UI rendering logic.

That said as long as we ensure that any new card(s) with real requirements added always have a `requirements` attribute in metadata it should be ok 👍

sienmich:
> That said as long as we ensure that any new card(s) with real requirements added always have a `requirements` attribute in metadata it should be ok.

Well, right now that's the only place to put the requirements (except the card description), so it will be hard to miss.
I understand the 'mixing UI and functionalities' concern.

Should I move the `requirements` from metadata to card itself? Maybe this would improve the readability?

nwai90:
@sienmich It is generally fine for me since the logic is still quite clear, let's see what others think about it

kberg:
> Functionally I believe this works fine. I have similar concerns to @kberg in that basing Tactician and CuttingEdgeTechnology's checks on having requirements in a card's metadata conflates business logic with UI rendering logic.
> 
> That said as long as we ensure that any new card(s) with real requirements added always have a `requirements` attribute in metadata it should be ok

Here are my concerns: first, this PR isn't really about fixing the tactician milestone, it's about restructiring the code. Which is fine, for sure.  

In another PR you can see the effect of this change, which incldues this line:

```
  if (this.properties.metadata.requirements?.satisfies(player) === false) {
```

This is **metadata**, which means it should store **data**. But this is **behavior** and so does not belong in metadata.

Finally, I'm reading tactician, and it has this line:

```
const hasRequirements = card.canPlay && (card.hasRequirements === undefined || card.hasRequirements);
```

It's not clear why `card.canPlay` is a prerequisite. A comment describing what this line is doing would be helpful.

Lynesth:
@kberg Because that code is a bit... meh.
Instead of having a simple `hasRequirements = true` for Tactician, it's actually checking if the card has a `canPlay` method.
Then, it assumes that it is a card with requirements, unless it has a `hasRequirements = false`.

kberg:
It's possible, I was reading from my VSCode instance.

kberg:
My first message was a little brusque, because I was trying to get back to the first PR which included the line that helped me see the concernse I had with the first one. So first, thanks for the fix, for certain. But also, please think about how you can move behavior out of metadata. Not now, but eventually.

sienmich:
@kberg 

> Finally, I'm reading tactician, and it has this line:
> 
> ```
> const hasRequirements = card.canPlay && (card.hasRequirements === undefined || card.hasRequirements);
> ```
> 
> It's not clear why `card.canPlay` is a prerequisite. A comment describing what this line is doing would be helpful.

Yep, that's the old version, the one I simplified to `const hasRequirements = card.metadata.requirements !== undefined;`, much cleaner imo.

> this PR isn't really about fixing the tactician milestone

It is. To fix the tactician I had to make the change described above. (Because earlier I deleted some `canPlay` methods, which were automaticly calculated from the metadata.requirements; my fault not thinking about the tactician, sry).
And now since the `hasRequirements` was unused, I deleted it.

> This is metadata, which means it should store data. But this is behavior and so does not belong in metadata.

Agree, I'll move the requirements level up, to the card soon.

kberg:
> > this PR isn't really about fixing the tactician milestone
> 
> It is. To fix the tactician I had to make the change described above. (Because earlier I deleted some `canPlay` methods, which were automaticly calculated from the metadata.requirements; my fault not thinking about the tactician, sry).
> And now since the `hasRequirements` was unused, I deleted it.

Sorry, I said the wrong thing here. I did mean that it was such a surprise that the meat of this was moving data around, and it wasn't clear what was going on. That comment about "because earler I deleted some 'canPlay' methods" is really useful and I think would have helped me dive in to review this. But no big deal, it's not as if I don't make mistakes (I make A LOT OF THEM.)

PR #2454, opened 2021-01-17
add solarphaseoption to gameoptionmodel

vsrisuknimit:
This allows WGT to display correctly here:

image


kberg (approved):

kberg:
Nice and simple. That's the kind of PR I like.

PR #2456, opened 2021-01-18
Moon #12: Add 9 more cards. Clean up a fair amount of reuse, and start to change how Units works, because it's looking like it's going to be a little backwards unless I think through carefully.

kberg:
Depends on #2444


src/cards/Card.ts
```diff
@@ -67,8 +67,8 @@ export abstract class Card {
   public get tags() {
     return this.properties.tags === undefined ? [] : this.properties.tags;
   }
-  public get productionDelta() {
-    return this.properties.productionDelta;
+  public get productionDelta(): Units {
```
bafolts, line 70: With the addition of `negative` this seems like `Units` is being setup to be immutable. Every place that `Units` is used I believe it is never mutated. I think we should reflect this within `Units`.

```
diff --git a/src/Units.ts b/src/Units.ts
index 02e3bc34..505c37af 100644
--- a/src/Units.ts
+++ b/src/Units.ts
@@ -10,12 +10,12 @@ import {Resources} from './Resources';
 // Units represents any value of each standard unit.
 // Could be positive or negative, depending on how it's used.
 export interface Units {
-  megacredits: number;
-  steel: number;
-  titanium: number;
-  plants: number;
-  energy: number;
-  heat: number;
+  readonly megacredits: number;
+  readonly steel: number;
+  readonly titanium: number;
+  readonly plants: number;
+  readonly energy: number;
+  readonly heat: number;
 }
```
Without this in place the `Units.EMPTY` would throw a runtime error if we tried to mutate it.

src/cards/moon/MoonCard.ts
```diff
@@ -0,0 +1,36 @@
...
+import {IMoonCard} from './IMoonCard';
+
+export interface MoonCardProperties {
+  reserveUnits?: Units,
+  tilesBuilt?: Array<TileType>
+}
+
+export abstract class MoonCard extends Card implements IProjectCard, IMoonCard {
+  constructor(properties: StaticCardProperties,
+    private moonCardProperties: MoonCardProperties) {
```
bafolts, line 15: This is a good way of separating these moon specific properties. This is going to store a new array for `tilesBuilt` and a new object for `reserveUnits`. Will have to use the `Map` like `staticCardProperties` of `Card.ts` to re-use one instance.

src/moon/MoonExpansion.ts
```diff
@@ -152,36 +153,36 @@ export class MoonExpansion {
...
+      tiles = moonData.moon.spaces.filter(
+        (space) => {
+          const spaceTileType = space.tile?.tileType;
+          let include: boolean = true;
+          if (tileType === TileType.MOON_COLONY) {
+            include = spaceTileType === TileType.MOON_COLONY || spaceTileType === TileType.LUNAR_MINE_URBANIZATION;
+          } else if (tileType === TileType.MOON_MINE) {
+            include = spaceTileType === TileType.MOON_MINE || spaceTileType === TileType.LUNAR_MINE_URBANIZATION;
+          } else {
+            include = include && space.tile?.tileType === tileType;
```
bafolts, line 174: The `space.tile?.tileType` is available in the `spaceTileType` variable on line 167.

bafolts (approved):
I had minor suggestions and optimizations. Making `Units` immutable by making every property `readonly` would most likely help with the pattern you are introducing with functions like `negative` and it should also prevent runtime errors as typescript would catch any place we were attempting to mutate `Units`. This would avoid runtime errors if we ever indirectly attempted to mutate `Units.EMPTY`.

:rocket: 
image

src/cards/Card.ts
kberg: I don't recommend units be fully immtuable. In fact, at some point before too long I am going to suggest making changes to player such that it has a Units representation for both individual resources and production, and those should be mutable.

bafolts: If the `Units.EMPTY` ends up being written to we will have a runtime exception. This is concerning to me since typescript will not enforce it. Can we allow `Units.EMPTY` to be mutable and return a new instance?

kberg: How will Units.EMPTY be written to?

bafolts: It may not be happening now but I could see it happening in the future. For example if someone were to have code similar to:

```
const randomMoonCard: MoonCard = this.getRandomMoonCard();
randomMoonCard.reserveUnits.megacredits += 1; 
```

This will compile without error but would only error during runtime on certain cards which had `Units.EMPTY` which can't have it's properties tested. Now this example I provided will most likely never be used but I used it to I hope reflect my concern. It isn't always clear when `Units.EMPTY` is used and typescript wont enforce the readonly restriction.

kberg: I suspect there no answer that is efficient and error-proof, and so perhaps making a definsive copy every time will have to do.

kberg: I'm going to do that in a follow-up.

src/cards/moon/MoonCard.ts
kberg: It is, but also it's ugly in the constructor, as you will see.

src/moon/MoonExpansion.ts
kberg: Ah, thanks!

src/cards/Card.ts
bafolts: Sounds good. I don't think it is worth blocking the PR. It is an optimization if anything. Having worked with javascript before typescript was there to help I want typescript to be able to help us whenever possible! Can't go back.

PR #2459, opened 2021-01-18
Fix Playwrights / Project Inspection interaction

Lynesth:
Should be good to fix https://github.com/bafolts/terraforming-mars/pull/2441
Wrote a verbose test to make sure things won't be broken again later.

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1092,15 +1092,6 @@ export class Player implements ISerializable<SerializedPlayer> {
     return Math.max(cost, 0);
   }
 
-  private addPlayedCard(card: IProjectCard): void {
-    this.playedCards.push(card);
-    this.lastCardPlayed = card;
-
-    // Playwrights hook for Conscription and Indentured Workers
-    // Cards "removed from play" can still have an effect
-    this.removedFromPlayCards = this.removedFromPlayCards.filter((card) => card.getCardDiscount === undefined);
-  }
-
   private canUseSteel(card: ICard): boolean {
     return card.tags.indexOf(Tags.BUILDING) !== -1;
   }
@@ -1201,7 +1192,7 @@ export class Player implements ISerializable<SerializedPlayer> {
     return 0;
   }
 
-  public playCard(selectedCard: IProjectCard, howToPay?: HowToPay): undefined {
+  public playCard(selectedCard: IProjectCard, howToPay?: HowToPay, addToPlayedCards: boolean = true): undefined {
     // Pay for card
     if (howToPay !== undefined) {
       this.steel -= howToPay.steel;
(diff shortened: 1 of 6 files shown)


src/cards/promo/ProjectInspection.ts
```diff
@@ -18,13 +19,14 @@ export class ProjectInspection implements IProjectCard {
...
 
-      if (
-        player.corporationCard !== undefined &&
-                player.getActionsThisGeneration().has(player.corporationCard.name) &&
-                player.corporationCard.action !== undefined &&
-                player.corporationCard.canAct !== undefined &&
-                player.corporationCard.canAct(player, game)) {
-        result.push(player.corporationCard);
+      if (player.corporationCard !== undefined && player.getActionsThisGeneration().has(player.corporationCard.name)) {
+        if (player.corporationCard.name !== CardName.PLAYWRIGHTS || (player.corporationCard as Playwrights).getCheckLoops() < 2) {
```
nwai90, line 23: As discussed, this probably should have a comment and/or define the `2` as a proper named value somewhere. The rest seems fine to me 👍

Lynesth:
I tried to improve it by using a boolean instead of a counter, but couldn't wrap my head around it (its late).

Also there's 2 edge cases that are not covered by the behavior:

#### Case 1
Setup: No blue action used yet. **Project Inspection** + another playable event in the event pile.

Result: If you use **Playwrights**, you cannot select **Project Inspection** even though you could then replay **Playwrights** into another event.

Reason: When checking if **Project Inspection** can be played, it finds no current used blue action, even though **Playwrights** one will be used.

#### Case 2
Setup: **Playwrights** and another blue action already used. No playable event in an event pile.

Result: If you play **Project Inspection** from your hand, you cannot select **Playwrights** even though you could then use it to replay **Project Inspection** into using another blue action.

Reason: Same as above except that it's because **Playwrights** cannot find any playable card in an event pile, even though **Project Inspection** will be in it.

-----

To have those work we would need to cheat by falsely adding **Project Inspection** to the playedCards when checking the actions it can use and set **Playwrights** as already used when checking the events it can replay.
Shouldn't be hard to fix, but not sure that's worth it.

Vote: Let's just remove that corp from the game :D

src/Player.ts
```diff
@@ -1235,6 +1226,7 @@ export class Player implements ISerializable<SerializedPlayer> {
     }
 
     if (selectedCard.cardType !== CardType.PROXY) {
```
bafolts, line 1228: Unrelated to this change but checking that `selectedCard.name !== CardName.SCIENCE_TAG_BLANK_CARD` would be easier to read here. That is the only `CardType.PROXY`. That would allow us to remove `CardType.PROXY`.

bafolts (approved):
A good case could be made to removing `Playwrights`. These two cards conflict so the solution is going to have to involve a comprimise between how they work together.

PR #2461, opened 2021-01-18
Remove redundant properties in playermodel

vsrisuknimit:
PlayerModel already includes GameOptionsModel. I remove all the properties from PlayerModel that already exists in GameOptionsModel and adjusted the call appropriately.


kberg (approved):
Thanks for scouring the code.

PR #2462, opened 2021-01-18
Update blue card sorting list with Ares and fix Earth Catapult ordering

vsrisuknimit:
- Previously we return index 0 for Earth Catapult. This causes `ActiveCardsSortingOrder.get(cardA.name as CardName) || size` to become `0 || size` for Earth Catapult which has index of 0 .The logic statement considers 0 to be false and picks size instead so Earth Catapult was pushed to the back of the blue cards list. @Lynesth helped me with adding +1 index to the map to prevent this.
- Remove size. Use -1 if card is not found instead. This way we make sure that corporation power is in the front, not back.
- Add Ares blue cards to the ordering list.

### DIFF
--- a/src/components/ActiveCardsSortingOrder.ts
+++ b/src/components/ActiveCardsSortingOrder.ts
@@ -2,8 +2,9 @@ import {CardName} from '../CardName';
 import {CardModel} from '../models/CardModel';
 
 export function sortActiveCards(inCards: Array<CardModel>): Array<CardModel> {
+  const firstCardIndex = -1;
   return inCards.slice().sort(function(cardA, cardB) {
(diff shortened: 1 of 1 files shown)


vsrisuknimit:
@kberg can you help check whether these are all the blue Ares cards? Are they in the good order with correct category?

nwai90 (approved):

kberg:
> @kberg can you help check whether these are all the blue Ares cards? Are they in the good order with correct category?

image

let's see: Bioengineering Enclosure, ecological Survey, Ecological Zone, Geological Survey, Industrial Center, Marketing Experts, Ocean Sanctuary, Restricted Area.

kberg (approved):

vsrisuknimit:
@kberg I think I got everything then.

PR #2463, opened 2021-01-18
Fixed CuttingEdgeTechnology

sienmich:
So turns out that I did fix the Tactician but not the Cutting Edge. Sorry for that.
More detailed explanation in https://github.com/bafolts/terraforming-mars/pull/2452

### DIFF
--- a/src/cards/promo/CuttingEdgeTechnology.ts
+++ b/src/cards/promo/CuttingEdgeTechnology.ts
@@ -19,7 +19,7 @@ export class CuttingEdgeTechnology implements IProjectCard {
     }
 
     public getCardDiscount(_player: Player, _game: Game, card: IProjectCard) {
-      if (card.canPlay && card.metadata.requirements !== undefined) return 2;
+      if (card.metadata.requirements !== undefined) return 2;
       return 0;
     }
 


kberg (approved):

kberg:
Is there a test that should have been written?

vsrisuknimit:
@kberg @sienmich Give some one cutting edge, 6 mc and colonizer training camp. They should be able to play it. The discount should applies before checking canPlay.

sienmich:
@vsrisuknimit 
> They should be able to play it. The discount should applies before checking canPlay.

What do you mean? Have you checked this interaction after this PR?
The discounts are applying correctly now from what I see.

sienmich:
I updated the `getCardDiscount` function that returns the discounts for other cards, the order of discount / checking / playing remains unchanged

vsrisuknimit:
@sienmich I mean as a test. To prevent detect the problem what this PR is fixing in the future.

sienmich:
Oh, I missed that comment. Yep, that sounds like a good test.

vsrisuknimit:
> Oh, I missed that comment. Yep, that sounds like a good test.

@sienmich can you add it? I am still not experienced with constructing a test.

sienmich:
Sure, I'll

PR #2464, opened 2021-01-18
Simplified Player.getPlayableCards()

sienmich:
Very small PR this time.
I'll try to keep it that way!


kberg (commented):
So I know this is a simple PR, but it will complicate reimplementing #2444

That's why I suggested a much smaller change. Just take the entire innards of the filter and make it a top-level function.

PR #2467, opened 2021-01-18
Show card ID on cards.

kberg:
image
image
image


vsrisuknimit:
@kberg Is it possible to show the ID only on the debug-ui page? I don't think we need it in the main game page. Or have an option in preferences for hiding it?

kberg:
I don't think this is worth a preference or a toggle. As it is a standard value on cards, I'd like to show it. If not, let's just get rid of the card numbers from the metadata.

chosta (commented):
While you did a great job with the visuals, I must admit that these numbers are just clutter. My vote is on removing them in general. I can't think of any use at the moment. They don't bring any value to the players or the devs. We can easily add them back if we find a reason in the future.

kberg:
Here's the value I have as a dev: I've been wanting this field throughout the entire Moon development cycle. I'm managing bringing 120-odd cards to life, and their visual cue would have been very helpful -- would _still_ be helpful.

I wish, back when, the game indexed cards by this number, and not by  their name. We could change that, but I don't have strong feelings that way.

As it is still helpful to me, and at least one other person thumbed up my perspective, then perhaps, yes, a toggle? I don't love it that option, but can accept it. :/

vsrisuknimit:
@kberg  Can we do it by CSS such that the ID number only shows up in debug-ui? Do you need to see the number during the game?

chosta:
I'm ok with numbers as an option. No need to be css patched just a preference. Also, I'm suspicious of the resource counter being overlapping with the number atm. Let's just keep them out of the game by default and turn them on if we need to debug during a game for some reason.

chosta (approved):
Thank you

kberg:
> Thank you

And you!

PR #2468, opened 2021-01-18
Add productionDelta for Space Elevator when using Robotic Workforce.

kberg:
I'll follow up with a test in a moment. This should get the active game going.


vsrisuknimit:
Looks good!

vsrisuknimit (approved):

PR #2469, opened 2021-01-18
Switch to static card metadata for ... a bunch ... of venus cards. I don't know how many.

kberg:


chosta (approved):
The easieast reviews are the hardest to make :)

PR #2470, opened 2021-01-18
Remove the array validation tests for the milestone award selector.

kberg:
This was more interesting a while ago, but the synergy list is on the verge of update with milestones and awards not related. The code that did this can be recovered and used if someone wants to provide a big update.

### DIFF
--- a/src/cards/base/SpaceElevator.ts
+++ b/src/cards/base/SpaceElevator.ts
@@ -8,6 +8,7 @@ import {Game} from '../../Game';
 import {Resources} from '../../Resources';
 import {CardName} from '../../CardName';
 import {CardRenderer} from '../render/CardRenderer';
+import {Units} from '../../Units';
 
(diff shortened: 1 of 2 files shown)


chosta (approved):
There is one commit that changes `SpaceElevator.ts`, please double check if you intended it for this PR.

src/cards/base/SpaceElevator.ts
```diff
@@ -16,6 +17,7 @@ export class SpaceElevator extends Card implements IActionCard, IProjectCard {
       name: CardName.SPACE_ELEVATOR,
       tags: [Tags.SPACE, Tags.BUILDING],
       cost: 27,
+      productionDelta: Units.of({titanium: 1}),
```
vsrisuknimit, line 20: This is already fixed in #2468.

kberg: Dangit, I misused a branch.

kberg:
A fresh pull should solve this.

PR #2474, opened 2021-01-18
Move the meat of getPlayableCards to canPlay(). This made some game instances superfluous.

kberg:
…nstances superfluous.


src/Player.ts
```diff
@@ -1594,63 +1594,65 @@ export class Player implements ISerializable<SerializedPlayer> {
...
-      // reserve costs that a player cannot afford.
-      if (card.reserveUnits !== undefined) {
-        const reserveUnits = MoonExpansion.adjustedReserveCosts(this, card);
-        // If there isn't enough steel to meet the purchase reserve, this isn't a playable card.
-        if (steel < reserveUnits.steel) {
-          return false;
-        }
-        // Set aside reserve units in case the card has a building tag.
-        steel = Math.max(0, steel - reserveUnits.steel);
+    const playableCards = candidateCards.filter((card) => this.canPlay(card));
```
chosta, line 1597: If `playableCards` is not used elsewhere can't we just do:
`return candidateCards.filter((card) => this.canPlay(card))`

chosta (approved):

src/Player.ts
```diff
@@ -1594,63 +1594,65 @@ export class Player implements ISerializable<SerializedPlayer> {
...
+    if (card.reserveUnits !== undefined) {
+      const reserveUnits = MoonExpansion.adjustedReserveCosts(this, card);
+      // If there isn't enough steel to meet the purchase reserve, this isn't a playable card.
+      if (steel < reserveUnits.steel) {
+        return false;
       }
-      if (canUseTitanium) {
-        maxPay += titanium * this.getTitaniumValue();
+      // Set aside reserve units in case the card has a building tag.
+      steel = Math.max(0, steel - reserveUnits.steel);
```
Lynesth, line 1618: This cannot be `< 0`. Same for titanium below.

kberg: Most probably not. But I'd like to do some kind of defensive thing so someone don't wind up in a situation with a negative steel count. Or are you saying that we should remove that safety net?

Lynesth: You have that net just above, that's what I mean.
```ts
      if (steel < reserveUnits.steel) {
        return false;
      }
```
So after that, you are sure that `steel - reserveUnits.steel` is `>= 0`.

kberg: Oh! Yes. Thanks.

PR #2477, opened 2021-01-19
show remaining mega credits for starting cards

bafolts:
Displays remaining megacredits during initial card selection.

image

### DIFF
--- a/src/components/SelectInitialCards.ts
+++ b/src/components/SelectInitialCards.ts
@@ -4,6 +4,7 @@ import Vue from 'vue';
 import {Button} from './common/Button';
 import {CardFinder} from '../CardFinder';
 import {CardName} from '../CardName';
+import * as constants from '../constants';
 import {CorporationCard} from '../cards/corporation/CorporationCard';
 import {PlayerInputModel} from '../models/PlayerInputModel';
 import {PlayerModel} from '../models/PlayerModel';
@@ -45,6 +46,15 @@ export const SelectInitialCards = Vue.component('select-initial-cards', {
       }
       return this.playerinput.options[idx];
     },
+    getRemainingMegacredits: function() {
+      if (this.selectedCorporation === undefined) {
+        return NaN;
+      }
+      let remaining = this.selectedCorporation.startingMegaCredits;
+      const cardCost = this.selectedCorporation.cardCost === undefined ? constants.CARD_COST : this.selectedCorporation.cardCost;
+      remaining -= this.selectedCards.length * cardCost;
+      return remaining;
+    },
     saveData: function() {
       const result: Array<Array<string>> = [];
       result.push([]);
@@ -70,10 +80,11 @@ export const SelectInitialCards = Vue.component('select-initial-cards', {
       this.selectedPrelude = cards;
     },
   },
-  template: `<div>
-    <select-card :player="player" :playerinput="getOption(0)" :showtitle="true" v-on:cardschanged="corporationChanged" />
+  template: `<div class="select-initial-cards">
+   <select-card :player="player" :playerinput="getOption(0)" :showtitle="true" v-on:cardschanged="corporationChanged" />
     <select-card v-if="hasPrelude()" :player="player" :playerinput="getOption(1)" :showtitle="true" v-on:cardschanged="preludesChanged" />
     <select-card :player="player" :playerinput="getOption(hasPrelude() ? 2 : 1)" :showtitle="true" v-on:cardschanged="cardsChanged" />
+    <div v-if="selectedCorporation" v-i18n>Starting Megacredits: <div class="megacredits">{{getRemainingMegacredits()}}</div></div>
     <Button v-if="showsave" :onClick="saveData" type="submit" :title="playerinput.buttonLabel" />
   </div>`,
 });
--- a/compile-vue-templates.js
+++ b/compile-vue-templates.js
@@ -197,7 +197,7 @@ checkComponent(
 checkComponent(
   'src/components/SelectInitialCards',
   require('./build/src/components/SelectInitialCards').SelectInitialCards,
-  [],
+  ['selectedCorporation'],
 );
 checkComponent(
   'src/components/SelectOption',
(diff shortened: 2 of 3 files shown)


kberg (approved):

vsrisuknimit:
Looks great!

vsrisuknimit:
@bafolts If we are not including preludes in the calculation yet but allow the number to show up when prelude expansion is used, should we have a short comment for now that it is not accounting for preludes.

Starting Megacredit: 48 (before preludes)

chosta (changes requested):
Looking good but I agree with @vsrisuknimit that preludes `+/-` MC amounts must be included in the result. I'll make it easier for you:
`Allied Banks` +3
`Aquifier Turbines` -3
`Business Empire` -6
`Donation` +21
`Galilean Mining` -5
`Huge Asteroid` -5
`Loan` +30
`Martian Industries` +6
`Nitrogen Shipment` +5

vsrisuknimit:
Also `startingMegaCredits` seems to take into account already the corp reveal effect for Vitor, Pharma U, and Splice. Do we need to tell user anything about it? Player might be confused and double count.

Perhaps a <div v-if> in template that says `corporation effect accounted` if the corp is one of those three?

vsrisuknimit:
> Looking good but I agree with @vsrisuknimit that preludes `+/-` MC amounts must be included in the result. I'll make it easier for you:
> `Allied Banks` +3
> `Aquifier Turbines` -3
> `Business Empire` -6
> `Donation` +21
> `Galilean Mining` -5
> `Huge Asteroid` -5
> `Loan` +30
> `Martian Industries` +6
> `Nitrogen Shipment` +5

I am fine to do it without prelude for this PR. But if you are going to do Prelude as well, I would show the number in two lines. 

Starting Megacredit: ? mc
After Preludes: ? mc

Technically you can start the game if the second number is negative. You just cannot play one of your preludes. 

The second number should take into account the following:
If you are Manutech, you get some cash back:
`Allied banks` give you +4 back immediately
`Business Empire` +6
`Dome Farming` +2
`Metal Company` +1
`Research Network` +1
`Self-sufficient settlement` +2
If you are Splice or Pharma U, you lose or gain more mc if you play `Ecoexperts` or `Biofuels`.
If you are Tharsis, you gain +3 mc for `Self-sufficient settlement` and `Early settlement`.

That should cover all corps and preludes interaction.

chosta:
We can merge it as it is in this case but I'm the first comment about this would be "What about preludes?" So why not give the players what they really need :)

chosta (approved):

src/components/SelectInitialCards.ts
```diff
@@ -70,10 +80,11 @@ export const SelectInitialCards = Vue.component('select-initial-cards', {
       this.selectedPrelude = cards;
     },
   },
-  template: `<div>
-    <select-card :player="player" :playerinput="getOption(0)" :showtitle="true" v-on:cardschanged="corporationChanged" />
+  template: `<div class="select-initial-cards">
+   <select-card :player="player" :playerinput="getOption(0)" :showtitle="true" v-on:cardschanged="corporationChanged" />
     <select-card v-if="hasPrelude()" :player="player" :playerinput="getOption(1)" :showtitle="true" v-on:cardschanged="preludesChanged" />
     <select-card :player="player" :playerinput="getOption(hasPrelude() ? 2 : 1)" :showtitle="true" v-on:cardschanged="cardsChanged" />
+    <div v-if="selectedCorporation" v-i18n>Starting Megacredits: <div class="megacredits">{{getRemainingMegacredits()}}</div></div>
```
vsrisuknimit, line 87: Please add `preludes not accounted` and `corporation effect accounted` for Vitor, PU and Splice.

vsrisuknimit (changes requested):

src/components/SelectInitialCards.ts
vsrisuknimit: Just to prevent any confusion. Let's work on Prelude in another PR.

bafolts: I can do preludes if you know the list of which preludes would impact this count. Easy enough to add now.

vsrisuknimit: @bafolts See the messages above from me and @chosta .

bafolts: I will push update tonight (GMT-7) that includes preludes.

src/components/SelectInitialCards.ts
```diff
@@ -39,12 +40,60 @@ export const SelectInitialCards = Vue.component('select-initial-cards', {
     };
   },
   methods: {
+    getAfterPreludes: function() {
+      let result = 0;
+      if (this.selectedPrelude.includes(CardName.ALLIED_BANKS)) {
+        result += 3;
```
bafolts, line 46: @vsrisuknimit is this list good enough to start?

We don't know which card from hand that `EcologyExperts` will trigger so we can't deduct the cost of that card.

vsrisuknimit: No need to deal with EcoExpert. It is definitely a good start but merging now will still cause confusion.

Please take into account Pharma Union and Splice with Biofuels and Ecoexperts (ignoring the card that the user will pay). Just the +2 or -4 from the microbe tag itself.

And then Manutech.

Lynesth: Wouldn't it be better to loop on `this.selectedPrelude` since there's only two of them at most and check the value instead of looping on them 10 times ?

bafolts: OK. Manutech added and I looped through `selectedPrelude`.

vsrisuknimit (approved):
Look good to me but let me test locally.

vsrisuknimit:
@bafolts Beautiful! This is a great feature. The only thing I might add is either (1) an (i) with <a href> link to a new page redirecting to the wiki about the calculation. OR (2) <span v-if="...">Accounted for corporation and prelude interaction</span> Next to the second number if we have manutech, tharsis, pharma u, or splice.

I prefer (2). But it is good to go without too.

vsrisuknimit:
You could have a vue method for that v-if. Only shows if Manutech is picked with one of the mc production prelude and so on. The switch cases you have toward the end of getAfterPreludes method.

PR #2478, opened 2021-01-19
Translate mixin

alrusdi:
Introduce mixin for template translations.
Reduce translation function imports and use TranslateMixin


chosta (approved):
Nice! Do we need to look for other instances of dynamic strings to replace?

alrusdi:
> Nice! Do we need to look for other instances of dynamic strings to replace?

I've tried to find them all, so i hope no. But it is better to use TranslateMixin in new code if v-i18n directive isn't the case

PR #2482, opened 2021-01-19
Fix wordings

### LINKED ISSUE #2357: Minor issues with card text
Enough with icons, now for text :-)

* Does it cause a problem to put the apostrophe into "CEOS FAVORITE PROJECT"?

* Search For Life: requirement text is missing

* Experimental Forest: missing "and raise oxygen 1 step"

* AI Central: "Effect:" -> "Action:"

* Research Coordination: this should be descriptive text (parenthesized mix-case) as it is on Septem Tribus

* Research Network: should also have that same descriptive text (about the wild symbol)

* Aridor: my card says "As your first action, put an additional Colony Tile of your choice into play", which I think makes a lot more sense

* Inventrix: "you" -> "your"

* Aquifer Turbines: "ocean" -> "ocean tile"

* Noctis City: "city" -> "city tile"

* Deimos Down Promo: "no other city tile" -> "no city tile"
 
* Pharmacy Union: Did we already decide to diverge from official wording, or is this revised official wording? Because corps aren't "revealed", they're "played", and "when this card is played" is implied anyway (by the fact that it *doesn't* say "as your first action"). It could leave all that out.

* Just a weird question: why do Rad-Chem and Rad-Suits appear out of order in debug-ui?

vsrisuknimit:
This PR can close #2357 .


src/CardName.ts
```diff
@@ -48,7 +48,7 @@ export enum CardName {
     CAPITAL = 'Capital',
     CARETAKER_CONTRACT = 'Caretaker Contract',
     CARTEL = 'Cartel',
-    CEOS_FAVORITE_PROJECT = 'CEOs Favorite Project',
```
kberg, line 51: Cards cannot be renamed. This will break all past games. Unfortunately this file can't be in the PR.

kberg (changes requested):

src/CardName.ts
kberg: This is tied, amusingly, to my comment wishing card numbers were our indexes.

vsrisuknimit: What is the problem with breaking all past games?

vsrisuknimit: Old games are deleted anyway. It is a demo server nothing is promised to be kept.

kberg: It breaks all games in progress.

vsrisuknimit: So? Updates did that in the past. People make new games.

kberg: Understood. I just spent a lot of December rewriting serialization in such a way that it maintained full backward compatibility over a period of time, and then removed the archiaic versions. In fact, I've held back on a turmoil backward compatibility serilaization fix because I realized when dismantling the old version after the requisite waiting period that I did it wrong, and that I'm going to need to repeat the process.

I understand that you want to rename these key values. Are you willing to let this file slip out of this PR for the time being, and I'll work on a safe channge like this?

chosta: We should never make a change that breaks existing games knowingly. That is just bad customer care :) Let's leave the game breaking file out of this PR as @kberg for now.

vsrisuknimit:
@chosta @kberg  I reverted the specific change on `CardName.ts`. The changes to the names are left on the translation files since `CEO's` is not in use anyway by `CEOs`.

BTW, Rad-chem vs Rad-Suits was why @kevinb9n noticed they appeared out of order in the debug-ui. The capitalization.

kberg:
> @chosta @kberg I reverted the specific change on `CardName.ts`. The changes to the names are left on the translation files since `CEO's` is not in use anyway by `CEOs`.
> 
> BTW, Rad-chem vs Rad-Suits was why @kevinb9n noticed they appeared out of order in the debug-ui. The capitalization.

I'd love to fix them, just ... with a little more work.

src/cards/prelude/ResearchCoordination.ts
```diff
@@ -16,9 +14,7 @@ export class ResearchCoordination extends Card implements IProjectCard {
 
       metadata: {
         cardNumber: 'P40',
-        renderData: CardRenderer.builder((b) => {
-          b.text('After being played, when you perform an action, the wild tag counts as any tag of your choice.', CardRenderItemSize.SMALL, true);
-        }),
+        description: 'After being played, when you perform an action, the wild tag counts as any tag of your choice.',
```
kberg, line 17: I'm OK with the changes overall, but wonder why this was done. (Not a wording change, but I also assume it wasn't accidental.)

vsrisuknimit: @kberg to make it consistent with other cards that describe wild tag. Check with @kevinb9n .

vsrisuknimit: Mostly that description is shown in parentheses which is appropriate for the wild tag explanation. This is already in use in Septem Tribus. So I made it the same for these two cards.

kevinb9n: Basically the all-caps sans-serif style is used for cases where there is *no* iconographic depiction for the instruction.

kberg (approved):

bafolts:
Merge this branch with `main` to pickup the workflow change that runs `build-linux` and `build-windows`.

vsrisuknimit:
@bafolts I think I did it.

bafolts:
Success! Sorry for the inconvenience during the transition. With the `build-windows` check we run the `npm run build` on a windows machine. This should help us support windows users and avoid inadvertently breaking their builds.

PR #2486, opened 2021-01-20
Follow the rulebook for neutral greenery placement

daniel-ziegler:
Instead of using RNG, discard cards to determine where neutral greeneries go.

This required updating `Board.getAdjacentSpaces` to return adjacent spaces in clockwise order, starting from the top left. I'll update the unit tests to verify that.


src/boards/Board.ts
```diff
@@ -59,16 +60,21 @@ export abstract class Board {
...
-        return space !== adj && adj.spaceType !== SpaceType.COLONY && (
-          (adj.x === leftSpace[0] && adj.y === leftSpace[1]) ||
-          (adj.x === rightSpace[0] && adj.y === rightSpace[1]) ||
-          (adj.x === topLeftSpace[0] && adj.y === topLeftSpace[1]) ||
-          (adj.x === topRightSpace[0] && adj.y === topRightSpace[1]) ||
-          (adj.x === bottomLeftSpace[0] && adj.y === bottomLeftSpace[1]) ||
-          (adj.x === bottomRightSpace[0] && adj.y === bottomRightSpace[1])
-        );
-      });
+      const coordsInOrder = [
```
kberg, line 63: For the next person reading this, I suggest dropping `coordsInOrder` to `coords` and just add a comment indicating they're in clockwise order.

// Coordinates are in clockwise order. Order only ever matters during solo game set-up when placing
// starting forests. Since that is the only case where ordering matters, it is adopted here.

src/boards/Board.ts
```diff
@@ -59,16 +60,21 @@ export abstract class Board {
...
+        topLeftSpace,
+        topRightSpace,
+        rightSpace,
+        bottomRightSpace,
+        bottomLeftSpace,
+        leftSpace,
+      ];
+      // Make lists with 0 or 1 elements for all adjacent positions (depending on
+      // whether they exist), then concatenate. (Like flatMap in ES2019.)
+      const listsToConcat = coordsInOrder.map(([x, y]) => this.spaces.filter((adj) =>
```
kberg, line 73: I don't think what you're writing is bad, though I am lost by how line 73 works. That's OK. Still, let me ask this question: what if this method stayed as is, and ordering occurred in the solo set-up? That way there wouldn't need to be special code here, which is called probably thousands of times a game, just to handle the two calls needed for set-up? If adding a method `Board.getAdjacentSpace(space:, dx, dy): ISpace | undefined` makes it easier, then I'm all for that.

Another alternative: Cache the adjacent space definitions during board construction,

```
private readonly adjacency: Map<SpaceId, Array<number>>

getAdjacentSpaces(...) {
  return adjacency.get(spaceId).map((idx) => this.spaces[idx]));
}
```

Perhaps then we get a double win. I think that's better only because it's a one-time set-up and doesn't have to be considered when reading later on.

kberg:
If you want to not knock yourself out regarding the tests, just do the ordering test in one or two places.

kberg:
By the way, thanks for doing this. I was glad just to have the card draw for the cities, and didn't think through the forests, so I'm glad to see this being done.

src/Game.ts
```diff
@@ -1485,20 +1485,25 @@ export class Game implements ISerializable<SerializedGame> {
     return this.getPlayers().some((p) => p.getProduction(resource) >= minQuantity) || this.isSoloMode();
   }
 
-  public getSpaceByOffset(direction: -1 | 1, type = 'tile') {
+  public discardForCost(toPlace: string) {
```
Lynesth, line 1488: Could this use TileType ?

src/boards/Board.ts
Lynesth: I agree that the ordering should go into the solo setup functions as we don't care about it here, and it's increasing this function's execution time by a substantial amount.

Lynesth (changes requested):

bafolts:
Also glad to see this done. The official rule book is always right.

Lynesth:
> Also glad to see this done. The official rule book is always right.

What about making cards drawn from the top of the deck then ? :)

daniel-ziegler:
Thanks for the feedback! Putting the ordering in the solo setup would make sense, but it seems fiddly to reorder after the spaces the fact and I also would rather not duplicate the coordinate manipulation code for finding adjacent spaces.

What if I kept the ordering in `getAdjacentSpaces` but set up a map at setup to make it quick to find up spaces by coordinates? That would also speed up the function for other callers.

Lynesth:
> What if I kept the ordering in `getAdjacentSpaces` but set up a map at setup to make it quick to find spaces by coordinates? That would also speed up the function for other callers.

Sure, that would work. We were actually thinking of doing this in the upcoming days. What kind of map are you thinking about ?
Something like `Map<SpaceId,Space[]>` ?

kberg:
> > What if I kept the ordering in `getAdjacentSpaces` but set up a map at setup to make it quick to find spaces by coordinates? That would also speed up the function for other callers.
> 
> Sure, that would work. We were actually thinking of doing this in the upcoming days. What kind of map are you thinking about ?
> Something like `Map<SpaceId,Space[]>` ?

Except the flavor of this project is Map<SpaceId, Array<ISpace>>.

If this is the mechanism you use rather than the index into the full array, please be sure the clone this array on every call so as to prevent someone manipulating it.

The rest of the algorithm was much more expensive than an array clone, I think we can settle on this much memory improvement.

daniel-ziegler:
It seemed to me that going from (x,y) to SpaceID was hard so I was instead going to make a nested `Map<number, Map<number, Space>>` indexed first by `x` and then by `y`. Anything wrong with that plan?

kberg:
Is it hard, or inefficient? Now that it occurs at setup or deserialization
it's possibly the second one. But you do what you think is best, I'm
confident your change won't be so far from the finished result.

Lynesth:
Oh I meant an adjacentSpaces map, that, given 1 space id would give all the spaces adjacent to it.

src/Game.ts
daniel-ziegler: Done. `getSpaceByOffset` is used in `AresHazards.ts` though so the logging behavior there has changed slightly (and I haven't tested that it looks reasonable)

daniel-ziegler: BTW is the Ares hazard placement supposed to avoid tiles next to cities? (I know nothing about Ares)

daniel-ziegler:
Okay, I switched to the plan where adjacent spaces are precomputed in the constructor and made the test check the clockwise ordering. Let me know what you think!

src/boards/Board.ts
daniel-ziegler: Done.

src/boards/Board.ts
daniel-ziegler: I went with the plan to cache adjacent space definitions.

src/boards/Board.ts
```diff
@@ -59,20 +61,42 @@ export abstract class Board {
...
+      for (const [x, y] of coords) {
+        const adj = this.spaces.find((adj) =>
+          space !== adj && adj.spaceType !== SpaceType.COLONY &&
+            adj.x === x && adj.y === y,
         );
-      });
+        if (adj !== undefined) {
+          spaces.push(adj);
+        }
+      }
```
Lynesth, line 84: I have not tested this piece of code at all, but it should probably be a bit less costly (need to benchmark it I guess):
```ts
/*
 * Returns true inside the loop when it's done searching adjacent spaces
 * Returns false inside the loop to continue to the next iteration
 * /
const spaces: Array<ISpace> = [];
this.spaces.some((adj) => {
  if (adj.spaceType === SpaceType.COLONY) {
    return false;
  }
  if (adj.y > space.y + 2) {
    return true;
  }
  if (adj.y < space.y - 2 || adj.x < space.x - 2 || adj.x > space.x + 2) {
    return false;
  }
  const index = coords.findIndex(([x, y]) => x === adj.x && y === adj.y));
  if (index === -1) {
    return false;
  }
  spaces[index] = adj;
  if (spaces.length === 6) {
    return true;
  }
  return false;
});

return spaces.flat();
```

src/boards/Board.ts
```diff
@@ -13,10 +13,12 @@ import {SerializedBoard, SerializedSpace} from './SerializedBoard';
 export abstract class Board {
   private maxX: number = 0;
   private maxY: number = 0;
+  private readonly adjacency = new Map<SpaceId, Array<ISpace>>();
   protected constructor(public spaces: Array<ISpace>) {
+    this.maxX = Math.max(...spaces.map((s) => s.x));
+    this.maxY = Math.max(...spaces.map((s) => s.y));
```
Lynesth, line 19: I don't think this is necessary to keep this.
At the moment, it is hardcoded anyway that a board is of a certain number of rows and and certain number of tiles per row `[5, 6, 7, 8, 9, 8, 7, 6, 5]` so you can safely assume that both `maxX` and `maxY` are `8`.

But let's get @kberg and/or @bafolts opinions.

Lynesth:
That's the idea, thank you. I left a couple of comments but globally it looks good to me.

src/Game.ts
```diff
@@ -1485,20 +1485,25 @@ export class Game implements ISerializable<SerializedGame> {
...
-  public getSpaceByOffset(direction: -1 | 1, type = 'tile') {
+  public discardForCost(toPlace: TileType) {
     const card = this.dealer.dealCard();
     this.dealer.discard(card);
-    this.log('Dealt and discarded ${0} (cost ${1}) to place a ${2}', (b) => b.card(card).number(card.cost).string(type));
+    this.log('Drew and discarded ${0} (cost ${1}) to place a ${2}', (b) => b.card(card).number(card.cost).tileType(toPlace));
+    return card.cost;
+  }
+
+  public getSpaceByOffset(direction: -1 | 1, toPlace: TileType) {
```
kberg, line 1495: Not related to what you did, but once this is done, I'm going to change getSpaceByOffset to take a number instead of being being responsible for dealing a card, and move it into Board where it should have been.

kberg: Don't go doing that, this is a nice simple PR for the moment. :D

src/GameSetup.ts
```diff
@@ -100,16 +100,11 @@ export class GameSetup {
...
     function getRandomForestSpace(spaces: Array<ISpace>): ISpace {
-      let idx = rng.nextInt(spaces.length);
-      for (let count = 0; count < spaces.length; count++) {
-        if (game.board.canPlaceTile(spaces[idx])) {
-          return spaces[idx];
-        }
-        idx = (idx + 1) % spaces.length;
-      }
-      throw new Error('Did not find space for forest');
+      let idx = game.discardForCost(TileType.GREENERY);
```
kberg, line 105: This function is so much less complicated than before, if you feel like it, move it all into placeCityAndForest and remove the separate function.

src/ares/AresHazards.ts
```diff
@@ -15,7 +15,7 @@ export class _AresHazardPlacement {
   }
 
   public static randomlyPlaceHazard(game: Game, tileType: TileType, direction: 1 | -1) {
-    const space = game.getSpaceByOffset(direction, 'hazard');
+    const space = game.getSpaceByOffset(direction, tileType);
```
kberg, line 18: Yes, this is fine. Good.

src/boards/Board.ts
kberg: The Moon's board is 4,5,6,5,4, which is why you can't assume size.

src/boards/Board.ts
kberg: Probably not nitpicking, but could also wait for a follow-up. Fortunately, the tests for adjacent spaces became a totally complete set, so it's easy to verify any breaks at all. :D

src/boards/Board.ts
```diff
@@ -13,10 +13,12 @@ import {SerializedBoard, SerializedSpace} from './SerializedBoard';
 export abstract class Board {
   private maxX: number = 0;
   private maxY: number = 0;
+  private readonly adjacency = new Map<SpaceId, Array<ISpace>>();
```
kberg, line 16: `adjacentSpaces`?

src/boards/Board.ts
kberg: For what it's worth, I think this new algorithm is harder to understand (I don't understand it yet) so how about making it more readable in a follow-up, and owning its correctness?

src/boards/Board.ts
```diff
@@ -59,20 +61,42 @@ export abstract class Board {
...
+        if (adj !== undefined) {
+          spaces.push(adj);
+        }
+      }
+      return spaces;
     }
     return [];
   }
 
+  // Returns adjacent spaces in clockwise order starting from the top left.
```
kberg, line 90: Put this comment up by the top: `stores adjacent spaces in clockwise order starting from the top left.`?

tests/moon/MoonBoard.spec.ts
```diff
@@ -19,7 +19,7 @@ describe('MoonBoard', function() {
 
   it('getAdjacentSpaces', () => {
     expect(board.getAdjacentSpaces(board.getSpace('m01'))).is.empty;
-    expect(board.getAdjacentSpaces(board.getSpace('m02')).map((s) => s.id)).deep.eq(['m03', 'm05', 'm06']);
+    expect(board.getAdjacentSpaces(board.getSpace('m02')).map((s) => s.id)).deep.eq(['m03', 'm06', 'm05']);
   });
 
   it('getAdjacentSpaces', () => {
```
kberg, line 25: Interesting: the test below doesn't need changing?

kberg (changes requested):
Fairly minor comments, really.

src/boards/Board.ts
daniel-ziegler: A simpler fast plan might be:
1. Build a map from (x,y) to ISpace
2. Look up each of the adjacent coords in that map
Anyway, I won't worry about any of that for this PR

tests/moon/MoonBoard.spec.ts
daniel-ziegler: It wasn't testing order so it still passed (but wasn't strict). I've updated it to test order as well.

src/boards/Board.ts
daniel-ziegler: Done

src/boards/Board.ts
daniel-ziegler: Renamed

src/GameSetup.ts
daniel-ziegler: Done

tests/moon/MoonBoard.spec.ts
kberg: Gah! Thanks.

kberg (approved):
Thank you! Not only did you resolve a game rule, you made the game faster. I couldn't imagine the millions of times that method is called.

kberg:
@Lynesth this PR is blocked on your review, so I'll leave it to you to merge when you are satisfied.

Lynesth (approved):

PR #2491, opened 2021-01-20
Fixes #2285

### LINKED ISSUE #2285: Warnings re: standard projects cards
Just played through a solo game and looked at the console. 100's of warnings show up there. I know there was recent changes with adding the SP cards and card lookup code and this likely has to do with one of those.

image

sienmich:


src/CardFinder.ts
```diff
@@ -35,50 +35,40 @@ export class CardFinder {
       return CardFinder.decks;
     }
 
-    public getStandardProjectCardByName(cardName: string): StandardProjectCard | undefined {
-      let found : (ICardFactory<StandardProjectCard> | undefined);
-      CardFinder.getDecks().some((deck) => {
-        found = deck.standardProjects.findByCardName(cardName as CardName);
-        return found !== undefined;
+    public getCardByName<T extends ICard>(cardName: string, decks: (manifest: CardManifest) => Array<Deck<T>>): T | undefined {
```
kberg, line 38: Make cardName `CardName` type here.

kberg: Unless that causes weird compiler issues for you.

src/CardFinder.ts
```diff
@@ -35,50 +35,40 @@ export class CardFinder {
...
       });
       if (found !== undefined) {
         return new found.Factory();
       }
-      console.warn(`standard project card not found ${cardName}`);
+      console.warn(`card not found ${cardName}`);
       return undefined;
     }
 
+    public getProjectOrStandardProjectCardByName(cardName: string) {
```
kberg, line 54: I think this is the wrong expression of the abstraction. This method serves only the log component, the log component should do this work.

kberg (commented):
It's a minor thing.

src/CardFinder.ts
sienmich: Agree, done

kberg (approved):

src/CardFinder.ts
sienmich: It sounds like a good idea, but this would require a lot of refactors.
I'd like to only fix the mentioned warnings in console bug in this PR, and leave cardName as string, like it was earlier in this set of functions.

kberg: Agree, that's fine.

PR #2496, opened 2021-01-20
Split takeAction() into takeAction() and getCommonActions().

sienmich:
I did two things here:

1. Split takeAction() into takeAction() and getCommonActions().
2. Change order of options to perform by a player from sorting alphabetically (quite funny that was optimal) to hard coded one.

The 1. is necessary to just read this code, takeAction() had 230 lines
The 2. is necessary for future standard actions import. I think that sorting them by name is funny ('Claim milestone' as first, 'Sell patents' at the end), but it shouldn't be here, it's hard to add new options


src/Player.ts
```diff
@@ -1763,11 +1762,33 @@ export class Player implements ISerializable<SerializedPlayer> {
       return;
     }
 
+    this.setWaitingFor(this.getCommonActions(), () => {
```
kberg, line 1765: So, I wouldn't call these the _common_ actions, it's all the actions. There are some exceptions, of course.

getActions() instead of getCommonActions()?

sienmich: Ok, done

kberg (approved):

kberg:
Thanks!

PR #2500, opened 2021-01-21
Standard actions once again

sienmich:
Hi, once again the Standard Actions - SAs. (I closed old messy https://github.com/bafolts/terraforming-mars/pull/2288 PR and opened this one, clearer imo).

In this PR I've:
1. Added the class StandardActionCard, which is meant to be a class for actions like converting plants / heat, turmoil actions, claiming milestone and similar. This time it is parallel to `StandardProjectCard`, those two do not extend each other. The most important part of it is:
  ```ts
  export abstract class StandardActionCard extends Card implements IActionCard, ICard {
  constructor(...) {
    super({
      cardType: CardType.STANDARD_ACTION,
      name: name,
      metadata: metadata,
    });
  }
    public abstract canAct(player: Player): boolean
    public abstract action(player: Player): PlayerInput | undefined
  }
  ```
2. Added first two SAs - Convert Plants and Convert Heat. I switched old code in `Player.ts` like `private convertPlantsIntoGreenery(): PlayerInput` to new using `ConvertPlants` methods.
3. Added displaying them as cards in log:
Screenshot from 2021-01-21 01-00-48

For now the SA cards are using the same css classes as standard projects, they are meant to look the same. Those classes are named after standard projects which I did not change to improve PR readability. If you think this is necessary, let me know, I'll add that.

In next PRs I want to:
1. Display them as cards in player input section (right now this part is unchanged)
2. Add more SAs

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -19,7 +19,6 @@ import {Colony} from './colonies/Colony';
 import {ISerializable} from './ISerializable';
 import {IMilestone} from './milestones/IMilestone';
 import {IProjectCard} from './cards/IProjectCard';
-import {ISpace} from './boards/ISpace';
 import {ITagCount} from './ITagCount';
 import {LogMessageDataType} from './LogMessageDataType';
 import {MiningCard} from './cards/base/MiningCard';
@@ -63,6 +62,8 @@ import {DrawCards} from './deferredActions/DrawCards';
 import {Units} from './Units';
 import {MoonExpansion} from './moon/MoonExpansion';
 import {StandardProjectCard} from './cards/StandardProjectCard';
+import {ConvertPlants} from './cards/base/standardActions/ConvertPlants';
+import {ConvertHeat} from './cards/base/standardActions/ConvertHeat';
 
 export type PlayerId = string;
 
@@ -1422,29 +1423,6 @@ export class Player implements ISerializable<SerializedPlayer> {
     return trade;
   }
 
-  private convertPlantsIntoGreenery(): PlayerInput {
-    return new SelectSpace(
-      `Convert ${this.plantsNeededForGreenery} plants into greenery`,
-      this.game.board.getAvailableSpacesForGreenery(this),
-      (space: ISpace) => {
-        this.game.addGreenery(this, space.id);
-        this.plants -= this.plantsNeededForGreenery;
-        this.game.log('${0} converted plants into a greenery', (b) => b.player(this));
-        return undefined;
-      },
-    );
-  }
-
-  private convertHeatIntoTemperature(): PlayerInput {
-    return new SelectOption(`Convert ${constants.HEAT_FOR_TEMPERATURE} heat into temperature`, 'Convert heat', () => {
-      return this.spendHeat(constants.HEAT_FOR_TEMPERATURE, () => {
-        this.game.increaseTemperature(this, 1);
-        this.game.log('${0} converted heat into temperature', (b) => b.player(this));
-        return undefined;
-      });
-    });
-  }
-
   private claimMilestone(milestone: IMilestone): SelectOption {
     return new SelectOption(milestone.name, 'Claim - ' + '('+ milestone.name + ')', () => {
       this.game.claimedMilestones.push({
@@ -1790,6 +1768,21 @@ export class Player implements ISerializable<SerializedPlayer> {
                       'available action.';
     action.buttonLabel = 'Take action';
 
+    // Convert Plants
+    const convertPlants = new ConvertPlants();
+    if (convertPlants.canAct(this)) {
+      action.options.push(convertPlants.action(this));
+    }
+
+    // Convert Heat
+    const convertHeat = new ConvertHeat();
+    if (convertHeat.canAct(this)) {
+      action.options.push(new SelectOption(`Convert ${constants.HEAT_FOR_TEMPERATURE} heat into temperature`, 'Convert heat', () => {
+        return convertHeat.action(this);
+      }));
+    }
+
+
     if (this.getPlayableCards().length > 0) {
       action.options.push(
         this.playProjectCard(),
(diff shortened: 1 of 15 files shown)


kberg:
Lucky #2500. Here I go.

src/cards/Card.ts
```diff
@@ -32,7 +31,7 @@ export abstract class Card {
       if (properties.cardType === CardType.CORPORATION && properties.startingMegaCredits === undefined) {
         throw new Error('must define startingMegaCredits for corporation cards');
       }
-      if (properties.cardType !== CardType.CORPORATION && properties.cardType !== CardType.PRELUDE && properties.cost === undefined) {
+      if ([CardType.CORPORATION, CardType.PRELUDE, CardType.STANDARD_ACTION].includes(properties.cardType) === false && properties.cost === undefined) {
```
kberg: I suggested a minor adjustment for readability. It is also a little more efficient.

src/cards/CardType.ts
```diff
@@ -1,10 +1,11 @@
 export enum CardType {
-    EVENT = 'red',
-    ACTIVE = 'blue',
-    AUTOMATED = 'green',
-    PRELUDE = 'pink',
-    CORPORATION = 'brown',
-    STANDARD_PROJECT = 'gray',
+    EVENT = 'event',
```
kberg, line 2: I do not recommend making a change like this right now. Just limit to adding standard_action, please.

src/cards/base/standardActions/ConvertHeat.ts
```diff
@@ -0,0 +1,47 @@
...
+
+
+export class ConvertHeat extends StandardActionCard {
+  constructor() {
+    super({
+      name: CardName.CONVERT_HEAT,
+      metadata: {
+        cardNumber: 'SA2',
+        renderData: CardRenderer.builder((b) =>
+          b.standardProject('Spend 8 Heat to raise temperature 1 step.', (eb) => {
```
kberg, line 17: Elswhere you used the heat_per_temperature constant (or whatever name it is)

src/cards/base/standardActions/ConvertPlants.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export class ConvertPlants extends StandardActionCard {
+  constructor() {
+    super({
+      name: CardName.CONVERT_PLANTS,
+      metadata: {
+        cardNumber: 'SA2',
+        renderData: CardRenderer.builder((b) =>
+          b.standardProject('Spend 8 Plants to place a greenery tile and raise oxygen 1 step.', (eb) => {
+            eb.plants(8).startAction.greenery().secondaryTag(AltSecondaryTag.OXYGEN);
```
kberg, line 21: Before, the user was explictly told "Spend 7 Plants" so I'm a little sad that's going to be removed.

src/Player.ts
```diff
@@ -1790,6 +1768,21 @@ export class Player implements ISerializable<SerializedPlayer> {
                       'available action.';
     action.buttonLabel = 'Take action';
 
+    // Convert Plants
+    const convertPlants = new ConvertPlants();
+    if (convertPlants.canAct(this)) {
+      action.options.push(convertPlants.action(this));
```
kberg, line 1774: Why does this line just have a call to action.options.push where the one for converting heat below pushes an option that says "Convert heat into temperature", and THEN call to action?

kberg (commented):
These are the tests I'd like to see, given how critical these are.

For ConvertHeat, a standard canPlay, and a play, and a canPlay that verifies how Reds policy impacts spending.
For ConvertPlants, a standard canPlay, and play, and another two for Ecoline. Those should be simple.

kberg:
This PR is definitely going to be much easier, and will also look great in the UI. Thank you!

kberg:
(If you want help with the tests, let me know. It genuinely gets simpler, I promise.)

src/cards/Card.ts
sienmich: Cool, thanks

src/Player.ts
sienmich: Because we want to be able to place the greenery without choosing (clicking) the action.
I could use SelectCard for this, but that adds 2 clicks for each plant conversion.
This will change when I'll add displaying cards in player input section

src/cards/base/standardActions/ConvertPlants.ts
sienmich: Right now the user still will see "Spend 7 Plants", here:
```ts
return new SelectSpace(
      `Convert ${player.plantsNeededForGreenery} plants into greenery`,
```
Just on the card it will say 8 plants.

src/cards/base/standardActions/ConvertHeat.ts
sienmich: Yep, but currently every card has action decryption in metadata hard coded. So I'll leave this.

src/cards/CardType.ts
sienmich: You mean `right now` as not this PR, or as "just don't"? I see no harm in this one, I've split it here: https://github.com/bafolts/terraforming-mars/pull/2509

kberg: > You mean `right now` as not this PR, or as "just don't"? I see no harm in this one, I've split it here: #2509

I meant what you did, so thank you for correctly inferring my intent from my confusing message.

src/cards/base/standardActions/ConvertHeat.ts
kberg: Ugh. Can you put a TODO in here? @chosta any thoughts?

src/cards/base/standardActions/ConvertPlants.ts
kberg: Oh! Then this is OK.

src/Player.ts
kberg: > Because we want to be able to place the greenery without choosing (clicking) the action.
> I could use SelectCard for this, but that adds 2 clicks for each plant conversion.
> This will change when I'll add displaying cards in player input section

OK. Thanks.

kberg (approved):

kberg:
Thank you!

PR #2501, opened 2021-01-21
Use windows build

bafolts:
Adds a windows machine to the build. This should provide some hope that windows users will be able to build as we change the build scripts.


kberg:
I can't parse this. Except it looks like this PR is all about CI?

bafolts:
Yes. With this change there is a `build-linux` and a `build-windows`. The `build-windows` runs on a windows server. This will ensure that with every change windows users can continue to build the server.

PR #2502, opened 2021-01-21
static card memory pattern for more promo cards

bafolts:
More promo cards modified to use static memory pattern for properties.

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -2142,10 +2142,7 @@ export class Player implements ISerializable<SerializedPlayer> {
         }
       }
       if (player.corporationCard instanceof PharmacyUnion) {
-        if (d.corporationCard.isDisabled === true) {
-          player.corporationCard.tags = [];
-          player.corporationCard.isDisabled = true;
-        }
+        player.corporationCard.isDisabled = Boolean(d.corporationCard.isDisabled);
       }
     } else {
       player.corporationCard = undefined;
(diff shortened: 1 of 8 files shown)


src/Player.ts
```diff
@@ -2142,10 +2142,7 @@ export class Player implements ISerializable<SerializedPlayer> {
         }
       }
       if (player.corporationCard instanceof PharmacyUnion) {
-        if (d.corporationCard.isDisabled === true) {
-          player.corporationCard.tags = [];
```
kberg, line 2146: Did you intend to remove this line?

src/cards/promo/Philares.ts
```diff
@@ -6,128 +6,132 @@ import {ISpace} from '../../boards/ISpace';
...
-    public startingMegaCredits: number = 47;
-    public cardType = CardType.CORPORATION;
+export class Philares extends Card implements CorporationCard {
+  constructor() {
+    super({
+      cardType: CardType.CORPORATION,
+      name: CardName.PHILARES,
+      tags: [Tags.BUILDING],
+      startingMegaCredits: 47,
+      initialActionText: 'Place a greenery tile and raise the oxygen 1 step',
```
kberg, line 24: Warning: initialActionText is all about rendering. It should be part of metadata, to prevent commingling..

kberg (approved):

src/Player.ts
bafolts: The `tags` property is only an accessor. While changing `PharmacyUnion` typescript wouldn't compile here. This is handled within `PharmacyUnion` now. When `this.isDisabled` is true the tags returns `[]`.

src/cards/promo/Philares.ts
bafolts: That is a good point. This happens to be used as the title for a `PlayerInput` so has been left behind. Should get moved to metadata.

PR #2504, opened 2021-01-21
Tsc

kberg:
Key changes:
1. styles.css is moved to build, so .gitignore can ignore it.
2. the two codegenned JSON files are moved to src/genfiles. the clean command and .gitignore changed appropriately.
3. Moved tsconfig.json into src, and tsconfig-test.json into tests
4. Converted both of those files so they would use project references.

One upshot of all of this is once you clone the repo you can run
```
$ npm run install
$ npm run test
```

Edited to add: this also restores test files to full participation in the tserver index.


Lynesth:
I missed a PR apparently, but why was the linter moved to the last spot ?
Can it be back to the front ?

Lynesth:
I'm not sure why you're moving settings and translations, but I think you also need to if you're doing it I think you also need to fix the routes for translations in server.ts.

src/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+    "compilerOptions": {
+        "outDir": "../build/src",
+        "module": "commonjs",
```
Lynesth, line 4: Is it needed since it's already declared in the base tsconfig ?

kberg:
> I missed a PR apparently, but why was the linter moved to the last spot ?
> Can it be back to the front ?

That was done slightly recently, because it's particularly slow -- I'd rather get compiler errors first.

kberg:
> > I missed a PR apparently, but why was the linter moved to the last spot ?
> > Can it be back to the front ?
> 
> That was done slightly recently, because it's particularly slow -- I'd rather get compiler errors first.

I'm not opposed to it, though.

kberg:
> I'm not sure why you're moving settings and translations, but if you're doing it I think you also need to fix the routes for translations in server.ts.

Yipe! This was in an earlier version I swear. I'll fix that immediately.

It's to put all the _source_ in _src_. And because the `src/tsconfig.json` must have all code in its root directory.

src/tsconfig.json
kberg: Thanks, fixed.

make_static_json.js
```diff
@@ -56,12 +58,15 @@ function getWaitingForTimeout() {
     return 5000;
 }
 
-fs.writeFileSync("assets/settings.json", JSON.stringify({
+if (!fs.existsSync('src/genfiles')) {
+    fs.mkdirSync('src/genfiles');
```
bafolts, line 62: Using these is helpful for ensuring this builds on every OS :+1:

src/components/App.ts
```diff
@@ -10,7 +10,7 @@ import {DebugUI} from './DebugUI';
 import {GameHomeModel} from '../models/GameHomeModel';
 import {HelpIconology} from './HelpIconology';
 
-import * as raw_settings from '../../assets/settings.json';
+import * as raw_settings from '../genfiles/settings.json';
```
bafolts, line 13: This is a great name change. Makes it clearer that we generate these files.

bafolts (approved):

bafolts:
I often run `npm run build` while I still have linting errors. Having the lint step last works for my workflow.

Lynesth:
Could we add --cache to the eslint command then ?

kberg:
> Could we add --cache to the eslint command then ?

Feel free! It sounds like we should have done that in the first place.

bafolts:
I may add the `--cache` for `eslint`. I work on underpowered linux machines and that could be a game changer.

PR #2505, opened 2021-01-21
Fix Minority Refuge & Pioneer Settlement

### LINKED ISSUE #2460: Minority Refuge / Pioneer Settlement should be playable when I have -4 or -5 mc production...
... as long as I can place the colony on Luna.

Lynesth:
Fixes #2460 

Could use some tests but I've been lazy this time.

Also, Pioneer Settlement somehow never checked that you had enough MC prod to play it. =/

### DIFF
--- a/src/cards/colonies/MinorityRefuge.ts
+++ b/src/cards/colonies/MinorityRefuge.ts
@@ -21,7 +21,10 @@ export class MinorityRefuge implements IProjectCard {
         return false;
       }
 
-      if (player.getProduction(Resources.MEGACREDITS) <= -4) {
+      const megaCreditsProduction = player.getProduction(Resources.MEGACREDITS);
+      if (megaCreditsProduction === -4 && player.isCorporation(CardName.POSEIDON)) {
+        return true;
+      } else if (megaCreditsProduction <= -4) {
         const lunaIsAvailable = player.game.colonies.some((colony) =>
           colony.name === ColonyName.LUNA &&
           colony.isColonyFull() === false &&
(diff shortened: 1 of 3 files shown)


sienmich:
I see 2 more edge cases:

1. `Aridor` with no Space tag,
2. `Lakefront Resort` and Europe colony (only if oceans are not maxed)

However, I'm not sure if we want to implementing them, sounds like a spaghetti to me.
(Maybe we need some automatic, reversible `canPlay` in the future :) )

Lynesth:
@sienmich Yes but:
1. For Aridor it means you are already at -4 MC and don't have a space tag, which is a probably pretty unlikely.
2. Forgot about Lakefront but that means you're already at -4 and didn't place a colony on Europa before that time.

Even the whole Luna thing I wrote in this PR is really unlikely to happen.

Regarding the reversible canPlay, I'm not for it at all, but that's another matter :)

Lynesth:
If you want to implement those 2 edge cases, feel free to do it though !
Then there's the one where you place a colony on Mercury and select to copy the trade from Europa to get 1MC prod. :D

sienmich:
Nice, I knew there has to be one more case. Ok, I'll pass that opportunity for now :)

sienmich (approved):

kevinb9n:
> Even the whole Luna thing I wrote in this PR is really unlikely to happen as you'd need to be at -4 or -5, without a colony already on Luna (which is probably one of the first you're putting something on), and with a free space on Luna.

It hit me twice in recent days because I've been on a strange quest to try to win solo with Terralabs. That means I take Loan whenever I can and boom I'm already at -4. Yet I can still be tight enough on cash and needing to funnel some prelude titanium into something useful and using one of these cards to get a colony on Luna should just leave my miserably low mc production the same.

So it is a bit of a special situation, yes.

PR #2509, opened 2021-01-21
Updated CardType enum

sienmich:
Values of this enum are unused and don't bring much new info.
I think it's better to have them the same way as in other enums, where values are lowercase of names.


sienmich:
Ok, so the #2500 merged also this one apparently, my bad.

kberg:
:D

Issue #2423, opened 2021-01-16
Agendas: proposed alternative chairperson variant


kevinb9n:
Here's an idea for a different chairperson variant for the Agendas expansion. @nwai90 

In this variant:

* When the game starts, each party has an initial ruling bonus and ruling policy, randomly chosen just like in the default variant.
* The incoming chairperson gets to change *either* the bonus or policy; the other remains the same.
* A neutral chairperson changes nothing.

We rarely have time to play over-the-table anymore so I can't say I have had a chance to test this idea out yet. But I *think* it would make for more interesting gameplay, and a *little* bit of continuity from generation to generation. It also preserves the quality, shared by all official expansions, that there is no in-game randomness except for the card decks.

nwai90:
Sounds good @kevinb9n, I like the idea of the new chairman having limited influence

nwai90:
Closed as already implemented

PR #2517, opened 2021-01-22
Update chairman mode for Agendas

nwai90:
Closes https://github.com/bafolts/terraforming-mars/issues/2423.

Main change:
- If AgendaStyle is CHAIRMAN and a player is the new chairman, either the ruling policy or ruling bonus can be changed, but not both
- If AgendaStyle is CHAIRMAN and a neutral delegate is the new chairman, nothing changes
- If AgendaStyle is STANDARD or RANDOM, get the predefined agenda from staticAgendas (i.e. no change)

Screenshot 2021-01-22 at 9 11 20 AM


kberg:
In the visual example above, what actually _is_ the ruling policy?

nwai90:
> In the visual example above, what actually _is_ the ruling policy?

@kberg The existing ruling policy and bonus are displayed on the Turmoil board. It re-renders dynamically depending on what was changed by the player

src/turmoil/PoliticalAgendas.ts
```diff
@@ -21,48 +20,41 @@ export interface Agenda {
 export interface PoliticalAgendasData {
   // The agenda for this generation.
   currentAgenda: Agenda;
-  // Standard and Random agendas are static through the game. CHAIRMAN changes
-  // by the will of the chairperson, so when staticAgendas is undefined, that means
-  // it's the CHAIRMAN agenda style
   staticAgendas: Map<PartyName, Agenda> | undefined;
```
kberg, line 23: If these are no longer static, then rename this. Can be a follow-up since it will meaningfull complicate readability of this PR.

src/turmoil/PoliticalAgendas.ts
```diff
@@ -133,11 +118,13 @@ export class PoliticalAgendas {
       return {
         currentAgenda: d.currentAgenda,
         staticAgendas: new Map(d.staticAgendas),
+        agendaStyle: d.agendaStyle,
```
kberg, line 121: This here is the problem: all of these changes are going to break current games. Their chairman's choice behavior will change mid-game -- that is --- if the game can be deserialized.

One alternative is to introduce a fourth option, making the thrd option invisible in the New Game UI and eventually removing it. (Though it makes past games not cloneable -- but long-term historically preserving cloning has never been a priority in this project.)

Another option is to think through the serialization problem more carfully, which I leave up to you to consider.

kberg (changes requested):

src/turmoil/PoliticalAgendas.ts
nwai90: So here's the thing: Current code lets chairman change both ruling bonus and policy. New code updates `ChoosePoliticalAgendas` to allow chairman to change only one of them.

We could in theory add a 4th value to `AgendaStyle` enum e.g. `CHAIRMAN_LIMITED` to reflect the new behaviour, but it's ugly to do that. It also means maintaining two versions of logic for `ChoosePoliticalAgendas` and having to deprecate the old one at some point when doing the migration.

About the concern of breaking existing games due to a new property being added to `deserialize` method, a quick and dirty fix could be to do `agendaStyle: d.agendaStyle === undefined ? AgendaStyle.STANDARD : d.agendaStyle`. Chairman behaviour for ongoing `AgendaStyle.CHAIRMAN` games would change in subsequent generations, which is fine imo.

kberg: > About the concern of breaking existing games due to a new property being added to `deserialize` method, a quick and dirty fix could be to do `agendaStyle: d.agendaStyle === undefined ? AgendaStyle.STANDARD : d.agendaStyle`. Chairman behaviour for ongoing `AgendaStyle.CHAIRMAN` games would change in subsequent generations, which is fine imo.

staticAgendas are serialized differently, too. How do you take that into account?

kberg: And are you saying, then, that active games will have the surprise that the chairman is suddenly half-powered? Which is what happen when this fix gets pushed to production. A fourth option is ugly, but that's the cost of repairing a moving train.

kberg: @nwai90 checking in -- is this stuck? Let me know.

nwai90: Yup @kberg

kberg: So, you have a choice: either break the prior games or keep them working. You've added a new serialized field `agendaStyle` on line 30. All historical games will not have this value. So, you can't just add it as

`agendaStyle: AgendaStyle;`

but have to make it optional to support loading historical games,

`agendaStyle: AgendaStyle | undefined`

Now you can know that all deserialized games without it are old ones. But you still don't know when reading it which agenda style is in place, particularly as you have reused existing fields.

So consider making this entire change in a couple of steps.

1. Add the agendaStyle field, make it optional.  Do nothing with it other than serialize and deserialize it.
2. Add a new map that is just like staticAgendas, but a new field. let's call it `agendas`. On newInstance, initialize it with the appropriate first agenda/policy for each part. When agenda type is CHAIRMAN, keep the map up to date. But never use it.
2a. `agendas` also must be `|undefined` in its definnition in `SerializedPoliticalAgendas`, because otherwise it breaks historical games. But keep deserializing it, and have some tests.
3. Wait some period of time. 10 days? All historical games will have `agendas` populated. And then you can include the new behavior for the chairman's agenda, and start ignoring `staticAgendas`.
4. At some point in the future (not long, actually, but in a separate PR, perhaps) remove `staticAgendas`.

What do you think?

kberg (approved):

kberg (changes requested):
Sorry, I'm trying to set the review to request changes instead of approve.

kberg:
In chatting with @nwai90 I went to check the database in the demo server to see how many active games use chairman mode. The answer ise 550.

```
psql postgresql://... -c "select game_id, game::json->'gameOptions'->'politicalAgendasExtension' from games where status='running'" > /tmp/result.sql

      1 (163879 rows)
    550  "Chairman"
      1   ?column?  
     27  "Random"
   7788  "Standard"
```

So just overwriting seems a bad idea.

kberg:
@nwai90 this has been settled in the main repo, yes? Close?

nwai90:
> @nwai90 this has been settled in the main repo, yes? Close?

Yes this PR is outdated, will close. The change is already done and merged in a newer PR

PR #2518, opened 2021-01-22
Make Clone error checking compatible with sqlite errors.

kberg:


daniel-ziegler:
Whoops, totally missed this when I wrote #2535! IMO that's a more complete fix.

src/database/Cloner.ts
```diff
@@ -16,7 +16,7 @@ export class Cloner {
     const response: {err: any, game: Game | undefined} = {err: err, game: undefined};
 
     try {
-      if (err === undefined && serialized !== undefined) {
```
daniel-ziegler, line 19: Why not change `SQLite.ts` to turn `null` into `undefined` instead? That way all the databases are consistent, and the comment at the top of `IDatabase` stays accurate: https://github.com/bafolts/terraforming-mars/blob/c766126a9366bcf84e866244ad0e9a480aa152e1/src/database/IDatabase.ts#L28-L29

kberg: I'm matching the original behvaior.

Lynesth (approved):

bafolts (approved):

PR #2522, opened 2021-01-23
Fix not paying enough

### LINKED ISSUE #2483: Bug: selectHowToPay is broken
image

- The amount of metal is not adjusted to max possible by default as it used to be.
- Can't pay for a card with zero cash even with enough Ti.

@kberg ?

vsrisuknimit:
Pay enough mc by default when playing a card. 

This is done by trying to overspend steel and Ti if possible. Then if we are overspending the mc at the end, spend less mc if possible.

This can close #2483 .


kberg (approved):
Interesting. If this is the case then the last parameter to deductUnits can be removed. But that also means I didn't fully understand the weird math the original algorithm was doing.

Either way, thanks for the fix.

PR #2525, opened 2021-01-23
More test and better edge case handling by default payment

vsrisuknimit:
In this case, it should spend 6 steel + 5 Ti. Not 7 steel + 5 Ti.

image


src/components/SelectHowToPayForProjectCard.ts
```diff
@@ -166,9 +166,22 @@ export const SelectHowToPayForProjectCard = Vue.component('select-how-to-pay-for
...
+        // Try to spend less mc if possible
+        if (this.megaCredits + megacreditBalance >= 0) {
+          this.megaCredits += megacreditBalance;
+        } else {
+        // If not, try to spend less steel if possible
+          // Overspend with little mc as possible
+          megacreditBalance += this.megaCredits;
+          this.megaCredits = 0;
+          // Then try to reduce the amount of steel if possible
+          while (megacreditBalance + this.player.steelValue <= 0 && this.steel > 0) {
```
kberg, line 180: This assumes you can always spend steel.

So what was wrong with the original choice, exactly? It spent less steel than it should have when it was available? There might be better ways to fix this. Let me look.

vsrisuknimit: @kberg The while loop only activates if this.steel >0 . Otherwise it still tries to minimize the amount of mc that the player is overspending.

vsrisuknimit: Nothing wrong with the previous PR. It just does not handle some edge cases well.

It tries to spend all the mc, all the steel and then all the Ti. It is likely that it will overspend too much in Ti since Ti value ranges from 3 to 7. The new PR tries to help soften the overspending.

Lynesth: Should probably do the same with microbes/floaters ?
- Paying for Venus Soil with microbes and floaters.
- Paying for Protected Valley with microbes and steel.
- Paying for Giant Solar Shade with floaters and titanium.
- And so on.

vsrisuknimit: Add a function to save resources and add more tests.

kberg: Thinking about this some more: The one thing I want to be careful about here is that there WAS a verison that worked and I broke it. 

So, what if instead, we reverted to the original algorithm, and kept all your tests?

Then I could go back and add the features I wanted, but have stronger guard rails.

What do you think?

vsrisuknimit: I think your new code is clean. The tests may not work with the old code as I am not sure of their expected behaviors. I think it is fine to have your code, my reducePayment section, and tests.

src/components/SelectHowToPayForProjectCard.ts
```diff
@@ -166,9 +181,13 @@ export const SelectHowToPayForProjectCard = Vue.component('select-how-to-pay-for
         this.heat = deductUnits(this.player.heat, 1);
       }
 
-      // If we are overspending in mc, spend less mc.
-      if (megacreditBalance < 0 && this.megaCredits + megacreditBalance >= 0) {
-        this.megaCredits += megacreditBalance;
+      // If we are overspending
+      if (megacreditBalance < 0) {
+        // Try to spend less resource if possible
+        this.steel -= saveOverSpendingUnits(this.steel, this.player.steelValue);
```
kberg, line 187: Titanium is missing from here,possibly heat?

kberg (commented):
I think this is good. A minor comment.

src/components/SelectHowToPayForProjectCard.ts
vsrisuknimit: No need. Titanium is paid last. We should never have to reduce the value. It is often overspent due to Ti.

vsrisuknimit: Heat is value at 1. And paid at the very very end. We would never overspend in heat.

kberg (commented):
I made an edit to some of the comments so they were more clear. There might be more places, they might even be tests I wrote that should also be updated.

src/components/SelectHowToPayForProjectCard.ts
kberg: Can you add a commenta about titanium? Merge when you're ready (or tell me.) Good work!

kberg (approved):

kberg (approved):

kberg:
Nice work.

PR #2527, opened 2021-01-23
Change DeferredActionsQueue to a Min-Heap

Lynesth:
A few things to note:
It doesn't change behavior at the moment but will permit better granularity in the future.
The Priority enum will change in an upcoming PR. Right now it's just like this to keep the cards behavior.

It's easier to read and understand this PR commit by commit.


kberg (approved):

kberg (commented):
Wait, wrong, wait.

src/cards/venusNext/SponsoredAcademies.ts
```diff
@@ -32,7 +32,7 @@ export class SponsoredAcademies extends Card {
   }
 
   public play(player: Player, game: Game) {
-    game.defer(new DiscardCards(player));
+    game.defer(new DiscardCards(player), -1);
```
kberg, line 35: Please give a value to this priority in the enum.

kberg (approved):
Having reviewed in commit order, you're right, it's pretty clear. One change requested. Nice work!

@bafolts may have thoughts about the new node package, but I'd like to use it and abandon my hand-built multiset.

src/cards/venusNext/SponsoredAcademies.ts
Lynesth: It's done.
Names in the enum are not final at all, it was just to keep the current behavior, they will be reworked in an upcoming PR.

bafolts:
I am cautious of taking on new dependencies. As the package only has one dependent and seems it is written with good goals in mind it seems fine. This will not impact the client side build size as it only runs on the server. I would replace the multiset we have once taking on this dependency. If we only want to use the heap there is also a stand alone priority queue with 0 dependents and more weekly downloads.

https://www.npmjs.com/package/tinyqueue

bafolts (approved):

Lynesth:
I didn't go with TinyQueue as it seemed much less performant based on a benchmark I unfortunatly cannot find anymore.

kberg:
> I am cautious of taking on new dependencies. As the package only has one dependent and seems it is written with good goals in mind it seems fine. This will not impact the client side build size as it only runs on the server. I would replace the multiset we have once taking on this dependency. If we only want to use the heap there is also a stand alone priority queue with 0 dependents and more weekly downloads.
> 
> https://www.npmjs.com/package/tinyqueue

Don't be so certain. Deferred actions code absolutely appears in the browser. That's why I was saying we should focus sometime on extracting what code is common between client and server.

PR #2533, opened 2021-01-23
player.canAfford reworked

sienmich:


src/Player.ts
```diff
@@ -1578,62 +1578,47 @@ export class Player implements ISerializable<SerializedPlayer> {
...
 
-      // Set aside reserve units in case the card has a space tag.
-      // If there isn't enough titanium to meet the purchase reserve, this isn't a playable card.
-      titanium = titanium - reserveUnits.titanium;
-      if (titanium < 0) {
-        return false;
-      }
-    }
+  public canAffordCard(card: IProjectCard, reserveUnits : Partial<Units> = Units.EMPTY): boolean {
+    return this.canAffordOptions(this.getCardCost(card), {
```
kberg, line 1585: What does canAffordOptions mean?

src/Units.ts
```diff
@@ -71,6 +71,17 @@ export namespace Units {
     };
   }
 
+  export function sum(a: Units, b: Units): Units {
```
kberg, line 74: These methods have a verb for their names, so "add"?

src/Player.ts
```diff
@@ -1578,62 +1578,47 @@ export class Player implements ISerializable<SerializedPlayer> {
...
-    const dirigibles = this.playedCards.find(
-      (playedCard) => playedCard.name === CardName.DIRIGIBLES);
-
-    if (dirigibles !== undefined &&
-        dirigibles.resourceCount &&
-        card.tags.indexOf(Tags.VENUS) !== -1) {
-      maxPay += dirigibles.resourceCount * 3;
-    }
+    // Return those removed units:
+    Units.deductUnits(Units.negative(reserveUnits), this);
```
kberg, line 1616: I'd like to avoid deducting negative units to result in positive units.

src/Player.ts
```diff
@@ -1578,62 +1578,47 @@ export class Player implements ISerializable<SerializedPlayer> {
...
+  }) {
+    const reserveUnits = Units.of(options?.reserveUnits || {});
+    if (!Units.hasUnits(reserveUnits, this)) {
+      return false;
     }
 
-    const psychrophiles = this.playedCards.find(
-      (playedCard) => playedCard.name === CardName.PSYCHROPHILES);
+    // Remove reserveUnits for a moment, to check if we can still afford the card:
+    Units.deductUnits(reserveUnits, this);
```
kberg, line 1607: What I'd like to head towards at some point is that a player's units are a Units object themselves, which would make this a lot easier. And once you start doing this kind of pre-emptive check, we'll need lots of tests to ensure Player's units are untouched when canPlay is false. So, while this is convenient, it's asking for subtle bugs.

kberg (commented):
What I'm not clear about with this PR is what your goal is. canAfford should eventually get the parameterization treatment, for sure, but I'm not sure what you're aining for  here.

sienmich:
@kberg, in fact there was too many things being changed at once. So I changed that, now there are no new functions or changed signatures, I just simplified their meat - counting maximal affordable cost.

src/Player.ts
```diff
@@ -1604,54 +1605,22 @@ export class Player implements ISerializable<SerializedPlayer> {
...
-    maxPay += this.spendableMegacredits();
-    return maxPay >= this.getCardCost(card) &&
-                (card.canPlay === undefined || card.canPlay(this, this.game));
+    return canAfford && (card.canPlay === undefined || card.canPlay(this, this.game));
   }
 
   public canAfford(cost: number, canUseSteel: boolean = false, canUseTitanium: boolean = false, canUseFloaters: boolean = false, canUseMicrobes : boolean = false): boolean {
-    let extraResource: number = 0;
-    if (canUseFloaters !== undefined && canUseFloaters) {
-      extraResource += this.getFloatersCanSpend() * 3;
```
kberg, line 1640: How have you verified removing this block doesn't cause breaks?

kberg (commented):
This is definitely much simpler!

src/Player.ts
sienmich: Are you asking why is this working or are you asking to include new tests? :)

I didn't remove this block, I just included it in the sum below (`(canUseFloaters ? this.getFloatersCanSpend() * 3 : 0) +
`)

kberg: I'm asking the former. How do you know it still works. But I'll look, thanks.

kberg (approved):
This seems okay to me. Adding a test to Diriglbles.spec.ts that matches Psychophiles.spec.ts would close the loop, I think. But OK for now.

sienmich:
Ok, I'll add `Diriglbles.spec.ts` with next changes

PR #2534, opened 2021-01-23
use dialog element instead of window.confirm

bafolts:
This should work to replace `window.confirm` with a dialog element.

image

Moved more logic into `methods` from the `mounted` function as well. Eventually we could probably work this all onto the board component.

### DIFF
--- a/src/components/SelectSpace.ts
+++ b/src/components/SelectSpace.ts
@@ -1,9 +1,8 @@
 import Vue from 'vue';
+import {ConfirmDialog} from './common/ConfirmDialog';
 import {PlayerInputModel} from '../models/PlayerInputModel';
-import {TranslateMixin} from './TranslateMixin';
 import {PreferencesManager} from './PreferencesManager';
-
-const dialogPolyfill = require('dialog-polyfill');
+import {TranslateMixin} from './TranslateMixin';
 
 export const SelectSpace = Vue.component('select-space', {
   props: {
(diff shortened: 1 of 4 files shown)


src/components/SelectSpace.ts
```diff
@@ -21,111 +22,114 @@ export const SelectSpace = Vue.component('select-space', {
...
+        return board.getElementsByClassName('board-space-selectable');
+      }
+      throw new Error('main board not found!');
+    },
+    onTileSelected: function(tile: HTMLElement) {
+      this.selectedTile = tile;
+      this.disableAvailableSpaceAnimation();
+      this.animateSpace(tile, true);
+      tile.classList.remove('board-space--available');
       const hideTileConfirmation = PreferencesManager.loadValue('hide_tile_confirmation') === '1';
```
kberg, line 86: How much extra work would it take to include the "hide" checkbox I prpopsed in the bug? That's the real goal.

bafolts: For you? No extra work at all! I will add it. It isn't work when you enjoy it so technically this is no work at all.

bafolts: @kberg I created a `ConfirmDialog` component for re-use. Attached new screenshot. The "hide" checkbox is now there.

src/components/SelectSpace.ts
```diff
@@ -21,111 +21,111 @@ export const SelectSpace = Vue.component('select-space', {
...
       if (hideTileConfirmation) {
-        return true;
+        this.confirmPlacement();
+      } else {
+        (this.$refs['confirmation'] as any).show();
```
bafolts, line 95: I want to remove this `any`. Meant to do this before I pushed.

kberg: Well I sure won't stop you.

kberg (approved):
Looks great upon visual inspection. I might play with it beforehand, but it hits the right notes.

PR #2535, opened 2021-01-23
Fix loading games from SQLite

daniel-ziegler:
The sqlite3 bindings return `err = null` when there's no error, but `Cloner.ts` checked `err === undefined`, so it thought there was an error.

This made it past the typechecker due to using `any` all over the place, which I have replaced with `Error | undefined` (or `Error | null` when using the sqlite3 bindings).

### DIFF
--- a/src/database/Cloner.ts
+++ b/src/database/Cloner.ts
@@ -10,10 +10,10 @@ export class Cloner {
     newGameId: GameId,
     players: Array<Player>,
     firstPlayerIndex: number,
-    err: any,
+    err: Error | undefined,
     serialized: SerializedGame | undefined,
     cb: DbLoadCallback<Game>) {
-    const response: {err: any, game: Game | undefined} = {err: err, game: undefined};
+    const response: {err?: Error, game: Game | undefined} = {err: err, game: undefined};
 
     try {
       if (err === undefined && serialized !== undefined) {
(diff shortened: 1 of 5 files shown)


kberg (commented):
So this is not just fixing games loaded from Sqlite, but changing the behavior of all the database implementations. I'm just going to ask that this one be for now, and https://github.com/bafolts/terraforming-mars/pull/2518 get approved at least to restore the broken behavior.

I do not have the postgresql knowledge to be able to approve this otherwise.

daniel-ziegler:
Well, the changes are almost all type annotations, so they can't change the behavior, only whether it compiles or not. The only behavior changes are a few null -> undefined conversions (with `?? undefined`) in `SQLite.ts` and turning a string into an actual `Error` in `LocalFilesystem.ts`. So this seems pretty safe to me.

That said, fixing the bug separately from improving the types seems reasonable; happy to update this after #2518 gets merged.

bafolts:
This change makes sense to me. As we have `@types/pg` and `@types/sqlite3` this change must be introducing the right type or typescript wouldn't compile.

bafolts:
@daniel-ziegler this has passing build. If you mark ready for review I will approve and merge. I used `any` when first writing database components as I wasn't sure what the type used in the callback would be. Since we have the types for pg and sqlite these must be the right types so this is a great change. Also this is the first time I am seeing `??`. Javascript is growing up!

kberg:
@daniel-ziegler I'm good if @bafolts is too. Plus now that I'm reading the postgresql innards, I can see why that works.

kberg (commented):
I don't know how to remove myself as a reviewer here -- leaving it up to Brian.

bafolts (approved):

PR #2543, opened 2021-01-24
Remove initial draft from playermodel, fix initial draft label

### LINKED ISSUE #2542: Minor bug: initial draft in gameOptions became false after end game.
Initial draft shows up correctly at the beginning of the game

image

But once the games ended, gameOptions.initialDraftVariant seems to be false.

image

image

vsrisuknimit:
I hope for this PR to close #2542 .

But so far I only removed initialDraft from playerModel since we already have initialdraftvariant in gameoptionsmodel.

I am still not sure what is causing the problem.

### DIFF
--- a/src/components/PlayerHome.ts
+++ b/src/components/PlayerHome.ts
@@ -330,15 +330,15 @@ export const PlayerHome = Vue.component('player-home', {
 
             <div class="player_home_block player_home_block--setup nofloat"  v-if="!player.corporationCard">
 
-                <div v-for="card in player.dealtCorporationCards" :key="card.name" class="cardbox" v-if="player.initialDraft">
+                <div v-for="card in player.dealtCorporationCards" :key="card.name" class="cardbox" v-if="player.gameOptions.initialDraftVariant">
                     <Card :card="card"/>
                 </div>
 
-                <div v-for="card in player.dealtPreludeCards" :key="card.name" class="cardbox" v-if="player.initialDraft">
+                <div v-for="card in player.dealtPreludeCards" :key="card.name" class="cardbox" v-if="player.gameOptions.initialDraftVariant">
                     <Card :card="card"/>
                 </div>
 
-                <div v-for="card in player.dealtProjectCards" :key="card.name" class="cardbox" v-if="player.initialDraft">
+                <div v-for="card in player.dealtProjectCards" :key="card.name" class="cardbox" v-if="player.gameOptions.initialDraftVariant">
                     <Card :card="card"/>
                 </div>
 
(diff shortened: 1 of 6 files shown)


vsrisuknimit:
There is something confusing between GameOptions (in Game.ts) and GameOptionsModel (in GameOptionsModel.ts).

vsrisuknimit:
Tagging @kberg in case he might know.

bafolts:
I believe the problem is here.

https://github.com/bafolts/terraforming-mars/blob/main/src/Game.ts#L880

For some reason we set `initialDraftVariant` to `false` here.

vsrisuknimit:
@bafolts good find! Thank you. I will take a look to see if we need it and remove it if not.

vsrisuknimit:
It is used here. 

https://github.com/bafolts/terraforming-mars/blob/04e9fdf5407464288a61530ffd984979a4d1a602/src/components/PlayerHome.ts#L333-L350

If we remove, the initial cards do go away when we are done with initial drafting.

image

bafolts:
When we are done with `Phase.DRAFTING` I would set the phase to `Phase.RESEARCH`. The checks in `PlayerHome` should then use `v-if="player.phase === getInitialPhase() && player.initialDraft"`.

I would set the phase to `Phase.RESEARCH` at the top of `gotoInitialResearchPhase`.

https://github.com/bafolts/terraforming-mars/blob/main/src/Game.ts#L669

vsrisuknimit:
Summary:
- I removed initialDraft from playerModel since we already have initialdraftvariant in gameoptionsmodel.
- I added a new Phase.InitialDraft. This is used to check whether to display the drafted cards during initial draft.
- I removed the manipulation of gameOptions.initialDraftVariant from game. Those values should be constant. It was previously used to determine whether the initial draft is done. So I rewrote that part a little.

src/components/PlayerHome.ts
```diff
@@ -165,6 +166,12 @@ export const PlayerHome = Vue.component('player-home', {
     isEventCardShown(): boolean {
       return this.hide_event_cards !== '1';
     },
+    isInitialDraftingPhase(): boolean {
+      console.log(this.player.phase);
+      console.log(this.player.generation);
+      console.log(this.player.gameOptions.initialDraftVariant);
```
bafolts, line 172: I would remove the `console.log` as they were probably used for debugging. Is the check that it is the first generation necessary? As long as `phase === Phase.INITIALDRAFTING` that should work fine.

vsrisuknimit: My bad. I added it after the test failed.

vsrisuknimit: Fixed. Removed gen 1 check. I was trying to be cautious.

bafolts (approved):

PR #2544, opened 2021-01-24
Remove setPlayerProductionForTest, and replace it with TestPlayer that subclasses Player.

kberg:


bafolts (approved):
:clap: :clap: :clap: :shipit:

PR #2550, opened 2021-01-24
Add a few bits of code used in moon13 to reduce its overall size.

kberg:


src/cards/moon/MoonCard.ts
```diff
@@ -7,6 +7,7 @@ import {IMoonCard} from './IMoonCard';
 
 export interface MoonCardProperties {
   reserveUnits?: Units,
+  // Really only interested in MOON_COLONY, MOON_MINE and MOON_ROAD.
   tilesBuilt?: Array<TileType>
```
bafolts, line 11: Typescript is pretty great. I think it can enforce the intentions here.

```
tilesBuilt?: Array<TileType.MOON_COLONY | TileType.MOON_MINE | TileType.MOON_ROAD>
```

bafolts (approved):

src/cards/moon/MoonCard.ts
kberg: > Typescript is pretty great. I think it can enforce the intentions here.
> 
> ```
> tilesBuilt?: Array<TileType.MOON_COLONY | TileType.MOON_MINE | TileType.MOON_ROAD>
> ```

In the future feel free to make an edit to that effect. Because that's fantastic. I'll do that now.

PR #2555, opened 2021-01-25
Log shuffling

Lynesth:


src/Dealer.ts
```diff
@@ -49,6 +49,7 @@ export class Dealer implements ISerializable<SerializedDealer> {
       }
 
       if (this.deck.length === 0) {
+        game.log('The discard has been shuffled to form a new deck.');
```
kberg, line 52: "The discard pile"

src/Dealer.ts
```diff
@@ -36,7 +36,7 @@ export class Dealer implements ISerializable<SerializedDealer> {
     public discard(card: IProjectCard): void {
       this.discarded.push(card);
     }
-    public dealCard(isResearchPhase: boolean = false): IProjectCard {
+    public dealCard(game: Game, isResearchPhase: boolean = false): IProjectCard {
```
kberg, line 39: I hate that just to log you have to pass the whole Game object in. I wish that Logger was a first-class object.

In the meantime, I also think what would be better would be for Game to be supplied to Dealer on construction ratrher than passing it around when needed, even if it was a logger. But, your call.

kberg (approved):

src/Dealer.ts
Lynesth: I agree, but not the scope of this small PR.

PR #2557, opened 2021-01-25
Adding disount to tags

chosta:
Continuing #2150

(1) Discount for Valley trust
valley
(2) Discount for Valley trust + Earth catapult
valley+ec

The reason this is labelled `do not merge` is because I was not sure how to expose the player discounts. I've considered several different approaches and would like to have an agreement to what we would like to use before commiting to it. 
1) Current approach.
 Just a map of all known discounts on cards in a map `DISCOUNTS_MAP`. The benefit is that it works immediatelly. Drawbacks - not pretty, not extendable (or extendable to an extent)
2) Adding a `discounts` static property to `Card.ts`.
(I've done it for the sake of review but the changes could fly if we don't choose this approach). If the `discounts` property is used, every card that has a discount would have an object as propery describing the tag and the amount for the discount. This will probably require a `getCardDiscount` to be refactored. Benefits - central control of discounts, pretty and extendable, no backend. Drawback - we need to use cardFinder.getCardByName (looping throuhg the player.playedCards array) in the frontend and potentially add unnecessary computations.
3) Adding a `discounts` property in the playerModel. 
Benefits - no unnecessary computations. Drawback - needs to be coded from scratch :)


kberg:
I would not include Earth Catapult. Global discounts should be made obvious elsewhere. As, I think should any tag discounts. Making discounts visible on cards does make more sense.

sienmich:
Maybe we could go for sth similar to metal bonus indicators, like this (I messed up the transparency):
Screenshot from 2021-01-25 15-58-25
It could save some space (although I'm not sure if it is necessary). What do you think?

sienmich:
> I would not include Earth Catapult. Global discounts should be made obvious elsewhere. As, I think should any tag discounts. Making discounts visible on cards does make more sense.

We could consider placing global discounts next to the wild tag, but maybe that's not that intuitive

assets/settings.json
```diff
@@ -0,0 +1 @@
+{"version":"701d717e Thu, 21 Jan 2021 07:42:36 -0700","waitingForTimeout":5000}
```
sienmich, line 1: Is this supposed to be here?

chosta: No, it's just something that was left behind in between commits. Just ignore it for the time being.

chosta:
> I would not include Earth Catapult. Global discounts should be made obvious elsewhere. As, I think should any tag discounts. Making discounts visible on cards does make more sense.

The reason we show this on the player board is for other players to be able to see what discounts you have. Imagine a 4p game that spans for 4 hours. Sometimes you skip a log entry that might be really important (like EC). This is much needed as well as what you suggest. Discount on cards when you play them should be a separate issue.

chosta:
> It could save some space (although I'm not sure if it is necessary). What do you think?

Space was supposed to created for discounts and player + corp name. One of the reasons for this rework was not having the space for discounts. I've considered your proposal and to be fair I even like it more that the current solution. I think we can fill the space above with identifiers of cards that are crucial to the game. I was thinking icons of penguins, ants, cutting edge (we can get creative).

chosta:
I also need to express my opinion on the way we work on issues. I've taken this one because of the great approval rating it got ( 6 rockets!). Now I see suggestions to completely change it. Should we have a better process in terms of deciding what and how to do it? I didn't 100% agree with the design but did it as it was given nevertheless because it got so much <3. As you know frontend work could be very time-consuming and this took me a while to make every pixel perfectly aligned..I'm not complaining, just suggesting to maybe find a way to find an agreemend between devs before we take on a task. Or we just roll with it and take the slack if any :)
This is just food for thought related to the future of this project. I'll focus on this task and will probably present a slightly different design approach, close to what @sienmich suggested because I also think it's simply better. Let me know what you think about the way discounts are exposed in this PR. Thanks.

sienmich:
Sorry for that, I understand it must be frustrating. At the time you proposed those changes they seemed optimal (rocket worthy!). In the meantime those metal indicators were released and after playing with them I think the small MC icon is very nice and fits here as well. It's just a thought / suggestion tho, feel free to go with current layout.

I think we need those discount counters, a lot of players were asking for them. And I agree, that discounts on cards are not working here, because we want to see other players discounts.
Thanks for all the front!

kevinb9n:
> I also need to express my opinion on the way we work on issues. I've taken this one because of the great approval rating it got ( 6 rockets!). Now I see suggestions to completely change it. Should we have a better process in terms of deciding what and how to do it?

Yes :-)

I wrote up my thoughts on this design here and in the following comment: https://github.com/bafolts/terraforming-mars/issues/2150#issuecomment-748725660

It doesn't work to show Earth Catapult as applying to every tag, because a per-card discount is different from a per-tag discount (where the discount applying to each tag on the card is added together). I think it works fine to show it above the "# cards in hand" indicator, which is even right nearby.

kevinb9n:
> Maybe we could go for sth similar to metal bonus indicators, like this (I messed up the transparency):
> Screenshot from 2021-01-25 15-58-25
> It could save some space (although I'm not sure if it is necessary). What do you think?

(For whatever it's worth, I still like the original proposal.)

nwai90:
@chosta I hear you. I think there is a bloat of issues (177!) such that it becomes hard for people to keep track of everything. This particular issue is one that has not been discussed in great detail. I'm guessing @kberg has the same concern about global discounts - that displaying the same -2 n times for n tags is repetitive UI. It may be possible to come up with something better but this is probably fine too

chosta:
> @chosta I hear you. I think there is a bloat of issues (177!) such that it becomes hard for people to keep track of everything. This particular issue is one that has not been discussed in great detail. I'm guessing @kberg has the same concern about global discounts - that displaying the same -2 n times for n tags is repetitive UI. It may be possible to come up with something better but this is probably fine too

I think showing the discount to you when playing the card is a separate issue. The player board should show what others have at a glance. We are discussing in discord, making these discounts optional, so I think it's going in the right direction.

This, I think is better as a design. We can add icons for key cards on the top. Just throwing that out there:
penguins (1)

chosta:
Actually there was a cool suggestion about the global discounts. Earth Catapult, Sky Docks, Anti-Grav and Research Outpost could be added to the card symbol to the left!

vsrisuknimit:
@chosta Can I propose an alternative this evening? I don't like the one above all tag icons either.

Should we move this to git discussion or keep talking in this PR?

chosta:
Here + post in general in discord too

vsrisuknimit:
rect872

@chosta the first four are possible for any corp. The last one is possible for three specific corp and not possible to come together. So total of five boxes maximum.

A VP icon on top of jovian tag. 1 for 1x multiplier. 2 for 2x. 3 is the maximum as we don't count Terraforming Ganymede.

vsrisuknimit:
image843-2

Remove card for global discount and add cutting edge tech.

sienmich:
Nice work! I think I still prefer (a bit) small discounts indicators on the tags. It's more minimalistic, less data to process.
The jovian VP is a  cool idea, this could be added with those small indicators too (bottom right corner of the tag, like usual?)
That cutting edge is tempting, since it's hard to show it in previous version.

vsrisuknimit:
multiplayer-discount

A big picture mockup with other players.

kevinb9n:
i do think that putting the discount amount right above where the tag already is is much more elegant. And it means you can easily see things like "the player I'm passing to has -2 on space, but damn, the player THEY pass to has -8." – it all lines up vertically.

The Jupiter VP symbol is a nice idea and since there aren't any Jovian discounts it doesn't have to fit below the discount row.

(Nothing to the right of Venus has any discounts either, so I would suggest *not* showing zeroes there, in case you decide you want to use that real estate for something else...)

chosta:
Just adding @ssimeonoff's suggestion to have them all at one place.
bar-simon

chosta:
Thanks for all the suggestion and feedback. I've decided to go for a version of the UI that combines many of the excellent suggestions. It will look like this:
current state
+ global discounts displayed on card
+ any other discount displayed on the corresponding tag
+ jovian multiplier on top of tag with VP icon

I'll probably switch back to the old "boxed" view with the updated design from @ssimeonoff. While, @vsrisuknimit's suggestion is the best way to fill the space on top of the player board, it does feel a bit cluttered and unnecessary.
The main reason we went for the player and corp name going top was to make space for discounts identificators. It turned out that there was no need for them to be that big of a size and therefore we can go back to having a compact view. I'll let this settle for a while before doing it though.

P.S. I still need that code review :)

vsrisuknimit:
multiplayer-superscript

Here is a mockup of the superscript design.

I will summarize the discussion on Discord.
- How would you represent cutting edge tech?
- Hard to do rebate.
- The discount above card can be confused with Terralabs ability / Polyphemos.
- But I guess we have to go with this if we are going back to the box design.
- As a player, I don't care to read vertically. There is no need to know all players' space discount at once. What is more important while playing is horizontal reading. I care about all the discount of one player (leader or the person immediately downstream of my draft, or the person after that). This is especially useful if all the discounts are together next to the player name.
- Whatever design we go with, I would consider putting discount next to name of the person downstream of my draft. Here is an example: 

draft-discount

chosta:
> * Whatever design we go with, I would consider putting discount next to name of the person downstream of my draft. Here is an example:

I quite like this idea, too. 

Let's roll with `superscript` + `jovians` and see where it takes us. Ultimately, we can have 2 or more UI views depending on the player needs / feedback.

vsrisuknimit:
Two more points I forgot:
- Try top center instead of superscript (top right). See below.
- Add a toggle to hide/show discount next to concise/full button.

multiplayer-superscript

chosta:
> * Try top center instead of superscript (top right). See below.
I don't like it. There is no consistancy with the alloy bonus. I can try and move the mc a bit to the left but for me right > center.
> * Add a toggle to hide/show discount next to concise/full button.
Not in this PR. If we have a button like that would it be hiding the alloy bonus too?

vsrisuknimit:
> I don't like it. There is no consistancy with the alloy bonus. I can try and move the mc a bit to the left but for me right > center.

We can move the alloy bonus toward the center if you would like. Superscript is fine for alloy since there is empty space on the right. Tags are much closer together, superscript looks less good there. I don't think we need consistency here. They mean different things.

> Not in this PR. If we have a button like that would it be hiding the alloy bonus too?

Don't think so. The alloy bonus does not add as much clutter as the tag discount.

chosta:
Sorry, I just don't agree with you. I'd like for this PR to go through and get feedback from the players. Then we can take action. If this is not widely accepted, I'll implement your suggestion. If it's moderately accepted, then I'll implement your suggestion and add a "compact" mode that will also enable discounts toggle. If people think its trash, I'll accept that too.
Also, I think the screenshot with the "full picture" is a bit deceiving. Currently, there is are NO rebates implemented and even if they were, they wouldn't be placed like that (it was just a thought I suggested on the go). Again, this PR is about discounts. I don't mind throwing work in the bin but I need actual player feedback before doing so.

ssimeonoff:
I am a bit worried that with all those MC(yellow) icons everywhere, the UI will get as colourful and bright as a xmas tree.

chosta:
> I am a bit worried that with all those MC(yellow) icons everywhere, the UI will get as colourful and bright as a xmas tree.

You started this :)

But seriously, a fully discounted game only comes in the last generations. I even think it's good on a psychological level.
Building an engine = Having nice stuff
Sir Lancelot's upgrades do look good, right :) ?
knights

I can't stress enogh how important is to roll this to players and actually see it in game. We can make a thousand guesses what is good or not but until we have the experience in a real game, no conclusions can be drawn, especially with a controvercial design solutions because they all are.
And to repeat myself, I don't mind rolling back or modifying any design as long as it's the right thing to do. But ... Do, we must.

src/components/overview/PlayerTagDiscount.ts
```diff
@@ -0,0 +1,29 @@
+import Vue from 'vue';
+
+export const PlayerTagDiscount = Vue.component('PlayerTagDiscount', {
+  props: {
+    amount: {
+      type: Number,
+    },
+  },
+  methods: {
+    getClasses: function(): string {
```
vsrisuknimit, line 10: Same here, do we need a method for this?

src/components/overview/PlayerTags.ts
```diff
@@ -3,8 +3,141 @@ import {TagCount} from '../TagCount';
...
+interface CardTagDiscount {
+  tag: InterfaceTagsType,
+  amount: number
+}
+type InterfaceTagsType = Tags | SpecialTags | 'all' | 'separator';
+
+const JOVIAN_MULTIPLIERS: Array<CardName> = [
+  CardName.IO_MINING_INDUSTRIES,
+  CardName.GANYMEDE_COLONY,
+  CardName.TERRAFORMING_GANYMEDE,
```
vsrisuknimit, line 21: TfG is not an ongoing multiplier. Once it is out of the hand, the TR is adjusted. It does not matter anymore if you pass that person more jovian.

src/components/overview/PlayerTags.ts
```diff
@@ -93,20 +237,49 @@ export const PlayerTags = Vue.component('player-tags', {
...
+      return tag === Tags.JOVIAN && this.playerJovianMultipliersCount() > 0;
+    },
+    playerJovianMultipliersCount: function(): number {
+      let multipliers = 0;
+      for (const card of this.player.playedCards) {
+        if (card !== undefined && JOVIAN_MULTIPLIERS.includes(card.name as CardName)) {
+          multipliers += 1;
+        }
+      }
+      console.log(multipliers, 'multi');
```
vsrisuknimit, line 250: remove console log?

vsrisuknimit:
> I can't stress enogh how important is to roll this to players and actually see it in game. We can make a thousand guesses what is good or not but until we have the experience in a real game, no conclusions can be drawn, especially with a controversial design solutions.

I made some comments on the code, but I am probably not qualified to review and approve this.

If we want to see player feedback sooner, we could try a beta server.

src/components/overview/PlayerTags.ts
chosta: yes

src/components/overview/PlayerTags.ts
chosta: Correct, my bad

src/components/overview/PlayerTagDiscount.ts
chosta: It used to need it because of something I had to add. Well spotted.

src/components/overview/PlayerTagDiscount.ts
```diff
@@ -0,0 +1,29 @@
...
+      const classes = ['player-tag-discount'];
+      return classes.join(' ');
+    },
+    getMegacreditsClasses: function(): string {
```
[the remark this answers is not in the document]
chosta: Deprecated, not used.

src/components/overview/JovianMultiplier.ts
```diff
@@ -0,0 +1,18 @@
...
+    },
+  },
+  methods: {
+    getClasses: function(): string {
+      const classes = ['player-jovian-multiplier'];
+      return classes.join(' ');
+    },
+  },
+  template: `
+    <div :class="getClasses()">{{ amount }}</div>
```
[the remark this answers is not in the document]
chosta: Not needed too

chosta:
Thanks for the comments. Last commit should resolve the issues. No need for beta testing this...It's not a crazy new feature, just a couple of helpful icons. I'm curious to see if people notice the jovian multiplier indicator. This needs to go in production without explanation, so we can have genuine reactions. Trust me.

assets/translations.json
```diff
@@ -0,0 +1 @@
...拥有16张手牌","de":"Mindestens 16 eigene Handkarten im Moment der Beanspruchung dieses Meilensteins.","fr":"Avoir au moins 16 cartes en main lorsque vous validez cet objectif","pl":"W momencie uzyskania tego Tytułu gracz musi mieć na ręce przynajmniej 16 kart","ru":"Соберите в руке не менее 16 карт"},"Polar Explorer":{"cn":"极地探险家","de":"Polarforscher","pl":"Odkrywca Polarny","ru":"Полярник"},"Requires that you have 3 tiles on the two bottom rows":{"cn":"底图最下方的两行上拥有至少3个版块","de":"Benötigt 3 eigene Plättchen auf den unteren beiden Gebietsreihen.","pl":"Gracz posiada w sumie przynajmniej 3 obszary w dwóch dolnych rzędach pól mapy","ru":"Выложите хотя бы 3 жетона на два нижних ряда"},"Rim Settler":{"cn":"木星定居者","de":"Grenzsiedler","pl":"Osadnik Rubieży","ru":"Пионер"},"Requires that you have 3 jovian tags":{"cn":"拥有3个木星标志","de":"Benötigt mindestens 3 eigene Jupitersymbole im Spiel.","pl":"Gracz posiada w grze przynajmniej 3 symbole Jowisza","ru":"На ваших картах должно быть не менее 3 меток Юпитера"},"Specialist":{"cn":"专业人士","de":"Spezialist","pl":"Specjalista","ru":"Специалист"},"Requires that you have at least 10 in production of any resource":{"cn":"拥有至少10点任意产能","de":"Eine Produktionsket
```
vsrisuknimit, line 1: Is this file supposed to be part of this PR?

compile-vue-templates.js
```diff
@@ -294,6 +294,7 @@ function checkComponent(name, component, dataProperties) {
   }
 
   if (result.tips.length > 0) {
+    console.log(result.tips);
```
vsrisuknimit, line 297: no need for console.log?

src/components/overview/PlayerTags.ts
```diff
@@ -3,8 +3,140 @@ import {TagCount} from '../TagCount';
...
+  Tags.EVENT,
+  SpecialTags.NONE,
+  Tags.WILDCARD,
+  SpecialTags.INFLUENCE,
+  SpecialTags.CITY_COUNT,
+  SpecialTags.COLONY_COUNT,
+];
+
+export const checkTagUsed = (tag: InterfaceTagsType, player: PlayerModel) => {
+  if (player.gameOptions.coloniesExtension === false && tag === SpecialTags.CITY_COUNT) {
```
vsrisuknimit, line 126: Is this supposed to be COLONY_COUNT instead of CITY_COUNT?

kevinb9n:
I'm happy with how all this is going!

assets/translations.json
chosta: It's not supposed to be there but it got in between commits. A merge with main fixes it, so don't worry about it.

src/components/overview/PlayerTags.ts
chosta: Actually it's supposed to be a logical `||`
This check should ignore cities and colonies

compile-vue-templates.js
chosta: no need, true

vsrisuknimit (approved):
Look good to me but let's see what others say.

sienmich (approved):

chosta:
Please, someone add a do not merge, Jeep pointed out a possible issue that
I'll check as soon as I can. Thanks

src/components/overview/PlayerTags.ts
```diff
@@ -3,8 +3,140 @@ import {TagCount} from '../TagCount';
...
+  if (player.gameOptions.coloniesExtension === false || tag === SpecialTags.CITY_COUNT) {
+    return false;
+  }
+  if ((player.turmoil !== undefined) === false && tag === SpecialTags.INFLUENCE) {
+    return false;
+  }
+  if (player.gameOptions.venusNextExtension === false && tag === Tags.VENUS) {
+    return false;
+  }
+  if (player.gameOptions.moonExpansion && tag === Tags.MOON) {
```
vsrisuknimit, line 135: Should not it be `player.gameOptions.moonExpansion === false && tag === Tags.MOON`?

src/components/overview/PlayerTags.ts
```diff
@@ -3,8 +3,140 @@ import {TagCount} from '../TagCount';
...
+  Tags.EVENT,
+  SpecialTags.NONE,
+  Tags.WILDCARD,
+  SpecialTags.INFLUENCE,
+  SpecialTags.CITY_COUNT,
+  SpecialTags.COLONY_COUNT,
+];
+
+export const checkTagUsed = (tag: InterfaceTagsType, player: PlayerModel) => {
+  if (player.gameOptions.coloniesExtension === false || tag === SpecialTags.CITY_COUNT) {
```
vsrisuknimit, line 126: Something is weird here. If colonies expansion is not in use, this function always return false. For all tags. That can't be right.

vsrisuknimit (changes requested):

chosta:
Yes, this needs to be changed. Might add a test as well.

src/components/overview/PlayerTags.ts
chosta: It's the same as `undefined` is considered falsy in javascript but for the sake of consistency it's better to have the === operator

src/components/overview/PlayerTags.ts
chosta: You were correct in your first comment here. It was a typo. I've made a mess of this...Next commit should be fine though.

src/components/overview/PlayerTags.ts
vsrisuknimit: @chosta it is not the equivalent. If Moon is not in use, player.gameOptions.moonExpansion would be undefined and you would get False for the whole if statement even if tag = Tags.MOON. You want it to be true there.

vsrisuknimit (approved):

vsrisuknimit:
Look good to my inexperienced eyes. Not sure if you want real programmers to take a look.

src/components/overview/PlayerTags.ts
chosta: That is wanted behavior. The order of `&&` and `||` resolves from left to right. So if the moon is undefined or false it will ignore the statement to the right.

chosta:
> Look good to my inexperienced eyes. But you may want real programmers to take a look still.

You found out enough :). It was already approved by a dev. It should be fine. Thanks.

src/components/overview/PlayerTags.ts
vsrisuknimit: It does not matter since you fixed it. But if the Moon expansion is not in use, don't you want this function to return false? The way it was written before, the if statement would be come false and you skip returning false here.

src/styles/players_overview.less
```diff
@@ -213,19 +263,19 @@
...
-    
     .players-overview {
         .player-info-bottom {
             flex-flow: column !important;
         }
         .player-status-and-res {
-            margin-bottom: 8px;
-            border-bottom: 2px solid rgba(170, 170, 170, 0.2);
+            margin-bottom: 48px;
+            // border-bottom: 2px solid rgba(170, 170, 170, 0.2);
```
vsrisuknimit, line 278: Should this be removed rather than commented out?

Issue #2538, opened 2021-01-24
Serialize deferredActions with Game


bafolts:
When we serialize a game we set `deferredActions` to an empty array. We should serialize the deferred actions and deserialize them. This will make the undo operation more reliable and allow us to eventually let the user undo any operation.

https://github.com/bafolts/terraforming-mars/blob/main/src/Game.ts#L394

Lynesth:
There won't be any when we serialize the game since we're saving at times when the deferred actions queue is empty.
Are we planning on saving more often ?

bafolts:
I would like to save more often. Ideally we would save after every player input. During server restarts we then wouldn't lose any data. I am thinking about this in isolation. When `Player` or `Game` is serialized and deserialized we should end up with the same instance.

PR #2558, opened 2021-01-25
replace waitingForCb with object

### LINKED ISSUE #2539: Serialize waitingForCb with Player
During an undo operation or whenever we save the game state we lose the `waitingForCb`. This means when we serialize a `Player` we are losing state.

https://github.com/bafolts/terraforming-mars/blob/main/src/Player.ts#L74

Since a function can not be serialized this will require some form of refactor. Instead of relying on a callback we will need to rely on some structure we can serialize.

This should allow for a more robust undo operation and fix bugs when restoring games from database.

bafolts:
This will start to close #2539. The `waitingForCb` on `Player` is a function. This can't be serialized and it relies on closures to keep state. For this change the `waitingForCb` is replaced with a `PlayerCallback` (probably a better name) interface. These objects can be serialized as we can store `PlayerId`, `GameId`, and `PlayerCallbackId` to store the callback. Then upon deserialization we can have `Player` where it was when serialized. This will give us more flexibility with when we can save game state and hopefully eventually save game state during every state update.

I also added new `TestingUtils` functions for running actions from the `DeferredActionsQueue`. These should help us change the API with `DeferredActionsQueue` without needing to update every place we use the API in tests. It also removes the separate test only function from `DeferredActionsQueue`.

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -781,7 +781,7 @@ export class Game implements ISerializable<SerializedGame> {
     this.deferredActions.runAll(new Game.DraftOrResearch(this));
   }
 
-  public goToDraftOrResearch() {
+  private goToDraftOrResearch() {
     this.generation++;
     this.log('Generation ${0}', (b) => b.forNewGeneration().number(this.generation));
     this.incrementFirstPlayer();
@@ -1065,7 +1065,7 @@ export class Game implements ISerializable<SerializedGame> {
     player.takeActionForFinalGreenery();
   }
 
-  public startActionsForPlayer(player: Player) {
+  private startActionsForPlayer(player: Player) {
     this.activePlayer = player.id;
     player.actionsTakenThisRound = 0;
 
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -26,12 +26,15 @@ import {ITile} from './ITile';
 import {LogBuilder} from './LogBuilder';
 import {LogHelper} from './LogHelper';
 import {LogMessage} from './LogMessage';
+import {NoopCallback} from './server/callbacks/NoopCallback';
 import {ALL_MILESTONES} from './milestones/Milestones';
 import {ALL_AWARDS} from './awards/Awards';
 import {OriginalBoard} from './boards/OriginalBoard';
 import {PartyHooks} from './turmoil/parties/PartyHooks';
 import {Phase} from './Phase';
 import {Player, PlayerId} from './Player';
+import {PlayerCallback} from './server/PlayerCallback';
+import {PlayerCallbackId} from './server/PlayerCallbackId';
 import {PlayerInput} from './PlayerInput';
 import {ResourceType} from './ResourceType';
 import {Resources} from './Resources';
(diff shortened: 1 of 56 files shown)


src/Game.ts
```diff
@@ -749,10 +778,10 @@ export class Game implements ISerializable<SerializedGame> {
   }
 
   private resolveTurmoilDeferredActions() {
-    this.deferredActions.runAll(() => this.goToDraftOrResearch());
+    this.deferredActions.runAll(new Game.DraftOrResearch(this));
   }
 
-  private goToDraftOrResearch() {
+  public goToDraftOrResearch() {
```
bafolts, line 784: I will change this back to `private`. With the `private static` property for the class this can remain `private`.

Lynesth:
_Note: Read it before bed, might have missed something._

I totally get where you want to go with #2538 and #2539 and it probably would be great, but isn't it a bit overkill ?
You already created a new class for each `run` and `runAll` callbacks (even had to do some inception on one of them), and you'll have to handle serializing/deserializing those, as well as serializing/deserializing all deferred actions (that means we can't use the generic one anymore ?), and do the same for player inputs ? 

I mean, server restarts are already pretty well handled as they are able to start back to the last unfinished action and we even save between draft rounds. And it's not like they should be happening anyway during a game, those restarts.

Just my own (and probably unpopular) opinion on this.

bafolts:
> I mean, server restarts are already pretty well handled as they are able to start back to the last unfinished action and we even save between draft rounds. And it's not like they should be happening anyway during a game, those restarts.

Heroku can restart the dyno whenever it needs to or wants to. The hobby plan dynos make no guarantee that the dyno will remain up. If you are in the middle of a game during the restart your game is taken to the last unfinished action. Every morning when I push up the latest version of the code I will be taking some user's games back to the last unfinished action without them being aware that it happened.

With this strategy this would mean that we have to name ever deferred action and player input. The end goal would be that we can restart the server at anytime and not lose game data.

I could very well be over complicating this. Can you think of another way to not lose game state during server restarts?

Lynesth:
Sure, dynos can be restarted by Heroku whenever they feel like it (free ones should not go to sleep though unless it's been idle for a long time, which means you're not in the middle of an action).

And no I cannot think of another way to preserve the game state during server restart, your approach is probably the good one for that.

But what I mean is that, most of the time, an action would prompt you once or twice, so if a restart happens at that time, well, I really don't think it's a big deal at all. And I think we already have a pretty good compromise between what we provide as a save mechanism and the "complexity" of the code. We probably are already doing more than most game servers in that regard.

I think it will be a big endeavor for something that is probably not needed.

Lynesth:
**Note:** Save mechanism is currently less good for games that do not allow undo as they will have the player restart its entire turn upon server restart instead of just the unfinished action. It's just a matter of removing 1 line in the code but it was someone's wish that it behave like this if I recall correctly (although I can't remember exactly).

bafolts:
Will close this. There are enough open issues as it is and no users are asking for this feature for what the work would involve.

PR #2559, opened 2021-01-25
Player.canAfford - further optimizations

sienmich:
I added `reserveUnits` to `Player.canAfford`.
This allows to simplify the `Player.canPlay` significantly.

In future it could be used a lot in cards:
The `LocalHeatTrapping` could change from:
```ts
  public canPlay(player: Player): boolean {
    const requiredHeatAmt = 5;

    // Helion must be able to pay for both the card and its effect
      return (player.heat >= requiredHeatAmt) && (player.heat + player.megaCredits >= requiredHeatAmt + player.getCardCost(this));
    }

    return player.availableHeat >= requiredHeatAmt;
  }
```
to
```ts
  public canPlay(player: Player): boolean {
    return player.canAfford(player.getCardCost(this), false, false, false, false, {heat: 5});
  }
```
and a lot of cards like `Asteroid`:
```ts
return player.canAfford(player.getCardCost(this) + REDS_RULING_POLICY_COST * stepsRaised, false, true);
```
could properly calculate the `canAfford()`. Right now they count the `REDS_RULING_POLICY_COST` as part of card price, which means it can be payed with titanium, but it can not. But that's not this PR.

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1594,6 +1594,7 @@ export class Player implements ISerializable<SerializedPlayer> {
   // and additionally pay the reserveUnits (no replaces here)
   // TODO(sienmich): use options parameter
   public canAfford(cost: number, canUseSteel: boolean = false, canUseTitanium: boolean = false, canUseFloaters: boolean = false, canUseMicrobes : boolean = false, reserveUnits: Partial<Units> = {}): boolean {
+    // Check if player has the reserveUnits - required resources
     const units = Units.of(reserveUnits);
     if (!Units.hasUnits(units, this)) {
       return false;
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1593,18 +1593,17 @@ export class Player implements ISerializable<SerializedPlayer> {
   // Checks if the player can afford to pay `cost` mc (possibly replaceable with steal, titanium etc.)
   // and additionally pay the reserveUnits (no replaces here)
   // TODO(sienmich): use options parameter
-  public canAfford(cost: number, canUseSteel: boolean = false, canUseTitanium: boolean = false, canUseFloaters: boolean = false, canUseMicrobes : boolean = false, reserveUnits: Partial<Units> = {}): boolean {
+  public canAfford(cost: number, canUseSteel: boolean = false, canUseTitanium: boolean = false, canUseFloaters: boolean = false, canUseMicrobes : boolean = false, reserveUnits: Units = Units.EMPTY): boolean {
     // Check if player has the reserveUnits - required resources
-    const units = Units.of(reserveUnits);
-    if (!Units.hasUnits(units, this)) {
+    if (!Units.hasUnits(reserveUnits, this)) {
       return false;
     }
 
     return cost <=
-      this.megaCredits - units.megacredits +
-      (this.canUseHeatAsMegaCredits ? this.heat - units.heat : 0) +
-      (canUseSteel ? (this.steel - units.steel) * this.steelValue : 0) +
-      (canUseTitanium ? (this.titanium - units.titanium) * this.getTitaniumValue() : 0) +
+      this.megaCredits - reserveUnits.megacredits +
+      (this.canUseHeatAsMegaCredits ? this.heat - reserveUnits.heat : 0) +
+      (canUseSteel ? (this.steel - reserveUnits.steel) * this.steelValue : 0) +
+      (canUseTitanium ? (this.titanium - reserveUnits.titanium) * this.getTitaniumValue() : 0) +
       (canUseFloaters ? this.getFloatersCanSpend() * 3 : 0) +
       (canUseMicrobes ? this.getMicrobesCanSpend() * 2 : 0);
   }
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1578,39 +1578,14 @@ export class Player implements ISerializable<SerializedPlayer> {
   }
 
   public canPlay(card: IProjectCard): boolean {
-    const canUseSteel = card.tags.includes(Tags.BUILDING);
-    const canUseTitanium = card.tags.includes(Tags.SPACE);
-    const canUseMicrobes = card.tags.includes(Tags.PLANT);
-    const canUseFloaters = card.tags.includes(Tags.VENUS);
-
-    let steel = this.steel;
-    let titanium = this.titanium;
-
-    // Adjust available steel based on reserve costs on Moon cards. Also reject cards with
-    // reserve costs that a player cannot afford.
-    if (card.reserveUnits !== undefined) {
-      const reserveUnits = MoonExpansion.adjustedReserveCosts(this, card);
-      // Set aside reserve units in case the card has a building tag.
-      // If there isn't enough steel to meet the purchase reserve, this isn't a playable card.
-      steel = steel - reserveUnits.steel;
-      if (steel < 0) {
-        return false;
-      }
-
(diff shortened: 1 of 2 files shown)


kberg:
Just reading the description alone,

```
  public canPlay(player: Player): boolean {
    return player.canAfford(player.getCardCost(this), false, false, false, false, {heat: 5});
  }
```

would be even better without all the repeated uses of `false`. Is that doable?

sienmich:
> Just reading the description alone,
> 
> ```
>   public canPlay(player: Player): boolean {
>     return player.canAfford(player.getCardCost(this), false, false, false, false, {heat: 5});
>   }
> ```
> 
> would be even better without all the repeated uses of `false`. Is that doable?

Of course, I'll do it in the next PR, as noted in TODO in code. Just wanted to keep thinks simpler, one change at a time

src/Player.ts
```diff
@@ -1578,47 +1578,32 @@ export class Player implements ISerializable<SerializedPlayer> {
...
-      (canUseSteel ? this.steel * this.steelValue : 0) +
-      (canUseTitanium ? this.titanium * this.getTitaniumValue() : 0) +
+  // Checks if the player can afford to pay `cost` mc (possibly replaceable with steal, titanium etc.)
+  // and additionally pay the reserveUnits (no replaces here)
+  // TODO(sienmich): use options parameter
+  public canAfford(cost: number, canUseSteel: boolean = false, canUseTitanium: boolean = false, canUseFloaters: boolean = false, canUseMicrobes : boolean = false, reserveUnits: Partial<Units> = {}): boolean {
+    const units = Units.of(reserveUnits);
+    if (!Units.hasUnits(units, this)) {
+      return false;
+    }
```
Lynesth, line 1600: At first glance, this piece of code doesn't make any sense. Could you use another variable name than `units` and add a comment above the `if` explaining what it does ?
This whole `Units` thing really puzzles me but I guess that's not a topic for this PR.

sienmich: `this` refers to the Player. This 'if' is checking if the player has the `units`. I'll that a comment, thanks

src/Player.ts
```diff
@@ -1578,47 +1578,32 @@ export class Player implements ISerializable<SerializedPlayer> {
...
+
+    return cost <=
+      this.megaCredits - units.megacredits +
+      (this.canUseHeatAsMegaCredits ? this.heat - units.heat : 0) +
```
Lynesth, line 1604: Why change this for now ?
The only cards using the `reserveUnits` property are from the Moon expansion and `MoonExpansion.adjustedReserveCosts()` is only managing steel and titanium anyway.

sienmich: For now it's true, I'd like this new function to be useful with cards like `LocalHeatTrapping` in the future. Since it is accepting the whole `Units` class, not only steel and titanium, I think it's ok to make it work for also for heat

sienmich: And `units.megacredits` can also be quite useful (as noted in the PR description) for all the reds policy cost.

src/Player.ts
Lynesth: Oh, I understood what it did, it just didn't make sense at all at first glance.
I mean, `units = Units.of` and `Units.hasUnits(units` really makes you wonder if the dev didn't have a stroke.

sienmich: It would be better to have `player.hasUnits(units)` than `Units.hasUnits(units, player)`. Maybe in next PR?

src/Player.ts
Lynesth: I guess... I'm still not fond of the Units thing. It does not seem right to me...
I'll make a new issue about it

kberg:
> > Just reading the description alone,
> > ```
> >   public canPlay(player: Player): boolean {
> >     return player.canAfford(player.getCardCost(this), false, false, false, false, {heat: 5});
> >   }
> > ```
> > 
> > 
> > would be even better without all the repeated uses of `false`. Is that doable?
> 
> @kberg. Of course, I'll do it in the next PR, as noted in TODO in code. Just wanted to keep thinks simpler, one change at a time

Great, thanks.

src/Player.ts
kberg: Yes, player.hasUnits is where this should go. In another PR (or issue) I made the same comment.

kberg (commented):
Here's what I don't understand -- what is this optimizing?

sienmich:
> Here's what I don't understand -- what is this optimizing?


So @kberg , earlier you wrote:
```ts
    let steel = this.steel;
    let titanium = this.titanium;

    // Adjust available steel based on reserve costs on Moon cards. Also reject cards with
    // reserve costs that a player cannot afford.
    if (card.reserveUnits !== undefined) {
      const reserveUnits = MoonExpansion.adjustedReserveCosts(this, card);
      // Set aside reserve units in case the card has a building tag.
      // If there isn't enough steel to meet the purchase reserve, this isn't a playable card.
      steel = steel - reserveUnits.steel;
      if (steel < 0) {
        return false;
      }

      // Set aside reserve units in case the card has a space tag.
      // If there isn't enough titanium to meet the purchase reserve, this isn't a playable card.
      titanium = titanium - reserveUnits.titanium;
      if (titanium < 0) {
        return false;
      }
```
There are those to if-s checking if player has those `reserveUnits`. Since i'd like this to work for other units as well (like MC for reds policy), we have to check if all reserved units are sufficient, that's what that new `if` is doing. Right?

kberg:
Aha

src/Player.ts
```diff
@@ -1578,47 +1578,33 @@ export class Player implements ISerializable<SerializedPlayer> {
...
   }
 
-  public canAfford(cost: number, canUseSteel: boolean = false, canUseTitanium: boolean = false, canUseFloaters: boolean = false, canUseMicrobes : boolean = false): boolean {
-    return cost <= this.spendableMegacredits() +
-      (canUseSteel ? this.steel * this.steelValue : 0) +
-      (canUseTitanium ? this.titanium * this.getTitaniumValue() : 0) +
+  // Checks if the player can afford to pay `cost` mc (possibly replaceable with steal, titanium etc.)
+  // and additionally pay the reserveUnits (no replaces here)
+  // TODO(sienmich): use options parameter
+  public canAfford(cost: number, canUseSteel: boolean = false, canUseTitanium: boolean = false, canUseFloaters: boolean = false, canUseMicrobes : boolean = false, reserveUnits: Partial<Units> = {}): boolean {
```
kberg, line 1596: Can reserveUnits be of type Units instead of Partial?

kberg (commented):
Minor comments.

src/Player.ts
sienmich: Why? This function (`player.canAfford`) is used in a lot of places, in cards. It would be nice to be able to call it like `player.canAfford(..., {megacredits: 3})` instead of `player.canAfford(..., Units.of({megacredits: 3}))`

sienmich: And since the Units are probably going to change soon that's not that big of a deal, right?

kberg: > And since the Units are probably going to change soon that's not that big of a deal, right?

I haven't participated in the discussion yet, so I can't say anything about that.

Can you please drop the Partial for now? It's easier to switch from Units to Partial<Units> than the other way around.

sienmich: Ok, done

kberg (approved):
Thanks @sienmich. I'm sorry this took as long as it did.

PR #2561, opened 2021-01-26
Make a proper Priority list for deferred actions

Lynesth:
First PR of a few that will then add priorities everywhere needed to finally try to get effects order as good as we can.

### DIFF
--- a/src/cards/colonies/AirRaid.ts
+++ b/src/cards/colonies/AirRaid.ts
@@ -22,8 +22,8 @@ export class AirRaid implements IProjectCard {
   }
 
   public play(player: Player, game: Game) {
-    game.defer(new RemoveResourcesFromCard(player, ResourceType.FLOATER, 1, true));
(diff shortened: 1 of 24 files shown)


bafolts (approved):

src/cards/colonies/AirRaid.ts
```diff
@@ -22,8 +22,8 @@ export class AirRaid implements IProjectCard {
   }
 
   public play(player: Player, game: Game) {
-    game.defer(new RemoveResourcesFromCard(player, ResourceType.FLOATER, 1, true));
     game.defer(new StealResources(player, Resources.MEGACREDITS, 5));
+    game.defer(new RemoveResourcesFromCard(player, ResourceType.FLOATER, 1, true));
```
Lynesth, line 26: Just to make it clear, changing the order of the `game.defer` isn't necessary since it'll be ordered properly in the queue.
But since stealing resources will trigger before removing the floater with the priorities that have been set, I thought it would make sense to have them in the correct order here as well.

PR #2565, opened 2021-01-26
Defer onTilePlaced effects + modify Eco/Geo Surveys 

Lynesth:
This is the first of many PRs that intends to add all effects to the queue in order to properly execute them in the correct order.
I started "small", with the cards that have an effect triggered when a tile is placed.

@kberg, I had to rewrite how **Ecological Survey** and **Geological Survey** worked and put the logic directly on the card, in an `onTilePlaced` method instead of relying on the AresHandler.

I added a couple of new utility functions in the tests, inspired by one of @bafolts latest PR.

-----

I'll stop working on this any further until I have a proper agreement from you guys that this is ok, as it introduces 2 main changes in how things will unfold in the game, both code wise and for the players.

#### Code wise:
The end goal is that all effects, even a simple "Gain 3 MC", will get deferred. While we could think that most of them should be fine in the current situation, with the amount of new cards and mechanics being added by fan made expansions and other community stuff, it will be much more reliable to apply the same "treatment" to all of them.

#### For the players:
Because the goal is to have optimal ordering between effects, independently of the situation*, some things might happen later than players expect them to. Playing a simple **Comet** won't give you the bonuses from your tile placement (either from space bonus or ocean adjacency) until the end of the action, which means after you've removed someone's plants.

_&ast; probably minus some exceptions._

-----

_Notes: Best viewed commit by commit._


src/cards/ares/GeologicalSurvey.ts
```diff
@@ -28,6 +35,31 @@ export class GeologicalSurvey extends Card implements IProjectCard {
...
+  }
+
+  public onTilePlaced(player: Player, space: ISpace) {
+    if (player.game.phase === Phase.SOLAR || player.id !== space.player?.id) {
+      return;
+    }
+
+    // Steel, Titanium and Heat
+    ([[Resources.STEEL, SpaceBonus.STEEL], [Resources.TITANIUM, SpaceBonus.TITANIUM], [Resources.HEAT, SpaceBonus.HEAT]] as [Resources, SpaceBonus][]).forEach(([resource, bonus]) => {
+      if ((resource === Resources.STEEL && space.spaceType !== SpaceType.COLONY) ||
```
Lynesth: This line was for Mars First interaction and has been fixed in the latest commit.

src/deferredActions/GainProduction.ts
```diff
@@ -0,0 +1,34 @@
...
+    public player: Player,
+    public resource: Resources,
+    public options: GainProduction.Options = {},
+  ) {}
+
+  public execute() {
+    if (this.options.count === undefined) {
+      this.options.count = 1;
+    } else {
+      this.options.count = Math.abs(this.options.count);
```
Lynesth, line 26: We can't force an argument's type to be something like a positive (or unsigned) integer.
So I either had to:
 - Throw an error if `count` was lower than 0
 - Make sure it's positive

I don't mind either way of doing it. Went with the latter since the deferred actions name are pretty clear, so negative number as argument would most likely be a simple mistake.

Let me know if you want this to change.

bafolts: I would prefer to throw an error if `count` is not what we expect. This will fail faster and ensure a negative value doesn't indirectly end up being provided.

tests/TestingUtils.ts
```diff
@@ -79,6 +79,29 @@ export const setRulingPartyAndRulingPolicy = function(game: Game, turmoil: Turmo
   game.phase = Phase.ACTION;
 };
 
+
+// Just shortcuts to some often called methods
+// related to the deferred actions queue
+export function runAllActions(game: Game) {
+  game.deferredActions.runAll(() => {});
```
bafolts, line 86: :+1: glad to see something from my PR survived!

src/cards/ares/EcologicalSurvey.ts
```diff
@@ -28,6 +37,43 @@ export class EcologicalSurvey extends Card implements IProjectCard {
     });
   }
 
+  private hasAdjacencyBonus(player: Player, space: ISpace, bonus: SpaceBonus) {
+    return player.game.board.getAdjacentSpaces(space).some((adj) => adj.adjacency?.bonus.includes(bonus));
+  }
+
+  public onTilePlaced(player: Player, space: ISpace) {
```
bafolts, line 44: This change makes sense to me. It still keeps the logic isolated to ares as it is in the ares card. Arguably it is easier to see that this card is impacted by the `onTilePlaced` where it is directly on the card.

src/deferredActions/GainProduction.ts
Lynesth: Done.

src/cards/ares/EcologicalSurvey.ts
kberg: How does this take into account bonuses accrued from adjacency tiles?

kberg: Or bonuses accrued from effects that may allow a player to gain, for instance, a plant, because they get a plant bonus when someone places an ocean?

tests/cards/turmoil/LakefrontResorts.spec.ts
```diff
@@ -2,19 +2,20 @@ import {expect} from 'chai';
 import {LakefrontResorts} from '../../../src/cards/turmoil/LakefrontResorts';
 import {Game} from '../../../src/Game';
 import {Resources} from '../../../src/Resources';
-import {TestPlayers} from '../../TestingUtils';
+import * as Utils from '../../TestingUtils';
```
kberg, line 5: Instead of Utils, can you either rename TestingUtils to Utils, or create a TestingUtils namespace?

src/cards/ares/EcologicalSurvey.ts
Lynesth: Arctic Algae and Mars First are both checked for in their respective card.
And yes, it checks for adjacency bonuses as well (using their private method `hasAdjacencyBonus`.

tests/cards/turmoil/LakefrontResorts.spec.ts
Lynesth: Using a namespace probably a good idea, but that means I need to modify all tests using TestingUtils, which is out of the scope of this PR.
How about I do that in a big batch after this one ?

nwai90:
LGTM. Deferring the effects so that there is a consistent and expected order makes sense overall. It might be slightly confusing to users initially that tile placement bonuses are not granted immediately, but I can't see a scenario besides Reds (which is the subject of another PR) where it would matter that the player receives these bonuses at the end of the action.

Lynesth:
> 
> 
> LGTM. Deferring the effects so that there is a consistent and expected order makes sense overall. It might be slightly confusing to users initially that tile placement bonuses are not granted immediately, but I can't see a scenario besides Reds (which is the subject of another PR) where it would matter that the player receives these bonuses at the end of the action.

I'll properly do something about Reds after the whole priority thing and cards changes to static are over.
Also, Reds effect will be in the LOSE_RESOURCE_OR_PRODUCTION priority, which triggers after gaining bonuses from tile placement anyway.

Lynesth:
Could I get @bafolts and/or @kberg opinion regarding the changes that this will bring (detailed in the PR's description) so that I know if I keep working on this or not ?
It doesn't have a clear approval from anyone yet, but no one is dismissing it either, so... I'm a bit confused.

tests/cards/turmoil/LakefrontResorts.spec.ts
kberg: I was going to do something like this, and it makes it possible to do it in pieces. Let me set that up for you.

src/cards/ares/GeologicalSurvey.ts
```diff
@@ -28,6 +38,31 @@ export class GeologicalSurvey extends Card implements IProjectCard {
...
+    // Steel, Titanium and Heat
+    ([[Resources.STEEL, SpaceBonus.STEEL], [Resources.TITANIUM, SpaceBonus.TITANIUM], [Resources.HEAT, SpaceBonus.HEAT]] as [Resources, SpaceBonus][]).forEach(([resource, bonus]) => {
+      if ((resource === Resources.STEEL && space.spaceType !== SpaceType.COLONY && PartyHooks.shouldApplyPolicy(player.game, PartyName.MARS, TurmoilPolicy. MARS_FIRST_DEFAULT_POLICY)) ||
+          space.bonus.includes(bonus) ||
+          this.hasAdjacencyBonus(player, space, bonus)) {
+        player.game.defer(new GainResources(
+          player,
+          resource,
+          {
+            logMessage: '${0} gained a bonus ${1} because of ${2}',
```
kberg, line 59: I would like to suggest that this and the logBuilder get simplified to a single paramter being the source of the change, and the GainResources handle everything about logging. This is somewhat related to my suggestion that all resource management start to be internalized to Player.

kberg (commented):
This is getting clear if the deferring priority is right. Some new questions, as promised! :D More reading on my part in the meantime.

kberg:
Thanks so much for your patience.

src/cards/ICard.ts
```diff
@@ -39,7 +40,7 @@ export interface ICard {
     getVictoryPoints?: (player: Player) => number;
     onCardPlayed?: (player: Player, card: IProjectCard) => OrOptions | void;
     onStandardProject?: (player: Player, projectType: StandardProjectCard) => void;
-    onTilePlaced?: (player: Player, space: ISpace) => void;
+    onTilePlaced?: (player: Player, space: ISpace) => DeferredAction | void;
```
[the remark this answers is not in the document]
Lynesth: Ok so, there's 2 ways to approach this and I simply picked one (but we can go with the other if you prefer, I don't mind).
First a quick explanation:

**1)** If an effect applies to the **active** player, then it is deferred with its default priority (which is set in the DeferredAction, eg. Priority.GAIN_RESOURCES_OR_PRODUCTION for GainResources deferred action).

**2)** If an effect applies to **another** player, then it needs to be deferred with a piority set to Priority.OPPONENT_TRIGGER instead of its default one (so that it happens before anything else).

So effects that fall into **1)** can be deferred directly in the `onTilePlaced` method of the card.
For those in **2)**, I need to check if it's going to trigger for the active player or another one and change the priority if needed.

Example: Rover Construction is in play and someone places a city. If Rover belong to the active player, then it's going to be deferred with its default priority, if Rover belong to another player then it will be deferred with the OPPONENT_TRIGGER priority.

Now, `onTilePlaced` receives two arguments: the player who owns the card and the space where the tile was placed.
So to deal with **2)** (check if active player or not), I have 2 solutions:
 - Add a 3rd argument to the method declaration to pass it the player who placed the tile, that way I can compare/defer inside the method.
 - Return the applicable deferred action and compare/defer it outside of the method, where I have access to both players (solution I picked, you can see the check line 1329 in the Game.ts of this PR)

The exact same question will be asked regarding `onCardPlayed` when I later come to it (if ever).
I really don't mind deferring inside methods only, it might make things cleaner for Game and Player.

src/Game.ts
```diff
@@ -1328,13 +1323,18 @@ export class Game implements ISerializable<SerializedGame> {
 
     // Mining Guild tile placement effects should resolve before removing space.player for Oceans
     this.players.forEach((p) => {
-      if (p.corporationCard !== undefined &&
-          p.corporationCard.onTilePlaced !== undefined) {
-        p.corporationCard.onTilePlaced(p, space);
+      if (p.corporationCard?.onTilePlaced !== undefined) {
```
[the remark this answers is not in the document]
Lynesth: See answer to the other comment below.

src/cards/ares/EcologicalSurvey.ts
```diff
@@ -28,6 +37,43 @@ export class EcologicalSurvey extends Card implements IProjectCard {
...
+    if (player.game.phase === Phase.SOLAR || player.id !== space.player?.id) {
+      return;
+    }
+
+    // Plants
+    if (space.bonus.includes(SpaceBonus.PLANT) ||
+        this.hasAdjacencyBonus(player, space, SpaceBonus.PLANT) ||
+        (space.tile?.tileType === TileType.OCEAN && player.playedCards.some((card) => card.name === CardName.ARCTIC_ALGAE))) {
+      player.game.defer(new GainResources(
```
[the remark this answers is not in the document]
Lynesth: That was because you were actually trying to compare the resources the player had before placing the tile, and those he had after the tile was placed.

Now however, we can't do that anymore since the effects triggered from placing a tile can be anywhere in the queue.
So here is what the onTilePlaced does for Eco/Geo Surveys:

- It looks at the space the tile was placed for any bonuses.
- It looks at the adjacent spaces for any adjacency bonuses.
- For both cases above, if microbes/animals, it checks that you have cards you can put them on.
- If you placed an ocean and you have Arctic Algae, it counts that you will get some plants.
- If Mars First is ruling, it will count that you will get some steel.

And using all that info it gives you 1 of each you will receive at some point.

src/cards/ares/GeologicalSurvey.ts
```diff
@@ -28,6 +38,31 @@ export class GeologicalSurvey extends Card implements IProjectCard {
...
+  }
+
+  public onTilePlaced(player: Player, space: ISpace) {
+    if (player.game.phase === Phase.SOLAR || player.id !== space.player?.id) {
+      return;
+    }
+
+    // Steel, Titanium and Heat
+    ([[Resources.STEEL, SpaceBonus.STEEL], [Resources.TITANIUM, SpaceBonus.TITANIUM], [Resources.HEAT, SpaceBonus.HEAT]] as [Resources, SpaceBonus][]).forEach(([resource, bonus]) => {
+      if ((resource === Resources.STEEL && space.spaceType !== SpaceType.COLONY && PartyHooks.shouldApplyPolicy(player.game, PartyName.MARS, TurmoilPolicy. MARS_FIRST_DEFAULT_POLICY)) ||
```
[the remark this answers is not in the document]
Lynesth: I think those cards (Geo/Eco Surveys) should only apply to tiles placed on Mars though.

src/cards/ares/EcologicalSurvey.ts
```diff
@@ -28,6 +37,43 @@ export class EcologicalSurvey extends Card implements IProjectCard {
     });
   }
 
+  private hasAdjacencyBonus(player: Player, space: ISpace, bonus: SpaceBonus) {
```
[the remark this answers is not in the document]
Lynesth: Yes that's what it does. I'll change the name indeed.

src/cards/ICard.ts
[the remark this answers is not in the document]
kberg: Not urgent for now. But eventually, if both things do the same thing, let's try to converge on one way. A goal to aim for, not a requirement.

kberg (approved):
Thanks for your patient patience.

bafolts (approved):

Lynesth:
Thank you both for your time reviewing this.
I'll fix the conflicts and merge.

PR #2572, opened 2021-01-27
The last cards from the first set of Moon cards. (With two exceptions.)

kberg:


src/cards/moon/MooncrateBlockFactory.ts
```diff
@@ -0,0 +1,33 @@
...
+    super({
+      name: CardName.MOONCRATE_BLOCK_FACTORY,
+      cardType: CardType.ACTIVE,
+      tags: [Tags.BUILDING],
+      cost: 8,
+
+      metadata: {
+        description: 'Effect: When you pay for a luna standard project (colony, road, mine), you spend 4MC less. / Requires 1 mine on the Moon.',
+        cardNumber: 'M38',
+        renderData: CardRenderer.builder((_b) => {}),
```
chosta, line 21: `renderData` is optional, you could probably skip this line

chosta: Is this mssing the effect actually?

chosta (approved):
LGTM. I'd like to also check the visuals after the merge.

src/cards/moon/MooncrateBlockFactory.ts
kberg: It probably is. There is an absolute need for the cards to get a viz polish.

chosta: I'll check them once you get them all up.

kberg: Thanks. Me too - it'll just be good to do them at once.

Issue #2355, opened 2021-01-12
Next batch of small card iconography issues


kevinb9n:
@chosta 

* Ecoline is missing the oxygen reminder icon

* Celestic: Floater icons should be square

* Arcadian Communities: missing "Effect" banner (space?). Also does it want to use an actual player cube icon perhaps?

* MIssing asterisk: Factorum (action 1), Capital (ocean icon), Zeppelins (city icon)

* Commercial District, Martian Rails: missing red outline and asterisk

* Decomposers has extra slash

* Recruitment and Vote of No Confidence: each should have both a minus and a plus (or skip the plus for both if you prefer).

* This small black icon is only showing up in "max" requirement boxes
<img width=500 src=https://user-images.githubusercontent.com/934551/104382585-377a3f80-54e3-11eb-8bcd-6e4f1612e6e6.png>

* Using a yellow box here suggests that it has something to do with "min" requirements specifically (not "max"):
<img width=250 src=https://user-images.githubusercontent.com/934551/104382913-d737cd80-54e3-11eb-8ae3-2e2ac9c598f2.png>

It could perhaps do something more like this, or just leave out the color entirely?
<img width=170 height=30 src=https://user-images.githubusercontent.com/934551/104383095-1a923c00-54e4-11eb-9034-765b7540a7f2.png>

Applies to Inventrix, Adaptation Tech, Special Design, Morning Star, and Cutting Edge Tech. And Ecology Experts (but maybe that one should just be text-only like the card. I missed that last time, but there really is no icon with the meaning "play a card from hand").

* Not clear why Standard Tech is using a yellow box (that nearly resembles a min-requirement) at all.

* The Trade Fleet icon usually looks like this, is there a reason we don't use it?

<img width="50" alt="Trade Fleet Symbol" src="https://user-images.githubusercontent.com/934551/104407921-9eb1e700-5517-11eb-8a5d-eae590601e02.png">

* On cards like Aquifer Pumping, it would be cool if the resource-icon-in-parens could be a smaller superscript kinda thing.

kevinb9n:
> * Ecoline is missing the oxygen reminder icon
confirmed fixed

> * Celestic: Floater icons should be square
Not fixed because round is supported and square isn't yet.
Cutting Edge has a rectangular one tho, howzat work?


> * Arcadian Communities: missing "Effect" banner (space?). Also does it want to use an actual player cube icon perhaps?
[...]

PR #2574, opened 2021-01-27
Minor fixes for cards

chosta:
#2355
@kevinb9n 
The things I haven't done and probably won't any time soon becaues of other priorities:

- `Celestic: Floater icons should be square` - Current (overall) secondary tag design is round, didn't want to add squares as an exception
- `Arcadian Communities: missing "Effect" banner (space?). Also does it want to use an actual player cube icon perhaps?` - Again, a special case is needed just for one card
- `MIssing asterisk: Factorum (action 1)` - I skipped it when I did the card because, again a special design for just that card might be needed (to have smaller arrows, or smaller MC icon, or custom position for the asterix)
- `Capital (ocean icon)` - Probaby, doable but not in this PR, requires a bit of customization
- `The Trade Fleet icon usually looks like this, is there a reason we don't use it?` - I'm ok for it to be changed but maybe in another PR
- `On cards like Aquifer Pumping, it would be cool if the resource-icon-in-parens could be a smaller superscript kinda thing.` - cool but not a priority

All other issues should be resolved. Feel free to open another issue or update this one, once you confirm the changes.

### DIFF
--- a/src/cards/base/CommercialDistrict.ts
+++ b/src/cards/base/CommercialDistrict.ts
@@ -29,7 +29,7 @@ export class CommercialDistrict extends Card implements IProjectCard {
         }).nbsp.nbsp.tile(TileType.COMMERCIAL_DISTRICT, true).br;
         b.vpText('1 VP per adjacent city tile.');
       }),
-      victoryPoints: CardRenderDynamicVictoryPoints.cities(1, 1),
+      victoryPoints: CardRenderDynamicVictoryPoints.cities(1, 1, true),
     },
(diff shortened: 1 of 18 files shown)


kberg (approved):

kevinb9n:
All good. Just to mention:

> * `Arcadian Communities: missing "Effect" banner (space?).` - Again, a special case is needed just for one card

(well, Stormcraft would want it too, it's just that in its case the print card has the same bug. again, fine to not change.)

PR #2578, opened 2021-01-28
Two Moon Preludes. And a refactoring of the MoonCardManifest imports. (And one file that leaked in.)

kberg:


src/moon/MoonCards.ts
```diff
@@ -0,0 +1,14 @@
+import {CardName} from '../CardName';
+
+export class MoonCards {
+  public static scienceCardsWith2VPorLess: Array<CardName> = [
```
bafolts, line 4: I would consider using `Set<CardName>`. It can be iterated and would allow for efficient retrieval.

bafolts (approved):

src/moon/MoonCards.ts
kberg: Good idea.

PR #2579, opened 2021-01-28
Remove superfluous uses of 'as' casting.

kberg:


src/cards/promo/Splice.ts
```diff
@@ -48,7 +47,15 @@ export class Splice extends Card implements CorporationCard {
     return undefined;
   }
 
-  public onCardPlayed(player: Player, card: IProjectCard | CorporationCard) {
+  public onCardPlayed(player: Player, card: IProjectCard) {
+    returnthis._onCardPlayed(player, card);
```
bafolts, line 51: This should be `return this`.

bafolts (approved):
Save "as" for similes.

src/cards/promo/Splice.ts
kberg: Done. Gah. Feel free to make an edit yourself next time if you prefer.

kberg:
> Save "as" for similes.

chuckle

PR #2581, opened 2021-01-28
Remove more unnecessary 'as' casting.

kberg:


chosta (approved):

kberg:
Huh, I think I need to restart the staus checks?

src/components/LogPanel.ts
```diff
@@ -74,19 +74,20 @@ export const LogPanel = Vue.component('log-panel', {
             }
           }
         } else if (data.type === LogMessageDataType.CARD) {
+          const cardName = data.value as CardName;
```
bafolts, line 77: This `CardName` needs to be imported for this file.

bafolts: @kberg if this file imports `CardName` the build will pass.

PR #2592, opened 2021-01-29
 Add Luna Project Office card

kberg:
Has a nice pair of tests that go through the draft phase, so that's helpful.


kberg:
Interesting! A flaky test. I'm on it.

kberg:
Figured out what was making this test fail.

src/cards/moon/LunaProjectOffice.ts
```diff
@@ -0,0 +1,46 @@
...
+import {Player} from '../../Player';
+import {CardType} from '../CardType';
+import {IProjectCard} from '../IProjectCard';
+import {Tags} from '../Tags';
+import {CardMetadata} from '../CardMetadata';
+import {CardRenderer} from '../render/CardRenderer';
+import {Cards} from '../Cards';
+import {MoonExpansion} from '../../moon/MoonExpansion';
+
+export class LunaProjectOffice implements IProjectCard {
```
sienmich, line 11: I think the new cards should be made static. `export class LunaProjectOffice extends Card implements IProjectCard`

kberg: Oh that was an oversight on my part. Thanks.

sienmich (approved):

PR #2595, opened 2021-01-29
Static cards - Colonies A-L

sienmich:


src/cards/colonies/AirRaid.ts
```diff
@@ -6,15 +6,27 @@ import {ResourceType} from '../../ResourceType';
...
-import {CardMetadata} from '../CardMetadata';
-import {CardRenderer} from '../render/CardRenderer';
+import {Card} from '../Card';
 import {CardRenderItemSize} from '../render/CardRenderItemSize';
+import {CardRenderer} from '../render/CardRenderer';
+
+export class AirRaid extends Card implements IProjectCard {
+  constructor() {
+    super({
+      cost: 0,
```
kberg, line 16: This is fine, but please take note ongoing that all the other cards are following an ordering of name, type, tags, cost.

kberg (approved):

bafolts (approved):

src/cards/colonies/AirRaid.ts
sienmich: Ok. I just mastered regexes, so If we will need to change this let me know

PR #2597, opened 2021-01-29
Static cards colonies M-Z

sienmich:
Second part of colonies static cards - https://github.com/bafolts/terraforming-mars/pull/2595


kberg (approved):

kberg:
Thank you! This is going to reduce memory consumption on Heroku instances.

PR #2601, opened 2021-01-29
Fix LogBuilder.tileType

### LINKED ISSUE #2591: "to place a 2"
image

daniel-ziegler:
Fixes  #2591

### DIFF
--- a/src/LogBuilder.ts
+++ b/src/LogBuilder.ts
@@ -87,7 +87,7 @@ export class LogBuilder {
     }
 
     public tileType(value: TileType): LogBuilder {
-      this.parameters.push(new LogMessageData(LogMessageDataType.TILE_TYPE, value.toString()));
+      this.parameters.push(new LogMessageData(LogMessageDataType.TILE_TYPE, TileType.toString(value)));
       return this;
     }
 


sienmich (approved):

src/LogBuilder.ts
```diff
@@ -87,7 +87,7 @@ export class LogBuilder {
     }
 
     public tileType(value: TileType): LogBuilder {
-      this.parameters.push(new LogMessageData(LogMessageDataType.TILE_TYPE, value.toString()));
+      this.parameters.push(new LogMessageData(LogMessageDataType.TILE_TYPE, TileType.toString(value)));
```
kberg, line 90: I appreciate that this alleviates the problem, but I would much prefer change be made on the client-side which has accesse to TileType.toString.

This change is not backward compatible, and doesn't fix any past games. The other path would have fixed all prior games.

Can you please switch to resolving this on the client side?

daniel-ziegler: Ah, I see how the system is meant to work. Will do.

kberg: :pray: thank you.

PR #2602, opened 2021-01-29
fix trailing delegate below red and yellow

vsrisuknimit:
image

### DIFF
--- a/src/styles/turmoil.less
+++ b/src/styles/turmoil.less
@@ -323,7 +323,7 @@
 
 .player-token {
     width: 50px;
-    height: 60px;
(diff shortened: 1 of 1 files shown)


kberg:
I'd be more than happy to approve and merge this, but I just don't know what the 'before' is. I'm not particularly turmoil-savvy.

vsrisuknimit:
@kberg Current height for the delegate class is too high. It overflows to cover the next element on the big Turmoil asset.

image


image

kberg:
Good enoughf or me!

kberg (approved):

PR #2604, opened 2021-01-29
add no negative turmoil event to setup details

vsrisuknimit:
image

### DIFF
--- a/src/components/GameSetupDetail.ts
+++ b/src/components/GameSetupDetail.ts
@@ -88,6 +88,11 @@ export const GameSetupDetail = Vue.component('game-setup-detail', {
             </li>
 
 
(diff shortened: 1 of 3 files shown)


kberg:
Am I mistaken or are we basically trying to avoid these kinds of changes where the solution is a configurable list of cards?

vsrisuknimit:
@kberg you cannot adjust turmoil event. This is one of the option for Turmoil.
image

kberg:
> @kberg you cannot adjust turmoil event. This is one of the option for Turmoil.
> image

Oh yeah of course. My bad.

kberg (approved):
Of course. Thank you. Sorry I misread it.

PR #2605, opened 2021-01-30
Resolves #2587 - Adds a configurable number of history lines in game log via .env variable

### LINKED ISSUE #2587: Allow configurable number of log lines?
The current log buffer is 50 lines. On my server, I run with 500 lines which is great for longer games that take several days. It should be fairly straightforward to add a .env variable for this to allow this to be configurable.

puuj:
This commit allows a variable number of lines of history in the game log. The number defaults to 50 to maintain the previous behavior, but can be configured using the new LOG_LENGTH variable noted in the updated README.
I used the same pattern as PostgreSQL.ts (for purge days) to define and parse the number of lines.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -29,3 +29,6 @@
 
 ## When non-empty, stores game states in JSON files. Good for local development and debugging, bad for hosting lots of games.
 # LOCAL_FS_DB=
+
+## Specifies the number of lines of history to be shown in the game log (default 50)
+# LOG_LENGTH=
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ A [.env](https://www.npmjs.com/package/dotenv) file allows you to store environm
 * `ASSET_CACHE_MAX_AGE`: (default 0) How many seconds should assets (fonts, stylesheets, images) be cached by browsers
 * `SERVER_ID`: (default random) Static pass phrase to restrict access to /games-overview and /api/games endpoints
 * `LOCAL_FS_DB`: When non-empty, stores game states in JSON files. Good for local development and debugging, bad for hosting lots of games.
+* `LOG_LENGTH`: (default 50) The number of lines of history shown in the game log 
 
 A `.env.sample` file can be used as a template. You can rename it to `.env` and it will take effect in your environment. Note that `.env` is ignored in source control.
 
--- a/make_static_json.js
+++ b/make_static_json.js
@@ -58,13 +58,21 @@ function getWaitingForTimeout() {
     return 5000;
 }
 
+function getLogLength() {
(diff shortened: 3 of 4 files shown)


kberg:
I'm good with this, but I am going to defer to @chosta who had a minor concern.

bafolts:
Looks good to me. I am glad we are finally using this data! This api route has been waiting to send back more data.

Lynesth:
Wouldn't it make more sense to process it like we do for WAITING_FOR_TIMEOUT, and save the value in settings.json instead of checking the process variable each time the component is mounted ?

bafolts (changes requested):
I believe @Lynesth is right. This compiles because we have the `node` type in our `tsconfig` but I wouldn't think it works. The `process` wont exist client side. This should throw an error that `process` is `undefined` whenever the `LogPanel` were mounted. This change has to use the strategy @Lynesth suggests.

Lynesth:
> This should throw an error that process is undefined

I think it just returns an empty `process`, no error, if I recall correctly.

kberg:
> Wouldn't it make more sense to process it like we do for WAITING_FOR_TIMEOUT, and save the value in settings.json instead of checking the environment variable each time the component is mounted ?
> 
> Edit: check `make_static_json.js` to see what I'm talking about.

@puuj how did you test this?

puuj:
Sorry. I tested this in a haphazard way and didn't realize the issue. I've updated based on the suggestion to use the static JSON file to capture the log length setting, and tested more thoroughly this time.

> > Wouldn't it make more sense to process it like we do for WAITING_FOR_TIMEOUT, and save the value in settings.json instead of checking the environment variable each time the component is mounted ?
> > Edit: check `make_static_json.js` to see what I'm talking about.
> 
> @puuj how did you test this?

kberg:
Still looks good. Hanging back on this though.

Lynesth:
LGTM

chosta (approved):

puuj:
@bafolts
Since this is stuck in "Changes requested" here's an update:
I made the updates to use settings.json rather than the node process. 

I also tested more thoroughly by getting a game with a long log, and loading it into a fresh DB and a newly rebuilt instance, counting the lines, building again with a lower log length number and running the test again.

Note that since the API endpoint for the log does a destructive log.splice() and stores the truncated log, changing the length parameter from a smaller number to a larger number will still only show the smaller number of lines until more actions are taken.

bafolts (approved):

bafolts:
Sorry. Didn't realize my early request changes was the only thing blocking this. This looks good!

PR #2607, opened 2021-01-30
use window event listener for keyboard navigation

bafolts:
Uses `window.addEventListener` for listening to keyboard events for navigation. This will avoid needing to have any element focused in the DOM.


src/components/PlayerHome.ts
```diff
@@ -84,22 +84,21 @@ export const PlayerHome = Vue.component('player-home', {
         let id: string | undefined = undefined;
         switch (event.code) {
         case KeyboardNavigation.GAMEBOARD:
-          id = 'shortkey-board';
+          id = 'board';
```
kberg, line 87: This makes the id's variable name not what it is. It's not the id, it's the id's suffix.

kberg (commented):
But otherwise I'm sure this is fine.

kberg (approved):

PR #2614, opened 2021-01-30
Add card Lunar Mine Urbanization.

kberg:
This also adds all the requirements entries for moon tile counts.


src/cards/CardRequirement.ts
```diff
@@ -159,8 +159,10 @@ export class CardRequirement {
     case RequirementType.TAG:
     case RequirementType.PARTY:
     case RequirementType.PRODUCTION:
-      throw 'Should use subclass satisfies()';
+      throw `Use subclass satisfies() for requirement type ${this.type}`;
     }
+
+    return false;
```
bafolts, line 165: The build seems to pass. I am surprised that one of two things did not happen here. If this was unreachable I am surprised typescript doesn't tell us this is unreachable. If it is reachable I am surprised typescript allowed us to have a code path which returned `undefined` and not a `boolean`.

bafolts: Ah ha!

This code is not reachable. Looks like we are missing a tsconfig option. Can you add `allowUnreachableCode: false`. With that on I see the following error with this addition:

```
src/cards/CardRequirement.ts:164:5 - error TS7027: Unreachable code detected.

164     return false;
        ~~~~~~~~~~~~~


Found 1 error.
```

Lynesth: Also, shouldn't
```ts
    case RequirementType.TAG:
    case RequirementType.PARTY:
    case RequirementType.PRODUCTION:
```
Just be 
```ts
    default:
```

bafolts (approved):
Can you add `allowUnreachableCode: false` to `tsconfig.json`? I figured we had that rule enabled.

src/cards/CardRequirement.ts
kberg: > Also, shouldn't
> 
> ```ts
>     case RequirementType.TAG:
>     case RequirementType.PARTY:
>     case RequirementType.PRODUCTION:
> ```
> 
> Just be
> 
> ```ts
>     default:
> ```

Well that's up to @chosta I suppose?

kberg:
> Can you add `allowUnreachableCode: false` to `tsconfig.json`? I figured we had that rule enabled.

In a follow-up if you do not mind.

kberg:
> > Can you add `allowUnreachableCode: false` to `tsconfig.json`? I figured we had that rule enabled.
> 
> In a follow-up if you do not mind.

Actually I'm doing it now.

kberg:
> > > Can you add `allowUnreachableCode: false` to `tsconfig.json`? I figured we had that rule enabled.
> > 
> > 
> > In a follow-up if you do not mind.
> 
> Actually I'm doing it now.

I added it, and nothing broke. Are you certain that false is unnecessary? Anyway, I'm adding that in a new place.

PR #2616, opened 2021-01-30
Moon UI fixes

kberg:


src/cards/moon/PrideoftheEarthArkship.ts
```diff
@@ -24,7 +24,7 @@ export class PrideoftheEarthArkship extends MoonCard implements IActionCard {
         renderData: CardRenderer.builder((b) => {
           b.action('Add 1 science resource here per every 5 science tags you have.', (eb) => {
             eb.empty().startAction.science(1).slash().text('5').science(); // TODO(kberg): these should be tags, not resources.
-          }),
+          }).br,
```
Lynesth, line 27: Never really looked in the card renderer before.
Why is there a comma here, instead of a semi colon ?

Also in the file below you did it like this: ` b.br;`

src/cards/moon/HypersensitiveSiliconChipFactory.ts
```diff
@@ -24,6 +24,7 @@ export class HypersensitiveSiliconChipFactory extends MoonCard {
         renderData: CardRenderer.builder((b) => {
           b.minus().titanium(2);
           b.production((pb) => pb.megacredits(4));
+          b.projectRequirements().moonMine().moonMine();
```
Lynesth, line 27: MOON_MINE tiles instead ?

Lynesth (approved):

src/cards/moon/HypersensitiveSiliconChipFactory.ts
kberg: Yeah, this is such a mess. Thank you for pointing this out.

src/cards/moon/PrideoftheEarthArkship.ts
kberg: Oops! Should be a semicolon.

kberg: Thanks, fixed.

Issue #2585, opened 2021-01-28
Extend timers to count up to 999 hours?


puuj:
Currently timers will wrap around after hitting 24 hours. This isn't a big deal, but sometimes it's interesting to see how long a player has taken cumulatively. To make timers a bit more transparent, I made a personal mod to use 3 digits for the hours. If this is something others would appreciate, I'd be happy to contribute.
![image](https://user-images.githubusercontent.com/58322/106200284-9985a680-616b-11eb-8dc4-9aec2b4189e4.png)

chosta:
We are actually trying to reduce the space for timers by but this makes sense for games that are played based on availability. The best solution imo is to add days to if it passes the 24h mark. What you currently show is an outdated version of the timer, please get your server updated to check the new design.

vsrisuknimit:
Timer does not make any sense for async games anyway and probably should be turned off. What does it mean if someone (Player A) gets 8 hr added to their timer just because the previous player took an action at midnight when player A already went to bed?

puuj:
It means we get to make fun of them for not being sufficiently dedicated to the game. It's all in fun for now. 
Although if you're interested, I have lots of ideas about how timers could include, for example, a schedule of when timers are running (e.g., M-W, 6p-9p) that would sidestep this issue. Probably overkill, but...

> Timer does not make any sense for async games anyway and probably should be turned off. What does it mean if someone (Player A) gets 8 hr added to their timer just because the previous player took an action at midnight when player A already went to bed?

puuj:
Sure, I update roughly weekly but I checked out the current Heroku implementation. Since the change is simply in how the time string is returned, I don't think it will alter anything in the code. 

Technically, 999 hours will cover ~40 days with one digit (and no separator) whereas supporting that number of days requires adding 2 digits and a separator, so more space. Days might be more user friendly, though. 
 
> We are actually trying to reduce the space for timers by but this makes sense for games that are played based on availability. The best solution imo is to add days to if it passes the 24h mark. What you currently show is an outdated version of the timer, please get your server updated to check the new design.

kberg:
> Timer does not make any sense for async games anyway and probably should be turned off. What does it mean if someone (Player A) gets 8 hr added to their timer just because the previous player took an action at midnight when player A already went to bed?

I've had async games where only until after it was done that on average 3 of the four players took ~8 hours delay, and one took 20 days. It was really useful.

vsrisuknimit:
[...]

PR #2619, opened 2021-01-30
Timer expand

puuj:
This update changes the timer behavior so it does not wrap after 24 hours. Instead, additional digits will be shown. Additionally, hours will not be shown until at least one hour has elapsed.

### DIFF
--- a/src/Timer.ts
+++ b/src/Timer.ts
@@ -57,12 +57,15 @@ export class Timer implements ISerializable<SerializedTimer> {
     this.sumElapsed += Timer.lastStoppedAt - this.startedAt;
   }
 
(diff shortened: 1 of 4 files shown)


vsrisuknimit:
@puuj Please change the test in build/tests/Timer.spec.js to match the new behavior. You can run 'npm run test' to run the test before submitting the PR. A screenshot of the new behavior would be nice.

vsrisuknimit:
#2585

puuj:
Updated the tests to include with and without hours. Here are the screenshots.
image
image

sienmich (approved):

PR #2620, opened 2021-01-31
Requirements in card

sienmich:
I moved the `card.metadata.requirements` to `card.requirements`.

This PR is in two commits.
1. All important files that you should check out.
2. Updating all the cards, just moving the requirements out of metadata. Was made in 95% automatically with some regexes. 

Why?
The `CardRequirements` are now not only the UI. They can check if the requirements are satisfied with `requirements.satisfies(player): boolean`. In next PR i'd like to update the rest of `card.canPlay()` with this `.satisfies`.


chosta (approved):
Looks good.

PR #2623, opened 2021-01-31
Database admin changes

kberg:
This should be read as 3 separate commits, as they build on each other. The last one is still broken, though.

In order, read these: 

1. https://github.com/bafolts/terraforming-mars/pull/2623/commits/8c69ab5e440b774f2d1aff1065b668442682bfc8
2. https://github.com/bafolts/terraforming-mars/pull/2623/commits/e004b04e49485b50a10510e4f9c18b9c6b585cb9
3. https://github.com/bafolts/terraforming-mars/pull/2623/commits/2113fa19845b035401b4326509a85948e6c182dc


kberg:
Ugh, this is 11 commits. Somehow I lost the smooth 3-commit setup I had planned.

README.md
```diff
@@ -106,7 +106,7 @@ Thanks goes to these wonderful people:
   <tr>
     <td align="center">
       <a href="https://github.com/kberg"><img src="https://avatars1.githubusercontent.com/u/413481?s=460&v=4" width="100px;" alt=""/><br />
-        <sub><b>Robert Konigsberg</b></sub><br />Ares fan expansion, infrastructure cleanup, code reviews, two opinions too many.</a> 
+        <sub><b>Robert Konigsberg</b></sub><br />Ares fan expansion, The Moon fan expansion, infrastructure cleanup, code reviews, two opinions too many.</a>
```
bafolts, line 109: This change should go in ASAP. A separate PR would help it along.

package.json
```diff
@@ -57,7 +57,6 @@
     "templates": "node compile-vue-templates.js && tsc --build tsconfig-vue.json",
     "test": "mocha --file build/tests/utils/Vue.js --recursive build/tests",
     "pretest": "npm run prebuild && tsc --build tests/tsconfig.json",
-    "testgiven": "tsc --build tsconfig.json && npm run webpack && tsc --build tests/tsconfig.json && mocha",
```
bafolts, line 60: Removing this makes sense to me. Not sure how it works.

kberg: It doesn't. alrudsi added it but it turned out not to work.

kberg:
@bafolts I'm mostly curious about your thoughts around the three database PRs.

src/server/DatabaseMaintenance.ts
```diff
@@ -0,0 +1,14 @@
...
+export class DatabaseMaintenance {
+  private static task() {
+    console.log('Starting database maintenance');
+    Database.getInstance().purgeUnfinishedGames();
+    console.log('Finished database maintenance');
+  }
+
+  public static initialize() {
+    const fifteenMinutesInMilliseconds = 15 * 60 * 1000;
+    setInterval(() => DatabaseMaintenance.task(), fifteenMinutesInMilliseconds);
```
bafolts, line 12: I would store this returned interval id. On subsequent calls to `DatabaseMaintenance.initialize` could then ignore them if this has already started or call `clearInterval` on any stored id and start a new `setInterval`.

src/database/SQLite.ts
```diff
@@ -104,20 +118,43 @@ export class SQLite implements IDatabase {
...
+      if (gameIds === undefined) {
+        return console.warn('games undefined in purge');
+      }
+      gameIds.forEach((gameId) => {
+        this.purgeFinishedGame(gameId);
+      });
+    });
+  }
+
+  purgeFinishedGame(gameId: GameId): void {
```
bafolts, line 143: There is a `purged` status that doesn't seem to be used. Is the `purged` status going to be used?

bafolts:
This change makes sense to me. The process to purge games now relies on someone restarting the node server. I happen to do this once a day. Users running the server that weren't running this once per day could end up with games collecting in the database.

kberg:
> The process to purge games now relies on someone restarting the node
server.
You sure? It's supposed to purge every fifteen minutes.

bafolts:
By "now" I meant prior to this change. After this change it would run every fifteen minutes.

kberg:
I added an implementation for the Postgres instance, but I need to locally test it.

kberg:
I'm going to close this out and start it fresh.

PR #2625, opened 2021-01-31
Debug UI tweaks

alrusdi:
1) Delay page refresh when user is typing in search box
2) Ability to pass search string parameter via url like this: `/debug-ui?search=ai central`

### DIFF
--- a/src/components/DebugUI.ts
+++ b/src/components/DebugUI.ts
@@ -46,6 +46,9 @@ export interface DebugUIModel {
   moon: boolean | unknown[],
   promo: boolean | unknown[],
 }
+
+let throttleTimeout = 0;
+
 export const DebugUI = Vue.component('debug-ui', {
   components: {
     Card,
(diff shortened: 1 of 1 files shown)


kberg (approved):
Thanks for both of these.

kevinb9n:
Hey @alrusdi, it's now happening that if I type a little too fast in the search box, some letters are getting erased/overtyped. Could that have been introduced with this change?

bafolts:
This is triggering only with the onkeyup event. If you were to type and then hold down a key I would think you could create the problem @kevinb9n sees. Using the onchange event may prove more reliable.

kberg:
That makes a lot of sense. onchange also supports behavior like pasting
into the text box.

>

alrusdi:
> Hey @alrusdi, it's now happening that if I type a little too fast in the search box, some letters are getting erased/overtyped. Could that have been introduced with this change?

Shurely it is. I will try to fix it ASAP

PR #2626, opened 2021-01-31
add button to hide tiles

### LINKED ISSUE #2050: Feature request: a button to hide tiles
A button next to the main game board. Once toggle, it makes all tiles 25% opacity so that users can look at what are underneath.

vsrisuknimit:
Normal image

Once click 'hide tiles'
image

It does not persist. As soon as you refresh or other player take an action, the view mode goes back to default. I think this is what we want.

Question: Not sure what the button name should be. 'hide tiles'? 'show bonus'?
Question: Should we show player cube during hide tiles?

Still have to adjust the position so it does show over the card during initial research phase.

This PR can close #2050 .

### DIFF
--- a/src/components/Board.ts
+++ b/src/components/Board.ts
@@ -59,6 +59,7 @@ export const Board = Vue.component('board', {
   data: function() {
     return {
       'constants': constants,
+      'isTileHidden': false,
     };
   },
   mounted: function() {
@@ -157,9 +158,16 @@ export const Board = Vue.component('board', {
     getGameBoardClassName: function():string {
       return this.venusNextExtension ? 'board-cont board-with-venus' : 'board-cont board-without-venus';
     },
+    hideTile: function() {
+      return this.isTileHidden = !this.isTileHidden;
+    },
+    checkHideTile: function():boolean {
+      return this.isTileHidden;
+    },
   },
   template: `
     <div :class="getGameBoardClassName()">
+        <div class="hide-tile-button" v-on:click.prevent="hideTile()">Hide tiles</div>
         <div class="board-outer-spaces">
             <board-space :space="getSpaceById('01')" text="Ganymede Colony"></board-space>
             <board-space :space="getSpaceById('02')" text="Phobos Space Haven"></board-space>
@@ -207,7 +215,7 @@ export const Board = Vue.component('board', {
         </div>
 
         <div class="board" id="main_board">
-            <board-space :space="curSpace" :is_selectable="true" :key="'board-space-'+curSpace.id" :aresExtension="aresExtension" v-for="curSpace in getAllSpacesOnMars()"></board-space>
+            <board-space :space="curSpace" :is_selectable="true" :key="'board-space-'+curSpace.id" :aresExtension="aresExtension" :isTileHidden="checkHideTile()" v-for="curSpace in getAllSpacesOnMars()"></board-space>
             <svg id="board_legend" height="550" width="630" class="board-legend">
                 <g v-if="boardName === 'tharsis'" id="ascraeus_mons" transform="translate(95, 192)">
                     <text class="board-caption">
--- a/src/components/Board.ts
+++ b/src/components/Board.ts
@@ -158,8 +158,11 @@ export const Board = Vue.component('board', {
     getGameBoardClassName: function():string {
       return this.venusNextExtension ? 'board-cont board-with-venus' : 'board-cont board-without-venus';
     },
-    hideTile: function() {
-      return this.isTileHidden = !this.isTileHidden;
+    toggleHideTile: function() {
+      this.isTileHidden = !this.isTileHidden;
+    },
+    toggleHideTileLabel: function(): string {
+      return this.isTileHidden ? 'show tiles' : 'hide tiles';
     },
     checkHideTile: function():boolean {
       return this.isTileHidden;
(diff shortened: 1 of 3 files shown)


vsrisuknimit:
@chosta help. I need the position of the button to be absolute so it does shift the terraforming tracks. But during initial draft, absolute just layover cards. What can I do?

image

src/components/Board.ts
```diff
@@ -157,9 +158,16 @@ export const Board = Vue.component('board', {
...
+    hideTile: function() {
+      return this.isTileHidden = !this.isTileHidden;
+    },
+    checkHideTile: function():boolean {
+      return this.isTileHidden;
+    },
   },
   template: `
     <div :class="getGameBoardClassName()">
+        <div class="hide-tile-button" v-on:click.prevent="hideTile()">Hide tiles</div>
```
kberg, line 170: "show map"?

kberg (commented):
IMO an excellent bit of work. While there are more elegant solutions, this is one of those cases where the perfect would be the enemy of the good.

How would you envision this applying to The Moon simultaneously? I wouldn't worry about this just yet, though, but I would like your opinion.

src/components/Board.ts
chosta: "show tiles"?

src/components/Board.ts
```diff
@@ -157,9 +158,16 @@ export const Board = Vue.component('board', {
     getGameBoardClassName: function():string {
       return this.venusNextExtension ? 'board-cont board-with-venus' : 'board-cont board-without-venus';
     },
+    hideTile: function() {
```
chosta, line 161: Devs like the word `toggle` in such cases.

src/components/Board.ts
```diff
@@ -157,9 +158,16 @@ export const Board = Vue.component('board', {
...
+      return this.isTileHidden = !this.isTileHidden;
```
chosta, line 162: I'd prefer to have this in two lines so we separate the concerns.
1) Assignment
2) return statement

chosta (changes requested):
If you want to have something that is positioned absolutely relative to the parent container, just add "position: relative" (to the parent)

chosta:
Also, you have to toggle the text too - Hide tiles should become Show tiles and vice versa (upon click)

src/styles/board.less
```diff
@@ -284,3 +284,18 @@
     margin-bottom: -1px;
 }
 
+.board-hidden-tile{
+	opacity: 0.6;
+}
+
+.hide-tile-button {
+	position: absolute;
+	top: 610px;
```
chosta, line 293: Better to offset it from "bottom"

src/styles/board.less
```diff
@@ -284,3 +284,18 @@
...
+	text-transform: uppercase;
+	border-radius: 4px;
+	font-size: 12px;
```
chosta, line 296: Please, use a variable for the font size from "variables.less"

src/styles/board.less
```diff
@@ -284,3 +284,18 @@
     margin-bottom: -1px;
 }
 
+.board-hidden-tile{
+	opacity: 0.6;
+}
+
+.hide-tile-button {
```
chosta, line 291: Please, consider reusing a class that has been used previously. Check out the buttons for "concise" and "full" view in the player board.

src/components/Board.ts
vsrisuknimit: @chosta if I use relative, all the global parameters will be pushed aside.

src/components/Board.ts
vsrisuknimit: I guess it does not need to return anything. my bad.

src/styles/board.less
vsrisuknimit: That class is inside another class. I cannot use it from here.

vsrisuknimit:
Before
image

After
image

src/styles/board.less
vsrisuknimit: use font small now

src/styles/board.less
vsrisuknimit: bottom now. with pseudoelement to be relative.

src/components/Board.ts
vsrisuknimit: changed the function name

src/components/Board.ts
vsrisuknimit: now show tiles and hide tile

vsrisuknimit:
> How would you envision this applying to The Moon simultaneously? I wouldn't worry about this just yet, though, but I would like your opinion.

@kberg I imagine the same button will hide tiles on the Moon map as well. It should be pretty easy to add.

vsrisuknimit:
Do we want to hide player cube as well?

image

@chosta I would like to apply translucent only to the tile but not the bonuses. Is there anyway to do it?

src/styles/board.less
chosta: My point was to use the style of an existing UI element for consistency. Plase, consider:
`    text-transform: uppercase;
    border-radius: 4px;
    color: black;
    background: #ccc;`

chosta (approved):
I don't think there is a way (not an easy one at least) to have an element without opacity inside another that is opaque. Just leave it as it is, it does the job and it's just fine.
Check my comment about styling the button before merge.

src/styles/board.less
vsrisuknimit: I have all those styles here as well except the background color. I don't want this button to be too bright next to the main map.

vsrisuknimit:
Two changes:
- Add transparency only to tiles, not bonuses
- Remove player cubes when hiding tiles

Before:
image

After:
image

chosta (approved):
Yeah, if it's inside a container you can stack opacities. Good work.

PR #2627, opened 2021-01-31
rename CEOS_FAVORITE_PROJECT, RAD_CHEM_FACTORY, and TITAN_FLOATING_LAUNCHPAD

bafolts:
Renamed three cards. Ensures that these games can be loaded from database on server side. After 10 days the old card names should be purged from database and all clients will most likely have pulled latest main.js. Can remove this mapping at that time.


src/Deck.ts
```diff
@@ -2,13 +2,24 @@ import {CardName} from './CardName';
 import {ICard} from './cards/ICard';
 import {ICardFactory} from './cards/ICardFactory';
 
+// TODO(bafolts): remove after 02/15/2021 before 03/01/2021
```
kberg, line 5: Could keep the structure around, for the next time this occurs.

kberg (approved):

src/Deck.ts
bafolts: I will put this comment per card it refers to.

PR #2629, opened 2021-01-31
use one image for planets

bafolts:
We make 9 separate web requests for images. Browsers are limited to the number of simultaneous network requests they can make. The usage of png leads to images that are far larger than needed. This change puts all of the planets in one jpg. This should speed up the load time of the front page and take load off the server.

Before:

```
-rw-r--r--  1 bfolts  753692654    56K Jan 18 11:11 planet1.png
-rw-r--r--  1 bfolts  753692654    85K Jan 18 11:11 planet2.png
-rw-r--r--  1 bfolts  753692654    93K Jan 18 11:11 planet3.png
-rw-r--r--  1 bfolts  753692654    92K Jan 18 11:11 planet4.png
-rw-r--r--  1 bfolts  753692654    68K Jan 18 11:11 planet5.png
-rw-r--r--  1 bfolts  753692654    49K Jan 18 11:11 planet6.png
-rw-r--r--  1 bfolts  753692654    22K Jan 18 11:11 planet7.png
-rw-r--r--  1 bfolts  753692654    77K Jan 18 11:11 planet8.png
-rw-r--r--  1 bfolts  753692654    63K Jan 18 11:11 planet9.png
```

After:

```
-rw-rw-r-- 1 algorithm99 algorithm99 89K Jan 31 15:05 planets.jpg
```


vsrisuknimit:
@bafolts how does the new home page now look with one image file?

Lynesth:
@vsrisuknimit It should still look the same. Just one big image and each menu entry is showing only specific part of it.

bafolts:
@vsrisuknimit it shouldn't look any different.

image

What should be noticed is how much faster the start screen loads. Instead of 9 separate requests of around 80kb there will be one request of 80kb. The images are also no longer scaled client side so they will render faster. When you request a website all of the assets are not downloading simultaneously. The browser will only run so many requests simultaneously. With out 9 image requests we are blocking loading of other resources as well.

https://stackoverflow.com/questions/7456325/get-number-of-concurrent-requests-by-browser

kberg (approved):

Issue #2631, opened 2021-02-01
Major Bug: Dirigibles is not taken into account for playing Venus cards


vsrisuknimit:
![image](https://user-images.githubusercontent.com/14239220/106404356-88a19300-6400-11eb-943b-8b721d24a915.png)

In this case, the player should be able to play Venus Governor by spending 1 mc and 1 floater from Dirigibles. It does not work.

vsrisuknimit:
This is a problem for all Venus cards and Dirigibles.

![image](https://user-images.githubusercontent.com/14239220/106404454-e0d89500-6400-11eb-896b-fa08243e9ec5.png)

vsrisuknimit:
This is a new bug. I believe it used to work fine before. Did someone change canPlay or canAfford?

vsrisuknimit:
Not just related to floater. This problem affects microbe canPay as well.

![image](https://user-images.githubusercontent.com/14239220/106405473-4ed28b80-6404-11eb-9c8d-f548fc2a817f.png)


In this case, you should be able to play Greenhouses with just 3 microbes from Psychophiles.

PR #2632, opened 2021-02-01
Fix argument order for canAfford

Lynesth:


Lynesth:
@bafolts This is currently "game breaking".
Windows test failed for some reason though.

vsrisuknimit:
#2631

nwai90 (approved):

kberg:
Given that this fixes a big bug, something deserves a unit test somewhere.

Lynesth:
Actually it would be simple tests with different cards and different resources and then just check that player.canPlay(card) returns the correct value.

Will try to do that today.

PR #2634, opened 2021-02-01
Fix Steel Market Monopolists and Titanium Market Monopolists, which were kind of broken.

kberg:


src/cards/moon/SteelMarketMonopolists.ts
```diff
@@ -9,7 +9,7 @@ export class SteelMarketMonopolists extends MarketCard {
   constructor() {
     super(
       Resources.STEEL,
-      {from: 3, to: 2, limit: 6},
+      {from: 3, to: 2, limit: 3},
```
kberg, line 12: This was just straight-up incorrect.

src/cards/moon/MarketCard.ts
```diff
@@ -98,17 +97,15 @@ export abstract class MarketCard extends Card implements IActionCard {
...
     }
     let limit = player.getResource(this.tradeResource);
-    limit = terms.limit === undefined ? limit : Math.max(limit, terms.limit);
-
-    // TODO(kberg): the messages and UI need work.
+    limit = terms.limit === undefined ? limit : Math.min(limit, terms.limit);
 
     return new SelectAmount(
-      `Select amount of ${this.tradeResource} to sell`,
+      `Select a number of trades (${terms.from} ${this.tradeResource} => ${terms.to} MC, max ${limit})`,
```
kberg, line 103: These text bits are meant to make the UI much simpler to understand.

tests/cards/moon/TitaniumMarketMonopolists.spec.ts
```diff
@@ -47,6 +48,51 @@ describe('TitaniumMarketMonopolists', () => {
     expect(card.canAct(player)).eq(true);
   });
 
-  // TODO(kberg): action tests, which are tricky.
```
kberg, line 50: Well, what have I learned?

src/cards/moon/MarketCard.ts
```diff
@@ -65,23 +65,22 @@ export abstract class MarketCard extends Card implements IActionCard {
...
         if (player.canUseHeatAsMegaCredits) {
           const howToPay = new SelectHowToPayDeferred(player, cashDue, {afterPay: () => {
-            player.setResource(this.tradeResource, tradesRequested);
+            player.setResource(this.tradeResource, unitsEarned);
           }});
           player.game.defer(howToPay);
         } else {
           player.setResource(Resources.MEGACREDITS, -cashDue);
-          player.setResource(this.tradeResource, tradesRequested);
+          player.setResource(this.tradeResource, unitsEarned);
```
Lynesth, line 83: You can probably just defer howToPay in both cases, it will do it automatically if the player cannot pay with anything else than MC.

src/cards/moon/MarketCard.ts
```diff
@@ -98,17 +97,15 @@ export abstract class MarketCard extends Card implements IActionCard {
       throw new Error('selling from !== 1 not yet supported.');
     }
     let limit = player.getResource(this.tradeResource);
-    limit = terms.limit === undefined ? limit : Math.max(limit, terms.limit);
-
-    // TODO(kberg): the messages and UI need work.
+    limit = terms.limit === undefined ? limit : Math.min(limit, terms.limit);
```
Lynesth, line 100: limit cannot be undefined anymore.

Lynesth (commented):
Did not review the tests.

src/cards/moon/MarketCard.ts
kberg: Meaning the heat?

src/cards/moon/MarketCard.ts
kberg: Good catch, done.

src/cards/moon/MarketCard.ts
kberg: Is this OK as a follow-up?

Lynesth: It sure is ok as a follow up.

Lynesth (approved):
Doesn't look like it'd break anything and the logic seems to make sense.

kberg:
> Did not review the tests.

They're the best part! :D

kberg:
Thank you.

PR #2635, opened 2021-02-01
Add Moon card Thorium Rush

kberg:


chosta (approved):

src/cards/moon/ThoriumRush.ts
```diff
@@ -0,0 +1,37 @@
...
+import {CardType} from '../CardType';
+import {IProjectCard} from '../IProjectCard';
+import {Tags} from '../Tags';
+import {CardMetadata} from '../CardMetadata';
+import {CardRenderer} from '../render/CardRenderer';
+import {PlaceMoonColonyTile} from '../../moon/PlaceMoonColonyTile';
+import {PlaceMoonRoadTile} from '../../moon/PlaceMoonRoadTile';
+import {PlaceMoonMineTile} from '../../moon/PlaceMoonMineTile';
+
+export class ThoriumRush implements IProjectCard {
```
sienmich, line 12: Again better to have a static card (`extends Card`), right?

sienmich (changes requested):

src/cards/moon/ThoriumRush.ts
kberg: Thanks for catching these.

The reason  you are seeing these is because the cards are _already written._ I wrote them a month or two ago, and now all I'm doing is bringing them into the main repo and adding tests.

So if you see these ongoing, it's because I forgot to convert them. I don't have to be convinced -- just  reminded. :D

Thanks.

sienmich (approved):

PR #2640, opened 2021-02-01
Improve and "harmonize" onTilePlaced logic

Lynesth:
After a comment by @kberg I thought it would be a good idea to streamline onTilePlaced a bit more.
To do that, I added a 3rd argument to the method and I tried to improve the logic here and there as well.
I also took care of converting Mining Guild to the new priority system, which I missed in the previous PR.

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -41,6 +41,7 @@ import {SelectHowToPayDeferred} from './deferredActions/SelectHowToPayDeferred';
 import {SelectInitialCards} from './inputs/SelectInitialCards';
 import {PlaceOceanTile} from './deferredActions/PlaceOceanTile';
 import {RemoveColonyFromGame} from './deferredActions/RemoveColonyFromGame';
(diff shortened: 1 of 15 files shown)


nwai90 (approved):
Functionally LGTM 👍

src/Game.ts
```diff
@@ -1316,33 +1316,19 @@ export class Game implements ISerializable<SerializedGame> {
       TurmoilHandler.resolveTilePlacementBonuses(player, spaceType);
 
       if (arcadianCommunityBonus) {
-        player.megaCredits += 3;
+        this.defer(new GainResources(player, Resources.MEGACREDITS, {count: 3}));
```
Lynesth, line 1319: I would really like to move this to an Arcadian Community `onTilePlaced` method instead of having it here, but it relies on the fact that there was a player marker on the space before placing the tile, and that's an info we don't have anymore once we hit `onTilePlaced`.

So as to not have to modify things again, just for one card, I left the logic as it was.

PR #2642, opened 2021-02-01
Replaced all `.indexOf(sth) !== -1` with `.includes(sth)`

sienmich:
(and same with `=== -1`)

It makes the code easier to read.


bafolts (approved):
I have been writing javascript since 2005. At one point there was only `indexOf` available and it wasn't available on `String` or `Array` I forget which one. Long story short I used `indexOf` here as I still worry about these old quirks. This change makes sense as it is far easier to read `includes` and less error prone than `indexOf`. This change wont work in IE but our contributing guide and code of conduct reminds users we don't strive to support every old browser. Support for `includes` is available in relatively new browsers so adding this is fine.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

PR #2643, opened 2021-02-01
DeferredActionsQueue

sienmich:
@Lynesth 
The actions queue does not care about the initial order of its elements, only about the priorities. (As pointed out in https://discordapp.com/channels/737945098695999559/742721510376210583/805509673289515008)

I added the test that confirms this behavior.

I see those possible solutions:
1. Have a separate (standard) queue for each of the possible priorities
2. Use a different data structure than `Heap from 'mnemonist'`
3. Add some priority-like counter to each action (first sort by priority, then by counters)

I think the best solution is 1. It's fast (without the logarithmic complexity) and does not require additional packages.
I can gladly implement it if you want.

What do you think?


Lynesth:
Thank you very much for your time !

As for the solution, I'm first going with number 3 as this is the one that requires the less changes for now.
The number of different priorities might change in the upcoming days, as I switch everything to deferred actions.
I'm not dismissing the idea of number 1 though, I think it might be great, but let's check it out once more when I'm done with what I have to do :)

I'll be making another PR with your test and the counter solution.

sienmich:
Ok, sure!

PR #2648, opened 2021-02-01
Add queueId to deferred actions + ordering test

Lynesth:


src/deferredActions/DeferredAction.ts
```diff
@@ -19,6 +19,7 @@ export enum Priority {
 }
 
 export class DeferredAction {
+  public queueId?: number;
```
sienmich, line 22: The '?' is for backward compatibility?

src/deferredActions/DeferredActionsQueue.ts
```diff
@@ -3,15 +3,24 @@ import {GiveColonyBonus} from './GiveColonyBonus';
...
+  private insertId: number = 0;
+
   private queue: Heap<DeferredAction> = new Heap((a, b) => {
-    return (a.priority < b.priority) ? -1 : ((a.priority > b.priority) ? 1 : 0);
+    if (a.priority < b.priority) {
+      return -1;
+    } else if (a.priority > b.priority) {
+      return 1;
+    } else {
+      return (a.queueId! < b.queueId!) ? -1 : 1;
```
Lynesth, line 14: Note: It should be a number at this point, but even if it's not a number, it will just compare with `undefined` and that returns false, so it's fine to `!` them.

src/deferredActions/DeferredActionsQueue.ts
```diff
@@ -3,15 +3,24 @@ import {GiveColonyBonus} from './GiveColonyBonus';
...
+    }
```
sienmich, line 15:
```suggestion
    if (a.priority === b.priority) {
      return a.queueId! - b.queueId!;
    }
    return a.priority - b.priority;
```
A bit simpler

tests/cards/venusNext/SponsoredAcademies.spec.ts
```diff
@@ -32,4 +34,24 @@ describe('SponsoredAcademies', function() {
     expect(player.cardsInHand).has.lengthOf(4);
     expect(player2.cardsInHand).has.lengthOf(1);
   });
+
+  it('triggers in right order', function() {
```
sienmich, line 38: I like those!

sienmich (approved):

src/deferredActions/DeferredAction.ts
Lynesth: If I don't put the `?` here, I have to declare `queueId` in each deferred action that extends DeferredAction.

src/deferredActions/DeferredActionsQueue.ts
Lynesth: I like this.

src/deferredActions/DeferredActionsQueue.ts
kberg: So, why is there a queue id now? Is that somewhere that I'm still not caught up on?

Lynesth: Oh sorry, I thought I linked the previous closed PR: #2643 
It is used to order the deferred actions &mdash; within the same priority &mdash; in the order they were deferred.

If you defer (pseudo code):
```
player1.drawCard();
player2.drawCard();
player3.drawCard();
player4.drawCard();
```
Before that PR, it would return them in any order since they are of the same priority.

While it fundamentally doesn't really matter, in some cases it helps make the order a bit more consistent.

For example, when we're deferring actions that make every player draw a card (eg. Sponsored Academies), users like to see that the cards are drawn in turn order (this is the case that brought us to this PR).

kberg: OK. So, by queue id, it's not the ID of the queue but the ID of a queue entry?

PR #2650, opened 2021-02-02
Add Darkside Mining Syndicate

kberg:


src/cards/moon/DarksideMiningSyndicate.ts
```diff
@@ -0,0 +1,39 @@
...
+      },
+    });
+  };
+
+  public canPlay(): boolean {
+    return true;
+  }
+
+  public play(player: Player) {
+    const productionBonus = (MoonExpansion.moonData(player.game).miningRate >= 2) ? 2 : 1;
```
sienmich, line 34: This if does not match the description

sienmich: (2 and 1 swapped)

bafolts: It does seem there is a discrepancy between what this function is doing and the description. I can't find this card on

https://www.dropbox.com/s/txeluw0z0h8nk36/TM%20-%20the%20Moon.zip?dl=0

To determine which is right. @kberg I could help with adding cards if you have list of cards left to add.

kberg: You were right that my logic was backward. The card text is right.

image

kberg: > It does seem there is a discrepancy between what this function is doing and the description. I can't find this card on
> 
> https://www.dropbox.com/s/txeluw0z0h8nk36/TM%20-%20the%20Moon.zip?dl=0
> 
> To determine which is right. @kberg I could help with adding cards if you have list of cards left to add.

I have begged the designer to update the first page to reference the more complete set of cards, including his recommendation to use  larger (!) board . All of which you can see is referenced at https://github.com/bafolts/terraforming-mars/wiki/The-Moon

kberg: In addition, @bafolts most of the cards have been implemented and are at https://github.com/kberg/terraforming-mars/tree/moon, I'm spending much of my time converting the card format and adding tests. (And other things.)

sienmich (approved):

PR #2652, opened 2021-02-02
Remove Cards.ts

nwai90:


kberg (changes requested):
Why? I put there to be used, not to be removed. Could you explain more, please?

nwai90:
@kberg These are minimally used and for very common Javascript functions. They also seem to be used mainly in tests - in which case they should be in `TestingUtils` if at all imo

kberg:
> @kberg These are minimally used and for very common Javascript functions. They also seem to be used mainly in tests - in which case they should be in `TestingUtils` if at all imo

They're minimally used because I just introduced them, like, a week ago, and they can be used elsewhere.

Lynesth:
I'm not saying we shouldn't use shorthand functions, but I'll have to agree with @nwai90 on this specific one though.

Those 2 are only turning
`(card) => card.name === ###` into `Cards.byCardName(###)`
and
`(card) => card.name` into `Cards.toCardName`

Their initial version is already easily readable/understandable, especially the second one.

For the first one (byCardName), a shorthanded version could be used I assume, but then I would probably do it like this instead:
`(card) => card.is(###)`

kberg:
> I'm not saying we shouldn't use shorthand functions, but I'll have to agree with @nwai90 on this specific one though.
> 
> Those 2 are only turning
> `(card) => card.name === ###` into `Cards.byCardName(###)`
> and
> `(card) => card.name` into `Cards.toCardName`
> 
> Their initial version is already easily readable/understandable, especially the second one.
> 
> For the first one (byCardName), a shorthand version could be used I assume, but then I would probably do it like this instead:
> `(card) => card.is(###)`

Well, OK. @nwai90 mind if I restore the branch and apply the changes you proposed?

nwai90:
Sure go ahead

kberg (approved):

Issue #2328, opened 2021-01-10
This sequence of undo steps seems to be bugged


kevinb9n:
I have seen this behavior several times but haven't determined under exactly what circumstances it happens. It doesn't happen every time.

1. Take some action
2. Undo it
3. Reload
4. Take a different action
5. Undo it
6. Reload
7. The first action I did in step 1 has magically returned.

kevinb9n:
Sorry I still don't have exact steps for guaranteed repro, but FYI @Lynesth I have seen this many many times lately. It might be triggered by doing a chain of two or more undos in a row, perhaps. A rabid undoer, I am.

kevinb9n:
This sequence seems to produce the bug every time in a solo game:

1. Do action A
2. Undo action A
3. Do action B
4. Do action C
5. Undo action C
6. Action A has returned to life

Lynesth:
Right, my dumb ass finally found out why.

kevinb9n:
Sounds like your dumb ass is smart then :-)

PR #2656, opened 2021-02-02
Use UPSERT in saveGame() to overwrite previous saves

Lynesth:
I don't know why I thought it was already the case, and the undo mechanism I wrote expected it to...

Anyway, so this allow overwriting saves when undoing actions, and fixes the [last known undo bug](https://github.com/bafolts/terraforming-mars/issues/2328) (solo games only).

Edit: I'm not sure what to do in the event the save failed. We only assumed that it would be because of a duplicate entry before that, so I only added a console output for now. Please tell me your thoughts @bafolts @kberg

### DIFF
--- a/src/database/PostgreSQL.ts
+++ b/src/database/PostgreSQL.ts
@@ -193,15 +193,17 @@ export class PostgreSQL implements IDatabase {
   }
 
   saveGame(game: Game): void {
+    const gameJSON = game.toJSON();
     this.client.query(
-      'INSERT INTO games(game_id, save_id, game, players) VALUES($1, $2, $3, $4)',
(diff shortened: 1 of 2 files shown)


kberg (commented):
Have you been able to test this works on postgres? (I recently added documentation on how to set up a local copy of postgres.)

kberg:
This is really good. `ON CONFLICT` is a new term for me, SQL in 2002 didn't have it. :D

kberg (approved):

bafolts:
I will sanity check this on postgres before merging.

bafolts:
This change works. I never use undo. It seemed I had to refresh the page after some undo operations. If there an issue to improve that user experience? We should be able to update this automatically without requiring a page refresh.

PR #2657, opened 2021-02-02
Added gen selector for log panel + pause/resume

chosta:
[WIP] http://fplbg.com:8080/ -> might be unstable, use for quick tests only
A 2 in 1 interface that:

- upon click (nr 1 to current gen) displays the log entries for the selected generation
- pauses the log polling (actually the whole game is paused for the active player)

As a side effect (but also a feature requested by users):

- pause / resume log polling that stops the interrupt and scroll to bottom on every log update

This is how it looks like atm:
log-gens

I'd like to hear suggestions about:

- the positioning of the Gen: 1 2 3 4 elements
- the design of the pause / resume actions (or maybe we don't do that)
- please think about edge cases or side effects that might be an issue

ToDo:

- solo mode add "of {nr_generations}"
- parsing all messages to look for 'Generation' string seems a bit weak to me. A model upgrade somehow?


sienmich:
I think the ideal solution would be to remove the `pause/resume` button and scroll to the bottom after an update if the scroll was at the bottom already. What do you think?
I like the whole idea!

kberg:
Historical log data isn't necessarily made available to players, so it won't show early generations in a long game.

chosta:
> I think the ideal solution would be to remove the `pause/resume` button and scroll to the bottom after an update if the scroll was at the bottom already (and selected gen is the current one). What do you think?
> I like the whole idea!

The problem is that when the user goes back to a previous gen, the polling stops and there HAS to be an indicator for that. Otherwise this might slow games if you forget to go back to the current gen. For me a button that feels safe and represents the state in a clear manner is acceptable. We'd considered putting an actions bar for the log anyway (search, toggle players, etc.)

chosta:
> Historical log data isn't necessarily made available to players, so it won't show early generations in a long game.

That's news :)
I thought we just loaded a small amount because of performance issues. I also thought the api @bafolts did for the log messages was to help solve that problem. Can we start storing all the information and what are the obstacles to do that?

kberg:
> > Historical log data isn't necessarily made available to players, so it won't show early generations in a long game.
> 
> That's news :)
> I thought we just loaded a small amount because of performance issues. I also thought the api @bafolts did for the log messages was to help solve that problem. Can we start storing all the information and what are the obstacles to do that?

You mean storing in the web client? I imagine that an unbounded log can slow down any app constantly modifying the DOM. (Having seen similar things in other frameworks.)

chosta:
> > > Historical log data isn't necessarily made available to players, so it won't show early generations in a long game.
> > 
> > 
> > That's news :)
> > I thought we just loaded a small amount because of performance issues. I also thought the api @bafolts did for the log messages was to help solve that problem. Can we start storing all the information and what are the obstacles to do that?
> 
> You mean storing in the web client? I imagine that an unbounded log can slow down any app constantly modifying the DOM. (Having seen similar things in other frameworks.)

No need to load the whole log but if a player needs to do it there shouldn't be an issue. I'm actually dealing with such issues in another project at the moment. The idea is to only show 50 (or whatever the setting is) lines of the log in the current gen when initially loading the player-home component. Should the user select another gen - only the messages for that gen should be shown (we can also add a limit on how many to load initially and do a lazy loading on scroll). The only time the user loads the whole log is when they decide to scroll back to the bottom of the message stack and slowly add (again lazy loading using the api) the log entries to the DOM. I can imagine this happening rarely (all the entries are added to the DOM) but even then the browser can handle up to 2k elements without any serious issues (most of the time).
So my question was - and correct me if I got your comment about the historical data wrong - Can we have all the log messages fetched by the `logs` api at any time or not?

kberg:
> So my question was - and correct me if I got your comment about the historical data wrong - Can we have all the log messages (or a chunk of them based on parameters) fetched by the `logs` api at any time or not?

Can it be done? Yes. With the current API? Probably. The API asks for 'n' rows, and it delivers them. Is that larger than the game's log? Unknown.

vsrisuknimit:
> I think the ideal solution would be to remove the `pause/resume` button and scroll to the bottom after an update if the scroll was at the bottom already (and selected gen is the current one). What do you think?

I am proposing another solution. The log scroll automatically when it is at the bottom. When you scroll up, it is paused but reveal a button that when click automatically scroll to the bottom. If this is difficult to code, I would prefer to have a pause/resume button so I don't have to scroll all the way back down to get an update. (What if I scroll almost all the way to the bottom but not at the bottom? Do I get an update in @sienmich's proposal?)

chosta:
> > I think the ideal solution would be to remove the `pause/resume` button and scroll to the bottom after an update if the scroll was at the bottom already (and selected gen is the current one). What do you think?
> 
> I am proposing another solution. The log scroll automatically when it is at the bottom. When you scroll up, it is paused but reveal a button that when click automatically scroll to the bottom. If this is difficult to code, I would prefer to have a pause/resume button so I don't have to scroll all the way back down to get an update. (What if I scroll almost all the way to the bottom but not at the bottom? Do I get an update in @sienmich's proposal?)

I don't mind doing the scroll thing. It should probably happen anyway. My issue is that we need an indicator that the polling is paused. Players can get confused when clicking through generations and waiting for their turn which will not come unless they go back to the current gen and scroll to top.

kberg:
> > > I think the ideal solution would be to remove the `pause/resume` button and scroll to the bottom after an update if the scroll was at the bottom already (and selected gen is the current one). What do you think?
> > 
> > 
> > I am proposing another solution. The log scroll automatically when it is at the bottom. When you scroll up, it is paused but reveal a button that when click automatically scroll to the bottom. If this is difficult to code, I would prefer to have a pause/resume button so I don't have to scroll all the way back down to get an update. (What if I scroll almost all the way to the bottom but not at the bottom? Do I get an update in @sienmich's proposal?)
> 
> I don't mind doing the scroll thing. It should probably happen anyway. My issue is that we need an indicator that the polling is paused. Players can get confused when clicking through generations and waiting for their turn which will not come unless they go back to the current gen and scroll to bottom.

One thing may be for the "resume" button to pulse whenever there's a new log entry.

chosta:
> One thing may be for the "resume" button to pulse whenever there's a new log entry.
That too. But the again @sienmich @vsrisuknimit want the buttons gone :)
Browser notifications are helping but not everyone has them on.

ssimeonoff:
the gen selector - maybe a dropdown on the same row as "game log" ? 
10+ gens game will result in many buttons and they are right into the user's face.

chosta:
> the gen selector - maybe a dropdown on the same row as "game log" ?
> 10+ gens game will result in many buttons and they are right into the user's face.

Why are you changing your designs after I finish them :dagger: :) ?
I kind of like this one. It represents progression. After 10 we can morph into something else.

vsrisuknimit:
> That too. But @sienmich @vsrisuknimit want the buttons gone :)
> Browser notifications are helping but not everyone has them on.

I think the button is fine. Just that if you want to be fancy, you don't need the pause button, you will still need a resume button.

kevinb9n:
For whatever it's worth, I'm seeing it like this:

Whether I manually pause, or auto-pause just by scrolling, either way it is very very helpful to see an indicator that new messages appeared since I did that.

And if that works, then I think manual pause won't have any advantage over auto-pause (aside from if it's just easier to get done, that is).

bafolts:
We have every game log available on the server. Originally started storing everything to allow for changes like this so glad to see this. We have the logs available in javascript memory so this shouldn't really be too much load on the node instance.

The impact on bandwidth is another story.

I have never looked but it does seem we have a 2TB bandwidth limit with the heroku instance.

https://devcenter.heroku.com/articles/limits#network

I have no idea if we are getting close to this limit and I don't think heroku provides this metric. We may want to come up with a strategy where the client side only pulls down log items they don't have. The game logs array should for the most part always grow. If we know how many entries in the game logs array a client has we should only need to send any they are missing. Since today we are always pulling down the last 50 items no matter what a change like this could lead to less bandwidth then we are using today. Once all of the logs are available client side the UI should be much simpler to implement as well. The UI could always assume it had every log message and a separate process could be attempting to make sure we have every log message on the client pulling down new items. The heroku dyno is storing hundreds of games in memory and the average memory usage looking right now is 125MB all things considered so I don't think the client side memory impact would be noticeable.

bafolts:
@chosta I hope my comment did not derail this PR. After implementing ETag support I think we can handle any potential bandwidth concerns by implementing caching of the log route. Odds are the logs are not going to change and a player is going to be requesting the same logs. I am going to work on that separate of this PR. If a new route shows up in the game log route to get logs by generation it will pick up the caching headers change.

chosta:
> I hope my comment did not derail this PR

@bafolts I'm just busier than normal these days. Your comments make sense and I've imagined that, only loading what is needed, should be the correct approach. These changes can be introduced when ready. For now I'll just go with what we have and try to release the functionality that I started doing with some minor tweaks based on the feedback.

chosta:
Was just thinking about the bandwidth concern. Currently, I'm just getting all the game logs and slicing the array between gen x and gen x +1. A better solution would be a query on DB level that does that. 
Maybe we can introduce a new Model for generation log messages or find a way to do it with sql?
I haven't dug into the DB design too much yet but might do it if you think this is reasonable.

bafolts:
The database should never end up being hit as we keep instances of `Game` in javascript memory once loaded. All the game logs should be available in javascript memory. If this is added and it ends up being a bandwidth problem we can remedy it. The limit of only sending the last 50 items today is a magic number. It wasn't introduced after we saw an issue. This PR is heading in the right direction. I am going to look into optimizing all the bandwidth we are using through better caching.

chosta:
- moved gen markers next to game log title
- added solo `of gen x` 
- updated icons
- fixed bugs

This should be ready to go. I've decided to not implement the scroll to bottom -> fetch batch of data and leave it for another PR. I'm curious to how players will react to the pause / resume log interaction. Image below is for solo mode. 
cool-story-bro

bafolts:
I will check this out tomorrow. Still excited to put the separate log api route to use!

src/routes/GameLogs.ts
```diff
@@ -3,12 +3,29 @@ import * as http from 'http';
...
     return url.startsWith('/api/game/logs?');
   }
+
+  public static getLogMessageIndexByGen(data: Array<LogMessage>, generation: number): number | undefined {
+    let result = undefined;
+    for (let i = 0; i < data.length; i++) {
+      const logMsg: LogMessage = data[i];
+      const isGen = logMsg.message.startsWith('Generation');
+      if (isGen) {
+        if (parseInt(logMsg.data[0].value) === generation) {
```
bafolts, line 20: I would check that `logMsg.data.length > 0` or use `?` operator. It `logMsg.data.length === 0` this will throw an error trying to read `value` property from `undefined`.

src/routes/GameLogs.ts
```diff
@@ -3,12 +3,29 @@ import * as http from 'http';
...
 export class GameLogs extends Route {
   public canHandle(url: string): boolean {
     return url.startsWith('/api/game/logs?');
   }
+
+  public static getLogMessageIndexByGen(data: Array<LogMessage>, generation: number): number | undefined {
+    let result = undefined;
+    for (let i = 0; i < data.length; i++) {
+      const logMsg: LogMessage = data[i];
+      const isGen = logMsg.message.startsWith('Generation');
```
bafolts, line 18: Instead of using `startsWith` I would check that `logMsg.message === 'Generation ${0}'`. This should be a more efficient check and also help ensure we aren't checking other messages that happen to start with `Generation`.

src/routes/GameLogs.ts
bafolts: Parsing this value is probably unnecessary. Every time we see `Generation ${0}` if we increment a counter we should be able to return when that counter reaches `generation`. Usually I wouldn't nit pick on performance but this route could end up iterating hundreds of items for multiple games simultaneously.

```
let counter = 0;
for (let i = 0; i < data.length; i++) {
    if (data[i].message === 'Generation ${0}') {
        counter++;
        if (counter === generation) {
            return i;
    }
}
````

File: src/routes/GameLogs.ts [hunk 3]
```diff
@@ -33,16 +51,32 @@ export class GameLogs extends Route {
...
+      let log = [...game.gameLog];
+
+      if (generation !== undefined && !Array.isArray(generation)) {
+        const theGen = parseInt(generation);
+
+        // find the index of the selected generation message
+        const startIndex = GameLogs.getLogMessageIndexByGen(log, theGen) || 0;
+
+        // find the next gen index
+        const endIndex = GameLogs.getLogMessageIndexByGen(log, theGen + 1);
```
bafolts, line 63: Providing a starting index to `getLogMessageIndexByGen` will avoid having to search from the start. I would change the method signature of `getLogMessageIndexByGen` to have a third argument of `startIndex`. When calling for `startIndex` we can provide `0`. When calling for `endIndex` we can provide `startIndex`. This could save us potentially iterating through the array twice.

src/routes/GameLogs.ts
```diff
@@ -33,16 +51,32 @@ export class GameLogs extends Route {
...
       }
 
-      const log = game.gameLog;
+      let log = [...game.gameLog];
+
+      if (generation !== undefined && !Array.isArray(generation)) {
+        const theGen = parseInt(generation);
+
+        // find the index of the selected generation message
+        const startIndex = GameLogs.getLogMessageIndexByGen(log, theGen) || 0;
```
bafolts, line 60: If this returns `undefined` I would return a 400 response or 404 instead of defaulting to 0. This will help catch errors earlier. Using 400 would make sense to me sense if we don't find the generation it is a bad request. Using 404 would make sense since we didn't find the generation.

src/routes/GameLogs.ts
```diff
@@ -33,16 +51,32 @@ export class GameLogs extends Route {
         return;
       }
 
-      const log = game.gameLog;
+      let log = [...game.gameLog];
```
bafolts, line 54: Well that will do it. This is most likely why we were still only seeing the last 50 logs stored. Pretty sure we would have been indirectly trimming this array here. I don't think we want to copy the entire array though as it could have hundreds of items. As long as we don't use `splice` we should be fine. We should change the usage of `splice` to use a `slice`.

src/components/LogPanel.ts
```diff
@@ -168,9 +177,69 @@ export const LogPanel = Vue.component('log-panel', {
...
+    },
+    getTitleClasses: function(): string {
+      const classes = ['log-title'];
+      classes.push(playerColorClass(this.color.toLowerCase(), 'shadow'));
+      return classes.join(' ');
+    },
+    getIconClass: function(): string {
+      return (this.$root as any).logPaused ? 'icon-play' : 'icon-pause';
+    },
+    getGenerationText: function(): string {
```
bafolts, line 228: Great change moving this in here :+1:

src/components/App.ts
```diff
@@ -49,6 +50,7 @@ export const mainAppSettings = {
       'millestones_list': true,
       'awards_list': true,
       'tags_concise': false,
+      'log_paused': false,
```
bafolts, line 53: Does this get used anywhere?

src/components/App.ts
```diff
@@ -81,6 +87,10 @@ export const mainAppSettings = {
       const currentPathname: string = window.location.pathname;
       const xhr = new XMLHttpRequest();
       const app = this as unknown as typeof mainAppSettings.data;
+
+      // fetching player data is ignored when the log data stream is paused
+      if (app.logPaused === true) return;
```
bafolts, line 92: We don't protect against this today and it warrants a separate issue but if an `updatePlayer` has an in progress `xhr` and a subsequent call comes in to `updatePlayer` we should abort the current `xhr`. Similarly here we would need to abort a running `xhr`.

src/components/LogPanel.ts
```diff
@@ -168,9 +177,69 @@ export const LogPanel = Vue.component('log-panel', {
     getCrossHtml: function() {
       return '<i class=\'icon icon-cross\' /i>';
     },
+    selectGeneration: function(gen: number): void {
+      this.selectedGeneration = gen;
+      // pause logging globally
+      (this.$root as any).changeLogPaused(true);
```
bafolts, line 183: To link this with `mainAppSettings` can you use `this.$root as unknown as typeof mainAppSettings.methods` here instead of `any`? This will ensure that if the method changes on `mainAppSettings` we update it here. Same for the other `as any` usages. I haven't found a more elegant way to link these with vue.

https://github.com/bafolts/terraforming-mars/blob/main/src/components/WaitingFor.ts#L81

Similarly for the `logPaused` you can use `this.$root as unknown as typeof mainAppSettings.data`.

src/components/LogPanel.ts
```diff
@@ -168,9 +177,69 @@ export const LogPanel = Vue.component('log-panel', {
...
+      fetch(`/api/game/logs?id=${this.id}&generation=${gen}`)
+        .then((response) => response.json())
+        .then((messages) => {
+          this.messages.splice(0, this.messages.length);
+          this.messages.push(...messages);
+          // if it's current gen go to default behavior
+          if (gen === this.generation) {
+            // resume logging globally
+            (this.$root as any).changeLogPaused(false);
+            (this.$root as any).updatePlayer();
```
bafolts, line 194: I would do this inside the `changeLogPaused` function body. Whenever `logPaused` is set to `false` we need to call `updatePlayer`.

bafolts (changes requested):
This is great work! This looks very close. I only have minor suggestions.

src/components/LogPanel.ts
```diff
@@ -168,9 +177,69 @@ export const LogPanel = Vue.component('log-panel', {
...
     },
+    selectGeneration: function(gen: number): void {
+      this.selectedGeneration = gen;
+      // pause logging globally
+      (this.$root as any).changeLogPaused(true);
+
+      fetch(`/api/game/logs?id=${this.id}&generation=${gen}`)
+        .then((response) => response.json())
+        .then((messages) => {
+          this.messages.splice(0, this.messages.length);
```
bafolts, line 188: Not important at all but while looking at this realize we can write this as:

```
this.messages.splice(0, this.messages.length, ...messages);
```

and avoid having to call `push` on next line. More of a cool typescript deal than anything else.

chosta:
@bafolts thank you for taking the to review the code, much appreciated! I'll try to polish things based on the feedback sometime this week.

src/routes/GameLogs.ts
chosta: Smart

src/routes/GameLogs.ts
chosta: Even smarter

src/components/App.ts
chosta: Yeah, I was feeling really weird doing this but couldn't find a quick way to do it. I guess we have to provide solution to the issue you mentioned in another PR.

src/components/LogPanel.ts
chosta: I kind of felt this comment was coming :) Was lazy for not looking for an efficient one-liner

src/routes/GameLogs.ts
```diff
@@ -33,16 +54,35 @@ export class GameLogs extends Route {
...
+      // TODO(chosta): ignore limit if gen is chosen until a scrollToTop batch loading is implemented (a.k.a. lazy load 50 lines at a time when user scrolls back)
+      } else {
+        if (limit !== undefined && !Array.isArray(limit)) {
+          const theLimit = parseInt(limit);
+          if (isNaN(theLimit)) {
+            this.badRequest(req, res);
+            return;
+          }
+          if (log.length > theLimit) {
+            log.splice(0, log.length - theLimit);
```
bafolts, line 84: The rest of this PR looks great but reading this line there is a bug which was here prior to this change where `splice` is used. I believe we are going to only keep the last 50 messages indirectly with this bug. Clicking to see the first generation once there were more than 50 log messages could end up providing the wrong generation. Going to try this locally to see if that is the case. Will need to change this to `slice` to not modify the `game.gameLog` array.

bafolts (approved):
This could have an impact to memory usage as we wont be indirectly trimming down to 50 messages. I will look at this tomorrow on terraforming-mars.herokuapp.com and see if it is going to be a memory issue. If it doesn't show a noticeable memory spike I will look into implementing some caching mechanism so we don't end up sending the same logs back.

nwai90:
This seems to crash ongoing games on server:

image

bafolts:
This problem was partly remedied with #2795 but there is a problem we will see with games currently running. The game logs route has a bug where a call to `splice` is modifying the `game.gameLogs` array and only storing the last 50 log items. For games that are far enough along where there are more than 50 log items when we call to `getLogMessageIndexByGen` it will return `undefined` until the bug is resolved. 

https://github.com/bafolts/terraforming-mars/blob/main/src/routes/GameLogs.ts#L85

For new games the new route should work without error and I don't think we will lose the old log messages. We will need some strategy for dealing with the games that are missing the previous items. I will hold off on pushing this change out to terraforming-mars.herokuapp.com for today.

Probably other ideas as well but we will either have to:

1. Keep the generation selector on the log component hidden for 10 days and push out a change to not lose old log messages. Then after 10 days every game should have every log message and the new component is fine.
2. From the log component attempt to determine when the server is missing the early generation log items. If the server is missing the log items then display only the log items we have.

I am leaning towards option 1 as it doesn't involve as much work and lets time flush out the missing log data for us versus someone having to do work.

PR #2660, opened 2021-02-02
fix version text

### LINKED ISSUE #2384: Feature request: More readable version details
image

- Redundant 'deploy'
- Space or new line to separate time. Label time zone if possible or convert to local time depending on the user.

vsrisuknimit:
Fix the version text when built on Heroku. This PR can close #2384.

Before
Screen Shot 2021-02-02 at 1 32 58 PM

After
Screen Shot 2021-02-02 at 1 41 46 PM

This problem does not happen when build locally. Not sure how to safely get rid of the `(Coordinated Universal Time)` at the end. Also do we want GMT +0? Probably right.

### DIFF
--- a/make_static_json.js
+++ b/make_static_json.js
@@ -41,7 +41,7 @@ function getAllTranslations() {
 function generateAppVersion() {
     // assumes SOURCE_VERSION is git hash
     if (process.env.SOURCE_VERSION) {
-        return process.env.SOURCE_VERSION.substring(0, 7) + " " + new Date().toUTCString();
+        return "deployed: " + new Date().toUTCString();
     }
     try {
         return child_process.execSync(`git log -1 --pretty=format:"%h %cD"`).toString();
--- a/make_static_json.js
+++ b/make_static_json.js
@@ -41,7 +41,7 @@ function getAllTranslations() {
 function generateAppVersion() {
     // assumes SOURCE_VERSION is git hash
     if (process.env.SOURCE_VERSION) {
-        return "deployed: " + new Date().toUTCString();
+        return process.env.SOURCE_VERSION.substring(0, 7) + " " + new Date().toUTCString();
     }
     try {
         return child_process.execSync(`git log -1 --pretty=format:"%h %cD"`).toString();
--- a/make_static_json.js
+++ b/make_static_json.js
@@ -41,7 +41,7 @@ function getAllTranslations() {
 function generateAppVersion() {
     // assumes SOURCE_VERSION is git hash
     if (process.env.SOURCE_VERSION) {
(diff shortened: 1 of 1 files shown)


vsrisuknimit:
What exactly does this `process.env.SOURCE_VERSION.substring(0, 7)` do? Can't I just replace with 'deployed'?

vsrisuknimit:
Simplify deploy. Switch to toUTCstring.

image

vsrisuknimit:
After switching to toUTCstring

image

make_static_json.js
```diff
@@ -41,7 +41,7 @@ function getAllTranslations() {
 function generateAppVersion() {
     // assumes SOURCE_VERSION is git hash
     if (process.env.SOURCE_VERSION) {
-        return process.env.SOURCE_VERSION.substring(0, 7) + " deployed " + new Date().toISOString();
+        return process.env.SOURCE_VERSION.substring(0, 7) + " " + new Date().toUTCString();
```
bafolts, line 44: Spent a surprising amount of time looking at this and still not entirely sure why the format with the `( )` suffix is seen on heroku. Heroku should be using node 12.x release. 

https://devcenter.heroku.com/articles/nodejs-support#specifying-a-node-js-version

My node 12.7 (on ubuntu linux and mac) locally does not include this value. According to MDN sometime in 2018 what the `toUTCString` function returned changed. 

> Prior to ECMAScript 2018, the format of the return value varied according to the platform. The most common return value was an RFC-1123 formatted date stamp, which is a slightly updated version of RFC-822 date stamps.

To remove the suffix I would use the following regular expression:

```
"Tue Feb 02 2021 19:00:05 GMT-0700 (Mountain Standard Time)"
new Date().toString().replace(/ \(.+\)/, '')
"Tue Feb 02 2021 19:00:08 GMT-0700"
```

bafolts (changes requested):

make_static_json.js
vsrisuknimit: @bafolts (time zone) is already removed with toUTCString. The picture shown above is with toUTCString from Heroku. Do I still need to change anything?

bafolts: However heroku is pulling `toUTCString` seems to be platform dependent. The time zone readable string is being included. I would add the regular expression to strip it on platforms which provide it.

bafolts: I used `toString` in my example as that was the only output which included what heroku is displaying when we use `toUTCString`.

vsrisuknimit: added reg ex

bafolts (approved):

PR #2661, opened 2021-02-03
add status for drafting and researching

### LINKED ISSUE #2260: hide the "active" and "next" labels during drafting?
It's confused me many a time. Wait, why am I active?
Just a small thing though.

vsrisuknimit:
This PR can close #2260 .

Everyone is drafting:
Screen Shot 2021-02-02 at 10 12 46 PM

Yellow is done with drafting the first pack:
Screen Shot 2021-02-02 at 10 12 57 PM

Everyone is researching phase (research phase after drafting):
Screen Shot 2021-02-02 at 10 11 53 PM

Yellow selected what cards to buy:
Screen Shot 2021-02-02 at 10 11 59 PM


src/Game.ts
```diff
@@ -630,6 +630,10 @@ export class Game implements ISerializable<SerializedGame> {
     });
   }
 
+  public finishedResearchPhase(player: Player): boolean {
+    return this.researchedPlayers.has(player.id);
```
bafolts, line 634: I think we already have this under a different name.

https://github.com/bafolts/terraforming-mars/blob/main/src/Game.ts#L797

src/components/overview/PlayersOverview.ts
```diff
@@ -64,6 +65,19 @@ export const PlayersOverview = Vue.component('players-overview', {
       return result.slice(0, -1);
     },
     getActionLabel(player: PlayerModel): string {
+      if (this.player.phase === Phase.DRAFTING) {
```
bafolts, line 68: This function has returns `ActionLabel.NONE` at the end which I think is helpful. If we didn't find any other situation where we returned then we return `ActionLabel.NONE`. I would modify this to only return `ActionLabel.DRAFTING` or `ActionLabel.RESEARCHING` when needed.

```typescript
if (this.player.phase === Phase.DRAFTING && player.needsToDraft) {
    return ActionLabel.DRAFTING;
}
if (this.player.phase === Phase.RESEARCH && player.needsToResearch) {
    return ActionLabel.RESEARCHING;
}
```

bafolts (changes requested):
I would re-use the `hasResearched` function on `Game` versus creating a separate function for this.

src/Game.ts
vsrisuknimit: Got it. Made it public to use here.

src/components/overview/PlayersOverview.ts
vsrisuknimit: ActionLabel.NONE is needed. Otherwise, ACTIVE player will be labeled active even when the label does not matter during drafting or research phase.

image

In this picture, Green already drafted.

bafolts (approved):

Issue #2221, opened 2020-12-29
Show cost of all standard projects


andystager:
I saw this was alluded to in #2192 discussion, but I wanted to highlight it specifically. There is no listing on the UI for costs of standard projects I can't afford at a given moment.

Scenario:
If I want to sell patents to buy a standard project that I cannot currently afford, I don't see a reference to confirm its cost (and thus how many cards I need to assess the sale of), since the projects above my current MC total are not displayed. Can they be shown but greyed out? Or add a reference around the game board for various costs?

I'm constantly second guessing myself on if a greenery costs 22 or 23 MC.

kevinb9n:
As a *general* rule I think dimming is "always" better than hiding, and made a short speech about it here:
https://github.com/bafolts/terraforming-mars/issues/2175#issuecomment-750410475
(a VERY short speech by my standards :-))
In this case, keeping them all there also lets you play a little bit more from muscle-memory. A very very small point.

By default I would keep everything in (even Aquifer after all the oceans are placed, since the new "warnings" system should address that?).

There can always be specific reasons that override a general rule though.

sienmich:
Yep, I still remember that one :)
Will work on this probably soon, thanks for reporting

PR #2664, opened 2021-02-03
Show all standard projects

sienmich:
When having 17MC it looks like this:
Screenshot from 2021-02-03 15-19-18

The gray ones are not clickable

Closes https://github.com/bafolts/terraforming-mars/issues/2221


vsrisuknimit:
What happen when you have 23 mc but Ocean and Temp are already maxed out?

sienmich:
> What happen when you have 23 mc but Ocean and Temp are already maxed out?

With maxed oceans (works! :) ):
Screenshot from 2021-02-03 15-39-11

vsrisuknimit:
Question is whether the grayed out buttons should be sorted out to the back end of the list. I am not sure.

sienmich:
> Question is whether the grayed out buttons should be sorted out to the back end of the list. I am not sure.

I think having them in the same order is better, we can get used to their constant position, muscle memory etc.
But if we want to, then probably in the next issue / PR.

sienmich:
Additionally, I think we could use the same method to merge the `Play card from hand` with `Cards in hand`, as someone suggested in discord lately (then some option button to sort by `canPlay` would be nice)

kberg:
We don't do this with other cards, so why is this special?

>

sienmich:
> We don't do this with other cards, so why is this special?
> […](#)

Because players want to see the cost of standard projects they can not afford (in #2221)

kberg:
> > We don't do this with other cards, so why is this special?
> > […](#)
> 
> Because players want to see the cost of standard projects they can not afford (in #2221)

Yeah that makes sense.

src/Player.ts
```diff
@@ -1634,26 +1634,25 @@ export class Player implements ISerializable<SerializedPlayer> {
       (canUseMicrobes ? this.getMicrobesCanSpend() * 2 : 0);
   }
 
-  // Public for testing
-  public getPlayableStandardProjects(): Array<StandardProjectCard> {
+  private getStandardProjects(): Array<StandardProjectCard> {
     // TODO(kberg): Filter playability based on the project's reserve units.
```
kberg, line 1638: This comment is irrelevant now.

src/inputs/SelectCard.ts
```diff
@@ -14,6 +14,7 @@ export class SelectCard<T> implements PlayerInput {
         public maxCardsToSelect: number = 1,
         public minCardsToSelect: number = 1,
         public selectBlueCardAction: boolean = false,
+        public canAct?: Array<boolean>,
```
kberg, line 17: This is a strange behavior, being a special case, so I suggest some changes.

1. Name this `enabled` so it's clear this isn't related to the `canAct` methods on cards.
2. Add a comment describing what this is and why it is optional. (selectBlueCardAction should have the same thing.)
2a. In fact, it's looking to be time to just turn all of those into a structure of parameters.

src/server/ServerModel.ts
```diff
@@ -215,7 +216,7 @@ function getCardsAsCardModel(
       resourceType: card.resourceType,
       calculatedCost: 0,
       cardType: CardType.AUTOMATED,
-      isDisabled: false,
+      isDisabled: canAct?.[index] === false,
```
kberg, line 219: Rename this to `enabled` and suddenly things will be even across the board.

src/components/card/Card.ts
```diff
@@ -106,7 +106,7 @@ export const Card = Vue.component('card', {
       const classes = ['card-container', 'filterDiv', 'hover-hide-res'];
       classes.push('card-' + card.name.toLowerCase().replace(/ /g, '-'));
 
-      if (this.actionUsed) {
+      if (this.actionUsed || card.isDisabled) {
```
kberg, line 109: The class being pushed is designed for `cards-action-was-used.` It might be convenient to resuse it, but please either rename the CSS class or something. Otherwise someone in the future is going to look at this and think "What does card.isDisabled have to do with action-was-used?"

kberg (commented):
Adding a test that shows how the SelectCard is populated is probably in order.

src/Player.ts
```diff
@@ -1634,26 +1634,25 @@ export class Player implements ISerializable<SerializedPlayer> {
...
+      .filter((card) => card.name !== CardName.SELL_PATENTS_STANDARD_PROJECT)
       .sort((a, b) => a.cost - b.cost);
   }
 
-  private getPlayableStandardProjectOption(): PlayerInput | undefined {
-    const standardProjects: Array<StandardProjectCard> = this.getPlayableStandardProjects();
-    if (standardProjects.length === 0) {
-      return undefined;
-    }
+  // Public for testing
```
bafolts, line 1645: Anything needed only for testing should be able to use the `TestPlayer` class. Then a needed function for testing can be made `protected` versus `private`.

src/inputs/SelectCard.ts
bafolts: Another option. Similar to `SelectInitialCards` you could create a `SelectStandardProject`. Then this could remain specific to standard projects and not have to go with the generic `SelectCard`.

src/Player.ts
```diff
@@ -1634,26 +1634,25 @@ export class Player implements ISerializable<SerializedPlayer> {
...
+  public getStandardProjectOption(): SelectCard<StandardProjectCard> {
+    const standardProjects: Array<StandardProjectCard> = this.getStandardProjects();
 
     return new SelectCard(
       'Standard projects',
       'Confirm',
       standardProjects,
       (card) => card[0].action(this),
```
bafolts, line 1653: Since we are not checking for `card.canAct` when allowing the user to select this card we will want to check here that `card.canAct` is `true`. Otherwise this could introduce a place where a user could cheat. If a clever user were to provide a card that was disabled we would run the action.

sienmich: Good point, thanks. Resolved in https://github.com/bafolts/terraforming-mars/pull/2664/commits/15b93cf90354143d34bdf6839042b6da510c4cc1

src/inputs/SelectCard.ts
sienmich: @kberg
1., 2. - done
2a - sounds good, in other PR

sienmich: @bafolts
I think that this option could be useful not only with standard projects, so I'd like to leave it that way.
For example, playing a card from hand could also display unplayable cards (as an option)

src/server/ServerModel.ts
sienmich: This one I'd rather not rename, it's used in some places (was here before me :) ).
Changing it to the opposite meaning would change more than I want

src/components/card/Card.ts
sienmich: Agree, changed the css name to `card-unavailable`

src/Player.ts
sienmich: Hm, I think not really. The protected function could be called from within the `TestPlayer`. In this case, there is some turmoil test that checks which SPs are available.
There are no `protected` methods in Player for now, so if we want to do this, then next PR preferably

sienmich:
@kberg, @bafolts, thanks for the review!
I think I addressed all of your concerns for now :smile:

src/inputs/SelectCard.ts
kberg: > Another option. Similar to `SelectInitialCards` you could create a `SelectStandardProject`. Then this could remain specific to standard projects and not have to go with the generic `SelectCard`.

I do like that option, but if it's another PR, that's fine too.

src/server/ServerModel.ts
kberg: Ah I didn't realize that. I'd still wish. Parameters in the negative aren't quite an _antipattern_ but they can be a little tricky. A thing for another day.

src/Player.ts
kberg: > Hm, I think not really. The protected function could be called from within the `TestPlayer`. In this case, there is some turmoil test that checks which SPs are available.
> There are no `protected` methods in Player for now, so if we want to do this, then next PR preferably

@sienmich will you own this? If not, drop a TODO.

src/Player.ts
sienmich:
```suggestion
```

src/Player.ts
sienmich:
```suggestion
  // Public for testing. TODO: make protected using the TestPlayer class.
```

sienmich: Added TODO.

src/Player.ts
```diff
@@ -1848,7 +1851,7 @@ export class Player implements ISerializable<SerializedPlayer> {
       action.options.push(remainingAwards);
     }
 
-    const standardProjectsOption = this.getPlayableStandardProjectOption();
+    const standardProjectsOption = this.getStandardProjectOption();
     if (standardProjectsOption !== undefined) {
       action.options.push(standardProjectsOption);
     }
```
sienmich, line 1857:
```suggestion
    action.options.push(standardProjectsOption);
```

sienmich:
So, any more concerns? :thinking:

src/Player.ts
```diff
@@ -832,6 +834,9 @@ export class Player implements ISerializable<SerializedPlayer> {
       const mappedCards: Array<ICard> = [];
       for (const cardName of input[0]) {
         mappedCards.push(this.getCard(pi.cards, cardName));
+        if (pi.enabled?.[pi.cards.map((card) => card.name).indexOf(cardName)] === false) {
```
bafolts, line 837: This usage of `.map.indexOf` is equivalent of `findIndex`. Once this uses `findIndex` this looks good to me.

`pi.cards.findIndex((card) => card.name === cardName)`

sienmich:
```suggestion
        if (pi.enabled?.[pi.cards.findIndex((card) => card.name === cardName)] === false) {
```
Ok, committing

bafolts (approved):

motoyugota:
> Additionally, I think we could use the same method to merge the `Play card from hand` with `Cards in hand`, as someone suggested in discord lately (then some option button to sort by `canPlay` would be nice)

If the reasoning for doing this with Standard Projects is to show the costs even if you can't afford one right now, then this I think is even more important. It is easy to forget that you have a particular card with a requirement that you are waiting to play, and not seeing that card when you go to "play a card" makes it very easy to forget about. You may play things differently to make that requirement happen.

PR #2666, opened 2021-02-04
Add vicftory point breakdown for The Moon.

kberg:


src/VictoryPointsBreakdown.ts
```diff
@@ -1,11 +1,14 @@
-export type VictoryPoints = 'terraformRating' | 'milestones' | 'awards' | 'greenery' | 'city' | 'victoryPoints';
+export type VictoryPoints = 'terraformRating' | 'milestones' | 'awards' | 'greenery' | 'city' | 'moon colony' | 'moon mine' | 'moon road' | 'victoryPoints';
```
avhetman, line 1: With this many strings in this type, converting this to an enum at this point seems to make sense. Thoughts?

kberg: At some point, perhaps. It's not going to meaningfully improve anything.

bafolts (approved):

PR #2678, opened 2021-02-05
Add RustEatingBacteria.

kberg:


src/cards/moon/RustEatingBacteria.ts
```diff
@@ -0,0 +1,57 @@
...
+        }),
+      },
+    });
+  };
+
+  public resourceCount = 0;
+
+  public canPlay(): boolean {
+    return true;
+  }
```
sienmich, line 38:
```suggestion
```

sienmich (approved):

src/cards/moon/RustEatingBacteria.ts
kberg: Oh I see, you suggest removing the lines. I'm going to hang back on removing it for the moment.

Issue #776, opened 2020-06-06
Improvement: Show discounts on card price


funkybaboon:
If I get any discount on a card price, It is not reflected in the price tag.

Suggestion: Add the discounted price to the price tag. Show the original price strike through, f.ex. like this:

<img width="659" alt="Screenshot 2020-06-06 at 12 58 56" src="https://user-images.githubusercontent.com/2966599/83942641-c3154a80-a7f5-11ea-9876-1f310c3f9814.png">

This was done with the chrome developer Tools changing the price div to

```
<div class="price ">
  <del>27</del>
  " 25"
</div>
```

and changing the width from 44px to 100px.

nwai90:
Related to https://github.com/bafolts/terraforming-mars/issues/412 as all displayed cost values are currently hardcoded in HTML_data

FrankJones314:
A related idea - someone suggested this one steam for the asmodee version:  Maybe an indicator of which cards are not yet playable or no loner playable because of global parameters.   As it stands now, when clicking "[p;lay a card" it hides ones that we cannot afford And that have global parameter issues.    That sometimes results in misplays of thinking, "ell, after I place the greenery next to three oceans I'll have the 6 MC for this card" except then realizing the oxygen is too high.  

It would be similar  to the showing the true cost and adjusted cost after discounts.   Maybe the global parameter requirement area of the card could turn green when it has been reached (including adjustments for adaptation tech and special design), and can turn permanent red once a parameter has been passed.  

Perhaps tag requirements can be done similarly.   When someone plays his 5th science tag, for example, something near the "requires 5 science tags" text on mass converter changes to green or something.

kevinb9n:
@FrankJones314 I just happened to file the same request, as part of bug 1407.

kevinb9n:
It may be hard to include both pieces of information (printed cost and effective cost) without it looking bad, so another option might be a toggle button somewhere (preferably not hidden away in settings?) between seeing cards-as-printed and cards-as-modified-by-effects. I think this should include cost mods and requirement mods. We should have a convention that the info is styled slightly differently (e.g. cost in dark green characters instead of black?) whenever it's modded. And maybe hovering on the cost or requirement should show the opposite mode......

kevinb9n:
[...]

PR #2679, opened 2021-02-05
Display card discounts

sienmich:
Closes https://github.com/bafolts/terraforming-mars/issues/776.

Currently looks like this (2 MC discount for science tags):
Screenshot from 2021-02-05 16-41-17

I think it's kinda ugly, I need some help with that UI. Ideally discounted cards could show both the old and new cost of the card. Any ideas on how to achieve that? (Maybe sth like here https://github.com/bafolts/terraforming-mars/issues/776#issue-632380560).

TODO: Display discounts only on cards in hand and draft cards, not on already played cards.

Edit:
Applied @vsrisuknimit suggestion:
Screenshot from 2021-02-05 21-10-12


vsrisuknimit:
@sienmich something like this?
image

vsrisuknimit:
Does this code apply discount to all cards?

I think discount should only be applied to cards in hand and select card to play. We don't care about discount once it is played by us or by other players.

kevinb9n:
The march of progress!! I love it...

If you're wanting to have the "design discussion" about the best longer-term design, then I think it's worth thinking about the broader set of card modifications people will want to see and having a consistent plan. That discussion, if anyone wants to have it, is what #2187 was meant to be for (also linked from this issue).

I think we should try to avoid using more space on the card (for example, if requirements change to icons they might still move up into the header like they are in print; enough cards are over-packed with content that I think @chosta is at least considering that change). And people don't really need to see both at once, so I have felt that an easy toggle is the way. I think advanced players will tend to leave the toggle on almost all the time and don't need to see the original cost taking up space.

vsrisuknimit:
> And people don't really need to see both at once

You need to see both price at once for things like Magnate, Credicor, Spinoff Department and Advertising.

sienmich:
@kevinb9n, while I like the #2187 goggles idea, I think that reduced card costs could be just always displayed.

I also want to put the requirements on the top of the card, but I think for now I'll go with Jeep's 3rd suggestion. One day we can put the requirements there and move those costs somewhere else.

vsrisuknimit:
image

@sienmich The discount screw up corporation and prelude. It also applies to all cards including played cards. We should not show discounted price on the played cards because it does not matter.

vsrisuknimit:
I personally prefer design 2 over 3. The arrow helps implying the transparent number is the original.

sienmich:
> @sienmich The discount screw up corporation and prelude. It also applies to all cards including played cards. We should not show discounted price on the played cards because it does not matter.

Yep, working on it

vsrisuknimit:
Stepping back a moment. I am wondering whether it would be cleaner to apply the discount not to CardModel but as an overlay directly from PlayerHome. There, you already have player and cards. So you can calculate the discount card by card. Also you can choose to apply the discount overlay only on the cards in hand. I am not sure if it will be difficult for select cards to play. Just think that we should not mess around too much with waitingFor when we might not need to.

kevinb9n:
> > And people don't really need to see both at once
> 
> You need to see both price at once for things like Magnate, Credicor, Spinoff Department and Advertising.

(Non-blocking:  We are thinking the same way, except for the "at once" part. I think the combination of the number being styled differently (green) with a toggle to turn it off would also get the same job done. I suspect it will be more practical in the long run but am totally fine with ANY way this feature happens for now!)

sienmich:
Fixed both issues.
> I am wondering whether it would be cleaner to apply the discount not to CardModel but as an overlay directly from PlayerHome.

That's an interesting idea. But then it would be hard to apply discounts in some player inputs, like selecting the card during the draft. I prefer sticking with my current implementation, the `waitingFor` is not an issue here (I think)

kevinb9n:
> It also applies to all cards including played cards. We should not show discounted price on the played cards because it does not matter.

Let's cover the bases. A card can be

* In play
* In events pile
* In hand
* In selection (deciding which to discard and which to take/buy)
* In draft (deciding which to pass)
* On robots
* Discarded

The discounts are *clearly* relevant to: in hand, in selection, and in robots.
The discounts are clearly not relevant to: in play, in events pile, discarded.

That leaves "in draft", where I would say: sure, the discount may be a distraction from the *hate-drafting* perspective, but is still *mostly* useful and desirable information. This is actually an easier call if we're showing both values - we should just do it.

(Will the discount for robot cards (from the resources on the card) be included? I believe it "should", because unlike other cases of expending resources toward a card purchase, you have no choice in the matter. The resources will go away and the discount will be a done deal. Don't block over this of course.)

vsrisuknimit:
@sienmich discount should not apply to player input other than play, draft, research, and SRR. So it should not apply when you select target for microbe, animals, floaters, Viron, Project Inspection, etc.

That was why I thought my suggestion might be cleaner. But again, I am not sure how easy it will be to implement from there.

image

src/components/card/CardCost.ts
```diff
@@ -14,8 +17,14 @@ export const CardCost = Vue.component('CardCost', {
...
     },
+    displayTwoCosts: function(): boolean {
+      return this.newCost !== undefined && this.newCost !== this.amount;
+    },
   },
   template: `
-        <div :class="getClasses()">{{ amount === null ? 0 : amount }}</div>
+    <div>
+        <div :class="getClasses()">{{ newCost !== undefined ? newCost : (amount === null ? 0 : amount) }}</div>
+        <div :class="'card-old-cost'" v-if="displayTwoCosts()">{{amount}}</div>
```
vsrisuknimit, line 27: It can just be `class="card=old-cost"` if it is not a function.

sienmich:
Ok, good point, didn't see that one. I'll think about it

vsrisuknimit:
Some suggestions for the cost transition arrow.

```
<template v-if="displayTwoCosts()">
          <div class="card-cost-transition"><<</div>
          <div class="card-old-cost">{{amount}}</div>
</template>
```


```
            .card-cost-transition {            
                position: absolute;
                left: 41px;
                top: 12px;
                color: white;
            }
```

Screen Shot 2021-02-05 at 4 33 28 PM
Screen Shot 2021-02-05 at 4 32 43 PM
Screen Shot 2021-02-05 at 4 30 48 PM
Screen Shot 2021-02-05 at 4 30 28 PM

vsrisuknimit:
With asset instead of unicode.

Screen Shot 2021-02-05 at 5 16 39 PM

Screen Shot 2021-02-05 at 5 15 15 PM

sienmich:
> @sienmich discount should not apply to player input other than play, draft, research, and SRR. So it should not apply when you select target for microbe, animals, floaters, Viron, Project Inspection, etc.

I think the best solution would be add new parameter to `SelectCard`, something like `showDiscounts`.
And then we can show those discounts for selecting a card to buy / play / draft / discard, but not when adding / removing resources.

That'll be a bigger PR than I expected, so more general changes are in https://github.com/bafolts/terraforming-mars/pull/2687.

chosta (commented):
Thoughts on design:
1. A smaller translucent original cost box (love it)
2. I prefer a simple symbol for showing the cost. Probably `>>` instead of the custom made icons. Let's not try to mimic the steam app.
3. Positioning - we have a problem called `Ecology Research`. Let's try and put it below like the steam app does :)

Thoughts on implementation:
1. No issues that I found in code concerning the frontend
2. Not sure about the backend part. I would have probably done it the same way but I want to hear from @bafolts, @Lynesth or @kberg  on this. Is this the optimal way of doing it?

Would be a great improvement in general!

sienmich:
Screenshot from 2021-02-07 11-48-30
Right, I forgot about that one. I agree that the old cost below the new one (instead of next to it) would be the best solution. @vsrisuknimit, would you like to make that happen?

vsrisuknimit:
image

@sienmich I made a PR to your branch.

@chosta

vsrisuknimit:
@sienmich I decided to change the size, fix the asset and move things up a bit to avoid colliding with the card text.

image

chosta:
The translucency dos not work well with the color of the card type. Maybe reduce it a bit or remove it in general?
Also the arrow passing through the MC icons is something I don't really fancy.

kevinb9n:
What will this do with Indentured/Conscription?

sienmich:
> What will this do with Indentured/Conscription?

The new price will be accordingly reduced (no need to change anything, that's how those cards work already)

vsrisuknimit:
> The translucency dos not work well with the color of the card type. Maybe reduce it a bit or remove it in general?

I can play around with the transparency or do slightly gray scale instead.

> Also the arrow passing through the MC icons is something I don't really fancy.

Do you prefer the double arrow unicode? @chosta

vsrisuknimit:
> The new price will be accordingly reduced (no need to change anything, that's how those cards work already)

@sienmich I think @kevinb9n means is "will the price on the top left of the cards change to be 8 lower after Indentured Worker is played?".

sienmich:
> @sienmich I think @kevinb9n means is "will the price on the top left of the cards change to be 8 lower after Indentured Worker is played?".

So do I, and by `The new price will be accordingly reduced` I meant yep, the price on the top left will be reduced. :smile:

vsrisuknimit:
@sienmich Cool. I did not remember that those cards make the graphical cost drops already.

sienmich:
> @sienmich Cool. I did not remember that those cards make the graphical cost drops already.

Well, did they not. My change is to display the cost a player would have to pay (instead of the normal cost). There is already a function that calculates that cost (includes all discounts, for tags, and for stuff like Conspiration). So there was nothing to be done, already works.

sienmich:
Screenshot from 2021-02-08 22-34-08

I think no opacity works good

vsrisuknimit:
@sienmich Try playing around with `filter: grayscale(60%);` 60%, 70% or 80%.

sienmich:
I think this is ready.
I'm closing this PR and reopen it to reduce scrolling.

PR #2680, opened 2021-02-05
Add Lunar Steel

kberg:


src/cards/moon/LunarSteel.ts
```diff
@@ -0,0 +1,41 @@
...
+          });
+        }),
+      },
+    });
+  };
+
+  public canPlay(): boolean {
+    return true;
+  }
+
```
sienmich, line 32:
```suggestion

```
We don't need those, right?

sienmich (approved):
That's a strong card

kberg:
> That's a strong card

Could be, yes. When The Moon is in play, steel is used for other things than buildings, so perhaps that adjusts its value a bit?

PR #2683, opened 2021-02-06
etag support for assets

bafolts:
This should make a major dent in the amount of bandwidth we are sending from production environments. Trying to only introduce this when `NODE_ENV=production` to not impact local development where these files may change.

The change is pretty great.

image

Only 19kb transferred for 7MB.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -39,7 +39,6 @@ zlib.gzip(styles, function(err, compressed) {
     console.warn('error compressing styles', err);
     return;
   }
-  console.log('adding the compressed styles');
   fileCache.set('styles.css.gz', compressed);
 });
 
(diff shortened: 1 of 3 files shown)


src/server.ts
```diff
@@ -31,6 +31,7 @@ const route = new Route();
 const gameLogs = new GameLogs();
 const assetCacheMaxAge = process.env.ASSET_CACHE_MAX_AGE || 0;
 const fileCache = new Map<string, Buffer>();
+const hashCache = new Map<string, string>();
```
kberg, line 34: Consider fileCache = new Map<String, {buffer: Buffer, hash: string}>

kberg (approved):
Otherwise looks great.

PR #2685, opened 2021-02-06
Add Orbital Power Grid

kberg:


src/cards/moon/OrbitalPowerGrid.ts
```diff
@@ -0,0 +1,40 @@
...
+        renderData: CardRenderer.builder((b) => {
+          b.production((pb) => pb.energy(1)).slash().city().asterix();
+        }),
+        victoryPoints: 1,
+      },
+    });
+  };
+
+  public play(player: Player) {
+    const amount = player.game.getCitiesInPlay() - player.game.getCitiesInPlayOnMars();
```
sienmich, line 31: So it counts only your cities. Maybe put that info in the description? Sound like it would count all the cities in game.

src/cards/moon/OrbitalPowerGrid.ts
```diff
@@ -0,0 +1,40 @@
...
+    player.addProduction(Resources.ENERGY, amount);
+    LogHelper.logGainProduction(player, Resources.MEGACREDITS, amount);
```
sienmich, line 33: Why log mc production?

src/cards/moon/OrbitalPowerGrid.ts
kberg: It should count all cities. I suspect you're confusing that `player.game` is fetching the game instance, and has nothing to do with the player. Otherwise, it'd be `player.game.getCitiesInPlay(player)`

src/cards/moon/OrbitalPowerGrid.ts
kberg: Well, I think we should be logging all production shifts.

src/cards/moon/OrbitalPowerGrid.ts
kberg: Does this need a test verifying it? Probably not? But if you think that would help, I will do it.

kberg:
Thanks for the feedback, @sienmich

src/cards/moon/OrbitalPowerGrid.ts
sienmich: Yep, but you change energy prod and log the MC prod

sienmich: While MC prod is unchanged

src/cards/moon/OrbitalPowerGrid.ts
sienmich: Ok, my bad here, I didn't see that `.game` indeed. No test necessary imo

src/cards/moon/OrbitalPowerGrid.ts
kberg: **this is why all resource and production changes should be behind an api. :D :D :D**

Thanks! Will fix momentarily.

src/cards/moon/OrbitalPowerGrid.ts
kberg: Thanks!

sienmich (approved):

PR #2687, opened 2021-02-06
Improved ServerModel

sienmich:
While working on https://github.com/bafolts/terraforming-mars/pull/2679 I noticed, that there are two functions `getCards` and `getCardsAsCardModel` doing the same thing. I merged them into one and added options parameter.

Additioanly I made `CardModel.calculatedCost` optional (earlier `getCardsAsCardModel` return 0 as `calculatedCost`, which is not the best solution).

I'd like to merge this PR before #2679.

### DIFF
--- a/src/components/SelectHowToPayForProjectCard.ts
+++ b/src/components/SelectHowToPayForProjectCard.ts
@@ -233,7 +233,11 @@ export const SelectHowToPayForProjectCard = Vue.component('select-how-to-pay-for
     },
     cardChanged: function() {
       this.card = this.getCard();
-      this.cost = this.card.calculatedCost;
+      if (this.card.calculatedCost === undefined) {
+        throw new Error('Unexpected undefined calculatedCost in SelectHowToPayForProjectCard');
+      } else {
+        this.cost = this.card.calculatedCost;
+      }
       this.tags = this.getCardTags();
 
       this.megaCredits = (this as unknown as typeof PaymentWidgetMixin.methods).getMegaCreditsMax();
--- a/src/components/SelectHowToPayForProjectCard.ts
+++ b/src/components/SelectHowToPayForProjectCard.ts
@@ -233,11 +233,7 @@ export const SelectHowToPayForProjectCard = Vue.component('select-how-to-pay-for
     },
     cardChanged: function() {
       this.card = this.getCard();
-      if (this.card.calculatedCost === undefined) {
-        throw new Error('Unexpected undefined calculatedCost in SelectHowToPayForProjectCard');
-      } else {
-        this.cost = this.card.calculatedCost;
-      }
+      this.cost = this.card.calculatedCost!; // TODO(sienmich): Remove the ! here. It could require splitting the CardModel into ProjectCardModel and CorporationCardModel.
       this.tags = this.getCardTags();
 
       this.megaCredits = (this as unknown as typeof PaymentWidgetMixin.methods).getMegaCreditsMax();
--- a/src/components/SelectHowToPayForProjectCard.ts
+++ b/src/components/SelectHowToPayForProjectCard.ts
@@ -233,7 +233,7 @@ export const SelectHowToPayForProjectCard = Vue.component('select-how-to-pay-for
     },
     cardChanged: function() {
       this.card = this.getCard();
-      this.cost = this.card.calculatedCost!; // TODO(sienmich): Remove the ! here. It could require splitting the CardModel into ProjectCardModel and CorporationCardModel.
+      this.cost = this.card.calculatedCost || 0;
       this.tags = this.getCardTags();
 
       this.megaCredits = (this as unknown as typeof PaymentWidgetMixin.methods).getMegaCreditsMax();
--- a/src/server/ServerModel.ts
+++ b/src/server/ServerModel.ts
@@ -253,7 +253,8 @@ function getWaitingFor(
     playerInputModel.canUseHeat = shtpfpc.canUseHeat;
     break;
   case PlayerInputTypes.SELECT_CARD:
-    playerInputModel.cards = getCards(player, (waitingFor as SelectCard<ICard>).cards);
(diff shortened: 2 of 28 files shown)


src/server/ServerModel.ts
```diff
@@ -349,20 +319,31 @@ function getWaitingFor(
   return playerInputModel;
 }
 
+function getCalculatedCost(card: ICard, player: Player) : number | undefined {
+  if ([CardType.ACTIVE, CardType.AUTOMATED, CardType.EVENT].includes(card.cardType)) {
```
kberg, line 323: You can simplify this to `return card.cost === undefined ? undefined : player.getCardCost(card as IProjectCard)` and also rename this to getCost(). There's nothing special about it being 'calculated'.

src/server/ServerModel.ts
```diff
@@ -349,20 +319,31 @@ function getWaitingFor(
...
+
 function getCards(
   player: Player,
-  cards: Array<IProjectCard>,
-  showResouces: boolean = true,
+  cards: Array<ICard>,
+  options: {
+    showResouces?: boolean,
+    reserveUnitMap?: Map<CardName, Units>,
+    enabled?: Array<boolean>, // If provided, then the cards with false in `enabled` are not selectable and grayed out
```
kberg, line 335: This can be simplified by naming the array required, and giving it a default value of `Array(cards.length).fill(true)`

src/server/ServerModel.ts
```diff
@@ -123,7 +123,7 @@ export class Server {
       turmoil: turmoil,
       venusScaleLevel: game.getVenusScaleLevel(),
       victoryPointsBreakdown: player.getVictoryPoints(),
-      waitingFor: getWaitingFor(player.getWaitingFor()),
+      waitingFor: getWaitingFor(player, player.getWaitingFor()),
```
kberg, line 126: No need to pass both parameters. Just pass `player`.

kberg (changes requested):

src/server/ServerModel.ts
sienmich: Not really, since there is the
```ts
    if (waitingFor.options !== undefined) {
      for (const option of waitingFor.options) {
        const subOption = getWaitingFor(player, option);
        if (subOption !== undefined) {
          playerInputModel.options.push(subOption);
        }
```
So, it is not always `getWaitingFor(player, player.getWaitingFor())`

src/server/ServerModel.ts
sienmich: Nice, agree, done

src/components/SelectHowToPayForProjectCard.ts
```diff
@@ -233,7 +233,11 @@ export const SelectHowToPayForProjectCard = Vue.component('select-how-to-pay-for
     },
     cardChanged: function() {
       this.card = this.getCard();
-      this.cost = this.card.calculatedCost;
+      if (this.card.calculatedCost === undefined) {
+        throw new Error('Unexpected undefined calculatedCost in SelectHowToPayForProjectCard');
```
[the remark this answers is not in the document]
sienmich: This should never happen. I changed that to `this.cost = this.card.calculatedCost!;`, should be simpler.

This comes from the fact, that we have one class `CardModel` to both `IProjectCards` and some `ICorporationCard`, which is not optimal here (ones do have cost, ones do have not).

src/server/ServerModel.ts
sienmich: I'm not sure, most of the time it is an unused parameter, so one undefined seems more optimal than an array of falses.
Anyway, I'm planning to change the `enabled` from `Array<bool>` to function `(card) => bool`, which should be cleaner.

src/components/SelectHowToPayForProjectCard.ts
```diff
@@ -233,7 +233,7 @@ export const SelectHowToPayForProjectCard = Vue.component('select-how-to-pay-for
     },
     cardChanged: function() {
       this.card = this.getCard();
-      this.cost = this.card.calculatedCost;
+      this.cost = this.card.calculatedCost!; // TODO(sienmich): Remove the ! here. It could require splitting the CardModel into ProjectCardModel and CorporationCardModel.
```
kberg, line 236: How about `this.card.calculatedCost || 0`?

src/server/ServerModel.ts
kberg: Ugh,yeah.

src/server/ServerModel.ts
kberg: All fine.

kberg (approved):
How do you know this works? This part of the system doesn't have a lot of tests.

src/components/SelectHowToPayForProjectCard.ts
bafolts: Using `0` makes sense as it was the behavior before and it makes sense as technically corporation cards have no calculated cost. Using `!` is always going to lead to side effects.

bafolts (changes requested):

src/components/SelectHowToPayForProjectCard.ts
sienmich: Sure
```suggestion
      this.cost = this.card.calculatedCost || 0;
```

sienmich:
> How do you know this works? This part of the system doesn't have a lot of tests.

That's true.
I think it's generally not the best part of the system. I think it would be cleaner to have some subclasses for each PlayerInput type than those switches.
I made minimal changes required to reasonably implement showing card discounts and show card owners in the future.
I don't really feel like writing tests to this.
Testing this manually (playing game) looks good.

kberg:
Why request an additional review? I think this is ready to go.

sienmich:
@bafolts requested changes, so until he approves the merge is blocked

bafolts (approved):

kberg:
Well, that makes sense, thanks.

>

PR #2688, opened 2021-02-06
One time link

vsrisuknimit:
Why: One time link allows user to share the admin link freely without having to worry other players will get on their player page and sabotage them.

This is just a draft. It only remember whether the link is clicked on client side only. It should be server side.

I will need help. I don't even know whether vue is the correct way to go about it. Can you access vue data from the browser with a pluggin?

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -195,6 +195,8 @@ export class Game implements ISerializable<SerializedGame> {
   // Crash Site promo project
   public someoneHasRemovedOtherPlayersPlants: boolean = false;
 
+  public clickedLinks: Array<string> = [];
+
   private constructor(
     public id: GameId,
     private players: Array<Player>,
@@ -417,6 +419,7 @@ export class Game implements ISerializable<SerializedGame> {
       researchedPlayers: Array.from(this.researchedPlayers),
       seed: this.seed,
       someoneHasRemovedOtherPlayersPlants: this.someoneHasRemovedOtherPlayersPlants,
+      clickedLinks: this.clickedLinks,
       temperature: this.temperature,
       unDraftedCards: Array.from(this.unDraftedCards.entries()).map((a) => {
         return [
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -393,6 +393,7 @@ export class Game implements ISerializable<SerializedGame> {
       awards: this.awards,
       board: this.board.serialize(),
       claimedMilestones: serializeClaimedMilestones(this.claimedMilestones),
+      clickedLinks: this.clickedLinks,
       colonies: this.colonies,
       colonyDealer: this.colonyDealer,
       dealer: this.dealer.serialize(),
@@ -419,7 +420,6 @@ export class Game implements ISerializable<SerializedGame> {
       researchedPlayers: Array.from(this.researchedPlayers),
       seed: this.seed,
       someoneHasRemovedOtherPlayersPlants: this.someoneHasRemovedOtherPlayersPlants,
-      clickedLinks: this.clickedLinks,
       temperature: this.temperature,
       unDraftedCards: Array.from(this.unDraftedCards.entries()).map((a) => {
         return [
--- a/src/SerializedGame.ts
+++ b/src/SerializedGame.ts
@@ -23,6 +23,7 @@ export interface SerializedGame {
     awards: Array<IAward>;
     board: SerializedBoard;
     claimedMilestones: Array<SerializedClaimedMilestone>;
+    clickedLinks: Array<string>,
     clonedGamedId?: string;
     colonies: Array<SerializedColony>;
     colonyDealer: ColonyDealer | undefined;
--- a/src/components/GameHome.ts
+++ b/src/components/GameHome.ts
@@ -66,16 +66,9 @@ export const GameHome = Vue.component('game-home', {
     copyUrl: function(playerId: string): void {
       copyToClipboard(window.location.origin + this.getHref(playerId));
       this.urlCopiedPlayerId = playerId;
-      this.game?.clickedLinks.push(playerId);
     },
     isPlayerUrlCopied: function(playerId: string): boolean {
       return playerId === this.urlCopiedPlayerId;
-    },
-    isLinkClicked: function(playerId: string): boolean {
-      return this.game ? this.game.clickedLinks.includes(playerId) : false;
-    },
-    clickLink: function(playerId: string): void {
-      this.game?.clickedLinks.push(playerId);
     },
   },
   template: `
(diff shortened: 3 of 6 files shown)


kberg:
So, a few opinions about this approach: it makes game sharing brittle. A player who loses their one-time link is in trouble. This could be worked around by providing an admin link of some kind.

I might also just consider changing the game's behavior to provide a login. That's NOT EASY but it would really alleveate so much. That's long-term, though.

ainokna:
I like the idea of one-time links!
I think it is a good candidate for being a toggleable option during new game creation.

I personally never saw people losing links in the game that I played, so don't know if it's a frequent issue for other players.

vsrisuknimit:
Can't easily lose a link. The moment you opened it, it is in your browser history.

This is the admin link we are talking about.

bafolts:
I am assuming this change is trying to take the load off of the player who hosts the game. The game host has to send players their links and has to act as an admin if a player were to lose their link. If a game has 6 players the game host has to be prepared to send players their links if they are lost. Assuming the goal of this change is to take this pressure off the game host we could consider another option.

When you click "Create Game" we create all the players and deal them their project cards right away. For multiplayer games we could consider delaying this. The game host would only specify they wanted a multiplayer game. The game admin page would then allow a player to show up and click to add a player until max players was reached. While adding their player they could provide a password. We would store this password with the player and they would have to provide it with their player link. We could add the password protection as a configurable option. If the other players are trusted the password protection is unnecessary security.

Allowing players to add themselves to the game could solve this problem and allow players to name themselves and select their own color. We could also consider allowing the players to modify the game options before the game has started. On the multiplayer game admin page we would add a "Start Game" button that any player could click to start the game.

vsrisuknimit:
> Allowing players to add themselves to the game could solve this problem and allow players to name themselves and select their own color. We could also consider allowing the players to modify the game options before the game has started. On the multiplayer game admin page we would add a "Start Game" button that any player could click to start the game.

That would be fantastic. I was proposing this as a stepping stone. It is mostly to solve the issue of a bad actor on Discord getting into players' links and sabotage their games.

bafolts:
I was giving this more thought today. Instead of one time link we could only provide the player links until every player has finished selecting their corporation card. This would re-use data we already store and I think solve the problem. Players would still be in charge of their player id but after the game has started someone couldn't see the game id posted on discord and find the player links. If some bad actor kept all of the game ids and player ids before the game had started we would still have a problem but I don't think we can solve that until larger re-factor of creating a separate screen where player's register themselves.

vsrisuknimit:
> I was giving this more thought today. Instead of one time link we could only provide the player links until every player has finished selecting their corporation card. This would re-use data we already store and I think solve the problem. Players would still be in charge of their player id but after the game has started someone couldn't see the game id posted on discord and find the player links. If some bad actor kept all of the game ids and player ids before the game had started we would still have a problem but I don't think we can solve that until larger re-factor of creating a separate screen where player's register themselves.

I don't quite follow this suggestion. @bafolts 

One time links just act as a security measure until we have login. And I don't even know if we should have a login. It can be too much troubles if everyone has to sign in for every game.

An alternative is an ability to set up a lock code. Once you are in the game, you can set up a lock code associated with the player id. Anyone who then use this player ID without the lock code would basically be in spectator mode. You can see everything except the hand and action block.

bafolts:
Setting up the lock code for a player is what I failed to get across in my previous comment. I think that is a good solution. We could allow a player to optionally specify it while selecting their project cards. Once it is specified it would be required as part of the player URL as a query string argument. This could very well allow for spectator mode as well but we will have to revisit what we send from the server.

alrusdi:
Let's wait for @kberg's experiment with OAuth? Account + Login is bulletproof, correct and universal solution for all the problems like this.

vsrisuknimit:
@bafolts Any suggestion on how to propagate the change from here to the server? I will try to get this to work first and then try the lock code.

I don't want this to be merged.

bafolts:
I agree that if someone is looking into oauth we may as well wait to see what happens with that.

If doing this as a learning exercise I would track when the player home route is hit on the server and update state to track the visit at that time.

:warning: Below as learning exercise only :warning:

When the link is visited the server will get a request with a url with prefix `/player?id=`. When we see this url we would then have to mark with the game that this player has visited their page.

https://github.com/bafolts/terraforming-mars/blob/main/src/server.ts#L59

For simplicity I would use a `Set<PlayerId>` on `Game` to track this so that anytime the `/player?id=` route is hit we call to add the player id to the set on game.

Then at this point we would have to provide this set through a `Model` to the client to know which links have been provided. If a player id is in this set we also wouldn't expose their player id in any response from the server as to not expose the secret.

vsrisuknimit:
@bafolts Thanks. I will check it out. I will likely push this to the community server where if we are going with OAuth.

vsrisuknimit:
Hmm, so it is now working, sort of.

- Once someone click on the admin page link, player?id is already there. If another user click the link on another machine, the id remains on this browser.
- If the server resets, links are clickable again. Something wrong with the serialization here. Unclear.

src/Game.ts
```diff
@@ -195,6 +195,8 @@ export class Game implements ISerializable<SerializedGame> {
   // Crash Site promo project
   public someoneHasRemovedOtherPlayersPlants: boolean = false;
 
+  public clickedLinks: Array<string> = [];
```
bafolts, line 198: Using `PlayerId` instead of `string` is helpful as when compiling typescript will determine if we are using a `PlayerId` versus any string. Also makes it easier to see what is being stored in the array. Instead of checking if the item exists in the array I would use a `Set` as that will naturally avoid duplicates.

bafolts:
> Once someone click on the admin page link, player?id is already there. If another user click the link on another machine, the id remains on this browser.

You wont be able to avoid this without adding more complexity. This is one of the reason people are waiting for an oauth type solution where the player has a known password they register with.

> If the server resets, links are clickable again. Something wrong with the serialization here. Unclear.

When serializing and deserializing `Game` make sure you read and write the new property. As old data in the database wont have this field you will need to defend against it being `undefined`.

src/Game.ts
vsrisuknimit: I tried Set at first but I had problem with serialization as I mentioned and read that `Set<>` is not serializable.

vsrisuknimit:
> When serializing and deserializing `Game` make sure you read and write the new property. As old data in the database wont have this field you will need to defend against it being `undefined`.

That should just be Game.ts and SerializedGame.ts, right? I thought I did. But I will double check to make sure if I miss anything.

vsrisuknimit:
Here is what I am thinking. 

1. Player has another property called password.
2. There is a lock icon on the sidebar for the playerhome page. Next to (?) help page and (i) game setup details page.
3. This lock icon is revealed and clickable only when the player.password is undefined.
4. Once click, it generates a password, randomly generated string like playerid and gameid.
5. From now on, if a user tries to access http:/.../player?id=xfdg2054dfas . It will just become spectator mode for that player.
6. To be able to take action, you need to attach a password to the url request http:/.../player?id=xfdg2054dfas?pass=rew515asdm3. Clicking on the lock the first time will also redirect the player to this new appended url.

This should not locked anyone out of existing games.
Provide simple security and spectator mode.

One problem I could think of is if a user plays on the phone, locks, and then runs out of battery. They might have a hard time continue playing unless their phone browser is synced with their laptop browser. Even more problematic if they use in cognito on the phone. But I think it is such an edge case that should not matter often.

Issue #2691, opened 2021-02-06
Don't offer standard projects if none are available.


kberg:
At the start of my turn, this was my default option.
![image](https://user-images.githubusercontent.com/413481/107126774-da726f00-687f-11eb-983b-6df6293037ed.png)

Don't show this as the default option if none are available, and make "confirm" hidden.

vsrisuknimit:
I thought the point of showing standard project was so that user know what their costs are and can make enough mc to be able to afford them by activating cards or selling cards.

I think the solution should be:
- Never select standard projects as a default option if none is available. Just go ahead to sell or pass if need to.
- Never show Buffer Gas outside of Solo game with 63TR mode.

bafolts:
Only showing buffer gas on the games where it can be used makes sense. The change to show all available actions was only intended to show ones you couldn't afford at the time with those game rules.

With never showing the standard projects that makes sense to me. The goals of that change were to provide a place where as a player you could see the cost of standard projects. It also ensured the standard project was always in the same place when selecting it. If we hide the row we will be removing the goal of showing the cost of the standard projects. We make a lot of assumptions that the player remembers these costs.

After hiding this row when none are available we should add the cost of these standard projects to the `/help-iconology` page so that at any point a player at least has some way to see the costs of a standard project.

kberg:
Sorry, I wasn't clear. I mean, show them. Make the option visible, but it
should not be the preferred option if no standard projects can be selected.

Sorry --- is it clear yet?

>

sienmich:
Yep, clear. Will do.

kberg:
Great, thanks.

sienmich:
Buffer gas in that PR.

I'm not sure about the 2nd idea though.
Changing the default select option to not the first sometimes seems ugly. Maybe I'll just move those SPs after the `Pass` button? Then they would not be automatically selected. Right?

PR #2700, opened 2021-02-07
StandardProjects option after Pass option

sienmich:
2/2 of #2691

I think just putting the SPs option after the Pass button is the best solution here.

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1853,12 +1853,9 @@ export class Player implements ISerializable<SerializedPlayer> {
       action.options.push(remainingAwards);
     }
 
-    const standardProjectsOption = this.getStandardProjectOption();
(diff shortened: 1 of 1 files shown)


kberg:
> 2/2 of #2691
> 
> I think just putting the SPs option after the Pass button is the best solution here.

Not best, but probably simplest. Stick with it, something better will come along.

bafolts (approved):

kberg:
> > 2/2 of #2691
> > I think just putting the SPs option after the Pass button is the best solution here.
> 
> Not best, but probably simplest. Stick with it, something better will come along.

Sorry, perhaps I wasn't clear -- I was asking for a better solution before merging.

kberg:
Do you think one is doable?

Issue #1607, opened 2020-10-13
Minor: consistently distinguish "in play" vs. "out of play" in UI?


kevinb9n:
It's hard for new players to remember the really subtle rules like how event cards are "not in play". (Ideally they would just always trust the Tag Tracker, but sometimes they're just gonna glance and see "oh I do have 2 science tags" and draft on.)

In the "played cards" section, could we gray out cards that are not in play?

* Played event cards
* Flipped Pharmacy Union (done in #1437)
* Cards linked to Self-Replicating Robots
* (I don't know of other examples.)

Complication: "graying out" is already in use for used Action cards. Would you consider graying out only the interior of the card and leaving the tags solid? That may seem strange, but the benefit is that then (together with the above), whether one looks at the cards or at the tag tracker one should always get the exact same answer. (Or a different means could be used to show the used-ness of the action.)

kberg:
I'm not sure I follow -- can you give a brief, but concrete example?

kevinb9n:
I'm not sure what you mean by a concrete example. I listed all the examples of cards that are not-in-play that I know.

Maybe what wasn't clear is why in-playness matters? I know of four reasons:

* It exactly governs whether the tags should count in your tag tracker or not
* It exactly governs whether any effects on the card are active (such as those of Pharmacy Union, which as mentioned *has* already been given this treatment)
* The Excentric award cares about it
* It governs when a colony tile like Titan *also* changes from not-in-play to in-play. (Picture for example I have dropped Titan Shuttles on my Self-Replicating Robots. That doesn't bring Titan into play because the card is itself not in play; when it does come into play then so does Titan.)

Does this even help, I wonder?

kevinb9n:
Game elements are sometimes dimmed to signify that they are no longer relevant - this could also be done more consistently.

* It dims milestones as they are claimed, but after the third one, the remaining two should be as well
* Likewise even though it doesn't and shouldn't dim awards as they are funded, once three are funded it should dim the unclaimed two
* When the global parameters are all maxed it could dim the global events as a reminder that the turmoil phase doesn't happen at the end.

kevinb9n:
Adding to the list: the ruling policy is only in play during the action phase. I've seen questions about this 3 times in the last month or so. Dimming that element during the other phases could help.

kevinb9n:
If a colony like Miranda is dimmed until it comes into play, it will be more obvious that no one can build a colony there yet.

kevinb9n:
[...]

PR #2702, opened 2021-02-07
Darker unavailable cards instead of grayscale

sienmich:
I have trouble compiling with my eyes which cards are gray and which are not, this change makes it easier to fast notice.

They are darker now (didn't change opacity).
We could also try grayscale + darker if u want, I don't think it's necessary. 

Before:
Screenshot from 2021-02-07 02-12-25
After:
Screenshot from 2021-02-07 02-12-01

Note, it also changes the used blue cards (intended):
Screenshot from 2021-02-07 02-13-05

### DIFF
--- a/src/styles/cards.less
+++ b/src/styles/cards.less
@@ -35,7 +35,7 @@
 }
 
 .card-unavailable {
-    filter: grayscale(1);
+    filter: brightness(0.4);
 }
 
 .nofloat {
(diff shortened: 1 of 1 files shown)


kberg:
I have to say this is a fantastic improvement.

chosta (approved):

kevinb9n:
Cool. I think dimming is the better motif.

If anyone ever wants to discuss consistent principles for how the UI employs dimming, I started that in #1607.

It's orthogonal to the current PR but I'll just mention it because this touches it:

I think we should not dim an action card to show used-ness *except* if we are specifically in the interface for choosing an action to play. When I'm just looking at my played cards in general, its tags, its ability to hold resources, and especially its effects are 100% alive. And those things are *not* alive for played event cards and a flipped Pharmacy, so dimming should be reserved for those cases.

So how should a used action card be depicted otherwise, then? Well... how about putting a player cube on the card?

kevinb9n:
Just going to tag @chosta @sienmich to point out this comment above, since the PR is already closed n'all.

PR #2704, opened 2021-02-07
Cache eslint results. Faster local builds, please.

kberg:


bafolts (changes requested):
This is a great change. My laptop is from 2012 and not getting any newer.

The cache is stored in a generated `.eslintcache` file. Can you add this to `.gitignore` so that it isn't added to a commit?

kberg:
> This is a great change. My laptop is from 2012 and not getting any newer.
> 
> The cache is stored in a generated `.eslintcache` file. Can you add this to `.gitignore` so that it isn't added to a commit?

Yes. Done.

bafolts (approved):

PR #2705, opened 2021-02-07
remove unused property in polling

bafolts:
I have been taking a closer look at the data we send between client and server in an attempt to minimize the bandwidth we use. When looking at the request we use for polling in multiplayer games I discovered a property we aren't using. Eventually I want to minimize this response body to rely on HTTP status codes versus sending any response data as it quickly adds up. Right now we send back roughly 150 bytes every 5 seconds. Over time this quickly becomes megabytes of data. Should be able to use status codes to map to `GO`, `WAIT`, and `REFRESH`. Will do that in separate PR.


kberg (approved):
Nice.

PR #2708, opened 2021-02-08
Add Undermoon Drug Lords Network

kberg:


src/cards/moon/UndermoonDrugLordsNetwork.ts
```diff
@@ -0,0 +1,45 @@
...
+      cost: 2,
+
+      metadata: {
+        description: 'Increase your MC production 1 step per each 2 Colony Rate.',
+        cardNumber: 'M81',
+        victoryPoints: -1,
+        renderData: CardRenderer.builder((b) => {
+          b.production((pb) => {
+            pb.megacredits(1);
+          }).slash().moonColonyRate(2);
```
sienmich, line 26: In 90% of the current cards, the `slash` is inside the production box (which does not make very sense to me, but that's how it is)

kberg: To be clear, are you asking for a change, approving the request, or just making a drive-by comment?

sienmich: I'm not sure :)
I think that's a more general comment to moon cards, it would be nice to change that in my opinion. Do you agree?
Could be changed later, since it's just not the most important layout.
The rest looks good

sienmich (approved):

src/cards/moon/UndermoonDrugLordsNetwork.ts
kberg: image

kberg: Thanks!

sienmich: Screenshot from 2021-02-08 13-13-41

sienmich: Normally the whole equation is inside the production box. But it's not a must, and it makes more sense to me. I think we can leave it for now.
Also, the `Colony Rate` is bigger than tags, so it would require more work

kberg: OH I SEE

kberg: Thanks for poiinting this out.

PR #2712, opened 2021-02-08
Add Rover Driver's Union.

kberg:


src/cards/moon/RoverDriversUnion.ts
```diff
@@ -0,0 +1,43 @@
...
+      tags: [Tags.MOON],
+      cost: 16,
+      requirements: CardRequirements.builder((b) => b.logisticRate(2)),
+
+      metadata: {
+        description: 'Requires 2 Logistic Rate. Raise Logistic Rate 1 step. Increase your MC production 1 step per each Logistic Rate.',
+        cardNumber: 'M78',
+        renderData: CardRenderer.builder((b) => {
+          b.moonLogisticsRate(1).br;
+          b.production((pb) => pb.megacredits(1)).slash().moonLogisticsRate(2);
```
sienmich, line 26: `slash().moonLogisticsRate(2)` vs `Increase your MC production 1 step per each Logistic Rate`
(per 1 or per 2)?

sienmich:
```suggestion
          b.production((pb) => pb.megacredits(1)).slash().moonLogisticsRate(1);
```

sienmich (approved):

src/cards/moon/RoverDriversUnion.ts
kberg: Thanks, good tip.

PR #2714, opened 2021-02-08
Display card discounts #2

sienmich:
Continuation of https://github.com/bafolts/terraforming-mars/pull/2679

It looks like this:
Screenshot from 2021-02-08 23-43-26

Those discount are displayed in:

- Play a card section
- Cards in hand
- Drafting (your discounts)
- Select cards to buy / draw

They are not displayed in for example

- Played cards section
- Select card to add resource

Additionally fixed showing resources only on played cards (opposite to discounts - on not played cards).


chosta (approved):
Let's roll!

kberg:
Can this have a unit test, please? See my comment on the matter. Please let me know if you need some help.

PR #2715, opened 2021-02-09
animate title when waiting for input

### LINKED ISSUE #430: Feature: Update document title upon poll response from the backend

bafolts:
This will close #430 . During multiplayer games animates the document title when user input is required. Most likely better animations. I selected `<^>v`. Suggestions welcome for better character sequences for the animation.

### DIFF
--- a/src/components/WaitingFor.ts
+++ b/src/components/WaitingFor.ts
@@ -67,7 +67,7 @@ export const WaitingFor = Vue.component('waiting-for', {
   },
   methods: {
     animateTitle: function() {
-      const sequence = '^>v<';
+      const sequence = '\u25f7\u25f6\u25f5\u25f4';
       const first = document.title[0];
       const position = sequence.indexOf(first);
       let next = sequence[0];
--- a/src/components/WaitingFor.ts
+++ b/src/components/WaitingFor.ts
@@ -74,7 +74,7 @@ export const WaitingFor = Vue.component('waiting-for', {
       if (position !== -1 && position < sequence.length - 1) {
         next = sequence[position + 1];
       }
-      document.title = next + ' ' + constants.APP_NAME;
+      document.title = next + ' ' + $t(constants.APP_NAME);
     },
     waitForUpdate: function() {
       const vueApp = this;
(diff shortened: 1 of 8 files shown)


src/components/WaitingFor.ts
```diff
@@ -64,6 +66,16 @@ export const WaitingFor = Vue.component('waiting-for', {
...
   methods: {
+    animateTitle: function() {
+      const sequence = '^>v<';
+      const first = document.title[0];
+      const position = sequence.indexOf(first);
+      let next = sequence[0];
+      if (position !== -1 && position < sequence.length - 1) {
+        next = sequence[position + 1];
+      }
+      document.title = next + ' ' + constants.APP_NAME;
```
kberg, line 77: Translateable?

kberg (approved):
Is this running somewhere?

kberg:
Actually -- is this going to mean the title is always flashing in a solo game?

vsrisuknimit:
@bafolt
1) ◴ ◷ ◶ ◵

kberg:
@vsrisuknimit's alternatives are better.

src/components/WaitingFor.ts
```diff
@@ -111,10 +124,15 @@ export const WaitingFor = Vue.component('waiting-for', {
     },
   },
   render: function(createElement) {
+    document.title = constants.APP_NAME;
+    window.clearInterval(documentTitleTimer);
     if (this.waitingfor === undefined) {
       this.waitForUpdate();
       return createElement('div', $t('Not your turn to take any actions'));
     }
+    if (this.player.players.length > 1 && this.player.waitingFor !== undefined) {
```
bafolts, line 133: @kberg this should ensure the title is only animated for games with more than one player. Solo games this wont animate.

kberg: Oh, of course. Excuse me for missing it.

bafolts:
Used the clock which looks much better and added translation wrapper around the `document.title`. I don't think we translate the string `Terraforming Mars` today but will work if and when we do.

PR #2720, opened 2021-02-10
Turning help-iconology into a full help page

vsrisuknimit:
I am turning the help-iconology page into a full on help page with information you could normally find on player-aid cards.

image

Work in progress.


vsrisuknimit:
image

vsrisuknimit:
image

vsrisuknimit:
Good for now. We could add more later. At the moment,I could not think of other stuffs to add unless @kevinb9n has more suggestions. I wrote the game phases tab to be more relevant to the web app (skipping setup for example).

vsrisuknimit:
Not totally happy with the way the tabs and collapsible list look but they serviceable. Tagging @ssimeonoff just in case he has suggestion.

alrusdi:
> Not totally happy with the way the tabs and collapsible list look but they serviceable. Tagging @ssimeonoff just in case he has suggestion.

Did you consider of tabs implementation build-in in our css framework (spectre)? https://picturepan2.github.io/spectre/components/tabs.html

vsrisuknimit:
> > Not totally happy with the way the tabs and collapsible list look but they serviceable. Tagging @ssimeonoff just in case he has suggestion.
> 
> Did you consider of tabs implementation build-in in our css framework (spectre)? https://picturepan2.github.io/spectre/components/tabs.html

No idea what that is and how to use it. Is it already in the repo? I thought we don't like a new library.

vsrisuknimit:
Ah, I see that it is already there. I will look into trying it out.

vsrisuknimit:
@alrusdi I gave it a try. I like my tabs better. For the collapsible list, spectre only has accordion. That is not good since opening one section collapses all others. Users might want to see more than one section at once.

alrusdi (approved):

src/components/help/HelpPhases.ts
```diff
@@ -0,0 +1,142 @@
...
+    <div class="help-page-phases-container">
+
+      <ul>
+
+        <li>
+        <input type="checkbox" id="initial-draft">
+        <label for="initial-draft" class="first"><span v-i18n>Initial Drafting (optional variant)</span></label>
+          <ul>
+            <li v-i18n>Draw 3 corporation cards</li>
+            <li v-i18n>Draw 4 prelude cards <div class="expansion-icon expansion-icon-prelude"></li>
```
bafolts, line 18: Reading through this HTML I spot `div` tags like this which seem like they are not closed. Believe there are `span` as well which are not closed. Going to look through this more tomorrow. Adding these new components to `compile-vue-templates.js` may help catch the HTML issues. Could also consider adding a unit test. The vue unit tests also runs another type checker that may catch these warnings as well.

bafolts (changes requested):

ssimeonoff:
The tabs are nice - however they look, in the future the action buttons should look the same.

src/components/help/HelpPhases.ts
vsrisuknimit: @bafolts I looked through it again. That was the only spot with unclosed div or span. It is kinda of hard to see when writing in one chunk of string.

puuj:
I think putting the keyboard shortcuts here might make sense, and possibly the standard actions (greenery, temperature)?

vsrisuknimit:
image

@puuj, I moved hotkeys to the help-page.

src/components/help/HelpPhases.ts
```diff
@@ -0,0 +1,164 @@
...
+                    <li v-i18n>convert plants to greenery</li>
+                    <li v-i18n>convert heat to temperature</li>
+                    <li v-i18n>claim milestone</li>
+                    <li v-i18n>fund award</li>
+                    <li>
+                      <span v-i18n>trade with 9 mc, 3 Ti, or 3 energy<span> 
+                      <div class="expansion-icon expansion-icon-colony"></div>
+                    </li>
+                    <li>
+                      <span v-i18n>lobby: one is free, 5 mc per delegate after<span>
```
kevinb9n, line 78: Their choice of the homonym "Lobby" for this action (when there is already a *place* called the "Lobby") is such a bad move that I would suggest avoiding and saying "send delegate" instead.

It's possible (though rare) that even your first delegate placed will cost 5 mc, so it's more accurate to just say that they're free if from lobby and 5 mc otherwise.

src/components/help/HelpPhases.ts
```diff
@@ -0,0 +1,164 @@
...
+            </li>
+
+            <li>
+              <input type="checkbox" id="solar-phase">
+              <label for="solar-phase"><span v-i18n>Solar Phase</span></label>
+              <ul>
+                <li><span v-i18n>i. Game end check</span>
+                  <ul>
+                    <li v-i18n>If temperature, oxygen, and oceans are maxed, skip the rest of solar phase to end game phase. </li>
+                    <li v-i18n>(If playing solo, Venus track must be completed as well.) </li>
```
kevinb9n, line 103: This should reference the "Mandatory Venus Terraforming" option.

In solo, the game end trigger is 12/14 generations, no relation to terraforming. But if the game end is triggered, then victory is judged immediately.

src/components/help/HelpPhases.ts
```diff
@@ -0,0 +1,164 @@
...
+                  </ul>
+                <li>
+                  <span v-i18n>ii. World Government Terraforming (start player chooses)</span> 
+                  <div class="expansion-icon expansion-icon-venus"></div>
+                </li>
+                <li><span v-i18n>iii. Colony production</span>
+                  <ul>
+                    <li v-i18n>Trade fleets return</li>
+                    <li v-i18n>Each colony marker goes up one step</li>
```
kevinb9n, line 112: Following Market Manipulation, I'd suggest something like "Each colony tile track is increased 1 step if possible". Or "Each unmaxed..." to avoid the "if possible" at the end.

src/components/help/HelpPhases.ts
```diff
@@ -0,0 +1,164 @@
...
+              </li>
+            </ul>
+        </li>
+
+        <li>
+          <input type="checkbox" id="initial-research">
+          <label for="initial-research" v-i18n><span v-i18n>Initial Research Phase</span></label>
+          <ul>
+            <li v-i18n>Select which corporation, 2 preludes and project cards to keep.</li>
+            <li v-i18n>In turn order, play the corporation and pay 3 mc for each of the kept project cards.</li>
```
kevinb9n, line 33: Recommend "then buy the project cards that weren't discarded", because Polyphemos/Terralabs.

kevinb9n, line 33: At this point, Generation 1 has already begun, so we should be under "each generation" by now.

Officially, the entire setup process is all part of the first generation.

src/components/help/HelpPhases.ts
```diff
@@ -0,0 +1,164 @@
...
+                <span v-i18n>Action phase</span> 
+                (<span v-i18n>Turmoil policy in effect</span>
+                <div class="expansion-icon expansion-icon-turmoil"></div>
+                ) 
+              </label> 
+              <ul>
+                <li><span v-i18n>Take 1 or 2 actions (mandatory 2 actions if playing with fast mode):</span>
+                  <ul>
+                    <li v-i18n>play card</li>
+                    <li v-i18n>use active (blue) card action</li>
```
kevinb9n, line 67: would remove "blue" (corps too).

src/components/help/HelpPhases.ts
vsrisuknimit: 3 mc is fine. People can deal with the exception themselves.

src/components/help/HelpPhases.ts
vsrisuknimit: No need to do Mandatory Venus Terraforming. Very weird variant anyway.

src/components/help/HelpPhases.ts
vsrisuknimit: This is easier to read than each generation but gen 1 is different.

src/components/help/HelpPhases.ts
vsrisuknimit: @bafolts this should be good to go.

src/components/help/HelpPhases.ts
kevinb9n: That's fine too, I just meant that it is what the sentence as written applies to, fine to remove it.

src/components/help/HelpPhases.ts
kevinb9n: Fair enough. Please consider keeping the word "buy" in there, as that is the game keyword that these corps reference.

kberg (commented):
This is a lot of feature work, and thank you, but consider this, what about just adding markdown to the github wiki instead?

src/components/help/HelpPhases.ts
kevinb9n: It is, and ease of reading is an important virtue. But I think it's only a virtue for correct information. In this case, the current text will lead to disappointed Pristar players and confused UNMI players, who will try to report bugs (it already happens).

src/components/help/HelpPhases.ts
```diff
@@ -100,7 +100,8 @@ export const HelpPhases = Vue.component('help-phases', {
                 <li><span v-i18n>i. Game end check</span>
                   <ul>
                     <li v-i18n>If temperature, oxygen, and oceans are maxed, skip the rest of solar phase to end game phase. </li>
-                    <li v-i18n>(If playing solo, Venus track must be completed as well.) </li>
+                    <li v-i18n>(If playing solo, the game is ended after 14 generations or 12 generations if Prelude expansion is used. </li>
+                    <li v-i18n>(In solo mode, the game also end when temperature, oxygen and oceans are maxed. If Venus Next is in use, the gaVenus track must be completed as well.) </li>
```
kevinb9n, line 104: Nope, again, this is not a game end trigger for solo.

kevinb9n:
> This is a lot of feature work, and thank you, but consider this, what about just adding markdown to the github wiki instead?

One key question here is whether this content should track with the version of the server that user happens to be on, or whether users will usually be best served by always seeing the latest version.

src/components/help/HelpPhases.ts
kevinb9n: source for ya:

kevinb9n: Getting your terraforming done early is a nice stretch goal, as it prevents the WG from stealing a TR. Sorry, done responding here now.

vsrisuknimit:
> This is a lot of feature work, and thank you, but consider this, what about just adding markdown to the github wiki instead?

Better for it to be self-contained all in one page as a player aid. No need to link to another Github page. Many users might also be afraid of Github.

src/components/help/HelpPhases.ts
vsrisuknimit: updated. was confused by your earlier statement.

kberg:
> > This is a lot of feature work, and thank you, but consider this, what about just adding markdown to the github wiki instead?
> 
> Many users might also be afraid of Github.

What does that mean?

kberg:
> > This is a lot of feature work, and thank you, but consider this, what about just adding markdown to the github wiki instead?
> 
> One key question here is whether this content should track with the version of the server that user happens to be on, or whether users will usually be best served by always seeing the latest version.

Someone with the engineering skill to make meaningful changes that reflect on the iconology will be able to change the link to their local repo, too.

vsrisuknimit:
> > > This is a lot of feature work, and thank you, but consider this, what about just adding markdown to the github wiki instead?
> > 
> > 
> > Many users might also be afraid of Github.
> 
> What does that mean?

@kberg 
This
image
is not as user-friendly as this

image

kberg:
Nope, not getting it. Using words, what are users afraid of?

>

vsrisuknimit:
> Nope, not getting it. Using words, what are users afraid of?
> […](#)

When users look at the help page, they want to know how the game is played. It works like an IRL player-aid cards. All the information are now together in one placed. If we have a link to Github, they now need to open two tabs. And now they are overloaded with information. They don't care about 'pull request', 'issues', 'discussion', etc.

kberg:
Gotcha. I don't think afraid is the right term, nor do I agree with the
concern, but I'm glad to understand your perspective.

PR #2722, opened 2021-02-10
reverse discount design, option to hide discount

vsrisuknimit:
image

image


vsrisuknimit:
Keeping the second price up high. Blocking card name is better than blocking card text or graphics. I like grayscale better than transparency.

sienmich (approved):
Thanks!

sienmich:
Is there some default value for this option? I think they can be turned on by default.

src/components/Preferences.ts
```diff
@@ -74,6 +74,7 @@ export const Preferences = Vue.component('preferences', {
       'smooth_scrolling': false as boolean | unknown[],
       'hide_tile_confirmation': false as boolean | unknown[],
       'show_card_number': false as boolean | unknown[],
+      'show_discount_on_cards': true as boolean | unknown[],
```
vsrisuknimit, line 77: @sienmich The discount is on by default.

sienmich: Thanks, I was looking for this one

PR #2729, opened 2021-02-11
show metal selection for mining cards

vsrisuknimit:
Does not show anything in hand.

Screen Shot 2021-02-11 at 2 25 47 PM

But once they are placed

image

steel is selected

Screen Shot 2021-02-11 at 2 26 12 PM

or titanium is selected

Screen Shot 2021-02-11 at 2 26 21 PM

What does it look like when trying to copy with Robotic Workforce:
image

I rely on the cardExtraContent like Search For Life.

What I would like to improve:
- I don't want to make bonusResource property for all CardModel but I don't know how to fetch IProjectCard given a CardName.
- The box is slightly off for the Ares t cards. Just a little off center but definitely look okay.

### DIFF
--- a/src/components/card/CardExtraContent.ts
+++ b/src/components/card/CardExtraContent.ts
@@ -1,6 +1,7 @@
 import Vue from 'vue';
 import {CardModel} from '../../models/CardModel';
 import {CardName} from '../../CardName';
+import {Resources} from '../../Resources';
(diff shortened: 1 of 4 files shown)


kberg:
What's the value proposition of this change? I do not understand why it is
useful.

>

PR #2731, opened 2021-02-11
Change RoboticWorkforce test to include preludes, corporation cards, and cards with wild tags.

kberg:


kberg:
nb: without pulling from head, this is the result:

```
  1 failing

  1) RoboticWorkforce
       Has all building cards set up:

      Research Network is missing in builderCardsNames
      + expected - actual

      -false
      +true
```

bafolts (approved):

PR #2733, opened 2021-02-12
Spectator mode

vsrisuknimit:
Beginning of spectator mode. I have no idea what I am doing especially with the server side. I will need help with that.

I can work on the front end for what spectator mode should see.

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -236,8 +236,8 @@ export class Game implements ISerializable<SerializedGame> {
     firstPlayer: Player,
     gameOptions: GameOptions = {...DEFAULT_GAME_OPTIONS},
     seed: number = 0,
-    spectatorId: SpectatorId
-    ): Game {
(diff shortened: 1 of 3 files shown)


vsrisuknimit:
What I imagine:
1) Upon game creation generate spectatorId. A random string like gameId and playerId.
2) You can enter `https://terraforming-mars.herokuapp.com/spectator?id=47367f402d54` to watch a game.
3) As a spectator, you have access to the game board, turmoil, colonies, log, and played cards. Spectator cannot see cards in hand of anyone or take any action.
4) Spectator when click show played cards, don't show the cards above the overview table but always show below.

Perhaps also have super spectator mode that can see players' hands too.

kevinb9n:
> Perhaps also have super spectator mode that can see players' hands too.

+1, long term, each player could see two copyable links at the bottom of their page: the normal spectation link, and a let-them-see-my-cards-too link.

Issue #2689, opened 2021-02-06
Is Aerospace Mission's test flaky?


kberg:
I had a test failure today, 

```
  1) AerospaceMission
       Should play:
     TypeError: Cannot read property 'cb' of undefined
      at Context.<anonymous> (build\tests\cards\community\AerospaceMission.spec.js:1:5009)
```

This came from https://github.com/bafolts/terraforming-mars/actions/runs/543427035 which is now hiding that result. :/

bafolts:
It is flaky! Just failed again.

https://github.com/bafolts/terraforming-mars/runs/1891687411?check_suite_focus=true

Was on the windows machine if this gets lost.

nwai90:
Will check. Likely something about test setup (e.g. if there are not 2 valid colonies to place on)

kberg:
Still flaky: https://github.com/bafolts/terraforming-mars/runs/1904365750?check_suite_focus=true

bafolts:
This is duplicate of #3165 . Answer to this question is yes, the aerospace mission's test is flaky.

PR #2735, opened 2021-02-13
Fix flaky test for AerospaceMission

nwai90:
Closes https://github.com/bafolts/terraforming-mars/issues/2689


kberg (approved):
Thanks!

bafolts:
Awesome! I spent an hour or so trying to recreate this or see what was going wrong last week when it was filed. Cool to see it get fixed.

PR #2739, opened 2021-02-13
Simplify tile placement logging message.

kberg:
image


kberg:
NOTE: for some reasons, the translation didn't catch.

image


What did I do wrong? Do translations work without adjacent spaces?

src/locales/fr/log_messages.json
```diff
@@ -28,7 +28,7 @@
     "${0}'s ${1} production increased by ${2}": "La production de ${1} de ${0} a augmenté de ${2}",
     "${0} received ${1} MC from ${2} owner (${3})": "${0} a reçu ${1} M€ de ${2} (appartenant à ${3})",
     "${0} copied ${1} production with ${2}": "${0} a copié la production de ${1} avec ${2}",
-    "${0} placed ${1} tile on row ${2} position ${3}": "${0} a placé une tuile ${1} sur la ligne ${2} position ${3}",
```
bafolts, line 31: I am not sure this translation key can be updated. The translation uses direct matching so it will look for `${0} ${1} ${2} on row ${3} position ${4}'`.

kberg: Oh, huh. Are the translations wrong?

I would not be surprised, by the way, if many of these translations were orphaned and inaccurate.

bafolts: These could very well be wrong and were never updated. If they were orphaned they will need to be updated. I will check this out.

nwai90:
Some context here: The old log message used to have format `(x, y)` and it caused some confusion due to the ambiguity. This was subsequently changed to `row x position y`  to provide more clarity for users.

kberg:
Really? Huh. Well, I won't push this any further. But, side-note -- imagine
if the coordinates were clickable and took you to a spot on the map. I
think in that case, (x, y) would be less ambiguous (but can wait for
another time.)

>

bafolts:
Separate from the message change I do think this uncovered a bug. Some of the translation keys need to be updated to use `${0} ${1} ${2}`. After that the substitutions need to be translated. Currently we only replace the values.

https://github.com/bafolts/terraforming-mars/blob/main/src/directives/i18n.ts#L10

Since it seems we changed to use this messaging format closing this issue makes sense.

PR #2741, opened 2021-02-13
Update server URL handling

kberg:
Move the massive if bracket into URL path processing.

This improves readability and comprehension, but more important it wil set up for the oauth API, where  more arbitrary query parameters might be attached to URLs. (Specifically, the root path, but why not get some clean code out of it.)

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -43,8 +43,10 @@ zlib.gzip(styles, function(err, compressed) {
 
 function processRequest(req: http.IncomingMessage, res: http.ServerResponse): void {
   if (req.url !== undefined) {
+    const url = new URL(req.url, `http://${req.headers.host}`);
+
     if (req.method === 'GET') {
-      if (req.url.replace(/\?.*$/, '').startsWith('/games-overview')) {
+      if (url.pathname === '/games-overview') {
         if (!isServerIdValid(req)) {
           route.notAuthorized(req, res);
           return;
@@ -52,41 +54,42 @@ function processRequest(req: http.IncomingMessage, res: http.ServerResponse): vo
           serveApp(req, res);
         }
       } else if (
-        req.url === '/' ||
-        req.url.startsWith('/new-game') ||
-        req.url.startsWith('/solo') ||
-        req.url.startsWith('/game?id=') ||
-        req.url.startsWith('/player?id=') ||
-        req.url.startsWith('/the-end?id=') ||
-        req.url.startsWith('/load') ||
-        req.url.startsWith('/debug-ui') ||
-        req.url.startsWith('/help-iconology')
+        url.pathname === '/' ||
+        url.pathname === '/new-game' ||
+        url.pathname === '/solo' ||
+        url.pathname === '/game' ||
+        url.pathname === '/player' ||
+        url.pathname === '/the-end' ||
+        url.pathname === '/load' ||
+        url.pathname === '/debug-ui' ||
+        url.pathname === '/help-iconology'
       ) {
         serveApp(req, res);
-      } else if (req.url.startsWith('/api/player?id=')) {
+      } else if (url.pathname === '/api/player') {
         apiGetPlayer(req, res);
-      } else if (req.url.startsWith('/api/waitingfor?id=')) {
+      } else if (url.pathname === '/api/waitingfor') {
         apiGetWaitingFor(req, res);
-      } else if (req.url.startsWith('/assets/translations.json')) {
+      } else if (url.pathname === '/assets/translations.json') {
         res.setHeader('Content-Type', 'application/json');
         res.setHeader('Cache-Control', 'max-age=' + assetCacheMaxAge);
         res.write(fs.readFileSync('build/genfiles/translations.json'));
         res.end();
       } else if (
-        req.url.startsWith('/assets/') ||
-        req.url === '/styles.css' ||
-        req.url === '/favicon.ico' ||
-        req.url === '/main.js' ||
-        req.url === '/main.js.map'
+        url.pathname.startsWith('/assets/') ||
+        url.pathname === '/styles.css' ||
+        url.pathname === '/styles.css' ||
+        url.pathname === '/favicon.ico' ||
+        url.pathname === '/main.js' ||
+        url.pathname === '/main.js.map'
       ) {
         serveAsset(req, res);
-      } else if (req.url.startsWith('/api/games')) {
+      } else if (url.pathname === '/api/games') {
         apiGetGames(req, res);
-      } else if (req.url.indexOf('/api/game?id=') === 0) {
+      } else if (url.pathname === '/api/game') {
         apiGetGame(req, res);
       } else if (gameLogs.canHandle(req.url)) {
         gameLogs.handle(req, res);
-      } else if (req.url.startsWith('/api/clonablegames')) {
+      } else if (url.pathname === '/api/clonablegames') {
         getClonableGames(res);
       } else {
         route.notFound(req, res);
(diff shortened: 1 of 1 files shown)


kberg:
I've tested this with manual testing of an app, and reading the translation file, but I didn't deal with testing the API.

src/server.ts
```diff
@@ -42,63 +42,101 @@ zlib.gzip(styles, function(err, compressed) {
...
-        req.url.startsWith('/solo') ||
-        req.url.startsWith('/game?id=') ||
-        req.url.startsWith('/player?id=') ||
-        req.url.startsWith('/the-end?id=') ||
-        req.url.startsWith('/load') ||
-        req.url.startsWith('/debug-ui') ||
-        req.url.startsWith('/help-iconology')
-      ) {
+  if (req.url === undefined) {
+    route.notFound(req, res);
```
bafolts, line 46: I don't know why `req.url` would be `undefined`. Using an internal server error versus a not found may make more sense. There is a `wtf` log level in android but for our purposes we should also have a `console.warn` here. Never expect to ever hit this. This is a good thing to have up top and out of the way.

src/server.ts
```diff
@@ -42,63 +42,101 @@ zlib.gzip(styles, function(err, compressed) {
...
+    return;
+  }
+
+  const url = new URL(req.url, `http://${req.headers.host}`);
```
bafolts, line 50: For our purposes the host is not important. I would hard code `http://magichost.com` or some other made up host. This should be slightly more efficient and also show we don't depend on this value and are providing it for the `URL` as a base.

src/server.ts
```diff
@@ -42,63 +42,101 @@ zlib.gzip(styles, function(err, compressed) {
...
-        req.url === '/styles.css' ||
-        req.url === '/favicon.ico' ||
-        req.url === '/main.js' ||
-        req.url === '/main.js.map'
-      ) {
+      }
+      break;
+
+    case '/':
+    case '/new-game':
```
bafolts, line 65: This is a great change in terms of readability as well as performance. Will gain a few CPU cycles comparing the full string versus having to use `startsWith`. :+1:

src/server.ts
```diff
@@ -42,63 +42,101 @@ zlib.gzip(styles, function(err, compressed) {
...
+    case '/api/player':
+      apiGetPlayer(req, res);
+      break;
+
+    case '/api/waitingfor':
+      apiGetWaitingFor(req, res);
+      break;
+
+    // TODO(kberg): move to serveAsset.
+    case '/assets/translations.json':
```
bafolts, line 85: I am pretty sure this route is never hit by our app. It might have been hit by the app at one point before we bundled it. Can probably just remove it. @alrusdi may remember more.

src/server.ts
```diff
@@ -42,63 +42,101 @@ zlib.gzip(styles, function(err, compressed) {
...
-    ) {
+      break;
+
+    default:
+      route.notFound(req, res);
+    }
+    break;
+
+  case 'POST':
+    if (req.url.indexOf('/player/input?id=') === 0) {
```
bafolts, line 139: This should be able to use `url.pathname === '/player/input'` with the new `url` object.

src/server.ts
kberg: A counterargument: how important is this slight efficiency? Will http://magichost.com be sufficiently more clear that the disconnect between that and the definition of a URL object makes sense?

I don't mind, but if you explained a little more, I'd be grateful. If this is going to have a magic string, I'd like to add a comment that makes enough sense to assuage future maintainers.

src/server.ts
kberg: Tell you what: how about I add an additional TODO saying "Someone verify that translations are served through another bit of the code path?" or something?

src/server.ts
kberg: Good catch, thanks.

bafolts (approved):
This change is great.

Would using a `Map` make this too difficult to read? With the `switch` statement we give precedence to whatever paths we happen to check first. Now that we can switch on `url.pathname` we could have.

```ts
// create this map once initially
var getRoutes = new Map<string, (req, res) => void>();
getRoutes.set("/", serveAsset);
getRoutes.set("/solo", serveAsset);

// when request comes in check the maps
if (req.method === GET) {
   if (getRoutes.contains(url.pathname)) {
     getRoutes.get(pathname)(req, res);
   } else {
     defaultGet(req, res);   
   }
}
```

This should still be easier to read then what we have today but decrease the time to look-up a route marginally.

We should be able to add unit tests for this file. If we export `processRequest` we could test it without having to prop up the server and leverage stub `req` and `res` classes. I can prop this up and put one test in place in a separate PR.

src/server.ts
kberg: It's defined this way.

```
/**
 * Only valid for request obtained from http.Server.
 */
url?: string;
```

kberg: Also, this matches existing behavior. I just moved a notFound up top. At last, that's what I _think_ I did.

src/server.ts
kberg: Oh, right. I didn't bother since there was only one post path. Still want it?

kberg (commented):
Thanks for the feedback. I pushed back probably a little more than I should have, so feel free to roll over any of my push-backs.

src/server.ts
bafolts: My comment isn't important. Feel free to disregard. Back in the day the `require("url")` had a `parse` that didn't require providing this base. Needing to provide this base is most likely putting a lot of developers in this situation.

src/server.ts
bafolts: Since we have `url.pathname` I would use it here.

src/server.ts
bafolts: Yes. Your changes are all fine. My question of not knowing why `req.url` would be `undefined` was directed towards myself. Would have to read through `http` module to see when it would be `undefined`. For our purposes it should be defined.

src/server.ts
alrusdi: Yes, it is safe to remove it

kberg: Thanks @alrusdi.

src/server.ts
kberg: > Yes. Your changes are all fine. My question of not knowing why `req.url` would be `undefined` was directed towards myself. Would have to read through `http` module to see when it would be `undefined`. For our purposes it should be defined.

Gotcha, thanks. Resolving.

kberg:
> This change is great.
> 
> Would using a `Map` make this too difficult to read? With the `switch` statement we give precedence to whatever paths we happen to check first. Now that we can switch on `url.pathname` we could have.
> 
> ```ts
> // create this map once initially
> var getRoutes = new Map<string, (req, res) => void>();
> getRoutes.set("/", serveAsset);
> getRoutes.set("/solo", serveAsset);
> 
> // when request comes in check the maps
> if (req.method === GET) {
>    if (getRoutes.contains(url.pathname)) {
>      getRoutes.get(pathname)(req, res);
>    } else {
>      defaultGet(req, res);   
>    }
> }
> ```
> 
> This should still be easier to read then what we have today but decrease the time to look-up a route marginally.
> 
> We should be able to add unit tests for this file. If we export `processRequest` we could test it without having to prop up the server and leverage stub `req` and `res` classes. I can prop this up and put one test in place in a separate PR.

Well, I agree that a map would help. But so would a quality battle tested web server framework. :D

Also, there's a bit of a hassle in doing that with the general-purpose assets/ directory, though it's not that hard to special case, still.

I'd really love to figure out how to unit test this stuff. I think doing that requires pulling processRequest into its own thing -- a lightweight framwork would again, really help.

PR #2742, opened 2021-02-14
Add Darkside Observatory

kberg:


src/cards/moon/DarksideObservatory.ts
```diff
@@ -0,0 +1,69 @@
...
+      'Select card to add 1 Science resource OR 2 Data resources',
+      'Add resource',
+      playableCards,
+      (card) => {
+        this.addResource(card[0], player);
+        return undefined;
+      });
+  }
+
+  play() {
```
nwai90, line 66: Missing `public`?

kberg: Huh. I guess? Seems like it isn't necessary.

kberg: But, done.

nwai90 (approved):

PR #2749, opened 2021-02-14
Add Nanotech Industries.

kberg:


bafolts (approved):
Will eventually want to use static properties. If the moon expansion is very popular it will save memory.

PR #2752, opened 2021-02-14
Rework the send delegate to area action using an options dictionary.

kberg:
This is composed of two commits. Reading each separately will probably make it easier to understand.

### DIFF
--- a/src/cards/turmoil/MartianMediaCenter.ts
+++ b/src/cards/turmoil/MartianMediaCenter.ts
@@ -30,7 +30,7 @@ export class MartianMediaCenter extends Card implements IProjectCard {
             pb.megacredits(2);
           });
         }),
-        description: 'Requires that Mars First is ruling or that you have 2 delegates there.Increase your MC production 2 steps.',
+        description: 'Requires that Mars First is ruling or that you have 2 delegates there. Increase your MC production 2 steps.',
       },
     });
   }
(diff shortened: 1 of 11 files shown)


src/cards/turmoil/MartianMediaCenter.ts
```diff
@@ -30,7 +30,7 @@ export class MartianMediaCenter extends Card implements IProjectCard {
             pb.megacredits(2);
           });
         }),
-        description: 'Requires that Mars First is ruling or that you have 2 delegates there.Increase your MC production 2 steps.',
+        description: 'Requires that Mars First is ruling or that you have 2 delegates there. Increase your MC production 2 steps.',
```
sienmich, line 33: When changing card description, change all the translations, else they are breaking

sienmich: I think better to leave it as it is and in nee PR change the description and translations if u want

sienmich (approved):

src/cards/turmoil/MartianMediaCenter.ts
kberg: Agreed, thank you. I fixed the translation in this PR.

sienmich (approved):
Thanks, i thought there were more than just polish one :)

PR #2753, opened 2021-02-14
Remove old and out of date fields from Serialized turmoil.

kberg:
And fix current global event, which still didn't serialize efficiently.

### DIFF
--- a/src/turmoil/Turmoil.ts
+++ b/src/turmoil/Turmoil.ts
@@ -14,6 +14,7 @@ import {ISerializable} from '../ISerializable';
 import {SerializedTurmoil} from './SerializedTurmoil';
 import {PLAYER_DELEGATES_COUNT} from '../constants';
 import {AgendaStyle, PoliticalAgendasData, PoliticalAgendas} from './PoliticalAgendas';
+import {CardName} from '../CardName';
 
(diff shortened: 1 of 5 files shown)


src/turmoil/Turmoil.ts
```diff
@@ -289,8 +290,10 @@ export class Turmoil implements ISerializable<SerializedTurmoil> {
         // Finally, award Chairman TR
         if (this.chairman !== 'NEUTRAL') {
           const player = game.getPlayerById(this.chairman);
-          player.increaseTerraformRating();
-          game.log('${0} is the new chairman and got 1 TR increase', (b) => b.player(player));
+          // Tempest Consultancy Hook (gains an additional TR when they become chairman)
```
nwai90, line 293: @kberg This is already un https://github.com/bafolts/terraforming-mars/pull/2754 which adds the Tempest Consultancy corp. Is it intended here? Same for comment on line 383 below

nwai90 (approved):

src/turmoil/Turmoil.ts
kberg: That was a mistake. I must have branched that PR from this one. Thanks.

PR #2770, opened 2021-02-15
First step in adjusting Political Agendas: Fix the serialization method.

kberg:
@nwai90 this is a work in progress. I need to spend a little more time making sure this works, and also fix the PoliticalAgendas test (which only tests serailization, how sad.)

But take a look at tell me your thoughts, please. This will make the transition much easier.


src/turmoil/PoliticalAgendas.ts
```diff
@@ -25,66 +25,73 @@ export interface PoliticalAgendasData {
...
 export interface SerializedPoliticalAgendasData {
   currentAgenda: Agenda;
   staticAgendas: Array<[PartyName, Agenda]> | undefined;
+  // Replacing staticAgendas and currentAgenda with agendas, eventually.
+  agendas: Array<[PartyName, Agenda]> | undefined;
+  agendaStyle: AgendaStyle | undefined;
 }
 
 export class PoliticalAgendas {
+  public static randomElement: (list: Array<any>) => any = PoliticalAgendas.defaultRandomElement;
```
bafolts, line 43: Typescript should infer this type so there is no need to use `any`. I would change this to:

```
public static randomElement = PoliticalAgendas.defaultRandomElement;
```

bafolts (approved):

src/turmoil/PoliticalAgendas.ts
kberg: I'm *really* happy with these tests.

kberg:
@nwai90 I think given our prior discussions you're fine if I merge this in. But take a peek at it anyway. The tests are useful.

PR #2771, opened 2021-02-15
Replace localized card discount reporting, moving it to the card. Add Moon Tether.

kberg:
image
image

### DIFF
--- a/src/cards/base/AntiGravityTechnology.ts
+++ b/src/cards/base/AntiGravityTechnology.ts
@@ -6,6 +6,7 @@ import {CardName} from '../../CardName';
 import {CardRenderer} from '../render/CardRenderer';
 import {CardRequirements} from '../CardRequirements';
 
+export const ALL = 'all';
 export class AntiGravityTechnology extends Card implements IProjectCard {
   constructor() {
     super({
@@ -25,6 +26,8 @@ export class AntiGravityTechnology extends Card implements IProjectCard {
       },
     });
   }
+  public cardDiscount = {amount: 2};
+
   public getCardDiscount() {
     return 2;
   }
--- a/src/cards/base/AntiGravityTechnology.ts
+++ b/src/cards/base/AntiGravityTechnology.ts
@@ -6,7 +6,6 @@ import {CardName} from '../../CardName';
 import {CardRenderer} from '../render/CardRenderer';
 import {CardRequirements} from '../CardRequirements';
 
-export const ALL = 'all';
 export class AntiGravityTechnology extends Card implements IProjectCard {
   constructor() {
     super({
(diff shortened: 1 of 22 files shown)


src/models/CardModel.ts
```diff
@@ -3,12 +3,14 @@ import {CardType} from '../cards/CardType';
...
 import {Units} from '../Units';
 import {CardName} from '../CardName';
+import {CardDiscount} from '../cards/ICard';
 
 export interface CardModel {
     name: CardName;
     resources: number | undefined;
     resourceType: ResourceType | undefined;
     calculatedCost?: number;
+    discount?: CardDiscount,
```
bafolts, line 13: This makes sense to me to place on `CardMetadata`. Placing the `discount` on `CardMetadata` will also help with memory per card.

src/cards/base/AntiGravityTechnology.ts
```diff
@@ -6,6 +6,7 @@ import {CardName} from '../../CardName';
 import {CardRenderer} from '../render/CardRenderer';
 import {CardRequirements} from '../CardRequirements';
 
+export const ALL = 'all';
```
bafolts, line 9: I don't think this is used anywhere. Probably leftover from when taking a different approach. Or I am missing something.

bafolts (changes requested):
Separating the discount mapping from `PlayerTags` is good change. The added `cardDiscount` property will either have to use the static memory pattern or be placed on the `metadata` property. It feels like it is metadata property to me so would make sense to place on `metadata`.

src/cards/base/AntiGravityTechnology.ts
kberg: A, removing

src/models/CardModel.ts
kberg: How about making that a follow-up change?

kberg:
> Separating the discount mapping from `PlayerTags` is good change. The added `cardDiscount` property will either have to use the static memory pattern or be placed on the `metadata` property. It feels like it is metadata property to me so would make sense to place on `metadata`.

I'm ... not so sure. But OK.

kberg:
> > Separating the discount mapping from `PlayerTags` is good change. The added `cardDiscount` property will either have to use the static memory pattern or be placed on the `metadata` property. It feels like it is metadata property to me so would make sense to place on `metadata`.
> 
> I'm ... not so sure. But OK.

Do you mean CardMetadata? CardMetadata has mostly been used for rendering. I'd like to keep those separate, and prefer to use the static memory pattern (and replace calls to getCardDiscount where reasonable.)

bafolts (approved):
Using the static memory pattern would also be fine. Doing in a separate pull request is fine.

kberg:
> Using the static memory pattern would also be fine. Doing in a separate pull request is fine.

Already done.

PR #2772, opened 2021-02-16
Split all URL processing into individual handlers. (does not compile, rfc)

kberg:


bafolts (commented):
Having looked at this at a high level it looks like exactly what I had in mind with the `src/routes/` namespace. It is worth getting this to compile.

kberg:
Aye aye.

kberg:
I'm going to abandon this PR and start from scratch so as to make sure individual pieces work. I see what routes was supposed to be, but since you need those functions sometimes, it doesn't seem like the Routes code is a good idea for a base class, but something like a "ReturnCodes" thing.

kberg:
Closing this, as it's being handled in steps.

PR #2777, opened 2021-02-17
Add Beginner mode, which shows default alloy values and milestone and award costs.

kberg:


src/components/Milestone.ts
```diff
@@ -35,16 +37,26 @@ export const Milestone = Vue.component('milestone', {
     toggleList: function() {
       this.showList = !this.showList;
     },
+    getAvailableMilestoneSpots: function(): Array<number> {
+      const count = this.milestones_list.filter((milestone) => milestone.player_name).length;
+      return Array(3 - count).fill(MILESTONE_COST);
```
bafolts, line 42: May be worth making this 3 a constant. When first reading this I wasn't sure why 3 was used. The 3 here would be `AVAILABLE_MILESTONES`.

src/components/overview/PlayerResource.ts
```diff
@@ -56,12 +57,25 @@ export const PlayerResource = Vue.component('player-resource', {
...
+        return tutorialModeOn || this.steelValue > DEFAULT_STEEL_VALUE;
+      case Resources.TITANIUM:
+        return tutorialModeOn || this.titaniumValue > DEFAULT_TITANIUM_VALUE;
+      case Resources.HEAT:
+        return this.canUseHeatAsMegaCredits;
+      default:
+        return false;
+      }
+    },
     isResourceUpgraded: function(): boolean {
```
bafolts, line 73: With the introduction of the `showResourceValue` function the `isResourceUpgraded` can be removed as it isn't used.

src/components/Milestone.ts
kberg: Follow-up? Because I'd apply that in a few places, like Game and Player. That's not appropriate for this change. The first half could go in here, just for the sake of moving code through the pipeline, I'll do it in a follow up.

src/constants.ts
```diff
@@ -1,6 +1,8 @@
 // Base constants
 export const CARD_COST: number = 3;
 export const MILESTONE_COST: number = 8;
+export const AWARD_COSTS: Array<number> = [8, 14, 20];
```
bafolts, line 4: This array can be re-used

https://github.com/bafolts/terraforming-mars/blob/main/src/Game.ts#L534

Should make the function easier to follow in `Game`.

bafolts (approved):

src/components/overview/PlayerResource.ts
kberg: Done.

src/constants.ts
kberg: viz otrher comment, I'd like to do that separately.

PR #2785, opened 2021-02-18
Fixes missing Mandatory Venus Terraforming option in game setup details panel

puuj:
This fixes an issue where selecting "Mandatory Venus Terraforming" as a game option is not reflected in the popup panel showing the game settings. I tested by creating games with and without the mandatory Venus option and it appropriately shows the correct message (or lack thereof).


kberg (approved):
Thank you!

PR #2787, opened 2021-02-18
Expand help page, based on nwai90/ui_improvements.

kberg:


kberg:
I captured everything except the keyboard shortcuts are a little off.

vsrisuknimit:
So much work. Why not put everything on the wiki instead? Easier to maintain that way?

https://github.com/bafolts/terraforming-mars/pull/2720#pullrequestreview-589566041

src/components/help/HelpStandardProjects.ts
```diff
@@ -0,0 +1,31 @@
...
+    Card,
+  },
+  methods: {
+    getStandardProjects: () => [
+      CardName.SELL_PATENTS_STANDARD_PROJECT,
+      CardName.POWER_PLANT_STANDARD_PROJECT,
+      CardName.ASTEROID_STANDARD_PROJECT,
+      CardName.AQUIFER_STANDARD_PROJECT,
+      CardName.GREENERY_STANDARD_PROJECT,
+      CardName.CITY_STANDARD_PROJECT,
```
vsrisuknimit, line 16: missing standard project colony

src/components/help/HelpStandardProjects.ts
```diff
@@ -0,0 +1,31 @@
...
+      CardName.AIR_SCRAPPING_STANDARD_PROJECT,
+      CardName.BUFFER_GAS_STANDARD_PROJECT,
+      CardName.MOON_COLONY_STANDARD_PROJECT,
```
vsrisuknimit, line 19: Separate non official expansion standard projects.

vsrisuknimit (changes requested):

chosta:
@kberg Is there a way to check this somewhere live? Code seems solid, just need to check visuals

src/components/help/HelpIconology.ts
```diff
@@ -0,0 +1,212 @@
+import Vue from 'vue';
+
+export const HelpIconology = Vue.component('help-iconology', {
+  components: {
```
chosta, line 4: I don't think components and methods is required but might be a new setting I'm unaware of. Seems like ghost code to me if its empty

src/components/help/HelpPhases.ts
```diff
@@ -0,0 +1,153 @@
+import Vue from 'vue';
+
+export const HelpPhases = Vue.component('help-phases', {
+  components: {
```
chosta, line 4: ghost code?

chosta (approved):
I agree with @vsrisuknimit's comments about missing colony and separation of expansions.

vsrisuknimit:
> @kberg Is there a way to check this somewhere live? Code seems solid, just need to check visuals

@chosta You can check on the community server. It is mostly the same just a few different names for things.

https://github.com/nwai90/terraforming-mars/pull/16
https://tfm-community.herokuapp.com/help-page

chosta:
> @chosta You can check on the community server. It is mostly the same just a few different names for things.

Lit!
That is some Pro-work

I am not sure how I feel about the accordion effect though. It's 2021 and jquery leftover fade effects should be srangled to death :)

kberg:
****

> @kberg Is there a way to check this somewhere live? Code seems solid, just need to check visuals

Ah right I forgot: terraforming-mars-kberg.herokuapp.com

kberg:
> So much work. Why not put all these pages on the Github wiki instead? Easier to maintain that way?
> [#2720 (review)](https://github.com/bafolts/terraforming-mars/pull/2720#pullrequestreview-589566041)

I assume by using a quote you're not necessarily asking the question, but I'll answer it as if you asked it.

The good news is the work is done (thanks!), so the work for me was moving it between repositories. My general criticism of your proposal was that your answer to this was that people were _afraid_ of github. I still disagree with that reasoning, but the work is done, so making it available to everybody is a win, especially if the only cost is moving code between repositories, and cleaning it up along the way.

That reasoning isn't applicable everywhere, like, when moving infrastructure code, so that wasn't meant to be a generally applicable statement.

vsrisuknimit:
@kberg you did not answer why not the wiki? Easier to maintain that way. That was your argument in my original PR.

kberg:
> So much work. Why not put all these pages on the Github wiki instead? Easier to maintain that way?
> [#2720 (review)](https://github.com/bafolts/terraforming-mars/pull/2720#pullrequestreview-589566041)
> 
> "What's the value proposition of this change? I do not understand why it is useful."
> [#2729 (comment)](https://github.com/bafolts/terraforming-mars/pull/2729#issuecomment-778672446)
> 
> "Don't add additional complexity, please. Gamers aren't going to need them, what's the meaningful extra value?" Perhaps just link to a BGG? They can find everything there.
> [#1637 (comment)](https://github.com/bafolts/terraforming-mars/pull/1637#discussion_r508213490)

I did say all those things, and in their contexts I stand behind them, including the one this PR is based on (which, of course, as we know, is mostly your work.) I responded to that one separately.

But two of those things are _not_ related to this PR, and so I have to ask why you chose to do that. As I mentioned to you in chat, there's a professional skill software engineers learn, to criticize the problem, not the person. Due to the structure of the comment, and that you put all of those in one place, I'm going to assert that you're not criticising the problem, but the person, that person being me. I further assert that your compilation here was a kneejerk emotional response, and by showing them all at once is your way of letting me get a taste of my own medicine.

My failure was failing to recognize that people not trained to understand this problem/person thing will find it difficult or unpleasant. As you suggested in chat, I should take a look at that, and as I have told you, I have!

Another thing we're taught professionally is that criticizing the problem can be done in public, criticizing the person is something to do in private. Consider that ongoing when airing grievances in public like this, and whom it serves well. Alterantively, as I've said to you before, go straight to Brian if you can't find relief from your concerns elsewhere.

vsrisuknimit:
"What's the value proposition of this change? I do not understand why it is useful."
https://github.com/bafolts/terraforming-mars/pull/2729#issuecomment-778672446

If you stand behind those statements, please do. But first read the PR and look at the images. It already explained by pictures what the PR tried to do (allowing user to know which card to copy). What is the question for then?

And in the PR where I first clashed with you, you already admitted you were wrong and misunderstood the PR. Do you still stand behind the statement?

It is constantly like this. Not understanding what the users want and just keep resisting and pushing back with sharp words.  Words have weight, @kberg. You can do what you want (OAuth) without wanting to listen to users, but when others try to push for changes that people requested, always pushback.

kberg:
@vsrisuknimit this is not the place for this discussion; it's not about this PR. Let's continue this offline.

chosta:
Guys, you should settle the dispute in a private chat. You are both great at what you do and I'm sure you'll find the strength to atually listen what the other party has to say. I'm actually glad that you are passionate about the project and care about it's present and future but I feel like it became too personal. The PR threads are not the right place to clear out the differences :)

vsrisuknimit:
"I don't think his reasoning is based on software development culture as much as software development in general. Any line of source code we add could potentially cause a bug that prevents the app from running. Anything using vue is also something else we have to update if we were to update vue or change to something else like react at some point. **Putting that page in wiki would avoid all that overhead**." - @bafolts (when I asked about the your resistance and your argument on the original PR.)

So I am still waiting to read about why it was bad when I did it and okay when you do it.

PR #2791, opened 2021-02-18
Show metal choice for Mining Cards

vsrisuknimit:
Problem: Once you played Mining Rights or Mining Area, you don't know what production you gained from it (other than reading the log). Other players may want to know that. If you play Robotic Workforce, you would like to know that so you can copy correctly. I added hide tile/show tile button to solve this problem but only partly. It still requires work to figure out what metals are underneath the tiles. Another problem is if you have one tile on steel and another on Ti, when you play Robotic Workforce you have no idea which is which.

Solution: I utilize extracardcontent.ts (the same way Search For Life shows the alien) to display a production box background over the other metal that was not mined. The position is carefully adjusted to work for both normal and Ares cards.

Potential future problem: If we change the way the cards look drastically, we will need to adjust the location here. Alternative method is to intercept the metadata and rerender it differently. But that is a lot of work and likely has to be done card by card.

Screen Shot 2021-02-18 at 5 49 48 PM


src/components/card/CardExtraContent.ts
```diff
@@ -13,8 +14,26 @@ export const CardExtraContent = Vue.component('CardExtraContent', {
     lifeFound: function(card: CardModel): boolean {
       return card.name === CardName.SEARCH_FOR_LIFE && card.resources !== undefined && card.resources > 0;
     },
+    isMiningTileOnMetal: function(card: CardModel, metal: string): boolean {
+      const miningCard = [CardName.MINING_RIGHTS, CardName.MINING_AREA, CardName.MINING_RIGHTS_ARES, CardName.MINING_AREA_ARES];
+      if (miningCard.includes(card.name)) {
+        if (metal === 'titanium') {
```
chosta, line 20: You can add an enum for metals

chosta (approved):
There is nothing wrong with your approach. For a minor addition like this we can live with the possibility of having to rework the feature if card design were to change.

src/components/card/CardExtraContent.ts
vsrisuknimit: replaced with resources enum

PR #2792, opened 2021-02-18
Improve Turmoil send delegate UI

### LINKED ISSUE #1689: Show parties when sending delegates
The way colony implementation has been updated is amazing. We should do the same for parties when sending delegates.

Instead of words and descriptions, just copy the entire party section whenever a player is sending / removing delegates. This would speed things along and not make the player look in two different locations when they're trying to make their decisions.

vsrisuknimit:
Problem: Sending delegate to party action is not intuitive. Too much text, not enough info.

This PR can close #1689.

Changes:
- Remove orOptions and Option from Turmoil. Added a new class based on SelectColony
- Display each turmoil party as selectable objects similar to how cards and colonies are done
- Dim the party that is not selectable (only for Recruitment)

Note:
- This does not work with Banned Delegates which does not utilize SendDelegateToArea. It requires not just selecting a party but selecting a delegate.

Working demo:
Check community server: https://tfm-community.herokuapp.com/

Example:

image

image


chosta (commented):
Wow, you are getting good at this :)
Front-end wise I don't see anything that bothers me but I'd like someone else to check the server logic before approval.
The only thing that really bothers me are the red arrows showing direction. This has to be a different color in my opinion. The direction is there to help and not to ring an alaram.  I'd choose one of the 50 shades of grey for it.

src/components/SelectPartyToSendDelegate.ts
```diff
@@ -0,0 +1,67 @@
...
+      type: Boolean,
+    },
+  },
+  data: function() {
+    return {
+      selectedParty: undefined as string | undefined,
+    };
+  },
+  components: {
+    'Button': Button,
```
chosta, line 29: This could be just  {Button, Party}

vsrisuknimit:
image

image

@chosta

src/components/SelectPartyToSendDelegate.ts
vsrisuknimit: done

alrusdi (approved):
Everything seems to be allright. Let's merge it

vsrisuknimit:
Since I have no idea what I am doing on the backend, let's wait for @bafolts or @kberg to or @Lynesth to take a look too.

src/server/ServerModel.ts
```diff
@@ -215,6 +216,7 @@ function getCorporationCard(player: Player): CardModel | undefined {
 function getWaitingFor(
   player: Player,
   waitingFor: PlayerInput | undefined,
+  game: Game | undefined,
```
bafolts, line 219: It isn't necessary to pass `game` as an argument. The `player.game` property should be used.

bafolts (changes requested):
We have moved away from passing `game` argument separately. If the `player` argument is provided we should use `player.game`. We are trying to do this throughout the code. In `ServerModel` can you revert the introduction of `game` argument to the `getWaitingFor` function and use `player.game`?

src/deferredActions/SendDelegateToArea.ts
```diff
@@ -39,35 +36,35 @@ export class SendDelegateToArea implements DeferredAction {
...
 
-    sendDelegate.options = parties.map((party) => new SelectOption(
-      party.name + ' (' + party.description + ')',
-      'Send delegate',
-      () => {
-        if (this.options.cost) {
-          this.player.game.defer(new SelectHowToPayDeferred(this.player, this.options.cost, {title: 'Select how to pay for send delegate action'}));
-        }
+    const availableParties:Array<PartyName> = [];
+    parties.forEach((party) => availableParties.push(party.name));
```
bafolts, line 43: Not a big deal but this is equivalent to `const availableParties = parties.map((party) => party.name);`. Using the `map` should be more efficient as well.

vsrisuknimit: done

src/server/ServerModel.ts
vsrisuknimit: removed game

bafolts (approved):

PR #2793, opened 2021-02-19
Hide un-actionable standard projects

### LINKED ISSUE #2691: Don't offer standard projects if none are available.

vsrisuknimit:
Situation: #2664 makes standard projects available at all time. This can be useful for new players but unnecessary and create visual overload for experienced players who can remember all standard project cost by heart. It also create the problem in #2691. Standard project is offered as the first choice when none is available.

This PR can close #2691.

Changes:
- Rename tutorial mode to learner mode. @kevinb9n mentioned tutorial mode is not exactly the right name. 'Beginner mode' sounds too similar to beginner corps, hence learner mode for people still learning how to play may need the extra info that experienced players can find unnecessary.
- Use the learner mode to add an extra class to standard project cards to hide them if they are not actionable.
- If no standard project is available at all, don't add the choice in components/orOptions

Little problem.
- It does not self-select pass as the first choice option. Not a big deal, but it would be nice if we can save one more click.
- I try to play around with how the inputs are generated instead but that is on server side has no access to PreferenceManager.
- I played around with idx but the cause sell to become pass, sell to becomes no action. Shift up 1 or down 1. I am not sure how to deal with it. The current solution is just not add the element when standard projects are unavailable and skipped.

Examples:

When learner mode is on:
image

When learner mode is off, and no standard projects is available:
image
Note that pass is not auto-selected.

When learner mode is off and some standard projects are available:
image


src/components/OrOptions.ts
```diff
@@ -119,29 +120,45 @@ export const OrOptions = Vue.component('or-options', {
...
+      if (option.title === 'Standard projects' && option.enabled !== undefined ) {
+        if (option.enabled.every((p: boolean) => p === false)) {
+          const learnerModeOff = PreferencesManager.loadValue('learner_mode') === '0';
+          if (learnerModeOff) {
+            skippedStandardProject = true;
+          }
+        }
+      }
+
+      // Only push this orOption element if we are not skipping standard projects
```
sienmich, line 135: I think it would be simpler and cleaner to not change OrOptions but to add one `if` in `Player.ts` when adding SPs as an option. That would be one line addition. What do you think?

vsrisuknimit: I do not think you can access PreferenceManager from Player.ts. That is server-side? But I did not try very hard.

sienmich: I see. Still, I think hiding the SPs card manualy is ugly in this case, this could lead to bugs in the future. I'll try to propose some better solution, ok?

sienmich: Maybe the Preference Menagerie should be accessible from the Player

vsrisuknimit: If we can add a new property (learner_mode) to player and change the value along with the PreferenceManager's value. But I think the current method is cleaner. All we do is not adding an element that is not going to be used. 

But I could always get behind a cleaner solution.

vsrisuknimit: The problem with putting something on the server-side like in player.ts is also that the learner mode preference would not be persistent across many games.

bafolts: > All we do is not adding an element that is not going to be used.

This is true. But what happens when there is another option we don't want to show in `learner_mode`? With this strategy we would have to continue to have places throughout the code where we check `learner_mode` each place we want to hide the option. For example, what if within an `AndOptions` we wanted to have this logic in the future? We would have to copy this code into `AndOptions` as well.

Adding a property for `learner_mode` to `Player` would make sense to me. In this situation the `Player` has not selected the beginner corporation but they are in learning mode. Regardless of which client they are using I assume they would want to be in `learner_mode`. To associate the property with the `Player` another strategy would be to provide in the query string to the `/player/input` route an additional query string argument. The url would then become:

`/player/input?id=playerid&learnerMode=true`

On the server when we are processing player input we would then know that we are processing the input for a learner. At this point could store on `Player` that the user is a learner. From there once we know the `Player` is a learner we can hide this option and in the future hide or show more options with ease. All the wiring would already be in place. I don't think there would be a need to serialize learner mode but could also add it to serialization.

vsrisuknimit: Adding learner mode as a property seems fine to me if that is how we want to go. Some questions
1) Some players may want to turn on learner_mode without playing Beginner Corporation. How do we set the value? Do we still have learner_mode value under PreferenceManager?
2) Can you update this value in Player mid way through the game through Preferences? As soon as you toggle or after clicking okay?

I am not sure how to do it with query string. I will leave that to someone else to try, but I can attempt to play around with adding the property to player and set things up Beginner Corp.

bafolts: Any player interaction goes through the `WaitingFor` component. The POST request is made [here](https://github.com/bafolts/terraforming-mars/blob/main/src/components/WaitingFor.ts#L147). On this URL is where you could provide the query string argument.

The request ends up on the server [here](https://github.com/bafolts/terraforming-mars/blob/main/src/server.ts#L164). From here you could set the property. The property would then be available on the server.

If a user were to toggle the learner mode preference after the server had already sent the option they would have to refresh the page to see the option. I think this is fine since someone with learner mode shouldn't be continually changing the preference during game play.

vsrisuknimit: What should happen after the toggle in Preferences? Redirect directly with `window.location.href`?

bafolts: Including a message that the user may need to refresh the page for their changes to take effect would be good enough for me. I wouldn't expect this option to change often. During a game if you were in learner mode I would assume you would be in learner mode until the end of that game.

vsrisuknimit: What I don't understand is how the toggle will propagate the data all the way to the server without changing the query url and how to do that with just manual refresh?

bafolts: Sorry if this feels like I am going back and forth but giving it some more thought I do think this will want to remain on the client side. I was trying to come up with a good answer to your question and was unable to. Updating the server after it has already created the "waiting for" would have more repercussions than anything we may happen to introduce on the client with this change. I will take another look at this with client side mindset. This will need support for one option within an `OrOptions` to be hidden. If there are only 2 total options and hiding one of them leaves us with one option we would need a way to display that. If the only available option in an `OrOptions` were hidden we would have to also have some user interface for the user. In that situation we would most likely need to display the option to the user anyway. We can only hide the option if there is at least one other option to pick.

src/models/PlayerInputModel.ts
```diff
@@ -14,6 +14,7 @@ export interface PlayerInputModel {
     canUseSteel: boolean | undefined;
     canUseTitanium: boolean | undefined;
     cards: Array<CardModel> | undefined;
+    enabled?: Array<boolean>;
```
bafolts, line 17: Instead of passing `enabled?: Array<boolean>` I think we should pass `learnerMode`. This would indicate that this input is meant for players using learner mode. Similar to `showOwner` I would again use `learnerMode?` and make the property optional so that we are only sending it when it is needed. When creating the `SelectCards` for the standard projects if all of the cards are disabled we would set `learnerMode` to `true`. With this in place in `OrOptions` we then only need to consider what to do when any option has `learnerMode` set to `true`.

src/components/OrOptions.ts
vsrisuknimit: Currently, the only time where you are offered an option that is un-actionable is here, only for standard projects due to #2664 . I do not intend for this hiding effect to take place in anything else.

So perhaps I should make this more specific to that.

src/models/PlayerInputModel.ts
vsrisuknimit: Does `learnerMode` need passing since it is already fetched from PreferenceManager from anywhere client side? `enabled?: Array<boolean> ` is still needed because that is how #2664 tell the client side which standard project is affordable.

bafolts: The `learnerMode` property from the server would indicate to the client that this `PlayerInputModel` should only be shown when the client has selected the `learner_mode` preference. With that in place `enabled?: Array<boolean>` could be moved server side to set the `learnerMode` property to indicate that when every card is not enabled then the option should only be shown in learner mode.

vsrisuknimit: I don't think I follow. `enable` is both server-side and client-side. Server side it tells which standard projects are available, actually actionable. Client-side it tell vue what cards to gray out. I don't think we can have it server-side only.  You are trying to replace enabled<> array with learnerMode when the array is all false. But the array is not always all false.

bafolts: > Client-side it tell vue what cards to gray out.

This is handled through [isDisabled](https://github.com/bafolts/terraforming-mars/blob/main/src/models/CardModel.ts#L16). 

I am not trying to _replace_ the enabled property on `src/inputs/SelectCard`. Instead of sending this data to the server on `PlayerInputModel` I am recommending to introduce a new property of `learnerMode`. With a property for `learnerMode` on `PlayerInputModel` we could re-use this property for any `PlayerInputModel` in the future that we only wanted to show the user client side if they were in learner mode.

vsrisuknimit: Ah, you are certainly right. Sorry, it has been a while ago since I made the PR so I sort of forgot. I will work on it tonight.

vsrisuknimit: @bafolts I removed `enabled` and added `showOnlyInLearnerMode` that is calculated at ServerModel.

vsrisuknimit:
image

Still a little problem of the 'pass' not being preselected. I am not quite sure how to fix this here.

src/components/OrOptions.ts
```diff
@@ -119,29 +120,45 @@ export const OrOptions = Vue.component('or-options', {
...
-                props: {
-                  title: $t(option.buttonLabel),
-                  type: 'submit',
-                  size: 'normal',
-                  onClick: () => {
-                    this.saveData();
+
+      // Label with title standard project will be skipped if no suboption is available and learner mode is off.
+      let skippedStandardProject = false;
+      if (option.title === 'Standard projects') {
```
bafolts, line 126: It shouldn't matter what the title is. As long as `option.showOnlyInLearnerMode` is `true` we can run the logic to skip this option. The variable name `skippedStandardProject` can be more generic and could be `skippedOption`.

src/components/OrOptions.ts
```diff
@@ -119,29 +120,45 @@ export const OrOptions = Vue.component('or-options', {
...
+        if (option.showOnlyInLearnerMode) {
+          const learnerModeOff = PreferencesManager.loadValue('learner_mode') === '0';
```
bafolts, line 128: The `loadValue` function returns the empty string if this value is not set and not `0`. I would add a `PreferencesManager.loadBooleanValue('learner_mode')` to the `PreferencesManager` to avoid having to check what the value happens to be.

vsrisuknimit: done. and simplified the check.

src/components/OrOptions.ts
vsrisuknimit: removed. I was hoping this would apply just to standard projects and nothing else. The check is to help with that, but showOnlyInLearnerMode should be false or undefined most of the time.

bafolts (approved):

src/components/OrOptions.ts
bafolts: That `loadBooleanValue` will go a long way for other places we are checking values for `=== '1'`

PR #2794, opened 2021-02-19
Move buffer gas out of Prelude Manifest

### LINKED ISSUE #2788: Buffer Gas SP availability should depend only on TR63 mode, not expansions
There's no connection between TR63 mode and the Prelude expansion, or between the Buffer Gas standard project and the Prelude expansion.

It just happened that these rules appear physically in the Prelude rulebook, but this fact is of no consequence.

Currently (as reported by Kowal on discord), the SP is missing when Prelude isn't in use.

vsrisuknimit:
This PR can close #2788 . It allows buffer gas to be available when Prelude is not used, as long as 63TR mode is on.


kberg (approved):
This makes sense in retrospect.

sienmich:
Just pointing out it's against the official rules (where `Buffer Gas` was introduced in Prelude expansion). Still, good for me.

sienmich (approved):

vsrisuknimit:
> Just pointing out it's against the official rules (where Buffer Gas was introduced in Prelude expansion). Still, good for me.

@sienmich see #2788

sienmich:
> @sienmich see #2788

Ok!

PR #2797, opened 2021-02-19
send all log messages from server, simplify LogPanel

bafolts:
This changes the log route to always return every game log. The JSON response is now gzip encoded which should limit the bandwidth we are sending. Eventually will also add `br` but will wait to see `gzip` working first. With every log message coming back from the server the `LogPanel` component now filters by the available generations. This should allow us to bridge the gap with the production data that only has the last 50 log messages.

I tested this on a new game as well as a game running on terraforming-mars.herokuapp.com. The games missing the game log entries show as many generations as they have available.


src/server.ts
```diff
@@ -528,7 +528,7 @@ function serveAsset(req: http.IncomingMessage, res: http.ServerResponse): void {
     const compressed = fileCache.get('styles.css.gz');
     contentType = 'text/css';
     file = 'styles.css';
-    if (compressed !== undefined && supportsEncoding(req, 'gzip')) {
+    if (compressed !== undefined && Route.supportsEncoding(req, 'gzip')) {
```
kberg, line 531: Moving 'Route' to a static class isn't a bad idea, between you and me that will happen naturally.

But this method is very different from all the other Route methods; this one is just about an incoming request, all the other ones are about outgoing responses. Is this the right place for this method?

kberg: And also as serverAsset becomes its own handler, this function will fit perfectly inside it. For now, no change, I suppose, but I'll probably move Route.supportsEncoding in ServeAsset.

src/components/LogPanel.ts
```diff
@@ -237,11 +239,10 @@ export const LogPanel = Vue.component('log-panel', {
     },
   },
   mounted: function() {
-    fetch(`/api/game/logs?id=${this.id}&limit=${raw_settings.logLength}&generation=${this.selectedGeneration}`)
+    fetch(`/api/game/logs?id=${this.id}`)
```
kberg, line 242: Every time an event occurs the entire log will be sent down? This sounds ... inefficient and expensive. What are your thoughts? I suspect logs for a single generation are still valuable as the game generally proceeds. But that can be reinstated as an optimization, because I can see edge cases where that can be done wrong.

kberg (commented):
I approve, but if I click 'Approve' someone might merge it. The ui portion deserves someone else's opinion.

chosta:
I quite like the direct approach. The pause/resume functionality was the wrong path to take, I should have considered discussing the options first. This seems ready to go.

chosta (approved):

alrusdi (approved):

src/components/LogPanel.ts
bafolts: This isn't ideal but we are missing generation data in production due to a bug where we end up with only the last 50 messages. After 10 days have passed the games with the missing logs should work their way out of the database through our purge process. This is inefficient and expensive but I can't think of a cheaper way until the games with the missing log data are out of production. Since some games right now are missing generational data, error handling would have to be implemented which isn't needed once these games are purged.

If this goes out and proves to increase CPU dramatically we could write the gzipped JSON for the log data to memory so that when this route it requested we only have to pull the logs from memory like any other asset. This is expensive but I think this provides enough value to games that it is worth the cost. When I first added the separate log route I wish I hadn't had the bug where only the last 50 messages were kept. This would at least have shown what the memory impact was going to look like first. With string interning I don't _think_ the memory impact will be that significant since most of the log messages are strings we re-use often.

src/server.ts
bafolts: I moved this to what seemed like the best fit at the time since I wanted to re-use the function and didn't want to import it from `server` creating a circular dependency. I agree this isn't ideal. Moving this to another place is recommended. Until this change is merged I can't update terraforming-mars.herokuapp.com without breaking currently running games so there was some urgency behind this.

kberg: > I moved this to what seemed like the best fit at the time since I wanted to re-use the function and didn't want to import it from `server` creating a circular dependency. I agree this isn't ideal. Moving this to another place is recommended. Until this change is merged I can't update terraforming-mars.herokuapp.com without breaking currently running games so there was some urgency behind this.

Perfectly fine.

src/components/LogPanel.ts
kberg: We shall see!

kberg (approved):

bafolts:
Was hoping to have all the log messages stored when the log route was added. Didn't realize we were only storing the last 50 log messages until looking at the @chosta PR. We have a high number or games played on some server but most of them are single player games. In a single player game you would have to log a lot per generation to end up with over 50 log messages. I will keep an eye on this throughout the day once this is merged.

The new log route using gzip is expensive for a CPU task but the route should only be hit on a page refresh or if another player has made a move. Heroku has an unwritten 2 terabyte limit so I added the gzip as it dramatically cuts down on the payload size. This may be an early optimization which backfires.

For users running the server locally or casually I don't expect any performance issues. For high traffic servers could very well be another PR from me in due time.

PR #2798, opened 2021-02-19
Card ownership 

### LINKED ISSUE #629: Show cards ownership for actions
When you do an action involving more than one card to add resources to it is hard to differ your cards from opponent cards. It is error prone behavior. We should show cards ownership in this case.

vsrisuknimit:
This can close #629 .

Work in progress. No need to review yet.


vsrisuknimit:
Too messy. I have a better PR coming up.

PR #2800, opened 2021-02-20
Replace the (admittedly, way more functional) card UI with debug ui, which is more up to date.

kberg:
We discussed this briefly on discord, but I wanted to bring  it here for further discussion. Yes? No?


src/components/DebugUI.ts
```diff
@@ -172,7 +172,10 @@ export const DebugUI = Vue.component('debug-ui', {
   },
   template: `
         <div class="debug-ui-container" :class="getLanguageCssClass()">
-            <h1>Debug UI</h1>
+            <h1>Cards List</h1>
+            <div style="position: absolute; top: 0; right: 0; margin: 12px">
```
alrusdi, line 176: No inline styles please!

alrusdi (changes requested):

src/components/DebugUI.ts
kberg: Of course, mea culpa. Thank you.

alrusdi (approved):

PR #2801, opened 2021-02-20
Add card ownership

### LINKED ISSUE #629: Show cards ownership for actions

vsrisuknimit:
This PR can close #629.

It add a 'showOwner 'property to input/SelectCard and InputModel. Only RemoveResourcesFromCard will specify this value to true. 

If 'showOwner' is true, components/SelectCard activates a method to find the owner and pass it to components/Card.ts to display.

One potential problem: the name will wrapped to the next line. If someone has incredibly long name, the label might get big enough that it overlaps the next cards below. I cannot change the padding on the go. I need to apply padding to cardbox.

Picture: 

normal SelectCard
image

SelectCard called by RemoveResourcesFromCard
image


vsrisuknimit:
image

image

Still deciding on the style.

I like 'below the card' because it is next to the resource count and VP per count. Often time, you decide to eat something so that the resource count goes below threshold. Made a poll on Discord now.

bafolts (commented):
Every card has an owner that we could calculate client side. Calculating the value client side would save resources on the server. So this part seems fine.

In terms of the `showOwner` property. While looking at this I am wondering. Could we also calculate that we want to show the owner? Any time as a player if you are selecting from cards where you are not the owner it makes sense to me to show the owner. Could we also calculate `showOwner` client side? If as a player you are not the owner of _any_ card in `SelectCard` we should show the owner. This would avoid having to manually add this everywhere we want to use it.

vsrisuknimit:
@bafolts RemoveResourcesFromCard is only place we would need this. 

We could calculate what you mentioned but that is not enough. If you use Ants and then only you have microbes, you should still show the ownership. If I did not misunderstand what you wrote, you would not want to show it in that case. But we need to show ownership for Ants even if the only targets are our own cards, just so we know they are our own cards.

vsrisuknimit:
@bafolts Alternatively I could go by the title of SelectCard instead. That is what I thought at first. But I was afraid that it would be too finicky.

src/deferredActions/RemoveResourcesFromCard.ts
```diff
@@ -45,7 +45,7 @@ export class RemoveResourcesFromCard implements DeferredAction {
         const owner = this.player.game.getCardPlayer(card.name);
         owner.removeResourceFrom(card, this.count, this.player.game, this.player);
         return undefined;
-      },
+      }, 1, 1, false, undefined, true, true,
```
bafolts, line 48: After giving this a lot of thought I do think that having `showOwner` property is probably among the best options here. What I don't like about this line is having to provide every other argument's default value in order to finally provide the 6th extra argument's value of `true`. If we ever changed these defaults we would have to come here and update all these defaults.

There is no reason we have to continue to set every property of `SelectCard` through it's constructor. Can you set `showOwner` after creating the instance of the class? This will avoid having to set all these arguments. What you have would become something like:

```
const selectCard = new SelectCard(this.title, 'Remove resource(s)' /* every argument we have prior to your change*/);
selectCard.showOwner = true;
```

This will avoid having to set so many arguments. I would also make the `showOwner` argument optional. Most times this property is not going to be set. Not setting the property will avoid taking up javascript memory on server and also save bandwidth as we won't have to send `showOwner: false` in the JSON from the server when the property is not set.

src/models/PlayerInputModel.ts
```diff
@@ -31,6 +31,7 @@ export interface PlayerInputModel {
     payProduction?: IPayProductionModel;
     aresData?: IAresData;
     selectBlueCardAction: boolean;
+    showOwner: boolean;
```
bafolts, line 34: Lets make this `showOwner?: boolean;`. The property not existing will save space. As most `PlayerInputModel` wont have this property it will save 10 or so bytes on every request with the resulting generated JSON. When the value is needed we can set to `true`.

src/server/ServerModel.ts
```diff
@@ -278,6 +279,7 @@ function getWaitingFor(
     playerInputModel.maxCardsToSelect = selectCard.maxCardsToSelect;
     playerInputModel.minCardsToSelect = selectCard.minCardsToSelect;
     playerInputModel.selectBlueCardAction = selectCard.selectBlueCardAction;
+    playerInputModel.showOwner = selectCard.showOwner;
```
bafolts, line 282: With the property optional lets change to:

```
if (selectCard.showOwner) {
  playerInputModel.showOwner = true;
}
```

src/server/ServerModel.ts
```diff
@@ -242,6 +242,7 @@ function getWaitingFor(
     payProduction: undefined,
     aresData: undefined,
     selectBlueCardAction: false,
+    showOwner: false,
```
bafolts, line 245: With `showOwner` defined as optional we don't have to provide a value here.

src/deferredActions/RemoveResourcesFromCard.ts
vsrisuknimit: fixed. I was thinking of making an option constructor like in SendDelegateToArea later but this works too. Good to know about false vs undefined on the server. I will try to be careful next time.

src/models/PlayerInputModel.ts
vsrisuknimit: done

src/server/ServerModel.ts
vsrisuknimit: done

src/server/ServerModel.ts
vsrisuknimit: removed

bafolts (approved):

PR #2803, opened 2021-02-20
Handler #1 Add handler for processeing requests, and start with ApiGames.

kberg:


src/routes/Route.ts
```diff
@@ -39,4 +39,9 @@ export class Route {
     res.write('Not authorized');
     res.end();
   }
+  public writeJson(res: http.ServerResponse, json: any) {
+    res.setHeader('Content-Type', 'application/json');
+    res.write(JSON.stringify(json));
```
bafolts, line 44: Calling `res.write` and then `res.end` should be the same as calling `res.end(JSON.stringify(json))`.

https://nodejs.org/api/http.html#http_response_end_data_encoding_callback

While this is moving we should take the time to simplify this.

kberg: ack

kberg: Thinking about it, my advice is wait on that -- I can do that as a follow-up. Would that be OK?

kberg: Once things settle down it'll be much simpler to write tests for all these implementations.

src/routes/Handler.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+export abstract class Handler implements IHandler {
+  private validateServerId: boolean = false;
+  constructor(options?: Handler.Options) {
+    this.validateServerId = options?.validateServerId === true;
+  }
+
+  private isServerIdValid(req: http.IncomingMessage, ctx: IContext): boolean {
+    const queryParams = querystring.parse(req.url!.replace(/^.*\?/, ''));
```
bafolts, line 18: This can make use of `ctx.url.searchParams` instead of parsing the url again and assuming req.url exists.

```
const serverId = ctx.url.searchParams.get('serverId');
if (serverId === null || serverId !== ctx.serverId) {
  console.warn('No or invalid serverId given');
  return false;
}
```

src/routes/Handler.ts
```diff
@@ -0,0 +1,57 @@
...
+      this.put(req, res, ctx);
+      break;
+    case 'POST':
+      this.post(req, res, ctx);
+      break;
+    }
+  }
+
+  get(req: http.IncomingMessage, res: http.ServerResponse, ctx: IContext): void {
+    ctx.route.notFound(req, res);
```
bafolts, line 49: We probably returned 404 prior to this change and this is only moving the logic, but why we are here using (and adding) a `notImplemented` (501) probably makes more sense. While looking at heroku logs it would be easier to spot these 500 based status codes instead of 400 as well.

bafolts (changes requested):
Of all my comments the only one I think we should change prior to merging is using `ctx.url.searchParams`. Since we are already parsing the URL this should also be easier to read than the regular expression and `querystring.parse`. This also avoids having to use `req.url!` and assuming `req.url` will be defined.

src/routes/Route.ts
bafolts: Waiting for this is fine. This is only an optimization.

kberg:
> Of all my comments the only one I think we should change prior to merging is using `ctx.url.searchParams`. Since we are already parsing the URL this should also be easier to read than the regular expression and `querystring.parse`. This also avoids having to use `req.url!` and assuming `req.url` will be defined.

This also is one where I'd like to write tests to fix that, separately. Do you feel strongly about this?

bafolts (approved):
If you don't have the unit tests in the works I could make my requested changes while adding tests today.

kberg:
If you like.

>

PR #2806, opened 2021-02-20
Handler #4: Convert the rest of the API into to Handlers. Depends on Handler #3

kberg:


src/routes/ApiCloneableGames.ts
```diff
@@ -0,0 +1,21 @@
...
+  public static readonly INSTANCE = new ApiCloneableGames();
+  private constructor() {
+    super();
+  }
+
+  public get(req: http.IncomingMessage, res: http.ServerResponse, ctx: IContext): void {
+    Database.getInstance().getClonableGames(function(err, allGames) {
+      if (err) {
+        console.warn('Could not load cloneable games: ', err);
+        ctx.route.internalServerError(req, res, err);
```
bafolts, line 16: This needs to call `return` after this line to avoid calling to `internalServerError` as well as `writeJson`.

kberg: Good eye, fixed.

bafolts (approved):

PR #2807, opened 2021-02-20
Fix #2802, load game ids prior to returning game ids.

### LINKED ISSUE #2802: /api/games returns no games until a game action occurs
Reproduce by starting a local server. Before doing anything else, send a valid  request to the /api/games endpoint (with a valid serverId). It returns an empty array.

Once you do something like create a new game, or anything that causes the GameLoader to initialize, the call to /api/games returns the full set of games.

kberg:
Tested manually.

### DIFF
--- a/src/database/GameLoader.ts
+++ b/src/database/GameLoader.ts
@@ -61,6 +61,7 @@ export class GameLoader {
   }
 
   public getLoadedGameIds(): Array<string> {
+    this.loadAllGameIds();
     return Array.from(this.games.keys());
   }
 
@@ -185,6 +186,7 @@ export class GameLoader {
     this.loadNextGame();
   }
 
+  // This method is re-entrant.
   private loadAllGameIds(): void {
     if (this.state !== State.WAITING) {
       return;


src/database/GameLoader.ts
```diff
@@ -61,6 +61,7 @@ export class GameLoader {
   }
 
   public getLoadedGameIds(): Array<string> {
+    this.loadAllGameIds();
```
bafolts, line 64: Which database did you use where you saw this work? For asynchronous databases like postgres I don't think this will work. The call to `loadAllGameIds` would asynchronously start the query to find all the game ids while the `getLoadedGameIds` function returns an empty array. For this to work with asynchronous databases a callback argument would need to be provided which was called once all the game ids were loaded.

kberg: This is what the other bits of code do -- see `getGameById` and `getPlayerById`.

It's true I was using the filesystem cache, but I don't know what else to do here.

kberg: And those methods will be similarly broken?

bafolts: The `getByGameId` and `getByPlayerId` both take callback arguments. They aren't broken because the callbacks are not called until the database has loaded all the game ids. This `getLoadedGameIds()` would need to be updated to take a callback pattern and have similar logic to `getByGameId` where the callbacks are stored until we have the game ids loaded from the database.

kberg: Tsk, you're probably right.  I'll close this. The issue, minor as it is, remain.

PR #2809, opened 2021-02-20
check custom corp count

### LINKED ISSUE #2789: Game creation with less than (players * corpsPerPlayer) corporations should be an error
The game creation interface should know deterministically how many corporations will be in play, and if that's not enough to give everyone the desired number, seems like that should be an error instead of reverting to 2 corps per player.

image

vsrisuknimit:
This PR can close #2789 .

### DIFF
--- a/src/components/CreateGameForm.ts
+++ b/src/components/CreateGameForm.ts
@@ -409,6 +409,7 @@ export const CreateGameForm = Vue.component('create-game-form', {
       const requiresMoonTrackCompletion = component.requiresMoonTrackCompletion;
       let clonedGamedId: undefined | GameId = undefined;
 
+      // Check custom colony count
       if (customColoniesList.length > 0) {
         const playersCount = players.length;
         let neededColoniesCount = playersCount + 2;
@@ -420,6 +421,16 @@ export const CreateGameForm = Vue.component('create-game-form', {
 
         if (customColoniesList.length < neededColoniesCount) {
           window.alert('Must select at least ' + neededColoniesCount + ' colonies');
+          return;
+        }
+      }
+
+      // Check custom corp count
+      if (customCorporationsList.length > 0) {
+        const neededCorpsCount = players.length * startingCorporations;
+
+        if (customCorporationsList.length < neededCorpsCount) {
+          window.alert('Must select at least ' + neededCorpsCount + ' corporations');
           return;
         }
       }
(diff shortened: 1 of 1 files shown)


src/components/CreateGameForm.ts
```diff
@@ -424,6 +425,16 @@ export const CreateGameForm = Vue.component('create-game-form', {
         }
       }
 
+      // Check custom corp count
+      if (customCorporationsList.length > 0) {
+        const neededCorpsCount = players.length * startingCorporations;
+
+        if (customCorporationsList.length < neededCorpsCount) {
+          window.alert('Must select at least ' + neededCorpsCount + ' corporations');
```
alrusdi, line 433: Can you please add i18n?

alrusdi (changes requested):

src/components/CreateGameForm.ts
vsrisuknimit: added

src/components/CreateGameForm.ts
```diff
@@ -430,7 +431,7 @@ export const CreateGameForm = Vue.component('create-game-form', {
         const neededCorpsCount = players.length * startingCorporations;
 
         if (customCorporationsList.length < neededCorpsCount) {
-          window.alert('Must select at least ' + neededCorpsCount + ' corporations');
+          window.alert($t('Must select at least ') + neededCorpsCount + $t(' corporations'));
```
alrusdi, line 434: I'm sorry it is little bit incorrect. You should pass a Message object to this function with [neededCorpsCount] as data and something like "Must select at least ${0} colonies" as message

vsrisuknimit: @alrusdi I don't follow. Can you do it in another PR or point me to a line that does something similar? None of the alert message in this file use v-18n so far.

alrusdi: Can you try something like this:
```$t({type: LogMessageDataType.RAW_STRING, data: [neededCorpsCount] message: 'Must select at least ${0} colonies'} as Message)```
?

vsrisuknimit: image

vsrisuknimit: @alrusdi would you like to try it later in a different PR?

alrusdi (approved):
ok, let's do i18n later

PR #2810, opened 2021-02-20
Handler #5: Turning CreateGame and LoadGame into handlers

kberg:
Warning, though, I cannot tell how to test LoadGame. I think it's hard to access.

**Update** I tested with `curl -X PUT -d "{\"game_id\":\"19d33604256e\"}" localhost:8080/load`


src/components/App.ts
```diff
@@ -23,6 +23,7 @@ interface MainAppData {
             'load' |
             'player-home' |
             'start-screen' |
+            'load-game-form' |
```
bafolts, line 26: This change seems unrelated. Isn't `load` used for showing the `load-game-form` component.

https://github.com/bafolts/terraforming-mars/blob/main/assets/index.html#L18

Why was this additional string needed?

kberg: Hmm. Why did I add this? I'll remove it and see what the compiler thinks.

kberg: You know what it is? I tried using that enum in the code below, and it didn't quite work out. Like I mentioned elsewhere, what's going on with LoadGameForm? Delete?

bafolts (approved):

PR #2812, opened 2021-02-20
keep trying if error while waiting

### LINKED ISSUE #2808: Unexpected server response
Playing async games on the demo server at herokuapp.org, I quite often get these pop up errors that steal focus from the active tab. I'm leaving the game open in a background tab on Firefox on MacOS.

Screen Shot 2021-02-20 at 8 47 57 AM

bafolts:
If there is an error response code from the server while waiting for a player to take their turn the server is most likely restarting. This has the failure cases retry while logging the error client side. On the server the non 200 response codes will be logged. If the client failed to reach the server they will try the request again after the update period. This should be a better user experience for multi-player games. This will close #2808


src/components/WaitingFor.ts
```diff
@@ -86,7 +86,8 @@ export const WaitingFor = Vue.component('waiting-for', {
         const xhr = new XMLHttpRequest();
         xhr.open('GET', '/api/waitingfor' + window.location.search + '&prev-game-age=' + this.player.gameAge.toString());
         xhr.onerror = function() {
-          alert('Error getting waitingfor data');
+          console.warn('unable to reach server waiting for');
```
alrusdi, line 89: Shoudn't it continue only if 500 http status is received?

alrusdi (approved):

src/components/WaitingFor.ts
kberg: While the alert isn't pretty, this leaves someone with no understanding of what's going on. Someone might think the server is frozen or something.

If the app up some kind of butter bar I'd be more comfortable with this.

kberg (changes requested):

bafolts:
I am going to modify this to use the custom `alert` we use elsewhere for the `WaitingFor` component. I will also make the error message more verbose to explain the situation that the server is most likely re-starting. Using the custom `alert` wont be as intrusive as the native `alert`. Once the alert is clicked we can then start the waiting for loop back checking with the server. When the server is restarting this will prevent requests from coming in. If someone has a tab open in the background the custom alert wont be as intrusive as the native alert. When they get back to the page they will see there was some problem and clicking ok will start their game back. This will be simpler than introducing a new UI component.

bafolts:
@kberg I took a simpler approach with this. I think it will improve the situation for #2808. Moving away from the native `window.alert` will avoid potentially stealing focus for users that are running a game in the background or a separate tab. I also improved the error message to make it less vague. This also moves the alert component to a place we can use it more often.

image

src/components/WaitingFor.ts
```diff
@@ -116,7 +116,7 @@ export const WaitingFor = Vue.component('waiting-for', {
             }
             vueApp.waitForUpdate();
           } else {
-            alert('Unexpected server response');
+            root.showAlert(`Received unexpected response from server (${xhr.status}). Server is most likely restarting.`, () => vueApp.waitForUpdate());
```
kberg, line 119: "This is often due to the server restating." is a little less potentially misleading, fwiw.

kberg (approved):

PR #2813, opened 2021-02-21
Move content type detection to a central location.

kberg:


bafolts (changes requested):
We make 50-60 network requests on some pages for assets. On a page refresh the server will end up looking through `CONTENT_TYPES` array 50-60 times. I think we should update `CONTENT_TYPES` to be a `Map`. If you map `.map` to `text/javascript` then passing in everything from the last `.` character to the end of the string can be used to look up the content type. This will be much more efficient.

kberg:
> We make 50-60 network requests on some pages for assets. On a page refresh the server will end up looking through `CONTENT_TYPES` array 50-60 times. I think we should update `CONTENT_TYPES` to be a `Map`. If you map `.map` to `text/javascript` then passing in everything from the last `.` character to the end of the string can be used to look up the content type. This will be much more efficient.

You're probably right, but given the rest of the work done that sounds like premature optimization, you see what I mean? I mean, once you're doing a series of string operations just to get to this point in the call, is this really that much of a loss?

But ... OK. It's not as if this optimization makes things harder to read or anything. Will do.

bafolts:
@kberg we have been working together on this long enough I knew as I wrote my requested change that you would be able to talk me out of it :-)

You are right that it is early optimization. I can mark this as approved and if you feel like changing to map that is fine.

bafolts (approved):

kberg:
> @kberg we have been working together on this long enough I knew as I wrote my requested change that you would be able to talk me out of it :-)
> 
> You are right that it is early optimization. I can mark this as approved and if you feel like changing to map that is fine.

Nah. You conviced ME. The only reason it didn't arrive earlier is that I'm writing a good test.

bafolts (approved):

PR #2815, opened 2021-02-21
Player boards rework

ssimeonoff:
## Changes
- Player boards are reverted as before - without a header
- "played cards" is moved to the right side of the resources
- button "show" is much bigger now (easy to hit on mobile)
- "played cards" icon represents the 3 cards' types
- (optional) if desired, the 3 "rectangles" can get bigger and display the actual played cards per type as well, making them even more similar to the "played cards" toggle filters. 
- "played cards" toggle filters minor change (to make them look closer to the "played cards" icon)
- "hide tiles" and toggle buttons minor change to match the  buttons main theme (dark/light grey)
- minor change to the "arrow" of energy to heat
- (optional) if desired, tag "none" can be removed so the bottom and top rows widths will match :)


vsrisuknimit:
I thought more about it. Tag none is used only for one card, Community Services. It is possible to how much production a player would gain from this card directly on the card or at the moment it is selected to play. That way we can remove the none tag.

However, during the draft, you need to know other players none tag to consider whether to pass this card or not. This cannot be alleviated by the solution above. 

Unfortunately, I think the none tag is necessary even though it is used for exactly one card.

vsrisuknimit:
image

I think it is okay to leave this empty space for now. I would like to add remaining blue card action there in the future.

chosta (commented):
Code looks good but would like to have @alrusdi have a quick check too.

alrusdi:
It needs fix of merging conflict. Otherwise it is looking good to me

kevinb9n:
The "blue green red" icon suggests a different count than the one the app is currently using. 

Today, every card that is in my actual possession (setting aside the strange case of cards on Self-Replicating Robots) is counted once and only once: it appears either as part of my "cards in hand", my "cards in play", or my "cards in event pile" (round yellow icon).

Just speaking for myself, I feel this is already working well and doesn't need a change. And I've given reasons, but I might be wrong. So I invited you on two occasions to explain why you want to do this. I guess this is the third. If you don't want to have that discussion, would it be fair to just not make the change yet?

@bafolts

kevinb9n:
To clarify: I think there are at least two possible changes being proposed here, maybe

1. it sounds like event cards played will start being counted twice, as part of two different totals

2. it sounds like there is some desire to treat corporation cards and prelude cards as being special/different from project cards

2 is the one I meant (that I've asked twice to understand the reasoning for), 1 is new to me.

ssimeonoff:
I can argue that "played cards" count should exclude the preludes and corporation, since they are not project cards. 
However the "automated played cards" filter counts the preludes as automated cards. 
For the sake of consistency the "played cards" count on the player boards also count the played preludes. 
The corporation is not counted anywhere. 
The sum of the 3 numbers displayed on the 3 "played cards" filters now 
Untitled
matches the "played cards" displayed on the player board.

nwai90:
@ssimeonoff The corporation name is also cut off in the above screenshot

kevinb9n:
> I can argue that "played cards" count should exclude the preludes and corporation, since they are not project cards.

Okay, this is a start. But it isn't a real argument; it only leads to the next question: does whether a played card is a project card or not *matter* for any purpose?

It turns out that:

* There are no rules anywhere in the game, or any official expansion or variant or promo card, that "care" whether a played card is a project card or not. (There are a scant few that care whether a card has the "automated" or "active" label which only project cards happen to have, but they are milestones/etc. with their own tracker in the UI.)

And it also turns out that:

* New players to the game routinely assume that there *is* something special about project cards vs. non-project cards. They aren't sure whether non-project cards count for Robotic Workforce, Community Services, etc. They aren't sure if corporations are "played" or whether the order of Stormcraft vs. Saturn matters, etc. Once these players learn the fact that there is NO important difference between these cards – you just get them in a different way and play them at a different time, that's all – then their understanding of the game *improves*. They can answer the Robotic Workforce question with, "well of *course* it counts."

The change you want to make works actively against that understanding, and for what it's worth, I still have no idea what advantage you see in it.

chosta:
@ssimeonoff did you also check the quick player resource preview on the top of the page. I remember I had to manually adjust it last time.
Agree with @nwai90, the player and corp name need at least a overflow hidden property with `...`

alrusdi (approved):

ssimeonoff:
Updated and tweaked the css. 

Regarding "played cards" counting the corporation. @nwai90 always thinks that way. 

Currently it doesn't.  
We count preludes as "automated" cards. 
If "played cards" count the corporation, then the corporation should be count as "active" card. 
Thus: from 1+2=3 to become 2+2=4.
Untitled

I cannot do that( at least don't know how yet). Can someone help?

kevinb9n:
> Regarding "played cards" counting the corporation. @nwai90 always thinks that way.
> 
> Currently it doesn't.

I believe it does, ever since #2127

ssimeonoff:
currently in this pr, with the current pushed commits

kberg:
Chatted with @ssimeonoff and going to merge this for now, and address the Corporation issue shortly. This will probably confuse a couple of people, but it's short-lived, and moving code is a good thing.

PR #2823, opened 2021-02-22
fix discount from corporation not showing

### LINKED ISSUE #2818: Discounts from corporation effects aren't appearing on player board
image

#2771 messed it up.

vsrisuknimit:
This can close #2818 .

image


kberg (approved):
Dangit! I looked in ServerModel assuming it was that simple, but didn't find this line. Thank you for fixing the thing I broke.

PR #2826, opened 2021-02-23
Move computing the last generation of solo games to a value supplied by the server.

kberg:


src/Game.ts
```diff
@@ -521,13 +521,41 @@ export class Game implements ISerializable<SerializedGame> {
...
+    let lastGeneration = 14;
+    const options = this.gameOptions;
+    if (options.preludeExtension) {
+      lastGeneration -= 2;
+    }
+
+    // Only add 2 more generations when using the track completion option
+    // and not the solo TR option.
+    //
+    // isSoloModeWin backs thsi up.
```
bafolts, line 534: The `thsi` on this line should be `this`.

src/components/help/HelpPhases.ts
```diff
@@ -93,8 +93,12 @@ export const HelpPhases = Vue.component('help-phases', {
                 <li><span v-i18n>i. Game end check</span>
                   <ul>
                     <li v-i18n>If temperature, oxygen, and oceans are maxed, skip the rest of solar phase to the end game phase. </li>
-                    <li v-i18n>(If playing solo, the game is ended after 14 generations or 12 generations if Prelude expansion is used. </li>
-                    <li v-i18n>(If Venus Next is in use, the Venus track must be completed as well for solo mode victory.) </li>
+                    <li v-i18n>(If playing solo, the game ends after 14 generations.)</li>
+                    <li v-i18n>Solo mode adjustments for expansions:</li>
+                    <li v-i18n>• Prelude: The game ends 2 generations earlier.</li>
```
bafolts, line 98: Seems we do this in other places in `HelpPhases` but this should be using a `<ul>` element. The `list-style-type` css attribute can be used to determine the style.

https://www.w3schools.com/css/css_list.asp

We should use this styling instead of writing it out ourselves. This will make translating simpler as well.

bafolts (approved):

src/components/help/HelpPhases.ts
kberg: I spent about 15 minutes on that, and gave up. I used to be really good at CSS. Thanks.

src/Game.ts
kberg: done.

kberg (commented):
Thank you!

src/components/help/HelpPhases.ts
bafolts: Since I know of this problem I can take a look at it at some point. I hope I am still pretty good with CSS.

PR #2827, opened 2021-02-23
Write test for ApiHandler

kberg:
I wrote this in two steps. The first was to test the existing behavior, which requried the majority of changes, and is in the first commit. The second commit is purely the one that replaces regex parsing with the simpler code.


tests/routes/FakeGameLoader.ts
```diff
@@ -0,0 +1,21 @@
+import {IGameLoader} from '../../src/database/IGameLoader';
+import {Game} from '../../src/Game';
+
+export class FakeGameLoader implements IGameLoader {
```
bafolts, line 4: :+1: dependency injection ftw

bafolts (approved):

tests/routes/FakeGameLoader.ts
kberg: I hoped you would like it.

PR #2828, opened 2021-02-23
Add corporation contest winner Curiosity II

nwai90:
Screenshot 2021-02-23 at 4 55 56 PM


src/CardName.ts
```diff
@@ -467,6 +467,7 @@ export enum CardName {
     PROJECT_WORKSHOP = 'Project Workshop',
     INCITE = 'Incite',
     PLAYWRIGHTS = 'Playwrights',
+    CURIOSITY_II = 'Curiosity II',
```
kberg, line 470: I assume the card names are in a specific order, so why put this one here?

src/cards/community/CuriosityII.ts
```diff
@@ -0,0 +1,67 @@
...
+              eb.emptyTile('normal', CardRenderItemSize.SMALL).startEffect.megacredits(3).arrow().cards(1);
+            });
+          });
+        }),
+      },
+    });
+  }
+
+  public onTilePlaced(cardOwner: Player, activePlayer: Player, space: ISpace) {
+    if (cardOwner.id === activePlayer.id && space.bonus.length > 0 && space.spaceType !== SpaceType.COLONY) {
```
kberg, line 43: According to the text as written, this should not trigger when placing the three special oceans. Those spaces are not empty, they have an ocean tile.

The same applies to placing a tile on top of a hazard.

kberg, line 43: As written, you will need to know when the space is on The Moon and not Mars. The way I've been distinguishing between them is to check the tile type. Here you would filter out TileType.MOON_COLONY, TileType.MOON_MINE, and TileType.MOON_ROAD. Ideally, onTilePlaced would include a "satellite" parameter, or perhaps ISpace would include one (that just repeats that it's a Mars or Moon space) but this works since those are the only three types of tiles in use on the Moon.

Or, permit this corporation to apply to The Moon. That may not be what Curiosity II was meant for (Mars, I get it, really I do) but consider that it can put the corporation at a disadvantage.

tests/cards/community/CuriosityII.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+  it('Can pay 3 MC to draw card when placing a tile on a non-empty space', function() {
+    const nonEmptySpace = game.board.getAvailableSpacesOnLand(player).find((space) => space.bonus.length > 0)!;
+    game.addCityTile(player, nonEmptySpace.id);
+    player.cardsInHand = [];
+
+    expect(game.deferredActions.length).to.eq(1);
+    const orOptions = game.deferredActions.pop()!.execute() as OrOptions;
+
+    orOptions.options[1].cb(); // Do nothing
+    expect(player.cardsInHand).has.lengthOf(0);
```
kberg, line 31: I think you can write .is.empty, but your call.

src/CardName.ts
kberg: This is so obviously a minor point, but you did it consciously, and I'm curious.

src/cards/community/CuriosityII.ts
kberg: The first is easy to test for because you can filter on the tile type. But the hazards will not, because by the time this function is called, the hazard has disappeared.

kberg: Wait, I'm, just realizing: this is for when you place a tile on a non-empty space? _non-empty?_

kberg: Now I see -- non-empty isn't about tiles or land claim, it's about space bonuses? I'd suggest rewriting the description either way, and help me understand the proper goal. (If it's space bonus, then the overplacement for hazards and oceans probably still applies.)

src/cards/community/CuriosityII.ts
```diff
@@ -0,0 +1,67 @@
...
+
+      metadata: {
+        cardNumber: '',
+        description: 'You start with 40 MC and 2 steel production.',
+        renderData: CardRenderer.builder((b) => {
+          b.br.br;
+          b.megacredits(40).nbsp.production((pb) => pb.steel(2));
+          b.corpBox('effect', (ce) => {
+            ce.effect('When you place a tile on a non-empty space ON MARS, you may pay 3 MC to draw a card.', (eb) => {
+              eb.emptyTile('normal', CardRenderItemSize.SMALL).startEffect.megacredits(3).arrow().cards(1);
```
kberg, line 34: I think add astersik before startEffect, since there's a special condition not conveyed in the iconography.

src/cards/community/CuriosityII.ts
nwai90: The description can be changed if there is a better way to word it. This is the original which is not quite catch-all either, maybe @kevinb9n can help with that. The term "resource" is probably not sufficient here since a card is not a resource.

image

kberg: The rulebook calls it a "placement bonus"

image

"When covering a placement bonus with a tile."?

kberg: Which reminds me: what will this corp do during WGT?

src/CardName.ts
nwai90: No real reason, will move it up

src/cards/community/CuriosityII.ts
nwai90: Resolved

alrusdi (approved):
It seems working. I've tried it in solo mode and didn't found any issues

src/cards/base/RoboticWorkforce.ts
```diff
@@ -146,6 +146,7 @@ export class RoboticWorkforce extends Card implements IProjectCard {
     CardName.MINING_GUILD,
     CardName.RECYCLON,
     CardName.UTOPIA_INVEST,
+    CardName.CURIOSITY_II,
```
kberg, line 149: Why is this applied here? This card doesn't have behavior Robotic Workforce would use.

src/cards/community/CuriosityII.ts
```diff
@@ -0,0 +1,80 @@
...
+      startingMegaCredits: 40,
+
+      metadata: {
+        cardNumber: '',
+        description: 'You start with 40 MC and 2 steel production.',
+        renderData: CardRenderer.builder((b) => {
+          b.br.br;
+          b.megacredits(40).nbsp.production((pb) => pb.steel(2));
+          b.corpBox('effect', (ce) => {
+            ce.effect('When you place a tile on an area ON MARS that has a resource placement bonus, you may pay 3 MC to draw a card.', (eb) => {
```
kberg, line 37: @kevinb9n for feedback on this one. A "resource placement bonus" is a new term, and might be just fine.

src/cards/community/CuriosityII.ts
```diff
@@ -0,0 +1,80 @@
...
+    });
+  }
+
+  public onTilePlaced(cardOwner: Player, activePlayer: Player, space: ISpace) {
+    const eligibleBonuses = [SpaceBonus.STEEL, SpaceBonus.TITANIUM, SpaceBonus.HEAT, SpaceBonus.PLANT, SpaceBonus.MEGACREDITS, SpaceBonus.ANIMAL, SpaceBonus.MICROBE];
+    const moonTileTypes = [TileType.MOON_COLONY, TileType.MOON_MINE, TileType.MOON_ROAD];
+    const tile = space.tile as ITile;
+
+    if (cardOwner.id !== activePlayer.id) return;
+    if (cardOwner.game.phase !== Phase.ACTION) return;
```
kberg, line 52: This  makes prelude an invalid phase. Suggest using === Phase.SOLAR.

src/cards/community/CuriosityII.ts
```diff
@@ -0,0 +1,80 @@
...
+      },
+    });
+  }
+
+  public onTilePlaced(cardOwner: Player, activePlayer: Player, space: ISpace) {
+    const eligibleBonuses = [SpaceBonus.STEEL, SpaceBonus.TITANIUM, SpaceBonus.HEAT, SpaceBonus.PLANT, SpaceBonus.MEGACREDITS, SpaceBonus.ANIMAL, SpaceBonus.MICROBE];
+    const moonTileTypes = [TileType.MOON_COLONY, TileType.MOON_MINE, TileType.MOON_ROAD];
+    const tile = space.tile as ITile;
+
+    if (cardOwner.id !== activePlayer.id) return;
```
kberg, line 51: A prior comment was resolved, but isn't. Specifically, about Ares-compatibility. Overplacing oceans and hazards will not reveal the space bonuses and should be filtered.

Oceans won't be hard to manage, but hazards will be tricky.

kberg (changes requested):

src/cards/base/RoboticWorkforce.ts
nwai90: It does. Two steel production is copy-able

kberg: You're right of course. Sorry, I was so focused on the effect, I hadn't looked at the initial behavior.

src/cards/community/CuriosityII.ts
nwai90: @kberg Does it matter here though? The effect is triggered when a tile is placed on a space that has some resource(s) on it. On a physical board this would be the equivalent of the "printed" resources on the tile.

My interpretation is that it is not necessary for you to actually gain the bonus resources on the tile itself. For example in the case of hazards, you will never gain the bonus resources when placing a tile on top of a hazard to remove it. Preventing the corporation effect from triggering would significantly weaken this corporation in an Ares game.

kevinb9n: I have assumed the behavior should be just like Mining Area/Rights, only widened from 2 resource types to all. How do those currently work in this sense?

src/cards/community/CuriosityII.ts
kevinb9n: It was my suggestion, based on Mining Area/Rights.

Or maybe "a placement bonus including at least one resource" would be extra extra clear?

src/cards/community/CuriosityII.ts
kberg: I do think it matters here, because I think it's thematically inconsistent. A player places a tile over a hazard, and activates the corp ability without gaining the benefit, it's a bit awkward.

Consider if this card was used in a physical game. What would it do?

What if the description read "When you gain a _resource placement bonus_?" This would work, and would work slightly different from `EcologicalSurvey` but they would be pretty similar. I don't understand exactly how the new priority queue works, but @Lynesth promises me that it does the right thing. So perhaps you can mimic Ecological Survey to achieve the this goal.

I realize I'm getting into card design, but my priorities in this discussion are a) consistent behavior, b) mildly reasoanble theme c) clear card description.

kberg: Whether this PR contains such a change or not is less important than us agreeing on what the behavior ought to be (or settling on letting it ride for now.)

vsrisuknimit: Sid (the card designer) does not play with Ares. He thinks @nwai90's current wording and the current code behavior is fine.

kberg: > Sid (the card designer) does not play with Ares. He thinks @nwai90's current wording and the current code behavior is fine.

I am not surprised given the wording that was the case.

vsrisuknimit: It is fine. It is the behavior we want. It is more balanced this way. So let's move on with it. Or do you not like it just because it is not thematic?

kberg: I will wait for @nwai90 to address my questions.

vsrisuknimit: I think going forward we need to be clear whether Ares-compatibility is the responsibility of the person trying to add a new card or it is something to be handled later by the maintainer of Ares. Not everyone plays with Ares and should be expected to understand the intricacy of Ares. I am asking this because there are more cards from the competition that have tile placement action/effect.

I also think that we should talk to the designer if we want to change the design. I did ask Sid. 

> I think you could make it work either way. I don't know what would be best. Probably let it still get the card draw bonus, because otherwise it would be too weak in Ares?

kberg: It's the responsibility of both the author and the reviewer to make sure the change is as broadly compatible as can be, Ares or not Ares. This is a prime example of the sort of "discuss before making a PR" that's now in CONTRIBUTING.md. This is the fundamental problem with incomplete contributions; they put an immense burden on maintainers who also like doing cool things like designing cards.

Did this contest include a commitment to provide a PR to the main repo? If so, I hope the entrants realize that it might have to be different from what they envisioned, either because they didn't consider it, or because it's far too structurally difficult to get right. The community server is a fine place for code with a different definition of compatibility.

I'm going to propose a PR update to contributors.md to indicate this. It will help.

kberg: Honestly, I think once nwai90 and I get a chance to talk this won't take long.

nwai90: There is a difference between merely placing a tile on a space that contains a resource bonus and having to actually gain that bonus. Current code triggers the corporation's ability based on the latter.

In Ares, the hazard covers the bonus, so you don't get those resources if you place there and remove the hazard. It does not mean the resources are not there though. They are still present, you just don't gain them. Allowing this corporation to trigger its effect to pay 3 MC to draw a card hardly seems game-breaking in these scenarios, given that its ability is about average.

The issue is purely pedantic - I will suggest to leave it as is and not further nerf this corporation.

kberg: As you know, the upgraded oceans are also not eligible for the bonuses underneath. That's set a precedent for how tile overlay works. And now Ares will have inconsistent behavior with this for. And that's why I think this isn't pedantic? For the person not playing Ares? Of course. For the person playing Ares, it's inconsistent, confusing, and solvable.

kberg: In a discussion with @nwai90 we came to some agreements.

1. @nwai90 expressed some detail about his concern about nerfing the corp with Ares. While I don't think that's so bad, and also imagine if I used this corp on a physical board, the fact that it's covered would change a lot.
2. In early review, suggested he decide whether to include The Moon, and he chose to not include it. I didn't ask why, but I can imagine it was thematic for our beaitful Mars rover. But in further discussion, by omitting this corp from bonuses on The Moon (and bonuses are coming with the new map) that nerfs the corp (I suspect) even worse than the concern for Ares.

We hit upon a compromise.

1. Allow overplacement on Mars.
2. Enable the corp behavior on The Moon
3. Change the text to clarify overplacement is valid. Two options:
a. When you place a tile on a space with a resource placement bonus, or cover a tile on a space with a space bonus, you may ... .
but that's janky.
b.  When you place a tile on a space with a resource placement bonus you may blah blah end of sentence.) <newline> (Placing a tile on top of another one when the space has a resource placement bonus triggers this effect, such as with Ares.)

What's the saying about compromise? In a good compromise everyone walks away a little unhappy?

kberg: Wait! New idea. Something even simpler in terms of text.

"When you place a tile on a space with a RESOURCE placement bonus, or place a tile on top of another tile, you may..."

(I capitalized RESOURCE because I suspect it's going to matter that people unclear won't get it.)

But that's not janky, remains thematic, and gives this corp it's freedom from being nerfed.

To make it work still requires the same code change I'm going to have to make, but it'll be easy to do and I'll follow up with it in the next day or two.

All we need now is agreement that this is the way to go. (already got an OK from @nwai90)

vsrisuknimit: @kberg Cards that are not Ares-specific should not have wording on them that are Ares-specific. That will just confuse the general users who will ask 'how do I place a tile on top of another tile?'.

kevinb9n: @vsrisuknimit I think that's a good point. It might (?) still be the most acceptable solution anyway, but it's a good point.

kberg (approved):

src/cards/community/CuriosityII.ts
kberg: I'd prefer it to be cleaner, too. Like I said before, it seems to be an acceptable compromise.

vsrisuknimit: image

The first sentence totally contains the second sentence, making the second one redundant.

Perhaps consider make an Ares version of the corp if you need a very specific wording for its interaction with Ares.

vsrisuknimit: I also thought the card logic was changed to ignore what is underneath if you place on top of another tile.

Or make it totally clear this is for Ares. 

> When you place a tile on an area* that has a RESOURCE placement bonus, you may pay 3 mc to draw a card.
> (Ares fan-made expansion only: or when you place a tile on top of another tile)

bafolts:
Is this waiting on something to merge?

nwai90:
@bafolts It's ready for merge

PR #2832, opened 2021-02-24
Make it possible to track covered tiles for Curiosity II.

kberg:


kberg:
This just demonstraes how simple it would be to address the tile placement rule. Heh, I wish I thought of that back in September.

Still needs tests.

src/cards/community/CuriosityII.ts
```diff
@@ -33,8 +33,9 @@ export class CuriosityII extends Card implements CorporationCard {
           b.megacredits(40).nbsp.production((pb) => pb.steel(2));
           b.corpBox('effect', (ce) => {
             ce.effect(
-              'When you place a tile on an area that has a RESOURCE placement bonus*, you may pay 3 MC to draw a card. ' +
-              '(or place a tile on top of another tile with a RESOURCE placement bonus underneath)', (eb) => {
+              'When you place a tile on an area that has a RESOURCE placement bonus, ' +
+              'or place a tile on top of another tile, you may pay 3 MC to draw a card. ',
```
nwai90, line 37: `When you place a tile on an area that has a RESOURCE placement bonus or on top of another tile, ...` suffices here imo

nwai90 (approved):

src/cards/community/CuriosityII.ts
```diff
@@ -50,7 +51,7 @@ export class CuriosityII extends Card implements CorporationCard {
     if (cardOwner.game.phase === Phase.SOLAR) return;
     if (space.spaceType === SpaceType.COLONY) return;
 
-    if (space.bonus.some((bonus) => eligibleBonuses.includes(bonus))) {
+    if (space.bonus.some((bonus) => eligibleBonuses.includes(bonus)) || space?.tile?.covers !== undefined) {
```
nwai90, line 54: @kberg There is potentially an issue here spotted by @vsrisuknimit . This would trigger the corp ability even when placing on top of an ocean or hazard tile that has no actual resources on it.

kberg: Right. That was what I was proposing as a compromise.

src/cards/community/CuriosityII.ts
kberg: OK. Would you also be OK with that _either_ has a RESOURCE placement bonus or _is_ on top? otherwise "or" is iffy.

PR #2834, opened 2021-02-24
clarify pharma union card effect

### LINKED ISSUE #2829: The app's Pharmacy Union card text is misleading
The app is leaving out the "or as much as possible" part, which leads to rules confusion like this:
https://github.com/bafolts/terraforming-mars/issues/2191#issuecomment-765442112

The text from the kickstarter image is something like this:

"(Effect: When ANY microbe is played, including these, add a disease here and lose 4 M€ or as much as possible. When you play a science tag remove 1 disease from here and raise your TR 1 step, OR, if there are no diseases here, you may raise your TR 3 steps and TURN THIS CARD FACE DOWN (lose this card's tags and effects).)"

Since space is tight, we may need to abridge. This is approximately the shortest abridgement that I think can work (while also fixing a couple other issues):

"(Effect: When ANY microbe tag is played, including these, add a disease here and lose as much of 4 M€ as possible. When you play a science tag: if there is any disease here, remove one and gain 1 TR; if not, you MAY put this card face down in your EVENTS PILE to gain 3 TR.)"

EDIT: Ideas if that can't fit:

- drop "including these," because it's only a reminder of how things already work
- drop "here" x2 until some other card can hold disease resources
- change "lose as much of 4 M€ as possible" to "lose up to 4 M€", which is technically ambiguous, but at least has only one *plausible* interpretation (being actually optional would be super weird)

vsrisuknimit:
This PR closes #2829 .

Current:
image

New change:
image


kevinb9n:
Would you be willing to include the "events pile" part of my suggestion?

As I understand it, that rule change happened sometime after the kickstarter image went up, and I would charitably presume that the final print card will probably address it.

vsrisuknimit:
image

@kevinb9n

chosta (approved):

PR #2835, opened 2021-02-25
beginning of spectator mode

vsrisuknimit:
Here is a draft of spectator mode.

When a game is created, spectatorId is randomly generated along side gameId and playerIds.

What I need help with:

- I would like `.../spectator?id=....` to redirect index.html to display spectator-home, which currently is just a copy of player-home. I will work on that later. I don't understand App.ts at all and would like help with that.

Once this routing is done and I can display `.../spectator?id=....` properly, I will work on the front end for what the spectator should be able to see. Ideally we would generate two id for spectators: normal and super. Normal you can watch everything that is played. Super you can watch everyone hand as well. Neither will allow you to take an action.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -107,6 +107,7 @@ function processRequest(req: http.IncomingMessage, res: http.ServerResponse): vo
     case '/solo':
     case '/game':
     case '/player':
+    case '/spectator':
     case '/the-end':
     case '/load':
     case '/cards':
@@ -291,6 +292,8 @@ function createGame(req: http.IncomingMessage, res: http.ServerResponse): void {
     try {
       const gameReq = JSON.parse(body);
       const gameId = generateRandomId();
+      const spectatorId = generateRandomId();
+      console.log(`gameId: ${gameId} | spectatorId: ${spectatorId}`);
       const players = gameReq.players.map((obj: any) => {
         return new Player(
           obj.name,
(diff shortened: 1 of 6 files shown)


vsrisuknimit:
Tagging @bafolts and @Lynesth for help.

bafolts:
This strategy of generating a spectator id makes sense to me. Visually controlling what can be seen should be straight forward -- with one caveat. The server is going to be sending back data that someone with only the spectator id shouldn't have. This will also need a separate API route that takes all of the information and filters it server side so that someone with only spectator id doesn't get all of the response. I haven't taken a look at any of the files changed yet. Leaving my initial thoughts.

vsrisuknimit:
That makes sense where do you think the filtering should be done?

vsrisuknimit:
@bafolts I still need help with the backend on this PR. But no rush, whenever convenient is fine.

src/server.ts
```diff
@@ -291,6 +292,8 @@ function createGame(req: http.IncomingMessage, res: http.ServerResponse): void {
     try {
       const gameReq = JSON.parse(body);
       const gameId = generateRandomId();
+      const spectatorId = generateRandomId();
```
bafolts, line 295: The `updatePlayer` function of `src/components/App.ts` retrieves a `PlayerModel` through a request to `/api/player?id=PLAYERID`. As a spectator is associated with a game it may make sense to add a `/api/spectator?id=SPECTATORID`. This will have to return a new `SpectatorModel` (or better name) that includes only the information we would want a spectator to see.

This will also need to update the `GameLoader`. A `getBySpectatorId` will need to be added so that we can query for a game by spectator id server side.

vsrisuknimit: I can attempt to do what you outline, but I don't understand this part of the code at all and wonder if you will have time to make the change to this PR.

bafolts: Can you put the branch on `bafolts` origin so I can write to it? Or provide me write access to your origin?

vsrisuknimit: I am trying to figure how to move my branch to the main repo? But I gave you access to my branch on the mean time.

bafolts:
This work moved to #2921. Once the services work is in place can implement the client side in smaller pull requests.

PR #2837, opened 2021-02-25
Move Agendas Scientists P4 code to correct place

nwai90:
From Discord:

`About fan expansion Agenda: When the Scientists rule and their Ruling Policy is 'Cards with :tag_science: requirements may be played with 1 :tag_science: less', 1 additional :tag_science: is here (which is a good thing) but Orbital Cleanup should not take this tag into account I think :smiley:`


kberg:
Nice.

kberg (approved):

PR #2849, opened 2021-02-26
Moon card 65: Syndicate Pirate Raids

kberg:
This card's behavior is to prevent players from retrieving their trade fleets this generation. It uses a behavior I am loathe to do, which is add a special field on the game instance.

My knowledge of Colonies is in reading only, and not playing.

No tests yet, I'm more interested in a verification that this makes sense, structurally. And if you know of any other project cards that retrieve trade fleets mid-game, tell me? Though a semantic scan for `.visitor` suggests otherwise.

**Edited: Tests are of course now here.**


src/Player.ts
```diff
@@ -936,7 +936,13 @@ export class Player implements ISerializable<SerializedPlayer> {
   public runProductionPhase(): void {
     this.actionsThisGeneration.clear();
     this.removingPlayers = [];
-    this.tradesThisTurn = 0;
+    // Syndicate Pirate Raids hook. If it is in effect, then only the syndicate pirate raider will
+    // retrieve their fleets.
+    // See Colony.ts for the other half of this effect, and Game.ts which disables it.
+    if (this.game.syndicatePirateRaider !== undefined && this.game.syndicatePirateRaider !== this.id) {
```
nwai90, line 942: I would rewrite this for legibility as

```if (this.game.syndicatePirateRaider === undefined) {
    this.trades = 0;
} else if (this.game.syndicatePirateRaider === this.id) {
    this.trades = 0;
}```

**Comment by nwai90** [comment]
@kberg LGTM, it works as intended

File: src/Player.ts (hunk 1 above)
Comment (kberg) [reply src/Player.ts:942]: In fact, my code is broken (as I'm discovering with tests) and yours works. So that's a good reason. :D

File: src/turmoil/Turmoil.ts [hunk 2]
```diff
@@ -454,21 +452,6 @@ export class Turmoil implements ISerializable<SerializedTurmoil> {
 
       turmoil.playersInfluenceBonus = new Map<string, number>(d.playersInfluenceBonus);
 
-      function globalEventName(object: any): string {
```
nwai90, line 457: Note: This PR seems to include changes from a previous PR as well

nwai90 (approved):

src/turmoil/Turmoil.ts
kberg: Actually, it was just a matter of merging with head. Sigh.

PR #2850, opened 2021-02-26
fix seeded solo colony

### LINKED ISSUE #2362: Replaying solo game from seed lets you keep all four colony tiles in play (a major advantage)
Started a game on main server from seed 920566a07a55. After choosing cards there was no prompt to choose a colony tile to discard. When I choose to Trade, it seems one colony is missing, which is surely [EDIT: no] the one that the player who originated this seed had chosen to discard. But it still appears at the bottom of the page.

vsrisuknimit:
This PR closes #2362 .

Problem: RemoveColonyFromGame was previously called from newInstance. This means it is not called when the game is cloned via cloner.

This problem did not exist before January. It was caused up by #2223 . New cloner was made without the remove colony check.

Solution: Move that part to gotoInitialResearchPhase. Remove the colony first thing after someone pick their corporation.


vsrisuknimit:
I tested a bit. But please help test more locally since this could impact a lot of players.

nwai90 (approved):
Tested locally, works great 👍

PR #2853, opened 2021-02-26
avoid collision for /game route temporarily

bafolts:
We are starting to migrate all of the code in `src/server.ts` into separate routes. This makes them easier to unit test. While moving the `/game` route seems we have overlooked a collision where this route name is hit with both PUT and GET. This avoids the collision temporarily.


bafolts:
I will do an emergency deployment of this once merged.

kberg (approved):

kberg:
Ah, I see the problem. Thanks.

PR #2856, opened 2021-02-26
Added confirmation dialog when no inital cards selected

### LINKED ISSUE #2831: Show an alert if a player didn't pick initial cards by mistake
From time to time, one player accidentally makes an error and just picks a corp and preludes, but skips picking any cards.

When this happens, can we add a prompt indicating it, which allows that player to cancel the submission and go back to pick cards?

AssafKr:
This closes #2831
Shows a confirmation dialog, same component as the select-space dialog, that prompts when a player did not buy any cards at the initial selection phase.
image
Can be disabled in preferences

### DIFF
--- a/src/components/common/ConfirmDialog.ts
+++ b/src/components/common/ConfirmDialog.ts
@@ -8,6 +8,10 @@ export const ConfirmDialog = Vue.component('confirm-dialog', {
     message: {
       type: String,
     },
+    enableDontShowAgainCheckbox: {
+      type: Boolean,
+      default: false,
+    },
   },
   data: function() {
     return {
@@ -32,7 +36,7 @@ export const ConfirmDialog = Vue.component('confirm-dialog', {
     },
   },
   mounted: function() {
-    dialogPolyfill.default.registerDialog(this.$refs['dialog']);
+    (dialogPolyfill.default || dialogPolyfill).registerDialog(this.$refs['dialog']);
   },
   template: `<dialog ref="dialog">
       <form method="dialog">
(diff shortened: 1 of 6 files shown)


vsrisuknimit:
@AssafKr I think, because such case is so incredibly rare, it is okay not to have the memory of the settings in PreferenceMangager. Any time someone decide to buy ZERO card, it is okay to ask them.

AssafKr:
@vsrisuknimit alright, I'll remove the settings

vsrisuknimit:
@AssafKr up to you, what do you think? We can also wait to hear from other people as well.

AssafKr:
@vsrisuknimit I agree. I only kept it since it to consistent with the select space dialog, but it makes sense since it's rarer

src/components/common/ConfirmDialog.ts
```diff
@@ -32,7 +36,7 @@ export const ConfirmDialog = Vue.component('confirm-dialog', {
     },
   },
   mounted: function() {
-    dialogPolyfill.default.registerDialog(this.$refs['dialog']);
+    (dialogPolyfill.default || dialogPolyfill).registerDialog(this.$refs['dialog']);
```
bafolts, line 39: Was this change of adding `|| dialogPolyfill` needed?

bafolts (approved):

src/components/common/ConfirmDialog.ts
AssafKr: Without it the tests break. I don't know exactly why, when when importing `dialogPolyfill` while running the tests, the object recieved does no contain the `default` attribute. The `default` object is spreded one level higher inside the `dialogPolyFill` object.
I can dig more into it to understand why if we wanna get rid of this hack

kevinb9n:
Just a note that it's unclear to me why certain alerts are handled this way while others are handled the "Show in-game alerts" way (like when you spend a steel for less than full value).

src/components/common/ConfirmDialog.ts
bafolts: This is due to how node requires a module versus browser. Cool :+1:

bafolts:
> Just a note that it's unclear to me why certain alerts are handled this way while others are handled the "Show in-game alerts" way (like when you spend a steel for less than full value)

I don't think there is any reason why this is not using the `show_alerts` preference. Part of the reason is an oversight on my part. I didn't realize that the tile confirmation was adding a separate preference only for the tile. Moving every confirmation under `show_alerts` would make sense. 

@AssafKr can you only show this confirmation if `show_alerts` is set to on? It does seem we put other native `confirm` under this. I can update all the other native `confirm` to use the new `ConfirmDialog`.

AssafKr:
@bafolts I conditioned the confirmation by the `show_alerts` perference

PR #2864, opened 2021-02-27
Replace the Moon board. This is a breaking change.

kberg:
image

Probably needs some tests around new space bonuses, and manual verification. Currently uploading to http://terraforming-mars-kberg.herokuapp.com


src/moon/MoonBoard.ts
```diff
@@ -49,43 +53,67 @@ export class MoonBoard extends Board {
...
+    this.x = startX;
+    return new Row(this);
+  }
+  public colony() {
+    this.spaces.push(Space.colony(this.nextId()));
+  }
+  public nextId(): string {
+    this.idx++;
+    let strId = this.idx.toString();
+    if (this.idx < 10) {
```
bafolts, line 97: There is `String.prototype.padStart` available. I would use `strId = strId.padStart(2, '0')`

bafolts (approved):

src/moon/MoonBoard.ts
kberg: The other board code uses this same technique, ack, that's good. Will do. Thanks.

PR #2869, opened 2021-02-28
Lots of good changes to Moon cards, mostly corps.

kberg:
The mechanism for identifying size could very much be simplified, I think.

## Before

image

## After

image


kberg:
This is running on terraforming-mars-kberg.herokuapp.com

kberg:
This is running on http://terraforming-mars-kberg.herokuapp.com/cards

src/components/card/CardCorporationLogo.ts
```diff
@@ -10,142 +10,157 @@ export const CardCorporationLogo = Vue.component('CardCorporationLogo', {
...
+      case CardName.APHRODITE:
+        return '<div class="card-aphrodite-logo">APHRODITE</div>';
+      case CardName.ARKLIGHT:
+        return '<div class="card-arklight-logo">ARKLIGHT</div>';
+      case CardName.POSEIDON:
+        return '<div class="card-poseidon-logo">POSEIDON</div>';
+      case CardName.SATURN_SYSTEMS:
+        return `<div class="card-saturn-logo">
+        'SATURN <span style="font-size:20px;display:inline-block;">&#x25CF;</span> SYSTEMS';
+        '</span>`;
```
bafolts, line 24: You didn't introduce this but this should be `</div>` and not `</span>`.

bafolts (approved):

src/components/card/CardCorporationLogo.ts
kberg: Good catch, fixing.

Issue #2839, opened 2021-02-25
Minor: English card name mismatches with the print versions


kevinb9n:
Sorry about this folks. Since "CEOS" was actually fixed to "CEO'S" without incident I figured I may as well just report these too. Do what thou wilt...

My English first-printings have these names (which I doubt have been changed since):

* "Allied Bank", singular
* "Refugee Camps", plural
* "Inventors' Guild", with startlingly correct grammar
* "Cryo-Sleep", hyphenated
* "Designed Microorganisms", not hyphenated
* To my dismay, these corp names also use camel-case: ThorGate, TerraLabs, AstroDrill. But I haven't seen anything showing Ecoline with a capital L in the middle.

Lastly the 17-cost standard project is "Build Colony", but if the change to "Colony" was intentional for space reasons, I get it. (I still think it should appear with its full name in the log.)

kberg:
![PXL_20220930_153424076 MP](https://user-images.githubusercontent.com/413481/193306135-259a1273-5dca-41ba-86e7-47c6027cdbf1.jpg)

kberg:
Also confirmed ThorGate and TerraLabs, but Astrodrill must be in the other room.

kevinb9n:
Sorry I didn't think clearly enough to include a photo myself the first time; the time it took you to pluck those out probably already exceeds the time I would have wanted this "issue" to waste.

kberg:
It was easier than you might think. And it's easy enough to correct. I just
wanted to confirm.

Borbarad2:
Could this be closed? Is everything done here?

kberg:
It is not fixed yet.

kevinb9n:
Verified: cards were fixed, no change yet to corp names or standard project names.

As they say, "literally unplayable" :-)

PR #2870, opened 2021-02-28
rename tradeThisTurn

vsrisuknimit:
The only thing this PR does is renaming player.tradesThisTurn to player.tradesThisGeneration. The variable was badly named since it counts how many times the trade happened this Generation, not Turn as it said.

It does not do much now. But it will be helpful for rebalancing in the future if we actually want to limit the number of trade per Turn, not per Gen.

Tested. Everything works fine. I renamed all instances that I found by searching.


vsrisuknimit:
@nwai90 pointed out this break existing games due to serialization shenanigans, so let's close this for now. But we should consider renaming it when we are merging other things such as renaming card names #2839 that will also break existing games.

kberg:
The card name renames shouldn't be breaking existing changes given how
Brian recently did it. Same here, this is easy to fix without breaking
existing games. I'll kick that off.

kberg (commented):
Also you can revive this PR and do everything you did except rename the SerializedPlayer, and leave that as as TODO.

vsrisuknimit:
@kberg I reopened. 

`did except rename the SerializedPlayer, and leave that as as TODO`

Don't think I understand this part.

kberg:
> @kberg I reopened.
> 
> `did except rename the SerializedPlayer, and leave that as as TODO`
> 
> Don't think I understand this part.

The short version is to roll back the change to SerializedPlayer. That means that the serialize and deserialize functions will look a little weird: in serialize, you'll assign tradesThisGeneration to tradesThisTurn, and in deserialize you'll assign tradesThisTurn to tradesThisGeneration. And that'll look peculiar and doesn't fully do what you wanted, but the next part, which slowly changes what the field is stored as in the database, I'll do for you.

Does that explain it? If not, just say OK, and I'll make a commit that demonstrates it.

vsrisuknimit:
@kberg please take a look. The test failed due to deserialization. I am not sure how to fix that one.

kberg:
Will do. I hate that test that's breaking right now, but it's a good one.

kberg (commented):
I removed the test, and added another brief expectation in the test below.

That particular test is good but brittle, and was written before we had more carefully rewritten how serialization goes. It doesn't help when trying to rename things, so it's fine to remove it.

kberg (approved):

PR #2875, opened 2021-02-28
Log that the Mars University Effect is being used.

kberg:
Otherwise this is what you get, two card drawn in a row, it looks like a bug.

image

### DIFF
--- a/src/cards/base/MarsUniversity.ts
+++ b/src/cards/base/MarsUniversity.ts
@@ -44,6 +44,7 @@ export class MarsUniversity extends Card implements IProjectCard {
             new SelectCard('Select a card to discard', 'Discard', player.cardsInHand, (foundCards: Array<IProjectCard>) => {
               player.cardsInHand.splice(player.cardsInHand.indexOf(foundCards[0]), 1);
               player.game.dealer.discard(foundCards[0]);
+              player.game.log('${0} is using their ${1} effect to draw a card.', (b) => b.player(player).card(this));
               player.drawCard();
               return undefined;
             }),


src/cards/base/MarsUniversity.ts
```diff
@@ -44,6 +44,7 @@ export class MarsUniversity extends Card implements IProjectCard {
             new SelectCard('Select a card to discard', 'Discard', player.cardsInHand, (foundCards: Array<IProjectCard>) => {
               player.cardsInHand.splice(player.cardsInHand.indexOf(foundCards[0]), 1);
               player.game.dealer.discard(foundCards[0]);
+              player.game.log('${0} is using their ${1} effect to draw a card.', (b) => b.player(player).card(this));
```
nwai90, line 47: Maybe  '${0} drew a card via ${1} effect .' or '${0} drew a card using ${1} effect .' ?

kberg: If I included that, then player.drawCard would have to not log, otherwise that introduces the same problem -- the same screenshot I provided in the bug report would show three card drawn instead of two.

kberg: If you're suggesting that I do that, then, OK, I can see adding a DoNotLog parameter to drawCards.

nwai90: This isn't really about an extra parameter, it's about a more concise log description. `drawCards` already has a parameter which controls whether the names of the drawn cards are publicly revealed or not

kberg: I do get it. What I'm trying to say though, is that the wording you suggest will actually not achieve what I'm trying to fix. Unless I remove one of the extraneous "Tatiana drew 1 card(s)" messages and replace it with what you suggested.

The propsal in the PR isn't saying that somene drew a card. It's saying that The Next Card Draw is due to Mars University.

kberg: Or, drawCards can include a "source" option which will log additional metadata. In that case, drawCards would log the message you propose, not MarsUniversity.

nwai90: I see. Maybe it is worth adding an extra and optional `source` parameter in that case. If provided, append `" via ${source}"` to the log message

bafolts (approved):

kberg:
Ah, this PR was on hold in order to address what nwai90 and I agreed on.

>

bafolts:
Ahhhhhhhhhhh. Sorry about that. I try to keep the pull request list low and I didn't see a requested changes.

kberg:
Nah, no worries.

>

Issue #2598, opened 2021-01-29
RFC: Big problems we can solve with logging


kberg:
After getting some time to read #1233 -- I really like where everything in this project is going, including that design.

There were a few places that I wanted to start taking the project, particularly as it pertains to logging, and all of it is related to the two underlying changes I started last month, introducing `Units` and removing `Game` where it is not necessary. It includes  @kevinb9n's notion of having a complete log. Although not a complete log in the technical sense, more in the experiential sense.

This is a lot of stuff, but I think it touches on all the pieces. Sorry about how long it is.

So now that 'game' has been removed from all the cards, the next step is to hide all the interesting player state behind methods. That means a couple of things:

1. Replacing the publicly accessible megaCredits, et cetera with property equivalents, and underneath, represent all of them by a `Units` instance.
2. Remove `game` from `player.setResource` and `player.setProduction`. `game` is only passed to indicate that the unit/production change is meant to be logged. So, ALWAYS log.
3. Units changes in 1 are also logged.

That can make a noisy log, I'll get to the log later.

3. Update the `player.setResource` and `player.setProduction` methods with `Units`-compatlble versions. This makes the code which does all those variable-resource and variable-production type changes much simpler to read, and easier to cluster into a single update (and a single log message -- eg `Robert reduced plant production 1 step and raised megacredit production 1 step.`
3a. @sienmich expressed interest in this refactoring. IMO now's a good time for it.
4. Define two more game log levels. Right now there are two: intergenerational, and everything else. The four levels might be "generation", "turn/action", "important action details", "minute action details".  Like what @Lynesth is doing with action priorities, this will take some discussion.
5. To deal with all this logging information, make these changes to the UI:
5a. Make the log bigger, a more vital part of the UI.
5b. Make it possible to collapse the 'minute action details' level like an IDE collapses blocks of code (maybe just for other players)
5c. Add a filter box that finds entries in the log

More logging information is also a scaling issue, so accordingly, I would make the following changes:
6. Server-side, put logging in its own table. It will reduce overall storage meaningfully, and make it easier to save the game much more frequently. Each log entry has an increasing number.
7. A client, when sending a request to the server, sends a number representing the last log entry it knows about. Server responses will only contain log entries newer than that log entry. This will reduce network bandwidth.
8. NOW, whatever the Undo operation does, the historical log stuff doesn't change. In other words, add a log entry "Player did an undo."
[...]

PR #2876, opened 2021-03-01
pull logs one generation at a time

bafolts:
This adds back in the ability to see all logs. On the first generation the logs before the first generation are shown. On the final generation the last 50 are shown. Also added tests! Had to expand the api test handler to support loading by player id.

### DIFF
--- a/src/components/LogPanel.ts
+++ b/src/components/LogPanel.ts
@@ -44,11 +44,6 @@ export const LogPanel = Vue.component('log-panel', {
   components: {
     Card,
   },
-  watch: {
-    selectedGeneration: function(generation: number) {
-      this.getLogsForGeneration(generation);
-    },
-  },
   methods: {
     scrollToEnd: function() {
       const scrollablePanel = document.getElementById('logpanel-scrollable');
@@ -185,9 +180,13 @@ export const LogPanel = Vue.component('log-panel', {
       return '<i class=\'icon icon-cross\' />';
     },
     selectGeneration: function(gen: number): void {
+      if (gen !== this.selectedGeneration) {
+        this.getLogsForGeneration(gen);
+      }
       this.selectedGeneration = gen;
     },
     getLogsForGeneration: function(generation: number): void {
+      const messages = this.messages;
       // abort any pending requests
       if (logRequest !== undefined) {
         logRequest.abort();
@@ -202,8 +201,8 @@ export const LogPanel = Vue.component('log-panel', {
       };
       xhr.onload = () => {
         if (xhr.status === 200) {
-          this.messages.splice(0, this.messages.length);
-          this.messages.push(...xhr.response);
+          messages.splice(0, messages.length);
+          messages.push(...xhr.response);
           if (generation === this.generation) {
             this.$nextTick(this.scrollToEnd);
           }
@@ -241,7 +240,6 @@ export const LogPanel = Vue.component('log-panel', {
           retText = '<span class=\'last-generation blink-animation\'>' + retText + '</span>';
         }
       }
-
       return retText;
     },
   },
--- a/src/components/LogPanel.ts
+++ b/src/components/LogPanel.ts
@@ -14,6 +14,8 @@ import {TileType} from '../TileType';
 import {playerColorClass} from '../utils/utils';
 import {Color} from '../Color';
 
+let logRequest: XMLHttpRequest | undefined;
+
 export const LogPanel = Vue.component('log-panel', {
   props: {
     id: {
@@ -41,6 +43,11 @@ export const LogPanel = Vue.component('log-panel', {
   },
   components: {
     Card,
+  },
+  watch: {
+    selectedGeneration: function(generation: number) {
+      this.getLogsForGeneration(generation);
+    },
   },
   methods: {
     scrollToEnd: function() {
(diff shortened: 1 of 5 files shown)


src/components/LogPanel.ts
```diff
@@ -42,6 +44,11 @@ export const LogPanel = Vue.component('log-panel', {
   components: {
     Card,
   },
+  watch: {
+    selectedGeneration: function(generation: number) {
+      this.getLogsForGeneration(generation);
```
bafolts, line 49: The type checking system with `compile-vue-templates` isn't working with this for some reason. This is why build is failing. Will have to address before I can merge.

tests/routes/FakeGameLoader.ts
```diff
@@ -12,8 +12,25 @@ export class FakeGameLoader implements IGameLoader {
   getByGameId(gameId: string, _bypassCache: boolean, cb: (game: Game | undefined) => void): void {
     cb(this.games.get(gameId));
   }
-  getByPlayerId(_playerId: string, _cb: (game: Game | undefined) => void): void {
-    throw new Error('Method not implemented.');
+  getByPlayerId(playerId: string, cb: (game: Game | undefined) => void): void {
+    let found = false;
+    this.games.forEach((game) => {
```
kberg, line 17: this.games.getPlayers().some()

src/routes/GameLogs.ts
```diff
@@ -1,31 +1,53 @@
...
+      // for most recent generation pull last 50 log messages
+      if (generation === null || Number(generation) === game.generation) {
+        logs = game.gameLog.slice(-50);
+      } else { // pull all logs for generation
+        logs = this.getLogsForGeneration(game.gameLog, Number(generation));
+      }
 
       res.setHeader('Content-Type', 'application/json');
-      res.end(JSON.stringify(game.gameLog.slice(-50)));
+      res.end(JSON.stringify(logs));
```
kberg, line 50: This is uncompressed? I thought you were looking to compress logs.

src/routes/GameLogs.ts
```diff
@@ -1,31 +1,53 @@
...
-
+    const generation = ctx.url.searchParams.get('generation');
     if (id === null) {
-      ctx.route.badRequest(req, res, 'invalid playerid');
+      ctx.route.badRequest(req, res, 'must provide playerid');
       return;
     }
 
-    GameLoader.getInstance().getByPlayerId(id, (game) => {
+    ctx.gameLoader.getByPlayerId(id, (game) => {
```
kberg, line 35: <3

src/routes/GameLogs.ts
```diff
@@ -1,31 +1,53 @@
...
       if (game === undefined) {
         ctx.route.notFound(req, res, 'game not found');
         return;
       }
+      let logs: Array<LogMessage> | undefined;
+
+      // for most recent generation pull last 50 log messages
+      if (generation === null || Number(generation) === game.generation) {
+        logs = game.gameLog.slice(-50);
```
kberg, line 44: So does this mean if I'm on, say, generation 3, my UI will have all logs for generations 1 and 2, and only the last 50 for generation 3? So the log view will have a gap representing the first part of the third generation.

Of course generation 3 isn';t going to have 50 operations, but later generations will.

tests/routes/FakeGameLoader.ts
kberg: I'm sure this can be simplified.

kberg (changes requested):

src/routes/GameLogs.ts
bafolts: That is what it would mean. As this is confusing I can update this to return only the logs for that generation. The first generation would also return any log messages from before the first generation was logged.

src/routes/GameLogs.ts
bafolts: When it was all the logs I was looking to compress as those results would be large and always growing. With breaking down by generation the size of what we return should be smaller. Both the `gzip` and `br` impacted the CPU and response times significantly. Sometime on Sunday the response times with `br` grew so large that the server unusable, CPU load was over 4 and response time was averaging 10 seconds.

I would like to consider optimizations once the generation change is in. As heroku doesn't log bandwidth usage I am assuming at least breaking up by generation will lead to smaller payloads then sending back every game log. Ideally there would be some algorithm to send only the logs you needed. This would be straight forward without undo. Without undo the list is always growing and determining which logs a client needs it simple. With undo there may also be a simple strategy for determining which logs a client needs to download. This should be more effective than using compression and put less CPU load on the server.

bafolts: So at one point I watched a lot of the online MIT CS courses for advanced data structures. One of these was on retroactive data structures. I think we will need some way on the client to know which timeline you have. As long as no undo operation is used your timeline doesn't change. As long as the timeline doesn't change and we know how many logs you have on the client for that generation and timeline we can send the logs you need. If the timeline has changed you would need all logs. We have `game.gameAge` but I think we may also need a `game.timeline`. When the undo operation is used we could increment `game.timeline` as now we are entering a different timeline. A client may still have the old timeline. With this in place we would then update the log route to also provide `timeline` as an argument. This would all be done in a separate PR but I am curious what you think about this timeline approach @kberg

kberg: I covered this a little bit in #2598 LMK your thoughts about that.

src/routes/GameLogs.ts
```diff
@@ -1,31 +1,53 @@
...
 export class GameLogs {
-  public handle(req: http.IncomingMessage, res: http.ServerResponse, ctx: IContext): void {
-    if (req.url === undefined) {
-      ctx.route.notFound(req, res, 'url not defined');
-      return;
+  private getLogsForGeneration(messages: Array<LogMessage>, generation: number) {
+    let foundStart = generation === 1;
+    const newMessages: Array<LogMessage> = [];
+    for (const message of messages) {
+      if (message.message === 'Generation ${0}') {
```
kberg, line 12: `if (message.type === MessageType.NEW_GENERATION)` ?  Possibly quicker depending on how often this is called. Or, premature optimization.

src/routes/GameLogs.ts
kberg: Your call.

kberg (approved):

src/routes/GameLogs.ts
bafolts: I didn't realize that `MessageType.NEW_GENERATION` was a thing. This is good change as it doesn't rely on comparing the string that someone could update.

kberg:
The regex is still necessary, but the searches are faster. Or you can swap
out the regex with counting the number of NEW_GENERATION log messages
received.

>

bafolts:
Counting the number of NEW_GENERATION messages would work. We had the bug where we were only storing the last 50 messages. On some servers we may not have the entire log available. Using the check for the value is slower but should prevent against these backwards compatible situations where we may not have all the old logs.

PR #2884, opened 2021-03-04
unit tests for api/waitingfor

bafolts:
Adds unit tests for the waiting for route. Also took the opportunity to use the search params instead of parsing the query string. This route gets hit quite often so worth being efficient.


alrusdi (approved):

kberg:
Thanks!

Issue #2738, opened 2021-02-13
FR: show preludes while initial players are choosing preludes


kberg:
![image](https://user-images.githubusercontent.com/413481/107857689-ea4cfe80-6dfd-11eb-9ac6-25d886f94cf0.png)

I'd like to see my preludes during this phase of the game.

kevinb9n:
+1 they should be in your hand

vincentneko:
Fixed

PR #2886, opened 2021-03-04
Display Preludes in cards in hand

vincentneko:
#2738

Only local "Cards In Hand" number is updated to avoid a display discrepancy, without messing with planner milestone.

show_preludes_in_hand


kberg (approved):
Thank you!

PR #2888, opened 2021-03-05
Move ServeAsset and ServeApp to handlers

kberg:
NOTE: `ServeAsset` isn't really a handler yet. I'm just refactoring to I can start to make some sense out of the intertwined components separately.


src/server.ts
```diff
@@ -26,42 +23,28 @@ import {Route} from './routes/Route';
...
-    // ['/debug-ui', ServeApp.INSTANCE],
-    // ['/help-iconology', ServeApp.INSTANCE],
-    // ['/styles.css', ServeAsset.INSTANCE],
-    // ['/favicon.ico', ServeAsset.INSTANCE],
-    // ['/main.js', ServeAsset.INSTANCE],
-    // ['/main.js.map', ServeAsset.INSTANCE],
+    ['/', ServeApp.INSTANCE],
+    ['/new-game', ServeApp.INSTANCE],
+    ['/solo', ServeApp.INSTANCE],
+    ['/game', ServeApp.INSTANCE],
```
bafolts, line 38: There is already an entry for `/game`

https://github.com/bafolts/terraforming-mars/blob/main/src/server.ts#L72

Will need to merge with `CreateGame` in some way.

bafolts (changes requested):
Will need to consider the duplicate `/game` entries in the `handlers` Map. Can implement the `get` of `CreateGame` and then also remove the work around.

https://github.com/bafolts/terraforming-mars/blob/0e0e31a90a2af98e828d0b339a29b89adb53e628/src/server.ts#L72

kberg:
> Will need to consider the duplicate `/game` entries in the `handlers` Map. Can implement the `get` of `CreateGame` and then also remove the work around.
> 
> https://github.com/bafolts/terraforming-mars/blob/0e0e31a90a2af98e828d0b339a29b89adb53e628/src/server.ts#L72

This is all easy to fix because Handler multiplexes based on the request method.

src/server.ts
kberg: Done.

bafolts (approved):

PR #2889, opened 2021-03-05
Refactor ServeAsset for readability: Relies on handler2

kberg:


src/routes/ServeAsset.ts
```diff
@@ -0,0 +1,151 @@
...
+        if (resolvedFile.startsWith(assetsRoot)) {
+          return {file: resolvedFile};
+        }
+      }
+    }
+
+    return {};
+  }
+
+  private supportedEncodings(req: http.IncomingMessage): Set<Encoding> {
```
bafolts, line 138: This can be written as:

```typescript
  private supportedEncodings(req: http.IncomingMessage): Set<Encoding> {
    return new Set(req.headers['accept-encoding'] as (Array<Encoding> | undefined));
  }
```

Should be more efficient and less complex.

kberg: **Thank you.**

bafolts (approved):

Issue #2577, opened 2021-01-27
Rethinking the Venus Milestone/Award config option


kevinb9n:
Assuming the Venus Next expansion is enabled. There are two different cases:

**When "Random Ms/As" is disabled:**

... Are we sure that users really mind sticking to the published rules? Being able to uncheck "Venus M/A" is a straight nerf to Morning Star, Celestic, Stormcraft, Titan, and expensive cards that add floaters quickly... none of which really need nerfing.

**When "Random milestones" is enabled:**

... In this case, sticking to the published rule (Venus M/A always included) would just feel too weird, so the cards mentioned above will be weakened. However, the Venus M/A should really *always* be in the pool to choose from – I don't think we need an option to turn them off. And it should always just choose 5 because there is no real reason to need 6. So, the cards are debuffed only in *some* games.

If the behavior was as described above, would we really need the option at all?

Some of the above should apply equally to Ares, it just doesn't have the same 'Ares milestone' config option to get rid of.

kevinb9n:
Saw this in #vote-for-issues. Note that removing an option isn't generally something people will vote for. Also this issue is about making things work in a different way so that the option is even less needed, which won't be clear there.

kevinb9n:
Happy to re-grab this if more votes come in, but for now, it's telling a story:

<img width=650 src=https://user-images.githubusercontent.com/934551/107828471-cc1fc980-6d3d-11eb-8bdc-f02c3225040c.png>

This addresses only the random-milestones case. So far it suggests that:

* My predictions that users don't really want 3 or 4 were probably right.
* My prediction that users don't really want 2 was wrong. There is interest in both 5-milestone and 6-milestone games. However, this very unlikely has anything to do with Venus Next specifically; they would just like an option to control the number in general, whenever they use randomization. (I asked for anyone to say otherwise if it was Venus-related, nothing yet.)

I didn't poll about what people want when NOT randomizing. I doubt we'd head hear from many discord users about that. I still think it unlikely that they actually care about banning Hoverlord/Venuphile against the published rules.

All this suggests that users may be happy with:

* NO "Venus Milestone/Award" option
* When randomizing is off, always add them
* When randomizing is on, always have them in the pool, and have an option for whether to pick 5 or 6

I think Ares could work the same way, even though it does mean some games will have *seven* milestones.

kberg:
That makes sense. The Moon will add one more of each.

I envision being able to move milestone and award selection fully to the
[...]

PR #2894, opened 2021-03-06
Refactor the MIlestoneAward selelctor

kberg:
This in part starts to make #2577 easier, but mostly it's so I can integrate the Moon's milestones and awards into this without too much trouble. It also clarifies and simplifies a few changes.

This change is best reviewed commit-by-commit.

I have a few more changes coming, but this is a good starting point.

### DIFF
--- a/src/MilestoneAwardSelector.ts
+++ b/src/MilestoneAwardSelector.ts
@@ -40,17 +40,24 @@ import {Tycoon} from './milestones/Tycoon';
 import {RandomMAOptionType} from './RandomMAOptionType';
 
 export namespace MilestoneAwardSelector {
-  // exported for testing.
-  export const MA_ITEMS = [
+  const MILESTONES = [
     ...ORIGINAL_MILESTONES,
     ...ELYSIUM_MILESTONES,
     ...HELLAS_MILESTONES,
     ...VENUS_MILESTONES,
+  ];
 
+  const AWARDS = [
     ...ORIGINAL_AWARDS,
     ...ELYSIUM_AWARDS,
     ...HELLAS_AWARDS,
     ...VENUS_AWARDS,
+  ];
+
+  // exported for testing.
+  export const MA_ITEMS = [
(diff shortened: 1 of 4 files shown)


src/MilestoneAwardSelector.ts
```diff
@@ -26,292 +29,384 @@ import {Generalist} from './milestones/Generalist';
...
-  bind(Terraformer, Landlord, 1);
-  bind(Terraformer, Thermalist, 1);
-  bind(Terraformer, DesertSettler, 1);
-  bind(Terraformer, EstateDealer, 1);
-  bind(Gardener, Ecologist, 1);
-  return array;
-}
+  class Synergies {
+    // This map uses keys "X|Y" and of course stores strings as both
+    // "Y|X" and "X|X"
```
bafolts, line 86: I am not sure why this says "of course stores" as it is strange to me that the key is stored for `Y|X and X|Y`. The comment indicates we store `Y|X and X|X` should this be `Y|X and X|Y`?

bafolts (approved):

src/MilestoneAwardSelector.ts
kberg: Gotcha. That needs to be expanded and made more clear.

kberg: Replaced it with

```
    // This map uses keys of the format "X|Y" where X and Y are MA names.
    // Entries are stored as "X|Y" and also "Y|X"; it just makes searching
    // slightly faster. There will also be entries of the type "X|X".
    //
    // I honestly don't remember why "X|X" is useful, and it's possible
    // it's no longer necessary. That's something that should be carefully conisdered
    // and possibly removed, and not just propagated because that's what we had to begin with.
    // In other words, someone figure out why, and preserve it, and document why, or
    // be certain it's unnecessary and remove this paragraph and the code below that sets "X|X" to 1000.
```

PR #2895, opened 2021-03-06
Replace number indexes with strings in the milestone award selector.

kberg:
Based on the #2894


kberg:
Ugh, I actually put all of this in the prior PR. Closing.

PR #2902, opened 2021-03-07
Seriously clean up card requriements in the moon. It's even fewer LOC.

kberg:
Thankfully the tests that were already in place carefully prove out all the requirements cases, so that helped a great deal here.


sienmich (approved):
Nice, gj!

kberg:
> Nice, gj!

Thanks for the framework to make it possible.

PR #2903, opened 2021-03-07
Moon viz fixes, addressing highest level issues.

kberg:
image

image
image
image
image
image
image
image
image
image
image
image
image
image
image
image


bafolts (approved):

src/cards/moon/CopernicusTower.ts
```diff
@@ -12,28 +12,28 @@ import {OrOptions} from '../../inputs/OrOptions';
...
+      cardType: CardType.ACTIVE,
       tags: [Tags.SCIENCE, Tags.MOON],
       cost: 36,
       resourceType: ResourceType.SCIENCE,
       requirements: CardRequirements.builder((b) => b.production(Resources.TITANIUM, 2)),
 
       metadata: {
-        description: 'Requires that you have 2 Titanium production. 1VP per each Moon tag you have.',
         cardNumber: 'M72',
         victoryPoints: CardRenderDynamicVictoryPoints.moon(1, 1),
```
chosta, line 29: The moon on the victory points seems too big in comparison with the other elements that usually go there. Please, consider introducing a small or medium size markup.

chosta (approved):
These are pretty good considering the cusom work needed! The only 2 things that bother me are:
1) The size of the moon on top of VP container
2) The requirements title for HE3 Production quota. Unfortunately, this looks like it needs some extra work and can't be just patched. We can let go for now and find a solution in another PR

src/cards/moon/CopernicusTower.ts
kberg: I'll fix that in another PR so this gets in.

kberg:
Thanks for the review, both of you.

kberg:
I'll fix the moon tag in the VP container shortly.

PR #2908, opened 2021-03-08
Add Ares to the MIlestone Award selector.

### LINKED ISSUE #2577: Rethinking the Venus Milestone/Award config option

kberg:
This doesn't close #2577 but it advances it.


bafolts (approved):

kberg:
I rewrote MA Selector to include an initial synergy from the two Ares MAs to all the other MAs based on @nwai90's suggestion.

bafolts (approved):

kberg:
DAMN, it still doesn't work. MA Selector isn't generating data.
```
Starting server on port 8081
#100
/home/kberg/git/terraforming-mars/build/src/MilestoneAwardSelector.js:257
            throw new Error('No limited synergy milestones and awards set was generated after ' + maxAttempts + ' attempts. Please try again.');
            ^

Error: No limited synergy milestones and awards set was generated after 5 attempts. Please try again.
    at getRandomMilestonesAndAwards (/home/kberg/git/terraforming-mars/build/src/MilestoneAwardSelector.js:257:19)
    at getRandomMilestonesAndAwards (/home/kberg/git/terraforming-mars/build/src/MilestoneAwardSelector.js:277:28)
```

kberg:
image

1,000,000 runs of limited synergy fails 1,525 times, or 00.15%, which means that it will reject ~ once every 655 times

Reducing the default synergy to  4 and 3 result in even worse failures:

4: 2989 errors (!) out of 1 million
3: 6685???

If I remove the default value altogether, there are 0 failures in 1,000,000 runs. Removing the default.

nwai90 (approved):

vsrisuknimit:
@kberg @nwai90 Could it be that something is changed and now wrong in the selection algorithm?

Here is the old data for why I set the number to 5 to prevent infinite loop. I believe I tried 1 million before and got similar data. 5 times should be more than enough.

https://github.com/bafolts/terraforming-mars/pull/1637#issuecomment-711411034

vsrisuknimit:
Ah, my bad, I see you remove the default value for all milestones, not default value for the 5 try limit. Sorry, only look at the comment, not the code.

PR #2911, opened 2021-03-08
Provide custom viz of Terraformer milestone when Turmoil is in effect

kberg:
This is a WIP. Wondering if there's any reason this would be a mistake to implement.


bafolts:
Code wise this makes sense to me. I never play the game to comment on the visual change.

kberg:
This is becoming a lot of work just to not show "or 26 when using Turmoil" in the description. I'm abandoning it.

PR #2921, opened 2021-03-09
Services work for spectator mode.

bafolts:
This change puts in place the services work to enable a spectator mode. This will allow creating a new component (SpectatorHome) which displays the game as if you were a spectator. The spectator id is generated along with game id and player ids when the game is created.

Will still need to put in polling so that the page updates when the game age has changed. I plan to also add a `timeline` property so that along with `gameAge` we know when the game state has changed. Due to the undo operation we will also need both `gameAge` and `timeline` to know the current age of the game on the most recent timeline. If `timeline` is larger than the timeline you have on client or the `gameAge` differs we will need to query the server for most recent game information.


src/database/GameLoader.ts
```diff
@@ -37,11 +37,14 @@ export class GameLoader implements IGameLoader {
   // player ids which we know exist mapped to game id
   private readonly playerIds = new Map<PlayerId, GameId>();
 
+  // spectator ids which we know exist mapped to game id
+  private readonly spectatorIds = new Map<SpectatorId, GameId>();
```
kberg, line 41: This map could merge with playerId, which could be Map<PlayerId|SpectatorId> /shrug

src/database/GameLoader.ts
```diff
@@ -37,11 +37,14 @@ export class GameLoader implements IGameLoader {
...
+
   // requests for game that are waiting to load
-  private readonly gameRequests: Array<[GameId, boolean, LoadCallback]> = [];
+  private readonly gameRequests: Array<() => void> = [];
 
   // requests for player that are waiting to load
-  private readonly playerRequests: Array<[PlayerId, LoadCallback]> = [];
+  private readonly entityRequests: Array<() => void> = [];
```
kberg, line 47: Comment above still says player.

Are you using this to define the term 'entity' as player or spectator? It's kind of general.

src/database/GameLoader.ts
```diff
@@ -37,11 +37,14 @@ export class GameLoader implements IGameLoader {
   // player ids which we know exist mapped to game id
   private readonly playerIds = new Map<PlayerId, GameId>();
 
+  // spectator ids which we know exist mapped to game id
+  private readonly spectatorIds = new Map<SpectatorId, GameId>();
+
   // requests for game that are waiting to load
-  private readonly gameRequests: Array<[GameId, boolean, LoadCallback]> = [];
+  private readonly gameRequests: Array<() => void> = [];
```
kberg, line 44: I wonder why the array type has changed.

kberg (commented):
Once we add a spectator, someone can use that link to get player IDs from the web request, and we're back to where we were before.

src/routes/Game.ts
```diff
@@ -115,6 +116,7 @@ export class GameHandler extends Handler {
         } else {
           const seed = Math.random();
           const game = Game.newInstance(gameId, players, players[firstPlayerIdx], gameOptions, seed);
+          game.spectatorId = spectatorId;
```
kberg, line 119: Can you put SpectatorId in the constructor?

src/routes/Game.ts
```diff
@@ -39,6 +39,7 @@ export class GameHandler extends Handler {
       try {
         const gameReq = JSON.parse(body);
         const gameId = generateRandomId();
+        const spectatorId = generateRandomId();
```
kberg, line 42: May I suggest adding a leading `s-` to spectator id?

kberg: (In fact, leading `s-`, `p-`, and `g-`.)

src/database/GameLoader.ts
kberg: Yeah, can you tell me a bit about why you made this change? I'm a little unclear why it's a good thing.

kberg (commented):
Good start.

bafolts:
> Once we add a spectator, someone can use that link to get player IDs from the web request, and we're back to where we were before.

As long as `SpectatorModel` never exposes this then it shouldn't be an issue. May need a separate log request so we don't expose player id through log messages.

src/routes/Game.ts
bafolts: Yes you may. This should also indirectly avoid collisions between the three.

kberg:
> > Once we add a spectator, someone can use that link to get player IDs from the web request, and we're back to where we were before.
> 
> As long as `SpectatorModel` never exposes this then it shouldn't be an issue.

OK.

src/database/GameLoader.ts
bafolts: This change was actually to allow for `entityRequests`. Instead of having to pass `SpectatorId` or `PlayerId` this avoids having to pass the data through altogether. Instead of `gameRequests` a better name may now be `gameRequestQueue`. It is storing FIFO callbacks which are executed once the database has loaded every game id.

src/database/GameLoader.ts
bafolts: Yes. Can you think of a better name than 'entity'?

src/database/GameLoader.ts
bafolts: I can do that with the change to use a `p` and `s` suffix in front of the `PlayerId` and `SpectatorId`. I was trying to avoid collisions between `SpectatorId` and `PlayerId` in the map.

src/database/GameLoader.ts
kberg: Well, a perhaps a generic term for player | spectator is _participant_. But consider just calling this map `loadRequests` anyway.

kberg: Actually, this isn't even a request for entities, it's now just an 'onGameIdsLoaded' set of callbacks.

bafolts: YES! That is the right answer. Will change to `onGameIdsLoaded`.

kberg: There you go. But also, I think the notion of _participant_ isn't bad, for future stuff.

src/database/GameLoader.ts
```diff
@@ -160,17 +175,9 @@ export class GameLoader implements IGameLoader {
 
   private loadNextGame(): void {
```
kberg, line 176: This method is now a little unclear. It's called loadNextGame() but it's just pulling a callback from onGameIdsLoaded. Change the name to ... /shrug?

kberg (approved):
One minor comment for clarity.

src/routes/Game.ts
```diff
@@ -11,9 +11,9 @@ import {Server} from '../server/ServerModel';
 import {ServeAsset} from './ServeAsset';
 
 // A copy from server.ts. Please dedupe.
```
kberg, line 13: Since there's a copy in server.ts, and they now behave different ... /shrug ?

src/database/GameLoader.ts
```diff
@@ -35,13 +37,10 @@ export class GameLoader implements IGameLoader {
...
-  private readonly playerIds = new Map<PlayerId, GameId>();
-
-  // requests for game that are waiting to load
-  private readonly gameRequests: Array<[GameId, boolean, LoadCallback]> = [];
+  private readonly onGameIdsLoaded: Array<GameIdCallback> = [];
 
-  // requests for player that are waiting to load
-  private readonly playerRequests: Array<[PlayerId, LoadCallback]> = [];
+  // participant ids which we know exist mapped to game id
+  private readonly participantIds = new Map<SpectatorId | PlayerId, GameId>();
```
kberg, line 43: I wish I could add comments to arbitrary lines. But, hey, while you're in here, since the constructor does no real work, you can simplify this a bit, and remove the '?' from `instance`'s definition.

```
  private static instance = new GameLoader();

  private constructor() {}

  public static getInstance(): IGameLoader {
    return GameLoader.instance;
  }
```

bafolts:
@vsrisuknimit this should allow to start filling in `SpectatorHome` with what we want to display on spectator page. Anything on `SpectatorModel` will be public so will have to contain public information only.

vsrisuknimit:
@bafolts Thank you. I will try to work on it.

PR #2933, opened 2021-03-11
only show buffer gas for solo games

bafolts:
The check to show buffer gas was relying on the game option versus checking for solo mode. This moves the check to `Player` and looks for solo mode. Unit test was added to verify this logic.

### DIFF
--- a/tests/TestPlayer.ts
+++ b/tests/TestPlayer.ts
@@ -31,6 +31,10 @@ export class TestPlayer extends Player {
 
   public tagsForTest: Partial<TagsForTest> | undefined = undefined;
 
+  public getStandardProjectOption() {
+    return super.getStandardProjectOption();
+  }
+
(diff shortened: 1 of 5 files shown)


tests/TestPlayer.ts
```diff
@@ -31,6 +31,10 @@ export class TestPlayer extends Player {
 
   public tagsForTest: Partial<TagsForTest> | undefined = undefined;
 
+  public getStandardProjectOption() {
```
kberg, line 34: Is this necessary? Interesting.

kberg (approved):

tests/TestPlayer.ts
bafolts: I believe it is necessary. It is now `protected` in `Player` and made `public` for testing here on `TestPlayer`.

PR #2935, opened 2021-03-11
Add Moon MAs

kberg:
**Note:** Baseline is PR 2908


vsrisuknimit:
Once you are all done with the Milestones and Awards change, can we rerun the test on the randomizer distribution? Just for sanity check.

Randomize with Venus, without Venus, with Moon, or with Ares.

kberg:
Actually -- we should do it **before.** A couple of observations during
development makes me nervous that I broke it, but didn't think about how to
deal with it, and that's exactly how - thanks.

kberg:
Ha! Not yet:

image

kberg:
Ah, fixed. A simple issue about naming.

image

bafolts (approved):

bafolts:
I :heart: this chart.

kberg:
> I  this chart.

Yeah, it's nice. More charts. I want more charts.

PR #2937, opened 2021-03-11
Add Analyze MA tool.

kberg:
image

### DIFF
--- a/src/tools/analyze_ma.html
+++ b/src/tools/analyze_ma.html
@@ -1,0 +1,81 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<html>
+<head>
+
+<!-- Load d3.js -->
+<script src="https://d3js.org/d3.v4.js"></script>
+
+<form>
+  <label for="runs">Runs: </label></label><input type="text" id="runs" value="10000" name="runs"></input><br/>
+  <label for="venus">Venus: </label></label><input type="checkbox" id="venus" name="venus"></input><br/>
+  <label for="ares">Ares: </label></label><input type="checkbox" id="ares" name="ares"></input><br/>
+  <label for="moon">The Moon: </label></label><input type="checkbox" id="moon" name="moon"></input><br/>
+
+  <input type="radio" id="none" name="type" value="NONE"><label for="none">None</label>&nbsp;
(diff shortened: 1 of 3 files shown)


bafolts:
Love that this project is far enough along that a page like this makes sense to add.

kberg:
It's not an additional page. It's a command-line tool.

kberg:
Still, I'm happy to have it.

src/tools/analyze_ma.html
```diff
@@ -0,0 +1,81 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
```
bafolts, line 2: I believe the `meta` element goes inside the `<head>` element

src/tools/analyze_ma.html
```diff
@@ -0,0 +1,81 @@
...
+<html>
+<head>
+
+<!-- Load d3.js -->
+<script src="https://d3js.org/d3.v4.js"></script>
+
```
bafolts, line 8: This needs a closing `</head>` and an opening `<body>` element.

src/tools/analyze_ma.html
```diff
@@ -0,0 +1,81 @@
...
+<form>
+  <label for="runs">Runs: </label></label><input type="text" id="runs" value="10000" name="runs"></input><br/>
```
bafolts, line 10: Duplicate `</label>` here and on next 3 lines.

bafolts (approved):

bafolts:
As it is a command line tool you may be able to generate the SVG and save it to a file versus using the HTML page. Taking in the arguments through CLI.

src/tools/analyze_ma.html
kberg: Done.

src/tools/analyze_ma.html
kberg: Done.

src/tools/analyze_ma.html
kberg: Fixed. Thanks!

kberg:
> As it is a command line tool you may be able to generate the SVG and save it to a file versus using the HTML page. Taking in the arguments through CLI.

Absolutely. Anyone who wants to take the next step now has something to bulid on.

PR #2942, opened 2021-03-12
Adding MoonExpansion.ifElseMoon, which should have been separate from ifMoon to begin with.

kberg:


src/cards/moon/LunaProjectOffice.ts
```diff
@@ -37,7 +37,7 @@ export class LunaProjectOffice extends Card implements IProjectCard {
         }
       }
       return false;
-    }) || false;
+    }, () => false);
```
bafolts, line 40: May not even work and probably not helpful but when looking at `() => false` I was wondering if that would be the equivalent of `Boolean`. I think could write `}, Boolean);` on this line. Would be far harder to read but cool of javascript to have this function ready.

kberg: Agreed on all counts. :D

bafolts (approved):

PR #2944, opened 2021-03-12
Trim the expansion suffix off card names.

kberg:
Before
image

After
image

### DIFF
--- a/src/components/GameEnd.ts
+++ b/src/components/GameEnd.ts
@@ -154,13 +154,18 @@ export const GameEnd = Vue.component('game-end', {
                                 <span :class="'log-player ' + getEndGamePlayerHighlightColorClass(p.color)"><a :href="'/player?id='+p.id+'&noredirect'">{{p.name}}</a></span>
                             </div>
                             <div v-for="v in p.victoryPointsBreakdown.detailsCards">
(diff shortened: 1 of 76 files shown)


src/components/GameEnd.ts
```diff
@@ -154,7 +154,8 @@ export const GameEnd = Vue.component('game-end', {
                                 <span :class="'log-player ' + getEndGamePlayerHighlightColorClass(p.color)"><a :href="'/player?id='+p.id+'&noredirect'">{{p.name}}</a></span>
                             </div>
                             <div v-for="v in p.victoryPointsBreakdown.detailsCards">
-                                {{v}}
+                                <!-- Removing any expansion-specific suffix-->
+                                {{v.split(':')[0]}}
```
nwai90, line 158: Is it also intended to remove the `: VP_AMOUNT` after each card's name?

bafolts: Guessing this wasn't intended. May need to strip this on the server.

kberg: Whoops.

kberg:
I've rewritten this to be a different kind of PR. Closing it until I test it satisfactorally manually.

PR #2948, opened 2021-03-13
Clean up TestingUtils

kberg:
Finishing up using a TestingUtils namespace, and removing TestPlayers from TestingUtils.

This PR is big. :/ I can split it in chunks, please advise your preference.


bafolts:
I will review this as-is with one commit through CLI.

bafolts (approved):

PR #2952, opened 2021-03-13
Add a (disabled) prototype for a sound when the log changes

kberg:
There's a lot that's not good about this, yet. I could spend more time working on the web audio API, but at least this will a) help make it easier for other developers to start thinking about how sounds should change, and b) it is hidden -- it won't make sound unless you have an `experimental=1` parameter in your browser.

Really what I want is a quarter second buzz. I've done sounds before, but not with the web API, so it'll be a while. Plus, it has to be basically battle-tested before considering enabling it.


kberg:
I thought about adding the `node-web-audio` package so I could write a command-line tool to test sounds but eh, one thing at a time. But I wish there were a faster development cycle.

src/components/LogPanel.ts
```diff
@@ -203,6 +205,9 @@ export const LogPanel = Vue.component('log-panel', {
         if (xhr.status === 200) {
           messages.splice(0, messages.length);
           messages.push(...xhr.response);
+          if (PreferencesManager.loadValue('enable_sounds') === '1' && window.location.search.includes('experimental=1') ) {
```
bafolts, line 208: There is a new `loadBooleanValue` which can be used here to avoid having to check for `=== '1'`.

src/components/SoundManager.ts
```diff
@@ -0,0 +1,84 @@
...
+    const oscillator = audioCtx.createOscillator();
+    oscillator.type = 'sine';
+    oscillator.frequency.value = frequency;
+    oscillator.connect(gainNode);
+
+    return oscillator;
+  }
+
+  function createAudioContext() {
+    const audioCtx = window.AudioContext || (window as any).webkitAudioContext ? new AudioContext() : undefined;
```
bafolts, line 44: You most likely only moved this code but the check for `(window as any).webkitAudioContext)` can be removed as it isn't being used. This code is checking that the `AudioContext` constructor or the older `webkitAudioContext` constructor exists but then only makes use of the `AudioContext` constructor. If the `webkitAudioContext` exists but `window.AudioContext` does not exist this line is going to throw a runtime error that `AudioContext` is not a function.

src/components/SoundManager.ts
```diff
@@ -0,0 +1,84 @@
...
+  function playSound(audioCtx: AudioContext, frequency: number, time: number, len: number) {
+    const gainNode = setupGainNode(audioCtx, time);
+    const oscillator = setupOscillator(audioCtx, frequency, gainNode);
+
+    oscillator.start(time);
+    gainNode.gain.exponentialRampToValueAtTime(0.001, time + len);
+    oscillator.stop(time + len);
+  }
+
+  function playInContext(cb: (audioCtx: any) => void) {
```
bafolts, line 59: This doesn't have to use `any`. This can use `AudioContext` as the type of the `audioCtx` argument.

src/components/SoundManager.ts
```diff
@@ -0,0 +1,84 @@
...
+    const audioCtx = createAudioContext();
+    if (audioCtx === undefined) return;
+
+    audioCtx.resume().then(() => {
```
bafolts, line 63: The server side has avoided promises thus far. In this situation the audio context is requiring the use of promises. I would use `await` and `async` versus calling `then`. Calling `then` can be harder to read and this also isn't calling to `catch` for errors.

```
async function playInContext(cb: (audioCtx: AudioContext) => void) {
  const audioCtx = createAudioContext();
  if (audioCtx === undefined) return;
  await audioCtx.resume();
  cb(audioCtx);
}
```

bafolts (changes requested):
Minor tweaks suggested. Moving `SoundManager` into `components` is good move.

bafolts:
Using and playing wav files may be larger code but should be simpler to develop. Storing them as assets would be simple enough. Could be far less time consuming to develop.

kberg:
> Using and playing wav files may be larger code but should be simpler to develop. Storing them as assets would be simple enough. Could be far less time consuming to develop.

I sure agree. /shrug

src/components/LogPanel.ts
kberg: Thanks, done.

src/components/SoundManager.ts
kberg: Thank you.

src/components/SoundManager.ts
kberg: Do you mind if I make this change in another PR?

src/components/SoundManager.ts
kberg: OK. Please double-check that I got this right

kberg: Because _this_ doesn't make sense,either.

```
    const audioCtx = window.AudioContext || new AudioContext();

    if (audioCtx === undefined) console.log('This web browser doesn\'t support Web Audio API');
    return audioCtx;
  }
```

kberg: Ah, the problem is that if I remove this, then typescript gets fussy. I'm undoing the changes I made based on your suggestion. If you want to dig in, please do! Or teach more, please.

bafolts: With our stated browser support I would assume that `AudioContext` exists.

https://caniuse.com/?search=AudioContext

Otherwise this would become:

```
const audioCtx = window.AudioContext ? new AudioContext() : undefined;
```

bafolts: The `window.AudioContext` and `AudioContext` are pointing to the same function. On the browser any variable declared globally will be a property of `window`.

bafolts (approved):

src/components/SoundManager.ts
kberg: OK, done.

PR #2953, opened 2021-03-13
use undoCount with gameAge for detecting changes

bafolts:
This change introduces an `undoCount` property to track undo operations. The `gameAge` does not suffice to detect changes if another player performs an undo. Plan to also use `undoCount` along with `gameAge` to generate an `ETag`. If neither `undoCount` or `gameAge` changes we can re-use responses.


src/Game.ts
```diff
@@ -154,6 +154,7 @@ export class Game implements ISerializable<SerializedGame> {
   public deferredActions: DeferredActionsQueue = new DeferredActionsQueue();
   public gameAge: number = 0; // Each log event increases it
   public gameLog: Array<LogMessage> = [];
+  public timeline: number = 0; // Each undo increases it
```
kberg, line 157: So, in one sentence, what is the timeline?

src/Game.ts
```diff
@@ -1638,6 +1640,7 @@ export class Game implements ISerializable<SerializedGame> {
     game.generation = d.generation;
     game.phase = d.phase;
     game.oxygenLevel = d.oxygenLevel;
+    game.timeline = d.timeline ?? 0;
```
kberg, line 1643: This is a nice solution in this case.

src/SerializedGame.ts
```diff
@@ -52,6 +52,7 @@ export interface SerializedGame {
     someoneHasRemovedOtherPlayersPlants: boolean;
     syndicatePirateRaider: PlayerId | undefined;
     temperature: number;
+    timeline: number;
```
kberg, line 55: Shouldn't this be `?: number` ?

src/components/WaitingFor.ts
```diff
@@ -84,7 +84,7 @@ export const WaitingFor = Vue.component('waiting-for', {
       clearTimeout(ui_update_timeout_id);
       const askForUpdate = () => {
         const xhr = new XMLHttpRequest();
-        xhr.open('GET', '/api/waitingfor' + window.location.search + '&prev-game-age=' + this.player.gameAge.toString());
+        xhr.open('GET', '/api/waitingfor' + window.location.search + '&prev-game-age=' + String(this.player.gameAge) + '&prev-timeline=' + String(this.player.timeline));
```
kberg, line 87: Is new String necessary?

`'/api/waitingfor${window.location.search}&prev-game-age=${this.player.gameAge}&prev-timeline=${this.player.timeline}`?

Pedantic question, why does the parameter have a `prev-` prefix?

src/Game.ts
bafolts: The value of `timeline` will be equal to the number of undo actions that have been used.

A better name may be `undosUsed`? The problem I am trying to solve here is with caching. The `gameAge` gets incremented with every log action. During an undo action it goes back to it's previous value. This means that in a multi player game if a client had `gameAge` of 50 they may not necessarily have the most recent `gameAge` of 50. If another player were to use an undo and take an action before they checked in with the server they may think they still have the new `gameAge` of 50 when they don't. Adding timeline here would catch the situation.

The `gameAge` combined with `timeline` can be used to hash a `Game`. I plan to join them in a string to form an `ETag` as well. The string `${spectatorId}:${gameAge}:${timeline}:waitingFor` can soon be used as an `ETag` on the waiting for route. As long as `gameAge` and `timeline` hasn't changed we can response with a 304.

I used `timeline` as name thinking of it in terms of a retroactive data structure. Whenever we do an undo and go into the past we change the future and create another timeline. Any client with the old timeline needs to update to the new timeline.

src/SerializedGame.ts
bafolts: This could be `timeline?: number`. I should add a todo to where I used `timeline ?? 0` while deserializing. I only introduced the `undefined` to get through the transition where old data in the database doesn't have this value defined. In a game that doesn't use `undo` the timeline would always be `0` representing the 0th timeline.

src/components/WaitingFor.ts
bafolts: Calling to `String` is unnecessary. The `prev-` prefix is also unnecessary. I copied it as it was being used with `prev-game-age`. I can change to `gameAge` and `timeline`.

src/Game.ts
kberg: I see the `timeline` idea now. Interesting. I think if you added a comment in its declaration saying 
`The number of times this game has had an undo action. Like the infinite universe theory and its many timelines.` then this would make more sense, but `rewindCount` or `undoCount` or `undoesUsed` might be easier to consume.

kberg (approved):
Approve modulo the changes discussed.

bafolts:
I want to look into what caused the outage yesterday before spending time on this. Plan to make the changes and merge this tomorrow or day after.

kberg:
Okie doke.

PR #2957, opened 2021-03-14
Make Moon tile placement resilient to onTilePlayed where cards think they're on Mars.

### LINKED ISSUE #2950: [Moon]  Broken Moon Road placement
> After playing Tycho Road Network i got to place additional roads after the first one:

image

>  after filling the moon I could not get out of this

image

> After playing Momentum Virum Habitat in an earlier generation the game proceeded, the global colony track did not increase, I got the production and kept the card in my hand

kberg:
Fixes #2950


src/boards/BoardId.ts
```diff
@@ -0,0 +1,4 @@
+export enum BoardId {
```
bafolts, line 1: Using `BoardType` may be clearer than `BoardId`. Originally thought this could be `BoardPlanet` but the moon is not a planet. Using `BoardCelestialBody` may be too verbose but may be most astronomically correct.

bafolts (approved):
I think `BoardType` would be clearer than `BoardId`. There is already `BoardName` so having a `BoardId` and a `BoardName` may lead to confusion. Even `BoardType` could fall into this category where it seems they are all the same thing. I have bias towards it but using `BoardCelestialBody` would be verbose and I don't think would cause confusion as it makes it clear that it represents the celestial body that the board takes place on.

src/boards/BoardId.ts
kberg: I've had those thoughts many times. I agree `BoardId` is bad. `Board` would be better, but it's taken. `BoardType` it is. (I almost wish it was `Board.Type`.

PR #2968, opened 2021-03-16
Finish up moving code to Handlers.

kberg:


kberg:
Once this goes in I will spend more time on oauth

src/routes/GamesOverview.ts
```diff
@@ -0,0 +1,18 @@
+import * as http from 'http';
+import {Handler} from './Handler';
+import {IContext} from './IHandler';
+import {ServeApp} from './ServeApp';
+import {ServeAsset} from './ServeAsset';
+
+// A strange way to get a games overview, by serving index.html with a validated server id.
+// Is this hackable?
```
bafolts, line 8: I don't think this is hackable. This secret data comes through `/api/games` which requires a `serverId` and also has `validateServerId` set to `true`.

bafolts (approved):

kberg:
Thank you!

Issue #2855, opened 2021-02-26
Active player indication


daniel-sousa-me:
There is no obvious way to see that it's my turn just by looking at the page. Taking inspiration from 18xx.games I made a modification on the server I'm running so that the left bar shows in a different color if it's your turn.

![image](https://user-images.githubusercontent.com/3337865/109355411-15603500-7877-11eb-947d-84053d1ce2fb.png)

Would you like me to make a PR with this change? (or some adaptation of it)

vsrisuknimit:
@daniel-sousa-me This looks cool but it may or may not be necessary. We current have three ways to tell someone it is their turn to take an action.

1) The browser title has a little clock running when it is your turn.
![image](https://user-images.githubusercontent.com/14239220/109367052-0a9ba500-7863-11eb-9af6-3d79b60cbe88.png)
2) The pop-up browser notification
![image](https://user-images.githubusercontent.com/14239220/109367081-21da9280-7863-11eb-9d43-dbfa607cc861.png)
3) The label on your player mat
![image](https://user-images.githubusercontent.com/14239220/109367118-37e85300-7863-11eb-90c9-251828400922.png)

Two things that I would consider before adding this feature: (A) will the bright color sidebar be too distracting? Perhaps that it is the point to know it is your turn, but (1) and (2) should already help with that. (B) Green/red color sidebar might not work well with the green checkmark.

I personally think we have enough ways to tell someone when it is their turn. But we could still add this feature or something along this line if you think that there are something missing in (1), (2) and (3) that could be helpful for the users.

kberg:
Thanks for the idea. I actually would have preferred that where the solid
purple square goes shows the active player. Your color is easy to remember
after a couple of rounds. It's certainly a better use of space than a
constant reminder of a piece of data that doesn't change all game.

vsrisuknimit:
> Your color is easy to remember after a couple of rounds.

You will be surprised how often people forgot their colors. Especially when they don't get to play their favorite colors. It is a frequent occurrence on Discord. I believe it is needed.

ssimeonoff:
When the whole sidebar is in a player colour  - it is a bit intrusive, tried that before and instead added the the "player cube" as a more subtle solution.

kevinb9n:
[...]

PR #2969, opened 2021-03-16
Sidebar player indicator (Issue #2855)

puuj:
This is an implementation of the sidebar-indicator proposed by @daniel-sousa-me with toned-down color proposed by  @vsrisuknimit. It uses the same logic as the highlight active CSS insertion, but excludes solo games where the color change is unnecessary.


src/components/Preferences.ts
```diff
@@ -137,6 +140,7 @@ export const Preferences = Vue.component('preferences', {
         this.$data[k] = PreferencesManager.preferencesValues.get(k);
         this.setPreferencesCSS(this.$data[k], k);
       }
+      this.setPreferencesCSS(this.active_player, 'active_player');
```
vsrisuknimit, line 143: It might be better to avoid using this method that is meant for something more complicated (the Preference toggles).

Alternatively, you can modify the first line of this vue template to:
`<div :class="'preferences_cont' + getActiveColorClass()" :data="syncPreferences()">`

And write a method to return a string depending on `active_player `. These strings are just css class names. One for not-active. One for active.

See `getBoardColorClass` for example.

src/components/PlayerHome.ts
```diff
@@ -104,17 +104,16 @@ export const PlayerHome = Vue.component('player-home', {
         }
       }
     },
+    isPlayerActive: function(player: PlayerModel) : boolean {
+      return player.players.length > 1 && (player.needsToDraft || (player.needsToDraft === undefined && player.isActive));
```
vsrisuknimit, line 108: Does this also cover need to discard from pluto and need to gain resource from Turmoil? Or are those under isActive already? I am not sure.

daniel-sousa-me: Iirc it doesn't cover Philares's power

src/components/Preferences.ts
vsrisuknimit: A few more lines to write but I think it will be a lot easier to understand in the future.

puuj: Updated as requested, agreed this is cleaner

src/components/PlayerHome.ts
puuj: Hm, this is the same logic used in the PlayerHome template to highlight the active player. The property comes from PlayerModel, and is set by the server code. Figuring out all the WaitingFor paths for  these cases is a bit tricky, but I would imagine that there is some deferredAction - waitingFor thing that happens to make isActive current.

puuj: Okay, I tested this and active is just the player whose turn it is. I updated the implementation to be based on whose action it is using the value of the waitingFor function

src/components/Preferences.ts
```diff
@@ -138,6 +141,13 @@ export const Preferences = Vue.component('preferences', {
         this.setPreferencesCSS(this.$data[k], k);
       }
     },
+    getActingPlayerClass: function(): string {
+      if (this.acting_player) {
```
vsrisuknimit, line 145: Ternary operator could be more concise here. It is used often through out the code base.

`return this.acting_player ?  'preferences_acting_player' : '';`

vsrisuknimit (approved):
Look good to me otherwise. But we should wait for real programmers to take a look.

src/components/Preferences.ts
vsrisuknimit: I might also set a background color anyway for non-active state. Currently it is a lack of background that is the color of the sidebar. That is not  good I think. If someone change the app main background in the future, we will inadvertently change the color of the side bar.

vsrisuknimit: `#303030` is the color of the current sidebar.

puuj: I made these changes in case you want to take a look.

vsrisuknimit: @puuj This looks good to me. Let's hope for a real dev to approve it.

bafolts (approved):

PR #2987, opened 2021-03-20
Game End page tables rework

ssimeonoff:
## Main goals
- icons in the table header 
- Combining the two wide cells (player+corporation) in one
- bigger numbers (so posted results in discord to be more visible without opening the original image)
- vp details columns of 4 players to fit the mobile width 
- vp details columns to easily show the points - vp are now aligned in their own sub-column
 
tables

Untitled2

## Further thoughts
- column MC -  I like the idea this to be completely gone and when there is a tie a MC icon with the value in it is displayed next to the Total vps for that player. E.g. 77 [50] - 77vp total and 50mc. I can't it at the moment, so I leave it for now as it is.
- time + mc columns position - I agree that those two show more of a meta info, not real VPs, but I think "Total" column being last looks better. 
- table VPs per generation - this is in the "pipeline" that will come in a different pr
Untitled3
A table that shows the total and gained vps per generation. The layout/visuals should be similar to the main table.

### DIFF
--- a/src/components/GameEnd.ts
+++ b/src/components/GameEnd.ts
@@ -25,9 +25,6 @@ export const GameEnd = Vue.component('game-end', {
     'Button': Button,
   },
   methods: {
-    getEndGamePlayerHighlightColorClass: function(color: string): string {
-      return playerColorClass(color.toLowerCase(), 'bg');
-    },
     getEndGamePlayerRowColorClass: function(color: string): string {
       return playerColorClass(color.toLowerCase(), 'bg_transparent');
     },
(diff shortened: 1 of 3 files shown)


kberg:
How will this look with The Moon?

kberg:
> column MC - I like the idea this to be completely gone and when there is a tie a MC icon with the value in it is displayed next to the Total vps for that player. E.g. 77 [50] - 77vp total and 50mc. I can't it at the moment, so I leave it for now as it is.

Why do you think those make sense on the left-hand side? Having them past the final score means they're less  in the way.

kberg:
In general, though this looks great.

ssimeonoff:
With the moon, the table will be wider, of course, but it won't be worse than before, just the same.
The "total" on the right as last column just feels right - having a "total" as row or column in a table is usually the last one, and I assume "the average user2 is quite accustomed to that.

kberg:
So first, a counterexample

image

I'll say that as the time and MC are secondary pieces of information, they are visual clutter. That's why I think the columns are exactly where they should be.

ssimeonoff:
Untitled
Untitled2

Current final look:
- no borders at all - keep it clean
- the meta info (mc and time) have opacity 0.5
- the "total" column is not last, but with the extra padding and font-size it should "pop out" enough
-todo: when Moon is released, the respective table header icons should be added for roads, colonies and mines

leggupdown:
Total on the far right makes the most sense to me for readability. When I look at a results page, I initially only care about player, corp, and VP. Far left and far right are the places readers look first. It also aligns with Steam (more players arriving by the day) and give an opportunity for future cool options like having the graphical elements like the VP animating to the right. 

The only choice in the OP I'm iffy on is the VP to the left of the card name on the bottom. It looks more like a numbered list in this format. Players look for cards, not points, so cards should be on the left. I care more if I see Antigrav than if I see "2." 

Love the reduced width.

src/components/GameEnd.ts
```diff
@@ -104,33 +101,34 @@ export const GameEnd = Vue.component('game-end', {
                     </a>
                 </div>
                 <div v-if="!isSoloGame() || player.isSoloModeWin" class="game-end-winer-announcement">
-                    <span v-for="p in getWinners()"><span :class="'log-player ' + getEndGamePlayerHighlightColorClass(p.color)">{{ p.name }}</span></span> won!
+                    <span v-for="p in getWinners()"><span :class="'log-player ' + getEndGamePlayerRowColorClass(p.color)">{{ p.name }}</span></span> won!
                 </div>
                 <div class="game_end_victory_points">
                     <h2 v-i18n>Victory points breakdown after<span> {{player.generation}} </span>generations</h2>
-                    <table class="table game_end_table">
+                    <table id="table" class="table game_end_table">
```
bafolts, line 108: Is this `id` attribute used? I can't see a reference to it from javascript or css.

bafolts (approved):

src/components/GameEnd.ts
ssimeonoff: oops, my bad, forgot to clean up that after I was done playing with html2canvas - in the future to have a download button, which when press, to save the table as png file.

PR #3012, opened 2021-03-24
Fix bugs of Moon Exp

jijiqw:
1. Topsoil Contract not work with Rust Eating Bacteria
2. Moonboard,I am  not sure if I am right, it seems  the space m11 should have 3 adjacent spaces: m06/ m12/ m17.
image


kberg (commented):
Great finds. Sorry for the issues!

kberg (approved):

kberg:
Thanks again.

PR #3017, opened 2021-03-24
Move runInput from Player to PlayerInputs.

kberg:
Now input validation will make more sense, and you won't have to deal with trying to make sense of the giant non-switch switch.

Tests fail and other things are strange. But what do you think about this change?

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -30,19 +30,14 @@ import {Phase} from './Phase';
 import {PlayerInput} from './PlayerInput';
 import {ResourceType} from './ResourceType';
 import {Resources} from './Resources';
-import {SelectAmount} from './inputs/SelectAmount';
 import {SelectCard} from './inputs/SelectCard';
 import {SellPatentsStandardProject} from './cards/base/standardProjects/SellPatentsStandardProject';
 import {SendDelegateToArea} from './deferredActions/SendDelegateToArea';
 import {DeferredAction} from './deferredActions/DeferredAction';
 import {SelectHowToPayDeferred} from './deferredActions/SelectHowToPayDeferred';
 import {SelectColony} from './inputs/SelectColony';
-import {SelectPartyToSendDelegate} from './inputs/SelectPartyToSendDelegate';
-import {SelectDelegate} from './inputs/SelectDelegate';
-import {SelectHowToPay} from './inputs/SelectHowToPay';
 import {SelectHowToPayForProjectCard} from './inputs/SelectHowToPayForProjectCard';
 import {SelectOption} from './inputs/SelectOption';
-import {SelectPlayer} from './inputs/SelectPlayer';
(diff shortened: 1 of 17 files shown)


src/Player.ts
```diff
@@ -1977,7 +1844,7 @@ export class Player implements ISerializable<SerializedPlayer> {
     this.waitingForCb = undefined;
     try {
       this.timer.stop();
-      this.runInput(input, waitingFor);
+      waitingFor.runInput(this, input);
```
kberg, line 1847: This is the key: trhe entire `runInput` method has been replaced with the `runInput` on the `watiingFor`.

src/Player.ts
```diff
@@ -799,126 +786,6 @@ export class Player implements ISerializable<SerializedPlayer> {
     }
   }
 
-  // This is only public for a test. It's not great.
-  // TODO(kberg): Fix taht.
-  public runInput(input: ReadonlyArray<ReadonlyArray<string>>, pi: PlayerInput): void {
-    if (pi instanceof AndOptions) {
```
bafolts, line 805: This change makes sense to me and should be well worth it. When adding a new `PlayerInput` we have to remember to add support for it here to this function. With that added `runInput` property that has to be defined it will require us to consider how the input is processed. Should also make it easier to unit test.

src/Player.ts
```diff
@@ -756,30 +749,24 @@ export class Player implements ISerializable<SerializedPlayer> {
...
     const foundCard = cards.find((card) => card.name === cardName);
     if (foundCard === undefined) {
       throw new Error('Card not found');
     }
     return foundCard;
   }
 
-  private runInputCb(result: PlayerInput | undefined): void {
+  // Not great that this is public. But not awful either
+  public runInputCb(result: PlayerInput | undefined): void {
```
bafolts, line 762: The `defer` function of `Game` is already public. Can probably move this `runInputCb` into the `PlayerInput` namespace.

src/Player.ts
```diff
@@ -756,30 +749,24 @@ export class Player implements ISerializable<SerializedPlayer> {
     return false;
   }
 
-  public getCard(cards: Array<IProjectCard>, cardName: string): IProjectCard {
+  // TODO(kberg): Make static, or put elsewhere. This doesn't belong here.
+  public getCard<T extends ICard> (cards: Array<T>, cardName: string): T {
```
bafolts, line 753: This is specific to `PlayerInput`. If no better place could put in the `PlayerInput` namespace. Could consider making `PlayerInput` an abstract class as well.

bafolts:
I think this change will be great. The `Player` class will shrink by a significant amount!

kberg:
I think the reason the tests are failing have to do with too much server side code creeping into the client. I'm going to abandon this PR for now. Also, this just means I think it's more urgent than before to separate code into client/server/common.

PR #3018, opened 2021-03-25
remove prev-game-age for good

bafolts:
I was dual writing `prev-game-age` and `gameAge` to handle the transition for clients which may had been polling with `prev-game-age`. Removing for good now.


kberg (approved):
Yay!

PR #3021, opened 2021-03-25
Update base_corporations.json

ltdstudio:
remove 5 corp to preludes.json


alrusdi (approved):
Thank you very much for this great work on chinese translation! Buе why don't you put all the files in one PR?

PR #3025, opened 2021-03-25
Make SelectHowToPayForProjectCard work with Moon cards.

kberg:
This commit does a few things: it deals with keeping track of what the
proper remaining steel and titanium a player may use and applies it
throughout the algorithm. That was the most complicated part -- no --
writing the tests was the most complicated part.

There's a bit of code in SHPTFPC which uses the player's played cards,
and includes the self replicating robots as its source for card data.
That just seems wrong. Of the many changes in this commit, that's the
one I'm most surprised is in the code, and the one I most want someone
to scrutinize and evaluate.

I rewrote the tests, but just a bit -- to take advantage of things like
Partial<>.

Finally, there's a server-side change which verifies that someone can't
overspend their steel, consequently, they can't go negative. Now, this
won't be possible anymore because of all the client-side changes, but
I have manually verified this with Luna Train Station.

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -756,14 +756,6 @@ export class Player implements ISerializable<SerializedPlayer> {
     return false;
   }
 
-  public getCard(cards: Array<IProjectCard>, cardName: string): IProjectCard {
-    const foundCard = cards.find((card) => card.name === cardName);
-    if (foundCard === undefined) {
-      throw new Error('Card not found');
-    }
-    return foundCard;
-  }
-
   private runInputCb(result: PlayerInput | undefined): void {
     if (result !== undefined) {
       this.game.defer(new DeferredAction(this, () => result));
@@ -800,7 +792,7 @@ export class Player implements ISerializable<SerializedPlayer> {
   }
 
   // This is only public for a test. It's not great.
-  // TODO(kberg): Fix taht.
+  // TODO(kberg): Fix that.
   public runInput(input: ReadonlyArray<ReadonlyArray<string>>, pi: PlayerInput): void {
     if (pi instanceof AndOptions) {
       this.checkInputLength(input, pi.options.length);
@@ -841,8 +833,17 @@ export class Player implements ISerializable<SerializedPlayer> {
       this.runInputCb(pi.cb());
     } else if (pi instanceof SelectHowToPayForProjectCard) {
       this.checkInputLength(input, 1, 2);
-      const foundCard: IProjectCard = this.getCard(pi.cards, input[0][0]);
+      const cardName = input[0][0];
+      const _data = PlayerInput.getCard(pi.cards, cardName);
+      const foundCard: IProjectCard = _data.card;
       const howToPay: HowToPay = this.parseHowToPayJSON(input[0][1]);
+      const reserveUnits = pi.reserveUnits[_data.idx];
+      if (reserveUnits.steel + howToPay.steel > this.steel) {
+        throw new Error(`${reserveUnits.steel} units of steel must be reserved for ${cardName}`);
+      }
+      if (reserveUnits.titanium + howToPay.titanium > this.titanium) {
+        throw new Error(`${reserveUnits.titanium} units of titanium must be reserved for ${cardName}`);
+      }
       this.runInputCb(pi.cb(foundCard, howToPay));
     } else if (pi instanceof SelectCard) {
       this.checkInputLength(input, 1);
@@ -854,7 +855,7 @@ export class Player implements ISerializable<SerializedPlayer> {
       }
       const mappedCards: Array<ICard> = [];
       for (const cardName of input[0]) {
-        mappedCards.push(this.getCard(pi.cards, cardName));
+        mappedCards.push(PlayerInput.getCard(pi.cards, cardName).card);
         if (pi.enabled?.[pi.cards.findIndex((card) => card.name === cardName)] === false) {
           throw new Error('Selected unavailable card');
         }
--- a/src/components/PaymentWidgetMixin.ts
+++ b/src/components/PaymentWidgetMixin.ts
@@ -44,7 +44,7 @@ export const PaymentWidgetMixin = {
     },
     addValue: function(target: string, to: number, max?: number): void {
       const currentValue: number = (this as any)[target];
-      let maxValue: number = max !== undefined ? max : (this as any).player[target];
+      let maxValue: number = max ?? (this as any).player[target];
 
       if (target === 'megaCredits') {
         maxValue = this.getMegaCreditsMax();
--- a/src/components/PaymentWidgetMixin.ts
+++ b/src/components/PaymentWidgetMixin.ts
@@ -77,7 +77,7 @@ export const PaymentWidgetMixin = {
     setMaxValue: function(target: string, max?: number): void {
       let currentValue: number = (this as any)[target];
       const cardCost: number = (this as any).$data.cost;
-      let amountHave: number = max !== undefined ? max : (this as any).player[target];
+      let amountHave: number = max ?? (this as any).player[target];
 
       let amountNeed: number = cardCost;
       if (['titanium', 'steel', 'microbes', 'floaters'].includes(target)) {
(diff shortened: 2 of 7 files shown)


src/Player.ts
```diff
@@ -756,14 +756,6 @@ export class Player implements ISerializable<SerializedPlayer> {
     return false;
   }
 
-  public getCard(cards: Array<IProjectCard>, cardName: string): IProjectCard {
```
kberg, line 759: I moved this over to PlayerInput, and also made it generic.

src/Player.ts
```diff
@@ -841,8 +833,17 @@ export class Player implements ISerializable<SerializedPlayer> {
       this.runInputCb(pi.cb());
     } else if (pi instanceof SelectHowToPayForProjectCard) {
       this.checkInputLength(input, 1, 2);
-      const foundCard: IProjectCard = this.getCard(pi.cards, input[0][0]);
+      const cardName = input[0][0];
```
kberg, line 836: This validates that a player can't overspend steel or titanium.

src/components/PaymentWidgetMixin.ts
```diff
@@ -74,10 +74,10 @@ export const PaymentWidgetMixin = {
 
       (this as any)['megaCredits'] = Math.max(0, Math.min(this.getMegaCreditsMax(), remainingMC));
     },
-    setMaxValue: function(target: string): void {
+    setMaxValue: function(target: string, max?: number): void {
```
kberg, line 77: This prevents clicking Max from solely relying on the player's steel units, but the maximum allowed by the caller (which will be this.availableSteel or this.availableTitanium.)

src/components/SelectHowToPayForProjectCard.ts
```diff
@@ -101,7 +100,7 @@ export const SelectHowToPayForProjectCard = Vue.component('select-how-to-pay-for
   },
   methods: {
     getCard: function() {
-      const card = this.player.cardsInHand.concat(this.player.selfReplicatingRobotsCards).find((c) => c.name === this.cardName);
+      const card = this.cards.find((c) => c.name === this.cardName); // this.player.cardsInHand.concat(this.player.selfReplicatingRobotsCards).find((c) => c.name === this.cardName);
```
kberg, line 103: I forgot to remove the original code. But honestly, I think this is correct, and the thing I would like others to confirm.

It's 2:30AM here so I won't do more manual testing tonight. More tomorrow.

src/inputs/SelectHowToPayForProjectCard.ts
```diff
@@ -2,19 +2,19 @@ import {PlayerInput} from '../PlayerInput';
...
 
+// TODO(kberg): Rename to SelectProjectCardToPlay
 export class SelectHowToPayForProjectCard implements PlayerInput {
   public inputType: PlayerInputTypes = PlayerInputTypes.SELECT_HOW_TO_PAY_FOR_PROJECT_CARD;
   public title = 'Play project card';
   public buttonLabel: string = 'Play card';
   public microbes: number;
   public floaters: number;
   public canUseHeat: boolean;
-  public reserveUnitsMap: Map<CardName, Units>;
```
kberg, line 17: I switched this from a Map to an Array solely because it seemed easier to think through when the indexes between card name, and this object lined up (as well as the data in the server model.)

tests/components/SelectHowToPayForProjectCard.spec.ts
```diff
@@ -254,14 +256,61 @@ describe('SelectHowToPayForProjectCard', function() {
...
-      cardsInHand: [{name: cardName, calculatedCost: cardCost, reserveUnits: Units.of({})}],
+    cardName: CardName,
+    cardCost: number,
+    playerFields: Partial<PlayerModel>,
+    playerInputFields: Partial<PlayerInputModel>,
+    reserveUnits: Units = Units.EMPTY) {
+    const player: Partial<PlayerModel> = Object.assign({
+      cards: [{name: cardName, calculatedCost: cardCost}],
       id: 'foo',
-      selfReplicatingRobotCards: [],
```
kberg, line 262: I removed this intentionally.

kberg:
I've manually tested this with a non-Moon game, using space cards and titanium and things worked as expected. I think this is good.

src/components/PaymentWidgetMixin.ts
```diff
@@ -42,9 +42,9 @@ export const PaymentWidgetMixin = {
 
       this.setRemainingMCValue();
     },
-    addValue: function(target: string, to: number): void {
+    addValue: function(target: string, to: number, max?: number): void {
       const currentValue: number = (this as any)[target];
-      let maxValue: number = (this as any).player[target];
+      let maxValue: number = max !== undefined ? max : (this as any).player[target];
```
bafolts, line 47: This can make use of my new favorite operator the null coalescing operator. This then becomes:

```
let maxValue: number = max ?? (this as any).player[target];
```

src/components/SelectHowToPayForProjectCard.ts
bafolts: It is 7:36PM here but I go to bed early! This all looks right to me but I want to double check it tomorrow in the light of day.

src/components/PaymentWidgetMixin.ts
kberg: Done!

src/Player.ts
```diff
@@ -841,8 +833,17 @@ export class Player implements ISerializable<SerializedPlayer> {
...
+      const _data = PlayerInput.getCard(pi.cards, cardName);
+      const foundCard: IProjectCard = _data.card;
       const howToPay: HowToPay = this.parseHowToPayJSON(input[0][1]);
+      const reserveUnits = pi.reserveUnits[_data.idx];
```
bafolts, line 840: During the transition games that are in progress are going to get an error with this rename and have to refresh the page. We could consider including that in our generic "error with player input" error messages. I don't think it is worth it to work around this. Something to keep in mind if once this is merged people claim they had issues paying for cards that appeared to go away when they refreshed the page.

src/components/PaymentWidgetMixin.ts
```diff
@@ -74,10 +74,10 @@ export const PaymentWidgetMixin = {
 
       (this as any)['megaCredits'] = Math.max(0, Math.min(this.getMegaCreditsMax(), remainingMC));
     },
-    setMaxValue: function(target: string): void {
+    setMaxValue: function(target: string, max?: number): void {
       let currentValue: number = (this as any)[target];
       const cardCost: number = (this as any).$data.cost;
-      let amountHave: number = (this as any).player[target];
+      let amountHave: number = max !== undefined ? max : (this as any).player[target];
```
bafolts, line 80: This again could be `max ?? (this as any).player[target]`. I need to see if there is a new linting rule to use this operator one way or another as it would be nice to enforce it's usage versus the ternary. Not a big deal either way.

src/components/SelectHowToPayForProjectCard.ts
```diff
@@ -31,7 +31,6 @@ import {PlayerModel} from '../models/PlayerModel';
 import {PreferencesManager} from './PreferencesManager';
 import {TranslateMixin} from './TranslateMixin';
 import {CardName} from '../CardName';
-// import {Units} from '../Units';
```
bafolts, line 34: Remove before flight.

src/components/SelectHowToPayForProjectCard.ts
bafolts: I also don't see any problem with this. This is far simpler so I am all for changing to this. Can you remove the commented out line before merging?

src/inputs/SelectHowToPayForProjectCard.ts
bafolts: This is good change. I think `Array` makes more sense as well because the order is important as it matches the order in the `cards` property.

bafolts (approved):

src/Player.ts
kberg: Agreed and assumed as much going into it.

src/components/PaymentWidgetMixin.ts
kberg: Thanks.

src/components/SelectHowToPayForProjectCard.ts
kberg: Ah right.

PR #3039, opened 2021-03-27
Clear out the server/ directory, it's not doing much.

kberg:
After this I'll experiment with moving all code into either client/ server/ or common/ and see if I can make some sense of the code.


bafolts (approved):
This is a great naming change.

PR #3041, opened 2021-03-27
add ETag for waiting for route

bafolts:
First usage of the `ETag` I think we can leverage for most routes. This implements for the waiting for route. During multi player games this should be hit often. This saves 100 bytes or so from the normal response. Over time this will add up to saved bandwidth!

image


kberg (approved):
Nice!

PR #3043, opened 2021-03-28
Fix the problem with selectHowToPay

### LINKED ISSUE #3042: Critical Bug: SelectHowToPay
Bug: Occasionally the app has a wrong default value that cannot be adjusted to spend properly except by overpaying.

This is a game breaking bug and should be fixed as soon as possible.

In the picture below, the default value should be 4 Ti (at 5 value) + 1 mc.

image

The app selects 3 mc. If you try to drop the mc down 1, it will complain.

vsrisuknimit:
This can close #3042.

When the app tries to reduce overspending, it should save steel only if it is actually spending steel. 

The bug was introduced by #3025. When the app sees that it is overspending by 2 mc and the player has steel, it tries to reduce the amount of spending steel by 1 even if that amount is already 0. As a result, the player is trying to pay with negative one steel forcing the player to have to overspend by 2 mc to cover that deficit.


vsrisuknimit:
See this

image

compared to this reported in #3042 

image

nwai90 (approved):

vsrisuknimit:
Added a test that should catch this problem.

image

bafolts:
Deploying this now. Thanks for quick fix.

PR #3045, opened 2021-03-28
send model when time to go

bafolts:
With the ETag I added to get waiting for route I completely broke when it is time to go sending "GO". This sends the "GO" when it is time to go. I added a regression test for this situation.

### DIFF
--- a/src/routes/ApiWaitingFor.ts
+++ b/src/routes/ApiWaitingFor.ts
@@ -11,8 +11,12 @@ export class ApiWaitingFor extends Handler {
     super();
   }
 
+  private timeToGo(player: Player): boolean {
+    return player.getWaitingFor() !== undefined || player.game.phase === Phase.END;
+  }
+
   private getWaitingForModel(player: Player, gameAge: number, undoCount: number): WaitingForModel {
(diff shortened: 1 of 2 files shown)


bafolts:
Already can see most requests are people waiting for their move. These have transitioned to 3XX which is good. This is less bandwidth the server is using. Should also be less CPU.

image

nwai90 (approved):

bafolts:
Deploying to terraforming-mars.herokuapp.com now

src/routes/ApiWaitingFor.ts
```diff
@@ -11,8 +11,12 @@ export class ApiWaitingFor extends Handler {
     super();
   }
 
+  private timeToGo(player: Player): boolean {
```
kberg, line 14: Why is it time to go when the phase is at the end, anyway?

bafolts: Prior to `gameAge` I think this is how we knew to update a player to take them to the game end screen. With `gameAge` it can most likely be removed assuming we logged during the final action. I am going to consider a different strategy for creating an hash for `Game`.

PR #3048, opened 2021-03-28
Do not use ETag header for waiting for route.

bafolts:
Will need a different strategy for generating the `ETag`. The strategy I was using relied on `gameAge` being incremented. Nothing must be logged during the initial drafting phase. Without `gameAge` being updated this hashing strategy doesn't work.

### DIFF
--- a/src/routes/ApiWaitingFor.ts
+++ b/src/routes/ApiWaitingFor.ts
@@ -34,14 +34,7 @@ export class ApiWaitingFor extends Handler {
         return;
       }
       try {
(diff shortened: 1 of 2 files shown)


vsrisuknimit (approved):
I don't understand the etag change but since this is the removal of the part that may cause bugs, I will approve so we can fix the bugs.

bafolts:
This will stay removed and I most likely wont revisit it. I didn't realize that `gameAge` wasn't always updated and was relying on that value to invalidate the cache. I also didn't realize how popular the draft variant was! Seems the draft variant doesn't increment `gameAge` which was being used to invalidate the cache.

PR #3050, opened 2021-03-28
Hidable animated side bar

vsrisuknimit:
This PR improves on #2969 .

image

It makes the color brighter but with gradient toward the bottom so the text on the top are not impossible to see. It also rotates the player cube slowly as movement should get attention of users.

It is possible to hide this animation from the Preferences panel.

Example

https://user-images.githubusercontent.com/14239220/112772426-b79e4480-8ffe-11eb-892c-cf326012b8e8.mov

### DIFF
--- a/compile-vue-templates.js
+++ b/compile-vue-templates.js
@@ -172,7 +172,7 @@ checkComponent(
     'hide_turnorder', 'hide_corporation_names', , 'hide_top_bar', 'small_cards', 'remove_background', 'magnify_cards',
     'magnify_card_descriptions', 'show_alerts', 'hide_ma_scores', 'hide_non_blue_cards', 'hide_log',
     'lang', 'langs', 'enable_sounds', 'smooth_scrolling', 'hide_tile_confirmation', 'show_card_number', 'show_discount_on_cards',
-    'learner_mode',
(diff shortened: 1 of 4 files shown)


vsrisuknimit:
@puuj

bafolts (changes requested):
The rotating of the cube seems great to me. Having a preference only for this animation seems like we are getting far too fine tuned. Every setting we add is another setting a user has to read through to find the setting they may want. Lets include the animation without the setting.

vsrisuknimit:
@bafolts The rotating cube and color gradient on the side bar will be on by default. The preference panel option is for hiding it. A few users prefer an option to have less distraction including @nwai90 .

vsrisuknimit:
I agree that we have too many options. We should remove many of the preference panel options (no one should want to hide the game log or player overview table for example). I will do that in another PR.

This is a bit more polarizing and is fine to have option to hide while being on by default.

puuj:
Sorry, didn't see this. I'm all for a brighter color or gradient. In my personal version I use the original, Mars-red color suggested. I find this very helpful when I'm playing a game while watching TV -- it's super-obvious when it's my turn.

bafolts (approved):
Approving with trade to remove some other preference in another PR.

PR #3063, opened 2021-03-29
Sabotage Global Event reduce only if has any

### LINKED ISSUE #2380: Bug: Inaccurate log for Turmoil
image

The player has no steel production. The log says steel production was decreased.

vsrisuknimit:
This PR can close #2380. It should try to reduce the steel/energy production if the player has any. This prevents inaccurate log report.

### DIFF
--- a/src/turmoil/globalEvents/Sabotage.ts
+++ b/src/turmoil/globalEvents/Sabotage.ts
@@ -12,8 +12,12 @@ export class Sabotage implements IGlobalEvent {
     public currentDelegate = PartyName.REDS;
     public resolve(game: Game, turmoil: Turmoil) {
       game.getPlayers().forEach((player) => {
-        player.addProduction(Resources.ENERGY, -1, game, undefined, true);
-        player.addProduction(Resources.STEEL, -1, game, undefined, true);
+        if (player.getProduction(Resources.ENERGY) >= 1) {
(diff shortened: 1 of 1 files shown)


kberg (commented):
While this works, I think we should really think about changing addProduction so it doesn't go negative there.

Also, I'd love for this to have a test - this is one of those changes where a test should hopefully be easy enough to write.

vsrisuknimit:
@kberg addProduction() should not allow a decrease of non-positive steel and energy production in general. It should strictly enforce that. 

I think it is fine for to do a check here as it is a special case. Turmoil does not enforce production decrease.

I have no idea how to test a log.

kberg:
I wouldn't test the log, I would test the production.

OH, you dont' want this to log? Oh, that's different. You should be able to do that by undoing what you've doine, and replacing  Just replace `game` with `undefined`. Awful, I know.

vsrisuknimit:
@kberg I would like this event to log, if and only if, the production was decreased by the event.

So replacing game with undefined would not solve it.

kberg:
OK, this makes sense now. I'll approve. Since all these things you're concerned about are not obvious from the code, please add this comment:

`// This conditional isn't to prevent negative production, but to prevent logging when the production diff is zero.`

kberg (approved):

kberg (approved):
Thanks

PR #3064, opened 2021-03-29
global event name in the log

### LINKED ISSUE #3057: Feature requesto: turmoil event in log
could there be any way we could get the turmoil event cards to show up in the log once they fall thru?

It would greatly benefit the log for seeing what has transpired.

Like we could click the event in the log and get to see that card just like how can click cards that have been played in the log atm.

vsrisuknimit:
This can close #3057, for now at least. 

We don't have a way to show the turmoil card once clicked yet. That is not how the turmoil event cards are rendered. 

For now, just showing the name of the global event that just happened should be helpful enough. Players can look them up later on Simon's database. But eventually we may want to consider rendering the global event as a unit of card.

### DIFF
--- a/src/turmoil/Turmoil.ts
+++ b/src/turmoil/Turmoil.ts
@@ -215,6 +215,7 @@ export class Turmoil implements ISerializable<SerializedTurmoil> {
 
       // 2 - Global Event
       if (this.currentGlobalEvent) {
+        game.log('Global event "' + this.currentGlobalEvent.name + '" has been resolved.');
         this.currentGlobalEvent.resolve(game, this);
       }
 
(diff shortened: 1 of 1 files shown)


vsrisuknimit:
image

src/turmoil/Turmoil.ts
```diff
@@ -215,6 +215,7 @@ export class Turmoil implements ISerializable<SerializedTurmoil> {
 
       // 2 - Global Event
       if (this.currentGlobalEvent) {
+        game.log('Global event "' + this.currentGlobalEvent.name + '" has been resolved.');
```
kberg, line 218: Can you add `// TODO: create a LogMessageDataType for Global Events, ref https://github.com/bafolts/terraforming-mars/issues/3057` above this, please?

kberg (approved):

src/turmoil/Turmoil.ts
vsrisuknimit: Added

PR #3071, opened 2021-03-30
Change logging order in compile-vue-templates.

kberg:
I wanted to give a good example here, but for some reason I don't know enough to force it to happen.

(I also renamed the primary method.)


compile-vue-templates.js
```diff
@@ -175,123 +174,125 @@ checkComponent(
...
-checkComponent(
+generateVueClass(
   'src/components/help/HelpStandardProjects',
   require('./build/src/components/help/HelpStandardProjects').HelpStandardProjects,
   [],
 );
 
-function checkComponent(name, component, dataProperties) {
+function generateVueClass(name, component, dataProperties) {
+  process.stdout.write(`generating the Vue typed template file for ${name} ... `);
```
alrusdi, line 294: Do we really need this output? It is kinda annoying to me :-)

kberg: I hear you. I wouldn't mind removing it, but without anny logging, when it fails I wouldn't know which class failed to generate.

And before this PR, you wouldn't know which class it was generating _when the error occurred._ The user would only know that the file _before it_ compiled successfully.

How annoying is it? If it's that annoying, I will follow up with a PR that keeps it silent UNLESS there are errors. How about that? Deal?

alrusdi: Yeah! It would be great!

kberg: Will do.

alrusdi (approved):

PR #3077, opened 2021-03-30
Fix promo card on end game page

### LINKED ISSUE #3027: Bug: promo cards show VP incorrectly on the final page & timer display bug
image

vsrisuknimit:
This PR closes #3027 .

Screen Shot 2021-03-30 at 1 49 31 PM

Screen Shot 2021-03-30 at 1 43 34 PM

image

It would be better if we rename the promo cards to `Great Dam (promo)` instead of `Great Dam:  promo`, but that will breaks existing games.

### DIFF
--- a/src/VictoryPointsBreakdown.ts
+++ b/src/VictoryPointsBreakdown.ts
@@ -11,7 +11,7 @@ export class VictoryPointsBreakdown {
     public moonRoads: number = 0;
     public victoryPoints = 0;
     public total = 0;
-    public detailsCards: Array<string> = [];
+    public detailsCards: Array<{cardName: string, victoryPoint: number}> = [];
     public detailsMilestones: Array<string> = [];
     public detailsAwards: Array<string> = [];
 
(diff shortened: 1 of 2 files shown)


bafolts (changes requested):
I think it is time to move past using type `string` for `detailsCards`.

https://github.com/bafolts/terraforming-mars/blob/main/src/VictoryPointsBreakdown.ts#L14

If we change this type to `Array<[CardName, number]>` we can then avoid needing to further add complexity with parsing data from the string.

Can you change the type of `detailsCards` to `Array<[CardName, number]>` (or some defined interface) and use that instead of trying to work with this string? At that point we can still parse out the colon and the number will be handled by it's type.

vsrisuknimit:
@bafolts, updated.

image

bafolts (approved):

bafolts:
If a player is playing a game without refreshing the page while this is rolled out they may run into a runtime exception thrown on the game end screen. Refreshing the page will solve this problem for them. I don't think it is worth putting in a fix for this. Be on lookout for this error soon if some unfortunate player playing a game as a server updates runs into this.

PR #3078, opened 2021-03-31
improve colony cube ui

vsrisuknimit:
This PR improves colony cube UI. The cube on the colonies are too small and sometimes hard to see that there is any cube there on some colonies. This UI improvement is inspired by the community server change, although the code is rewritten a bit.
image

### DIFF
--- a/src/components/Colony.ts
+++ b/src/components/Colony.ts
@@ -22,9 +22,6 @@ export const Colony = Vue.component('colony', {
     getColonyXPosition: (index: number): number => {
       return index * 56 + 16;
     },
-    getColonyCubeYOffset: function(): number {
-      return 7;
-    },
     getCubeYPosition: (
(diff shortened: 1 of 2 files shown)


vsrisuknimit:
Compared to the current cubes

src/components/Colony.ts
```diff
@@ -20,24 +20,37 @@ export const Colony = Vue.component('colony', {
       return colony.trackPosition * 56 + 27;
     },
     getColonyXPosition: (index: number): number => {
-      const offset: number = 5;
-      return index * 56 + 27 + offset;
+      return index * 56 + 16;
+    },
+    getColonyCubeYOffset: function(): number {
+      return 7;
```
alrusdi: Magic... Why don't we use css for positioning?

nwai90 (approved):

PR #3079, opened 2021-03-31
Fix typo & show how many cards being bought.

kberg:
image

image

Shows how many cards are being bought or selected. This is particular useful for Business Network:

image
image

### DIFF
--- a/src/components/SelectCard.ts
+++ b/src/components/SelectCard.ts
@@ -1,4 +1,3 @@
-
 import Vue from 'vue';
 import {Button} from '../components/common/Button';
 import {Message} from '../Message';
@@ -114,7 +113,7 @@ export const SelectCard = Vue.component('select-card', {
       return undefined;
     },
     buttonLabel: function(): string {
-      return this.playerinput.buttonLabel + ' ' + this.$data.cards.length;
+      return this.playerinput.buttonLabel + ' ' + this.cardsSelected();
     },
   },
   template: `<div class="wf-component wf-component--select-card">
--- a/src/components/SelectCard.ts
+++ b/src/components/SelectCard.ts
@@ -108,10 +108,13 @@ export const SelectCard = Vue.component('select-card', {
     getOwner: function(card: CardModel): OwnerModel | undefined {
       for (const player of this.player.players) {
         if (player.playedCards.find((c) => c.name === card.name)) {
-          return {name: player.name, color: player.color} as OwnerModel;
+          return {name: player.name, color: player.color};
         }
       }
       return undefined;
+    },
+    buttonLabel: function(): string {
+      return this.playerinput.buttonLabel + ' ' + this.$data.cards.length;
     },
   },
   template: `<div class="wf-component wf-component--select-card">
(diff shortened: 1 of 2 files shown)


vsrisuknimit:
For Business Network and Inventor Guild, I believe it is better to have two buttons: 'Buy' and 'Discard'. It would be ideal if the color for 'Discard' button is red-shade. Same for 'Skip this action'. But we can do that in another PR.

src/deferredActions/DrawCards.ts
```diff
@@ -97,7 +97,7 @@ export namespace DrawCards {
     const min = options.paying ? 0 : options.keepMax;
     const msg = options.paying ? (max === 0 ? 'You cannot afford any cards' : 'Select card(s) to buy') :
       `Select ${max} card(s) to keep`;
-    const button = max === 0 ? 'Oh' : (options.paying ? 'Buy' : 'Select');
+    const button = max === 0 ? 'Ok' : (options.paying ? 'Buy' : 'Select');
```
vsrisuknimit, line 100: 'Oh' might be accurate, ... if you cannot afford any card.

kberg: I just don't think that's the case because there's nothing that makes sense. In fact one of my friends saw that and thought "Uh, 'Oh'?"

vsrisuknimit: Perhaps change it to 'Skip this action' then.

kberg: I'm actually pretty confident this was a mistaken "OK." and I think it's an appropriate bit of text. Do you feel strongly?

vsrisuknimit: Not at all. I think you are right.

src/components/SelectCard.ts
```diff
@@ -108,11 +108,14 @@ export const SelectCard = Vue.component('select-card', {
...
       for (const player of this.player.players) {
         if (player.playedCards.find((c) => c.name === card.name)) {
-          return {name: player.name, color: player.color} as OwnerModel;
+          return {name: player.name, color: player.color};
         }
       }
       return undefined;
     },
+    buttonLabel: function(): string {
+      return this.playerinput.buttonLabel + ' ' + this.$data.cards.length;
```
bafolts, line 117: This should use the `cardsSelected` function instead of `this.$data.cards.length`. I think when vue-js uses the radio button that `this.$data.cards` will be the name of the card and not an array. This may be compiling because `string | Array` would have a `length` property. When `playerinput.minCardsToSelect === 1 && playerinput.maxCardsToSelect === 1` it probably isn't even worthwhile to display the number of cards selected as it will always be one. I think we should use `cardsSelected` here and not display when one card has to be selected.

bafolts (changes requested):

src/components/SelectCard.ts
kberg: > This should use the `cardsSelected` function instead of `this.$data.cards.length`. I think when vue-js uses the radio button that `this.$data.cards` will be the name of the card and not an array. This may be compiling because `string | Array` would have a `length` property.

That may be, but then why is it actually functioning correctly? Look at the screenshots.

> When `playerinput.minCardsToSelect === 1 && playerinput.maxCardsToSelect === 1` it probably isn't even worthwhile to display the number of cards selected as it will always be one. I think we should use `cardsSelected` here and not display the count when one card has to be selected.

It is. My friend lost a game yesterday because they didn't realize they hadn't selected the one Business Network card.

bafolts: The only example I can find is `Invention Contest` where `minCardsToSelect === 1` and `maxCardsToSelect === 1`. In this situation with `this.$data.cards.length` we end up displaying:

image

Using `this.cardsSelected()` resolves the issue.

kberg: Thanks. I fixed this, and tested it locally with the same card.

bafolts (approved):

Issue #2990, opened 2021-03-21
[Moon] Include moon tiles in Landlord.


kberg:
![image](https://user-images.githubusercontent.com/413481/111910508-0c0e5680-8a38-11eb-8fff-b0714175c508.png)

PR #3080, opened 2021-03-31
Fix bug #2990. Count Moon tiles for Landlord.

kberg:


src/awards/Landlord.ts
```diff
@@ -1,12 +1,22 @@
...
+import {MoonExpansion} from '../moon/MoonExpansion';
 
-// TODO(kberg): Landlord should take into account The Moon.
 export class Landlord implements IAward {
     public name: string = 'Landlord';
     public description: string = 'Owning the most tiles in play'
     public getScore(player: Player): number {
-      return player.game.board.spaces.filter((space) => space.tile !== undefined && space.tile.tileType !== TileType.OCEAN && space.player === player).length;
+      const marsSpaces = player.game.board.spaces.filter(
+        // Don't simplifiy this to "space.tile?.tileType !== TileType.OCEAN" because that will make
```
bafolts, line 11: This comment is helpful :+1:

bafolts: You only moved this but I don't think the check for `space.tile.tileType !== TileType.OCEAN` is necessary. When an ocean tile is placed the `player` property is set to `undefined` as oceans don't have an owner.

https://github.com/bafolts/terraforming-mars/blob/main/src/Game.ts#L1358

kberg: You're probably right, but [Chesterton's Fence](https://fs.blog/2020/03/chestertons-fence/) says it should stay.

PR #3083, opened 2021-03-31
clean up preferences panel

### LINKED ISSUE #2936: Feature request: Simplify Preference Panel
image

There are way too many options. Most of these should never be a toggle to begin with. Some options should be grouped together. Perhaps more than one column would be useful.

vsrisuknimit:
This PR can close #2936 .

Before

Screen Shot 2021-03-31 at 9 57 02 AM

After

Screen Shot 2021-03-31 at 11 05 01 AM

I would like to remove Hide Milestones and Awards and Hide Cards in Hand as well but I can imagine edge cases where users might want the options.


compile-vue-templates.js
```diff
@@ -167,11 +167,10 @@ generateVueClass(
   'src/components/Preferences',
   require('./build/src/components/Preferences').Preferences,
   [
-    'ui', 'hide_corporation', 'hide_hand', 'hide_cards', 'hide_awards_and_milestones', 'hide_tag_overview',
-    'hide_turnorder', 'hide_corporation_names', , 'hide_top_bar', 'small_cards', 'remove_background', 'magnify_cards',
-    'magnify_card_descriptions', 'show_alerts', 'hide_ma_scores', 'hide_non_blue_cards', 'hide_log',
-    'lang', 'langs', 'enable_sounds', 'smooth_scrolling', 'hide_tile_confirmation', 'show_card_number', 'show_discount_on_cards',
-    'learner_mode', 'hide_animated_sidebar',
+    'ui', 'hide_hand', 'hide_awards_and_milestones',
```
kberg, line 170: The list of changes here looks good, including changing one preference from `show` to `hide`.

src/components/card/CardCost.ts
```diff
@@ -19,8 +19,8 @@ export const CardCost = Vue.component('CardCost', {
       return classes.join(' ');
     },
     displayTwoCosts: function(): boolean {
-      const showDiscount = PreferencesManager.loadValue('show_discount_on_cards') === '1';
-      return this.newCost !== undefined && this.newCost !== this.amount && showDiscount;
+      const hideDiscount = PreferencesManager.loadBooleanValue('hide_discount_on_cards');
```
kberg, line 22: Does loadBooleanValue take a second optional default parameter? Huh, it doesn't. But in a follow-up that would be helpful.

kberg (approved):

src/components/card/CardCost.ts
vsrisuknimit: loadBooleanValue just load up from the memory. true is '1', false if '0' or undefined.

I don't think it needs a second default parameter.

kberg: I'll show you, in a follow-up. It's a standard pattern for fetching configuration parameters.

Issue #3081, opened 2021-03-31
Featuro request: LOG: seeing the resources gained from placement


aws0m:
Could we be able to see in the log what resources a player aquires when for example placing down an city or a ocean? The resources that are marked on the board.

Keep up the brilliant work!

kberg:
Yeah, I'll do that after #3090 goes in.

kberg:
This was fixed a while ago.

PR #3090, opened 2021-04-02
Space bonuses are granted en masse.

kberg:
image

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -65,6 +65,7 @@ import {TurmoilHandler} from './turmoil/TurmoilHandler';
 import {Random} from './Random';
 import {MilestoneAwardSelector} from './MilestoneAwardSelector';
 import {BoardType} from './boards/BoardType';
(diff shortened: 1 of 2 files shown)


vsrisuknimit:
while you are here, would it be difficult to add log for non-card resource? 

https://github.com/bafolts/terraforming-mars/issues/3081

kberg:
I will, in the next one.

src/Game.ts
```diff
@@ -1317,8 +1318,9 @@ export class Game implements ISerializable<SerializedGame> {
     // Part 5. Collect the bonuses
     if (this.phase !== Phase.SOLAR) {
       if (!coveringExistingTile) {
-        space.bonus.forEach((spaceBonus) => {
-          this.grantSpaceBonus(player, spaceBonus);
+        const bonuses = new Multiset(space.bonus);
```
bafolts, line 1321: This is great change and usage of `Multiset`.

bafolts (approved):

PR #3091, opened 2021-04-02
Introduce the notion of a private message to a player.

kberg:
This is only going to be used for helping players see what cards they were dealt, but as a demo, it's giving an introductory "good luck" message.

This seems to work between server restarts.

image
image
image
image

Don't use this feature for cards yet, let's make sure these messages can't accidentally leak into other browsers. I'm fairly certain they can't but anything can happen. Like bad deserialization. (I manually tested deserialization btw)

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -369,6 +369,10 @@ export class Game implements ISerializable<SerializedGame> {
       game.log('The id of this game is ${0}', (b) => b.rawString(id));
     }
 
+    players.forEach((player) => {
+      game.log('Good luck ${0}!', (b) => b.player(player), {reservedFor: player});
+    });
+
     game.log('Generation ${0}', (b) => b.forNewGeneration().number(game.generation));
 
     // Initial Draft
@@ -1470,12 +1474,14 @@ export class Game implements ISerializable<SerializedGame> {
     return player.cardsInHand.filter((card) => card.cardType === cardType);
   }
 
-  public log(message: string, f?: (builder: LogBuilder) => void) {
+  public log(message: string, f?: (builder: LogBuilder) => void, options?: {reservedFor?: Player}) {
     const builder = new LogBuilder(message);
     if (f) {
       f(builder);
     }
-    this.gameLog.push(builder.logMessage());
+    const logMessage = builder.build();
+    logMessage.playerId = options?.reservedFor?.id;
+    this.gameLog.push(logMessage);
     this.gameAge++;
   }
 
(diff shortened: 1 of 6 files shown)


vsrisuknimit:
It would be nice if the private message have a different icon, so the users don't panic and call easily tell which messages are private and which are public. Perhaps something like this, 💬. https://emojipedia.org/speech-balloon/

kberg:
Nice idea. I'll follow up with that, keeping things simple.

>

src/routes/GameLogs.ts
```diff
@@ -26,25 +26,30 @@ export class GameLogs {
...
         return;
       }
       let logs: Array<LogMessage> | undefined;
 
+      const messagesForPlayer = ((message: LogMessage) => message.playerId === undefined || message.playerId === playerId);
+
       // for most recent generation pull last 50 log messages
       if (generation === null || Number(generation) === game.generation) {
-        logs = game.gameLog.slice(-50);
+        logs = game.gameLog.slice(-60)
```
bafolts, line 48: This will have to run `game.gameLog.filter(messagesForPlayer).slice(-50)` so that the messages are removed that are for another player.

bafolts:
This change and approach makes sense to me. There is a side effect that the `log` function of `Game` updates `gameAge`.

https://github.com/bafolts/terraforming-mars/blob/main/src/Game.ts#L1485

Will need to consider if incrementing `gameAge` for private messages will have unintended consequences.

src/routes/GameLogs.ts
kberg: I was hoping to avoid that, but yeah.

kberg:
As @vsrisuknimit suggested, I used the speech bubble unicode character, 
image

kberg:
Thinking through this, I think the right way to use this for now is opt-in -- create a game option in the UI "experimental" or something like that, and that can control whether the game even records private messages. It allows players to test and find issues.

src/Game.ts
```diff
@@ -369,6 +369,10 @@ export class Game implements ISerializable<SerializedGame> {
       game.log('The id of this game is ${0}', (b) => b.rawString(id));
     }
 
+    players.forEach((player) => {
+      game.log('Good luck ${0}!', (b) => b.player(player), {reservedFor: player});
```
bafolts, line 373: Was this only intended for demonstration purposes?

kberg: Yes, and, to keep it in for now.

bafolts (approved):

Issue #3088, opened 2021-04-01
[Moon] Lunar Mine Urbanization issues


kberg:
> "Lunar Mine Urbanization": Sorry this card still causing issues. It seems diificult to implement this special case. It still has no icongraphy on the mmon map. And the VP calculation is still wrong or not clear. In the screenshot you will see, that the calculation for the colony, mines and road is correctly done without considering the LMU. Which should at least give in this situation additional 3 VP or even 6VP. I will post screenshot of the sitution in the discord moon-bug-report channel.

kberg:
Closing this because it's generally fixed.

PR #3094, opened 2021-04-02
Moon tile testing rewrite & Fix Grand Luna Capital Group.

### LINKED ISSUE #3086: [Moon] Grand Luna Capital Group calculating wrong VP
From report:
> Grand Luna Capital: We now understood the VP calculation for our situation with three colonies in triangle shape. This is fine for us now, and the implementation is correct. But now we had again a game with this corporation and we didn't understood the VP calculation at the end of the game comparing the game result with the end game moon map situation (yellow player with Grand Luna corp). Could someone explain how these VPs are calculated? Even my complete play group didn't understand it. I will post screenshot of the sitution in the discord moon-bug-report channel.

kberg:
There's a couple of rewrites that make moon tile testing better in general, including dealing with bug #3088, but it fixes #3086

### DIFF
--- a/src/cards/moon/TheDarksideofTheMoonSyndicate.ts
+++ b/src/cards/moon/TheDarksideofTheMoonSyndicate.ts
@@ -7,7 +7,6 @@ import {CardRenderer} from '../render/CardRenderer';
 import {ResourceType} from '../../ResourceType';
 import {Player} from '../../Player';
 import {ISpace} from '../../boards/ISpace';
-import {TileType} from '../../TileType';
 import {MoonExpansion} from '../../moon/MoonExpansion';
 import {Multiset} from '../../utils/Multiset';
 import {Resources} from '../../Resources';
@@ -88,10 +87,7 @@ export class TheDarksideofTheMoonSyndicate implements CorporationCard {
       return undefined;
     }
     const game = activePlayer.game;
-    switch (space.tile.tileType) {
-    case TileType.MOON_COLONY:
-    case TileType.MOON_MINE:
-    case TileType.MOON_ROAD:
+    if (MoonExpansion.MOON_TILES.includes(space.tile.tileType)) {
       const costs = new Multiset<Player>();
       MoonExpansion.moonData(game).moon.getAdjacentSpaces(space).forEach((space) => {
         if (space.tile !== undefined && space.player !== undefined && space.player !== activePlayer) {
--- a/src/cards/moon/TheDarksideofTheMoonSyndicate.ts
+++ b/src/cards/moon/TheDarksideofTheMoonSyndicate.ts
@@ -87,7 +87,7 @@ export class TheDarksideofTheMoonSyndicate implements CorporationCard {
       return undefined;
     }
     const game = activePlayer.game;
-    if (MoonExpansion.MOON_TILES.includes(space.tile.tileType)) {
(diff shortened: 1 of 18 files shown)


src/cards/moon/TheDarksideofTheMoonSyndicate.ts
```diff
@@ -88,10 +87,7 @@ export class TheDarksideofTheMoonSyndicate implements CorporationCard {
       return undefined;
     }
     const game = activePlayer.game;
-    switch (space.tile.tileType) {
-    case TileType.MOON_COLONY:
-    case TileType.MOON_MINE:
-    case TileType.MOON_ROAD:
+    if (MoonExpansion.MOON_TILES.includes(space.tile.tileType)) {
```
bafolts, line 90: The `MOON_TILES` array contains `LUNA_` and `LUNAR_` tile types as well which is different than the logic that is here. Is that change intentional?

src/moon/MoonExpansion.ts
```diff
@@ -22,17 +22,17 @@ import {BoardType} from '../boards/BoardType';
...
-  TileType.MOON_COLONY,
-  TileType.MOON_ROAD,
-  TileType.LUNA_TRADE_STATION,
-  TileType.LUNA_MINING_HUB,
-  TileType.LUNA_TRAIN_STATION,
-  TileType.LUNAR_MINE_URBANIZATION,
-];
-
 export class MoonExpansion {
+  public static readonly MOON_TILES = [
```
bafolts, line 26: This could be a `Set`. Can change to it with ease.

```
public static readonly MOON_TILES = new Set([
```

src/cards/moon/TheDarksideofTheMoonSyndicate.ts
kberg: It is, and it is a specific bugfix

src/cards/moon/TheGrandLunaCapitalGroup.ts
```diff
@@ -51,20 +51,26 @@ export class TheGrandLunaCapitalGroup implements CorporationCard {
...
     cardOwner.setResource(Resources.MEGACREDITS, filtered.length * 2, cardOwner.game);
   }
 
   public getVictoryPoints(player: Player) {
     const moon = MoonExpansion.moonData(player.game).moon;
-    const colonyTiles = MoonExpansion.tiles(player.game, TileType.MOON_COLONY, true);
-    const tilesAdjacentToPlayer = colonyTiles.filter((tile) =>
-      moon.getAdjacentSpaces(tile).some((neighbor) => neighbor.player?.id === player.id));
+    const neighboringColonyTiles: Set<SpaceId> = new Set();
+    const colonyTiles = MoonExpansion.tiles(player.game, TileType.MOON_COLONY, {ownedBy: player});
```
bafolts, line 65: The `true` argument before this change represented `surfaceOnly`. Is this change intentional to change to `ownedBy`?

bafolts (approved):

src/cards/moon/TheGrandLunaCapitalGroup.ts
kberg: Yes. Good catch. It was unnecessary.

src/moon/MoonExpansion.ts
kberg: Thanks.

PR #3095, opened 2021-04-02
Fix lunar mine urbanization visualization issues.

kberg:


src/styles/moon.less
```diff
@@ -42,9 +42,9 @@
 	background-image: url(./assets/moon/luna-train-station.png);
 }
 
-// .board-space-tile--lunar-mine-urbanization {
-// 	background-image: url(./assets/moon/lunar-mine-urbanization.png);
-// }
+.board-space-tile--lunar-mine-urbanization {
+	background-image: url(./assets/moon/lunar-mine-urbanization.png);
```
bafolts, line 46: This image is 200kb and it seems we are scaling client side. I will upload smaller versions of these soon to limit bandwidth usage.

bafolts (approved):

PR #3096, opened 2021-04-03
Add dangerous button and fix optional button

### LINKED ISSUE #3010: FR: Rename buttons and how Business Network works
image

vsrisuknimit:
This PR can close #3010. It does two main things.
1) When you are allowed to select zero card, it offers two buttons one for buy and one for skipping like during the research phase. I remove the restriction so that it applies to Business Network and Inventor Guild.
2) It labels certain buttons RED as a warning to the users that this can action is dangerous such as 'Skip', 'Discard', and 'Sell'. Sometimes users do not know that they are discarding a card due to Pluto trade or a global event. A red button will pauses them for a sec to notice what action they are doing.

Problem: It reads the title directly from Button. If I do it only in Selectcard, it will not apply to Sell and some forms of Discard. Not sure where the translation of the button title happened.

Discarding a card from Mars U
image

Business Network Action
image

Selling a card
image

Pluto does not work due to the number label
image

Sell does not work in another language
image

### DIFF
--- a/src/components/SelectCard.ts
+++ b/src/components/SelectCard.ts
@@ -127,7 +127,7 @@ export const SelectCard = Vue.component('select-card', {
         </label>
         <div v-if="hasCardWarning()" class="card-warning">{{ $t(warning) }}</div>
         <div v-if="showsave === true" class="nofloat">
-            <Button :disabled="isSelectionOptional() && cardsSelected() === 0" type="submit" :onClick="saveData" :title="buttonLabel()"/>
+            <Button :disabled="isSelectionOptional() && cardsSelected() === 0" type="submit" :onClick="saveData" :danger="playerinput.buttonDanger" :title="buttonLabel()" />
             <Button :disabled="isSelectionOptional() && cardsSelected() > 0" v-if="isSelectionOptional()" :onClick="saveData" type="submit" :danger="true" :title="$t('Skip this action')" />
         </div>
     </div>`,
--- a/src/components/SelectCard.ts
+++ b/src/components/SelectCard.ts
@@ -42,7 +42,6 @@ export const SelectCard = Vue.component('select-card', {
   },
   data: function() {
     return {
-      // Preselect the card if there is only one card as an option
       cards: [],
       warning: undefined,
     } as SelectCardModel;
--- a/src/components/SelectCard.ts
+++ b/src/components/SelectCard.ts
@@ -42,6 +42,7 @@ export const SelectCard = Vue.component('select-card', {
   },
   data: function() {
     return {
+      // Preselect the card if there is only one card as an option
       cards: [],
       warning: undefined,
     } as SelectCardModel;
@@ -87,10 +88,8 @@ export const SelectCard = Vue.component('select-card', {
       }
       return false;
     },
-    isOptionalToManyCards: function(): boolean {
-      return this.playerinput.maxCardsToSelect !== undefined &&
-             this.playerinput.maxCardsToSelect > 1 &&
-             this.playerinput.minCardsToSelect === 0;
+    isSelectionOptional: function(): boolean {
+      return this.playerinput.minCardsToSelect === 0;
     },
     getData: function(): Array<CardName> {
       return Array.isArray(this.$data.cards) ? this.$data.cards.map((card) => card.name) : [this.$data.cards.name];
@@ -128,8 +127,8 @@ export const SelectCard = Vue.component('select-card', {
         </label>
         <div v-if="hasCardWarning()" class="card-warning">{{ $t(warning) }}</div>
         <div v-if="showsave === true" class="nofloat">
-            <Button :disabled="isOptionalToManyCards() && cardsSelected() === 0" type="submit" :onClick="saveData" :title="buttonLabel()" />
-            <Button :disabled="isOptionalToManyCards() && cardsSelected() > 0" v-if="isOptionalToManyCards()" :onClick="saveData" type="submit" :title="$t('Skip this action')" />
+            <Button :disabled="isSelectionOptional() && cardsSelected() === 0" type="submit" :onClick="saveData" :title="buttonLabel()"/>
+            <Button :disabled="isSelectionOptional() && cardsSelected() > 0" v-if="isSelectionOptional()" :onClick="saveData" type="submit" :danger="true" :title="$t('Skip this action')" />
         </div>
     </div>`,
 });
--- a/src/components/common/Button.ts
+++ b/src/components/common/Button.ts
@@ -1,4 +1,5 @@
 import Vue from 'vue';
+import {PreferencesManager} from '../PreferencesManager';
 
 export const Button = Vue.component('Button', {
   props: {
@@ -74,8 +75,10 @@ export const Button = Vue.component('Button', {
         classes.push('btn-submit');
       }
 
-      if (this.danger || this.title === 'Discard' || this.title === 'Sell') {
-        classes.push('btn-dangerous');
+      if (this.danger || this.title.includes('Discard' || this.title === 'Sell')) {
+        if (PreferencesManager.loadBooleanValue('color_label_buttons')) {
+          classes.push('btn-dangerous');
+        }
       }
 
       // align
--- a/src/components/common/Button.ts
+++ b/src/components/common/Button.ts
@@ -75,8 +75,8 @@ export const Button = Vue.component('Button', {
         classes.push('btn-submit');
       }
 
-      if (this.danger || this.title.includes('Discard' || this.title === 'Sell')) {
-        if (PreferencesManager.loadBooleanValue('color_label_buttons')) {
+      if (this.danger || this.title === 'Discard' || this.title === 'Sell') {
+        if (!PreferencesManager.loadBooleanValue('hide_color_code_buttons')) {
           classes.push('btn-dangerous');
         }
       }
--- a/src/components/common/Button.ts
+++ b/src/components/common/Button.ts
@@ -40,6 +40,10 @@ export const Button = Vue.component('Button', {
           'plus',
         ].includes(item),
     },
+    danger: {
+      type: Boolean,
+      default: false,
+    },
     onClick: {
       type: Function,
       default: () => {},
@@ -68,6 +72,10 @@ export const Button = Vue.component('Button', {
         classes.push('btn-action');
       } else if (this.type === 'submit') {
         classes.push('btn-submit');
+      }
+
+      if (this.danger || this.title === 'Discard' || this.title === 'Sell') {
+        classes.push('btn-dangerous');
       }
 
       // align
--- a/src/components/common/Button.ts
+++ b/src/components/common/Button.ts
@@ -75,7 +75,7 @@ export const Button = Vue.component('Button', {
         classes.push('btn-submit');
       }
 
-      if (this.danger || this.title === 'Discard' || this.title === 'Sell') {
+      if (this.danger) {
         if (!PreferencesManager.loadBooleanValue('hide_color_code_buttons')) {
           classes.push('btn-dangerous');
         }
--- a/src/models/ServerModel.ts
+++ b/src/models/ServerModel.ts
@@ -223,7 +223,7 @@ function getWaitingFor(
   const playerInputModel: PlayerInputModel = {
     title: waitingFor.title,
     buttonLabel: waitingFor.buttonLabel,
-    buttonDanger: waitingFor.buttonLabel === 'Sell' || waitingFor.buttonLabel === 'Discard' ? true : undefined,
+    buttonDanger: waitingFor.buttonDanger,
     inputType: waitingFor.inputType,
     amount: undefined,
     options: undefined,
--- a/src/models/ServerModel.ts
+++ b/src/models/ServerModel.ts
@@ -223,6 +223,7 @@ function getWaitingFor(
   const playerInputModel: PlayerInputModel = {
     title: waitingFor.title,
     buttonLabel: waitingFor.buttonLabel,
+    buttonDanger: waitingFor.buttonLabel === 'Sell' || waitingFor.buttonLabel === 'Discard' ? true : undefined,
     inputType: waitingFor.inputType,
     amount: undefined,
     options: undefined,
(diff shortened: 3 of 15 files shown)


vsrisuknimit:
Changed the default to be showing color, with an option to hide the color code in preferences.

Most users seem to like this change. A few think they are distracting. (Same folks who don't like the sidebar notification.)

image

src/components/SelectCard.ts
```diff
@@ -87,10 +88,8 @@ export const SelectCard = Vue.component('select-card', {
       }
       return false;
     },
-    isOptionalToManyCards: function(): boolean {
-      return this.playerinput.maxCardsToSelect !== undefined &&
-             this.playerinput.maxCardsToSelect > 1 &&
-             this.playerinput.minCardsToSelect === 0;
+    isSelectionOptional: function(): boolean {
```
kberg, line 91: This is a good renaming. "isOptionalToManyCards" made no sense to me earlier this week. "isSelectionOptional" makes more sense.

kberg (changes requested):
So, I have a couple of general concerns. First, the notion of a change being 'dangerous' or not has to be removed. Actions are dangerous when they're particularly destructive, and in this game, that's not really an option.

Second, please don't make this an option. It's not going to provide any real value.

I'd like to start removing the notion that prefences should be named as 'hiding' things. That's not a good pattern. It makes reading code confusing "if !hide" is harder to read and mainain than "if show". This is why I recommended having a loadBoolean with a default value. But that can be dealt with later on.

I don't have a good suggestion for a name yet, but once you remove the preferences and just color the button this PR will be much simpler.

vsrisuknimit:
> So, I have a couple of general concerns. First, the notion of a change being 'dangerous' or not has to be removed. Actions are dangerous when they're particularly destructive, and in this game, that's not really an option.

1) I don't get it. What change do you want?

> Second, please don't make this an option. It's not going to provide any real value.

2) Don't make this an option. So close this PR? Or do you mean make this change for everyone? No option to turn it off. As mentioned, some minority do feel strongly that this is a distraction. It is the same reason we offer option to disable sound, remove background image, disable in game alert, etc.

> I'd like to start removing the notion that prefences should be named as 'hiding' things. That's not a good pattern. It makes reading code confusing "if !hide" is harder to read and mainain than "if show". This is why I recommended having a loadBoolean with a default value. But that can be dealt with later on.

I agree it is harder to read.

Load boolean with default values is fine. I wrote it but then I realize I have no way to make the preference panel toggled by default. Like if I open the app in incognito, some toggles should be on or off by default. I cannot do that right now, and that will just confuse the users.

image

image

src/components/common/Button.ts
```diff
@@ -70,6 +75,12 @@ export const Button = Vue.component('Button', {
         classes.push('btn-submit');
       }
 
+      if (this.danger || this.title === 'Discard' || this.title === 'Sell') {
```
bafolts, line 78: Could we pass this through `SelectCard` in some way? When `minCardsToSelect` is `0` we could provide a string to show as the button label when there are no cards selected. For example on `Mars University` we would create the `SelectCard` instance and could then set a property named `noSelectionLabel`. This would be the label when there were no selection. This could avoid having to check the `title` and could help with translation efforts. I think it could also avoid introducing the colored button an user preference.

I agree with @kberg that there is nothing _dangerous_ about discarding no project cards on `Mars University`. This action does not put the user in danger. I would prefer to solve this problem by introducing a `noSelectionButtonLabel` on `SelectCard` if possible.

vsrisuknimit: @bafolts  buying no card with Inventor Guild is 'dangerous'. The current implementation makes it really easy to buy no card when you are intended to buy one. From Mars University, what is dangerous is not 'discarding no project card' but the fact that you are 'discarding a card'. After users play a science card, on their second action, they may not realize that their next action is to discard something, they may select a card they want to play instead. I have seen this many times. That is dangerous. Same for Pluto.

I would like to do it in selectCard as well. But the Save button for 'SELL' and 'DISCARD' is outside select card. Please advise on how to proceed.

vsrisuknimit: image

kberg:
I've had an allergy attack all weekend and have been able to do nothing but
binge TV. I'll look again once I feel better.

>

kberg:
Sorry I've been away, I had an awful hay fever attack for most of the weekend.

> > So, I have a couple of general concerns. First, the notion of a change being 'dangerous' or not has to be removed. Actions are dangerous when they're particularly destructive, and in this game, that's not really an option.
> 
> 1. I don't get it. What change do you want?

I think you did it. Thanks.

> > Second, please don't make this an option. It's not going to provide any real value.
> 
> 1. Do you mean make this change for everyone? No option to turn it off. As mentioned, some minority do feel strongly that this is a distraction. It is the same reason we offer option to disable sound, remove background image, disable in game alert, etc.

I've seen a lot of UX changes, and a lot of people having opinions about UI changes. I've *been* that person for sixteen years at Google complaining about how the old thing was better, and the new UI isn't. And every time I got used to it. Every single time. It's nice that you're trying to make everybody happy, but then you wind up with the overloaded set of preferences. The majority of people who call it distracting -- I bet they'll soon forget about it.

If you still insist a preference is better, I get it, but my money is on that if there wasn't a preference, everyone would be just fine.

> > I'd like to start removing the notion that prefences should be named as 'hiding' things. That's not a good pattern. It makes reading code confusing "if !hide" is harder to read and mainain than "if show". This is why I recommended having a loadBoolean with a default value. But that can be dealt with later on.
> 
> I agree it is harder to read.
> 
> Load boolean with default values is fine. I wrote it but then I realize I have no way to make the preference panel toggled by default. Like if I open the app in incognito, some toggles should be on or off by default. I cannot do that right now, and that will just confuse the users.

That's fair - the framework should support that.
> 
> image
> 
> image

src/components/common/Button.ts
vsrisuknimit: Something to with `showsave`?

bafolts: This happens [here](https://github.com/bafolts/terraforming-mars/blob/main/src/components/OrOptions.ts#L145). The selection option's buttons text is passed to a different `Button` component. If the red button is always going to be displayed we would have to pass the property from the option here. This will avoid needing to check for the title in `Button`.

bafolts: I didn't realize "danger" was a name used for buttons which are red like this. The color of this button and name property makes sense to me.

https://getbootstrap.com/docs/4.0/components/buttons/

Checking the title for "Sell" or "Discard" is the part I would like to replace with a property on `PlayerInput`. Similar to `buttonLabel` it would be a property named `buttonDanger` to indicate that the button should use the danger button.

vsrisuknimit: image

image

@bafolts I moved the title check to playerInputModel that is done on ServerModel. But I still have the problem of these buttons not turning red.

src/models/ServerModel.ts
```diff
@@ -223,6 +223,7 @@ function getWaitingFor(
   const playerInputModel: PlayerInputModel = {
     title: waitingFor.title,
     buttonLabel: waitingFor.buttonLabel,
+    buttonDanger: waitingFor.buttonLabel === 'Sell' || waitingFor.buttonLabel === 'Discard' ? true : undefined,
```
bafolts, line 226: Still don't think we should check the `buttonLabel`. We can set `buttonDanger` on the `PlayerInput` directly. Can you use an approach like below?

```diff
diff --git a/src/PlayerInput.ts b/src/PlayerInput.ts
index 64dcdf19c..82f99b2c0 100644
--- a/src/PlayerInput.ts
+++ b/src/PlayerInput.ts
@@ -4,6 +4,7 @@ import {PlayerInputTypes} from './PlayerInputTypes';
 
 export interface PlayerInput {
     inputType: PlayerInputTypes;
+    buttonDanger?: boolean;
     buttonLabel: string;
     options?: Array<PlayerInput>;
     title: string | Message;
diff --git a/src/cards/base/standardProjects/SellPatentsStandardProject.ts b/src/cards/base/standardProjects/SellPatentsStandardProject.ts
index 7a157eead..34449e227 100644
--- a/src/cards/base/standardProjects/SellPatentsStandardProject.ts
+++ b/src/cards/base/standardProjects/SellPatentsStandardProject.ts
@@ -30,7 +30,7 @@ export class SellPatentsStandardProject extends StandardProjectCard {
   }
 
   public action(player: Player): SelectCard<IProjectCard> {
-    return new SelectCard(
+    const select = new SelectCard(
       'Sell patents',
       'Sell',
       player.cardsInHand,
@@ -50,5 +50,7 @@ export class SellPatentsStandardProject extends StandardProjectCard {
         return undefined;
       }, player.cardsInHand.length,
     );
+    select.buttonDanger = true;
+    return select;
   }
 }
diff --git a/src/deferredActions/DiscardCards.ts b/src/deferredActions/DiscardCards.ts
index 317d0c485..564cb8e28 100644
--- a/src/deferredActions/DiscardCards.ts
+++ b/src/deferredActions/DiscardCards.ts
@@ -17,7 +17,7 @@ export class DiscardCards implements DeferredAction {
       cards.forEach((card) => this.player.game.dealer.discard(card));
       return undefined;
     }
-    return new SelectCard(
+    const select = new SelectCard(
       this.title,
       'Discard',
       this.player.cardsInHand,
@@ -31,5 +31,7 @@ export class DiscardCards implements DeferredAction {
       this.count,
       this.count,
     );
+    select.buttonDanger = true;
+    return select;
   }
 }
diff --git a/src/inputs/SelectCard.ts b/src/inputs/SelectCard.ts
index 9ff029e8f..838b331b5 100644
--- a/src/inputs/SelectCard.ts
+++ b/src/inputs/SelectCard.ts
@@ -5,7 +5,7 @@ import {PlayerInputTypes} from '../PlayerInputTypes';
 
 export class SelectCard<T> implements PlayerInput {
     public inputType: PlayerInputTypes = PlayerInputTypes.SELECT_CARD;
-
+    public buttonDanger?: boolean;
     constructor(
         public title: string | Message,
         public buttonLabel: string = 'Save',
diff --git a/src/models/ServerModel.ts b/src/models/ServerModel.ts
index cb01a2c9a..01452af51 100644
--- a/src/models/ServerModel.ts
+++ b/src/models/ServerModel.ts
@@ -223,7 +223,7 @@ function getWaitingFor(
   const playerInputModel: PlayerInputModel = {
     title: waitingFor.title,
     buttonLabel: waitingFor.buttonLabel,
-    buttonDanger: waitingFor.buttonLabel === 'Sell' || waitingFor.buttonLabel === 'Discard' ? true : undefined,
+    buttonDanger: waitingFor.buttonDanger,
     inputType: waitingFor.inputType,
     amount: undefined,
     options: undefined,
```

vsrisuknimit: @bafolts this definitely looks cleaner. But something still does not work. I suspect something here:

image

image

image

vsrisuknimit:
This should be working now. It still does not work for SELL but that is another PR. Sell does not use a button with SelectCard but under orOption instead.

bafolts (approved):

vsrisuknimit:
@bafolts I updated two things. (1) remove unused comment, (2) pass buttonDanger to orOption buttons. This allows SELL to be color coded too.

vsrisuknimit:
@kberg I insist on the option for color coding.

kberg:
> @kberg I insist on the option for color coding.

Well, I hear you. I disagree.

vsrisuknimit:
@kberg It is problematic when you think you know what is best for the users, without talking to them to see why they would want this option, and just hold a PR hostage.

kberg:
I'm not dipping into another acrimonious debate with you.

alrusdi:
> @kberg It is problematic when you think you know what is best for the users, without talking to them to see why they would want this option, and just hold a PR hostage.

If most of the players wants color coding why it is important to have an option to switch it off?

vsrisuknimit:
> If most of the players wants color coding (great feature btw!) then why it is important to have an option to switch it off?

Because the color coding can be a distraction for some users (for example players with ADHD). These users prefer to have uniformed and muted color scheme. The red can bring too much attention that can be problematic.

https://uxpamagazine.org/designing_children_adhd/

vsrisuknimit:
> I'm not dipping into another acrimonious debate with you.

@kberg again I just want you to listen to people and see why they want things certain way instead of forming an opinion on your own.

bafolts:
@vsrisuknimit I don't think anyone is holding this PR "hostage". The word "hostage" implying there is hostility here. There is a disagreement from what you are trying to introduce with this change. As far as I can tell the major hold-up is the introduction of the _option_ to hide the color and not the color itself. I agree with not adding another option for the colors. 

From your referenced article:

```
Guideline 2: Provide a calm environment with soothing colors, free of decorations and distractions.
Again, although intended for designing a home or classroom environment, this suggestion could apply to a software environment. The advice on minimal distractions is in line with standard usability guidelines and seems suitable for all users since it helps avoid confusing displays and unnecessary attention-grabbing features. The advice on “calm” and “soothing colors” may be of note to graphical designers, particularly for the design of children’s software where bright colors are often used.
```

If the concern is that the color red selected for the danger button is not soothing would a different color of red work for you to avoid needing an option? My concern with adding the option would be that even if you were a player that didn't want the color of red selected for danger buttons you would have to know to look for the option to disable the colors and what that option meant. Every option we add comes along with additional help pages we are lacking and support. The recommended ages for the board game is also 12+ which would be teenagers and not children.

vsrisuknimit:
@bafolts 
- I already toned this red down as much as I can that still look aesthetically pleasing. I consulted with @ssimeonoff on the color.
- ADHD does not go away with age. And not the only reason for an option to mute the color coding.
- We have option to mute sound or hide tile confirmation. People know how to tune things.
- I removed like 10 unnecessary options from Preferences panel. I think more can be removed but I believe this option is important.

kberg:
I understand. Here's how I see it: there are two devs asking for changes, and of the four devs participating in this thread, none have marked an approval. If the devs are convinced by your points, that's great. Until then my review remains the only one. This is where https://github.com/bafolts/terraforming-mars/blob/main/CONTRIBUTING.md#if-your-change-is-declined comes in, although I'm not implying that you should fork the repo or that I want you to fork the repo.

vsrisuknimit:
Yes it was approved four days ago and we have been waiting.

Brian does not get the point of the color coding to begin with.

> I still am not sure the danger button color is going to help anything. https://github.com/bafolts/terraforming-mars/issues/3104#issuecomment-818349088

So one of you does not understand the point of this change while the other say this change has to be mandatory, removing the option to turn it off- the option which I believe will be beneficial to the mental state of a small group of people.

I guess we cannot progress.

kberg:
> I guess we cannot progress.

OK. If you have a change of heart please reopen the PR.

PR #3097, opened 2021-04-03
smaller images for moon corporation cards

bafolts:
Scales the images for the moon corporation cards on the server. This will use less bandwidth and not require the client to scale the image.


kberg:
I'm not sure. This loses fidelity and readability.

image

bafolts:
I can try this again. There should be no loss in quality assuming the algorithm used to scale the image is the same being used by the browser.

bafolts:
Proving difficult to find the scaling algorithm chromium is using. Will close until I can find a tool that uses the scaling algorithm that chromium is using.

PR #3101, opened 2021-04-05
calculate award properly

### LINKED ISSUE #3100: Feature request: Milestones and Awards predictor
For certain awards, the current number and how the real-time VP is calculated do not represent real players' standings. 

image

Thermalist: It should count heat + heat production + energy, until the end of the game where it only counts heat.

Industrialist: it should count steel + steel production + energy production, until the end of the game where it only counts steel + energy.

Miner: it should count steel + steel prod + titanium + titanium prod, until the end of the game where it only counts steel and titanium.

vsrisuknimit:
This PR can close #3100.

The awards that count resources should getScore that accounts for the final production as if the game is to end now.

image

See Thermalist is 7 for Red Player instead of 3.

### DIFF
--- a/src/awards/Industrialist.ts
+++ b/src/awards/Industrialist.ts
@@ -1,13 +1,12 @@
 import {IAward} from './IAward';
 import {Player} from '../Player';
-import {Phase} from '../Phase';
 import {Resources} from '../Resources';
 
 export class Industrialist implements IAward {
     public name: string = 'Industrialist';
     public description: string = 'Having most steel and energy resources'
     public getScore(player: Player): number {
-      if (player.game.phase === Phase.END || player.game.isDoneWithFinalProduction()) {
+      if (player.game.isDoneWithFinalProduction()) {
         return player.steel + player.energy;
       } else {
         return player.steel + player.getProduction(Resources.STEEL) + player.getProduction(Resources.ENERGY);
--- a/src/awards/Industrialist.ts
+++ b/src/awards/Industrialist.ts
@@ -1,10 +1,20 @@
 import {IAward} from './IAward';
 import {Player} from '../Player';
+import {Phase} from '../Phase';
+import {Resources} from '../Resources';
 
 export class Industrialist implements IAward {
     public name: string = 'Industrialist';
     public description: string = 'Having most steel and energy resources'
     public getScore(player: Player): number {
-      return player.steel + player.energy;
+      if (player.game.phase === Phase.END) {
+        return player.steel + player.energy;
+      } else if (player.game.gameIsOver() && player.game.phase === Phase.PRODUCTION) {
+        // This if check is for the final greenery placement step, we should calculate the score as if the game has end
+        // because the final production already happened.
+        return player.steel + player.energy;
+      } else {
+        return player.steel + player.getProduction(Resources.STEEL) + player.getProduction(Resources.ENERGY);
+      }
     }
 }
--- a/src/awards/Industrialist.ts
+++ b/src/awards/Industrialist.ts
@@ -4,7 +4,7 @@ import {Resources} from '../Resources';
 
 export class Industrialist implements IAward {
     public name: string = 'Industrialist';
-    public description: string = 'Having most steel and energy resources'
+    public description: string = 'Having most steel and energy resources (after final production round)'
     public getScore(player: Player): number {
       if (player.game.isDoneWithFinalProduction()) {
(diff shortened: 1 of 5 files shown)


kberg:
This feature seems useful for experienced players, but not for beginners. It also demonstrates a behavior that makes awards different from each other without a visual cue. If this showed both values, or gave very clear visual cue that these are special and different, I would approve such a change.

vsrisuknimit:
@kberg All awards are different from each other. Some count tags, some count tiles. These three awards are the only that count resources. As long as all three behave the same way that should be no problem.

- Showing two numbers are confusing and not compact.
- The current implementation on the other hand should confuse beginner players greatly. "The app says I am winning Thermalist. Why in the world did I lose it at the end of the game? Oh the app did not take final production into account." The current number are next to useless (for these three awards) and make the awards swing and live-vp calculation swing greatly for no reason.
- We can make it different colors, but there is no reason to do that. As I mentioned, all awards are different.

bafolts:
I think the number may be beneficial to beginners as well as experienced players. As a beginner it may help to show that production matters. A beginner who doesn't know that production is going to be added once mars is terraformed could come to learn they did not end up with the award.

src/awards/Industrialist.ts
```diff
@@ -1,10 +1,20 @@
...
+import {Phase} from '../Phase';
+import {Resources} from '../Resources';
 
 export class Industrialist implements IAward {
     public name: string = 'Industrialist';
     public description: string = 'Having most steel and energy resources'
     public getScore(player: Player): number {
-      return player.steel + player.energy;
+      if (player.game.phase === Phase.END) {
+        return player.steel + player.energy;
```
bafolts, line 11: There is code duplication among these three awards. Lets create a function in `src/awards/Awards.ts` for this logic. Can you add an `Awards.pastLastProductionPhase`?

```
function pastLastProductionPhase(game: Game) {
  return game.gameIsOver() && (game.phase === Phase.END || game.phase === Phase.PRODUCTION);
}
```

bafolts (changes requested):

kberg:
> I think the number may be beneficial to beginners as well as experienced players. As a beginner it may help to show that production matters. A beginner who doesn't know that production is going to be added once mars is terraformed could come to learn they did not end up with the award.

It _could_ be if it was clear what the number meant. But someone has to unwind it from first principles by identifying the three sources (in the worst case.) If this included an easy way to clarify what the value comes from, I'd be more comfortable with the change.

src/awards/Industrialist.ts
vsrisuknimit: Done. I moved it to Game instead.

src/awards/Industrialist.ts
```diff
@@ -1,10 +1,16 @@
...
 import {Player} from '../Player';
+import {Phase} from '../Phase';
+import {Resources} from '../Resources';
 
 export class Industrialist implements IAward {
     public name: string = 'Industrialist';
     public description: string = 'Having most steel and energy resources'
     public getScore(player: Player): number {
-      return player.steel + player.energy;
+      if (player.game.phase === Phase.END || player.game.isDoneWithFinalProduction()) {
```
bafolts, line 10: Sorry if my previous comment was confusing but I would include `player.game.phase === Phase.END` in the `isDoneWithFinalProduction` function. Whenever `player.game.phase === Phase.END` was are also done with the final production. This will make the function correct and cut down on the remaining duplication.

bafolts:
> > I think the number may be beneficial to beginners as well as experienced players. As a beginner it may help to show that production matters. A beginner who doesn't know that production is going to be added once mars is terraformed could come to learn they did not end up with the award.
> 
> It _could_ be if it was clear what the number meant. But someone has to unwind it from first principles by identifying the three sources (in the worst case.) If this included an easy way to clarify what the value comes from, I'd be more comfortable with the change.

Including how these values are calculated should help even with what we are doing today. Even today I don't think we show how these values are calculated.

src/awards/Industrialist.ts
vsrisuknimit: Done

bafolts (approved):

vsrisuknimit:
> It _could_ be if it was clear what the number meant. But someone has to unwind it from first principles by identifying the three sources (in the worst case.) If this included an easy way to clarify what the value comes from, I'd be more comfortable with the change.

Added `(after final production round)` in the award explanation.

PR #3103, opened 2021-04-06
Fixing Hellas Map ocean bonus

vsrisuknimit:
This should reduce the cost of the ocean bonus spot so that canAfford becomes true. From console.log that happens, so it seems the space got filtered out even before Hellas.filter somehow.

In theory this code should work. I am not sure what part I am missing.

### DIFF
--- a/src/boards/HellasBoard.ts
+++ b/src/boards/HellasBoard.ts
@@ -8,6 +8,10 @@ import {SpaceType} from '../SpaceType';
 import {BoardBuilder} from './BoardBuilder';
 import {SerializedBoard} from './SerializedBoard';
 import {Random} from '../Random';
+import {TileType} from '../TileType';
+import {PartyName} from '../turmoil/parties/PartyName';
+import {GREENS_RULING_POLICY_REBATE} from '../constants';
+import {CardName} from '../CardName';
 
 export class HellasBoard extends Board {
   public static newInstance(shuffle: boolean, rng: Random, includeVenus: boolean): HellasBoard {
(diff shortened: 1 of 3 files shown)


src/boards/HellasBoard.ts
```diff
@@ -51,24 +55,41 @@ export class HellasBoard extends Board {
...
+    }
+    if (tileType === TileType.LAVA_FLOWS) {
+      if (player.cardIsInEffect(CardName.MEDIA_GROUP)) cost -= 2;
+    }
+    if (tileType === TileType.DEIMOS_DOWN) {
+      if (player.cardIsInEffect(CardName.MEDIA_GROUP)) cost -= 2;
+      if (player.cardIsInEffect(CardName.OPTIMAL_AEROBRAKING)) cost -= 3;
+    }
+    console.log(`tile: ${tileType}, const: ${cost}, canAfford: ${player.canAfford(cost)}`);
+    return player.canAfford(cost) ? spaces : spaces.filter((space) => space.id !== SpaceName.HELLAS_OCEAN_TILE);
```
bafolts, line 76: This does seem like it should work. The `spaces` array should contain the results of `super.getAvailableSpacesForCity`. Your logging indicates we would return `spaces`. I would also log `spaces` to see what `spaces` were provided.

kberg: I would like to see if it were possible to not create another place, like RoboticWorkforce, where custom details are crammed into a corner of the app we wouldn't necessarily not know about. For example, this doesn't take Ares tiles into consideration, and my point is the right answer isn't just to work all that code into here.

PR #3106, opened 2021-04-06
Correct amount decreased

sauloachkar:
Some cards remove more resources than players have in board. Log was showing the full number, instead of the correct number removed.


src/deferredActions/CorrosiveRainDeferredAction.ts
```diff
@@ -19,14 +19,16 @@ export class CorrosiveRainDeferredAction implements DeferredAction {
             card.resourceCount !== undefined &&
             card.resourceCount >= 2);
 
+    const amountLost = Math.min(10, this.player.megaCredits);
```
bafolts, line 22: This `setResource` function _should_ be protecting against this.

https://github.com/bafolts/terraforming-mars/blob/main/src/Player.ts#L310

vsrisuknimit: @bafolts he is talking accurate logging.

bafolts:
Can you fix this from within the `Player.setResource` function? This will avoid having to fix it every place we call `setResource`.

https://github.com/bafolts/terraforming-mars/blob/main/src/Player.ts#L329

sauloachkar:
> Can you fix this from within the `Player.setResource` function? This will avoid having to fix it every place we call `setResource`.
> 
> https://github.com/bafolts/terraforming-mars/blob/main/src/Player.ts#L329

I guess that would be a better fix indeed.

sauloachkar:
Not quite happy with the code, but it works.
Unexpectedly found a bug LawSuit since it would allow it to be played even if no resources were removed.

vsrisuknimit:
```ts
let availableAmountToRemove = 0;
    switch(resource) {
      case Resources.MEGACREDITS:
        availableAmountToRemove = this.megaCredits;
        this.megaCredits = Math.max(0, this.megaCredits + amount);
        break;
      case Resources.STEEL:
        availableAmountToRemove = this.steel;
        this.steel = Math.max(0, this.steel + amount);
        break;
      case Resources.TITANIUM:
        availableAmountToRemove = this.titanium;
        this.steel = Math.max(0, this.steel + amount);
        break;
//... continued for other three
    }
```

`.number(amount >= 0 ? amount : Math.min(availableAmountToRemove, amount))`

Would this be cleaner? No need to change any other number for each and every function.

vsrisuknimit:
What is the bug about Law Suits? The test allows it to go negative. I don't see a bug on the card.

bafolts:
> ```ts
> let availableAmountToRemove = 0;
>     switch(resource) {
>       case Resources.MEGACREDITS:
>         availableAmountToRemove = this.megaCredits;
>         this.megaCredits = Math.max(0, this.megaCredits + amount);
>         break;
>       case Resources.STEEL:
>         availableAmountToRemove = this.steel;
>         this.steel = Math.max(0, this.steel + amount);
>         break;
>       case Resources.TITANIUM:
>         availableAmountToRemove = this.titanium;
>         this.steel = Math.max(0, this.steel + amount);
>         break;
> //... continued for other three
>     }
> ```
> 
> ```ts
> .number(amount >= 0 ? amount : Math.min(availableAmountToRemove, amount))
> ```
> 
> Would this be cleaner? No need to change any other number for each and every function.

I think this approach would be cleaner. I was trying to follow through the current approach before seeing this comment. Was about to suggest this as well.

sauloachkar:
> What is the bug about Law Suits? The test allows it to go negative. I don't see a bug on the card.

```
    if (game !== undefined && fromPlayer !== undefined && **amount** < 0) {
    if (game !== undefined && fromPlayer !== undefined && **variation** < 0) {
      if (fromPlayer !== this && this.removingPlayers.includes(fromPlayer.id) === false) {
        this.removingPlayers.push(fromPlayer.id);
      }
```
This is where LawSuit could fail. Sometimes amount could be negative, but variation wasn't. The test was failing because player1 had no megacredit. I setted to 1, so it'd pass.

vsrisuknimit:
> > What is the bug about Law Suits? The test allows it to go negative. I don't see a bug on the card.
> 
> ```
>     if (game !== undefined && fromPlayer !== undefined && **amount** < 0) {
>     if (game !== undefined && fromPlayer !== undefined && **variation** < 0) {
>       if (fromPlayer !== this && this.removingPlayers.includes(fromPlayer.id) === false) {
>         this.removingPlayers.push(fromPlayer.id);
>       }
> ```
> 
> This is where LawSuit could fail. Sometimes amount could be negative, but variation wasn't. The test was failing because player1 had no megacredit. I setted to 1, so it'd pass.

Law Suits does not fail. It is the new code utilizing variation that might cause an error. This should not happen if we go with the method I suggested above.

kberg:
@sauloachkar what's your plan for this PR?

sauloachkar:
> @sauloachkar what's your plan for this PR?

Sorry about the delay. I hope it's alright now.

src/Player.ts
```diff
@@ -337,7 +358,7 @@ export class Player implements ISerializable<SerializedPlayer> {
           b.player(this)
             .string(resource)
             .string(modifier)
-            .number(Math.abs(amount))
+            .number(amount >= 0 ? amount : Math.min(availableAmountToRemove, amount))
```
bafolts, line 361: If a player has 5 megacredits and `amount` were -10 this would log that the amount was decreased by `-10`. When `amount` is less than `0` I believe this should be `Math.min(availableAmountToRemove, Math.abs(amount))`. Then this would log that megacredits were decreased by `5` and not `-10`.

bafolts (changes requested):
This is still not quite right. When `amount` is negative need to reconsider what is logged.

kberg:
> This is still not quite right. When `amount` is negative need to reconsider what is logged.

I could swear I wrote a response that suggested a version of this that was shorter and I think worked out. It also had a spiel about adding tests for good measure.

kberg:
> > This is still not quite right. When `amount` is negative need to reconsider what is logged.
> 
> I could swear I wrote a response that suggested a version of this that was shorter and I think worked out. It also had a spiel about adding tests for good measure.

Here, I suggested something close to this:

```
  public addResource(resource: Resources, amount: number, options? : { log: boolean, from? : Player | GlobalEventName}) {
    const delta = Math.max(0, this.getResource(resource) + amount);

    if (resource === Resources.MEGACREDITS) this.megaCredits = += delta;
    if (resource === Resources.STEEL) this.steel = += delta;

    if (resource === Resources.TITANIUM) this.titanium = += delta;
    if (resource === Resources.PLANTS) this.plants = += delta;
    if (resource === Resources.ENERGY) this.energy = += delta;
    if (resource === Resources.HEAT) this.heat = += delta;

    if (options?.log === true) {
      const modifier = delta > 0 ? 'increased' : 'decreased';

      if (options?.from !== undefined && options.from instanceof Player && delta < 0) {
        const from: Player = options.from;
        if (from !== this && this.removingPlayers.includes(from.id) === false) {
          this.removingPlayers.push(from.id);
        }

        // Crash site cleanup hook
        if (from !== this && resource === Resources.PLANTS && delta < 0) {
          this.game.someoneHasRemovedOtherPlayersPlants = true;
        }

        this.game.log('${0}\'s ${1} amount ${2} by ${3} by ${4}', (b) =>
          b.player(this)
            .string(resource)
            .string(modifier)
            .number(Math.abs(delta))
            .player(from));
      }

      // Global event logging
      if (options?.from !== undefined && ! (options.from instanceof Player) && amount !== 0) {
        this.game.log('${0}\'s ${1} amount ${2} by ${3} by Global Event', (b) =>
          b.player(this)
            .string(resource)
            .string(modifier)
            .number(Math.abs(delta);
      }

      // Mons Insurance hook
      if (amount < 0 && options?.from !== undefined && options.from !== this) {
        this.resolveMonsInsurance();
      }
    }
  }
```

src/Player.ts
```diff
@@ -311,47 +311,47 @@ export class Player implements ISerializable<SerializedPlayer> {
...
 
   public addResource(resource: Resources, amount: number, options? : { log: boolean, from? : Player | GlobalEventName}) {
-    if (resource === Resources.MEGACREDITS) this.megaCredits = Math.max(0, this.megaCredits + amount);
-    if (resource === Resources.STEEL) this.steel = Math.max(0, this.steel + amount);
-
-    if (resource === Resources.TITANIUM) this.titanium = Math.max(0, this.titanium + amount);
-    if (resource === Resources.PLANTS) this.plants = Math.max(0, this.plants + amount);
-    if (resource === Resources.ENERGY) this.energy = Math.max(0, this.energy + amount);
-    if (resource === Resources.HEAT) this.heat = Math.max(0, this.heat + amount);
+    const delta = Math.max(0, this.getResource(resource) + amount) - this.getResource(resource);
```
kberg, line 314: I will be really surprised if the - this.getResource(resource) works.

src/Player.ts
sauloachkar: you're right. changed the whole thing again.

src/Player.ts
sauloachkar: hehe. that was kind of actually my original idea, but getResource() could really made the changes easiear and cleaner

sauloachkar:
made some tests with actual gaming using Hired Raiders, it seems to be working fine now.

src/Player.ts
```diff
@@ -311,13 +311,13 @@ export class Player implements ISerializable<SerializedPlayer> {
...
```
bafolts, line 314: This looks right to me now. Without unit tests can someone double check this?

When amount is 0 this will end up with `delta` as 0.
When amount is positive this will end up with `delta` as the positive value.
When amount is negative and sum is greater than 0 this will end up with `delta` as the negative amount.
When amount is negative and sum is less than 0 this will end up with `delta` as the amount we were able to remove.

@sauloachkar could you add a unit test for these 4 situations?

Can you add this unit test in `tests/Player.spec.ts`?

```
it('adds resources', () => {
  const player = TestPlayers.BLUE.newPlayer();
  player.megaCredits = 10;
  // adds any positive amount
  player.addResource(Resources.MEGACREDITS, 12);
  expect(player.megaCredits).eq(22);
  // removes more than we have
  player.addResource(Resources.MEGACREDITS, -23);
  expect(player.megaCredits).eq(0);
  // adds any positive amount
  player.addResource(Resources.MEGACREDITS, 5);
  expect(player.megaCredits).eq(5);
  // removes less than we have
  player.addResource(Resources.MEGACREDITS, -4);
  expect(player.megaCredits).eq(1);
  // makes no change
  player.addResource(Resources.MEGACREDITS, 0);
  expect(player.megaCredits).eq(1);
});
```

sauloachkar: done

src/Player.ts
kberg: > hehe. that was kind of actually my original idea, but getResource() could really made the changes easiear and cleaner

You know, when I looked at your code and thought "this could be simpler" I hadn't thought of getResource at first, and then it clicked. So, no surprise really. Anyway, we're reaching a good end, at least.

src/Player.ts
kberg, line 314: I see now that the expression above _is_ correct. Sorry about that. But it's also a little confusing on first glance. If you agree, consider replacing it with this:

`const delta = (amount >= 0) ? amount : Math.max(amount, -this.getResource(...)`

I just confirmed that Math.max of two negative values does what I expect:

```
Math.max(-1, -4)
-1
```
so it's possibly correct. Or don't bother, and I'll do it later.

Either way, this looks good.

sauloachkar: that's more elegant, i agree. it's done.

kberg:
image
I think you're good to go once you resolve this.

kberg (approved):

kberg:
Thank you, @sauloachkar ! That was a lot of commits and back and forth, but thank  you. Just waiting on @bafolts to approve.

sauloachkar:
> Thank you, @sauloachkar ! That was a lot of commits and back and forth, but thank you. Just waiting on @bafolts to approve.

Thank you guys for the patience, this is my first PR. Glad to help, as I love the game and have been playing this version a lot latelly.

bafolts (approved):
Sorry for the delay! Thanks for adding the unit test.

PR #3109, opened 2021-04-08
Fix card number on button

### LINKED ISSUE #3104: Bug: Misleading button labels
#3079 introduced this visual bug. Now all buttons have number on them when it does not make any sense.

image

image

@kberg

vsrisuknimit:
If working properly, this PR can close #3104.

It only adds a number to the button for selectCard only if selectCard allows it to select more than one card.

It works properly for 'Buy' during Research Phase.

image

But somehow it does not work for 'Sell'. 

image

Again it has some weird thing to do about the 'Sell' button not really being a part of SelectCard.

### DIFF
--- a/src/components/SelectCard.ts
+++ b/src/components/SelectCard.ts
@@ -113,7 +113,7 @@ export const SelectCard = Vue.component('select-card', {
       return undefined;
     },
     buttonLabel: function(): string {
-      return this.playerinput.buttonLabel + ' ' + this.cardsSelected();
+      return this.playerinput.maxCardsToSelect !== 1 ? this.playerinput.buttonLabel + ' ' + this.cardsSelected() : this.playerinput.buttonLabel;
     },
   },
   template: `<div class="wf-component wf-component--select-card">
--- a/src/components/SelectCard.ts
+++ b/src/components/SelectCard.ts
@@ -112,15 +112,18 @@ export const SelectCard = Vue.component('select-card', {
       }
       return undefined;
     },
+    isSelectOnlyOneCard: function() : boolean {
+      return this.playerinput.maxCardsToSelect === 1 && this.playerinput.minCardsToSelect === 1;
+    },
     buttonLabel: function(): string {
-      return this.playerinput.maxCardsToSelect !== 1 ? this.playerinput.buttonLabel + ' ' + this.cardsSelected() : this.playerinput.buttonLabel;
+      return this.isSelectOnlyOneCard() ? this.playerinput.buttonLabel : this.playerinput.buttonLabel + ' ' + this.cardsSelected();
     },
   },
   template: `<div class="wf-component wf-component--select-card">
         <div v-if="showtitle === true" class="nofloat wf-component-title">{{ $t(playerinput.title) }}</div>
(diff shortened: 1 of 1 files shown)


vsrisuknimit:
Similar problem that what I had here: https://github.com/bafolts/terraforming-mars/pull/3096#discussion_r606858901

bafolts:
When the `SelectCard` is part of `OrOptions` the save button is created in the `OrOptions` component.

https://github.com/bafolts/terraforming-mars/blob/main/src/components/OrOptions.ts#L143

This logic would need to share the button component with `SelectCard` instead of creating a new button.

vsrisuknimit:
> When the `SelectCard` is part of `OrOptions` the save button is created in the `OrOptions` component.
> 
> https://github.com/bafolts/terraforming-mars/blob/main/src/components/OrOptions.ts#L143
> 
> This logic would need to share the button component with `SelectCard` instead of creating a new button.

Let me know if you have a suggestion on how I might do that.

src/components/SelectCard.ts
```diff
@@ -113,7 +113,7 @@ export const SelectCard = Vue.component('select-card', {
       return undefined;
     },
     buttonLabel: function(): string {
-      return this.playerinput.buttonLabel + ' ' + this.cardsSelected();
+      return this.playerinput.maxCardsToSelect !== 1 ? this.playerinput.buttonLabel + ' ' + this.cardsSelected() : this.playerinput.buttonLabel;
```
bafolts, line 116: When `minCardsToSelect` is `0` and `maxCardsToSelect` is `1` it may still be helpful to display the number of cards selected. If this is introduced to remind the user to select a card when `minCardsToSelect` is `0` they could still forget to select a card. Can we change this to `!(this.playerinput.maxCardsToSelect === 1 && this.playerinput.minCardsToSelect === 1`? This check could also be moved into another method as we use the check [here](https://github.com/bafolts/terraforming-mars/blob/main/src/components/SelectCard.ts#L123) as well. The only time I don't think it makes sense to show the count is when one card has to be selected. The server will error if one card is not selected so the user can't make a mistake.

bafolts:
> > When the `SelectCard` is part of `OrOptions` the save button is created in the `OrOptions` component.
> > https://github.com/bafolts/terraforming-mars/blob/main/src/components/OrOptions.ts#L143
> > This logic would need to share the button component with `SelectCard` instead of creating a new button.
> 
> Let me know if you have a suggestion on how I might do that.

Not sure there will be a straight forward solution but it seems we will want to share it. This comment is unrelated to the change in this pull request. The change to share the button between the components will have to be a separate PR.

src/components/SelectCard.ts
vsrisuknimit: This PR should be merged after #3096. That PR already handle `minCardsToSelect = 0` by offering two buttons instead of one. 

See this: https://user-images.githubusercontent.com/14239220/113463228-ae441c00-93f2-11eb-8a2e-b0e0e2e044a6.png

vsrisuknimit:
@bafolts let's do that in another PR then. This is ready to go after #3096.

bafolts (approved):

PR #3115, opened 2021-04-09
Fix recruitment

### LINKED ISSUE #3114: Bug: Recruitment should not rearrange Dominance
With the current implementation, if two parties tie for dominance and a player use Recruitment card on the current dominant party, the dominance will switch the the second party.

This is because we remove a delegate and then add it back. This should not happen. Recruitment is one atomic action as @kevinb9n would say.

vsrisuknimit:
This PR can close #3114 .

Switching Mars First
Screen Shot 2021-04-08 at 8 27 12 PM

Mars First still dominant
Screen Shot 2021-04-08 at 8 27 22 PM

Test with Banned Delegate

Screen Shot 2021-04-08 at 8 31 10 PM
Screen Shot 2021-04-08 at 8 31 03 PM
Screen Shot 2021-04-08 at 8 31 21 PM

In this case, dominance check happens properly.

### DIFF
--- a/src/turmoil/Turmoil.ts
+++ b/src/turmoil/Turmoil.ts
@@ -145,16 +145,29 @@ export class Turmoil implements ISerializable<SerializedTurmoil> {
     }
 
     // Use to remove a delegate from a specific party
-    public removeDelegateFromParty(playerId: PlayerId | NeutralPlayer, partyName: PartyName, game: Game, dominanceCheck: boolean = true): void {
+    public removeDelegateFromParty(playerId: PlayerId | NeutralPlayer, partyName: PartyName, game: Game): void {
       const party = this.getPartyByName(partyName);
       if (party) {
         this.delegateReserve.push(playerId);
         party.removeDelegate(playerId, game);
+        this.checkDominantParty(party);
+      } else {
+        throw 'Party not found';
+      }
+    }
 
-        // Skip party dominant party check during removal part of Recruitment
-        if (dominanceCheck) {
-          this.checkDominantParty(party);
-        }
+    // Use to replace a delegate from a specific party with another delegate with NO DOMINANCE CHANGE
+    public replaceDelegateFromParty(
+      outgoingPlayerId: PlayerId | NeutralPlayer,
+      incomingPlayerId: PlayerId | NeutralPlayer,
+      source: 'lobby' | 'reserve' = 'lobby',
+      partyName: PartyName,
+      game: Game): void {
+      const party = this.getPartyByName(partyName);
+      if (party) {
+        this.delegateReserve.push(outgoingPlayerId);
+        party.removeDelegate(outgoingPlayerId, game);
+        this.sendDelegateToParty(incomingPlayerId, partyName, game, source);
       } else {
         throw 'Party not found';
       }
--- a/src/turmoil/Turmoil.ts
+++ b/src/turmoil/Turmoil.ts
@@ -145,12 +145,16 @@ export class Turmoil implements ISerializable<SerializedTurmoil> {
     }
 
     // Use to remove a delegate from a specific party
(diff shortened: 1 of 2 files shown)


src/turmoil/Turmoil.ts
```diff
@@ -145,12 +145,16 @@ export class Turmoil implements ISerializable<SerializedTurmoil> {
     }
 
     // Use to remove a delegate from a specific party
-    public removeDelegateFromParty(playerId: PlayerId | NeutralPlayer, partyName: PartyName, game: Game): void {
+    public removeDelegateFromParty(playerId: PlayerId | NeutralPlayer, partyName: PartyName, game: Game, dominanceCheck: boolean = true): void {
```
bafolts, line 148: Instead of adding the `dominanceCheck` argument can you call to `checkDominantParty` after calling `removeDelegateFromParty` when the argument would be `true`? The `checkDominantParty` function will have to be updated to work with `partyName`. I think this will help us in the long run.

bafolts (changes requested):

src/turmoil/Turmoil.ts
vsrisuknimit: @bafolts I don't follow your suggestion. Dominance check is still needed for Banned Delegate where dominant needed to be check immediately. We should not remove it from there.

Currently removeDelegateFromParty is only called by two places, Banned Delegate and Recruitment (under SendDelegateToArea).

vsrisuknimit: Mainly, I find it weird that sendDelegateToParty auto-adjust dominance but not removeDelegateFromParty.

bafolts: The call to `checkDominantParty` would be removed from the `removeDelegateFromParty` function. In `Banned Delegate` we would then call to:

```
removeDelegateFromParty(playerId, partyName, game);
checkDominantParty(partyName);
```

This will avoid having to introduce the boolean argument only to control what the function calls internally. Usually a function doing more than one thing eventually will lead to difficult to change functions. By using the boolean argument we know the function is doing more than one thing and are trying to have the function only do one thing under certain circumstances.

These links may better explain what I am thinking:
https://www.informit.com/articles/article.aspx?p=1392524
https://martinfowler.com/bliki/FlagArgument.html
https://medium.com/@amlcurran/clean-code-the-curse-of-a-boolean-parameter-c237a830b7a3

vsrisuknimit: @bafolts I understand but then I should also separate checkDominantParty from sendDelegateToParty() function as well, since the two functions are parallel. They should behave the same way. 

`sendDelegateToParty()` is called in many more places. If you insist, I will change but then I will have to add `checkDominantParty()` to all these calls below as well.

image

vsrisuknimit: An alternative is to write a new function specifically to replace delegate. It will copy most of the code from removeDelegateFromParty() though. Not sure about code duplication.

But this way send/removeDelegate still work as we should expect it to: that is to check dominance immediately afterward ... if you really don't want a flag argument.

```ts
    // Use to replace a delegate from a specific party with another delegate with NO DOMINANCE CHANGE
    public replaceDelegateFromParty(
      outgoingPlayerId: PlayerId | NeutralPlayer,
      incomingPlayerId: PlayerId | NeutralPlayer,
      source: 'lobby' | 'reserve' = 'lobby',
      partyName: PartyName,
      game: Game): void {
      const party = this.getPartyByName(partyName);
      if (party) {
        this.delegateReserve.push(outgoingPlayerId);
        party.removeDelegate(outgoingPlayerId, game);
        this.sendDelegateToParty(incomingPlayerId, partyName, game, source);
      } else {
        throw 'Party not found';
      }
    }
```

vsrisuknimit:
@bafolts I created a function just for replacement.

bafolts (approved):

PR #3119, opened 2021-04-10
fix start screen french translation.

MattG5:
Hi, here is my first contribution, I begin with a small and easy one.


pierrehilbert (approved):

MattG5:
sorry, I am a beginner in github, I did a mitsake with this branch (which is my main actually). I pushed an other branch as pull request. 
Moreover, the fix here is so small I can add it in another pull request

PR #3120, opened 2021-04-10
fix french translation for base cards and base corporation

MattG5:
I basically used english keyword input from chinese file as reference for translation. 
I let the traduction unchanged, i changed only the keys.
But I noticed some mistake, so some differences could appear.


MattG5:
@pierrehilbert : Salut Pierre, sur le discord, on me dit que je ne devrais pas changer les clefs alors que l'erreur semble venir de là généralement. Une idée ?
Matthieu

bafolts (approved):

PR #3131, opened 2021-04-11
set players by index

### LINKED ISSUE #3061: Bug: json upload does not allow player count change to be higher (but lower is allowed)
image

See JSON file here: https://discord.com/channels/737945098695999559/742721510376210583/826172700262531133

image

image

@alrusdi ?

bafolts:
Sets the player objects by index while uploading JSON for a game. This will close #3061

### DIFF
--- a/src/components/CreateGameForm.ts
+++ b/src/components/CreateGameForm.ts
@@ -174,21 +174,25 @@ export const CreateGameForm = Vue.component('create-game-form', {
       const refs = this.$refs;
       const file = (refs.file as any).files[0];
       const reader = new FileReader();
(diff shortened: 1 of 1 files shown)


kberg:
How do you know this works?

alrusdi (approved):

bafolts:
To verify this works I uploaded the provided json sample from the issue. After uploading that file I then incremented the number of players and saw the added player rows were added.

kberg:
> To verify this works I uploaded the provided json sample from the issue. After uploading that file I then incremented the number of players and saw the added player rows were added.

Thanks. Since there wasn't a unit test I assumed it was like that.

Issue #3116, opened 2021-04-09
Multiple milestones could not be claimed in same generation


TMJunkie:
Claimed the diversifier MS, after reaching 8 different tags but on next action could not claim Specialist MS despite having the credits and having achieved 13 energy production. For the next two turns it did not appear in my action choices, despite having the means to purchase it and having reached the MS goal. As soon as the generation changed, however, I was able to claim it. Nowhere in the rules does it state that players may only claim 1 MS per generation, so I assume it's a bug.

bafolts:
@TMJunkie was this game played on terraforming-mars.herokuapp.com? Do you have the game id?

TMJunkie:
Hi Brian,
 Unfortunately, it was played on a friend’s private server. We do this to avoid having games interrupted on updates. He updates the client in between games so we tend to be a little behind the most recent update, though not by much. Was this bug addressed on a recent update that we weren’t aware of?
Cheers,
Mike


Sent from Yahoo Mail for iPhone


On Tuesday, April 13, 2021, 8:48 PM, Brian Folts ***@***.***> wrote:




@TMJunkie was this game played on terraforming-mars.herokuapp.com? Do you have the game id?

kberg:
Not that I can think of.

>

TMJunkie:
Unfortunately, the game was wiped after the server owner updated. We will try and save games for longer in the future if it helps in tracking bugs.
You guys are doing a great job btw! We love your version of the game. Much better than what’s on steam.


Sent from Yahoo Mail for iPhone


On Wednesday, April 14, 2021, 12:42 AM, Robert Konigsberg ***@***.***> wrote:



Not that I can think of.

bafolts:
@TMJunkie I can't recreate this. Do you have more steps to reproduce? In my steps to recreate I have ensured I can fund both milestones. I see both milestones available and can then purchase both in a generation?

![image](https://user-images.githubusercontent.com/2707843/114798645-0b28c600-9d53-11eb-97a0-0c69d90c9dac.png)

Could this have perhaps been user error?

bafolts:
I couldn't recreate this.

PR #3140, opened 2021-04-15
check name when comparing award and milestone

bafolts:
While trying to recreate #3116 I noticed we compare `milestone` and `award` with the `===` operator. I don't think this is causing the bug in the ticket but it requires us to pass the same instance of the `milestone` and `award` when calling to see if a milestone or award has been claimed or funded. This changes to check the `name` property.


kberg (approved):
Excellent, thanks.

PR #3141, opened 2021-04-15
Show the None and Invert options when selecting corps

kberg:
I also moved the create game form to its own directory.


kberg:
Before: 
image
After:
image

bafolts (approved):

PR #3142, opened 2021-04-15
Add 'full' parameter to gameloader, returning participant ids along with player ids.

kberg:


kberg:
This can be used when debugging a server, or when trying to associate game ids and player ids.

kberg:
Perhaps there's a better way to do this.

src/database/GameLoader.ts
```diff
@@ -70,8 +70,17 @@ export class GameLoader implements IGameLoader {
     }
   }
 
-  public getLoadedGameIds(): Array<string> {
-    return Array.from(this.games.keys());
+  public getLoadedGameIds(): Array<{id: string, participants: Array<string>}> {
+    const entries: Map<string, Array<string>> = new Map();
+    this.participantIds.forEach((gameId, participantId) => {
```
bafolts, line 75: This may be a good opportunity to use one of the data structures that mnemonist provides with `MultiMap`. This would then become:

```diff
diff --git a/src/database/GameLoader.ts b/src/database/GameLoader.ts
index 63da9fb90..adac0c406 100644
--- a/src/database/GameLoader.ts
+++ b/src/database/GameLoader.ts
@@ -2,6 +2,7 @@ import {Database} from './Database';
 import {Game, GameId, SpectatorId} from '../Game';
 import {PlayerId} from '../Player';
 import {IGameLoader} from './IGameLoader';
+import {MultiMap} from 'mnemonist';
 
 type LoadCallback = (game: Game | undefined) => void;
 
@@ -70,16 +71,12 @@ export class GameLoader implements IGameLoader {
     }
   }
 
-  public getLoadedGameIds(): Array<{id: string, participants: Array<string>}> {
-    const entries: Map<string, Array<string>> = new Map();
+  public getLoadedGameIds(): Array<{id: GameId, participants: Array<string>}> {
+    const map = new MultiMap<GameId, string>();
     this.participantIds.forEach((gameId, participantId) => {
-      if (!entries.has(gameId)) {
-        entries.set(gameId, [participantId]);
-      } else {
-        entries.get(gameId)!.push(participantId);
-      }
+      map.set(gameId, participantId);
     });
-    const arry: Array<[string, Array<string>]> = Array.from(entries.entries());
+    const arry: Array<[string, Array<string>]> = Array.from(map.associations());
     return arry.map(([id, participants]) => ({id: id, participants: participants}));
   }
```

src/database/GameLoader.ts
```diff
@@ -70,8 +70,17 @@ export class GameLoader implements IGameLoader {
     }
   }
 
-  public getLoadedGameIds(): Array<string> {
-    return Array.from(this.games.keys());
+  public getLoadedGameIds(): Array<{id: string, participants: Array<string>}> {
```
bafolts, line 73: Instead of string this can use `{id: GameId, participants: Array<SpectatorId | PlayerId>}`.

src/routes/ApiGames.ts
```diff
@@ -10,6 +10,10 @@ export class ApiGames extends Handler {
 
   public get(_req: http.IncomingMessage, res: http.ServerResponse, ctx: IContext): void {
     const ids = ctx.gameLoader.getLoadedGameIds();
-    ctx.route.writeJson(res, ids);
+    if (ctx.url.searchParams.has('full')) {
```
bafolts, line 13: Would make sense to me to always provide what `full` is providing. I am fine updating this to always return the new `full`. What is thought process behind having the separate argument? Would like to avoid complexity if we can. I don't think this route is hit that often so performance shouldn't be an issue.

src/database/GameLoader.ts
kberg: Good idea.

src/database/GameLoader.ts
kberg: Excellent.

src/routes/ApiGames.ts
kberg: The thought process was that I didn't know if this was used elsewhere. I actually think this path is currently unused.

kberg: But I did as you suggested and removed 'full'. It was a good idea.

kberg (commented):
Thanks for the feedback!

bafolts (approved):

PR #3145, opened 2021-04-15
 Replace long text offering to take an action.

kberg:
Fixes #2958

PR #3147, opened 2021-04-16
Revert " Replace long text offering to take an action."

bafolts:
Reverts bafolts/terraforming-mars#3145

Until we fix case for single player games I think we should revert this.


kberg:
You know what? Let's keep it and move forward. I suggest: When solo, and
action is not the first, use "Take your next action." This is easy to fix.

Your call though. If you still want the revert, just say so.

>

kberg:
#3149 fixes forward.

PR #3148, opened 2021-04-16
[Moon] Many visualization fixes, mostly around using `any` when annotating moon cards

kberg:
Also:
Fixed Luna Train Station which was supposed to reward for adjacent road tiles and not mine tiles.
Lots of work on Darkside of the Moon Syndicate, and it still needs more.
Introduced simpler api for sizing viz components.


kberg:
You can see these cards at http://terraforming-mars-kberg.herokuapp.com/cards

kberg:
Click "Toggle all" and then click "Moon"

src/cards/render/CardRenderDynamicVictoryPoints.ts
```diff
@@ -74,16 +74,23 @@ export class CardRenderDynamicVictoryPoints {
...
+    item.anyPlayer = any;
     return new CardRenderDynamicVictoryPoints(item, points, points);
   }
-  public static colonyTile(points: number): CardRenderDynamicVictoryPoints {
+  public static moonColonyTile(points: number): CardRenderDynamicVictoryPoints {
     const item = new CardRenderItem(CardRenderItemType.MOON_COLONY);
     item.size = CardRenderItemSize.SMALL;
     return new CardRenderDynamicVictoryPoints(item, points, 1);
   }
+  public static moonRoadTile(points: number, any: boolean = false): CardRenderDynamicVictoryPoints {
```
bafolts, line 88: Naming this argument `any` may be difficult to read as `any` is also a typescript type. Recommend changing the argument name `any` to `anyPlayer`.

bafolts (approved):

src/cards/render/CardRenderDynamicVictoryPoints.ts
kberg: Since they're all named `any`, I can do that en masse later.

chosta: I'm late to the party but agree with the `any` to `anyPlayer` switch. The idea of `any` was introduced when implementing the requirements and I was looking for the least verbose option. Then came all other implementations about rendering elements and `any` is now kind of confusing.

PR #3153, opened 2021-04-16
canAfford now uses a struct.

kberg:
This makes calls to canAfford easier to read.

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1747,24 +1747,35 @@ export class Player implements ISerializable<SerializedPlayer> {
   public canPlay(card: IProjectCard): boolean {
     const canAfford = this.canAfford(
       this.getCardCost(card),
-      this.canUseSteel(card),
-      this.canUseTitanium(card),
-      this.canUseFloaters(card),
-      this.canUseMicrobes(card),
-      MoonExpansion.adjustedReserveCosts(this, card),
-    );
+      {
+        steel: this.canUseSteel(card),
+        titanium: this.canUseTitanium(card),
+        floaters: this.canUseFloaters(card),
+        microbes: this.canUseMicrobes(card),
+        reserveUnits: MoonExpansion.adjustedReserveCosts(this, card),
+      });
 
     return canAfford && (card.canPlay === undefined || card.canPlay(this));
   }
 
   // Checks if the player can afford to pay `cost` mc (possibly replaceable with steal, titanium etc.)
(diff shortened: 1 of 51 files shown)


src/Player.ts
```diff
@@ -1747,25 +1747,36 @@ export class Player implements ISerializable<SerializedPlayer> {
...
-  public canAfford(cost: number, canUseSteel: boolean = false, canUseTitanium: boolean = false, canUseFloaters: boolean = false, canUseMicrobes : boolean = false, reserveUnits: Units = Units.EMPTY): boolean {
-    // Check if player has the reserveUnits - required resources
+  public canAfford(cost: number, options?: {
+    steel?: boolean,
+    titanium?: boolean,
+    floaters?: boolean,
+    microbes?: boolean,
+    reserveUnits?: Units
+  }) {
+    const reserveUnits = options?.reserveUnits || Units.EMPTY;
```
bafolts, line 1770: I would use the nullish coalescing operator versus `||`. We want to use `Units.EMPTY` when `options?.reserveUnits` is `undefined` and not when it is falsey.

src/Player.ts
```diff
@@ -1747,25 +1747,36 @@ export class Player implements ISerializable<SerializedPlayer> {
...
     if (!this.hasUnits(reserveUnits)) {
       return false;
     }
 
+    const canUseSteel: boolean = options?.steel || false;
```
bafolts, line 1775: I would use the nullish coalescing operator here as well.

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing

bafolts: You can probably also simply use:

```
const canUseSteel = Boolean(options?.steel);
```

This will naturally convert `undefined` to `false` instead of having to write out `false`.

bafolts (approved):
Double checked all the arguments to the object. :+1:

src/Player.ts
kberg: Tell me more: under what conditions cound `reserveUnits` be falsey besides `undefined`?

kberg:
> Double checked all the arguments to the object.

Thanks for that extra check.

src/Player.ts
kberg: I'll switch to ?? anyway, but still, please tech.

kberg:
@sienmich FYI.

PR #3160, opened 2021-04-18
Remove 'CardRenderItemSize' and just use a constrained type.

kberg:


kberg:
This doesn't work yet

PR #3162, opened 2021-04-18
Rename CardRenderItemSize to Size

kberg:
And drop my idea of replacing the enum with strings.


kberg:
I tested by spot-checking cards.

bafolts (approved):
Double checked these. :shipit:

kberg:
Thank you!

PR #3164, opened 2021-04-18
First step standardizing on M€

kberg:
The cases I handled were "MC production" "steal MC" and "MC from".

Interesting that br translate MC to M$. /shrug


kberg:
My PR removed a bunch of blank lines and things from the JSON. Bad?

nwai90 (approved):

alrusdi (approved):

alrusdi:
> My PR removed a bunch of blank lines and things from the JSON. Bad?

No, it's ok. We don't use that JSONs directly anyway

PR #3168, opened 2021-04-18
use better key for logged cards

### LINKED ISSUE #2264: Log cards panel doesn't properly update
When you open more than one at the time and start closing them without starting by the last one.

https://user-images.githubusercontent.com/5318258/103494209-b5a66880-4e89-11eb-896d-d40ccfd5e7a1.mp4

bafolts:
Uses a key for the displayed cards. Using only `index` was leading to the bugs seen in the video provided with the bug. This fixes #2264 . Also saw we weren't passing in a boolean as expected. Vue was logging a warning regarding the `enableDontShowAgainCheckbox`. Updated that to use a boolean.

### DIFF
--- a/src/components/LogPanel.ts
+++ b/src/components/LogPanel.ts
@@ -161,7 +161,7 @@ export const LogPanel = Vue.component('log-panel', {
     },
     cardClicked: function(message: LogMessage) {
       const datas = message.data;
-      for (const data of datas) {
+      datas.forEach((data: LogMessageData) => {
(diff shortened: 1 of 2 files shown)


kberg:
Why is for.. in OK better than foreach? I thought foreach was better.

kberg (approved):

bafolts:
I can change this back to `forEach`. I added the `break` to leave the for loop once we had a hit. This is an optimization and isn't necessary.

PR #3169, opened 2021-04-18
Meaningful viz update for moon-related tiles.

kberg:
Properly separate tiles that should only show a moon tile versus those that should show a moon tile with its rate increase.

DO NOT REVIEW until #3162 is merged.


bafolts:
Can you re-base this? Now that #3162 is merged this should be far easier to review.

kberg:
> Can you re-base this? Now that #3162 is merged this should be far easier to review.

Yep, thanks! Done.

bafolts (approved):

PR #3171, opened 2021-04-18
[Moon]: Fix Lunar Project Office; it shouldn't have relied on resourceCount for tracking generations.

kberg:


kberg:
Instead IMoonData tracks the last generation it's valid.

src/cards/moon/LunaProjectOffice.ts
```diff
@@ -28,24 +27,19 @@ export class LunaProjectOffice extends Card implements IProjectCard {
...
-  public play() {
-    this.resourceCount = 2;
-    return undefined;
+  // Returns true when the current player has played Luna Project Office and the card is still valid
+  public static isActive(player: Player): boolean {
+    return MoonExpansion.ifElseMoon(player.game, (moonData) => {
+      if (!player.playedCards.some((card) => card.name === CardName.LUNA_PROJECT_OFFICE)) {
+        return false;
+      }
+      return player.game.generation <= (moonData.lunaProjectOfficeLastGeneration || -1);
```
bafolts, line 42: Unless this check wants a value of `0` for `moonData.lunaProjectOfficeLastGeneration` to use `-1` I would make this check more explicit. The number `0` is falsey and this would then use `-1`. I would use `moonData.lunaProjectOfficeLastGeneration ?? -1` if the `-1` is used when `lunaProjectOfficeLastGeneration` is `undefined`.

bafolts (approved):

src/cards/moon/LunaProjectOffice.ts
kberg: Yeah I really need to get used to ??

bafolts: I stumbled upon `??` recently and haven't looked back.

PR #3172, opened 2021-04-20
More MC to M€ conversion.

kberg:
Would you like this in smaller chunks?


src/cards/base/RadSuits.ts
```diff
@@ -20,7 +20,7 @@ export class RadSuits extends Card implements IProjectCard {
         renderData: CardRenderer.builder((b) => {
           b.production((pb) => pb.megacredits(1));
         }),
-        description: 'Requires two cities in play. Increase your MC up 1 step.',
+        description: 'Requires two cities in play. Increase your M€ 1 step.',
```
nwai90, line 23: Minor unrelated: Should be `Increase your M€ production 1 step` here. The word `production` is missing for some reason.

kberg: Good catch. I stared at the card just so far as to see that "up" was indeed not necessary.

kberg: Fixed.

nwai90 (approved):

PR #3173, opened 2021-04-20
Brazilian portuguese

sauloachkar:
More brazilian portuguese translation


kberg (commented):
One problem might be that I'm in the middle of renaming a lot of text to replace MC with M€. You'll have a fair bit of merges to resolve. Sorry about that.

bafolts (approved):

PR #3176, opened 2021-04-21
use searchParams for player input

bafolts:
I plan to start improving the action we currently know as undo. On my quest I stumbled upon this small clean up. Uses the search params instead of parsing the id out of the query string. Added unit test for the route to verify my addition of added bad request response.


kberg (approved):
yessss.

PR #3183, opened 2021-04-22
set usedUndo synchronously

bafolts:
This is part 2 (of 3 or 4) on a quest to have undo work without needing to click refresh. After this change after performing an undo the user will see the message "Not your turn to take any actions". Seconds later once we again query for the updated player the latest player object will be returned. This is still not perfect but better. In the change after this I plan to wait to send the updated player response until the database has rolled back. Once all of these commits are in when a player performs the undo last action they wont have to remember to click refresh.

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1663,10 +1663,17 @@ export class Player implements ISerializable<SerializedPlayer> {
   // Propose a new action to undo last action
   private undoTurnOption(): PlayerInput {
     return new SelectOption('Undo last action', 'Undo', () => {
-      GameLoader.getInstance().restoreGameAt(this.game.id, this.game.lastSaveId - 2, (game) => {
-        if (game !== undefined) {
-          this.usedUndo = true; // To prevent going back into takeAction()
+      this.usedUndo = true; // To prevent going back into takeAction()
+      GameLoader.getInstance().restoreGameAt(this.game.id, this.game.lastSaveId - 2, (err) => {
+        // If there is an error with restoring the game from the database this undo action has failed.
+        // We need a mechanism to tell the user this has failed. By now the `res` has been sent.
+        // For now we will keep this player instance going and hope player discovers what has happened.
+        if (err) {
+          this.takeAction();
+          return;
         }
+        // If there was no error the GameLoader has loaded the old version of `Player`
+        // This instance of `Player` will eventually be garbage collected.
       });
       return undefined;
     });
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1669,6 +1669,7 @@ export class Player implements ISerializable<SerializedPlayer> {
         // We need a mechanism to tell the user this has failed. By now the `res` has been sent.
         // For now we will keep this player instance going and hope player discovers what has happened.
         if (err) {
+          this.game.log('Unable to perform undo operation', () => {}, {reservedFor: this});
           this.usedUndo = false;
           this.takeAction();
           return;
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1669,6 +1669,7 @@ export class Player implements ISerializable<SerializedPlayer> {
         // We need a mechanism to tell the user this has failed. By now the `res` has been sent.
         // For now we will keep this player instance going and hope player discovers what has happened.
         if (err) {
+          this.usedUndo = false;
           this.takeAction();
           return;
         }
@@ -1811,8 +1812,13 @@ export class Player implements ISerializable<SerializedPlayer> {
   }
 
   public takeAction(): void {
+    /**
+     * Once an undo has been used we switch to
+     * a different instance of `Player` pulled from
+     * database. The instance where the undo was performed
+     * should no longer take actions
+     */
     if (this.usedUndo) {
-      this.usedUndo = false;
       return;
     }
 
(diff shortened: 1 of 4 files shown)


src/Player.ts
```diff
@@ -1656,10 +1656,17 @@ export class Player implements ISerializable<SerializedPlayer> {
   // Propose a new action to undo last action
   private undoTurnOption(): PlayerInput {
     return new SelectOption('Undo last action', 'Undo', () => {
-      GameLoader.getInstance().restoreGameAt(this.game.id, this.game.lastSaveId - 2, (game) => {
-        if (game !== undefined) {
-          this.usedUndo = true; // To prevent going back into takeAction()
+      this.usedUndo = true; // To prevent going back into takeAction()
```
bafolts, line 1659: This _has_ to run synchronously. Before this change we were sending the response from the server before running the query to load the previous game from the database. With this now running synchronously when the response comes back to the player there is no longer a waiting for. With no waiting for the user sees the message, "Not your turn to take any actions". After 5 seconds (or the configured polling interval) the client will then query and pull down the version we rolled back to from the database. This is not ideal but better. Plan to improve this further in next commit. Want to do this in chunks in case some part is wrong. The next step is largest change.

src/database/GameLoader.ts
```diff
@@ -116,23 +116,21 @@ export class GameLoader implements IGameLoader {
...
-          console.log(err);
-          cb(undefined);
-        }
-      });
-    } catch (error) {
-      console.log(error);
-      cb(undefined);
-    }
+  public restoreGameAt(gameId: GameId, saveId: number, cb: (err?: any) => void): void {
+    Database.getInstance().restoreGame(gameId, saveId, (err, game) => {
```
bafolts, line 120: This function should not throw so the `try...catch` is unnecessary. Any errors should come through the callback function.

src/Player.ts
```diff
@@ -1656,10 +1656,17 @@ export class Player implements ISerializable<SerializedPlayer> {
...
+      GameLoader.getInstance().restoreGameAt(this.game.id, this.game.lastSaveId - 2, (err) => {
+        // If there is an error with restoring the game from the database this undo action has failed.
```
kberg, line 1661: An error could also mean that the game is in a partial state, this seems therefore potentially messy.

src/Player.ts
```diff
@@ -1656,10 +1656,17 @@ export class Player implements ISerializable<SerializedPlayer> {
...
+        // We need a mechanism to tell the user this has failed. By now the `res` has been sent.
+        // For now we will keep this player instance going and hope player discovers what has happened.
+        if (err) {
```
kberg, line 1664: Such a failed undo should be communicated with the caller.

kberg (commented):
How do you know this works?

src/Player.ts
bafolts: Thankfully javascript is single threaded. We will always have a complete game object in javascript memory. The error should only be if we had some database error querying the database or we couldn't find the game by the last save id in the database.

src/Player.ts
bafolts: Lets merge the private message change first. In this situation I could log a private message to the user.

bafolts:
I tested this manually and saw it work. i should add a unit test for this since `usedUndo` needs to be set just right. If someone moves in the future the test should fail.

src/Player.ts
kberg: Because the GameLoader cache won't replace the Game instance?

kberg (approved):
I assume you've manually verified this.

src/Player.ts
bafolts: By the time `restoreGameAt` callback is called without error that function _should_ have replaced the game instance.

https://github.com/bafolts/terraforming-mars/blob/main/src/database/GameLoader.ts#L124

Once javascript progresses past that line the game is in memory completely. Since javascript is single threaded even any incoming requests for this game instance are waiting for the line to execute where we update the game in `GameLoader`.

bafolts:
I will add a unit test before merging.

src/Player.ts
kberg: Gotcha.

src/Player.ts
kberg: I'll do that now.

bafolts:
Still planning on adding final touches to this. Will merge this weekend.

kberg:
Sorry just clarifying: is this ready for a re-review @bafolts?

bafolts:
Yes. This now has unit tests and I added a private game log message when the rollback operation fails. Ready for review again.

kberg (approved):

src/database/GameLoader.ts
bafolts: I can keep this as it was and revert my change to remove the try...catch. It isn't needed for the change of setting `usedUndo` synchronously.

src/Player.ts
bafolts: Yes. This is a very unique situation to every other part of the code. I will add larger comment in the code. Almost warrants a TODO as there is most likely a better way to achieve what this `usedUndo` flag is being used for.

PR #3189, opened 2021-04-22
More MC -> M€ conversion.

kberg:


src/styles/turmoil.less
```diff
@@ -352,7 +352,7 @@
     margin: auto;
 }
 
-.label-5mc {
+.label-5M€ {
```
alrusdi, line 355: Is it intended? css class names with non-latin symbols are not recommended

kberg: Damn! I thought I checked them all. Sorry.

kberg: Also, I think this CSS class is unused.

kberg: Should be OK now. Sorry about that. I'm going to make smaller PRs ongoing.

alrusdi (approved):

PR #3195, opened 2021-04-23
Get the Moon secondary tag to go where it belongs, above the empty tile.

kberg:
image


kberg:
(This depends on the `wups` branch going in, but I figure you don't mind if I do both of these individually.

alrusdi (approved):

PR #3201, opened 2021-04-23
Miners ON Strike

sauloachkar:


kberg:
This won't work because it breaks all current games that will look for the previously-spelled card name. Look at `src/deck/Deck.ts` and `tests/deck/Deck.spec.ts`. This is why I said it was not trivial on the discord.

kberg:
If you mimic what those do for global event cards, you should be okay.

sauloachkar:
Don't think I can achieve this yet. Thanks for the heads up

kberg:
Understood! One step at a time.

PR #3207, opened 2021-04-25
Part of 1 server->client model restructuring

kberg:
Some very basic changes in this one.
1. Extract game-specific data from PlayerModel, call it GameModel.
2. Move Rename GameHomeModel and PlayerHomeModel to SimpleGameModel and SimplePlayerModel.
3, Finally, removed the "as PlayerModel" clause that surprised me.

Next PR will start to separate the notion of player data that is publicly knowable and player data that is only known privately.

I tested this simply with a manual test of both a solo and 2-player game, although I didn't test the 2-player game thoroughly. Any thoughts or concerns? I'll do a little more testing tomorrow.


src/components/App.ts
```diff
@@ -120,7 +120,7 @@ export const mainAppSettings = {
           app.player = xhr.response as PlayerModel;
           app.playerkey++;
           if (
-            app.player.phase === 'end' &&
+            app.player.game.phase === 'end' &&
```
kberg, line 123: You'll see this `player.game` everywhere. That's because public game data has been moved into its own data structure.

src/components/App.ts
```diff
@@ -170,7 +170,7 @@ export const mainAppSettings = {
             `${constants.APP_NAME} - Game`,
             '/game?id=' + xhr.response.id,
           );
-          app.game = xhr.response as GameHomeModel;
+          app.game = xhr.response as SimpleGameModel;
```
kberg, line 173: This rename was simply to help clarify the difference between `GameModel` and what was known previously as `GameHomeModel`. Home probably meant "This model is for the page that shares player links, but since Home isn't a term that has managed to propagate throughout the system, it doesn't seem like it needs to stay that way.

src/models/PlayerModel.ts
```diff
@@ -1,87 +1,59 @@
...
-    titanium: number;
-    titaniumProduction: number;
-    titaniumValue: number;
-    tradesThisGeneration: number;
-    turmoil: TurmoilModel | undefined;
-    undoCount: number;
-    venusScaleLevel: number;
-    victoryPointsBreakdown: VictoryPointsBreakdown;
-    waitingFor: PlayerInputModel | undefined;
+  actionsTakenThisRound: number;
```
kberg, line 10: I realphabetized this list, which makes the PR a little tricky. Sorry about that! But I promise that this + GameModel represent all the fields in PlayerModel.

src/components/App.ts
bafolts: A change like this will require clients to download this updated javascript. This wouldn't be the first time we have made a change that required the client to refresh the page without telling them. We should make an issue to handle this better. If at any point we see the server has a different `main.js` we should tell the client to refresh the page or attempt to update the client side logic for them. Writing this here as something I was thinking about while reading this change and not as a requested change.

src/components/App.ts
bafolts: Originally this `GameHomeModel` was the model which was supposed to only contain the data needed for the `GameHome` component.

src/components/App.ts
kberg: I thought the same thing. I guess the only thing to worry about is what happens if the player doesn't refresh. I can add that as a prequisite PR but this is starting to look like yak shaving. I say "starting to look like." It's probably pretty easy to do this by just constantly feeding the "version" of the binary, and if it's changed, make the suggestion. /shrug, your thoughts?

src/components/App.ts
kberg: Do you want me to change it back?

src/models/PlayerModel.ts
```diff
@@ -1,87 +1,59 @@
...
-import {FundedAwardModel} from './FundedAwardModel';
-import {Phase} from '../Phase';
 import {PlayerInputModel} from './PlayerInputModel';
-import {SpaceModel} from './SpaceModel';
-import {IAresData} from '../ares/IAresData';
 import {SerializedTimer} from '../SerializedTimer';
-import {MoonModel} from './MoonModel';
+import {GameModel} from './GameModel';
 
 export interface PlayerModel {
```
bafolts, line 9: While this change is rolling out we may want to have `PlayerModel extend GameModel` and dual write the data for at least one release of the client change. This will give clients time to update. I would dual write the properties with a `TODO` to come back and remove this after at least one client release. A week from now probably good enough. Going out as is most games in progress would have to refresh after running into errors when their clients expecting properties on `player` didn't come back from the server.

src/components/App.ts
bafolts: No. I was stating the original thought behind the name as your comment said it _probably_ meant something else.

src/components/App.ts
bafolts: Dual writing the data for at least one release should be a good enough interim solution. Using a version will have challenges. If the server were to update to a client version the differed from the one you had locally right when you went to take an action we would need to potentially capture the action save data again as it may have changed. Dual writing has worked well for us on server. This will be a rare client side dual write situation.

kberg: Take a peek at what I wrote to ask players to update their browser. If you STILL want the dual-write, please say so? I'll happily add it.

Maybe I'm convincing myself to do it anyway, but tell me your thoughts, please.

bafolts (approved):

PR #3208, opened 2021-04-25
Rework the preferences API.

kberg:
1. Limit valid fields to a key type. Avoiding using the namespace/enum solution as it seems namespaces and enums are deprecated. Huh.

2. Shortened the method names. Shorter and easier to read. 

3. Added default value parameters. These will help us standardize on `show` versus `hide` as a parameter name. In the long run.

4. Encompassed boolean management inside PreferencesManager. Each bit of code did their own thing, but all of them used `1` as a true, so it was easy to internalize.

### DIFF
--- a/src/components/PlayerHome.ts
+++ b/src/components/PlayerHome.ts
@@ -25,28 +25,28 @@ import {Phase} from '../../src/Phase';
 import * as raw_settings from '../genfiles/settings.json';
 
 export interface PlayerHomeModel {
-  hideActiveCards: boolean;
-  hideAutomatedCards: boolean;
-  hideEventCards: boolean;
+  showActiveCards: boolean;
+  showAutomatedCards: boolean;
+  showEventCards: boolean;
 }
 
 export const PlayerHome = Vue.component('player-home', {
   data: function(): PlayerHomeModel {
     return {
-      hideActiveCards: PreferencesManager.loadBoolean('hide_active_cards'),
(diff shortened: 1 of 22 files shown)


src/components/PlayerHome.ts
```diff
@@ -127,22 +127,22 @@ export const PlayerHome = Vue.component('player-home', {
       return fleetsRange;
     },
     toggleActiveCardsHiding() {
-      this.hide_active_cards = this.isActiveCardShown() ? '1': '';
+      this.hideActiveCards = this.isActiveCardShown();
```
bafolts, line 130: Prior to this change the `hide_active_cards` value would have been toggled. I _think_ this code was negating the value of `hide_active_cards`. If the value was `'1'` this code would change it to `''`. If `hideActiveCards` is a `boolean` this will need to _toggle_ (negate) the value as this is written. This would then become:

```
toggleActiveCardsHiding() {
  this.hideActiveCards = !this.hideActiveCards;
}
```

kberg: Of course.

kberg: I'll test these better.

kberg: Actually, this is correct! But this is what happens when you have variables called "hide" mapping to variables named "show." There's going to be blurred lines about translating the values between each other.

I'll make another fix.

bafolts (approved):

PR #3209, opened 2021-04-25
use a map for turmoil global events

bafolts:
Uses a `Map` instead of `Array` for the global event dealer. This will improve the look-up time of `getGlobalEventByName`.

### DIFF
--- a/src/turmoil/globalEvents/GlobalEventDealer.ts
+++ b/src/turmoil/globalEvents/GlobalEventDealer.ts
@@ -41,98 +41,93 @@ import {SerializedGlobalEventDealer} from './SerializedGlobalEventDealer';
 import {ISerializable} from '../../ISerializable';
 import {LeadershipSummit} from './LeadershipSummit';
 
-export interface IGlobalEventFactory<T> {
-    globalEventName: GlobalEventName;
-    Factory: new () => T
-}
+const COLONY_ONLY_POSITIVE_GLOBAL_EVENTS = new Map<GlobalEventName, new() => IGlobalEvent>([
+  [GlobalEventName.JOVIAN_TAX_RIGHTS, JovianTaxRights],
+]);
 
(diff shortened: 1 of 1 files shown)


src/turmoil/globalEvents/GlobalEventDealer.ts
```diff
@@ -142,28 +137,28 @@ export class GlobalEventDealer implements ISerializable<SerializedGlobalEventDea
...
+      if (game.gameOptions.coloniesExtension) events.push(...Array.from(COLONY_ONLY_NEGATIVE_GLOBAL_EVENTS));
 
       if (game.gameOptions.venusNextExtension && game.gameOptions.coloniesExtension) {
-        events.push(...VENUS_COLONY_NEGATIVE_GLOBAL_EVENTS);
+        events.push(...Array.from(VENUS_COLONY_NEGATIVE_GLOBAL_EVENTS));
       };
     }
 
-    if (game.gameOptions.venusNextExtension) events.push(...VENUS_POSITIVE_GLOBAL_EVENTS);
+    if (game.gameOptions.venusNextExtension) events.push(...Array.from(VENUS_POSITIVE_GLOBAL_EVENTS));
```
kberg, line 151: In fact, the CardManifest's notion of a Compatability parameter would work with this, too, and eliminate the repetition.

kberg (approved):

kberg:
My first comment seems to have disappeared. I was saying that I think if Dealer, CardFinder and CardLoader weren't so fuzzy and unclear, I'd suggest dumping most of this custom work and use that infrastructure. There's nothing special about this deck of cards.

src/turmoil/globalEvents/GlobalEventDealer.ts
bafolts: Yes. All good points. My goal with this change was sparing the poor heroku CPU from constantly having to iterate through `ALL_EVENTS` when looking for these events. I happened to see this source file for the first time yesterday when renaming the "Miners On Strike" event. Felt bad for the CPU.

PR #3210, opened 2021-04-26
Add alert when the server version updates.

kberg:
What do you think? Something more sophisticated would suggest a refresh when the server model changes, but I think that's probably not ideal.

Testing this was interesting -- I finally had to do it by editing `make_static_json.js`. But it works well. Refreshing the page does what you expect, and the server restarting with a new version does what you would expect.

### DIFF
--- a/src/components/App.ts
+++ b/src/components/App.ts
@@ -42,6 +42,7 @@ interface MainAppData {
     isServerSideRequestInProgress: boolean;
     componentsVisibility: {[x: string]: boolean};
     game: GameHomeModel | undefined;
+    lastServedVersion: string;
 }
 
 export const mainAppSettings = {
@@ -64,6 +65,7 @@ export const mainAppSettings = {
     } as {[x: string]: boolean},
     game: undefined as GameHomeModel | undefined,
     logPaused: false,
+    lastServedVersion: raw_settings.version,
   } as MainAppData,
   'components': {
     'start-screen': StartScreen,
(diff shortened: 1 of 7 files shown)


kberg:
Also, this is structured such that it is fully backward-compatible -- the browser will just ignore both uses of the new `version` field.

src/components/App.ts
```diff
@@ -148,6 +151,19 @@ export const mainAppSettings = {
...
+    testServerVersion: function(version: string) {
+      const app = this as unknown as typeof mainAppSettings.data;
+      const player = app.player;
+      if (player === undefined) {
+        return;
+      }
+      if (version !== undefined && version.length > 0) {
+        if (app.lastServedVersion !== version) {
+          app.lastServedVersion = version;
+          alert('A new version of the app is available. Please refresh your browser so it remains compatible with the server.');
```
bafolts, line 163: This change opens up a can of worms.

We ignore this problem today. For long running multi-player games if we were to update the server in a way which broke your client you may never know it is your turn. We don't know the server update is going to cause a problem. Looking at the server traffic on heroku when I deploy everyday there are usually at least 10-20 requests per second so I assume multi-player games are in progress. Every player of that game waiting would have to interact with this alert on a daily basis.

I would be in favor of continuing to ignore this problem. We usually do some due diligence to avoid causing problems during updates. It is rare to have a change that breaks clients but after this change we would require users to refresh when they may not need to take any action.

Instead of putting this alert here it may make more sense to add a `try...catch` where we attempt to update the player with the latest server response. This is where we would run into a runtime error. We aren't catching it today. Showing some updated user model that showed the error in an error modal and recommended refreshing the page may be more effective and simpler than this change.

https://github.com/bafolts/terraforming-mars/blob/main/src/components/App.ts#L120

A try catch around that entire block should catch any issues with the server response.

kberg: I agree with you. The 10-20 users disrupted per day was one I was also concerned about but didn't tie a number to it.

I'll close this PR. I enjoyed writing it, though.

PR #3213, opened 2021-04-27
Prelude cards filter; Create game form page styling fixes

alrusdi:
Backend already supports prelude filtering via cardsBlackList, so i've done only client side part.
Added small visual hint to distinguish project cards and preludes in cards filter widget: 
prelude_filter

### DIFF
--- a/compile-vue-templates.js
+++ b/compile-vue-templates.js
@@ -54,8 +54,8 @@ generateVueClass(
   [],
 );
 generateVueClass(
-  'src/components/CardsFilter',
-  require('./build/src/components/CardsFilter').CardsFilter,
+  'src/components/create/CardsFilter',
+  require('./build/src/components/create/CardsFilter').CardsFilter,
   ['selectedCardNames', 'foundCardNames', 'searchTerm'],
 );
 generateVueClass(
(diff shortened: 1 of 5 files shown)


compile-vue-templates.js
```diff
@@ -54,8 +54,8 @@ generateVueClass(
   [],
 );
 generateVueClass(
-  'src/components/CardsFilter',
-  require('./build/src/components/CardsFilter').CardsFilter,
+  'src/components/create/CardsFilter',
```
kberg, line 57: Thank you. I look forward to moving other components to the create directory. This is nice.

src/components/create/CardsFilter.ts
```diff
@@ -22,6 +22,9 @@ export const CardsFilter = Vue.component('cards-filter', {
   },
   mixins: [TranslateMixin],
   methods: {
+    isPrelude: function(cardName: CardName) {
+      return ALL_PRELUDE_CARD_NAMES.includes(cardName);
```
kberg, line 26: While I suspect this is inefficient, the reality is that it just has to be faster than someone typing.

kberg (approved):
This seems reasonable to me.

PR #3224, opened 2021-04-29
hide scores for awards and milestones

### LINKED ISSUE #2090: Bug: Planner count should be hidden until the game starts
image

You can see how many cards players keep under Planner. This might influence players who have not picked from going for or against planners.

This count should be hidden until the game starts.

bafolts:
Hides the scores for the awards and milestones during initial card selection. Closes #2090

### DIFF
--- a/src/components/Award.ts
+++ b/src/components/Award.ts
@@ -8,6 +8,10 @@ export const Award = Vue.component('award', {
     awards_list: {
       type: Array as () => Array<FundedAwardModel>,
     },
+    show_scores: {
+      type: Boolean,
+      default: true,
+    },
   },
   data: function() {
     const showDescription: {[x: string]: boolean} = {};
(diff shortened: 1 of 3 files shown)


kberg (commented):
Two minor problems with this PR which you may not have considered:

First, the scores are still accessible to a cheater who can fetch results. I suggest instead just not bothering to include these in ServerModel until completing the initial draft.

Also, no matter what, this problem is also exposed during the research phase. (Not only that, I've noticed that the app logs how may cards a single player kept even though others haven't finished their final purchase in the draft. But that's another bug.)

What would you like to do?

bafolts:
> First, the scores are still accessible to a cheater who can fetch results. I suggest instead just not bothering to include these in ServerModel until completing the initial draft.

Not including them in the results until the initial draft is complete works. I had considered the cheating angle but didn't take it as serious.

May need to handle the research phase issues in a separate PR. For the research phase I assume you would want to see the values from before the research phase started.

kberg:
> > First, the scores are still accessible to a cheater who can fetch results. I suggest instead just not bothering to include these in ServerModel until completing the initial draft.
> 
> Not including them in the results until the initial draft is complete works. I had considered the cheating angle but didn't take it as serious.

I'll say that listening in on the moderator's channel, there's plenty of that around. But OK!

> May need to handle the research phase issues in a separate PR. For the research phase I assume you would want to see the values from before the research phase started.

OK

kberg (approved):

PR #3225, opened 2021-04-29
Fix reds and chairman

vsrisuknimit:
image

Chairman uses deferredAction to resolve the policy, so TR increase should be deferredAction too. This way TR from being a chairman is not raised until after the new ruling bonus is resolved.

### DIFF
--- a/src/turmoil/Turmoil.ts
+++ b/src/turmoil/Turmoil.ts
@@ -15,6 +15,7 @@ import {SerializedTurmoil} from './SerializedTurmoil';
 import {PLAYER_DELEGATES_COUNT} from '../constants';
 import {AgendaStyle, PoliticalAgendasData, PoliticalAgendas} from './PoliticalAgendas';
 import {CardName} from '../CardName';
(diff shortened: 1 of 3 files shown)


src/turmoil/Turmoil.ts
```diff
@@ -313,7 +314,13 @@ export class Turmoil implements ISerializable<SerializedTurmoil> {
...
           const steps = player.corporationCard?.name === CardName.TEMPEST_CONSULTANCY ? 2 :1;
-          player.increaseTerraformRatingSteps(steps);
+
+          // Raise TR but after resolving the new policy
+          game.defer(new DeferredAction(player, () => {
+            player.increaseTerraformRatingSteps(steps);
+            return undefined;
+          }));
+
           game.log('${0} is the new chairman and gained ${1} TR', (b) => b.player(player).number(steps));
```
alrusdi, line 324: Log message should be inside the defered call

vsrisuknimit: fixed

nwai90 (approved):

PR #3228, opened 2021-04-30
Log drawn cards privately for a player.

kberg:
I've tested this in a game on my heroku instance. But what I'd like is some suggestions on what you would want to see in a good test. I have some ideas.


kberg:
I think this is ready.

bafolts (approved):

PR #3232, opened 2021-04-30
Minor optimization on waitingFor API, and related simplifications.

kberg:
Reviewer: the key is that the new function signature for `setWaitingFor` which sets a default second parameter:

`public setWaitingFor(input: PlayerInput, cb: () => void = () => {}): void {`

### DIFF
--- a/src/deferredActions/DeferredActionsQueue.ts
+++ b/src/deferredActions/DeferredActionsQueue.ts
@@ -40,9 +40,7 @@ export class DeferredActionsQueue {
 
     const input = action.execute();
     if (input !== undefined) {
-      action.player.setWaitingFor(input, () => {
-        cb();
-      });
+      action.player.setWaitingFor(input, cb);
     } else {
       cb();
     }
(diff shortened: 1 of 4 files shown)


src/deferredActions/DeferredActionsQueue.ts
```diff
@@ -40,9 +40,7 @@ export class DeferredActionsQueue {
 
     const input = action.execute();
     if (input !== undefined) {
-      action.player.setWaitingFor(input, () => {
-        cb();
-      });
+      action.player.setWaitingFor(input, cb);
```
kberg, line 43: No need for the extra dressing.

tests/Player.spec.ts
```diff
@@ -123,7 +123,7 @@ describe('Player', function() {
     const mockOption = new SelectOption('Mock select option', 'Save', () => {
       return mockOption2;
     });
-    player.setWaitingFor(mockOption, () => done());
+    player.setWaitingFor(mockOption, done);
```
kberg, line 126: No need for the extra dressing.

bafolts (approved):

PR #3235, opened 2021-05-01
Make logging adding resources part of adding resources.

kberg:


src/Player.ts
```diff
@@ -654,7 +655,10 @@ export class Player implements ISerializable<SerializedPlayer> {
       }
     }
   }
-  public addResourceTo(card: IResourceCard, count: number = 1): void {
+
+  public addResourceTo(card: IResourceCard & ICard, options: number | {qty?: number, log?: boolean} = 1): void {
```
kberg, line 659: This method works where `options` can still be a number, but of course I would understand droping that (honestly I'd rather just always log _unless indciated to not log_.)

src/ares/AresHandler.ts
```diff
@@ -75,7 +74,7 @@ export class AresHandler {
       const availableCards = player.getResourceCards(resourceType);
       if (availableCards.length === 0) {
       } else if (availableCards.length === 1) {
-        player.addResourceTo(availableCards[0]);
+        player.addResourceTo(availableCards[0], {log: true});
```
kberg, line 77: This wasn't logging, now it is.

src/cards/base/CEOsFavoriteProject.ts
```diff
@@ -34,8 +34,14 @@ export class CEOsFavoriteProject extends Card implements IProjectCard {
       'Add resource',
       player.getCardsWithResources().concat(robotCards.map((c) => c.card)),
       (foundCards: Array<ICard>) => {
-        player.addResourceTo(robotCards.find((c) => c.card.name === foundCards[0].name) ?? foundCards[0]);
-        LogHelper.logAddResource(player, foundCards[0]);
+        // if the user selected a robot card, handle it here:
```
kberg, line 37: This is more lines, but I think it's easier to undertand.

tests/Player.spec.ts
```diff
@@ -327,227 +328,257 @@ describe('Player', function() {
...
 
-  player.adjustProduction(Units.of({heat: -10}));
-  expect(asProductionUnits(player)).deep.eq({
-    megacredits: 10,
-    steel: 9,
-    titanium: 8,
-    plants: 7,
-    energy: 6,
-    heat: 5,
+  it('addResourceTo', () => {
```
kberg, line 554: This file makes more sense if you hide the whtiespace changes, because it turns out some of the tests weren't actually being run (see line 330 and line 583.)

src/cards/moon/DarksideObservatory.ts
```diff
@@ -43,10 +43,10 @@ export class DarksideObservatory extends Card implements IProjectCard, IActionCa
 
   private addResource(card: ICard, player: Player): void {
     if (card.resourceType === ResourceType.DATA) {
-      player.addResourceTo(card, 2);
+      player.addResourceTo(card, {qty: 2, log: true});
```
kberg, line 46: Was not logged. Fixed.

src/cards/moon/DarksideObservatory.ts
```diff
@@ -43,10 +43,10 @@ export class DarksideObservatory extends Card implements IProjectCard, IActionCa
...
     }
     if (card.resourceType === ResourceType.SCIENCE) {
-      player.addResourceTo(card, 1);
+      player.addResourceTo(card, {qty: 1, log: true});
```
kberg, line 49: Was not logged. Fixed.

src/Player.ts
```diff
@@ -654,7 +655,10 @@ export class Player implements ISerializable<SerializedPlayer> {
...
+    const count = typeof(options) === 'number' ? options : (options.qty ?? 1);
```
bafolts, line 660: The force is strong with this one. Excellent usage of typescript operators.

src/Player.ts
bafolts: I am always hesitant of boolean arguments like `log`. Another option is to have a public function on `Player` for `addAndLogResource` which then calls to `addResourceTo`.

```ts
public addAndLogResource(card: IResourceCard & ICard, count: number = 1) {
  this.addResourceTo(card, count);
  LogHelper.logAddResource(this, card, count);
}
```

I would prefer this to the separate argument to `addResourceTo`.

bafolts: As a big fan of java I often ask myself WWJD (what would java do) when writing typescript. Would prefer some solution which doesn't have to check the argument types at runtime. What would you do in java with this situation?

bafolts (approved):
This is a good change and will also decrease the build size. I don't want to block this by putting in a requested change over the API change. Would like to avoid having to check the type of argument at runtime if possible and perhaps use a separate function or some other alternative.

src/Player.ts
kberg: In Java I would have two methods that overload each other:

`addBlah(card, count)` and `addBlah(count, card, log: boolean)` where the first calls the second. Effective Java suggests replacing booleans like this with an enum. But keep in mind that enums are not recommended Typescript (I learned in recent weeks.)

kberg:
> This is a good change and will also decrease the build size. I don't want to block this by putting in a requested change over the API change. Would like to avoid having to check the type of argument at runtime if possible and perhaps use a separate function or some other alternative.

The runtime type check can be removed if I follow up and aggresively remove the number part of the `number | {options}` thing.

PR #3238, opened 2021-05-01
Rename Preferences to Sidebar.

kberg:
This is the first of a few changes that simplify Preferences a bit, separate components, clean up CSS and also make room for The Moon global parameters on the sidebar.


alrusdi (approved):

kberg:
Thanks!

Issue #3199, opened 2021-04-23
[Moon] the Mars is Terraformed popup doesn't take Moon into account.


kberg:

kberg:
This is fixed.

PR #3239, opened 2021-05-01
Move the "Mars is Terraformed!" announcement to PlayerHome, and get the result from the server, not the client.

kberg:


kberg:
I tested this manually.

kberg:
Fixes #3199

src/components/PlayerHome.ts
```diff
@@ -182,6 +186,11 @@ export const PlayerHome = Vue.component('player-home', {
   },
   mounted: function() {
     window.addEventListener('keydown', this.navigatePage);
+    if (this.player.game.isTerraformed && TerraformedAlertDialog.shouldAlert && PreferencesManager.load('show_alerts') === '1') {
```
bafolts, line 189: When this model change rolls out clients will have the client side code checking model properties which exist. This will allow older clients to catch up.

:+1:

src/components/Board.ts
```diff
@@ -43,9 +36,6 @@ export const Board = Vue.component('board', {
     temperature: {
       type: Number,
     },
-    shouldNotify: {
```
bafolts, line 46: This should also remove the `shouldNotify` attribute from the `GameEnd` component.

https://github.com/bafolts/terraforming-mars/blob/main/src/components/GameEnd.ts#L191

bafolts (changes requested):
I can't be certain but with the `shouldNotify` property being removed from `Board` we may end up showing the `alert` twice with this change. We were providing `shouldNotify` with `false` from the `GameEnd` component.

The only requested change I have is to remove the `shouldNotify` attribute being set on `Board` from the `GameEnd` component. Since the property was removed we should remove it from every place it was used.

src/components/Board.ts
kberg: Good catch. Why didn't the compiler catch this?

kberg:
> I can't be certain but with the `shouldNotify` property being removed from `Board` we may end up showing the `alert` twice with this change. We were providing `shouldNotify` with `false` from the `GameEnd` component.
> 
> The only requested change I have is to remove the `shouldNotify` attribute being set on `Board` from the `GameEnd` component. Since the property was removed we should remove it from every place it was used.

I think it might happen if a player refreshes the page, now that you mention it. But then again, the same thing would have happened with the old code, wouldn't it?

bafolts:
With the old code if you were on the `GameEnd` screen where we provide the `shouldNotify` property a value of `false` I don't think the alert would have been seen. We used to show the alert and then transition to the `GameEnd` screen.

kberg:
The alert isn't something shown just _before_ the game end screen, since Mars can be terraformed long before the generation ends.

Unless i'm missing something.

Perhaps restate plainly: what do you want me to change or what do you want me to confirm before approving this change?

bafolts:
The only requested change I have is to remove the shouldNotify attribute being set on Board from the GameEnd component. Since the property was removed we should remove it from every place it was used.

src/components/Board.ts
bafolts: This is one of my least favorite answers but I don't know. The vue template compiler must not consider properties being provided to an element that aren't defined as an error.

kberg:
> The only requested change I have is to remove the shouldNotify attribute being set on Board from the GameEnd component. Since the property was removed we should remove it from every place it was used.

Oh, of course. That makes sense now. Thanks.

bafolts (approved):

PR #3240, opened 2021-05-01
Alert dialog is now in fixed position.

kberg:
This means if, for instance, you get a server error, it won't scroll all the way to the top of the page.

Here's where this happens all the time: select "Take an action from a played card" but don't select a card. Before, the whole page scrolled up. Now the page doesn't move just to show the error dialog.
Before:
image

After: 
image


alrusdi (approved):

kberg:
Thank you!

PR #3241, opened 2021-05-01
Move global parameters to a dedicated sidebar css class.

kberg:
This is the second step in reworking the sidebar a bit.

### DIFF
--- a/src/components/Sidebar.ts
+++ b/src/components/Sidebar.ts
@@ -266,73 +266,73 @@ export const Sidebar = Vue.component('sidebar', {
     <div class="preferences_panel" v-if="ui.preferences_panel_open">
       <div class="preferences_panel_item">
         <label class="form-switch">
-          <input type="checkbox" v-on:change="updatePreferences" v-model="hide_hand"></input>
+          <input type="checkbox" v-on:change="updatePreferences" v-model="hide_hand">
           <i class="form-icon"></i> <span v-i18n>Hide cards in hand</span>
         </label>
       </div>
       <div class="preferences_panel_item">
         <label class="form-switch">
-          <input type="checkbox" v-on:change="updatePreferences" v-model="hide_awards_and_milestones"></input>
+          <input type="checkbox" v-on:change="updatePreferences" v-model="hide_awards_and_milestones">
           <i class="form-icon"></i> <span v-i18n>Hide awards and milestones</span>
         </label>
       </div>
       <div class="preferences_panel_item">
         <label class="form-switch">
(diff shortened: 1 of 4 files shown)


src/components/Sidebar.ts
```diff
@@ -193,161 +193,167 @@ export const Sidebar = Vue.component('sidebar', {
...
-                      <game-setup-detail :gameOptions="gameOptions" :playerNumber="playerNumber" :lastSoloGeneration="lastSoloGeneration"></game-setup-detail>
+<div :class="'sidebar_cont sidebar '+getSideBarClass()" :data="syncPreferences()">
+  <div class="tm">
+    <div class="gen-text">GEN</div>
+    <div class="gen-marker">{{ getGenMarker() }}</div>
+  </div>
+  <div v-if="gameOptions.turmoilExtension">
+    <div :class="'party-name party-name-indicator party-name--'+rulingPartyToCss()" v-html="getRulingParty()"></div>
+  </div>
+  <div class="global_params">
```
kberg, line 204: 1. Be sure to hide whitespace changes
2. Note dropping the "preferences" prefix as this is using the sidebar namespace.

src/components/Sidebar.ts
```diff
@@ -193,161 +193,167 @@ export const Sidebar = Vue.component('sidebar', {
...
+          <i class="form-icon"></i>
+          <span v-i18n>Learner Mode (req. refresh)</span>
+          <span class="tooltip tooltip-left" data-tooltip="Show information that can be helpful\n to players who are still learning the games">&#9432;</span>
+        </label>
+      </div>
+      <div class="preferences_panel_item form-group">
+        <label class="form-label"><span v-i18n>Language</span> (<a href="javascript:document.location.reload(true);" v-i18n>refresh page</a> <span v-i18n>to see changes</span>)</label>
+        <div class="preferences_panel_langs">
+          <label class="form-radio" v-for="language in langs">
+            <input name="lang" type="radio" v-on:change="updatePreferences" v-model="lang" :value="language.id"></input>
```
bafolts, line 345: I don't believe it is permitted for the `input` element to have a closing tag.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input

```
Tag omission | Must have a start tag and must not have an end tag.
```

kberg: I removed it from everywhere. What a weird rule.

bafolts: I had to look this up. I vaguely remembered this rule and then took the time to find the proof. Is very strange.

bafolts (approved):

PR #3249, opened 2021-05-02
Show Global Events when clicking on them in the log.

kberg:
image

### DIFF
--- a/src/components/LogPanel.ts
+++ b/src/components/LogPanel.ts
@@ -16,6 +16,10 @@ import {Color} from '../Color';
 import {SoundManager} from './SoundManager';
 import {PreferencesManager} from './PreferencesManager';
 import {GlobalEventName} from '../turmoil/globalEvents/GlobalEventName';
+import {GlobalEvent} from './GlobalEvent';
+import {getGlobalEventByName} from '../turmoil/globalEvents/GlobalEventDealer';
+import {GlobalEventModel} from '../models/TurmoilModel';
+import {PartyName} from '../turmoil/parties/PartyName';
 
 let logRequest: XMLHttpRequest | undefined;
 
@@ -49,6 +53,7 @@ export const LogPanel = Vue.component('log-panel', {
   },
   components: {
     Card,
+    GlobalEvent,
   },
   methods: {
     scrollToEnd: function() {
@@ -198,7 +203,7 @@ export const LogPanel = Vue.component('log-panel', {
             this.cards.splice(index, 1);
           }
         }
-        if (data.type === LogMessageDataType.CARD) {
+        if (data.type === LogMessageDataType.GLOBAL_EVENT) {
           const globalEventName = data.value as GlobalEventName;
           const index = this.globalEventNames.indexOf(globalEventName);
           if (index === -1) {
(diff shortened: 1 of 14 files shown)


kberg:
Fixes #3057

kevinb9n:
random: the new boxed card name style is nice. If it helps, the physical cards have a background color of approx. #a8d8e8 and I think that color would work well here.

src/components/LogPanel.ts
```diff
@@ -267,20 +313,23 @@ export const LogPanel = Vue.component('log-panel', {
               {{ n }}
             </div>
           </div>
-          <span class="label-additional" v-html="getGenerationText()"></span>
+          <span class="label-additional" v-if="players.length === 1"><span :class="lastGenerationClass">of {{this.lastSoloGeneration}}</span></span>
```
bafolts, line 316: Winner! Every `v-html` removed is a step in right direction.

src/turmoil/Turmoil.ts
```diff
@@ -233,9 +233,11 @@ export class Turmoil implements ISerializable<SerializedTurmoil> {
 
       // 2 - Global Event
       // TODO: create a LogMessageDataType for Global Events, ref https://github.com/bafolts/terraforming-mars/issues/3057
-      if (this.currentGlobalEvent) {
-        game.log('Global event "' + this.currentGlobalEvent.name + '" has been resolved.');
-        this.currentGlobalEvent.resolve(game, this);
+      if (this.currentGlobalEvent !== undefined) {
+        const currentGlobalEvent: IGlobalEvent = this.currentGlobalEvent;
+        game.log('Resolving global event ${0}', (b) => b.globalEvent(currentGlobalEvent));
+        // game.log('Global event "' + this.currentGlobalEvent.name + '" has been resolved.');
```
bafolts, line 239: Remove comment before flight.

bafolts (approved):

kberg:
> random: the new boxed card name style is nice. If it helps, the physical cards have a background color of approx. #a8d8e8 and I think that color would work well here.

Probably true, not the simplest thing for me to do, amusingly. I'll follow up with this.

kberg:
image
OK.

src/components/Turmoil.ts
```diff
@@ -234,26 +235,17 @@ export const Turmoil = Vue.component('turmoil', {
...
-            <div class="event-party event-party--top" :class="'event-party--'+partyNameToCss(turmoil.current.revealed)" v-i18n>{{ turmoil.current.revealed }}</div>
-            <div class="event-party event-party--bottom" :class="'event-party--'+partyNameToCss(turmoil.current.current)" v-i18n>{{ turmoil.current.current }}</div>
-            <div class="event-content" v-i18n>{{ turmoil.current.description }}</div>
-          </div>
+        <global-event v-if="turmoil.distant" :globalEvent="turmoil.distant" type="distant"></global-event>
+        <global-event v-if="turmoil.coming" :globalEvent="turmoil.coming" type="coming"></global-event>
+        <global-event v-if="turmoil.current" :globalEvent="turmoil.current" type="current"></global-event>
       </div>
-
+^
```
alrusdi, line 248: Is it real symbol or just github glitch?

kberg: Dang.

kberg (commented):
I'll fix that.

src/components/Turmoil.ts
kberg: Fixed, thanks.

PR #3257, opened 2021-05-05
Make timer hours the same height as minutes and seconds

alrusdi:


kberg (approved):
@ssimeonoff this OK with you?

PR #3261, opened 2021-05-07
perform undo asynchronously

bafolts:
This is a step towards a more robust and reliable undo. This change waits to send the response of the action until we have rolled the game back within the database and javascript memory. It also moves the logic out of `Player` making it clearer how this works.

I still need to update the unit tests.

### DIFF
--- a/src/components/WaitingFor.ts
+++ b/src/components/WaitingFor.ts
@@ -94,19 +94,15 @@ export const WaitingFor = Vue.component('waiting-for', {
             if (result.result === 'GO') {
               root.updatePlayer();
 
-              // Only show notification for multi-player games
-              // todo bafolts remove once undo refactor complete
-              if (this.player.players.length > 1) {
-                if (Notification.permission !== 'granted') {
-                  Notification.requestPermission();
-                }
-                if (Notification.permission === 'granted') {
-                  new Notification(constants.APP_NAME, {
-                    icon: '/favicon.ico',
-                    body: 'It\'s your turn!',
-                  });
-                }
+              if (Notification.permission !== 'granted') {
+                Notification.requestPermission();
+              } else if (Notification.permission === 'granted') {
+                new Notification(constants.APP_NAME, {
+                  icon: '/favicon.ico',
+                  body: 'It\'s your turn!',
+                });
               }
+
               const soundsEnabled = PreferencesManager.load('enable_sounds') === '1';
               if (soundsEnabled) SoundManager.playActivePlayerSound();
 
@@ -134,12 +130,7 @@ export const WaitingFor = Vue.component('waiting-for', {
     window.clearInterval(documentTitleTimer);
     if (this.waitingfor === undefined) {
       this.waitForUpdate();
-      let message = 'Not your turn to take any actions';
-      // todo bafolts remove once undo refactor complete
-      if (this.players.length === 1) {
-        message = '';
-      }
-      return createElement('div', $t(message));
+      return createElement('div', $t('Not your turn to take any actions'));
     }
     if (this.player.players.length > 1 && this.player.waitingFor !== undefined) {
       documentTitleTimer = window.setInterval(() => this.animateTitle(), 1000);
--- a/src/database/GameLoader.ts
+++ b/src/database/GameLoader.ts
@@ -116,25 +116,25 @@ export class GameLoader implements IGameLoader {
     this.getByParticipantId(spectatorId, cb);
   }
 
-  public restoreGameAt(gameId: GameId, saveId: number, cb: (err?: any) => void): void {
+  public restoreGameAt(gameId: GameId, saveId: number, cb: LoadCallback): void {
     try {
       Database.getInstance().restoreGame(gameId, saveId, (err, game) => {
         if (err) {
           console.error('error while restoring game', err);
-          cb(err);
+          cb(undefined);
         } else if (game !== undefined) {
           Database.getInstance().deleteGameNbrSaves(gameId, 1);
           this.add(game);
           game.undoCount++;
-          cb();
+          cb(game);
         } else {
           console.error('game not found while restoring game', err);
-          cb(new Error('game not found'));
+          cb(undefined);
         }
       });
     } catch (error) {
       console.log(error);
-      cb(error);
+      cb(undefined);
     }
   }
 
--- a/src/routes/PlayerInput.ts
+++ b/src/routes/PlayerInput.ts
@@ -43,10 +43,8 @@ export class PlayerInput extends Handler {
 
   private isWaitingForUndo(player: Player, entity: Array<Array<string>>): boolean {
     const waitingFor = player.getWaitingFor();
-    return waitingFor instanceof OrOptions &&
-           entity.length > 0 &&
-           entity[0].length > 0 &&
-           waitingFor.options[Number(entity[0][0])] instanceof UndoActionOption;
+    return entity.length > 0 && entity[0].length > 0 &&
+           waitingFor instanceof OrOptions && waitingFor.options[Number(entity[0][0])] instanceof UndoActionOption;
   }
 
   private performUndo(res: http.ServerResponse, ctx: IContext, player: Player): void {
--- a/src/routes/PlayerInput.ts
+++ b/src/routes/PlayerInput.ts
@@ -1,9 +1,11 @@
 import * as http from 'http';
-import {GameLoader} from '../database/GameLoader';
+import {Game} from '../Game';
 import {Player} from '../Player';
 import {Server} from '../models/ServerModel';
 import {Handler} from './Handler';
 import {IContext} from './IHandler';
+import {OrOptions} from '../inputs/OrOptions';
+import {UndoActionOption} from '../inputs/UndoActionOption';
 
 export class PlayerInput extends Handler {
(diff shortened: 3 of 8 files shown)


src/routes/PlayerInput.ts
```diff
@@ -39,6 +41,26 @@ export class PlayerInput extends Handler {
     });
   }
 
+  private isWaitingForUndo(player: Player, entity: Array<Array<string>>): boolean {
+    const waitingFor = player.getWaitingFor();
+    return waitingFor !== undefined &&
+           waitingFor instanceof OrOptions &&
+           entity.length > 0 &&
+           entity[0].length > 0 &&
+           waitingFor.options[Number(entity[0][0])].title === 'Undo last action';
```
bafolts, line 50: Not a fan of how this has to spot the undo action. Will most likely create a specific `UndoAction` and check the type here.

src/routes/PlayerInput.ts
```diff
@@ -39,6 +41,26 @@ export class PlayerInput extends Handler {
...
+  }
+
+  private performUndo(res: http.ServerResponse, ctx: IContext, player: Player): void {
+    GameLoader.getInstance().restoreGameAt(player.game.id, player.game.lastSaveId - 2, (game: Game | undefined) => {
+      if (game === undefined) {
+        player.game.log('Unable to perform undo operation', () => {}, {reservedFor: player});
+        player.takeAction();
```
bafolts, line 57: Note to self. This has to send the response with `player`.

src/components/WaitingFor.ts
```diff
@@ -94,19 +94,16 @@ export const WaitingFor = Vue.component('waiting-for', {
...
-                if (Notification.permission === 'granted') {
-                  new Notification(constants.APP_NAME, {
-                    icon: '/favicon.ico',
-                    body: 'It\'s your turn!',
-                  });
-                }
+              if (Notification.permission !== 'granted') {
+                Notification.requestPermission();
               }
+              if (Notification.permission === 'granted') {
```
kberg, line 100: `else`?

src/components/WaitingFor.ts
```diff
@@ -134,12 +131,7 @@ export const WaitingFor = Vue.component('waiting-for', {
     window.clearInterval(documentTitleTimer);
     if (this.waitingfor === undefined) {
       this.waitForUpdate();
-      let message = 'Not your turn to take any actions';
-      // todo bafolts remove once undo refactor complete
-      if (this.players.length === 1) {
-        message = '';
-      }
-      return createElement('div', $t(message));
+      return createElement('div', $t('Not your turn to take any actions'));
```
kberg, line 134: Side note: I was looking at this the other day, and thought that we shold replace this text entirely with a small visualization of the N players, whose turn it is, and who the first player is. In the same space where this "not your turn" was, and it would keep someone from having to look at the player boards.

src/database/GameLoader.ts
```diff
@@ -116,25 +116,25 @@ export class GameLoader implements IGameLoader {
     this.getByParticipantId(spectatorId, cb);
   }
 
-  public restoreGameAt(gameId: GameId, saveId: number, cb: (err?: any) => void): void {
+  public restoreGameAt(gameId: GameId, saveId: number, cb: LoadCallback): void {
```
kberg, line 119: Is this just a change you made while you happened to work on this?

src/routes/PlayerInput.ts
kberg: Oh god yes, I hate this too. `// TODO` please.

src/routes/PlayerInput.ts
```diff
@@ -39,6 +41,26 @@ export class PlayerInput extends Handler {
     });
   }
 
+  private isWaitingForUndo(player: Player, entity: Array<Array<string>>): boolean {
+    const waitingFor = player.getWaitingFor();
+    return waitingFor !== undefined &&
+           waitingFor instanceof OrOptions &&
```
kberg, line 47: If waitingFor is an instance of something then it's undefined. You can remove the check one line above, can't you?

src/routes/PlayerInput.ts
kberg: Help me here -- your note to self makes me think this isn't done yet, so how should I consider it?

src/routes/PlayerInput.ts
```diff
@@ -39,6 +41,26 @@ export class PlayerInput extends Handler {
...
+           entity.length > 0 &&
+           entity[0].length > 0 &&
+           waitingFor.options[Number(entity[0][0])].title === 'Undo last action';
+  }
+
+  private performUndo(res: http.ServerResponse, ctx: IContext, player: Player): void {
+    GameLoader.getInstance().restoreGameAt(player.game.id, player.game.lastSaveId - 2, (game: Game | undefined) => {
```
kberg, line 54: Mind adding a comment here explaining why it's -2 and not -1? I always think it.

src/routes/PlayerInput.ts
```diff
@@ -39,6 +41,26 @@ export class PlayerInput extends Handler {
...
+      if (game === undefined) {
+        player.game.log('Unable to perform undo operation', () => {}, {reservedFor: player});
```
kberg, line 56: "Unable to undo, read the log for details" ?

kberg: Or "Unable to undo, this could be a database problem so check the log"?

kberg (commented):
I don't see anything bad, but it sounded like you had more work to do, so just comments only for now.

src/database/GameLoader.ts
bafolts: The `IGameLoader` interface defines this as `LoadCallback`. It actually took me a few minutes to track down what the issue was here. Changed this to satisfy the interface.

src/routes/PlayerInput.ts
bafolts: I will need to see why it is -2 myself. I will then add comment.

src/routes/PlayerInput.ts
bafolts: This actually came up in #3263 . Some clearer message here will be helpful for everyone. Users hosting locally with SQLite may never see this error. With the PostgreSQL from the hobby basic plan users will see this.

bafolts:
@kberg this is now ready for review. With this change when performing an undo it will appear to happen right away as a user. Refreshing the page is no longer needed. This can close #3263 as it should prevent the situation where user losers their turn on a failure. The new error message also indicates they can retry.

This could be worth a local trial run. I was able to undo an operation fine with local testing. Also forced an error and tested the error where the undo operation fails.

src/routes/PlayerInput.ts
```diff
@@ -39,6 +41,32 @@ export class PlayerInput extends Handler {
     });
   }
 
+  private isWaitingForUndo(player: Player, entity: Array<Array<string>>): boolean {
+    const waitingFor = player.getWaitingFor();
+    return waitingFor instanceof OrOptions &&
```
kberg, line 46: For readability consider

```
return entity.length > 0 && entity[0].length > 0 &&
    waitingFor instanceof OrOptions && waitingFor.options[Number(entity[0][0])] instanceof UndoActionOption;
```

tests/Player.spec.ts
```diff
@@ -291,43 +288,6 @@ describe('Player', function() {
     srr.targetCards.push({card: new LunarBeam(), resourceCount: 0});
     expect(player.getSelfReplicatingRobotsTargetCards().length).eq(1);
   });
-  it('uses undo', function() {
```
kberg, line 294: If you use `() =>` instead of `function()` here (but more important with the call to restoresGameAt in the next test) `this` has ... I don't have the word for it ... a better context.

tests/routes/PlayerInput.spec.ts
```diff
@@ -28,4 +33,62 @@ describe('PlayerInput', function() {
...
+    ctx.url = new URL('http://boo.com' + req.url);
+    player.beginner = true;
+    const game = Game.newInstance('foo', [player], player);
+    const undo = Game.newInstance('old', [player], player);
+    ctx.gameLoader.add(game);
+    game.gameOptions.undoOption = true;
+    player.process([['1'], ['Power Plant:SP']]);
+    const options = player.getWaitingFor() as OrOptions;
+    options.options.push(new UndoActionOption());
+    ctx.gameLoader.restoreGameAt = function(__gameId: string, __lastSaveId: number, cb: (game: Game | undefined) => void) {
```
kberg, line 49: You only need one underbar.

tests/routes/PlayerInput.spec.ts
```diff
@@ -28,4 +33,62 @@ describe('PlayerInput', function() {
...
+    };
+    PlayerInput.INSTANCE.post(req, res.hide(), ctx);
+    req.emit('data', JSON.stringify([[String(options.options.length - 1)], ['']]));
+    req.emit('end');
+    const model = JSON.parse(res.content);
+    expect(game.gameAge).not.eq(undo.gameAge);
+    expect(model.game.gameAge).eq(undo.gameAge);
+  });
+
+  it('reverts to old if undo fails', () => {
```
kberg, line 60: Reverts to old what?

kberg (approved):

PR #3262, opened 2021-05-07
Remove more v-html

kberg:


kberg:
Manually tested.

kberg:
Never mind -- I forgot that I did an even better job on this that I was waiting to send out.

PR #3270, opened 2021-05-09
More static card conversion.

kberg:


kberg:
There is a very good chance these are the last holdouts for static property management.

bafolts (approved):

bafolts:
> There is a very good chance these are the last holdouts for static property management.

Considering only a few months ago the heroku dyno would run out of memory and restart the process this is great news.

PR #3271, opened 2021-05-09
Massively simplify Robotic Workforce by distributing the authority to the cards.

kberg:
There is no way this is 100% right. I need to very carefully review this. But I also think this is a good change that keeps RoboticWorkforce simple and clear.

This will need documentation, feedback, etc.


src/cards/base/RoboticWorkforce.ts
```diff
@@ -286,76 +63,18 @@ export class RoboticWorkforce extends Card implements IProjectCard {
...
-        b.player(player).card(foundCard).card(this));
-
+        b.player(player).card(card).card(this));
+
+      if (card.produce) {
+        card.produce(player, false);
+      } else if (card.productionBox) {
+        player.adjustProduction(card.productionBox);
+      } else {
+        // Ugh
```
bafolts, line 76: Will this `Ugh` become a logging of `logIllegalState`?

kberg: It could be!

src/cards/base/MedicalLab.ts
```diff
@@ -28,8 +28,13 @@ export class MedicalLab extends Card implements IProjectCard {
     });
   }
 
+  public produce(player: Player, includeThis: boolean) {
```
bafolts, line 31: This change makes sense to me and will avoid code duplication between RoboticWorkforce and the other project cards. I think we can avoid the `includeThis` argument. We can calculate `includeThis`. Calling to `player.cardIsInEffect(this.name)` will return `true` if the card has been played. Before it has been played is when we are passing in `true` for `includeThis`. I would change each `(includeThis ? 1 : 0)` with `player.cardIsInEffect(this.name) ? 0 : 1)`.

kberg: Agreed.

kberg: In fact, I don't entirely agree. IMO the real problem is that play() is called before the card is in the tableau. The behavior of the card should be that it is now in the tableau and executed for its one and only time.

So the proper fix is to put the card in the tableau, AND THEN call play(). That would take the "including this" into account naturally.

bafolts (approved):

PR #3252, opened 2021-05-04
Defer all possible requirements tests to super.canPlay()

kberg:
Fixes #3234

PR #3282, opened 2021-05-10
fix cultural metropolis canPlay bug

vsrisuknimit:
#3252 broke this with a sloppy deletion of canPlay() from cultural metropolis.

This PR adds back the requirement check for energy and delegate count. It also improves the tests on this card.

### DIFF
--- a/src/cards/turmoil/CulturalMetropolis.ts
+++ b/src/cards/turmoil/CulturalMetropolis.ts
@@ -34,13 +34,22 @@ export class CulturalMetropolis extends Card implements IProjectCard {
   }
 
   public canPlay(player: Player): boolean {
+    if ( ! super.canPlay(player)) {
+      return false;
+    }
+
+    if (player.getProduction(Resources.ENERGY) < 1) {
+      return false;
+    }
+
+    // This card requires player has 2 delegates available
     const turmoil = player.game.turmoil;
     if (turmoil !== undefined) {
-      // This card requires player has 2 delegates available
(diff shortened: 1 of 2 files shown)


kberg (commented):
Yeah, I see it now. The energy production was tucked in there. Can you move the energy production and super.canPlay() part outside of the turmoil test? Because those are part of a comment that "this card requires player has 2 delegates available." This is what confused me.  And thanks for the tests.

vsrisuknimit:
There is no point to move those out. No need to check energy prod or the ruling party before knowing that turmoil is even in play.

I find `if (! super.canPlay(player)) return false;` to be ugly and less readable. Adding energy check as another if statement even more so.

bafolts:
@vsrisuknimit the deletion was not _sloppy_ but an oversight. We will all encounter this while writing software as we are human. I do not think referring to it as _sloppy_ is constructive. The point of moving them out is also an optimization and nearly every other card is using `super.canPlay`. When moving between other cards this should then make every card similar. This one card may be harder to read but as a whole when moving between multiple cards they will all read the same.

bafolts (approved):

bafolts:
@vsrisuknimit this bug fix is helpful. In our attempts to continually re-factor the code base to make adding new features simpler we are going to introduce minor bugs such as this one -- it is inevitable. It isn't helpful to anyone to refer to another person's code change as a "sloppy deletion". None of us are making money from this project and are all doing the work for either the love of the board game, the love of programming, or both. Even if it wasn't intended I felt your comments were coming from an aggressive place versus a collaborative place. In the future we should all try to be more collaborative.

PR #3283, opened 2021-05-11
add methods to DeferredActionsQueue to act for Player

### LINKED ISSUE #2498: Suggestion: Helion to pick the payment at the card picking time
Another suggestion from someone that plays with people in different time zones and it affects the waiting time considerably.
The fact Helion need to choose how to pay at THE END of the card picking phase makes it slower as requiring an extra action from someone out of order.

Say player Helion picks his card. Player Tharsis is the 1st player, and also pick his card. Now Tharsis needs to wait Helion to come back just so he can say how much heat he wants to use to pay for his cards. Waiting for a player to play his turn is fine. But additional waiting just for something that could've been solved when picking, seems unnecessary, if it can be avoided (of course). 

For now, our group is even removing Helion just because of that.

Cheers from Down Under. Love your work guys.

PS: Sorry for the typo at the title

bafolts:
This can close #2498 . During the research phase if a player has deferred actions those actions will be set for that player before they are done with their research phase. Will no longer run the deferred action for all after research phase.

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -876,16 +876,14 @@ export class Game implements ISerializable<SerializedGame> {
   }
 
   public playerIsFinishedWithResearchPhase(player: Player): void {
-    if (this.deferredActions.hasActionFor(player)) {
-      this.deferredActions.runAllFor(player, () => this.playerIsFinishedWithResearchPhase(player));
-      return;
-    }
-    this.researchedPlayers.add(player.id);
-    if (this.allPlayersHaveFinishedResearch()) {
-      this.phase = Phase.ACTION;
-      this.passedPlayers.clear();
-      this.startActionsForPlayer(this.first);
-    }
+    this.deferredActions.runAllFor(player, () => {
+      this.researchedPlayers.add(player.id);
+      if (this.allPlayersHaveFinishedResearch()) {
+        this.phase = Phase.ACTION;
+        this.passedPlayers.clear();
+        this.startActionsForPlayer(this.first);
+      }
+    });
   }
 
   public playerIsFinishedWithDraftingPhase(initialDraft: boolean, player: Player, cards : Array<IProjectCard>): void {
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -876,13 +876,15 @@ export class Game implements ISerializable<SerializedGame> {
   }
 
   public playerIsFinishedWithResearchPhase(player: Player): void {
+    if (this.deferredActions.hasActionFor(player)) {
+      this.deferredActions.runAllFor(player, () => this.playerIsFinishedWithResearchPhase(player));
+      return;
+    }
     this.researchedPlayers.add(player.id);
     if (this.allPlayersHaveFinishedResearch()) {
-      this.deferredActions.runAll(() => {
-        this.phase = Phase.ACTION;
-        this.passedPlayers.clear();
-        this.startActionsForPlayer(this.first);
-      });
+      this.phase = Phase.ACTION;
+      this.passedPlayers.clear();
+      this.startActionsForPlayer(this.first);
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/Game.ts
```diff
@@ -876,13 +876,15 @@ export class Game implements ISerializable<SerializedGame> {
   }
 
   public playerIsFinishedWithResearchPhase(player: Player): void {
+    if (this.deferredActions.hasActionFor(player)) {
```
bafolts, line 879: I plan to test this manually prior to merging. As I don't play the game anymore the only corporation I can think of which may have to use a deferred action during research phase is Helion as they have to provide how they are going to pay for the card. Does anyone else know of another situation to test? There would have to be a project card that triggered a deferred action from the research phase. I am assuming the `deferredActions.runAll` was being used here since there wasn't a function to run only for one player.

@kevinb9n @kberg @vsrisuknimit @nwai90

nwai90: @bafolts I did a quick scan of all official corps. It seems that only Helion would have to make a choice during the research phase. Other corporations' deferred actions reside in other phases.

bafolts: I will test with this situation before marking this PR ready for review.

bafolts:
@kberg this is now ready for review. I tested it locally with Helion and it now runs as requested in the issue. A player can select how they will pay for their cards while other players are still selecting their cards.

src/deferredActions/DeferredActionsQueue.ts
```diff
@@ -46,6 +45,33 @@ export class DeferredActionsQueue {
     }
   }
 
+  public hasActionFor(player: Player): boolean {
+    return this.queue.toArray().some((da) => da.player.id === player.id);
+  }
+
+  public runAllFor(player: Player, cb: () => void): void {
+    let action: DeferredAction | undefined;
```
kberg, line 53: Add a comment, like

```
// pop all elements off the heap until finding one for 'player'. Then
// execute player and push the other elements back into the queue.
```

src/deferredActions/DeferredActionsQueue.ts
```diff
@@ -46,6 +45,33 @@ export class DeferredActionsQueue {
     }
   }
 
+  public hasActionFor(player: Player): boolean {
+    return this.queue.toArray().some((da) => da.player.id === player.id);
+  }
+
+  public runAllFor(player: Player, cb: () => void): void {
```
kberg, line 52: My one concern for this is that it introduces a use case that violates the heap data structure, as evidenced by your pop/restore mnechanism. If the queue was ever large, that would be pretty inefficient. Is Heap still the right data structure? Or could we get away with an underlying data structure of an array of priorities?

kberg (approved):

src/Game.ts
```diff
@@ -876,13 +876,15 @@ export class Game implements ISerializable<SerializedGame> {
...
+      this.deferredActions.runAllFor(player, () => this.playerIsFinishedWithResearchPhase(player));
```
kberg, line 880: Can you explain briefly why this is correct? If someone comes reading this they'll want to know why the research phase is special in this case, and why there might be other actions in the queue.

src/deferredActions/DeferredActionsQueue.ts
bafolts: An array of priorities should work fine. The `pop` and `peek` functions need to call to `sort` but those functions are only used during unit tests. The `runAll` can find the action with highest priority and run it in `O(n)`.

src/Game.ts
bafolts: Any player is not finished with the research phase until their deferred actions triggered while selecting projects cards to purchase are finished. During the action phase every deferred action in the queue is usually ran as an action that a player takes could push a deferred action to the queue for another player. Using the `runAll` could actually lead to an issue. We always call to `setWaitingFor` for a player even if that player was already waiting for something else. 

https://github.com/bafolts/terraforming-mars/blob/main/src/deferredActions/DeferredActionsQueue.ts#L43

We assume here that this player doesn't have a waiting for already. If this player were to already have a waiting for it would be lost. We should assert here that `player.getWaitingFor()` is undefined before we try to set a player input. Could very well have bugs that we aren't aware of today. It may make sense to use this `runAllFor` in other places where `runAll` is being used.

PR #3285, opened 2021-05-13
consider delta for Mons Insurance and Lawsuit

### LINKED ISSUE #3281: Comet For Venus not checking if player has M$
Comet For Venus is only checking if a player has Venus Tags and not if it has any M$ to be lost.
That caused Mons to play $3 to a player that lost no M$.

That's where probably it should be checked:
```
if (venusTagPlayers.length > 0) { ... 
}

```
WhatsApp Image 2021-05-09 at 22 37 20

bafolts:
Considers how much is removed from a player for the mons insurance hook. Fixes #3281 

Added regression test.

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -347,7 +347,7 @@ export class Player implements ISerializable<SerializedPlayer> {
     else if (resource === Resources.ENERGY) this.energy += delta;
     else if (resource === Resources.HEAT) this.heat += delta;
     else {
-      throw new Error('tried to add unsupported resource');
+      throw new Error(`tried to add unsupported resource ${resource}`);
     }
 
     if (options?.log === true) {
@@ -376,7 +376,7 @@ export class Player implements ISerializable<SerializedPlayer> {
     else if (resource === Resources.ENERGY) this.energyProduction += delta;
     else if (resource === Resources.HEAT) this.heatProduction += delta;
     else {
-      throw new Error('tried to add unsupported production');
+      throw new Error(`tried to add unsupported production ${resource}`);
     }
 
     if (options?.log === true) {
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -366,17 +366,8 @@ export class Player implements ISerializable<SerializedPlayer> {
   }
 
   public addProduction(resource: Resources, amount : number, options? : { log: boolean, from? : Player | GlobalEventName}) {
-    let delta = amount;
-    if (amount < 0) {
-      if (resource === Resources.MEGACREDITS) {
-        const production = this.getProduction(resource);
-        if (production + amount < -5) {
-          delta = -(production + 5);
-        }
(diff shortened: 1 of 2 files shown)


src/Player.ts
```diff
@@ -357,29 +360,44 @@ export class Player implements ISerializable<SerializedPlayer> {
...
   public addProduction(resource: Resources, amount : number, options? : { log: boolean, from? : Player | GlobalEventName}) {
-    if (resource === Resources.MEGACREDITS) this.megaCreditProduction = Math.max(-5, this.megaCreditProduction + amount);
-    if (resource === Resources.STEEL) this.steelProduction = Math.max(0, this.steelProduction + amount);
-    if (resource === Resources.TITANIUM) this.titaniumProduction = Math.max(0, this.titaniumProduction + amount);
-    if (resource === Resources.PLANTS) this.plantProduction = Math.max(0, this.plantProduction + amount);
-    if (resource === Resources.ENERGY) this.energyProduction = Math.max(0, this.energyProduction + amount);
-    if (resource === Resources.HEAT) this.heatProduction = Math.max(0, this.heatProduction + amount);
+    let delta = amount;
+    if (amount < 0) {
+      if (resource === Resources.MEGACREDITS) {
```
bafolts, line 371: @kberg I couldn't see a simpler way to write this. Suggestions welcome.

kberg:
```
const adj = resource === Resources.MEGACREDITS ? 5 : 0;
const delta = (amount >= 0) ? amount : Math.max(amount, -(this.getResource(resource) + adj));
```

Something very close to this. One or two of the (implied) positive or (overt) negative terms might have to switch. A test would help.

bafolts: There seem to be plenty of tests that fail when you get this wrong. With enough trial and error tests passing and logic seems sounds.

src/Player.ts
```diff
@@ -357,29 +360,35 @@ export class Player implements ISerializable<SerializedPlayer> {
...
+    const delta = (amount >= 0) ? amount : Math.max(amount, -(this.getProduction(resource) - adj));
+
+    if (resource === Resources.MEGACREDITS) this.megaCreditProduction += delta;
+    else if (resource === Resources.STEEL) this.steelProduction += delta;
+    else if (resource === Resources.TITANIUM) this.titaniumProduction += delta;
+    else if (resource === Resources.PLANTS) this.plantProduction += delta;
+    else if (resource === Resources.ENERGY) this.energyProduction += delta;
+    else if (resource === Resources.HEAT) this.heatProduction += delta;
+    else {
+      throw new Error('tried to add unsupported production');
```
kberg, line 379: Might as well the enum value in the error.

kberg (approved):
Minor: the PR talks about Mons Insurance but also applies to Lawsuit.

PR #3288, opened 2021-05-16
A series of resource management changes that prevent going negative.

kberg:


kberg:
"Prepare to track bad calls" is because I'm still getting reports of steel going negative, and I have no good way to track this. This change will at least forgive the error, and also improves readability when removing resources.

src/Player.ts
```diff
@@ -351,7 +387,7 @@ export class Player implements ISerializable<SerializedPlayer> {
     }
 
     if (options?.log === true) {
-      this.logUnitDelta(resource, amount, 'amount', options.from);
+      this.logUnitDelta(resource, delta, 'amount', options.from);
```
kberg, line 390: While in here I also fixed this, which it's the amount that was actually taken.

src/Game.ts
```diff
@@ -1644,4 +1645,16 @@ export class Game implements ISerializable<SerializedGame> {
...
+  public logIllegalState(description: string, metadata: {}) {
+    const gameMetadata = {
+      gameId: this.id,
+      lastSaveId: this.lastSaveId,
+      logAge: this.gameLog.length,
+      currentPlayer: this.activePlayer,
+
+      metadata: metadata,
+    };
+    console.warn('Illegal state: ' + description, JSON.stringify(gameMetadata, null, ' '));
```
bafolts, line 1658: Logging a warning for the time being makes sense. Eventually storing these in the database would be helpful. I rarely look at heroku logs and heroku only keeps a limited number of logs. Putting in database could then query for these and others.

kberg: I thought about a database but wow, didn't want to start thinking about that. Also, I have another small commit lined up which adds a long entry  that says "Hey, there's an error. Go tell your admin."

src/Player.ts
```diff
@@ -337,8 +337,47 @@ export class Player implements ISerializable<SerializedPlayer> {
...
+    // But if the intent is to remove the amount requested (spending 8 plants to place a greenery) then there
+    // better be 8 units. The code outside this call is responsible in those cases for making sure the player
+    // has enough resource units to pay for an action.
+    //
+    // In those cases, if the player calls this, but the logic is wrong, the player could wind up with a
+    // negative amount of units. This will break other actions in the game. So instead, this method deducts as
+    // much as possible, and lots that there was a game error.
+    //
+    // The shortcut for knowing if this is the case is when `options.from` is undefined.
+    if (delta !== amount && options?.from === undefined) {
```
bafolts, line 376: May make sense to leverage `console.assert` instead of introducing `logIllegalState`. With `console.assert` could avoid this separate function and write this as.

```
console.assert(delta === amount || options?.from !== undefined, `Adjusting ${amount} ${resource} when player has ${playerAmount}`, { player: {color: this.color, id: this.id, name: this.name}, resource, amount });
```

kberg: My goal of writing an additional function is to make it very easy for someone to log specific details about a game that we try to hunt down while investigating issues.

kberg: Make sense?

bafolts (approved):

src/Player.ts
bafolts: Yes.

src/Game.ts
bafolts: Yes. The problem here is really that heroku doesn't keep logs for long. I should write a script I run locally to collect logs from the terraforming-mars host on a daily basis.

PR #3291, opened 2021-05-18
Add message to log about illegal state.

kberg:


src/Player.ts
```diff
@@ -377,6 +377,7 @@ export class Player implements ISerializable<SerializedPlayer> {
       this.game.logIllegalState(
         `Adjusting ${amount} ${resource} when player has ${playerAmount}`,
         {player: {color: this.color, id: this.id, name: this.name}, resource, amount});
+      this.game.log('An illegal state has occurred; please ask the server owner to read the server log.');
```
bafolts, line 380: Calling to `this.log` from within the `logIllegalState` function would avoid having to call this every place we call to `this.game.logIllegalState`.

kberg: Good idea.

kberg: You know what? Rather than run the risk of polluting the log with this message, let's wait -- I'll put together something that prevents that.

kberg: (IOW put this PR on hold./)

PR #3293, opened 2021-05-18
Remove logGainStandardResource and logGainProduction, logging is part of addResource and addProduction.

kberg:
The only sad part is that logGainStandardResource's text is better English.


src/cards/base/MartianRails.ts
```diff
@@ -36,9 +36,8 @@ export class MartianRails extends Card implements IActionCard, IProjectCard {
     return player.energy >= 1;
   }
   public action(player: Player) {
-    player.energy--;
+    player.addResource(Resources.ENERGY, -1);
```
bafolts, line 39: This should probably be `deductResource`.

src/cards/base/PowerInfrastructure.ts
```diff
@@ -40,9 +39,8 @@ export class PowerInfrastructure extends Card implements IActionCard, IProjectCa
       'Select amount of energy to spend',
       'Spend energy',
       (amount: number) => {
-        player.energy -= amount;
-        player.megaCredits += amount;
-        LogHelper.logGainStandardResource(player, Resources.MEGACREDITS, amount);
+        player.addResource(Resources.ENERGY, -amount);
```
bafolts, line 42: This should use `deductResource`.

bafolts (approved):

src/cards/base/MartianRails.ts
kberg: Yep.

src/cards/base/PowerInfrastructure.ts
kberg: :+1:

PR #3294, opened 2021-05-18
Add Moon to the sidebar

kberg:
More importantly, it moves individual some of the sidebar material to its own component.


src/tools/analyze_ma.html
```diff
@@ -6,10 +6,10 @@
   </head>
   <body>
     <form>
-      <label for="runs">Runs: </label><input type="text" id="runs" value="10000" name="runs"></input><br/>
-      <label for="venus">Venus: </label><input type="checkbox" id="venus" name="venus"></input><br/>
-      <label for="ares">Ares: </label><input type="checkbox" id="ares" name="ares"></input><br/>
-      <label for="moon">The Moon: </label><input type="checkbox" id="moon" name="moon"></input><br/>
+      <label for="runs">Runs: </label><input type="text" id="runs" value="10000" name="runs"><br/>
```
alrusdi: Shouldn't the input tag be closed?

vsrisuknimit: Void elements. They cannot have a closing tag. https://www.w3.org/TR/2011/WD-html-markup-20110113/syntax.html#void-element

alrusdi: Ups! My fault. I didn't read all PR

kberg: But.. new question: how did you make a comment on a file completely unrelated to this PR? That's interesting.

kberg: :laughing:

src/components/GlobalParameterValue.ts
```diff
@@ -0,0 +1,62 @@
+import Vue from 'vue';
+import {TranslateMixin} from './TranslateMixin';
+import {MAX_OCEAN_TILES, MAX_OXYGEN_LEVEL, MAX_TEMPERATURE, MAX_VENUS_SCALE} from '../constants';
+import {GlobalParameter} from '../GlobalParameter';
+
+export const GlobalParameterValue = Vue.component('global-parameter-value', {
+  props: {
+    param: {
+      type: Number, // GlobalParameter
```
bafolts, line 9: I _think_ you can do `type: Number as GlobalParameter`.

src/components/Sidebar.ts
```diff
@@ -202,28 +183,22 @@ export const Sidebar = Vue.component('sidebar', {
...
-    <div class="global_params_value" v-html="getTemperatureCount()"></div>
-    <div class="oxygen-tile"></div>
-    <div class="global_params_value" v-html="getOxygenCount()"></div>
-    <div class="ocean-tile"></div>
-    <div class="global_params_value" v-html="getOceanCount()"></div>
-    <div v-if="gameOptions.venusNextExtension">
-      <div class="venus-tile"></div>
-      <div class="global_params_value" v-html="getVenusCount()"></div>
-    </div>
+    <global-parameter-value :param="this.globalParameter.TEMPERATURE" :value="this.temperature"></global-parameter-value>
```
bafolts, line 186: I don't _think_ you need to use `this` in templates. Vue-js may warn against this in console.

bafolts (approved):

src/components/GlobalParameterValue.ts
kberg: I hit a couple of snags. It had to be `Number as unknown as GlobalParameter` but then vue complained 

```
No overload matches this call.
  The last overload gave the following error.
    Type '{ param: { type: GlobalParameter; }; value: { type: NumberConstructor; }; }' is not assignable to type 'PropsDefinition<DefaultProps> | undefined'.
      Types of property 'param' are incompatible.
        Type '{ type: GlobalParameter; }' is not assignable to type 'PropValidator<any> | undefined'.
          Types of property 'type' are incompatible.
            Type 'GlobalParameter' is not assignable to type 'PropType<any> | undefined'.ts(2769)
```

If you know much about next steps, I'm happy to hear it.

PR #3296, opened 2021-05-18
Restore the whimsical "Oh" when you can't afford to buy a card.

epsalon:
This was "fixed" as a "typo", but it was in fact intentional. The "Oh" is the reaction to not being able to buy any cards.

### DIFF
--- a/src/deferredActions/DrawCards.ts
+++ b/src/deferredActions/DrawCards.ts
@@ -80,7 +80,7 @@ export class DrawCards<T extends undefined | SelectCard<IProjectCard>> implement
     const min = options.paying ? 0 : options.keepMax;
     const msg = options.paying ? (max === 0 ? 'You cannot afford any cards' : 'Select card(s) to buy') :
       `Select ${max} card(s) to keep`;
-    const button = max === 0 ? 'Ok' : (options.paying ? 'Buy' : 'Select');
+    const button = max === 0 ? 'Oh' : (options.paying ? 'Buy' : 'Select');
     const cb = (selected: Array<IProjectCard>) => {
       if (options.paying && selected.length > 0) {
         player.game.defer(


kberg:
This joke reminds me of the NYT Crossword's little joke when you fill all the squares but don't get it quite right. I wonder if this was inspired by that. But it's sort of a weird spot for a joke, gets missed (I have seen it) so I'd guess it doesn't quite work on its own. It's not quite overt enough, and so from my view it succeeds more as an inside joke than something everyone would get. An example of some of the jokes that I think do work well are the occasional icons which pop up (eg the alien in search for life, Santa flying over Mars, et cetera.)

If we were to keep this, then I'd suggest changing "Oh" to "Oh." which IMO works better as the utterance such a button is meant to convey. But we also need to update the translations, probably by restoring the "Oh" translation, either to something local which matches the joke, or to the local version of "Ok."

I like whimsy. In fact, Alon, you can independently verify this because we work at the same company. Go search around to see some of my track record on the matter. I still regret that my former TL didn't allow me to have an utterly adorable Oscar the Grouch pop up over a trashcan icon on the company monitoring dashboard. (I look forward to your DM. ;) )

bafolts (changes requested):
So at one point this was "Ok" but changed to "Oh".

https://github.com/bafolts/terraforming-mars/commit/e938fe1fa0baff36b3b777c193ef012e1da7c433

It was then changed back to "Ok" with the understanding that it was a typo. There are translation keys associated with this "Ok" that need to be updated if changing back to "Oh".

https://github.com/bafolts/terraforming-mars/blob/main/src/locales/br/preferences.json#L33
https://github.com/bafolts/terraforming-mars/blob/main/src/locales/cn/preferences.json#L28
https://github.com/bafolts/terraforming-mars/blob/main/src/locales/es/preferences.json#L31

Mapping to `"Ok"` as some of them are doing is fine. Without this mapping users would see the string `"Oh"` and not the translation.

kberg:
Since there's been no response in over a month, closing.

PR #3301, opened 2021-05-20
clear ended games on the hour

### LINKED ISSUE #3300: Implement some form of memory collector to free heap space.
The heroku instance (terraforming-mars.herokuapp.com) is still quite popular. With more being logged to the game logs we are also logging more. This is causing the memory to climb again. Starting two days ago was the first day I have seen the heroku instance run out of memory and restart the dyno. This restarting of the dyno has proven risky. Last winter this would sometimes prevent the dyno from restarting and the host would become unreachable.

image

One strategy which I think will be effective is running an interval. This interval will look for games in the `GameLoader` that have ended but we are storing in javascript memory. It would remove these from javascript memory. A player that is still using the instance could request the game. If the game were requested we would pull it from the database.

bafolts:
This can close #3300 . Every 60 minutes it removes all games with phase of ENDED from javascript memory.


src/database/GarbageCollector.ts
```diff
@@ -0,0 +1,28 @@
...
+  public start(): void {
+    this.interval = global.setInterval(() => {
+      this.run();
+    }, 1000 * 60 * 60 /* every hour */);
+  }
+  protected run(): void {
+    console.log('Running garbage collector');
+    this.gameLoader.getLoadedGames().forEach((game, gameId) => {
+      if (game !== undefined && game.phase === Phase.END) {
+        console.log(`GarbageCollector: Removing ${gameId} from game loader`);
```
kberg, line 18: "Unloading"? :D

kberg: Consider adding an x-minute delay on games that might have ended, or were accessed, less than a minute ago.

src/database/GarbageCollector.ts
```diff
@@ -0,0 +1,28 @@
...
+  constructor(private gameLoader: IGameLoader) {
+  }
+  public start(): void {
+    this.interval = global.setInterval(() => {
+      this.run();
+    }, 1000 * 60 * 60 /* every hour */);
+  }
+  protected run(): void {
+    console.log('Running garbage collector');
+    this.gameLoader.getLoadedGames().forEach((game, gameId) => {
```
kberg, line 16: Move this inside gameLoader, call it `purge` or something, and then you don't have to expose all the game instances, nor the remove method.

kberg (commented):
No concerns, but take a look at my comments. I suspect you'll change one or two things.

src/database/GarbageCollector.ts
bafolts: Do we know when a game was ended or last accessed? I couldn't find a place we had this information.

src/database/GarbageCollector.ts
bafolts: I originally had this inside `GameLoader` and moved it out. Will move it back in. It was more straight forward to unit test outside of `GameLoader` since I passed in the dependency to `GarbageCollector`.

kberg: Oh, huh.

src/database/GarbageCollector.ts
kberg: Not yet ..... ;)

bafolts: Since this is running every hour I think it may be fine for the time being. A game is only going to end once. If a game happens to end just prior to this script running we would have to pull that game from the database. At that point we would remove it an hour later. If someone then requests the game again we would again pull it from the database. Even once these are unloaded we still have them available in the database. Since this only removes the game instance from javascript memory we can still pull from the database when needed. I am assuming once most games have ended users don't come back often.

kberg: Right, OK, that makes sense.

kberg: Actually, people will look at game results for at least a minute after the game is over, including the act of fetching the game to  deliver it to the front-end for the final time.

Miniscule, yes.

kberg (approved):

src/database/GarbageCollector.ts
bafolts: This can probably be tweaked. Once a game transitions to ended we can start a `setTimeout` to remove it from memory after an hour. This should alleviate this problem and still clear games that have ended since the server started. I can migrate to this. I am curious to see how this impacts the memory load as-is.

PR #3305, opened 2021-05-20
Chairman's choice now limited to one of the two options.

kberg:


src/turmoil/PoliticalAgendas.ts
```diff
@@ -82,11 +82,18 @@ export class PoliticalAgendas {
     // Agendas are static unless it's chosen by a chairperson, in which case
     // defer the selection.
     if (politicalAgendasData.agendaStyle === AgendaStyle.CHAIRMAN && chairman !== 'NEUTRAL') {
-      game.defer(new ChoosePoliticalAgenda(game.getPlayerById(chairman), rulingParty, (nextAgenda) => {
-        turmoil.politicalAgendasData.agendas.set(rulingParty.name, nextAgenda);
```
bafolts, line 86: This was calling to `turmoil.politicalAgendasData.agendas.set` prior to this change. Is this still needed?

kberg: No. The same thing occurs in the lines below. Because the functional change means the player chooses one option and not both, the ChoosePoliticalAgenda deferred action has two callbacks.

bafolts (approved):

PR #3308, opened 2021-05-21
More Robotic Workforce tests, and also block cards without a building…

kberg:
… tag.


src/cards/base/RoboticWorkforce.ts
```diff
@@ -33,6 +33,9 @@ export class RoboticWorkforce extends Card implements IProjectCard {
   }
 
   private isCardApplicable(card: ICard, player: Player): boolean {
+    if (!card.tags.includes(Tags.BUILDING)) {
```
nwai90, line 36: 👍

nwai90 (approved):

PR #3314, opened 2021-05-23
Fully separate player's public and private models.

kberg:
This is still a prototype, it doesn't fully build yet. But I'm calling for comments.

Once this is done, a spectator link is easy and generally safer: it's a model with no private data. (NB that player ids are still in the public data model /shrug.)
The impetus for this change was for adding some data related to tracking whether the host has peeked at everyone's cards; this will be easier to work with, now.

### DIFF
--- a/src/components/App.ts
+++ b/src/components/App.ts
@@ -119,7 +119,7 @@ export const mainAppSettings = {
         if (xhr.status === 200) {
           app.player = xhr.response as PlayerModel;
           app.playerkey++;
-          const playerId = app.player.players[app.player.playersIndex].id;
+          const playerId = app.player.privateModel.id;
 
           if (
             app.player.game.phase === 'end' &&
--- a/src/components/App.ts
+++ b/src/components/App.ts
@@ -119,6 +119,8 @@ export const mainAppSettings = {
         if (xhr.status === 200) {
           app.player = xhr.response as PlayerModel;
           app.playerkey++;
+          const playerId = app.player.players[app.player.playersIndex].id;
+
           if (
             app.player.game.phase === 'end' &&
                         window.location.search.includes('&noredirect') === false
@@ -128,7 +130,7 @@ export const mainAppSettings = {
               window.history.replaceState(
                 xhr.response,
                 `${constants.APP_NAME} - Player`,
-                '/the-end?id=' + app.player.id,
+                '/the-end?id=' + playerId,
               );
             }
           } else {
@@ -137,7 +139,7 @@ export const mainAppSettings = {
               window.history.replaceState(
                 xhr.response,
                 `${constants.APP_NAME} - Game`,
-                '/player?id=' + app.player.id,
+                '/player?id=' + playerId,
               );
             }
           }
--- a/src/components/GameEnd.ts
+++ b/src/components/GameEnd.ts
@@ -1,5 +1,5 @@
 import Vue from 'vue';
-import {PlayerModel, PublicPlayerModel} from '../models/PlayerModel';
+import {PlayerModel, publicModelOf, PublicPlayerModel} from '../models/PlayerModel';
 import {Board} from './Board';
 import {LogPanel} from './LogPanel';
 import {Button} from '../components/common/Button';
@@ -12,6 +12,11 @@ export const GameEnd = Vue.component('game-end', {
   props: {
     player: {
       type: Object as () => PlayerModel,
+    },
+  },
+  computed: {
+    publicPlayer: function(): PublicPlayerModel {
+      return publicModelOf(this.player);
     },
   },
   data: function() {
@@ -74,7 +79,7 @@ export const GameEnd = Vue.component('game-end', {
                             <ul class="game_end_list">
                                 <li v-i18n>Try to win with expansions enabled</li>
                                 <li v-i18n>Try to win before the last generation comes</li>
-                                <li><span v-i18n>Can you get</span> {{ player.victoryPointsBreakdown.total + 10 }}<span v-i18n>+ Victory Points?</span></li>
+                                <li><span v-i18n>Can you get</span> {{ this.publicPlayer.victoryPointsBreakdown.total + 10 }}<span v-i18n>+ Victory Points?</span></li>
                             </ul>
                         </div>
                     </div>
@@ -194,7 +199,7 @@ export const GameEnd = Vue.component('game-end', {
                         :temperature="player.game.temperature"></board>
                 </div>
                 <div class="game_end_block--log game-end-column">
-                  <log-panel :color="player.color" :generation="player.game.generation" :id="player.id" :lastSoloGeneration="player.game.lastSoloGeneration" :players="player.players"></log-panel>
+                  <log-panel :color="this.publicPlayer.color" :generation="player.game.generation" :id="this.publicPlayer.id" :lastSoloGeneration="player.game.lastSoloGeneration" :players="player.players"></log-panel>
                 </div>
               </div>
             </div>
--- a/src/components/WaitingFor.ts
+++ b/src/components/WaitingFor.ts
@@ -16,7 +16,7 @@ import {$t} from '../directives/i18n';
 import {SelectPartyPlayer} from './SelectPartyPlayer';
 import {SelectPartyToSendDelegate} from './SelectPartyToSendDelegate';
 import {PlayerInputModel} from '../models/PlayerInputModel';
-import {PlayerModel} from '../models/PlayerModel';
+import {PlayerModel, publicModelOf} from '../models/PlayerModel';
 import {PreferencesManager} from './PreferencesManager';
 import {SoundManager} from './SoundManager';
 import {SelectColony} from './SelectColony';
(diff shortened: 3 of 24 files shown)


src/components/App.ts
```diff
@@ -119,6 +119,8 @@ export const mainAppSettings = {
         if (xhr.status === 200) {
           app.player = xhr.response as PlayerModel;
           app.playerkey++;
+          const playerId = app.player.players[app.player.playersIndex].id;
```
bafolts, line 122: If the `id` of a player is on the `PublicPlayerModel` this could be a security risk. This would send the `id` of the other players in the game. The `id` should be on the `PrivatePlayerModel`.

bafolts:
This separation makes sense. I added `SpectatorModel` in the hopes of it being used for a separate page for spectators. The work here may make that irrelevant.

src/components/App.ts
kberg: It's been a security risk since day one. ;) It's also an index into most pieces of data passed down. Until we can replace playerid with color throughout there isn't much to do. I recommend that as a completely separate change.

kberg:
> This separation makes sense. I added `SpectatorModel` in the hopes of it being used for a separate page for spectators. The work here may make that irrelevant.

I thought about that, and it probably isn't irrelevant.

src/components/App.ts
bafolts: This was a security risk at one point. We then migrated to using color in most places. The `id` referenced here actually ends up returning `color`.

https://github.com/bafolts/terraforming-mars/blob/main/src/models/ServerModel.ts#L414

I _think_ we put color here as a safety measure while we were changing everything to use color instead of id.

kberg:
All tests now pass, though I haven't executed manual tests. So now it's time for a structural review: is this the right structure? Are things named correctly? Does this still make sense? There are some places where I use `computed` and some where I do not.This is because the compiler doesn't seem to like how `computed` works under some circumstances and the only pattern I could see is that it is related to mixins.

Some changes I could see making:
1. Rename PlayerModel to PlayerViewModel.
2. Move id from PublicPlayerModel to the top object.

Thoughts?

kberg:
Also, @bafolts if you like, I can break this up into pieces.

src/models/ServerModel.ts
```diff
@@ -86,87 +86,36 @@ export class Server {
...
       dealtCorporationCards: this.getCards(player, player.dealtCorporationCards),
       dealtPreludeCards: this.getCards(player, player.dealtPreludeCards),
       dealtProjectCards: this.getCards(player, player.dealtProjectCards),
       draftedCards: this.getCards(player, player.draftedCards, {showNewCost: true}),
-      energy: player.energy,
-      energyProduction: player.getProduction(Resources.ENERGY),
-      fleetSize: player.getFleetSize(),
-      heat: player.heat,
-      heatProduction: player.getProduction(Resources.HEAT),
-      id: player.id,
```
bafolts, line 117: The `PrivatePlayerModel` should have the `id` property. The `id` property is used on the client but it is mapped to `Color` . Most places we are using `id` on the client should be updated to `color`.

Here is where we are mapping to `color` where we want the `id` to be protected.

https://github.com/bafolts/terraforming-mars/blob/main/src/models/ServerModel.ts#L414

Unless changing all the `id` usages on the client which could be `color` to `color` with this change I would create a `PrivatePlayerModel` `id` and a `PublicPlayerModel` `id`. The `PrivatePlayerModel` will always have the player id. The `PublicPlayerModel` will have the logic on line 414 of `ServerModel`. Eventually every `PublicPlayerModel::id` should change to `PublicPlayerModel::color`.

src/components/App.ts
```diff
@@ -128,7 +130,7 @@ export const mainAppSettings = {
               window.history.replaceState(
                 xhr.response,
                 `${constants.APP_NAME} - Player`,
-                '/the-end?id=' + app.player.id,
```
bafolts, line 131: This will need to be `app.privatePlayer.id`. The current `app.player.id` is the equivalent of what I think `app.privatePlayer.id` should become. This is always the private player id that other players shouldn't be able to know.

src/components/GameEnd.ts
```diff
@@ -194,7 +199,7 @@ export const GameEnd = Vue.component('game-end', {
                         :temperature="player.game.temperature"></board>
                 </div>
                 <div class="game_end_block--log game-end-column">
-                  <log-panel :color="player.color" :generation="player.game.generation" :id="player.id" :lastSoloGeneration="player.game.lastSoloGeneration" :players="player.players"></log-panel>
+                  <log-panel :color="this.publicPlayer.color" :generation="player.game.generation" :id="this.publicPlayer.id" :lastSoloGeneration="player.game.lastSoloGeneration" :players="player.players"></log-panel>
```
bafolts, line 202: We don't have a linting rule in place but I remember vue-js logging at one point that `this` was discouraged from within templates. There seems to be an eslint rule but I am not sure if we are running these.

https://eslint.vuejs.org/rules/this-in-template.html

bafolts, line 202: The `id` attribute passed to `log-panel` I think will need to be `privatePlayer.id`. This may work because we happen to set the public player id to the actual id when the game has ended.

https://github.com/bafolts/terraforming-mars/blob/main/src/models/ServerModel.ts#L414

src/components/WaitingFor.ts
```diff
@@ -145,7 +145,7 @@ export const WaitingFor = Vue.component('waiting-for', {
       }
 
       root.isServerSideRequestInProgress = true;
-      xhr.open('POST', '/player/input?id=' + this.player.id);
+      xhr.open('POST', '/player/input?id=' + publicModelOf(this.player).id);
```
bafolts, line 148: As a rule of thumb I believe any `this.player.id` will need to be mapped to `this.player.privateModel.id` with this change.

src/components/overview/PlayersOverview.ts
```diff
@@ -109,7 +112,7 @@ export const PlayersOverview = Vue.component('players-overview', {
             <overview-settings />
             <div class="other_player" v-if="player.players.length > 1">
                 <div v-for="(otherPlayer, index) in getPlayersInOrder()" :key="otherPlayer.id">
-                    <other-player v-if="otherPlayer.id !== player.id" :player="otherPlayer" :playerIndex="index"/>
+                    <other-player v-if="otherPlayer.id !== publicPlayer().id" :player="otherPlayer" :playerIndex="index"/>
```
bafolts, line 115: This never migrated to using `color` but this check was supposed to be migrated to use `otherPlayer.color !== player.color`.

src/models/PlayerModel.ts
```diff
@@ -61,8 +60,21 @@ export interface PlayerModel extends PublicPlayerModel {
...
+// public information about all the players,
+// your private information, and your index in the public players array.
+export interface PlayerModel {
+  game: GameModel;
+  players: Array<PublicPlayerModel>;
+  privateModel: PrivatePlayerModel;
+  playersIndex: number,
+}
+
+export function publicModelOf(playerModel: PlayerModel): PublicPlayerModel {
```
bafolts, line 78: Technically this function could return `undefined`. If `playerModel.playersIndex` were out of bounds of the array this function will return `undefined`. Throwing an error in that case may help with debugging.

bafolts:
This change looks great. Hopefully I was able to get the point across about how some `player.id` should actually be `player.color`. When the security risk was spotted a few years back we changed the now public `player.id` to send back color from the server. We never went through and updated every `player.id` to `player.color` though. Unfortunately I think this change will have to check on every `player.id` to see if it should be the public or private player. I think I left comments on every instance that should be the private player id.

src/components/GameEnd.ts
kberg: Not sure. The problem is that if I don't use `this.` the second compile stage for Vue doesn't work.

src/models/PlayerModel.ts
kberg: Once it returns undefined everything else gets more complicated.

kberg:
I made another pass which should address all the 'id' issues. I know because I deleted the id field, and only reintroduced it to support `GameEnd.ts` where all player IDs will be 'invalid' until game end.

There are currently two broken tests but I'm going to sleep; they won't be hard to fix with a little focus and attention.

src/models/ServerModel.ts
```diff
@@ -416,91 +365,49 @@ export class Server {
...
+      citiesCount: player.getCitiesCount(),
+      coloniesCount: player.getColoniesCount(),
+      color: player.color,
+      corporationCard: this.getCorporationCard(player),
+      energy: player.energy,
+      energyProduction: player.getProduction(Resources.ENERGY),
+      fleetSize: player.getFleetSize(),
+      heat: player.heat,
+      heatProduction: player.getProduction(Resources.HEAT),
+      id: player.game.phase === Phase.END ? player.id : 'invalid',
```
bafolts, line 385: I don't think `PublicPlayerModel` should expose `id`. I think you have successfully ported everything to using `color` so you should be able to remove this property. A comment above this indicating why it is still around may help.

bafolts (approved):
This all looks right now. Finally finished the migration to always use `color` publicly!

kberg:
Glad I didn't use this one. It took longer, but this is correct. So, correct is good.

PR #3319, opened 2021-05-26
Remove check that is not needed

bafolts:
Was looking into when we change game phase to `Phase.END`. Spotted an unnecessary check for `Phase.END` left over from a migration.


kberg (approved):
Yeah I saw that one but didn't bother. Good catch.

PR #3327, opened 2021-05-29
try removing ended games again

bafolts:
The first try with garbage collection I ended up blocking the main thread iterating through every game. The strategy this time is to set a timeout when the game transitions to the end phase. This will avoid having to iterate through thousands of games at once. It will also allow for a finished game to remain in javascript memory for an hour.

The goal of this ticket is to cut down on the amount of memory being used on the heroku instance. With no process in place today to every free memory we eventually use enough that heroku tries to restart the dyno. We have seen in the past that when it restarts the dyno it sometimes fails to start.

### DIFF
--- a/src/database/GameLoader.ts
+++ b/src/database/GameLoader.ts
@@ -116,6 +116,10 @@ export class GameLoader implements IGameLoader {
     this.getByParticipantId(spectatorId, cb);
   }
 
+  public remove(gameId: GameId): void {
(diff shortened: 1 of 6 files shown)


alrusdi (approved):

bafolts:
Last time this we merged there were CPU spikes. I will monitor this throughout the day. The server has rebooted everyday the past week due to memory. Hoping this starts cutting down on memory throughout the day.

PR #3333, opened 2021-06-02
use a Pool instead of one Client for postgres

### LINKED ISSUE #3331: Some responses take 30 seconds to respond.
Finally captured one of the 30 second timeouts seen a few times per day on the terraforming-mars.herokuapp.com.

```
2021-06-01T00:48:23.259421+00:00 heroku[router]: at=error code=H12 desc="Request timeout" method=POST path="/player/input?id=p9a90c59fc1f0" host=terraforming-mars.herokuapp.com request_id=6cdba4de-2e8a-4ac6-a931-647fe35a2753 fwd="204.14.36.17" dyno=web.1 connect=0ms service=30004ms status=503 bytes=0 protocol=https
```

Seems to be when a player was attempting to take some action. Will close this issue in 10 days once this data isn't available in the database. Will try to post more details. Once a player hits this issue they will see the error modal. After this they then usually try the action again. This leads to multiple 30 second responses in a row.

The CPU is not pegged and other requests make it through during this time. It is surprising that other requests make it through during this time. There are only a few places we have asynchronous operations. One of these asynchronous operations must time out. I am assuming this is some operation with the database timing out.

bafolts:
This should close #3331 . I believe eventually the one `Client` we use for Postgres throws an error and we never reconnect. Using a `Pool` should always open a new connection for us when needed.

Tested this locally with the terraforming-mars.herokuapp.com postgres connection string.

### DIFF
--- a/src/database/PostgreSQL.ts
+++ b/src/database/PostgreSQL.ts
@@ -3,10 +3,10 @@ import {Game, GameId, GameOptions, Score} from '../Game';
 import {IGameData} from './IDatabase';
 import {SerializedGame} from '../SerializedGame';
 
-import {Client, ClientConfig, QueryResult} from 'pg';
+import {Pool, ClientConfig, QueryResult} from 'pg';
 
 export class PostgreSQL implements IDatabase {
-  private client: Client;
+  private client: Pool;
 
   constructor() {
     const config: ClientConfig = {
@@ -18,8 +18,7 @@ export class PostgreSQL implements IDatabase {
         rejectUnauthorized: false,
       };
     }
-    this.client = new Client(config);
-    this.client.connect();
+    this.client = new Pool(config);
     this.client.query('CREATE TABLE IF NOT EXISTS games(game_id varchar, players integer, save_id integer, game text, status text default \'running\', created_time timestamp default now(), PRIMARY KEY (game_id, save_id))', (err) => {
       if (err) {
         throw err;


src/database/PostgreSQL.ts
```diff
@@ -18,8 +18,7 @@ export class PostgreSQL implements IDatabase {
         rejectUnauthorized: false,
       };
     }
-    this.client = new Client(config);
-    this.client.connect();
+    this.client = new Pool(config);
```
kberg, line 21: Is this config still correct?

https://node-postgres.com/api/pool suggests that the config has fewer fields, but you tell me.

alrusdi (approved):

bafolts:
I believe this is still correct. The documentation is hidden in comments regarding this.

```
// all valid client config options are also valid here
// in addition here are the pool specific configuration parameters:
```

When the `Pool` creates `Client` instances it uses the same configuration properties we used to use.

kberg:
> I believe this is still correct. The documentation is hidden in comments regarding this.
> 
> ```
> // all valid client config options are also valid here
> // in addition here are the pool specific configuration parameters:
> ```
> 
> When the `Pool` creates `Client` instances it uses the same configuration properties we used to use.

Is there a separate postgresql instance with https that can verify this config parameter? If not, give this a little babysitting after you push it to make sure things don't break.

bafolts:
I plan to babysit this as well as reverting the other PR. When testing locally I used the terraforming-mars.herokuapp.com database connection string and saw every game load from database. I then went into a random game and saw the client load. If I don't see the 30 second timeouts today which I assume are database related may then try reintroducing the PR to try garbage collecting. The good news with this app is that once it starts up access to the database is not critical. Should know once the server has restarted whether there is a problem. This is another example of this app growing larger than ever expected. When I first put in `Client` I didn't give it much thought and it seemed to work fine.

PR #3338, opened 2021-06-03
Exclude landclaimed Ares tiles

### LINKED ISSUE #3337: Landclaimed Ares tiles should not count toward milestones and awards
image

vsrisuknimit:
Exclude Landclaimed Ares tiles from land-based awards and milestones.

This PR closes #3337 .

### DIFF
--- a/src/TileType.ts
+++ b/src/TileType.ts
@@ -90,6 +90,11 @@ export namespace TileType {
   }
 }
 
+// Ares Tiles handling
+
+export const HAZARD_TILES = new Set([TileType.DUST_STORM_MILD, TileType.DUST_STORM_SEVERE, TileType.EROSION_MILD, TileType.EROSION_SEVERE]);
+export const OCEAN_UPGRADE_TILES = new Set([TileType.OCEAN_CITY, TileType.OCEAN_FARM, TileType.OCEAN_SANCTUARY]);
+
 export function isAresTile(tile: TileType): boolean {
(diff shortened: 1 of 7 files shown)


nwai90 (approved):
LGTM. I would also add a small unit test for it (to assert that a Land-claimed hazard tile does not count towards the milestone/award score)

nwai90 (approved):

src/TileType.ts
```diff
@@ -89,3 +89,7 @@ export namespace TileType {
     return TO_STRING_MAP.get(tileType) || `(unnamed tile, id ${tileType})`;
   }
 }
+
+export function isAresTile(tile: TileType): boolean {
+  return [TileType.EROSION_MILD, TileType.EROSION_SEVERE, TileType.DUST_STORM_MILD, TileType.DUST_STORM_SEVERE].includes(tile);
```
bafolts, line 94: This array is equivalent to the `HAZARD_TILES` array exported from `AresHandler`. Doesn't seem anything is using the export from `AresHandler`. Could be a good opportunity import and use it here. Using a `Set` to store this would also make sense to me and allow for using `Set.prototype.has` which will be slightly more efficient.

bafolts (approved):
Using `Array` is fine. Want to allow time to consider using `Set` before I merge. I think the optimization would be worth it. In `AresHandler` I would make:

```
export const HAZARD_TILES = new Set([TileType.DUST_STORM_MILD....
```

Then would leverage this from the `TileType` function.

vsrisuknimit:
@bafolts that will cause circular dependency. I moved things around a bit.

PR #3341, opened 2021-06-04
Fix prettier

Argeento:
Fix prettier to works along side with eslint
Format code base with prettier


kberg:
In short, first, pulling off a 1,800 PR just for formatting may be too much to ask. It makes merging pulling from the main repo harder, with no real value. I like well formatted code, no question, but I do not think this is the way to do it.

Argeento:
How else would You like to format the code?
Hard reset to initial commit and push all the files?
New repo?

If you want I can search for prettier config that modify as few files as possible. Bu still it will be hundreds of them. That only shows how mess is the code-base right now.

For me making prettier works is first and obligatory step to go further. I want to write a script to convert all ts components to SFC. 

Why this changes are so important? Yes, formatting code and rewrite vue components are no real value for product in the first place. But what it does is improve development experience. 

I am vue dev. Me and my friends (as well as community) have plenty of ideas to enhance UX. I look to the code. I see these.. things... JS in HTML in TS and I want to scream, run out and never come back again.

I know, I'm coming from nowhere and turning upside down your repo, but these are necessary steps that should be done long time ago

kberg:
I appreciate the value of a well-formatted codebase. I've done exactly this kind of thing on even larger codebases. But I think this change just isn't going to fly.

> How else would You like to format the code?

Well, at the moment we aren't having meaningful formatting problems, and although in the past devs have wanted to rely on a standard more than they I wouldn't do this at all. Bit by bit when I am actively working on a component, I might reformat it, but only when the formatting does not get in the way of understanding the change itself.

> If you want I can search for prettier config that modify as few files as possible. Bu still it will be hundreds of them. That only shows how mess is the code-base right now.

I'm not saying the code isn't messy. It is, but not because it is poorly formatted.

> For me making prettier works is first and obligatory step to go further. I want to write a script to convert all ts components to SFC.

Is reformatting a file a mandatory step before replacing it with another file? I haven't ever seen that as a useful case before.

I would love for you to be a contributor, but you have never contributed to this repository before, and you're asking us to start by having all the code reformatted. We, as reviewers, also have limited time on what we want to focus on. If you wanted to start by providing something of more concrete value, such as showing how an SFC would work, that would help a lot. I would love to see this be easier to read, maintain and test, and I believe SFC could be a good step in that direction.

Finally, keep in mind that dozens of people have used their unpaid peresonal time to build this because they love it. There are better ways of introducing yourself by calling this code a mess, or that it makes you want to scream and run away. I'm sure you didn't mean it literally, of course, but it's just not a way we talk about things here.

kberg:
I'm going to close this because the general PR is just much too large, and so this won't make it in. 

There is a small piece in your PR where you fixed some broken HTML and that on its own would be a very welcome commit / PR. I like small PRs; they are easy to read and understand. Would you please provide that as a separate PR?

Closing the PR doesn't mean I'm ending the discussion. If you have more you want to add please do.

Argeento:
Lets beign with SFC then :rocket:

PR #3343, opened 2021-06-05
Refactor build process

Argeento:
I got rid of `compile-vue-templates.js` and moved vue app generation to webpack

Moreover, small build improvements:
- In the build command, now the linting is the first task
- Fixed double makeJson run
- Added nvmrc


alrusdi (commented):
Vue templates compilation was added to check types inside the templates. Can you clarify of how current approach does it?

Argeento:
What do you mean @alrusdi ? Inside these strings?
image

Argeento:
Yea, I check this out. Yes, inside these strings. Let me fix that.

Argeento:
For now, there is no type checking in templates.

This build modification is require to SFC

While all components will be rewrite to SFC, we will use [vtc](https://github.com/Yuyz0112/vue-type-check) and the types saftiness will return

image

webpack.config.js
```diff
@@ -1,33 +1,18 @@
...
 module.exports = {
-  devtool: 'source-map',
-  mode: process.env.NODE_ENV === 'production' ? 'production' : 'development',
-  entry: [
-    './build/src/script.js',
-  ],
-  plugins: [
-    new CompressionPlugin(),
-    new CompressionPlugin({
-      algorithm: 'brotliCompress',
```
bafolts, line 16: This brotli compression provides production with significantly smaller file for the built javascript. On the terraforming-mars.herokuapp.com server we are probably pushing our luck with bandwidth usage. AFAIK brotli compression is not offered by default without it included here.

package.json
```diff
@@ -36,14 +35,15 @@
...
+    "nyc": "^14.1.1",
+    "ts-loader": "^9.2.2"
   },
   "nyc": {
     "extends": "@istanbuljs/nyc-config-typescript"
   },
   "scripts": {
-    "build": "npm run compile && npm run templates && npm run webpack && npm run lint",
-    "prebuild": "npm run makeJson && npm run makeCss",
+    "build": "npm run lint && npm run compile && npm run webpack",
```
bafolts, line 45: The linting was put last with the thought process that if the linting were to fail that isn't always something worth preventing a build locally. Often times when I am working on a change my code will compile but will have linting issues. I would prefer to still have lint  as the final step that runs for the build script.

.nvmrc
```diff
@@ -0,0 +1 @@
+v14.16.1
```
bafolts, line 1: We don't specify this version of node in `package.json`. When heroku starts the service it will use the value listed in `package.json`. If introducing a separate `.nvmrc` this value should also be added to `package.json` so heroku is using the version developers are using. If making this change using a separate PR would be helpful and could help if there were a need to roll back. I don't think there would be any issues with this change but if heroku failed to start for whatever reason after introducing a node change then may have to roll back.

package.json
```diff
@@ -36,14 +35,15 @@
...
-    "nyc": "^14.1.1"
+    "nyc": "^14.1.1",
+    "ts-loader": "^9.2.2"
   },
   "nyc": {
     "extends": "@istanbuljs/nyc-config-typescript"
   },
   "scripts": {
-    "build": "npm run compile && npm run templates && npm run webpack && npm run lint",
-    "prebuild": "npm run makeJson && npm run makeCss",
```
bafolts, line 46: This change to avoid running `makeJson` twice will be more efficient. Using a separate pull request for that change would be helpful as it will increase the odds of that change going in.

webpack.config.js
Argeento: I compare bundle size and there are no differernce between them. 

If you want oprimize bundle, start with async translation for each language. 
Target modern JS, and you will cut the size in the half.
Thats only the beginning. I can reduce the size to <1Mb

giziped bundle analyzis:
image

Run different compression algs against bundle is not the way to optymize it :see_no_evil:

.nvmrc
Argeento: And what version heroku is now using? I cant find it

package.json
Argeento: > Often times when I am working on a change my code will compile but will have linting issues. 

If these are formatting issues, please use prettier, and your IDE will automatically fix them
If these are logical issues, are you sure you want to run this build?

If you need separate build options for dev, why there is no buld-dev command?

package.json
Argeento: Not too much efficient improvement. `makeJson` takes less than a fraction of a second. Compered to building an entire app this is literally nothing

webpack.config.js
Argeento: Oh. I know why this plugin is "significantly" better. Becouse [you dont specify NODE_ENV, and run build in dev mode](https://github.com/bafolts/terraforming-mars/blob/main/webpack.config.js#L9)

It is fixed in this PR

Argeento:
Ok - I am ready with [SFC](https://github.com/Argeento/terraforming-mars/tree/sfc), but waiting for this PR

.nvmrc
bafolts: From [heroku](https://devcenter.heroku.com/articles/nodejs-support#specifying-a-node-js-version) 

> If a Node version isn’t specified in the engine, the 14.x release will be used.

I am not sure which value _the_ 14.x release correlates to.

package.json
bafolts: Is this change needed for the SPC change? This seems like a distraction for SPC.

Argeento: No problem, we can fix it later

bafolts:
> Ok - I am ready with SFC, but waiting for this PR

I am excited to have SFC. I don't think there is some way to introduce one component at a time but that would help with putting in this change. Is there a way we could move one component at a time? The change will be difficult to review with every component going in simultaneously and we could miss something. I will look at this PR further tonight.

.nvmrc
Argeento: OK. I added the node version to package.json too

package.json
```diff
@@ -2,14 +2,16 @@
   "name": "terraforming-mars",
   "version": "1.0.0",
   "description": "Terraforming Mars Game",
+  "engines": {
+    "node": "14.x"
```
bafolts, line 6: This is a good change as it will ensure we control which version of node heroku runs without relying on the default. I happened to have node v10.x locally so this change could introduce a hurdle for contributors. I believe it only shows up as a warning and doesn't block anything.

webpack.config.js
bafolts:
```
-rw-rw-r--  1 algorithm99 algorithm99 2.5M Jun  7 19:14 main.js
-rw-rw-r--  1 algorithm99 algorithm99 368K Jun  7 19:14 main.js.br
-rw-rw-r--  1 algorithm99 algorithm99 540K Jun  7 19:14 main.js.gz
```

The brotli compression is 32% smaller than gzip compression. There are other rooms for improvement but for the time being we send a 368k to some users. With this single page app this payload loads most of the app. We have thousands of games played with users all over the world from a handful of hobby plan heroku instances. I think this saving in bytes goes a long way. I don't think this has to be changed with this PR as it is unrelated to the SFC change. Can you change this part back?

package.json
Argeento: All contributors should use the same version of node. 
We know that 14.x is tested, (by heroku builds), so I added this version.
Please use [nvm](https://github.com/nvm-sh/nvm)
14.x means the newest version of 14

webpack.config.js
Argeento: I don't know what these files mean.
But I do know what are the real values in the browser:

Current build on heroku: 2.6MB
image

Build without brotli plugin: 2.6MB
image

If now you can explain me why this plugin is in use, then I will revert it.

bafolts: This must have broken as we were migrating to use _routes_. We should be sending the `br` file [here](https://github.com/bafolts/terraforming-mars/blob/main/src/routes/ServeAsset.ts#L110). This is a bug we have not noticed. The 32% was there at some point.

bafolts: Can you revert this even if it has a bug now? I can make an issue to fix the bug separately.

Argeento: :+1:

Argeento:
I also delete `mode: production` in webpack. Production is set by default

bafolts (approved):

bafolts:
@Argeento I think with the next change we can convert one component at a time to use the vuejs extension. This will be easier to review than having to go through every component at once.

Argeento:
Lets argue in the next PR :)

bafolts:
There really is going to be no ideal way to merge the single file components. I don't think anyone has large changes on the way. Anyone with some front end change will enter "merge hell" once the single file component PR shows up. Mostly don't want to miss anything when moving each template. I may write a script to verify the template doesn't change when it moves so we don't reintroduce some bug.

bafolts:
Having the single file components will make it easier for anyone developing the front end. Glad to see us heading that way.

PR #3351, opened 2021-06-08
SFC

Argeento:
Rewrite vue `.ts` components to [Single File Components](https://vuejs.org/v2/guide/single-file-components.html)


Argeento:
In every file I run this steps:
 1. Determine if the file is vue component (by searching `Vue.component` string), if so:
 2. Find component name by searching export line `/export const (.*?) = Vue./`
 3. Change `export` to `default export Vue.extends`
 4. Add `name` key to component settings
 5. Search for template string, and if found, move it to the `<template> tag to the top of the document
 6. Move rest of the code to the `<script>` tag
 7. Add empty `<style>` tag to the bottom
 
Moreover, SFC is requiring to have only one default export, so the script also fixed all the imports in all components. You don't have to check them. If project runs, then everything with import/exports is OK.

Every change in templates I will comment in this PR

src/components/OtherPlayer.vue
```diff
@@ -1,13 +1,48 @@
...
-import {StackedCards} from './StackedCards';
+import StackedCards from './StackedCards.vue';
 import {PlayerMixin} from './PlayerMixin';
 import {PlayerModel} from '../models/PlayerModel';
-import {hidePlayerData} from './overview/PlayerStatus';
 import {mainAppSettings} from './App';
-import {Card} from './card/Card';
+import Card from './card/Card.vue';
+
+const hidePlayerData = (root: typeof mainAppSettings.methods, playerIndex: number) => {
```
Argeento, line 40: Here is the manual change. This function was imported from another component. SFC can only have one export, so I added it here

src/components/SelectCard.vue
```diff
@@ -17,12 +37,13 @@ interface SelectCardModel {
   warning: string | Message | undefined;
 }
 
-export interface OwnerModel {
+interface OwnerModel {
```
Argeento, line 40: Next manual change. I had to implement this interface once again for this component

src/components/WaitingFor.vue
```diff
@@ -126,15 +129,19 @@ export const WaitingFor = Vue.component('waiting-for', {
     },
   },
   render: function(createElement) {
+    console.log('render function in WaitingFor is fired!');
```
Argeento, line 132: Oh wait.. I left some debug code...

Argeento:
My recommendation: Please DO NOT CHECK THIS BY HAND. It is waste of time. Run the tests, play some games

src/components/App.ts
```diff
@@ -120,6 +120,7 @@ export const mainAppSettings = {
           app.player = xhr.response as PlayerModel;
           app.playerkey++;
           if (
+            // @ts-ignore
```
Argeento, line 123: In App.ts are some types error. For now I only ignore them, but the code itself didn't change

Argeento:
Here is [the script](https://gist.github.com/Argeento/1f54e856d91c0f56023ca93bcae6d0e6) that rewrites components

bafolts:
> My recommendation: Please DO NOT CHECK THIS BY HAND. It is waste of time. Run the tests, play some games

I am going to write my own script to check the moving of all the templates. Prior to merging will run the script.

kberg (changes requested):
As I had feared, this includes moving a lot of contents around, making it hard to verify if any accidental changes made it in. This is why I asked to start with one component.

kberg:
> My recommendation: Please DO NOT CHECK THIS BY HAND. It is waste of time. Run the tests, play some games

Thanks for your recommendation, but I think I laid ou t my reasons for manual verification on Discord. Add to that that we do not have enough tests to not manually verify.

bafolts:
> As I had feared, this includes moving a lot of contents around, making it hard to verify if any accidental changes made it in. This is why I asked to start with one component.

I had mentioned this in #3343 as well. I agree that starting with one component will help with this transition and spot any issues early before every component is converted.

package.json
```diff
@@ -59,10 +67,12 @@
     "makeCss": "lessc src/styles/common.less build/styles.css",
     "makeJson": "node make_static_json.js",
     "start": "node build/src/server.js",
-    "test": "mocha --file build/tests/utils/Vue.js --recursive build/tests",
+    "test": "rm -rf build/tests/components/ && mocha --file build/tests/utils/Vue.js --recursive build/tests",
```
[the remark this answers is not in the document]
bafolts: This separation of tests make sense. I vaguely remember some issue with the `exclude` option with `mocha`. Did the `exclude` option not work? Most of our unit tests are for the back end game server classes and not the front end components.

src/components/App.ts
bafolts: I am curious what these type errors are. Ignoring these errors could mean in the future we end up ignoring a real error. Or we are ignoring a real error now.

package.json
```diff
@@ -59,10 +67,12 @@
...
+    "test:components": "mochapack --mode development --webpack-config webpack.config.js --require tests/components/setup.ts tests/components/*.spec.ts",
     "pretest": "npm run prebuild && tsc --build tests/tsconfig.json",
     "watch": "webpack --watch --mode=development",
-    "webpack": "webpack --no-stats"
+    "webpack": "webpack --no-stats",
+    "check-templates": "vtc --workspace . --srcDir ./src/components/ --onlyTemplate"
```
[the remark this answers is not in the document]
bafolts: This `check-templates` command is great. We should turn this on and then can introduce one template at a time. This would allow us to address the issues as we went. Looking quickly at the list most of the issues seem like things we forgot to remove during migrations.

One example:

```
Error in file:///home/algorithm99/Projects/terraforming-mars/src/components/GameSetupDetail.vue
62:37 Property 'undoOption' does not exist on type 'GameOptionsModel'.
  60 |               <div v-if="gameOptions.showTimers" class="game-config timer" v-i18n>timer</div>
  61 |               <div v-if="gameOptions.showOtherPlayersVP" class="game-config realtime-vp" v-i18n>real-time vp</div>
> 62 |               <div v-if="gameOptions.undoOption" class="game-config undo" v-i18n>undo</div>
     |                                      ^^^^^^^^^^
  63 |             </li>
```

src/components/App.ts
```diff
@@ -1,16 +1,16 @@
...
-import {DebugUI} from './DebugUI';
-import {SimpleGameModel} from '../models/SimpleGameModel';
-import {Help} from './help/Help';
+import GameEnd from './GameEnd.vue';
+import CreateGameForm from './create/CreateGameForm.vue';
+import GameHome from './GameHome.vue';
+import GamesOverview from './GamesOverview.vue';
+import PlayerHome from './PlayerHome.vue';
+import SpectatorHome from './SpectatorHome.vue';
+import PlayerModel from '../models/PlayerModel.vue';
```
bafolts, line 7: The reason for the `ts-ignore` and the compile errors are due to importing `PlayerMode.vue`. This should be imported as:

```
import {PlayerModel} from '../models/PlayerModel';
```

This will avoid the need for `ts-ignore`.

Argeento: Fixed

bafolts:
I am trying to put together a solution that will allow us to pull in a handful of these templates at a time. I don't think this is straight forward since once we use the `.vue` extension we need webpack compiling some of our files and not `tsconfig`.

tests/LogHelper.spec.ts
```diff
@@ -1,11 +1,11 @@
 
 import {expect} from 'chai';
-import {Algae} from '../../src/cards/base/Algae';
```
bafolts, line 3: This change can go in separate. This seems to pull in a reference to the database which prevents the front end tests from running.

package.json
```diff
@@ -36,11 +35,20 @@
...
+    "less-loader": "^9.0.0",
     "mocha": "^8.3.2",
+    "mochapack": "^2.1.2",
+    "mutation-observer": "^1.0.3",
     "nyc": "^14.1.1",
     "ts-loader": "^9.2.2",
+    "vue-loader": "^15.9.7",
+    "vue-style-loader": "^4.1.3",
+    "vue-template-compiler": "^2.6.10",
+    "vue-type-check": "^1.1.0",
```
bafolts, line 51: The `vue-type-check` has a peer dependency of typescript 3. When I try to install without the package-lock file I get an error.

```
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! 
npm ERR! While resolving: terraforming-mars@1.0.0
npm ERR! Found: typescript@4.3.2
npm ERR! node_modules/typescript
npm ERR!   typescript@"^4.1.2" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer typescript@"^3.9.6" from vue-type-check@1.1.0
npm ERR! node_modules/vue-type-check
npm ERR!   dev vue-type-check@"^1.1.0" from the root project
```

How did you avoid this error?

kberg:
This can be closed, right? Is there anything else in this which we need?

bafolts:
This work is slowly getting done a few components at a time. A lot of the major concepts have already been merged.

PR #3355, opened 2021-06-10
Reduce allowable heat when using Helion with Local Heat Trapping

### LINKED ISSUE #3326: Bug with Helion and "Local Heat Trapping"
I believe this should not work: If playing "Local Heat Trapping" as Helion when only having 5 heat remaining, and paying for the card (1 MC) with heat, it still works (and thus converts 4 instead of 5 heat to 4 plants). I would think that you'd need to be able to pay 1 MC/heat+5 heat to play Local Heat Trapping.

kberg:
Fixes #3326.

This PR has 3 commits; the first moves `reserveUnits` to a thing called `ProjectCard` (there seems to be some difference between what an `IProjectCard` and a `Card` can do. A little pecuilar, but OK. To make that work I had to fix a brittle test in L.T.F. Privileges, OK.

The second one moves the `Local Heat Trapping` special reserve behavior to a `reserveUnits` and ties everything else up. To make the tests meaningful again, I replaced card.canPlay() with player.getPlayableCards(), which takes into account things like money + units in reserve. This PR also fixes an issue with Playwrights that is not compatible with the Moon Expansion.

The third commit updates the UI to respect reserved heat units.

### DIFF
--- a/src/components/SelectHowToPayForProjectCard.ts
+++ b/src/components/SelectHowToPayForProjectCard.ts
@@ -192,7 +192,7 @@ export const SelectHowToPayForProjectCard = Vue.component('select-how-to-pay-for
       }
     },
     canUseHeat: function(): boolean {
-      return (this.playerinput.canUseHeat ?? true) && this.player.heat > 0;
+      return this.playerinput.canUseHeat === true && this.player.heat > 0;
     },
     canUseSteel: function() {
       if (this.card !== undefined && this.available.steel > 0) {
--- a/src/components/SelectHowToPayForProjectCard.ts
+++ b/src/components/SelectHowToPayForProjectCard.ts
@@ -15,8 +15,7 @@ interface SelectHowToPayForProjectCardModel {
   microbes: number;
   floaters: number;
   warning: string | undefined;
-  availableSteel: number;
-  availableTitanium: number;
+  available: Units;
 }
 
 import {HowToPay} from '../inputs/HowToPay';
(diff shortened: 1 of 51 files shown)


kberg:
I see now that this is even more complicated. ProjectCard is ignored in LocalHeatTrapping, which in some ways is fine since it does a `spendHeat()` at the bottom, but that's not great.

However, what it reveals is that Stormcraft Incorporated needs to be included in the calculus, too.

bafolts:
I haven't looked into the Stormcraft issue but the rest looks right so far.

bafolts:
We may need a `getSpendableHeat` function on `Player` that returns how much heat you could spend.

src/components/SelectHowToPayForProjectCard.ts
```diff
@@ -191,19 +191,19 @@ export const SelectHowToPayForProjectCard = Vue.component('select-how-to-pay-for
         this.megaCredits -= saveOverSpendingUnits(this.megaCredits, 1);
       }
     },
-    canUseHeat: function() {
-      return this.playerinput.canUseHeat && this.player.heat > 0;
+    canUseHeat: function(): boolean {
+      return (this.playerinput.canUseHeat ?? true) && this.player.heat > 0;
```
bafolts, line 195: If `this.playerinput.canUseHeat` is `undefined` I would expect this to use `false` and not `true`. Should this be:

`this.playerinput.canUseHeat ?? false` ?

bafolts: Or `this.playerinput.canUseHeat === true && this.player.heat > 0`

kberg: > If `this.playerinput.canUseHeat` is `undefined` I would expect this to use `false` and not `true`. Should this be:
> 
> `this.playerinput.canUseHeat ?? false` ?

Yipe! Good catch.

kberg: Fixed.

bafolts (approved):

PR #3356, opened 2021-06-10
move CardTag to a single file component

bafolts:
This starts small with moving the `CardTag` component to a single file component. Took many parts of this from #3351 . This also updates the build to run the `test:components`. Tested this by running a game and see the tags on cards still show up.


Argeento:
If you do it one by one, You can also move `.less` code to the component. Scoped styled are awesome! No more BEM and fighting with specificity

Argeento:
Moreover, can you now use prettier to format files one by one?

bafolts:
When I didn't use `const CardType = require('./cardType.vue')` typescript couldn't find the module. I think even now typescript may not find the module and it's type but it is unimportant as we don't use these vue modules for their types. Once every module is converted it wont matter. Should only have to use this `require` approach until every component is updated.

package.json
```diff
@@ -52,17 +56,18 @@
...
     "precover": "tsc --build tests/tsconfig.json",
     "fix": "eslint --fix src tests --ext ts",
     "lint": "eslint --cache src tests --ext ts",
     "makeCss": "lessc src/styles/common.less build/styles.css",
     "makeJson": "node make_static_json.js",
     "start": "node build/src/server.js",
-    "test": "mocha --file build/tests/utils/Vue.js --recursive build/tests",
-    "pretest": "npm run prebuild && tsc --build tests/tsconfig.json",
+    "test": "rm -rf build/tests/components/ && mocha --require build/tests/utils/setup.js --recursive build/tests",
+    "test:components": "npm run compile && mochapack --mode development --webpack-config webpack.config.js --require tests/components/setup.ts tests/components/*.spec.ts",
```
kberg, line 67: Much of this probably belongs in `pretest:components` I bet.

package.json
```diff
@@ -52,17 +56,18 @@
...
+    "cover": "nyc mocha --file build/tests/utils/setup.js --recursive build/tests",
     "precover": "tsc --build tests/tsconfig.json",
     "fix": "eslint --fix src tests --ext ts",
     "lint": "eslint --cache src tests --ext ts",
     "makeCss": "lessc src/styles/common.less build/styles.css",
     "makeJson": "node make_static_json.js",
     "start": "node build/src/server.js",
-    "test": "mocha --file build/tests/utils/Vue.js --recursive build/tests",
-    "pretest": "npm run prebuild && tsc --build tests/tsconfig.json",
+    "test": "rm -rf build/tests/components/ && mocha --require build/tests/utils/setup.js --recursive build/tests",
```
kberg, line 66: Is this double-testing everything in tests/components?

kberg (commented):
This looks great modulo the pending comments.

src/components/card/CardTags.ts
```diff
@@ -1,12 +1,12 @@
 import Vue from 'vue';
-import {CardTag} from './CardTag';
+const CardTag = require('./CardTag.vue');
```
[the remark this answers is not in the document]
kberg: I just tried it with a new widget I'm building and it doesn't work for me, either. (Nor does vscode like it.)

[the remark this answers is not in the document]
bafolts: This works for me with the included `vue-shims.d.ts`.

package.json
bafolts: This ends up testing only the back end code. The components are now only tested with the `test:components` script.

kberg:
I'm not sure this is working yet.

```
$ npm run start

> terraforming-mars@1.0.0 start
> node build/src/server.js

[2021-06-10 20:27:06 EDT] Starting server on port 8080
[2021-06-10 20:27:06 EDT] Starting local database at /home/kberg/git/terraforming-mars/db/files
[2021-06-10 20:27:06 EDT] version 0.X
[2021-06-10 20:27:06 EDT] 
The secret serverId for this server is bacece8d27fb. Use it to access the following administrative routes:

[2021-06-10 20:27:06 EDT] * Overview of existing games: /games-overview?serverId=bacece8d27fb
[2021-06-10 20:27:06 EDT] * API for game IDs: /api/games?serverId=bacece8d27fb

[2021-06-10 20:27:13 EDT] internal server error /main.js { [Error: ENOENT: no such file or directory, open 'build/main.js.br']
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: 'build/main.js.br' }
^C
```

bafolts:
> I'm not sure this is working yet.

Seeing that error it appears `build/main.js.br` is missing. When I run:

```
rm -rf build
npm run build
npm run start
``

I do not see this problem. Can you try running an `npm install` and re-installing `node_modules`?

kberg:
> > I'm not sure this is working yet.
> 
> Seeing that error it appears `build/main.js.br` is missing. When I run:
> 
> ```
> rm -rf build
> npm run build
> npm run start
> ```
> 
> I do not see this problem. Can you try running an `npm install` and re-installing `node_modules`?

Yes, that took care of it. Thanks!

package.json
bafolts: That is a safe bet. Will move.

bafolts:
Probably will not end up introducing prettier with this change. This project has a large number of files. Prettier, or any linting change, requires updating multiple files which is time consuming for the reviewer or anyone with an open change that may have to merge.

kberg:
What is holding up submitting this?

alrusdi (approved):

bafolts:
> What is holding up submitting this?

Need an approval to merge. Was also out of town on the weekend and away from computer.

kberg (approved):

bafolts:
Once this is merged I will continue to convert more components every day. Until https://github.com/Yuyz0112/vue-type-check/blob/master/package.json#L43 is updated for typescript 4.x I don't think we can use that type checker. I am going to put up a pull request to their repository. If for some reason they don't take the PR I will fork and create a tr-vue-type-check which has the updated typescript version for our needs.

kberg:
I'm working on SelectHowToPay and SelectHowToPayForProjectCard, so if you
want to hold off on those, I'd be grateful.

kberg:
Then again, if you really want to get those going, GO AHEAD.

>

bafolts:
The first components will be components without child components. As enough of the children get converted will then work my way up to their parents. Both SelectHowToPay and SelectHowToPayForProjectCard have enough children that I would think they are weeks away. Want to also get the vue-type-check in place sooner rather than later. That tool type checks better than the jalopy I had written.

kberg:
🎉

PR #3357, opened 2021-06-10
Add Ecoexperts interaction with deco-eco

### LINKED ISSUE #3344: minor rules error: EcoEx+Decomposers should get 3 microbes
image

image

Decomposers [should get ](https://boardgamegeek.com/thread/2096075/article/30501757#30501757) one microbe for its own microbe tag and one for the microbe tag on Eco Experts.

vsrisuknimit:
This PR can closes #3344 .

Note that EcoZone's requirement is not overridden by EcoExperts but it is possible to get a greenery tile from Experimental Forest and then play EcoExperts ignoring its benefit just so you can play EcoZone during the Preludes phase. If we are already talking about terrible plays (EcoExpert on Decomposers), we might as well fix this unimaginably bad play.

¯\\\_(ツ)\_/¯

### DIFF
--- a/src/cards/base/Decomposers.ts
+++ b/src/cards/base/Decomposers.ts
@@ -9,6 +9,7 @@ import {IResourceCard} from '../ICard';
 import {CardRequirements} from '../CardRequirements';
 import {CardRenderer} from '../render/CardRenderer';
 import {CardRenderDynamicVictoryPoints} from '../render/CardRenderDynamicVictoryPoints';
+import {Phase} from '../../Phase';
 
 export class Decomposers extends Card implements IProjectCard, IResourceCard {
   constructor() {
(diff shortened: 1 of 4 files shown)


nwai90 (approved):

bafolts (changes requested):
Can this use `player.lastCardPlayed`? Does it matter if the phase is `Phase.PRELUDES`? If at any point the last card was played was `EcologyExperts` I would think this logic would hold.

vsrisuknimit:
> Can this use player.lastCardPlayed? 

It cannot. lastCardPlayed will be Decomposers / EcoZone.

> Does it matter if the phase is Phase.PRELUDES? If at any point the last card was played was EcologyExperts I would think this logic would hold.

Yes. Someone could play EcoExperts as a second prelude to do nothing. Then try to play Decomposers first during the normal game.

@bafolts

bafolts (approved):
> It cannot. lastCardPlayed will be Decomposers / EcoZone.

It does seem we write to `lastCardPlayed` before we call to `card.play()`. Hard to know if anything relies on that ordering. This change relies on us calling to `playedCards.push` after we call to `card.play()`. Could be worthwhile in a separate change to try to set `lastCardPlayed` after we call to `card.play()` but that could break something. This change relies on how `player.playCard` runs but looks like the unit test is using `player.playCard`. This should ensure that if someone changes `player.playCard` they don't break this.

PR #3361, opened 2021-06-13
Move most of EcologicalSurvey and GeologicalSurvey to a common abstract class.

kberg:
Review after #3360


bafolts (approved):
This is a good change as I didn't realize these two cards were practically doing the same thing.

PR #3364, opened 2021-06-15
add vti for template type checking

bafolts:
This adds `vti` to type check our vue templates.

Sample output on an error I introduced.

```
File : /home/algorithm99/Projects/terraforming-mars/src/components/card/CardTag.vue:2:16
Error: Property 'getClassest' does not exist on type 'CombinedVueInstance<{ getClasses: () => string; } & { index: number; type: string; } & Vue, object, object, object, Record<never, any>>'. Did you mean 'getClasses'?
  1 | <template>
> 2 |   <div :class="getClassest()" />
    |                ^^^^^^^^^^^
  3 | </template>
  4 |
  5 | <script lang="ts">

====================================
VTI found 1 error
```

This will run on every build. As templates are ported to single file components they will have to pass this check.


kberg (commented):
Wow, this is nice.

One thing people will have to start remembering is to run "npm run test:components" from time to time. I will certainly forget some.

I wonder if it would be better to move "test" to "test:server" and have "test" run "test:server" and "test:components"

kberg (approved):

bafolts:
> I wonder if it would be better to move "test" to "test:server" and have "test" run "test:server" and "test:components"

This would be a useful change. How about if `test` runs `test:server && test:components`? This will then allow developers to only test the part they are working on.

kberg:
> > I wonder if it would be better to move "test" to "test:server" and have "test" run "test:server" and "test:components"
> 
> This would be a useful change. How about if `test` runs `test:server && test:components`? This will then allow developers to only test the part they are working on.

Yeah, that's exactly what I meant! Thank you. You want to do it, or shall I?

bafolts:
> > > I wonder if it would be better to move "test" to "test:server" and have "test" run "test:server" and "test:components"
> > 
> > 
> > This would be a useful change. How about if `test` runs `test:server && test:components`? This will then allow developers to only test the part they are working on.
> 
> Yeah, that's exactly what I meant! Thank you. You want to do it, or shall I?

You can add it. Adding component linting and another component.

PR #3369, opened 2021-06-16
Vastly simplify taking final greenery action, making it philares-compatible

kberg:
Instead of relying on knowing that Philares is involved and shunting it into the existing code, this makes the assumption that if a player has a `waitingFor` they just aren't done. It also abandons precomputing whether a player may place greeneries during the phase start in favor of determining whether the player can place a greenery when it's that player's turn.

It's kind of unfortunate that this doesn't just use the existing code about active players and transitions, making it understandably difficult to parse and understand. I hope this code is easier to understand, made even easier by comments. Add more comments as need be. I will probably add more.

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -1068,7 +1068,8 @@ export class Game implements ISerializable<SerializedGame> {
   // Well, this isn't just "go to the final greenery placement". It finds the next player
   // who might be able to place a final greenery.
   // Rename to takeNextFinalGreeneryAction?
-  public gotoFinalGreeneryPlacement(): void {
+
+  public /* for testing */ gotoFinalGreeneryPlacement(): void {
     // this.getPlayers returns in turn order -- a necessary rule for final greenery placement.
     const players = this.getPlayers();
     for (let idx = 0; idx < players.length; idx++) {
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -771,6 +771,7 @@ export class Game implements ISerializable<SerializedGame> {
     });
 
     if (this.gameIsOver()) {
+      this.log('Final greenery placement', (b) => b.forNewGeneration());
       this.gotoFinalGreeneryPlacement();
       return;
     }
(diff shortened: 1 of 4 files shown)


kberg:
I don't want to merge this just yet -- do some more play testing first. But see if you can spot anything awful.

src/Game.ts
```diff
@@ -1051,51 +1052,40 @@ export class Game implements ISerializable<SerializedGame> {
...
-    const players: Player[] = [];
-
-    this.players.forEach((player) => {
+  // Well, this isn't just "go to the final greenery placement". It finds the next player
+  // who might be able to place a final greenery.
+  // Rename to takeNextFinalGreeneryAction?
+  public gotoFinalGreeneryPlacement(): void {
+    // this.getPlayers returns in turn order -- a necessary rule for final greenery placement.
+    const players = this.getPlayers();
+    for (let idx = 0; idx < players.length; idx++) {
```
kberg, line 1074: An old-fashioned for loop! How antiquated. But sometimes lambdas are harder to read. This is just such a case.

bafolts: If you don't need the `idx` you can write this as `for (const player of players)`.

src/Game.ts
```diff
@@ -1051,51 +1052,41 @@ export class Game implements ISerializable<SerializedGame> {
...
+      const player = players[idx];
+      if (this.donePlayers.has(player.id)) {
+        continue;
+      }
       if (this.canPlaceGreenery(player)) {
-        players.push(player);
+        this.startFinalGreeneryPlacement(player);
+        return;
+      } else if (player.getWaitingFor() !== undefined) {
```
bafolts, line 1083: This seems different from what we had before. Is this part of the change that now if any player who is not done has a waiting for we wait for that action?

src/Game.ts
kberg: Yes, I keep forgetting about that expression form. Thanks!

src/Game.ts
kberg: Correct. If you are waiting for the next possible player to do something, then they're not actually done for the game, yet.

bafolts (approved):

PR #3370, opened 2021-06-16
package.json test feature enhancements

kberg:
1. Split server testing to test:server.
2. Run tests without having to delete any part of them.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,9 +63,10 @@
     "makeCss": "lessc src/styles/common.less build/styles.css",
     "makeJson": "node make_static_json.js",
     "start": "node build/src/server.js",
-    "test": "rm -rf build/tests/components/ && mocha --require build/tests/utils/setup.js --recursive build/tests",
+    "test": "npm run test:server && npm run test:components",
+    "test:server": "rm -rf build/tests/components/ && mocha --require build/tests/utils/setup.js --recursive build/tests",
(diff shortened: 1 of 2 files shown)


bafolts:
The github workflow is running `test` and `test:components`. Can you change those to only run `test` with this change?

https://github.com/bafolts/terraforming-mars/blob/main/.github/workflows/main.yml#L19
https://github.com/bafolts/terraforming-mars/blob/main/.github/workflows/main.yml#L33

bafolts (approved):

package.json
```diff
@@ -56,16 +56,17 @@
     "clean": "rm -r ./build/ ./src/genfiles/ || true",
     "compile": "tsc --build src/tsconfig.json",
     "precompile": "npm run makeJson",
-    "cover": "nyc mocha --file build/tests/utils/setup.js --recursive build/tests",
+    "cover": "nyc mocha --file build/tests/utils/setup.js --recursive build/tests --exclude 'build/tests/components/**'",
```
bafolts, line 59: This is a good find as it avoid writing to disk. The `rm -rf` would have involved file i/o. If I am only user of older devices I thank you for us all!

kberg:
> The github workflow is running `test` and `test:components`. Can you change those to only run `test` with this change?
> 
> https://github.com/bafolts/terraforming-mars/blob/main/.github/workflows/main.yml#L19
> https://github.com/bafolts/terraforming-mars/blob/main/.github/workflows/main.yml#L33

Will do.

PR #3372, opened 2021-06-17
single file component: move Tag.ts to Tag.vue, move TagCount.ts to TagCount.vue

bafolts:
Moves two more components to single file components.


kberg (approved):
Thank you. This is good stuff.

PR #3374, opened 2021-06-18
single file component: move CardCost, CardTags, and CardTitle to vue.

bafolts:
Moves three of the Card* to single file components.


src/components/card/CardTags.vue
```diff
@@ -0,0 +1,23 @@
+<template>
```
kberg, line 1: This was a delete/insert versus a file move.

kberg (approved):
LG modulo the issue I mentioned.

src/components/card/CardTags.vue
bafolts: Yes. I need to make this a move. I thought I did the first time. Will take the time to do this so git history is not lost.

PR #3377, opened 2021-06-20
single file component: CardContent, CardCorporationLogo, and CardDescription.

bafolts:
Three more components converted to single file components.


alrusdi (approved):

kberg (approved):
Yeah, this is nice.

PR #3378, opened 2021-06-22
single file component: Board.ts, BoardSpace.ts, and Colony.ts 

bafolts:
The vti component doesn't support typing mixins. I changed from mixin to using the spread operator.


src/components/Board.vue
```diff
@@ -51,8 +192,8 @@ export const Board = Vue.component('board', {
       'isTileHidden': false,
     };
   },
-  mixins: [TranslateMixin],
   methods: {
+    ...TranslateMixin.methods,
```
bafolts, line 196: This is where I used spread operator instead of a mixin as vti doesn't parse the types from the mixin. Without this vti throws an error that it can't find `$t`.

Argeento: :heavy_check_mark:  Added to my list to fix that later

PR #3379, opened 2021-06-23
single file component: Bonus

bafolts:
I rewrote Bonus to use a template string in place of a render function as well. Tested in browser.


alrusdi (approved):

src/components/Bonus.vue
```diff
@@ -10,11 +19,8 @@ export const Bonus = Vue.component('bonus', {
   data: function() {
     return {};
```
Argeento, line 20: Empty data is still needed in SFC?

bafolts: I will check. This came along when it was moved. Will remove if not needed.

Argeento (approved):

PR #3380, opened 2021-06-24
single file component: GameEnd

bafolts:
Tested this by skipping every generation until I arrived at the end screen. Saw we used `player-end` and not `game-end` in `App`. Not sure why this worked before this change but with this change `game-end` was not found. There is no `player-end` as far as I can tell.


kberg:
I'm sorry, because it's easier to test than that. I have written some notes
on the subject which I was going to make into a video, but clearly I have
to just post it in the wiki.

kberg (approved):

bafolts:
Thankfully I believe in patience. Wasn't that painful to get to the end screen. Now I see why there is a push to have the game always run on one screen without scrolling though. Had to continually scroll down or up to either skip a generation or not buy cards during research phase.

kberg:
I also believe in patience, but lazy is nice to believe in, too. :)

>

PR #3382, opened 2021-06-24
Remove `game` prefix from the local filesystem db

kberg:
Now that game ids start with `g` this is better.


bafolts (approved):
Any game in progress while someone updates might be lost. Not sure how many people use the local file system that this will impact.

kberg:
I expect this is impact One Person. :D

PR #3383, opened 2021-06-25
single file component: GameHome, GameSetupDetails, and GamesOverview

bafolts:
During this change vetur spotted that we did not pass `undoOption` or `requiresMoonTrackCompletion` to the game options model. Added those properties to the model to allow build to pass.


kberg (approved):
Nice work.

PR #3386, opened 2021-06-25
Add Bonus.vue to the Moon. Somehow this wasn't part of The Moon. Huh.

### LINKED ISSUE #3385: Moon board no longer showing bonuses
image

@bafolts I believe this is related to vue file migration, so I'm assigning it to you, but also I can peek at this.

kberg:
Fixes #3385


kberg:
image

nwai90 (approved):

bafolts:
I think when we aren't using the single file components we ended up declaring components globally. When we created the `Board` class I am assuming that set it up for use in any other template. As I continue the migration I will grep for any place using the components being moved and make sure they are registered.

kberg:
:thumbsup

PR #3387, opened 2021-06-26
single file component: DebugUI

bafolts:
Moves DebugUI to single file component.


src/components/App.ts
```diff
@@ -185,7 +185,7 @@ export const mainAppSettings = {
       app.screen = 'create-game-form';
     } else if (currentPathname === '/load') {
       app.screen = 'load';
-    } else if (currentPathname === '/debug-ui' || currentPathname === '/cards') {
```
bafolts, line 188: It doesn't seem like `debug-ui` is used in the path. This took me a few cycles to see that `/cards` had to be uses. The `debug-ui` always seems to 404 as no server route is listening.

kberg: It used to be debug-ui, and changed to cards. But I think you know that.

kberg (approved):

PR #3393, opened 2021-06-30
single file component: Button

bafolts:
This change involved finding places where `<Button` was being used in templates. I grepped for `<Button` and found a few components where I added `Button` to their `components`. The `vti` tool is also starting to spot issues with our templates. I had to make `title` and `align` optional for the `Button` props.


src/components/common/Button.vue
```diff
@@ -1,9 +1,20 @@
...
+
 import Vue from 'vue';
 
-export const Button = Vue.component('Button', {
+export default Vue.extend({
+  name: 'Button',
   props: {
     title: {
       type: String,
+      required: false,
```
Argeento, line 17: Optional props must have a default value.
Current [eslint config](https://github.com/bafolts/terraforming-mars/blob/main/.eslintrc.js#L8) should [throw the error here](https://eslint.vuejs.org/rules/require-default-prop.html).

src/components/common/Button.vue
```diff
@@ -12,6 +23,7 @@ export const Button = Vue.component('Button', {
     align: {
       type: String,
       validator: (align) => ['right', 'left'].includes(align),
+      required: false,
```
Argeento, line 26: Another optional props without a default value

Argeento (changes requested):

src/components/common/Button.vue
bafolts: We are using `plugin:vue/essential` and not `plugin:vue/recommended`. When I put in `plugin:vue/recommended` I see 2000+ warnings including this. I will put a default here and we will need to enable `plugin:vue/recommended` as it seems to be catching countless warnings similar to this.

Argeento: Oh. Yea, You are right, sorry

Argeento (approved):

kberg (approved):
Modulo @Argeento's comment.

src/components/common/Button.vue
bafolts: With the 2000+ warnings that showed up once I changed to _recommended_ from _essential_ let me try turning on one rule at a time and fixing those warnings. I wish we had used _recommended_ from the start. Pretty sure we only have _essential_ because that must be what was in some vue setup I copy and pasted when first setting up vue. Thank you for filling in as linter for time being!

PR #3394, opened 2021-06-30
Improve components testing

Argeento:
Removed: `--mode development`
There is no [option](https://sysgears.github.io/mochapack/docs/installation/cli-usage.html) `mode` in mochapack cli

Removed: `--webpack-config webpack.config.js`
webpack.config.js is set by default

Removed: `npm run compile`
Seems like this command is unnecessary here 🤔

Added watch mode to component testing. No more waiting 15s+ to check a test

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "test:server": "mocha --require build/tests/utils/setup.js --recursive build/tests --exclude 'build/tests/components/**'",
     "pretest:server": "tsc --build tests/tsconfig.json",
     "test:components": "vti diagnostics && mochapack --require tests/components/setup.ts tests/components/*.spec.ts",
+    "test:components:watch": "mochapack --watch --require tests/components/setup.ts tests/components/*.spec.ts",
     "watch": "webpack --watch --mode=development",
     "webpack": "webpack"
   },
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -66,7 +66,8 @@
     "test": "npm run test:server && npm run test:components",
     "test:server": "mocha --require build/tests/utils/setup.js --recursive build/tests --exclude 'build/tests/components/**'",
     "pretest:server": "tsc --build tests/tsconfig.json",
-    "test:components": "vti diagnostics && npm run compile && mochapack --mode development --webpack-config webpack.config.js --require tests/components/setup.ts tests/components/*.spec.ts",
+    "test:components": "vti diagnostics && mochapack --require tests/components/setup.ts tests/components/*.spec.ts",
```
kberg, line 69: Does `:` have a semantic meaning for npm? In other words, is there a way to say that any test runs `pretest:*` so that `tsc` can run regardless of whether we run test:server or test:components? Right now we have pretest:server, but if I don't want to test server components it isn't going to help.

Argeento: > Does : have a semantic meaning for npm? 
No, it is a divider and behaves like any other normal char

bafolts: This will speed up the github action workflow process so makes sense to me.

bafolts (approved):

PR #3401, opened 2021-07-01
Use visually consistent icons for Ares MAs

nwai90:
Before:
Screenshot 2021-07-01 at 3 37 56 PM

After:
Screenshot 2021-07-01 at 4 09 58 PM


alrusdi (approved):

kberg:
Why does the email version have what looks like hazard tiles? I was so confused!

nwai90:
> Why does the email version have what looks like hazard tiles? I was so confused!

@kberg I uploaded the wrong icons at first (with hazard tiles) and replaced them later :D

kberg:
This is one of the many small visual changes I thought, "What I've done is
good enough and someone will come along and fix it eventually." Mission
accomplished! (Thank you!)

>

bafolts:
This looks great!

PR #3404, opened 2021-07-01
[Moon] Add card Mooncrate Convoys to Mars.

kberg:


src/moon/SellSteel.ts
```diff
@@ -0,0 +1,39 @@
...
+export class SellSteel implements DeferredAction {
+  public priority = Priority.DEFAULT;
+
+  constructor(
+    public player: Player,
+    public title: string = 'Sell your steel for 3M€ each.',
+  ) {}
+
+  private logSale(unitsSold: number) {
+    this.player.game.log('${0} sold ${1} units of steel', (b) => b.player(this.player).number(unitsSold));
```
nwai90, line 15: Perhaps drop "units of" from the log message for brevity?

nwai90 (approved):

src/moon/SellSteel.ts
kberg: You're right.

PR #3405, opened 2021-07-02
single file component: PlayerInputFactory

bafolts:
This is one I have been messing with to try to get right for the past few days. The `PlayerInputFactory` was never a vue component but a strange factory. This change makes it a vue component with a template. Putting a do not merge label on this because I want to test it some more. I have already done some testing and everything seems fine. Once this is in place I plan to change `OrOptions` and `AndOptions` to also use templates to avoid the render functions.


src/components/AndOptions.ts
```diff
@@ -33,7 +32,7 @@ export const AndOptions = Vue.component('and-options', {
     saveData: function() {
       for (let i = 0; i < this.$data.childComponents.length; i++) {
         const componentInstance = this.$data.childComponents[i]
-          .componentInstance;
+          .componentInstance.$children[0];
```
kberg, line 35: Erm, what's this?

kberg (commented):
I can look at this tomorrow. Bedtime now.

src/components/AndOptions.ts
bafolts: I want to refactor this as this took me half an hour or so to debug. The `AndOptions` and `OrOptions` go looking for the `PlayerInput` component to call the `saveData`. I can update this so on the `PlayerInputFactory` component when you call to `saveData` it handles finding the component to call `saveData`. This will avoid having to go looking for the component in `AndOptions` and `OrOptions`.

bafolts:
Plan to finish this as next single file component. Will test more and then remove the do not merge label.

src/components/AndOptions.ts
```diff
@@ -66,17 +65,16 @@ export const AndOptions = Vue.component('and-options', {
...
+              players: this.players,
+              player: this.player,
+              playerinput: option,
+              onsave: (out: Array<Array<string>>) => {
                 this.$data.responded[idx] = out[0];
               },
-              false,
-              true,
-            ),
+              showsave: false,
```
kberg, line 75: Love these (effectively) named parameters. Thank you.

src/components/App.ts
```diff
@@ -66,6 +67,7 @@ export const mainAppSettings = {
     logPaused: false,
   } as MainAppData,
   'components': {
+    'player-input-factory': PlayerInputFactory,
```
kberg, line 70: I notice you sometimes remove these kebab-case aliases. Are you keeping them?

src/components/PlayerInputFactory.vue
```diff
@@ -0,0 +1,109 @@
+<template>
```
kberg, line 1: FWIW this isn't a move. LMK if you don't want to deal with that and I'll just do the comparison.

kberg (approved):
Seems OK minus my minor comments.

src/components/App.ts
bafolts: The vue recommended linting rule recommends this lower case version but the upper case version works so great when plopping into `components`. May just stick with the good thing we have going and keep this as `PlayerInputFatory`.

src/components/PlayerInputFactory.vue
bafolts: So I am not sure what happens here entirely. I have been doing the `git mv` first as a separate commit and then rewriting the component later. On the "do the move first" commit I move `PlayerInputFactory.ts` to `PlayerInputFactory.vue`. Not sure why this sometimes seems to work and other times it doesn't.

src/components/AndOptions.ts
bafolts: I love this entire change since moving `PlayerInputFactory` to use a template is better than the custom jalopy I had written prior. This is going to be the riskiest of the update to vue templates but we should not be covered by the linting and veture tool when using the `PlayerInputFactory`.

bafolts:
Given this change is riskier than the others I am going to write at least one unit test to check the `saveData` wiring I have in place.

kberg:
> Given this change is riskier than the others I am going to write at least one unit test to check the `saveData` wiring I have in place.

Huzzah! Adding "Do not merge" label.

bafolts:
Added unit tests that verifies every child component of the factory runs and ends up with the `saveData` function we need. Tested this manually through a few of the common select options. This is by far the riskiest of the conversions to single file component.

PR #3406, opened 2021-07-02
Add Moon card Luna Archives

kberg:
Tests are pending. But FYI

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -874,6 +874,7 @@ export class Player implements ISerializable<SerializedPlayer> {
       megaCredits: 0,
       microbes: 0,
       floaters: 0,
+      science: 0,
     };
     try {
       const howToPay: HowToPay = JSON.parse(json);
@@ -1339,8 +1340,16 @@ export class Player implements ISerializable<SerializedPlayer> {
       totalToPay += howToPay.floaters * DEFAULT_FLOATERS_VALUE;
     }
 
+    if (howToPay.science ?? 0 > 0) {
+      totalToPay += howToPay.science;
+    }
+
     if (howToPay.megaCredits > this.megaCredits) {
       throw new Error('Do not have enough M€');
+    }
+
+    if (howToPay.science !== undefined) {
+      totalToPay += howToPay.science;
     }
 
     totalToPay += howToPay.megaCredits;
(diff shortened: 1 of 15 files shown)


src/Player.ts
```diff
@@ -1373,6 +1382,13 @@ export class Player implements ISerializable<SerializedPlayer> {
     return 0;
   }
 
+  public getSpendableScienceResources(): number {
+    const lunaArchives = this.playedCards.find((card) => card.name === CardName.LUNA_ARCHIVES);
```
nwai90, line 1386: Suggestion: `if (this.cardIsInEffect(CardName.LUNA_ARCHIVES)) ...`

src/Player.ts
```diff
@@ -1339,10 +1340,18 @@ export class Player implements ISerializable<SerializedPlayer> {
       totalToPay += howToPay.floaters * DEFAULT_FLOATERS_VALUE;
     }
 
+    if (howToPay.science ?? 0 > 0) {
```
bafolts, line 1343: Was this intentional that `totalToPay` is incremented by `howToPay.science` here and on line 1352?

src/cards/moon/LunaArchives.ts
```diff
@@ -0,0 +1,47 @@
...
+  };
+
+  public resourceCount = 0;
+
+  public canAct() {
+    return true;
+  }
+
+  public action(player: Player) {
+    const count = player.game.getPlayers().map((p) => p.getTagCount(Tags.MOON)).reduce((p, v) => p + v, 0);
```
bafolts, line 39: The action for this card reads `Add 1 science resource here for each Moon tag you have`. This code seems to add a science resource for a moon tag that any player has. Is this intentional?

src/components/PaymentWidgetMixin.ts
```diff
@@ -81,6 +83,7 @@ export const PaymentWidgetMixin = {
         amountHave = (this as any).playerinput.floaters;
         if (this.isStratosphericBirdsEdgeCase()) amountHave--;
       }
+      if (target === 'science') amountHave = (this as any).playerinput.science;
```
bafolts, line 86: Unrelated to this PR but I will update this to use types instead of `any`. This is an accident waiting to happen during a refactor.

kberg: Might be tricky. But if you turned that into aa sub-data structure, I'd be s pleased.

src/Player.ts
kberg: Almost works. The problem is `lunaArchives` still has to be passed to the connected statement.

src/Player.ts
kberg: Yes, it matches the statements above it. Or am I mistaken?

src/cards/moon/LunaArchives.ts
kberg: Well, worked when I played solo! :b whoops.

bafolts (approved):

PR #3410, opened 2021-07-03
Add Turmoil.ifTurmoil, Turmoil.ifTurmoilElse, Turmoil.getTurmoil.

kberg:
While there are places where this change is slightly more complicated, it eliminates the need for if(there's a turmoil object) guards, and more important the (I know there's turmoil, so `game.turmoil!`) cases. Those will still fail with an error, but the error will report which game was the problem.

In a couple of places, it looks like a huge change, and it's just whitespace. I tried to avoid those but the linter insisted. Use the github code review feature that ignores whitespace - I'm sure you know how to do that.

There are some places where it seems I removed guards. For example `Septum Tribus`, which had a `canPlay` test. But Septum Tribus is a Turmoil card. `game.turmoil` WILL be there. In fact, the card won't even be in the deck without it. That kind of made the test useless, which is why you'll see I deleted it.

Speaking of cards in the deck, that means I also removed cards from other manifests that had the Turmoil compatibility attribute. So, I think all-in-all this is a safe change.

But if you think this requires some extra tests, let me know.

### DIFF
--- a/src/cards/CardRequirement.ts
+++ b/src/cards/CardRequirement.ts
@@ -7,6 +7,7 @@ import {ResourceType} from '../ResourceType';
 import {TileType} from '../TileType';
 import {GlobalParameter} from '../GlobalParameter';
 import {MoonExpansion} from '../moon/MoonExpansion';
+import {Turmoil} from '../turmoil/Turmoil';
 
 const firstLetterUpperCase = (s: string): string => s.charAt(0).toUpperCase() + s.slice(1);
 
@@ -91,7 +92,7 @@ export class CardRequirement {
   public satisfies(player: Player): boolean {
     switch (this.type) {
     case RequirementType.CHAIRMAN:
-      return player.game.turmoil?.chairman === player.id;
+      return Turmoil.getTurmoil(player.game).chairman === player.id;
 
     case RequirementType.CITIES:
       if (this._isAny) {
@@ -117,11 +118,9 @@ export class CardRequirement {
       return this.satisfiesInequality(greeneries);
 
     case RequirementType.PARTY_LEADERS:
-      if (player.game.turmoil !== undefined) {
-        const parties = player.game.turmoil.parties.filter((party) => party.partyLeader === player.id).length;
-        return this.satisfiesInequality(parties);
-      }
-      return false;
+      const turmoil = Turmoil.getTurmoil(player.game);
+      const parties = turmoil.parties.filter((party) => party.partyLeader === player.id).length;
+      return this.satisfiesInequality(parties);
 
     case RequirementType.OCEANS:
       return this.checkGlobalRequirement(player, GlobalParameter.OCEANS, this.amount, this.isMax);
@@ -257,9 +256,6 @@ export class PartyCardRequirement extends CardRequirement {
     return this.party.toLowerCase();
   }
   public satisfies(player: Player): boolean {
-    if (player.game.turmoil !== undefined) {
-      return player.game.turmoil.canPlay(player, this.party);
-    }
-    return false;
+    return Turmoil.getTurmoil(player.game).canPlay(player, this.party);
   }
 }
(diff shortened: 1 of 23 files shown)


kberg:
Oh right, there are also a couple of spots where I completely removed the turmoil guard, because the code inside didn't even do turmoil things! So, you might spot those, too.

kberg:
Finally, I'm happy to break this up into smaller, safer chunks. Any thoughts on that are also welcome.

src/cards/CardRequirement.ts
```diff
@@ -257,9 +256,6 @@ export class PartyCardRequirement extends CardRequirement {
     return this.party.toLowerCase();
   }
   public satisfies(player: Player): boolean {
-    if (player.game.turmoil !== undefined) {
```
bafolts, line 260: This seems like it could lead to a bug. If calling to `satisfies` and `player.game.turmoil` is `undefined` this will throw an error versus returning `false`.

src/cards/turmoil/SeptumTribus.ts
```diff
@@ -36,15 +37,14 @@ export class SeptumTribus extends Card implements IActionCard, CorporationCard {
     return undefined;
   }
 
-  public canAct(player: Player): boolean {
-    return player.game.gameOptions.turmoilExtension;
+  public canAct(): boolean {
+    return true;
```
nwai90, line 41: @kberg This could cause an issue in the odd scenario that a game is played without Turmoil expansion, but the players used Custom corps list to manually select Septum Tribus. The corp action will always throw an error in this scenario, though it is very unlikely to occur. WDYT?

src/cards/CardRequirement.ts
bafolts: I reviewed the rest of these and this is the only change that stands out as a potential problem. Using a `Turmoil.ifTurmoilElse` here makes sense to me. Otherwise any caller to `satisfies` needs the turmoil expansion or a runtime error will be thrown.

kberg: Cards declared with a party requirement not declared with turmoil SHOULD throw an error, see?

kberg: Cards declared with a party requirement not declared to be compatible with Turmoil SHOULD throw an error.

src/cards/turmoil/SeptumTribus.ts
kberg: I hear you and I agree that under current conditions, this could cause an error, so good catch.

But I also think it's right for the app to not allow this condition to occur -- this corp should not be allowed to be included in a game without Turmoil, so when receiving a CreateGame request, reject corps not compatible with the current game parameters.

So what do you think is the right thing, here? Would settling on having the corp be outlawed without Turmoil be an acceptable long-term outcome? And if so, would you be okay having this code remain as-is as long as I follow up quickly with another PR that addresses that?

src/cards/CardRequirement.ts
kberg: I'll expand: cards that require Turmoil need to have one of two conditions: it needs to be part of the Turmoil manifest, or, part of another manifest specifying `compatibility: GameModule.Turmoil` so they're only included when game.turmoil is included.

kberg (commented):
Thanks for these deep analyses.

src/cards/turmoil/SeptumTribus.ts
bafolts: Not allowing the game to be created makes sense to me. I don't think any of the turmoil project cards are of any value without the turmoil expansion. I would assume it is a mistake if a user created a game with this setup.

bafolts (approved):

src/cards/turmoil/SeptumTribus.ts
nwai90: `Would settling on having the corp be outlawed without Turmoil be an acceptable long-term outcome?`

Absolutely, this is a very illogical setup albeit a valid one. Preventing the game from entering such a state is a desirable outcome.

kberg: OK, I'll submit this with a fast follow tonight.

PR #3411, opened 2021-07-03
Add Moon Card An Offer You Can't Refuse.

kberg:


src/turmoil/Turmoil.ts
```diff
@@ -124,32 +131,24 @@ export class Turmoil implements ISerializable<SerializedTurmoil> {
       game: Game,
       source: 'lobby' | 'reserve' = 'lobby'): void {
       const party = this.getPartyByName(partyName);
-      if (party) {
```
kberg, line 127: Huh, apaprently I broke turmoil. Back in a moment.

kberg: Ah, no, I didn't I just removed the if (party) check which is unnecessary since `getPartyByName` no longer returns `undefined`. View with whitespace changes off.

alrusdi (approved):

PR #3413, opened 2021-07-05
single file components: CardNumber, CardExpansion, and CardVictoryPoints

bafolts:
The CardVictoryPoints required updates to the type of the prop. It needed to be `[Number, Object]` once vetur was checking it.


kberg (approved):
Nice work.

Issue #3414, opened 2021-07-06
Could not add floaters from 'Hydrogen to Venus' to Celestic


sfurlani:
From my understanding the Celestic Corporation is both a `Venus` and a `Card` so Hydrogen to Venus _should_ have given me the option to add the 3 `Floaters` to it.

Most of the other cards/actions seemed to allow me to add Floaters to Celestic including:
* Trading with Titan
* Celestic Action
* Floater Technology 
* Jupiter Floating Station 
* Atmoscoop

## Code Snippits (From Chrome Inspector)

Celestic.js
```
var Celestic = (function (_super) {
    __extends(Celestic, _super);
    function Celestic() {
        var _this = _super.call(this, {
            name: CardName_1.CardName.CELESTIC,
            tags: [Tags_1.Tags.VENUS],
            startingMegaCredits: 42,
            resourceType: ResourceType_1.ResourceType.FLOATER,
            cardType: CardType_1.CardType.CORPORATION,
```

HydrogenToVenus.js
```
    HydrogenToVenus.prototype.play = function (player) {
        var jovianTags = player.getTagCount(Tags_1.Tags.JOVIAN);
        var floatersCards = player.getResourceCards(ResourceType_1.ResourceType.FLOATER).filter(function (card) {
            return HydrogenToVenus.venusCardsWithFloaters.has(card.name);
        });
        if (jovianTags > 0) {
            if (floatersCards.length === 1) {
                player.addResourceTo(floatersCards[0], { qty: jovianTags, log: true });
            }
            if (floatersCards.length > 1) {
                return new SelectCard_1.SelectCard('Select card to add ' + jovianTags + ' floater(s)', 'Add floater(s)', floatersCards, function (foundCards) {
                    player.addResourceTo(foundCards[0], { qty: jovianTags, log: true });
                    player.game.increaseVenusScaleLevel(player, 1);
                    return undefined;
                });
            }
        }
        player.game.increaseVenusScaleLevel(player, 1);
        return undefined;
    };
    HydrogenToVenus.venusCardsWithFloaters = new Set([
        CardName_1.CardName.AERIAL_MAPPERS,
        CardName_1.CardName.CELESTIC,
        CardName_1.CardName.DEUTERIUM_EXPORT,
        CardName_1.CardName.DIRIGIBLES,
        CardName_1.CardName.EXTRACTOR_BALLOONS,
        CardName_1.CardName.FLOATING_HABS,
[...]

PR #3415, opened 2021-07-06
Change Hydrogen To Venus name-based filter to tag-based

sfurlani:
Related to issue #3414

The name-based filter from `HydrogenToVenus.venusCardsWithFloaters` was unable to perform a text-compare on `CardName.CELESTIC` vs. `CardName.CELESTIC_REBALANCED`

Replaced with a check to see if the card contained any tags (copied from `Player.getTagCount()` https://github.com/bafolts/terraforming-mars/blob/26eecb167bed1595405193c555ea48fe6e04a01a/src/Player.ts#L770

### DIFF
--- a/src/cards/venusNext/HydrogenToVenus.ts
+++ b/src/cards/venusNext/HydrogenToVenus.ts
@@ -41,9 +41,7 @@ export class HydrogenToVenus extends Card {
 
   public play(player: Player) {
     const jovianTags: number = player.getTagCount(Tags.JOVIAN);
-    const floatersCards = player.getResourceCards(ResourceType.FLOATER).filter((card) => {
-      return card.tags.filter((cardTag) => cardTag === Tags.VENUS).length > 0
-    });
+    const floatersCards = player.getResourceCards(ResourceType.FLOATER).filter((card) => card.tags.includes(Tags.VENUS));
     if (jovianTags > 0) {
       if (floatersCards.length === 1) {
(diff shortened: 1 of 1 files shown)


sfurlani:
Note: I haven't spun up a server and tested this fix myself?

kberg:
Thanks for the patch. Since you are interested in it working, please at least run the tests. You can do this with:

`npm run test`

This assumes of course that you've done everything else to set up a dev environment. But more important, adding a test case that verifies that this is fixed would be great.

The thing about the original code is that I don't understand why it was written that way. And it makes me think of the Chesterton's Fence adage: "If there's a fence in the middle of the road, and you think 'surely it shouldn't be there, so I should take it down,' find out why it was put up in the first place."

Oh, I don't know how easy that will be to do, but perhaps someone can say so, so at least we understand what's happening.

sfurlani:
OK! I will see if I can spin it up and run it later this week

kberg:
> OK! I will see if I can spin it up and run it later this week

Well the good news is as-is none of the current tests fail. It's kind of late at night -- let's see if other people respond tomorrow.

nwai90:
@sfurlani This refactor is fine if there is a test for it. The easier way is to add `CardName.CELESTIC_REBALANCED` to the `venusCardsWithFloaters` Set, but that change does not belong here.

vsrisuknimit (approved):
Looks good to me. This simplification is basically what we have now at Air-Scrapping Expedition.

PR #3418, opened 2021-07-07
Better fit of card titles in Russian and Spanish languages

alrusdi:
Calculate card title size from translated text instead on English only.


kberg:
This looks great. If you have screenshot I'd be grateful, but this seems pretty simple.

bafolts (approved):

Issue #3419, opened 2021-07-07
i (probably) broke the how to pay again


kberg:
![image](https://user-images.githubusercontent.com/413481/124768874-06654f80-df07-11eb-8e97-e9ccf0ffa6b3.png)

kberg:
Fixed. Adding tests and server-side validation.

PR #3423, opened 2021-07-08
remove any from PaymentWidgetMixin

bafolts:
This change removes the usage of `any` from `PaymentWidgetMixin`. It is still not as good as having everything typed and relies on casting to `unknown` but should help avoid issues like we saw with #3419. The mixins with typescript are difficult to type so we need to migrate away from using mixins. The vetur tool also doesn't work with mixins.

In places where without `any` we would have typescript compilation errors I threw a runtime exception.


src/components/PaymentWidgetMixin.ts
```diff
@@ -1,98 +1,162 @@
...
+import {Units} from '../Units';
+
+export interface SelectHowToPayModel {
+    card?: CardModel;
+    cost: number;
+    heat: number;
+    megaCredits: number;
+    steel: number;
+    titanium: number;
+    microbes: number;
```
kberg, line 17: Something I just discovered is that microbes and floaters are not used in SHTP at all, and I bet you can remove it from here, and the code itself.

kberg: The only reason they were in the model was because they had to accommodate this mixin with its anyness.

kberg: ... so move them from this model to the parent model.

src/components/PaymentWidgetMixin.ts
```diff
@@ -1,98 +1,162 @@
...
+  player: PlayerModel;
+  playerinput: PlayerInputModel;
+}
 
 export const PaymentWidgetMixin = {
   'name': 'PaymentWidgetMixin',
   'methods': {
     getMegaCreditsMax: function(): number {
-      return Math.min((this as any).player.megaCredits, (this as any).$data.cost);
+      return Math.min((this as unknown as PaymentWidgetModel).player.megaCredits, (this as unknown as PaymentWidgetModel).$data.cost);
```
kberg, line 47: Some well-placed newlines will make this easy to read.

```
return Math.min(
    (this as unknown as PaymentWidgetModel).player.megaCredits,
    (this as unknown as PaymentWidgetModel).$data.cost));
```

or 

```
const model = this as unknown as PaymentWidgetModel;
return Math.min(model.player.megaCredits, model.$data.cost)
```

src/components/PaymentWidgetMixin.ts
```diff
@@ -1,98 +1,162 @@
...
     },
-    getResourceRate: function(resourceName: string): number {
+    getResourceRate: function(resourceName: 'steel' | 'titanium' | 'heat' | 'microbes' | 'floaters' | 'science'): number {
```
kberg, line 49: How funny I did the exact same thing, but I made this a `type` definition. See https://github.com/bafolts/terraforming-mars/pull/3424/files#diff-87064214149a74384a7f6395bbaf9d63841eb2a633fd98d8d0d0f0b3021b9604R6 and see if you can use any of it.

src/components/PaymentWidgetMixin.ts
```diff
@@ -1,98 +1,162 @@
...
       }
       return rate;
     },
-    reduceValue: function(target: string, to: number): void {
-      const currentValue: number = (this as any)[target];
+    reduceValue: function(target: 'steel' | 'titanium' | 'heat' | 'megaCredits' | 'microbes' | 'floaters' | 'science', to: number): void {
+      const currentValue: number | undefined = (this as unknown as PaymentWidgetModel)[target];
+
+      if (currentValue === undefined) {
+        throw new Error(`can not reduceValue for ${target} on this`);
```
kberg, line 66: :+1:

src/components/PaymentWidgetMixin.ts
```diff
@@ -1,98 +1,162 @@
...
+        amountHave = (this as unknown as PaymentWidgetModel).player[target];
+      }
 
       let amountNeed: number = cardCost;
       if (['titanium', 'steel', 'microbes', 'floaters'].includes(target)) {
         amountNeed = Math.floor(cardCost/this.getResourceRate(target));
       }
 
-      if (target === 'microbes') amountHave = (this as any).playerinput.microbes;
+      if (target === 'microbes') amountHave = (this as unknown as PaymentWidgetModel).playerinput.microbes;
```
kberg, line 141: You wrote "this as unknown as PaymentWidgetModel" so much, just add 

asModel: function(): PaymentWidgetModel {
  return this as unknown as PaymentWidgetModel;
}

Then all of this will read better:

amountHave = this.asModel().playerinput.microbes;

kberg (changes requested):
This looks like a good start. Clicking Request Changes viz our discussion earlier. Since I just wrote a test for shtp, it will guide you through this rocky period. Consider adopting my change first and merging it in to here, making your code less risky?

kberg:
This change is terrific.

src/components/PaymentWidgetMixin.ts
bafolts: Yes. I will merge yours first and then update mine accordingly.

src/components/PaymentWidgetMixin.ts
```diff
@@ -1,65 +1,127 @@
...
+export interface SelectHowToPayForProjectCardModel extends SelectHowToPayModel {
+  cardName: CardName;
+  card: CardModel;
+  cards: Array<CardModel>;
+  tags: Array<Tags>
+  science: number;
+  available: Units;
+}
+
+interface PaymentWidgetModel extends SelectHowToPayModel {
```
kberg, line 31: Interesting. Why is this interface necessary and why does it both extend SHTPM and why is it one of its attributes?

src/components/PaymentWidgetMixin.ts
```diff
@@ -1,65 +1,127 @@
...
-      const currentValue: number = (this as any)[target];
-      let maxValue: number = max ?? (this as any).player[target];
+    addValue: function(target: ResourcesWithRates | 'heat' | 'megaCredits' | 'science', to: number, max?: number): void {
+      const currentValue: number | undefined = this.asModel()[target];
+
+      if (currentValue === undefined) {
+        throw new Error(`can not addValue for ${target} on this`);
+      }
+
+      let maxValue: number | undefined = max;
```
kberg, line 94: There's a switch in there.

```
let maxValue: number | undefined = max;

if (maxValue === undefined && target !== 'microbes' && target !== 'floaters' && target !== 'science') {
  maxValue = this.asModel().player[target];
}

switch(target) {
case 'megaCredits':
  maxValue = this.getMegaCreditsMax();
  break;
case 'microbes':
  maxValue = this.asModel().playerinput.microbes;
  break;
case 'floaters':
  maxValue = this.asModel().playerinput.floaters;
  if (maxValue !== undefined && this.isStratosphericBirdsEdgeCase()) maxValue--;
  break;
case 'sceince':
  maxValue = this.asModel().playerinput.science;
  break;
}

 if (currentValue === maxValue) return;

if (maxValue === undefined) {
  throw new Error(`unable to determine maxValue for ${target}`);
}
```

src/components/PaymentWidgetMixin.ts
```diff
@@ -73,30 +135,40 @@ export const PaymentWidgetMixin = {
...
+        amountHave = this.asModel().player[target];
+      }
 
       let amountNeed = cardCost;
       if (target !== 'science' && target !== 'heat') {
         amountNeed = Math.floor(cardCost / this.getResourceRate(target));
       }
 
-      if (target === 'microbes') amountHave = (this as any).playerinput.microbes;
+      if (target === 'microbes') amountHave = this.asModel().playerinput.microbes;
```
kberg, line 153: This looks pretty similar to addValue. Is there some reuse? Can be later.

kberg (changes requested):

src/components/PaymentWidgetMixin.ts
bafolts: The interface is capturing what is used on `this`. The `this` has all the properties of `SelectHowToPayModel`.

src/components/PaymentWidgetMixin.ts
bafolts: Trying to minimize what logic changes with this change. Definitely re-use throughout this object.

src/components/PaymentWidgetMixin.ts
kberg: > The interface is capturing what is used on `this`. The `this` has all the properties of `SelectHowToPayModel`.

Huh, OK then. Thanks!

kberg (approved):
Thanks for the persistence.

PR #3424, opened 2021-07-08
Add test for SelectHowToPay.

kberg:
Made it more sophsiticated so it's easier to read and more clicks can be verified.

### DIFF
--- a/src/components/PaymentWidgetMixin.ts
+++ b/src/components/PaymentWidgetMixin.ts
@@ -63,7 +63,7 @@ export const PaymentWidgetMixin = {
       const heatMc = ta['heat'] ?? 0;
       const titaniumMc = (ta['titanium'] ?? 0) * this.getResourceRate('titanium');
       const steelMc = (ta['steel'] ?? 0) * this.getResourceRate('steel');
-      const microbesMc = (ta['microbes'] ?? 0) * this.getResourceRate('steel');
+      const microbesMc = (ta['microbes'] ?? 0) * this.getResourceRate('microbes');
       const floatersMc = (ta['floaters'] ?? 0) * this.getResourceRate('floaters');
       const scienceMc = ta['science'] ?? 0;
 
@@ -72,14 +72,14 @@ export const PaymentWidgetMixin = {
 
       ta['megaCredits'] = Math.max(0, Math.min(this.getMegaCreditsMax(), remainingMC));
     },
-    setMaxValue: function(target: ResourcesWithRates | 'science', max?: number): void {
+    setMaxValue: function(target: ResourcesWithRates | 'science' | 'heat', max?: number): void {
       let currentValue: number = (this as any)[target];
       const cardCost: number = (this as any).$data.cost;
       let amountHave: number = max ?? (this as any).player[target];
 
-      let amountNeed: number = cardCost;
-      if (!(target === 'science')) {
-        amountNeed = Math.floor(cardCost/this.getResourceRate(target));
+      let amountNeed = cardCost;
+      if (target !== 'science' && target !== 'heat') {
+        amountNeed = Math.floor(cardCost / this.getResourceRate(target));
       }
 
       if (target === 'microbes') amountHave = (this as any).playerinput.microbes;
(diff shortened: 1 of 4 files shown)


src/components/PaymentWidgetMixin.ts
```diff
@@ -3,13 +3,14 @@ import {CardName} from '../CardName';
 import {CardModel} from '../models/CardModel';
 import {ResourceType} from '../ResourceType';
 
+type ResourcesWithRates = 'titanium' | 'steel' | 'microbes' |'floaters';
```
bafolts, line 6: We can eventually improve upon having to provide these types manually but for the time being doing this is great.

src/components/PaymentWidgetMixin.ts
```diff
@@ -58,23 +59,26 @@ export const PaymentWidgetMixin = {
...
-              (this as any)['titanium'] * this.getResourceRate('titanium') -
-              (this as any)['steel'] * this.getResourceRate('steel') -
-              (this as any)['microbes'] * this.getResourceRate('microbes') -
-              (this as any)['floaters'] * this.getResourceRate('floaters') -
-              ((this as any)['science'] ?? 0);
+      const ta = this as any;
+      const heatMc = ta['heat'] ?? 0;
+      const titaniumMc = (ta['titanium'] ?? 0) * this.getResourceRate('titanium');
+      const steelMc = (ta['steel'] ?? 0) * this.getResourceRate('steel');
+      const microbesMc = (ta['microbes'] ?? 0) * this.getResourceRate('steel');
```
bafolts, line 66: This should read `const microbesMc = (ta['microbes'] ?? 0) * this.getResourceRate('microbes');`

src/components/PaymentWidgetMixin.ts
```diff
@@ -58,23 +59,26 @@ export const PaymentWidgetMixin = {
...
     },
-    setMaxValue: function(target: string, max?: number): void {
+    setMaxValue: function(target: ResourcesWithRates | 'science', max?: number): void {
       let currentValue: number = (this as any)[target];
       const cardCost: number = (this as any).$data.cost;
       let amountHave: number = max ?? (this as any).player[target];
 
       let amountNeed: number = cardCost;
-      if (['titanium', 'steel', 'microbes', 'floaters'].includes(target)) {
+      if (!(target === 'science')) {
```
bafolts, line 81: The `target` will also be `heat`. When it is `heat` this block used to return `false` and will now return `true`.

src/components/SelectHowToPay.ts
```diff
@@ -169,7 +170,7 @@ export const SelectHowToPay = Vue.component('select-how-to-pay', {
       }
 
       const requiredAmt = this.playerinput.amount || 0;
-      const totalSpentAmt = htp.heat + htp.megaCredits + (htp.steel * this.player.steelValue) + (htp.titanium * this.player.titaniumValue) + (htp.microbes * 2) + (htp.floaters * 3);
+      const totalSpentAmt = htp.heat + htp.megaCredits + (htp.steel * this.player.steelValue) + (htp.titanium * this.player.titaniumValue);
```
bafolts, line 173: Great change. Checks out.

src/components/common/Button.vue
```diff
@@ -74,6 +74,10 @@ export default Vue.extend({
       // type
       if (this.type === 'max') {
         classes.push('btn-max'); // #TODO max_button?
+      } else if (this.type === 'plus') {
+        classes.push('btn-plus');
```
bafolts, line 78: Does this `btn-plus` classname exist in CSS? I can't find it.

bafolts (changes requested):
Love seeing a component test. Want to learn more about btn-plus and btn-minus. Reading more I think they are related to the component test. Spotted at least one issue in `PaymentWidgetMixin` will need to address before merging.

src/components/PaymentWidgetMixin.ts
kberg: Thanks.

src/components/PaymentWidgetMixin.ts
kberg: Eep!

kberg:
> Love seeing a component test. Want to learn more about btn-plus and btn-minus. Reading more I think they are related to the component test. Spotted at least one issue in `PaymentWidgetMixin` will need to address before merging.

btn-minus and btn-plus are added so it's easier for the test to find those buttons and press them.

bafolts (approved):

Issue #3398, opened 2021-07-01
[Moon] Match parameter order between sidebar and tracks


kberg:
New Moon parameter indicator on the left sidebar: The Order of the numbers from the moon parameter indicator (left bar) should have the same order as the coresponding tracks on the moon map or vise versa

kberg:
Fixed.

PR #3426, opened 2021-07-08
Two moon board changes

kberg:
1. The second colony rate reward is now 1 energy production. This required a lot of css changes to shift everything to the right.
2. The sidebar values are in the same order with the  three graphs, plus a little color.


kberg:
image

kberg:
image

kberg:
Fixes #3398

src/components/MoonGlobalParameterValue.ts
```diff
@@ -29,7 +25,9 @@ export const MoonGlobalParameterValue = Vue.component('moon-global-parameter-val
       <img src="/assets/misc/checkmark.png" class="checkmark" :alt="$t('Completed!')">
     </div>
     <div v-else class="moon_params_value">
-      {{this.value()}}
+      <span class="colony">{{this.moonData.colonyRate}}</span>
```
bafolts, line 28: We do it all the time but fwiw I don't think it is recommended to use `this` within a vue template. The linter will eventually enforce not using `this.` in templates once we update to the recommended vue rules.

bafolts (approved):

src/components/MoonGlobalParameterValue.ts
kberg: OK I removed it and playtested it4.

PR #3430, opened 2021-07-09
Robotic Workforce Gyropolis energy check

vsrisuknimit:
Check that the player has enough energy to duplicate with Robotic Workforce.

### DIFF
--- a/src/cards/base/RoboticWorkforce.ts
+++ b/src/cards/base/RoboticWorkforce.ts
@@ -42,6 +42,9 @@ export class RoboticWorkforce extends Card implements IProjectCard {
     if (card.name === CardName.HEAT_TRAPPERS) {
       return player.game.someoneHasResourceProduction(Resources.HEAT, 2);
     }
+    if (card.name === CardName.GYROPOLIS) {
+      return player.getProduction(Resources.ENERGY) >= 2;
+    }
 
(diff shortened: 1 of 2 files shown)


src/cards/base/RoboticWorkforce.ts
```diff
@@ -42,6 +42,9 @@ export class RoboticWorkforce extends Card implements IProjectCard {
     if (card.name === CardName.HEAT_TRAPPERS) {
       return player.game.someoneHasResourceProduction(Resources.HEAT, 2);
     }
+    if (card.name === CardName.GYROPOLIS) {
```
kberg, line 45: This shouldn't be necessary since Gyropolis has a card.produce method below.

kberg: No, wait, that's not enough since the produce doesn't take energy production into account. Because canProduce can't be statically defined.

nwai90: It's good to add a unit test for it: https://github.com/nwai90/terraforming-mars/commit/356eb658d2680ac9dd05ef0813c13d12d5e8d69c

bafolts: Can we get unit test in this commit before merging?

bafolts (approved):

PR #3434, opened 2021-07-10
SFC for SelectProductionToLose and ShiftAresGlobalParameters.

kberg:
I need to manually re-test, but shouldn't take long.


kberg:
(Yeah, the manual re-test was surprisingly easy as I'd set up the conditions already. :D)

alrusdi (approved):

PR #3435, opened 2021-07-10
Update MoonBoard.ts

kberg:
Forgot to close the element correctly.


bafolts (approved):

bafolts:
vetur or lint would have caught this once everything is sfc.

PR #3440, opened 2021-07-11
Centralize the wide list of inputs as a single type

kberg:
I think this is worthwhile, but if you don't think so, let me know?


bafolts:
This changes makes sense to me. Since vue.js doesn't have a way to extend class or implement interfaces with the components I don't think we can use an interface or base class here instead.

PR #3442, opened 2021-07-13
Ecological Survey and Geological Survey stopped working with normal placement bonuses.

kberg:


tests/cards/base/EcologicalZone.spec.ts
```diff
@@ -18,7 +18,7 @@ describe('EcologicalZone', function() {
     game = Game.newInstance('foobar', [player, redPlayer], player);
   });
 
-  it('Can\'t play', function() {
+  it('Cannot play', function() {
```
kberg, line 21: Didn't mean to include it but since it's harmless I'm not going to `git commit --amend`

nwai90 (approved):

src/Game.ts
```diff
@@ -1372,6 +1372,8 @@ export class Game implements ISerializable<SerializedGame> {
       player.addResource(Resources.TITANIUM, count, {log: true});
     } else if (spaceBonus === SpaceBonus.HEAT) {
       player.addResource(Resources.HEAT, count, {log: true});
+    } else {
+      throw new Error('Space bonus ' + spaceBonus + ' is not supported yet.');
```
nwai90, line 1376: @kberg This is the change that causes an error with `SpaceBonus.OCEAN` and possibly others. Should we revert just this part?

Issue #3373, opened 2021-06-17
LogPanel regex only goes to 9


kberg:
![image](https://user-images.githubusercontent.com/413481/122322741-f522b780-cef3-11eb-8fa7-7de3bcfd0d7e.png)

This also applies to one of the test classes I wrote and the proposal at #3359 which copy/pastes the algorithm.

PR #3443, opened 2021-07-13
Fix for logPanel regex only going up to 9

nwai90:
Screenshot 2021-07-13 at 11 37 32 AM

Closes https://github.com/bafolts/terraforming-mars/issues/3373


kberg (approved):
/headslap of course it's that simple. Nice.

PR #3447, opened 2021-07-16
single file component: AndOptions

bafolts:
This converts one of the components I didn't think would ever use a template. I added a unit test given it isn't as straight forward as other components.


kberg:
The file move didn't occur. Looks like delete/add.

tests/components/AndOptions.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+        },
+        showsave: true,
+        showtitle: true,
+      },
+    });
+    const buttons = component.findAllComponents({name: 'Button'});
+    await buttons.at(0).findAllComponents({
+      name: 'button',
+    }).at(0).trigger('click');
+    expect(savedData).to.deep.eq([['1'], ['1']]);
```
kberg, line 43: FYI, I think .eql is the same as .to.deep.eq. Don't bother, but I do think it's nice.

kberg (approved):
Much nicer!

src/components/AndOptions.vue
```diff
@@ -0,0 +1,80 @@
...
+  methods: {
+    ...TranslateMixin.methods,
+    playerFactorySaved: function(idx: number) {
+      return (out: Array<Array<string>>) => {
+        console.log('setting the value to', out[0]);
+        this.$data.responded[idx] = out[0];
+      };
+    },
+    saveData: function() {
+      console.log(this);
```
bafolts, line 68: Will remove before flight, was debugging.

PR #3449, opened 2021-07-16
Don't place Moon tiles during WGT, just raise the rates.

### LINKED ISSUE #3448: Make Moon WGT bonuses raise parameters instead of placing tiles + raise parameters
Playing The Moon with WGT means placing tiles on The Moon as the WGT action. The problem is that's free tile placement which can give players big VP bonuses, which seems to be against the spirit of WGT.

Because WGT for The Moon was not made by the designer who realized after the fact that yes, Moon parameters should be part of WGT, I feel it is acceptable to make this change.

kberg:
I will follow up with a couple more tests, as I noticed that the actual WGT operations aren't tested.

Fixes #3448

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -65,9 +65,6 @@ import {ConvertPlants} from './cards/base/standardActions/ConvertPlants';
 import {ConvertHeat} from './cards/base/standardActions/ConvertHeat';
 import {Manutech} from './cards/venusNext/Manutech';
 import {LunaProjectOffice} from './cards/moon/LunaProjectOffice';
-import {PlaceMoonMineTile} from './moon/PlaceMoonMineTile';
-import {PlaceMoonColonyTile} from './moon/PlaceMoonColonyTile';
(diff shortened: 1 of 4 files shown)


src/Player.ts
```diff
@@ -1124,26 +1121,26 @@ export class Player implements ISerializable<SerializedPlayer> {
     MoonExpansion.ifMoon(game, (moonData) => {
       if (moonData.colonyRate < constants.MAXIMUM_COLONY_RATE) {
         action.options.push(
-          new SelectOption('Place a colony tile on the Moon', 'Increase', () => {
```
bafolts, line 1127: This will impact translations.

bafolts (approved):

src/Player.ts
kberg: Ah right I forgot. Thanks.

kberg: While I think we'll be mostly okay, new translations are just going to be necessary. Looks like it, cn, are the only two languages.

PR #3450, opened 2021-07-16
Centralize parsing the log message.

kberg:
I verified that this works in a local game in English and another langauge. Plus the unit tests themselves rely on the TestUtils to work as expected.

### DIFF
--- a/src/Log.ts
+++ b/src/Log.ts
@@ -1,0 +1,10 @@
+import {LogMessageData} from './LogMessageData';
+import {Message} from './Message';
+
+export class Log {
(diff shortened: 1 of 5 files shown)


bafolts (approved):

nwai90 (approved):
👍 I might put the applyData method in `LogHelper.ts` instead of creating a new `Log.ts` file with a single method, but otherwise LGTM

kberg:
I thought about that. LogHelper is a server-side only piece of code, and so does not belong being included on the client-side. (The fact that nearly the entire server implementation is packaged into the client is a bit of a code smell.

PR #3451, opened 2021-07-16
Prototype: report when log message placeholders don't match data parameters.

kberg:
I found a case where log message parameters don't match their data, and thought: surely I can write a test for this. And so you see what I've done so far. A prototype that's right in the path for logging messages. I don't mind it _terribly_ but I'd rather do something that allows us to replace Game.log in tests, or something. Thoughts?

(PS that bit of extra logging should be removed; it leaks data players shouldn't see.)

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -66,6 +66,7 @@ import {Random} from './Random';
 import {MilestoneAwardSelector} from './MilestoneAwardSelector';
 import {BoardType} from './boards/BoardType';
 import {Multiset} from './utils/Multiset';
+import {Log} from './Log';
 
 export type GameId = string;
 export type SpectatorId = string;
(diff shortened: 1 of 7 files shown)


kberg:
Depends on #3450

src/Game.ts
```diff
@@ -1477,7 +1478,19 @@ export class Game implements ISerializable<SerializedGame> {
     if (f) {
       f(builder);
     }
+
+    const intest = typeof (global as any).it === 'function';
```
bafolts, line 1482: In other classes we have leveraged a test class which extends the class. In this case we could create a `TestGame` class that ran this check. If creating a `TestGame` class is a lot of work one could also override `Game.prototype.log` which could run this check. I would prefer doing something in the tests versus adding this to the source to check the tests.

If in the tests you were to write.

```
const originalGameLog = Game.prototype.log;
Game.prototype.log = function (message, f, options) {
   // run the check throwing error here
   originalGameLog.apply(this, arguments);
}
```
This would avoid having to modify the source.

PR #3453, opened 2021-07-16
export_game now accepts a player id

kberg:


src/database/PostgreSQL.ts
```diff
@@ -116,6 +116,26 @@ export class PostgreSQL implements IDatabase {
     });
   }
 
+  // TODO(kberg): throw an error if two game ids exist.
+  getGameId(playerId: string, cb: (err: Error | undefined, gameId?: GameId) => void): void {
+    const sql =
+      `SELECT game_id
+      FROM games, json_array_elements(CAST(game AS JSON)->'players') AS e
```
bafolts, line 123: With the number of rows in production I could see this query running slow as it has to parse the JSON on every row. This isn't exposed through an HTTP route so only an admin should hit it I suppose. We eventually need to make a table that maps player id to game id which would be far more efficient.

kberg: Fortunately we have a few things in our favor: this isn't used in the node app so node's single threadiness doesn't apply here. The save_id=0 made the query much, much faster, about 3-4 seconds. 

If the database record was of type 'jsonb' we could add json indexes, and everything you just said about special tables would not need to apply.

kberg: That said, your comment is on the money.

bafolts (approved):

PR #3461, opened 2021-07-19
A prototype of rewriting the model.

kberg:


kberg:
This is now reasonably out of date and served its purpose.

PR #3463, opened 2021-07-20
single file component: PlayerStatus

bafolts:


src/components/overview/PlayerInfo.ts
```diff
@@ -73,7 +70,7 @@ export const PlayerInfo = Vue.component('player-info', {
       }
       for (let i = 0; i < hiddenPlayersIndexes.length; i++) {
         if (hiddenPlayersIndexes.includes(i)) {
-          hidePlayerData(this.$root as unknown as typeof mainAppSettings.methods, i);
+          (this.$root as unknown as typeof mainAppSettings.methods).setVisibilityState('pinned_player_' + i, false);
```
kberg, line 73: I see why you are reintroducing duplication. Can you just add a TODO to find a way to deal with common code?

kberg (approved):
Minor request, otherwise lgtm

src/components/overview/PlayerInfo.ts
bafolts: Need to see if the single file component pattern will allow us to gain type safety with `this.$root`.

kberg:
It doesn't look as if this test was run. :'-(

bafolts:
> It doesn't look as if this test was run. :'-(

The checks show the tests run.

kberg:
This specific one? Because it wasn't passing. Help me understand, am I
missing something?

>

bafolts:
Ah I was missing something. I didn't realize this test wasn't running. As my generation would say, "my bad".

kberg:
My generation says it now, too. :b

PR #3465, opened 2021-07-21
Change Syndicate Pirate Raids description based on feedback.

kberg:


kberg:
Gonna close this one. I think the card is fine as is, at least for now.

PR #3468, opened 2021-07-23
Remove incorrect line in Water Treatment Complex

nwai90:
Bug report from Discord. The card does not seem like it should decrease plant production. The reserved unit cost of 1 Ti is already covered by `player.deductUnits(adjustedReserveUnits)` in `ProjectCard`


kberg (approved):
Yipe! What was that doing there?

PR #3470, opened 2021-07-23
Fixed issues with purging old running games in sqlite database

AndrewSav:
- made sqlite implementation consistent with postgresql one in that by default running games are purged after 10 days
- fixed the purge query to work

### DIFF
--- a/src/database/SQLite.ts
+++ b/src/database/SQLite.ts
@@ -132,14 +132,14 @@ export class SQLite implements IDatabase {
   }
 
   purgeUnfinishedGames(): void {
-    // Purge unfinished games older than MAX_GAME_DAYS days. If this .env variable is not present, unfinished games will not be purged.
-    if (process.env.MAX_GAME_DAYS) {
-      this.db.run('DELETE FROM games WHERE created_time < strftime(\'%s\',date(\'now\', \'-? day\')) and status = \'running\'', [process.env.MAX_GAME_DAYS], function(err: Error | null) {
-        if (err) {
-          return console.warn(err.message);
-        }
-      });
-    }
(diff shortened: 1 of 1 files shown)


nwai90 (approved):
LGTM. @bafolts Do you recall any reason why SQLite does not use the `MAX_GAME_DAYS` variable and instead keeps unfinished games in perpetuity?

bafolts:
It has been this way since it was first introduced [here](https://github.com/bafolts/terraforming-mars/pull/1556). Bringing this inline with Postgres makes sense. @AndrewSav the other option would be to have the postgres database not purge by default. Did you run into a situation where not purging the games posed a problem?

AndrewSav:
@bafolts not technical problems. The game list screen grows uncontrollably is all. As long as there is the env var to control this I'm fine either way, let me know if you want me to make the changes, but if we do that will be incompatible and could be surprising to people. 

I personally would support https://github.com/bafolts/terraforming-mars/pull/1377 but consensus was that it's not happening and (one of) the reason was given is auto-cleanup after 10 days. So I just wanted _something_ that would allow me to trim on the number of abandoned games.

bafolts:
Fixing the SQLite query only with this pull request will avoid changing the default behavior. Can you change this PR to only fix the SQLite query for purging?

AndrewSav:
@bafolts sorry, I do not understand what you are asking. This PR does not touch the Postgres code it's already limited to SQLite only. If you mean the 10 days default, then this is not what I was referring to. Making both database implementations consistent here, IMO is a worthy goal. I did not work before anyway, so it won't make anyone surprised, the default behaviour will also not change.

PR #3472, opened 2021-07-23
Move help files to SFC

kberg:


src/components/help/HelpIconology.vue
```diff
@@ -5,7 +5,7 @@ export const HelpIconology = Vue.component('help-iconology', {
   },
   methods: {
   },
-  template: `
+<template>
```
bafolts, line 8: This `<template>` needs to be moved.

bafolts (approved):

PR #3474, opened 2021-07-24
single file component: PlayerTags

bafolts:


src/components/overview/PlayerInfo.ts
```diff
@@ -39,7 +39,7 @@ export const PlayerInfo = Vue.component('player-info', {
   components: {
     Button,
     'player-resources': PlayerResources,
-    'player-tags': PlayerTags,
```
kberg, line 42: I'm not clear when we're doing this and when we're not (removing the component string key, that is.)

kberg (approved):

PR #3475, opened 2021-07-24
An attempt to fix #2802

### LINKED ISSUE #2802: /api/games returns no games until a game action occurs

AndrewSav:
Hello, everyone,

this is an attempt to fix #2802, unfortunately it comes with rather major refactoring of the `GameLoader` class.  Please be gentle with me, because I have not written for nodejs before, and I have not used typescript before, so I have certain gaps in the knowledge. I do understand how nodejs works in principle, though.

I tried to keep the spirit of the functionality of the class intact, and only change what is required for better flow, and for adding the [necessary callback](https://github.com/bafolts/terraforming-mars/pull/2807#discussion_r579861467) for fixing the issue.

The need to refactor the `GameLoader` class came from the desire to streamline and simplify state tracking and callbacks that were used there before the refactoring.

The main idea behind the refactoring is that we moved the following entities in their own class, `GameIds`:

- `loadAllGameIds` function
- `games` field
- `participantIds` field

This new `GameIds` class is an [EventEmitter](https://nodejs.org/api/events.html#events_class_eventemitter), and its main goal is to make sure that loading game ids from the database happens before they are used in the `GameLoader` class.  Thie `GameIds` class has the `load` function, which is called from the constructor of the `GameLoader` class. It also has the `getGames` function which returns a promise which is only resolved when the database load has finished. The promise contains `games` and  `participantIds` fields, I mentioned above.

With this promise based flow, we now can get rid of the following entities of the old class:

- `State` enum
- `state` field
- `loadingGame` field
- `onGameIdsLoaded` field
- `onGameLoaded` function
- `runGameIdLoadedCallback` function
- `allGameIdsLoaded` function

This significantly simplifies the flow and improves the class readability.

#2802 also warrants a single interface change for the Game Loader, and this is the callback for the `getLoadedGameIds` function as was discussed in the earlier PR linked above. This interface change feels self-explanatory, but please do not hesitate to ask if you need this or anything else in the PR clarified.

Naturally, `FakeGameLoader` from tests needed a corresponding change to conform the interface change. Same can be said about the `ApiGames` class.

Finally, there is a question of tests. My belief (which I'm happy to be corrected in if I'm wrong), is that the tests that I have changed always had a race. That is, it's because of luck/timing they have not failed before. There were two major problems with the tests in `GameLoader.spec.ts`.

Firstly, consider this code from the current file (before the PR):

```
  it('gets undefined when game will never exist', function() {
    let actualGame1: Game | undefined = game;
    let actualGame2: Game | undefined = game;
    GameLoader.getInstance().getByPlayerId('foobar', (game1) => {
      actualGame1 = game1;
    });
    GameLoader.getInstance().getByPlayerId('never', (game2) => {
      actualGame2 = game2;
    });
    GameLoader.getInstance().getByGameId('foobar', false, (game1) => {
      expect(game1).not.is.undefined;
    });
    expect(actualGame1).is.undefined;
    expect(actualGame2).is.undefined;
  });
```

In here the last two `expect` lines are not guaranteed to run after the call backs above. They can run before, which was the case in my testing. Because of this I had to move all similar `expect` occurrences in this the file inside the callbacks.

Secondly, the following excerpt may also problematic:

```
  it('gets no game when game goes missing from database', function() {
    const originalGetGame = Database.getInstance().getGame;
    GameLoader.getInstance().getByGameId('never', false, (game1) => {
      expect(game1).is.undefined;
      Database.getInstance().getGame = function(_gameId: string, cb: (err: any, serializedGame?: SerializedGame) => void) {
        cb(undefined, undefined);
      };
    });
    GameLoader.getInstance().getByGameId('foobar', false, (game1) => {
      expect(game1).is.undefined;
    });
    Database.getInstance().getGame = originalGetGame;
  });
```
What happens here, is that we try to substitute `Database.getInstance().getGame` with a temporary function, and then restore it. Again, in my tests, _subsequent tests_ (not this one) manage to run _before_ the restoration has taken place, and were running "bad" substituted function causing failed tests.

I had do use the `done` parameter in those test to make sure this does not happen.

Apologies, that this is too lengthy, the intention is to help, and share work. Please let me know your thoughts.

### DIFF
--- a/src/database/GameIds.ts
+++ b/src/database/GameIds.ts
@@ -33,9 +33,10 @@ export class GameIds extends EventEmitter {
   }
 
   private async getAllInstances(allGameIds: Array<GameId>) : Promise<void> {
-    for (const gameId of allGameIds) {
-      await this.getInstance(gameId);
-    }
+    const promises = allGameIds.map((x) => {
+      this.getInstance(x);
+    });
+    await Promise.all(promises);
   }
 
   public load() : void {
--- a/src/database/GameIds.ts
+++ b/src/database/GameIds.ts
@@ -32,11 +32,10 @@ export class GameIds extends EventEmitter {
     });
   }
 
-  private async getAllInstances(allGameIds: Array<GameId>) : Promise<void> {
-    const promises = allGameIds.map((x) => {
-      this.getInstance(x);
-    });
-    await Promise.all(promises);
+  private async getAllInstances(allGameIds: Array<GameId>) : Promise<void[]> {
+    return Promise.all(allGameIds.map((x) => {
+      return this.getInstance(x);
+    }));
   }
 
   public load() : void {
--- a/src/database/GameIds.ts
+++ b/src/database/GameIds.ts
@@ -1,0 +1,62 @@
+import {Database} from './Database';
+import {Game, GameId, SpectatorId} from '../Game';
+import {PlayerId} from '../Player';
+import {once} from 'events';
+import {EventEmitter} from 'events';
+
+export class GameIds extends EventEmitter {
+  private loaded = false;
+  private readonly games = new Map<GameId, Game | undefined>();
+  private readonly participantIds = new Map<SpectatorId | PlayerId, GameId>();
+  private getInstance(gameId: GameId) : Promise<void> {
+    return new Promise<void>((resolve) => {
+      Database.getInstance().getGame(
+        gameId,
+        (err, game) => {
+          if (err || (game === undefined)) {
+            console.error(`unable to load game ${gameId}`, err);
+          } else {
+            console.log(`load game ${gameId} with ${game.spectatorId}`);
+            if (this.games.get(gameId) === undefined) {
+              this.games.set(gameId, undefined);
+              if (game.spectatorId !== undefined) {
+                this.participantIds.set(game.spectatorId, gameId);
+              }
+              for (const player of game.players) {
+                this.participantIds.set(player.id, gameId);
+              }
+            }
+          }
+          resolve();
+        });
+    });
+  }
+
+  private async getAllInstances(allGameIds: Array<GameId>) : Promise<void> {
+    for (const gameId of allGameIds) {
+      await this.getInstance(gameId);
+    }
+  }
+
+  public load() : void {
+    Database.getInstance().getGames((err, allGameIds) => {
+      if (err) {
+        console.error('error loading all games', err);
+        this.loaded = true;
+        this.emit('loaded');
+        return;
+      }
+      this.getAllInstances(allGameIds).then(() =>{
+        this.loaded = true;
+        this.emit('loaded');
+      });
+    });
+  }
+
+  public async getGames(): Promise<{games:Map<GameId, Game | undefined>, participantIds:Map<SpectatorId | PlayerId, GameId>}> {
+    if (!this.loaded) {
+      await once(this, 'loaded');
+    }
+    return {games: this.games, participantIds: this.participantIds};
+  }
+}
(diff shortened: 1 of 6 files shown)


bafolts:
No fear! I restart the terraforming-mars.herokuapp.com server most mornings and this code is crucial to that restart going well. It will take me time to read through this. At quick glance nothing standing out as a problem.

src/database/GameIds.ts
```diff
@@ -0,0 +1,62 @@
...
+            }
+          }
+          resolve();
+        });
+    });
+  }
+
+  private async getAllInstances(allGameIds: Array<GameId>) : Promise<void> {
+    for (const gameId of allGameIds) {
+      await this.getInstance(gameId);
```
bafolts, line 37: These _could_ run in parallel. Was this intended for them to run sequentially?

AndrewSav: I considered this and decided it does not really matter. Do you think it does?

bafolts: I restart the demo server terraforming-mars.herokuapp.com most mornings. While the server is restarting it will need to wait for these game ids to finish loading before doing anything. Having to do these one at a time will slow down how long that takes. I think it is worth using a `Promise.all` here and running these in parallel.

AndrewSav: You are probably right, I just wanted to confirm that there is an evidence, that it's taking longer than required. Can you confirm that you tried it and it significantly slowed things down? On a side note, if performance here is an issue, then we probably want to change the database interface to query for this information in one go, rather than in `N + 1` calls. More work, of course than just updating with `Promise.all`.

AndrewSav: Alternatively you can give me a copy of your Postgres database, so I could test it myself.

AndrewSav: I've made the requested change

bafolts:
I am going to test this with the terraforming-mars.herokuapp.com database.

src/database/GameIds.ts
```diff
@@ -0,0 +1,63 @@
...
+            }
+          }
+          resolve();
+        });
+    });
+  }
+
+  private async getAllInstances(allGameIds: Array<GameId>) : Promise<void> {
+    const promises = allGameIds.map((x) => {
+      this.getInstance(x);
```
bafolts, line 37: This function needs to return the `Promise`. This is returning `undefined`. This leads to the function returning before every `getInstance` promise has resolved.

bafolts (changes requested):
Tested this with postgres data from terraforming-mars.herokuapp.com. The `getAllInstances` running in parallel needs to be updated to wait for each Promise.

src/database/GameIds.ts
AndrewSav: Yes, I see the problem. Thanks.

AndrewSav:
> Tested this with postgres data from terraforming-mars.herokuapp.com. The `getAllInstances` running in parallel needs to be updated to wait for each Promise.

I this different from the review above or is this the same issue?

bafolts:
> > Tested this with postgres data from terraforming-mars.herokuapp.com. The `getAllInstances` running in parallel needs to be updated to wait for each Promise.
> 
> I this different from the review above or is this the same issue?

Same issue.

src/database/GameIds.ts
AndrewSav: I think I need to add a test to catch this, which may be quite difficult due to timings. Any suggestions?

bafolts: If a unit test were to check when a call to `getAllInstances` resolved that every game id has loaded to would be a good test. I recreated this by calling the api route to get game ids. I would see it return a partial list of game ids and more game ids every time I refreshed. It should return every game id once they have all loaded.

AndrewSav: This has now been addressed

bafolts (approved):

PR #3476, opened 2021-07-24
fix interplanetary trade

vsrisuknimit:
Planetary Trade (card) should not count Venus tag or Moon tag by default.

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -831,7 +831,10 @@ export class Player implements ISerializable<SerializedPlayer> {
       }
     }
     if (countWild) {
-      return uniqueTags.size + wildcardCount;
+      let maxTagCount = constants.DEFAULT_TAG_COUNT;
+      if (this.game.gameOptions.venusNextExtension) maxTagCount++;
+      if (this.game.gameOptions.moonExpansion) maxTagCount++;
+      return Math.min(uniqueTags.size + wildcardCount, maxTagCount);
     } else {
       return uniqueTags.size;
     }
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -831,7 +831,7 @@ export class Player implements ISerializable<SerializedPlayer> {
       }
     }
     if (countWild) {
-      let maxTagCount = constants.DEFAULT_TAG_COUNT;
+      let maxTagCount = 10;
       if (this.game.gameOptions.venusNextExtension) maxTagCount++;
       if (this.game.gameOptions.moonExpansion) maxTagCount++;
       return Math.min(uniqueTags.size + wildcardCount, maxTagCount);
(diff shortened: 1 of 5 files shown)


src/constants.ts
```diff
@@ -4,6 +4,7 @@ export const MILESTONE_COST: number = 8;
 export const MAX_MILESTONES = 3;
 export const AWARD_COSTS: Array<number> = [8, 14, 20];
 export const MAX_AWARDS = 3;
+export const DEFAULT_TAG_COUNT = 10;
```
kberg, line 7: This only applies to the one card and it's ready. Seems like it should go there and not be a global constant. Are there other uses for this? If not, please hang it off the card.

src/cards/promo/InterplanetaryTrade.ts
```diff
@@ -30,8 +31,10 @@ export class InterplanetaryTrade extends Card implements IProjectCard {
   public play(player: Player) {
     // This card tag is counting as well
     const availableTags = player.getDistinctTagCount(true, Tags.SPACE);
-    // Only count wildcards up to the max amount of tag types existing (minus events and wildcards)
-    const existingTags = Object.keys(Tags).length - 2;
+    // There is only 10 tags in the base game. One more for Venus and one more for Moon.
+    let existingTags = DEFAULT_TAG_COUNT;
+    if (player.game.gameOptions.venusNextExtension) existingTags++;
```
nwai90, line 36: While the code works, it is currently placed in the wrong scope, and would be better moved into the `getDistinctTagCount` function in `Player.ts` to handle any existing or future scenarios that require similar logic. Ideally `player.getDistinctTagCount(true, Tags.SPACE)` should return the correct result by considering which expansions are in play, i.e.

Screenshot 2021-07-24 at 11 05 01 PM

kberg: Good catch. That makes senes.

src/Player.ts
```diff
@@ -831,7 +831,10 @@ export class Player implements ISerializable<SerializedPlayer> {
       }
     }
     if (countWild) {
-      return uniqueTags.size + wildcardCount;
+      let maxTagCount = constants.DEFAULT_TAG_COUNT;
```
kberg, line 834: This is the right place for this logic, but please, still remove the constant unless you expect it to be used elsewhere.

src/constants.ts
bafolts: If we ever find another use case for this number we could create a constant at that time.

bafolts (changes requested):
Can you change the `DEFAULT_TAG_COUNT` from a constant to a local variable?

nwai90 (approved):

bafolts (approved):

PR #3479, opened 2021-07-25
single file component: OrOptions

bafolts:
Of the remaining components to convert this is probably the most challenging. Plan to add unit tests similar to `AndOptions`. Creating draft pull request so someone doesn't attempt this simultaneously.


kberg (approved):
Well done.

bafolts:
I want to test what this does when there are nested `OrOptions`. I don't trust vue-js to do the right thing in regards to the radio buttons.

bafolts:
I tested this by creating a nested `OrOptions` and it seemed to work even without providing the `name` to the form element but I couldn't explain why it worked. I added the `name` attribute for my sanity.

PR #3484, opened 2021-07-26
speed up the build for webpack

bafolts:
Since switching to single file components the build time has increased. This should improve the situation.

https://github.com/TypeStrong/fork-ts-checker-webpack-plugin

Found this on the world wide web while looking for solutions to speed up build.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ dist/
 pid
 *.log
 .eslintcache
+tsconfig.tsbuildinfo
 
 # Local Packages
 node_modules/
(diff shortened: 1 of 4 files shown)


kberg:
> This should improve the situation.

Before:

```
real    1m13.151s
user    1m46.788s
sys     0m1.203s
```

After:

```
real    0m27.924s
user    1m1.822s
sys     0m1.077s
```

kberg (commented):
I can't reasonably verifiy this is correct, but see my speed test above.

kberg (approved):

bafolts:
This should lead to the same generated code. The `transpileOnly` option tells typescript to not bother checking types. The new dependency and plugin checks the types in a separate process. This allows it to run in parallel which webpack is transpiling with ts-loader. I added an obvious typo to one of the `.vue` files and saw the added plugin check the type and fail the build.

PR #3487, opened 2021-07-26
Move CreateGameForm to SFC

kberg:
Still needs some work, but I'm licking the cookie.


bafolts (changes requested):
When I run this locally I see warnings logged to the console.

```
vue.esm.js:628 [Vue warn]: Duplicate keys detected: '1'. This may cause an update error.

found in

---> <CreateGameForm> at src/components/create/CreateGameForm.vue
       <Root>
```

Should address these warnings before merging.

kberg:
> When I run this locally I see warnings logged to the console.
> 
> ```
> vue.esm.js:628 [Vue warn]: Duplicate keys detected: '1'. This may cause an update error.
> 
> found in
> 
> ---> <CreateGameForm> at src/components/create/CreateGameForm.vue
>        <Root>
> ```
> 
> Should address these warnings before merging.

Oh, good catch.  Sorry - I usually leave my console open.

bafolts:
Something is off with the color selection for multiple players:

before:

image

after:

image

kberg:
Oh this is interesting. This is due to adding the <div> to support v-bind.
OK I'll have to dig into this.

>

kberg:
> Oh this is interesting. This is due to adding the <div> to support v-bind. OK I'll have to dig into this.
> […](#)

OK, PTAL.

bafolts (approved):

PR #3496, opened 2021-07-29
single file component: SelectCard

bafolts:


src/components/OwnerModel.ts
```diff
@@ -0,0 +1,6 @@
+import {Color} from '../Color';
+
+export interface OwnerModel {
```
kberg, line 3: Independent of anything, this file doesn't make sense. Can you add some comments, please?

kberg (approved):

PR #3498, opened 2021-07-29
improved GameLoader tests

AndrewSav:
During my last [PR](https://github.com/bafolts/terraforming-mars/pull/3475) I have borked a few of GameLoader tests, which was not caught during the review, because they were passing...

This PR is fixing this situation. Problems being fixed:

- Most of the test need to become async, because the API became truly async, it used to be a sync with callback before, and the tests relied on it
- In the PR I did a very poor job addressing the above, so many tests became impossible to fail they would come up green no matter what
- Those tests that _would_ fail would give a generic timeout error instead of the real failure reason
- While PR did an attempt to fix problems with async substitution/restoration of the mock database interface functions, it did not do it correctly in all cases

Only the tests has been changing, the application code is not touched in this PR. Any questions, please feel free to ask.

### DIFF
--- a/tests/database/GameLoader.spec.ts
+++ b/tests/database/GameLoader.spec.ts
@@ -14,6 +14,7 @@ describe('GameLoader', function() {
   const player = TestPlayers.BLUE.newPlayer();
   const player2 = TestPlayers.RED.newPlayer();
   const game = Game.newInstance('foobar', [player, player2], player);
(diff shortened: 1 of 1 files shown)


bafolts (approved):
While reviewing I was actually thinking if the tests would run anymore since I saw `expect` inside code which may not execute before the test finished. Good to see this improvement!

PR #3502, opened 2021-07-30
fixed syntax error in the SQLite purge query

AndrewSav:
Since https://github.com/bafolts/terraforming-mars/pull/3470 was reverted, this is a lighter fix that does not change the timing. It just fixes the syntax error in the SQLite query.

Note: the timing issue might be present both before and after the fix and will need to be addressed separately.

### DIFF
--- a/src/database/SQLite.ts
+++ b/src/database/SQLite.ts
@@ -134,7 +134,7 @@ export class SQLite implements IDatabase {
   purgeUnfinishedGames(): void {
     // Purge unfinished games older than MAX_GAME_DAYS days. If this .env variable is not present, unfinished games will not be purged.
     if (process.env.MAX_GAME_DAYS) {
-      this.db.run('DELETE FROM games WHERE created_time < strftime(\'%s\',date(\'now\', \'-? day\')) and status = \'running\'', [process.env.MAX_GAME_DAYS], function(err: Error | null) {
+      this.db.run(`DELETE FROM games WHERE created_time < strftime('%s',date('now', '-' || ? || ' day')) and status = 'running'`, [process.env.MAX_GAME_DAYS], function(err: Error | null) {
         if (err) {
           return console.warn(err.message);
         }


src/database/SQLite.ts
```diff
@@ -134,7 +134,7 @@ export class SQLite implements IDatabase {
   purgeUnfinishedGames(): void {
     // Purge unfinished games older than MAX_GAME_DAYS days. If this .env variable is not present, unfinished games will not be purged.
     if (process.env.MAX_GAME_DAYS) {
-      this.db.run('DELETE FROM games WHERE created_time < strftime(\'%s\',date(\'now\', \'-? day\')) and status = \'running\'', [process.env.MAX_GAME_DAYS], function(err: Error | null) {
+      this.db.run('DELETE FROM games WHERE created_time < strftime(\'%s\',date(\'now\', \'-\' || ? || \' day\')) and status = \'running\'', [process.env.MAX_GAME_DAYS], function(err: Error | null) {
```
kberg, line 137: This is good.  This seems a little unnecessary, but this might be easier to read if sql wrapped in backticks, eg

const sql = `DELETE FROM games WHERE created_time < strftime('%s',date('now', '-' || ? || ' day')) and status = 'running'`;

kberg (commented):
I'll leave approval up to @bafolts.

src/database/SQLite.ts
AndrewSav: Github markdown screws up with back ticks, can you fix your example please so I could understand the idea? You will need to use two backticks to surround your back-ticked text:  e.g.  ``` `` `back-ticked test` `` ``` gives you `` `back-ticked test` ``.

AndrewSav: I think I got it - I've made the requested change.

kberg: Yes, that - thank you!

bafolts (approved):

PR #3504, opened 2021-07-31
SelectHowToPay to SFC.

kberg:


kberg:
I just reconfirmed that it works as expected.

bafolts (approved):

PR #3506, opened 2021-07-31
SelectHowToPayForProjectCard as SFC.

kberg:
Don't approve this. I want to break it into smaller pieces -- one that rewrites the testing, and also have this follow behind SelectHowToPay. But at least now this is here.


kberg:
woo!

bafolts (approved):

PR #3508, opened 2021-07-31
use string for key with v-for

bafolts:
Fixes a vue warning.

```
vue.esm.js:628 [Vue warn]: Duplicate keys detected: '[object Object]'. This may cause an update error.

found in

---> <PreferencesDialog> at src/components/PreferencesDialog.vue
       <Sidebar> at src/components/Sidebar.vue
         <PlayerHome> at src/components/PlayerHome.vue
           <Root>
```


kberg (approved):
Good catch.

PR #3509, opened 2021-07-31
Vastly clarify and simplify the SHTP and SHTPFPC tests.

kberg:


tests/components/PaymentTester.ts
```diff
@@ -0,0 +1,85 @@
...
+import {expect} from 'chai';
+import {SelectHowToPayModel} from '../../src/components/PaymentWidgetMixin';
+
+export type Unit = 'heat' | 'steel' | 'titanium' | 'floaters' | 'microbes' | 'megaCredits' | 'science';
+
+
+export class PaymentTester {
+  private model: SelectHowToPayModel;
+  constructor(private wrapper: Wrapper<Vue>) {
+    this.model = this.wrapper.vm as unknown as SelectHowToPayModel;
```
kberg, line 12: This is disappointing -- if I could figure out a way to define wrapper as something like `Wrapper<Vue & SHTPM>` and thread that correctly through the tests I'd be happier. But for now, this is acceptable.

bafolts (approved):

Issue #3507, opened 2021-07-31
Wait for asynchronous database queries in constructors.


bafolts:
When creating SQLite the queries used to create the `games` table does not wait for the query to complete. This leads to situations where we may attempt to query the table before it exists. We should update IDatabase implementations to wait until the tables are created before running any queries.

Issue #3500, opened 2021-07-30
No such table: games


ap3h3ad:
Hi,

I cloned the project and started it, but I get this error:

https://abload.de/img/bug22kdq.png

Has anyone a clue what the problem could be?

ap3h3ad:
I don't know, it  just disappeared

kberg:
Verified this happens.

kberg:
This is probably due to #3475

kberg:
@bafolts @AndrewSav can you please take a look?

bafolts:
@kberg how did you recreate this error?

bafolts:
We call to create the database in the constructor of most `IDatabase` implementations but we don't necessarily wait until the operation has finished. This is prone to error.

kberg:
I cloned the repo from source, ran npm install ; npm run build ; npm run
start

>

bafolts:
I am pretty sure this was caused by https://github.com/bafolts/terraforming-mars/pull/3470. We call to `Database.getInstance().purgeUnfinishedGames()` now when we weren't before. This has exposed the issue with our `SQLite` database not necessarily being created before that function runs.

AndrewSav:
@bafolts I wonder if we could re-open it? It still will happen in SQLite with `MAX_GAME_DAYS` set. (and before https://github.com/bafolts/terraforming-mars/pull/3502 a different (syntax error) would happen instead)

kberg:
Sure thing.

bafolts:
Re-opening is fine. Probably duplicates or would be addressed by #3507

AndrewSav:
Oh yeah, I did nor realised you opened #3507.

bafolts:
I plan to put in a fix for this.

PR #3511, opened 2021-08-02
add initialize to IDatabase

bafolts:
Allows database to initialize. I started with implementation for SQLite. Can do the others in separate commits.

### DIFF
--- a/src/database/GameLoader.ts
+++ b/src/database/GameLoader.ts
@@ -13,7 +13,7 @@ type ListLoadCallback = (list: Array<{id: GameId, participants: Array<SpectatorI
  * Loads games from database sequentially as needed
  */
 export class GameLoader implements IGameLoader {
-  private static readonly instance = new GameLoader();
+  private static instance?: GameLoader;
 
   private idsContainer = new GameIds();
 
@@ -27,6 +27,9 @@ export class GameLoader implements IGameLoader {
   }
 
   public static getInstance(): IGameLoader {
+    if (GameLoader.instance === undefined) {
+      GameLoader.instance = new GameLoader();
+    }
     return GameLoader.instance;
   }
 
--- a/src/database/PostgreSQL.ts
+++ b/src/database/PostgreSQL.ts
@@ -42,6 +42,10 @@ export class PostgreSQL implements IDatabase {
     });
   }
 
+  async initialize(): Promise<void> {
+
+  }
+
   getClonableGames(cb: (err: Error | undefined, allGames: Array<IGameData>) => void) {
     const allGames: Array<IGameData> = [];
     const sql = 'SELECT distinct game_id game_id, players players FROM games WHERE save_id = 0 order by game_id asc';
--- a/src/database/IDatabase.ts
+++ b/src/database/IDatabase.ts
@@ -43,6 +43,11 @@ export type DbLoadCallback<T> = (err: Error | undefined, game: T | undefined) =>
 export interface IDatabase {
 
     /**
+     * Creates any tables needed
+     */
+    initialize(): Promise<void>;
+
+    /**
      * Pulls most recent version of game
      * @param game_id the game id to load
      * @param cb called with game if exists. If game is undefined err will be truthy.
--- a/src/database/LocalFilesystem.ts
+++ b/src/database/LocalFilesystem.ts
@@ -20,6 +20,10 @@ export class Localfilesystem implements IDatabase {
     }
   }
 
+  async initialize(): Promise<void> {
+
+  }
+
   _filename(gameId: string): string {
     return path.resolve(dbFolder, `${gameId}.json`);
   }
--- a/src/database/SQLite.ts
+++ b/src/database/SQLite.ts
@@ -18,8 +18,26 @@ export class SQLite implements IDatabase {
       fs.mkdirSync(dbFolder);
     }
     this.db = new sqlite3.Database(dbPath);
-    this.db.run('CREATE TABLE IF NOT EXISTS games(game_id varchar, players integer, save_id integer, game text, status text default \'running\', created_time timestamp default (strftime(\'%s\', \'now\')), PRIMARY KEY (game_id, save_id))');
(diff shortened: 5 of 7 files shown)


src/database/GameLoader.ts
```diff
@@ -13,7 +13,7 @@ type ListLoadCallback = (list: Array<{id: GameId, participants: Array<SpectatorI
  * Loads games from database sequentially as needed
  */
 export class GameLoader implements IGameLoader {
-  private static readonly instance = new GameLoader();
```
bafolts, line 16: This static property was self-executing. I moved this to execute when needed so that it can run after `Database::initialize`.

kberg (commented):
Is this getting complicated for the wrong reasons? I don't know enough but my code smell sense is going off.

AndrewSav:
> Is this getting complicated for the wrong reasons? I don't know enough but my code smell sense is going off.

Getting complicated in which way? Async is traditionally not very straightforward, and this is a nodejs web app. Having said that, is there anything in particular that looks especially complicated?

src/database/PostgreSQL.ts
```diff
@@ -42,6 +42,10 @@ export class PostgreSQL implements IDatabase {
     });
   }
 
+  async initialize(): Promise<void> {
+
+  }
```
AndrewSav, line 47: Are we sure that the empty body is sufficient here?

kberg: > > Is this getting complicated for the wrong reasons? I don't know enough but my code smell sense is going off.
> 
> Getting complicated in which way? Async is traditionally not very straightforward, and this is a nodejs web app. Having said that, is there anything in particular that looks especially complicated?

Good question. Well, what problem is the app currently exhibiting? I reported a startup issue (#2802) which we then only saw appeared with the filesystem database (if my recollection is correct), and so closed the bug. Are you saying that the issue also appears in other databases?

AndrewSav: > Good question. Well, what problem is the app currently exhibiting? I reported a startup issue (#2802) which we then only saw appeared with the filesystem database (if my recollection is correct), and so closed the bug. Are you saying that the issue also appears in other databases?

The issues it aims to resolve are on the sidebar of this page those are #3500 and #3507.

bafolts: This change is ensuring that database tables are created before we query them. The bug that appeared in #3500 was a symptom of a larger problem. If you delete the SQLite database file and run `npm run start` we have an error that goes unreported. 

https://github.com/bafolts/terraforming-mars/blob/main/src/database/SQLite.ts#L49

We ignore the error there and only call the callback if there are rows. If you modify this to not ignore the error you will see the error logged. For now I only made the change for SQLite hoping to minimize how much was changing with one PR.

I don't have to use async. I originally was going a different way with this PR and figured using promises would make this simpler. I can make `initialize` utilize a callback like we do elsewhere.

bafolts:
> Is this getting complicated for the wrong reasons? I don't know enough but my code smell sense is going off.

I don't think so. Another option we could consider is creating the database tables during the build. If the tables are created during the build process then we wouldn't have to create them at runtime. Creating during build time would probably be clearer.

AndrewSav:
> > Is this getting complicated for the wrong reasons? I don't know enough but my code smell sense is going off.
> 
> I don't think so. Another option we could consider is creating the database tables during the build. If the tables are created during the build process then we wouldn't have to create them at runtime. Creating during build time would probably be clearer.

It then will need to be done for all three backends, and it may complicate the build for those who host it independently. If my vote counted I would vote against doing it during the build. (FWIW where I work software self-create tables it uses on the first run, for dozens of apps, makes deployment a lot simple)

src/database/PostgreSQL.ts
AndrewSav: > I originally was going a different way with this PR and figured using promises would make this simpler. 

I agree with this.

kberg:
I agree, I don't love the idea of creating tables during the build. Thanks
for explaining.

>

kberg:
@AndrewSav @bafolts what's the next step here?

AndrewSav:
@kberg I was wondering if @bafolts would like to make any changes here. I had a question about the empty bodies that has not been answered so far. Other than that it's looking fine, however I have not tested it.

bafolts:
I am waiting for someone to approve or leave requested changes. For the empty function bodies I plan to implement those in separate PRs per database implementation.

kberg:
> I am waiting for someone to approve or leave requested changes. For the empty function bodies I plan to implement those in separate PRs per database implementation.

OK, @AndrewSav says there's an open question, so ball's in your court?

bafolts:
The empty function bodies for the initialize functions I plan to implement after this is merged. Trying to start small with one database to limit the scope of this PR.

AndrewSav:
> The empty function bodies for the initialize functions I plan to implement after this is merged. Trying to start small with one database to limit the scope of this PR.

@bafolts  Sorry, I probably was not clear as much as intended: to me it looks like if this is implemented, then between the time this is pushed and the time the Postgres/LocalFS implemented the code will be broken, in particular, it looks to me that it could make currently working Postgres fail, because the timing of the calls will change. I just wanted to double check with you, that you looked at it and is satisfied that this is not an issue.

bafolts:
The new `initialize` function for `PostgreSQL` will resolve when called. The `PostgreSQL` constructor should run as it does today with the same issue where we may not create the tables before they are needed. For good measure I ran this locally pointing to the `terraforming-mars.herokuapp.com` postgres database and I see the service start up and I can load a game from the database. Someone making a new instance of `PostgreSQL` where the tables don't exist will most likely have the same bug as `SQLite` until the `initialize` function is implemented. They would have this bug already today without this change.

AndrewSav:
@bafolts okay fair enough, this fully addresses my concern. It LGTM with the caveat that I have not tested it.

AndrewSav (approved):

bafolts:
This now needs someone with write access to approve.

kberg (approved):

PR #3517, opened 2021-08-17
Clean up SelectPlayer a bit, and add an awesome test.

kberg:


kberg:
Depends on #3516

alrusdi (approved):

bafolts (commented):
I confirm that the test is awesome.

PR #3523, opened 2021-08-18
Clarify PlayerTags & PlayerInfo.

kberg:
Depends on  #3522

First: activePlayer is not a great term: the active player is the one taking the current action, but here it means "me, the player with the brower."

Second: Anyone who is NOT "me, the active player" is of type PublicPlayer, not PlayerView, so using the right types will clear things up.

There's a minor change for a component key to line it up with the others, too.

### DIFF
--- a/src/components/overview/PlayerTags.vue
+++ b/src/components/overview/PlayerTags.vue
@@ -37,7 +37,8 @@
 import Vue from 'vue';
 import TagCount from '../TagCount.vue';
 import {ITagCount} from '../../ITagCount';
-import {PlayerViewModel} from '../../models/PlayerModel';
+import {PlayerViewModel, PublicPlayerModel} from '../../models/PlayerModel';
+import {GameModel} from '../../models/GameModel';
 import {Tags} from '../../cards/Tags';
 import {CardName} from '../../CardName';
 import {SpecialTags} from '../../cards/SpecialTags';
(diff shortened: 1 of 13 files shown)


src/components/overview/PlayerTags.vue
```diff
@@ -134,23 +135,26 @@ export default Vue.extend({
...
+      return this.playerView.game.gameOptions.moonExpansion;
     },
     getTagsPlaceholders: function(): Array<InterfaceTagsType> {
       const tags = PLAYER_INTERFACE_TAGS_ORDER;
       return tags.filter((tag) => {
-        return checkTagUsed(tag, this.player);
+        return isTagInGame(tag, this.playerView.game);
       });
     },
+    isThisPlayer: function(): boolean {
```
Argeento, line 155: [use computed instead](https://vuejs.org/v2/guide/computed.html#Computed-Caching-vs-Methods)

kberg: As a general rule, yes, good catch. Since it's only used in one place, it doesn't change performance barely at all. That said, I'll change that in a follow-up.

Argeento: The whole project is using methods instead of computed values. 
You're right, this one is irrelevant to performance.
Yea... with computed value there is a contradiction between the consistency and the correctness of code.
Imo correctness is more important, but no problem, I am glad you are aware now

kberg: Yeah, thank you for raising it though! I learned about it while trying to do a big separate refactor (that failed in so many ways) but I liked what it provided. More of those, please.

kberg: Ah screw it, I'll change it later today since you raised it. ;D

kberg: Done.

bafolts (approved):
This re-factor brought to you by. "Single File Components" and vetur.

Issue #3520, opened 2021-08-17
Iapetus colony incorrect status bar cost reduction when building colony after trade


PitrPi:
Happens only when trading with Iapetus AND THEN building colony there

Steps to recreate bug:
1) Trade with Iapetus - 🕑You spent 9 M€ to trade with Iapetus
2) Build a colony on Iapetus - 🕑You built a colony on Iapetus

Result:
- no cost reduction on cards played (CORRECT)
- toolbar shows cost reduction (INCORRECT)
![image](https://user-images.githubusercontent.com/11059560/129743269-349727df-38a8-4b46-a832-c2c1e9e9ccc5.png)

kberg:
Thanks for the report, and the superb steps to recreate! Things have slowed
down here a bit, so it might get resolved today, might get resolved in a
couple of weeks.

>

bafolts:
@PitrPi if you build a colony on Iapetus don't you gain a card discount? Following the steps to reproduce I don't see why a player wouldn't gain a card discount in this situation.

PitrPi:
For the bug to happen trade must occur BEFORE building colony.
If you have a colony on Iapetus when someone trades with it you get a discount and proper -1 in player bar - all good here
If someone trades with Iapetus and THEN you build colony on it, you don't get a discount, but -1 in player bar still shows up.

It is not a bug that influences game (card cost when playing is correct in both cases) it only affects frontend (displayed discount in the player bar).

Hope it is clearer.

PitrPi:
I am not TypeScript fluent, but it seems that source of bug is here:
https://github.com/bafolts/terraforming-mars/blob/d71e4bbb01c9df675b85a6f8ca3050dd634060e2/src/components/overview/PlayerTags.vue#L195
It only checks if there is Player's colony on Iapetues and if there is visitor on it. It does not check if there was a colony BEFORE visitor.

bafolts:
We also always seem to give a colony bonus unless copying a trade.

https://github.com/bafolts/terraforming-mars/blob/main/src/colonies/Colony.ts#L125

PitrPi:
Colony bonus for player is calculated at the time of "Trade with colony" action, therefore card discount is awarded iff you have colony at that point in time. - That is correct
Player status bar before Iapetus was independent on order of player actions, (either you have a discount or not, you have a tag or not), so it seems to be calculating from current state of the game. However with Iapetus you need to know if trade occurs before or after colony build.

[...]

PR #3525, opened 2021-08-20
use player cardDiscount from server for client

bafolts:
This is for #3520 . The `player.cardDiscount` value is now used on the client to reflect the player's card discount. This avoids having to calculate the value client side.


kberg (approved):
This is a good change.

PR #3526, opened 2021-08-21
My most recent attempt to move public fields out of the private view.

kberg:
FEEL FREE TO READ, BUT THIS IS NOT FOR SUBMISSION

This one worked all the way through manual testing, so this is probably the one I'm going model against, but once more, as a series of incremental changes.

After this Spectator Mode will come pretty easily and my original trigger for this, a feature that prevents the game creator from peeking  at other players' starting hands before choosing a game.


kberg:
This is now reasonably out of date and served its purpose.

PR #3527, opened 2021-08-21
Three minor structural changes to the PlayerModels.

kberg:
Each commit can be reviewed separately (including the one that moves fields around.)


src/models/ServerModel.ts
```diff
@@ -91,29 +91,39 @@ export class Server {
   public static getPlayerModel(player: Player): PlayerViewModel {
     const game = player.game;
 
+    const players: Array<PublicPlayerModel> = this.getPlayers(game.getPlayers(), game);
+    const thisPlayerIndex: number = players.findIndex((p) => p.color === player.color);
```
bafolts, line 95: I _think_ this can use `const thisPlayer = players.find` and avoid having to create `thisPlayerIndex`. It doesn't seem like `thisPlayerIndex` is used anywhere else.

bafolts: Instead of calling to `getPlayers` and then finding the one public player we need I think we should create a `getPlayer` that can return the public model for one player. Then the `getPlayers` function would call to `getPlayer`. This will avoid having to map every other player.

kberg: `thisPlayerIndex` isn't used yet, but I plan to use it as an optimization so that the same player is not sent down the wire twice. I could remove it now and add a TODO. Your thoughts?

> Instead of calling to getPlayers and then finding the one public player we need I think we should create a getPlayer that can return the public model for one player. Then the getPlayers function would call to getPlayer. This will avoid having to map every other player.

It doesn't matter, actually. I've read through this code and compiled and changed it a few times over the last few months, and we'll be fine without it.

bafolts (approved):

Issue #3530, opened 2021-08-22
Stormcraft + Kelvinist special action does not allow paying heat with the floater


kberg:
>  Corp Stormcraft vs special Kelvinist action 6 heat res into 1 temp: it is not possible to pay the heat with the floater. Possible bug: the "howToPay" function is not been coded with the Kelvinist special 6heat in 1 temp action.

![image](https://user-images.githubusercontent.com/413481/130361171-7cf29533-2475-47b3-bac2-84341870cb7a.png)

![image](https://user-images.githubusercontent.com/413481/130361176-30d35720-5134-4844-b859-c217b733e418.png)

the game id from the (artificial) game was gb7057d4ee37

![image](https://user-images.githubusercontent.com/413481/130361182-eda9a95b-8eab-4a76-bd14-a800bda0d60e.png)

PR #3532, opened 2021-08-23
Fix: Kelvinists P3 should be reusable with Stormcraft

nwai90:
Closes https://github.com/bafolts/terraforming-mars/issues/3530

### DIFF
--- a/src/turmoil/Policy.ts
+++ b/src/turmoil/Policy.ts
@@ -3,6 +3,7 @@ import {Game} from '../Game';
 import {OrOptions} from '../inputs/OrOptions';
 import {ISpace} from '../boards/ISpace';
 import {Player} from '../Player';
+import {PlayerInput} from '../PlayerInput';
 
 export type PolicyId = string;
 
@@ -12,7 +13,7 @@ export interface Policy {
   isDefault: boolean;
   onTilePlaced?: (player: Player, space: ISpace) => void;
   onCardPlayed?: (player: Player, card: IProjectCard) => void;
-  action?: (player: Player) => OrOptions | undefined;
+  action?: (player: Player) => PlayerInput | OrOptions | undefined;
   canAct?: (player: Player) => boolean;
   apply?: (game: Game) => void;
 }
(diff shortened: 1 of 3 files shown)


src/turmoil/Policy.ts
```diff
@@ -12,7 +13,7 @@ export interface Policy {
   isDefault: boolean;
   onTilePlaced?: (player: Player, space: ISpace) => void;
   onCardPlayed?: (player: Player, card: IProjectCard) => void;
-  action?: (player: Player) => OrOptions | undefined;
+  action?: (player: Player) => PlayerInput | OrOptions | undefined;
```
kberg, line 16: Do you still need OrOptions?

kberg (approved):
Wow, I thought this was going to be a lot of hacky local changes. This is great!

src/turmoil/Policy.ts
nwai90: Yep, we still need that for several other Policy actions

kberg: Sure, that's fine, but since OrOptions is a subclass of PlayerInput, it only serves as documentation.

src/turmoil/parties/Kelvinists.ts
```diff
@@ -89,17 +90,18 @@ class KelvinistsPolicy03 implements Policy {
...
     const game = player.game;
     game.log('${0} used Turmoil Kelvinists action', (b) => b.player(player));
     game.log('${0} spent 6 heat to raise temperature 1 step', (b) => b.player(player));
 
-    player.deductResource(Resources.HEAT, 6);
-    game.increaseTemperature(player, 1);
-    return undefined;
+    return player.spendHeat(6, () => {
+      game.increaseTemperature(player, 1);
+      return undefined;
```
Argeento, line 103: Functions in JS by default return `undefined`. No need to return it explicitly

kberg: Not with our typescript configuration.

PR #3533, opened 2021-08-23
Three minor model changes

kberg:
The commits can be broken up as #585919b, which uses PublicPlayerModel, #fb5545f, which also uses PublicPlayerModel, and #3ce7128, which is the change to PlayerHome. It introduces two computed properties, but only uses one of them; that one is easier for you, the reviewer, to consume and understand. The next one will use `thisPlayer`.

### DIFF
--- a/src/components/PlayerHome.vue
+++ b/src/components/PlayerHome.vue
@@ -1,8 +1,8 @@
 <template>
-        <div id="player-home" :class="(player.game.turmoil ? 'with-turmoil': '')">
+        <div id="player-home" :class="(game.turmoil ? 'with-turmoil': '')">
             <top-bar :player="player" />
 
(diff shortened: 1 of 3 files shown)


src/components/PlayerHome.vue
```diff
@@ -289,6 +290,14 @@ export default Vue.extend({
       type: Object as () => typeof raw_settings,
     },
   },
+  computed: {
+    thisPlayer: function(): PublicPlayerModel {
+      return this.player.thisPlayer;
+    },
+    game: function(): GameModel {
```
bafolts, line 297: This is great change. If nothing else it leads to less code we have to write.

bafolts (approved):

PR #3536, opened 2021-08-24
Last minor thisPlayer changes before payment

kberg:
Payment is a big change, and the next ones after are also pretty big. I'm excited!

### DIFF
--- a/src/components/GameEnd.vue
+++ b/src/components/GameEnd.vue
@@ -15,7 +15,7 @@
                             <ul class="game_end_list">
                                 <li v-i18n>Try to win with expansions enabled</li>
                                 <li v-i18n>Try to win before the last generation comes</li>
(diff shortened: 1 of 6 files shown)


src/components/GameEnd.vue
```diff
@@ -15,7 +15,7 @@
                             <ul class="game_end_list">
                                 <li v-i18n>Try to win with expansions enabled</li>
                                 <li v-i18n>Try to win before the last generation comes</li>
-                                <li><span v-i18n>Can you get</span> {{ player.victoryPointsBreakdown.total + 10 }}<span v-i18n>+ Victory Points?</span></li>
+                                <li><span v-i18n>Can you get</span> {{ player.thisPlayer.victoryPointsBreakdown.total + 10 }}<span v-i18n>+ Victory Points?</span></li>
```
bafolts, line 18: Can most likely create the `computed` property for `thisPlayer` in this component as well.

bafolts (approved):

PR #3542, opened 2021-08-27
An attempt at a canSave UI.

kberg:
Not sure this is necessary as much as it's probably just worth fixing selectColony. /shrug


kberg:
Oh, I don't think this works, really.

>

PR #3543, opened 2021-08-27
Remove prettier config

Argeento:
Current config fires formatting in IDE with prettier extensions.


kberg:
After speaking some more with @Argeento, removing this makes sense. It can always be restored.

kberg (approved):

PR #3545, opened 2021-08-27
Hide tiles on the moon

Argeento:
[https://github.com/bafolts/terraforming-mars/issues/3399](https://github.com/bafolts/terraforming-mars/issues/3399)


tests/components/overview/PlayerStatus.spec.ts
```diff
@@ -1,34 +1,34 @@
-import {createLocalVue, mount} from '@vue/test-utils';
```
Argeento, line 1: These tests are commented because they don't pass

kberg: Thanks for catching the broken test and the bad format for the test parameters. Let's fix that first, separately. I'll pick on up that if you don't want to.

Argeento: Yes please. The code that this test was covering is now probably in `PlayerInfo` component.

kberg: Now in #3548.

src/components/Board.vue
```diff
@@ -1,7 +1,9 @@
 <template>
     <div :class="getGameBoardClassName()">
         <div class="hide-tile-button-container">
-        <div class="hide-tile-button" v-on:click.prevent="toggleHideTile()">{{ toggleHideTileLabel() }}</div>
+          <div class="hide-tile-button" @click="$emit('toggleTileHidden')" data-test="hide-tiles-button">
```
kberg, line 4: I see -- data-test is the equivalent of an ID that doesn't have the semantics of a DOM id but can be used in tests? I like it. The phrase "data-test" isn't available online, so I wonder, is that just a local style you learned or a standard used elsewhere?

src/components/Board.vue
```diff
@@ -1,7 +1,9 @@
...
+            {{ isTileHidden ? 'show' : 'hide' }} tiles
```
kberg, line 5: note to self: Ths isn't internationalized.

src/components/Board.vue
```diff
@@ -56,7 +58,16 @@
...
         <div class="board" id="main_board">
-            <board-space :space="curSpace" :is_selectable="true" :key="'board-space-'+curSpace.id" :aresExtension="aresExtension" :isTileHidden="checkHideTile()" v-for="curSpace in getAllSpacesOnMars()"></board-space>
+            <board-space
+              v-for="curSpace in getAllSpacesOnMars()"
+              :key="curSpace.id"
+              :space="curSpace"
+              :is_selectable="true"
+              :aresExtension="aresExtension"
+              :isTileHidden="isTileHidden"
+              data-test="board-space"
```
kberg, line 68: Now I'm not sure what board-space represents, it's not an ID, it is however a semantic-free marker. So, that's helpful, too.

src/components/Board.vue
kberg: Minor: `toggleTileHidden` -> `toggleHideTiles`

src/components/PlayerHome.vue
```diff
@@ -256,6 +258,7 @@ export interface PlayerHomeModel {
   showActiveCards: boolean;
   showAutomatedCards: boolean;
   showEventCards: boolean;
+  isTileHidden: boolean;
```
kberg, line 261: 'is' is a standard for booleans, but here, I'd prefer `hideTiles`. Actually, I think everything should be expressed in their positive form (showTiles) regardless of their initial state, but that's not necessary here.

kberg (commented):
This is really nice! I haven't finished reading it, but I like what I see, and I am learning alot. Thank you!

src/components/moon/MoonSpace.vue
```diff
@@ -35,31 +40,45 @@ export default Vue.extend({
...
   },
-  methods: {
-    getVerboseTitle(tileType: TileType | undefined): string {
-      let ret: string = '';
-      if (tileType === TileType.LUNA_TRADE_STATION) {
+  computed: {
+    verboseTitle(): string {
+      let ret = '';
+      if (this.space.tileType === TileType.LUNA_TRADE_STATION) {
         ret = 'Luna Trade Station';
```
kberg, line 55: Another reminder to internationlize these. Can you add a TODO for this and the spot above. please?

src/components/moon/MoonSpace.vue
```diff
@@ -1,9 +1,14 @@
 <template>
-        <div :class="getMainClass()" :data_space_id="space.id" :title="getVerboseTitle(space.tileType)">
-            <div class="board-space-text" v-if="text" v-i18n>{{ text }}</div>
-            <bonus :bonus="space.bonus" v-if="space.tileType === undefined"></bonus>
-            <div :class="'board-cube board-cube--'+space.color" v-if="space.color !== undefined"></div>
-        </div>
+  <div :class="mainClass" :data_space_id="space.id" :title="verboseTitle">
```
kberg, line 2: This is nice.

kberg (commented):
This change is fantastic. If you can make the naming changes, it can go right in. Thank you!

src/components/Board.vue
Argeento: It's often in use but with a different names - check `data-test-id` or `data-testid`. Some tools use their own conventions, like [`data-cy` in cypress](https://docs.cypress.io/guides/references/best-practices#Selecting-Elements). But the idea is the same.

src/components/Board.vue
Argeento: image
In test I had to somehow distinguish tiles on the moon from the around ones.

src/components/PlayerHome.vue
Argeento: imo expression `doSomething` (`showTiles` in this case) is reserved for methods names

kberg: Understood. Verbs are used as properties (parameters) as well as methods. Thanks!

src/components/moon/MoonSpace.vue
Argeento: Done

src/components/Board.vue
Argeento: Todo added

src/components/Board.vue
kberg: Sorry, by board-space I meant data-test. But still, it's fine. :D

src/components/PlayerHome.vue
kberg: Sorry perhaps I wasn't clear: I'm saying that hideTiles is an appropriate parameter name, and asking for the change.

kberg (commented):
Thanks, just this last bit, then.

kberg (approved):
Thank you! borbarad will be thrilled.

PR #3548, opened 2021-08-28
Restore recursive component tests.

kberg:
Also, Fix PlayerStatus test which really tested something in PlayerInfo.


tests/components/overview/PlayerInfo.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+          getVisibilityState: function() {},
+        },
+      },
+      propsData: {
+        player: thisPlayer,
+        playerView: playerView,
+        playerIndex: 0,
+      },
+    });
+    const test = playerStatus.find('div[class*="played-cards-count"]');
```
Argeento, line 58: Please use `data-test` to find DOM elements. - data attributes are resilient to UI changes.

bafolts (approved):

tests/components/overview/PlayerInfo.spec.ts
kberg: I'll do that as a follow-up since the goal is to restore the test and unblock the other change. Thanks. Again, this will be a good follow-up.

PR #3549, opened 2021-08-29
Async translations loading

Argeento:
Reduce bundle size by 45% (from 387KB to 213KB)

### DIFF
--- a/make_static_json.js
+++ b/make_static_json.js
@@ -77,3 +77,32 @@ fs.writeFileSync('src/genfiles/settings.json', JSON.stringify({
 fs.writeFileSync('src/genfiles/translations.json', JSON.stringify(
   getAllTranslations(),
 ));
+
+/**
+ * Generate translation files in `/assets/locales/*.json` to load them async by the client
+ */
+function generateTranslations() {
+  const localesDir = path.join(__dirname, 'src/locales');
+  const localesCodes = fs.readdirSync(localesDir);
+  const destinationPath = path.join(__dirname, 'assets/locales');
+
+  if (!fs.existsSync(destinationPath)) {
+    fs.mkdirSync(destinationPath);
+  }
+
+  localesCodes.forEach((localeCode) => {
+    const localeDir = path.join(localesDir, localeCode);
+    const localeFiles = fs.readdirSync(localeDir);
+
+    const localeObject = localeFiles.reduce((localeObject, localeFile) => {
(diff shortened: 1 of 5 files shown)


kberg:
This is nice. Paging @alrusdi

alrusdi:
I've added some comments. Brian should help with locales generation part.

make_static_json.js
```diff
@@ -77,3 +77,32 @@ fs.writeFileSync('src/genfiles/settings.json', JSON.stringify({
 fs.writeFileSync('src/genfiles/translations.json', JSON.stringify(
```
bafolts, line 77: This change seems to deprecate `src/genfiles/translations.json`. Can you update the `check_locales.ts` tool to use the `src/locales`?

make_static_json.js
```diff
@@ -77,3 +77,32 @@ fs.writeFileSync('src/genfiles/settings.json', JSON.stringify({
...
+  if (!fs.existsSync(destinationPath)) {
+    fs.mkdirSync(destinationPath);
+  }
+
+  localesCodes.forEach((localeCode) => {
+    const localeDir = path.join(localesDir, localeCode);
+    const localeFiles = fs.readdirSync(localeDir);
+
+    const localeObject = localeFiles.reduce((localeObject, localeFile) => {
+      const filePath = path.join(localeDir, localeFile);
```
bafolts, line 98: This should check that the file extension ends in `.json`.

make_static_json.js
Argeento: I left this function to not destroy `check_locales` - Ok, I will update this tools

make_static_json.js
Argeento: What else could it be if the folder structure is clear? Whatever.. I will add this check

make_static_json.js
Argeento: @bafolts the question is - what should be a source of truth? How can I tell, which keys are missing and which one are additional and unnecessary?

Should I create a union of keys from every translation, and then listing the missing keys in individual translations?

or

Should I create a detached key list and check every translation against it?

Argeento: In the second solution we can take advantage of TS types checking - every translation file will be a partial type of object of keys. This will prevent any typo.

Argeento: There is a third solution - left this tool and translation.json generation - everything works the same as before

bafolts (approved):

PR #3550, opened 2021-08-29
Refactor button component

Argeento:
I don't know why $root mock doesn't work in tests (tbh I have never seen $root in real vue before)

Btw guys: RTFM - [Accessing the Root Instance](https://vuejs.org/v2/guide/components-edge-cases.html#Accessing-the-Root-Instance)

> This can be convenient for demos or very small apps with a handful of components. However, the pattern does not scale well to medium or large-scale applications, so we strongly recommend using Vuex to manage state in most cases.

I will fix these tests later - with `app.ts` refactor

### DIFF
--- a/src/components/AndOptions.vue
+++ b/src/components/AndOptions.vue
@@ -10,7 +10,7 @@
       :showsave="false"
       :showtitle="true" />
     <div v-if="showsave" class="wf-action">
-      <Button :title="playerinput.buttonLabel" type="submit" size="normal" :onClick="saveData" :disabled="!canSave()"/>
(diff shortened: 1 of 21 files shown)


kberg:
> Btw guys: RTFM - [Accessing the Root Instance](https://vuejs.org/v2/guide/components-edge-cases.html#Accessing-the-Root-Instance)
> 
> > This can be convenient for demos or very small apps with a handful of components. However, the pattern does not scale well to medium or large-scale applications, so we strongly recommend using Vuex to manage state in most cases.

Maybe this is a language difference, but hey, there are nicer ways to say that.

kberg (approved):
This is a good change. Thank you for the test!

PR #3551, opened 2021-08-29
fix player timer

### LINKED ISSUE #3547: Timer isn't working properly
Hi,

I updated the project toady and started a new game and now the timer seems to be broken. First it shows "00:00" for everyone, now it shows the same time for everyone. It doesn't run, it seems to just set the time.

https://abload.de/img/bug0zjet.png

Cheers!

bafolts:
This should fix #3547 .

I simplified the player status component a lot and only passed in what was needed. Test this with a quick multi player and single player game where I saw the timer tick. Also tested with show timers off and did not see the timer.

### DIFF
--- a/src/components/overview/PlayerStatus.vue
+++ b/src/components/overview/PlayerStatus.vue
@@ -14,13 +14,13 @@
 import Vue from 'vue';
 import {ActionLabel} from './ActionLabel';
 import PlayerTimer from './PlayerTimer.vue';
-import {SerializedTimer} from '../../SerializedTimer';
+import {TimerModel} from '../../models/TimerModel';
 
 export default Vue.extend({
   name: 'player-status',
   props: {
     timer: {
-      type: Object as () => SerializedTimer,
+      type: Object as () => TimerModel,
     },
     actionLabel: {
       type: String,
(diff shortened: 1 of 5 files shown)


src/components/overview/PlayerStatus.vue
```diff
@@ -13,28 +13,20 @@
...
 import Vue from 'vue';
 import {ActionLabel} from './ActionLabel';
-import {mainAppSettings} from '../App';
-import {PlayerViewModel} from '../../models/PlayerModel';
 import PlayerTimer from './PlayerTimer.vue';
-
-export const hidePlayerData = (root: typeof mainAppSettings.methods, playerIndex: number) => {
-  root.setVisibilityState('pinned_player_' + playerIndex, false);
-};
+import {SerializedTimer} from '../../SerializedTimer';
```
kberg, line 17: Shouldn't there be a TimerModel, and not a SerializedTimer?

kberg (approved):

kberg:
This is a nice simplification.

src/components/overview/PlayerStatus.vue
bafolts: Yes. While we are here I will add it with this. Even if it has similar properties of `SerializedTimer` the `Serialized` is used for storing to database and not for client side.

PR #3558, opened 2021-09-01
Refactor award component

Argeento:


src/components/Award.vue
```diff
@@ -1,94 +1,57 @@
...
-                    <span v-i18n>{{ award.award.name }}</span>
-                    <span class="ma-player-cube"><i :class="'board-cube board-cube--'+award.player_color" /></span>
-                </span>
-                <span v-if="isLearnerModeOn()">
-                  <span v-for="spotPrice in getAvailableAwardSpots()" :key="spotPrice" class="milestone-award-inline unpaid">
-                    <div class="milestone-award-price">{{spotPrice}}</div>
-                  </span>
-                </span>
-            </div>
+  <div @click="toggleDescription" class="ma-block" title="Press to show or hide the description">
```
kberg, line 2: I don't know if removing the capital P breaks internationalization, but can you make sure all the text you've included retains the original case, or, map accordingly?

Note that there's two uses of that phrase, and one translation:

image

(This is a bit why smaller changes are better. I only happened to see this. Are there other text changes?)

src/components/Award.vue
```diff
@@ -1,94 +1,57 @@
...
+    <div class="ma-player" v-if="award.player_name">
+      <i :title="award.player_name" class="board-cube" :class="`board-cube--${award.player_color}`" />
```
kberg, line 4: Rewriting the components to use `` instead of string concatenation is nice, but it means I have extra review work to do. Please try to constrain these ongoing. I know you like reformatting, and I'm not trying to argue that the existing format is easier to read, but it makes changes harder to review.

src/components/Award.vue
```diff
@@ -1,94 +1,57 @@
...
+    </div>
+
+    <div class="ma-name ma-name--awards award-block" :class="maAwardClass" v-i18n>
+      {{ award.award.name }}
+      <div class="ma-scores player_home_block--milestones-and-awards-scores">
```
kberg, line 9: What happened to show_scores?

src/components/Award.vue
```diff
@@ -1,94 +1,57 @@
...
-      });
-      return AWARD_COSTS.slice(numFundedAwards);
+  },
+  computed: {
+    maAwardClass(): string {
+      return 'ma-name--' + this.award.award.name.replace(/ /g, '-').toLowerCase();
     },
-    isLearnerModeOn(): boolean {
-      return PreferencesManager.loadBoolean('learner_mode');
+    sortedScores(): FundedAwardModel['scores'] {
```
kberg, line 53: I can't parse what type `FundedAwardModel['scores']` is. Can you help, please?

src/components/Award.vue
```diff
@@ -1,94 +1,57 @@
...
+      return [...this.award.scores].sort((s1, s2) => s2.playerScore - s1.playerScore);
```
kberg, line 54: I'm pretty sure you need the array and spread operators.

src/components/Award.vue
```diff
@@ -1,94 +1,57 @@
...
+
+    <div class="ma-name ma-name--awards award-block" :class="maAwardClass" v-i18n>
+      {{ award.award.name }}
+      <div class="ma-scores player_home_block--milestones-and-awards-scores">
+        <p
+          v-for="score in sortedScores"
+          :key="score.playerColor"
+          class="ma-score"
+          :class="`player_bg_color_${score.playerColor}`"
+          v-text="score.playerScore"
```
kberg, line 15: Why have you chosen this instead of what we do elsewhere, to embed this as text inside the <p>?

kberg (changes requested):
I really like the restructuring, but the reformatting and other changes makes it hard to review these. Breaking changes like this into several small commits or a series of small PRs would go a long way toward faster reviews. Please consider that ongoing.

src/components/Award.vue
kberg: > I'm pretty sure you need the array and spread operators.
I mean "don't need"

src/components/Award.vue
Argeento: Deleted - templet didn't use this variable

src/components/Award.vue
Argeento: `sort` is stupid method that breaks basic principle of FP - do not mutate.
Without additional array wrap - sort method is sorting original array.
Props should not be modified in a child component.
So yes. This array and spread operators are needed.

src/components/Award.vue
Argeento: Shorter code / better looking code in column of attributes like this

src/components/Award.vue
Argeento: You can't? or your IDE can't? Its an array of PlayerScore. I can import this interface explicitly if you wan.

src/components/Award.vue
Argeento: I restored original letter. I will not change any text again - sorry.

src/components/Award.vue
Argeento: No way :P I will always replace old concatenations to templates string while refactoring. If you want I can crate PR that changes all concatenations in all components at once ([similar to this one](https://github.com/bafolts/terraforming-mars/pull/3550/files))

src/components/Award.vue
kberg: Don't get me wrong, it's fine to do in context. I've made that mistake with text plenty of times. :D

src/components/Award.vue
kberg: I hear you. The time I would spend reviewing that change does not make it worthwhile, sorry.

src/components/Award.vue
kberg: I see it referenced externally here

image

and used inside the template here

image

So what am I missing?

kberg: Huh, yeah, I wonder if something changed beforehand that changed the default values, or something. I'm not sure that's right. It is this kind of multi-purpose change that I find particularly tricky. When submitting a PR, making clear that you're removing features like this in advance would make things easier so I don't have to map intently at both sets of code.

src/components/Award.vue
kberg: I cannot tell. I haven't seen a value used as a type before.

src/components/Award.vue
kberg: I understand. But nothing was broken. Keep it for now, though.

src/components/Award.vue
kberg: I think it's OK, I like the simplicity. If v-text was a string, would it respond correctly to the i18n attributes we use to internationalize? If not, it just means that sometimes we do one thing and sometimes another. I assume it will support i18n, though.

src/components/Award.vue
Argeento: It's not a value used as a type. Look at [indexed access type](https://www.typescriptlang.org/docs/handbook/2/indexed-access-types.html)

src/components/Award.vue
Argeento: Text changes are irrelevant to these code refactors. I should avoid it. Moreover I've checked i18n directive and you are right. Translations are case sensitive.

The `p` letter was restored

src/components/Award.vue
kberg: Rereading my last comment I want to make sure I was clear: don't change it back. :D

src/components/Award.vue
kberg: #til that's neat.

src/components/Award.vue
kberg: OK, so it wasn't clear: I was OK with the indexed type, I thought it was a reasonable use. But thank you.

kberg (approved):

kberg:
I think there were one or two more comments, but let's get this in - it's a massive improvement. Thank you!

PR #3563, opened 2021-09-02
Make package.json Windows-compatible.

kberg:


kberg:
Huh! What's this now?

Argeento:
These files:
`@/tests/components/FakeLocalStorage.ts`
`@/tests/components/PaymentTester.ts`
`@/tests/components/setup.ts`
are now run by mocha as a normal tests.

The previous config allows only `*.spec.ts` files

kberg:
> These files:
> `@/tests/components/FakeLocalStorage.ts`
> `@/tests/components/PaymentTester.ts`
> `@/tests/components/setup.ts`
> are now run by mocha as a normal tests.
> 
> The previous config allows only `*.spec.ts` files

Beforehand, `npm run test:components` ran zero component tests under Windows. Between that and the effect you describe, this seems preferable.

Argeento:
I have windows machine (windows 10 pro) with git bash shell and everything works correct.
Maybe try [different quotes around glob](https://github.com/sysgears/mochapack/blob/master/docs/installation/setup.md)
If this doesn't work, try move mochapack config to separate file, and inside this file define glob pattern if it's possible

kberg:
Then what am I doing wrong? Huh. OK. I'll abandon this for now, and extract
it from the PRs I sent out today.

bafolts:
One of the github actions runs on a windows machine. Windows has a few options for running with powershell, bash, cygwin, or cmd.exe. I wouldn't be surprised if one of these didn't work.

kberg:
Oh right - which does it use?

kberg:
FWIW I'm using Mingw32, which is part of _Git for Windows_

Argeento:
oh.. https://github.com/bafolts/terraforming-mars/pull/3558/checks?check_run_id=3507074607#step:6:4768

kberg:
Huh! OK, so keep this?

>

Argeento:
Have you tried with `"` instead of `'`?

kberg:
I haven't looked at the test output in GREAT detail, but both Windows and Linux for this run ran 93 component tests: https://github.com/bafolts/terraforming-mars/pull/3563/checks?check_run_id=3508320577

Argeento (approved):

kberg:
Thanks. Pending approval from writer, but I'm not blocked, I just use `npx mochapack...` for now.

bafolts (approved):

PR #3564, opened 2021-09-02
s/player/playerView for everything but player input stuff.

kberg:

PR #3565, opened 2021-09-02
Have input selection use playerView

kberg:
Depends on #3563 and #3564

Rather than rename all the fields in all the components at once, this commit uses computed to ease the transition. Then I can do the change for each one individually. But these have to be in a bunch because there's a cross-reference between them thanks to the relationship between PlayerInputFactory, AndOptions and OrOptions.

### DIFF
--- a/assets/index.html
+++ b/assets/index.html
@@ -34,14 +34,14 @@
                 ></game-home>
                 <player-home
                     v-else-if="screen === 'player-home'"
-                    :player="player"
+                    :player-view="playerView"
                     :key="playerkey"
                     :settings="settings"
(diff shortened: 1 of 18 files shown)


nwai90 (approved):
Approved, but please verify that a 2-player game with some cards that invoke `AndOptions` and `OrOptions` can run to the end and render correctly

kberg:
Will do.

kberg:
Dang, wow, nope. New issues. Manual testing now.

kberg:
For the record, I did run this through a game, but I suspect recent changes got in the way. Let's see.

PR #3566, opened 2021-09-03
Provide Spectator view

kberg:
Log auto refresh still doesn't work, though.


kberg:
While reviews are backing up, I added the rest of the features. Automatic refresh is still broken, but that's enough to call it Preview Mode.

bafolts (approved):

PR #3568, opened 2021-09-03
Improve DX

Argeento:
It's a little workaround because backend doesn't have separation of prod/dev mode.

# What's changed
1. If there is no `main.js.br` - the server sends original js file. This uncompressed file is necessary in front dev mode
2. The `npm run watch` command now deletes `.js*.br` and `.js*.gz` files before it runs
3. Webpack compression plugin is only available in production mode
4. `NODE_ENV` is now explicitly set for each of build/dev command

# Why?
1. Users will no longer build their instances in webpack development mode. (to build in production mode you must have NODE_ENV set to the `production` and nobody except heroku has)
2. Recompiling components in `npm run watch` now takes 20x faster
3. Rerun test in `npm run test:componets:watch` now is instant

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "@typescript-eslint/eslint-plugin": "^4.8.2",
     "@typescript-eslint/parser": "^4.8.2",
     "console-stamp": "3.0.0-rc4.2",
+    "cross-env": "^7.0.3",
     "dialog-polyfill": "^0.5.1",
     "dotenv": "^8.2.0",
     "eslint": "^6.6.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -72,10 +74,11 @@
     "test": "npm run test:server && npm run test:components",
     "test:server": "mocha --require build/tests/utils/setup.js --recursive build/tests --exclude 'build/tests/components/**'",
     "pretest:server": "tsc --build tests/tsconfig.json",
-    "test:components": "vti diagnostics && mochapack --require tests/components/setup.ts 'tests/components/**/*.spec.ts'",
-    "test:components:watch": "mochapack --watch --require tests/components/setup.ts 'tests/components/**/*.spec.ts'",
-    "watch": "webpack --watch --mode=development",
-    "webpack": "webpack"
+    "test:components": "cross-env NODE_ENV=development vti diagnostics && mochapack --require tests/components/setup.ts 'tests/components/**/*.spec.ts'",
+    "test:components:watch": "cross-env NODE_ENV=development mochapack --watch --require tests/components/setup.ts 'tests/components/**/*.spec.ts'",
+    "prewatch": "rimraf build/main.js*.{br,gz}",
```
kberg, line 79: I think we can do jsut fine with rm like we do in `clean`, I've been running clean for some time using git bash with no issue.  But that's minor.

src/routes/ServeAsset.ts
```diff
@@ -115,6 +115,13 @@ export class ServeAsset extends Handler {
         encoding = 'gzip';
         file += '.gz';
       }
+
+      // Return not-compressed .js files for development mode
+      if (!fs.existsSync(file)) {
```
kberg, line 120: This is good.

kberg (commented):
I'm gonna pull this in and test it to prevent another rollback, but this looks good.

kberg:
Sanity test on Windows looks good.

kberg (approved):

PR #3572, opened 2021-09-04
Use path alias to import files from src dir

Argeento:
It's the first step to fully separate client code from API.

The dir structure should be rearranged to nicely introduce:
 - Vuex (to get rid of `this.$root` / `import {mainAppSettings} from App.ts`
 - Vue-router (to separate front-end routing from the server)

The clinet code will be look like this:
```
| -📁 src/
   | -📁 client/
      | - 📁 components/
      | - 📁 directives/
      | - 📁 locales/
      | - 📁 routes/
      | - 📁 store/
      | - 📁 utils/
      | - main.ts
      | - webpack.config.js
      | - tsconfig.json
```

If you don't want to change imports in all files in the project I can provide PR with imports changed only in `src/components` and `tests/components` dirs.

You don't need to check the changes of imports manually, if:
- `npm run build`
- `npm run test`
- `npm run cover`
- `npm run test:components:watch`
- `npm run start`
- `npm run watch`

are working, then every import is correct


kberg:
I gave it a good try, but github stopped providing diffs after some number of files (isn't it 250? I seem to recall that.)

image

I just can't do a review where I have to click 1,200 times to see the changes. Can you please break this into groups of 200?

Argeento:
Let's start with client files. It's enough for me

package.json
```diff
@@ -62,17 +66,17 @@
     "build": "npm run compile && npm run webpack && npm run lint",
     "prebuild": "npm run makeCss",
     "clean": "rm -r ./build/ ./src/genfiles/ || true",
-    "compile": "tsc --build src/tsconfig.json",
+    "compile": "tsc --build src/tsconfig.json && tsc-alias -p src/tsconfig.json",
```
kberg, line 69: What is this addition do?

Argeento: tsc by default doesn't resolve imports. There are to solutions:
1. Resolve them after build (like above)
2. Resolve them at runtime by ts-node/register and tsconfig-paths/register

number 2 takes too long to start the server

kberg: That makes sense.

kberg (commented):
I approve merging this change, but want to let @bafolts have the last look.

package.json
bafolts: Why does this need to change from using relative imports? The introduction of `tsc-alias` takes on a dependency I would prefer not to introduce. This requires someone familiar with typescript to learn how `tsc-alias` works.

kberg:
Good question. On the surface it is just a clean-up,but it helps with
refactoring - imports don't need to change when files move, and in order to
get a better router, and separation of concerns (client-server) a lot of
files are going to move. I think this is a good thing.

Also, the tsc-alias seems pretty standard.

>

package.json
Argeento: Are you asking about package tsc-alias or about using path aliases?

tsc-alias can be removed. In this pr at the end I don't replace server's imports. 

Path aliases are much cleaner to use, especially in tests files. 

People are familiar with aliases. Every popular framework use them. Moreover they use path aliases in the same form: @ is pointing at src dir.

Argeento:
@bafolts `tsc-alias` has been removed

bafolts (approved):

kberg:
Is there something I need to add to make this compatible with vscode?

image

Argeento:
> Is there something I need to add to make this compatible with vscode?

No. Just restart vscode / TS server

kberg:
Yeah, I see that now! Works great. Thank you!

>

PR #3574, opened 2021-09-05
Remove return value from game.increaseVenusScaleLevel

kberg:
Since it doesn't actually return SelectSpace, make it return nothing.

### DIFF
--- a/src/cards/venusNext/GiantSolarShade.ts
+++ b/src/cards/venusNext/GiantSolarShade.ts
@@ -36,7 +36,8 @@ export class GiantSolarShade extends Card {
   }
 
   public play(player: Player) {
-    return player.game.increaseVenusScaleLevel(player, 3);
+    player.game.increaseVenusScaleLevel(player, 3);
+    return undefined;
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/cards/venusNext/GiantSolarShade.ts
```diff
@@ -36,7 +36,8 @@ export class GiantSolarShade extends Card {
   }
 
   public play(player: Player) {
-    return player.game.increaseVenusScaleLevel(player, 3);
+    player.game.increaseVenusScaleLevel(player, 3);
+    return undefined;
```
Argeento, line 40: Why `return undefined`? JS functions by default return `undefined`

kberg: Good question. Because if not, we get this error:

image

```
Property 'play' in type 'MicroMills' is not assignable to the same property in base type 'IProjectCard'.
Type '(player: Player) => void' is not assignable to type '(player: Player) => PlayerInput | undefined'.
Type 'void' is not assignable to type 'PlayerInput | undefined'.
```

Argeento: I figured it out. This comment was caused by my misunderstanding of the difference between `void` and `undefined`.  Sorry.
([This is a good explanation why TS throws an error here](https://stackoverflow.com/questions/58885485/why-does-typescript-have-both-void-and-undefined/58885486#58885486) if you are curious)

bafolts (approved):

PR #3576, opened 2021-09-06
Add the Moon standard project variant.

### LINKED ISSUE #2951: [Moon] Consider the softer variant for moon standard projects to bootstrap players
From the original rulebook:

Variant 1: You can use steel as 2MC to pay for “Build Road” standard action and use titanium as 3MC to pay for “Build Mine” and “Build Colony” actions.

Variant 2: You can pay 3MC instead of paying 1 Steel for “Build Road” and pay 4MC instead of paying 1 Titanium for “Build Mine” and “Build Colony” actions.

kberg:
Fixes #2951

### DIFF
--- a/src/cards/moon/MoonColonyStandardProject.ts
+++ b/src/cards/moon/MoonColonyStandardProject.ts
@@ -12,6 +12,16 @@ import {AltSecondaryTag} from '../render/CardRenderItem';
 
 
 export class MoonColonyStandardProject extends StandardProjectCard implements IMoonCard {
+  // See if it's possible to move PROPERTIES out of being a private static.
+  // ref: https://github.com/bafolts/terraforming-mars/pull/3576#discussion_r705835458
+  // " Was this change made for ease of reading? Making this a static property this code
+  // will run while the  server or client is starting versus only when the constructor is
+  // called. It will also remain in memory. If the change was made for ease of reading I
+  // would use a function getProperties."
+  //
+  // I tried making a readable overridable subclass, but that didn't work.
+  //
+  // Note: This applies to all three Moon standard projects.
   private static PROPERTIES = {
     name: CardName.MOON_COLONY_STANDARD_PROJECT,
     cost: 22,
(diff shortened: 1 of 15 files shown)


src/cards/moon/MoonColonyStandardProject.ts
```diff
@@ -10,22 +10,25 @@ import {IMoonCard} from './IMoonCard';
...
-      metadata: {
-        cardNumber: '',
-        renderData: CardRenderer.builder((b) =>
-          b.standardProject('Spend 22 M€ and 1 titanium to place a colony on the moon and raise your M€ production 1 step.', (eb) => {
-            eb.megacredits(22).titanium(1).startAction.moonColony().secondaryTag(AltSecondaryTag.MOON_COLONY_RATE).production((pb) => pb.megacredits(1));
-          }),
-        ),
-      },
-    });
+  private static PROPERTIES = {
```
bafolts, line 15: Was this change made for ease of reading? Making this a static property this code will run while the server or client is starting versus only when the constructor is called. It will also remain in memory. If the change was made for ease of reading I would have use a function `getProperties`.

bafolts (changes requested):
I would revert the change of using the static properties. Static properties will execute when the javascript is first executed versus only when the constructor is called. They will also remain in memory even if the constructors are never called.

src/cards/moon/MoonColonyStandardProject.ts
kberg: Unfortuantely it was not for reading, it was to make it work. I'll add a TODO for this request, though.

kberg:
> I would revert the change of using the static properties. Static properties will execute when the javascript is first executed versus only when the constructor is called. They will also remain in memory even if the constructors are never called.

I wish I could. Since this is only two cards, I'm going to merge first, and then I'll see if I can come up with something better this afternoon. I realize the "since this is only two cards" argument is a slippery slope once we apply it to all the cards. :D

bafolts (approved):

kberg:
Thanks! I'll merge this and get right to it.

PR #3577, opened 2021-09-07
Remember which modules are selected in the card ui URL.

kberg:
Turned out to be pretty easy after all.

### DIFF
--- a/src/components/DebugUI.vue
+++ b/src/components/DebugUI.vue
@@ -146,7 +146,18 @@ ALL_CARD_MANIFESTS.forEach((manifest) => {
   });
 });
 
-const ALL_MODULES = 'bcpvot*ram';
+const BASE = 'b';
+const CORP = 'c';
+const PRELUDE = 'p';
+const VENUS = 'v';
+const COLONIES = 'o';
+const TURMOIL = 't';
+const COMMUNITY = '*';
+const PROMO = 'r';
+const ARES = 'a';
+const MOON = 'm';
+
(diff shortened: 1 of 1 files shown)


kberg:
(Oh, it could be cleaner, but I'm really pleased. :D)

src/components/DebugUI.vue
```diff
@@ -190,16 +192,80 @@ export default Vue.extend({
     if (searchString) {
       this.filterText = searchString;
     }
+    const modules = urlParams.get('m') || ALL_MODULES;
+    this.base = modules.includes('b');
```
bafolts, line 196: If `watch` runs whenever `base`, `corporateEra`, or the other properties are modified then this code may be highly inefficient. Does it end up calling to `upateUrl` 10 times as each property is set?

src/components/DebugUI.vue
```diff
@@ -146,6 +146,8 @@ ALL_CARD_MANIFESTS.forEach((manifest) => {
   });
 });
 
+const ALL_MODULES = 'bcpvot*ram';
```
bafolts, line 149: I would create a constant for each module instead of using the separate characters.

For example:

```
const MODULE_BASE = 'b';
const MODULE_CORPORATE_ERA = 'c';
const ALL_MODULES = `${MODULE_BASE}${MODULE_CORPORATE_ERA}`;
```

bafolts (approved):

src/components/DebugUI.vue
kberg: On the initial call? It might, but it's also not early as inefficient as loading the page overall. I did not notice anything being particularly slower. Will take a peek again.

src/components/DebugUI.vue
kberg: Fair.

PR #3578, opened 2021-09-07
Rename CardMetadata -> ICardMetadata. 

kberg:
As I start to detach the card metadata interface, I'm going to take advantage of the (inconsistently applied) standard in this project to start interfaces with I. As I've been prototyping a solution, not knowing what's a class and what's an interface has been a significant point of confusion.

Note that some of the changes include removing CardMetadata for some cards and renaming the interface in others. For some reason I don't yet understand, some of those parameters need the declaration. So I'm not sure if I'm causing a breakable issue by doing this.

### DIFF
--- a/src/cards/Card.ts
+++ b/src/cards/Card.ts
@@ -1,4 +1,4 @@
-import {ICardMetadata} from './CardMetadata';
+import {ICardMetadata} from './ICardMetadata';
 import {CardName} from '../CardName';
 import {CardType} from './CardType';
 import {IAdjacencyBonus} from '../ares/IAdjacencyBonus';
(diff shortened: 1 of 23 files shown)


Argeento:
[Maybe this inconsistently applied standard should be dropped, not develop](https://stackoverflow.com/a/2814831)

kberg:
> [Maybe this inconsistently applied standard should be dropped, not develop](https://stackoverflow.com/a/2814831)

Thanks. Though I haven't read the article, I'm in the pro I-prefix camp.

bafolts:
I am also in the "pro I-prefix camp". With matters of styling such as whether an interface name starts with a capital I or not I would leave this up to a linter. This is going to be a matter of preference between developers. If we don't have a linter enforcing this one way or the other we will continue to see both patterns used.

PR #3580, opened 2021-09-09
Make Spectator Mode refresh.

kberg:
playerkey was not documented, and I wondered what purpose it served.
Now I know!

### DIFF
--- a/src/components/App.ts
+++ b/src/components/App.ts
@@ -40,6 +40,9 @@ interface MainAppData {
      */
     spectator?: SpectatorModel;
     playerView?: PlayerViewModel;
+    // playerKey might seem to serve no function, but it's basically an arbitrary value used
+    // to force a rerender / refresh.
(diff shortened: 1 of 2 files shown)


src/components/App.ts
```diff
@@ -164,12 +167,15 @@ export const mainAppSettings = {
...
         if (xhr.status === 200) {
           app.spectator = xhr.response as SpectatorModel;
+          app.playerkey++;
           app.screen = 'spectator-home';
-          window.history.replaceState(
-            xhr.response,
-            `${constants.APP_NAME} - Game`,
-            '/spectator?id=' + app.spectator.id,
-          );
+          if (window.location.pathname !== '/spectator') {
```
kberg, line 172: This if block wasn't strictly necessary now, but it matches the player behavior, and will prevent endless unnessecary window history.

bafolts (approved):

PR #3583, opened 2021-09-09
Separate client files from server

Argeento:
Nothing is changed, only files moved. Now the tree look like this:
```
| -📁 src/
   | -📁 client/
      | - 📁 components/
      | - 📁 directives/
      | - 📁 mixins/
      | - 📁 utils/
      | - main.ts
      | - types.ts
      | - vue-shims.d.ts
```

(The styles dir is missing here, I will fix this later with webpack css generating)


kberg (approved):
Thise is a lot of changes at once, but it's very easy to review. This change looks good.

PR #3584, opened 2021-09-09
Remove BasePlayerModel, and replace its only use.

kberg:
Feature is preserved: 
image

I simplified Card by removing the owner benavior from it and used Vue slots: https://vuejs.org/v2/guide/components-slots.html putting all the heavy lifting in SelectCard.

I also added an optimization - as it was, `getOwner` was called twice for each card. A naive solution would call `getOwner` three times, and I wanted something more efficient, so I cached the results each time the card list changes.


kberg:
Dangit, I forgot to run the tests. On it.

src/components/SelectCard.vue
```diff
@@ -32,9 +38,15 @@ import {PlayerInputModel} from '@/models/PlayerInputModel';
 import {sortActiveCards} from '@/components/ActiveCardsSortingOrder';
 import {TranslateMixin} from '@/components/TranslateMixin';
 
+interface Owner {
```
Argeento, line 41: Missing I-preffiex

src/components/SelectCard.vue
```diff
@@ -6,8 +6,13 @@
               <input v-if="isSelectOnlyOneCard()" type="radio" v-model="cards" :value="card" />
               <input v-else type="checkbox" v-model="cards" :value="card" :disabled="playerinput.maxCardsToSelect !== undefined && Array.isArray(cards) && cards.length >= playerinput.maxCardsToSelect && cards.includes(card) === false" />
             </template>
-            <Card v-if="playerinput.showOwner && getOwner(card) !== undefined" :card="card" :owner="getOwner(card)" />
-            <Card v-else :card="card"  />
+            <Card :card="card">
+              <template v-if="playerinput.showOwner">
+                <div :class="'card-owner-label player_translucent_bg_color_'+ getOwner(card).color">
```
Argeento, line 11: Please do not create classes like that. It's hard to read and easy to break something.
```vue
<div class="card-owner-label" :class="`player_translucent_bg_color_${getOwner(card).color}`">
```

src/components/SelectCard.vue
```diff
@@ -6,8 +6,13 @@
...
+                  {{getOwner(card).name}}
```
Argeento, line 12: Be consistent in putting spaces around brackets or use prettier. Interpolations above and below have these spaces

src/components/card/Card.vue
```diff
@@ -12,11 +12,7 @@
             <CardExpansion :expansion="getCardExpansion()" :isCorporation="isCorporationCard()"/>
             <CardResourceCounter v-if="card.resources !== undefined" :amount="getResourceAmount(card)" />
             <CardExtraContent :card="card" />
-            <template v-if="owner !== undefined">
-              <div :class="'card-owner-label player_translucent_bg_color_'+ owner.color">
-                {{owner.name}}
-              </div>
-            </template>
+            <slot></slot>
```
Argeento, line 15: It's
```vue
<slot />
```

```vue
<slot>slot is when you set a default value</slot>
```

src/components/SelectCard.vue
kberg: ack.

src/components/SelectCard.vue
kberg: This is coming from straight from Card.vue, so I am preserving the original behavior that I find not critical to change. I realize that you have different perspectives on `` vs + than I do, or the repository as it was, but we do not have such a standard in this repo.

src/components/SelectCard.vue
kberg: Thanks for the observation. This kind of consistency is not something we call out in this repository; in general we strive for some form of readability, but we don't have this kind of standard in the repo.

src/components/card/Card.vue
kberg: Gotcha.

src/components/SelectCard.vue
Argeento: The problem is not in string concatenation but with classes definition. Keep static class in `class` attribute. If for some reason you must put static classes into to dynamic definition: pass an array of classes.

```vue
<div :class="['card-owner-label', 'player_translucent_bg_color_' + getOwner(card).color}]">
```

src/components/SelectCard.vue
Argeento: Noticed. No more comments like this

bafolts (approved):

PR #3585, opened 2021-09-10
Make ApiWaitingFor spectator-friendly.

kberg:
This triggered a couple of changes: the expectation that players start
with 'p' and spectators start with 's' means that TestPlayer has to
generate its players correctly, which had a follow-on effect to some
other tests. Nothing so major that it deserved its own PR.


bafolts (changes requested):
I would make a separate function for spectators instead of having a value of `undefined` for the `player` argument represent a spectator. I think this can be broken up as follows.

```
private getWaitingForModel(game: Game, gameAge: number, undoCount: number): WaitingForModel {
  if (game.gameAge > gameAge || game.undoCount > undoCount) {
    return {result: 'REFRESH'};
  }
  return {result: 'WAIT'};
}

private getWaitingForModelPlayer(player: Player, gameAge: number, undoCount: number): WaitingForModel {
  if (this.timeToGo(player)) {
    return {result: 'GO'};
  }
  return this.getWaitingForModel(game, gameAge, undoCount);
}
```

This will avoid the situation where something calling to the function ends up with player being `undefined` by mistake and getting the spectator result.

kberg:
> I would make a separate function for spectators instead of having a value of `undefined` for the `player` argument represent a spectator. I think this can be broken up as follows.
> 
> ```
> private getWaitingForModel(game: Game, gameAge: number, undoCount: number): WaitingForModel {
>   if (game.gameAge > gameAge || game.undoCount > undoCount) {
>     return {result: 'REFRESH'};
>   }
>   return {result: 'WAIT'};
> }
> 
> private getWaitingForModelPlayer(player: Player, gameAge: number, undoCount: number): WaitingForModel {
>   if (this.timeToGo(player)) {
>     return {result: 'GO'};
>   }
>   return this.getWaitingForModel(player.game, gameAge, undoCount);
> }
> ```
> 
> This will avoid the situation where something calling to the function ends up with player being `undefined` by mistake and getting the spectator result.

Since it's a private method, this felt pretty safe. But, this makes sense.

src/routes/ApiWaitingFor.ts
```diff
@@ -17,15 +17,23 @@ export class ApiWaitingFor extends Handler {
...
+  private getPlayerWaitingForModel(player: Player, game: Game, gameAge: number, undoCount: number): WaitingForModel {
+    if (this.timeToGo(player)) {
       return {result: 'GO'};
     } else if (game.gameAge > gameAge || game.undoCount > undoCount) {
       return {result: 'REFRESH'};
     }
     return {result: 'WAIT'};
   }
 
+  // When player is undefined, caller is a spectator.
```
bafolts, line 29: This comment is no longer relevant.

bafolts (approved):

PR #3586, opened 2021-09-10
Show resource count when clicking on cards in LogPanel

nwai90:
Display resource count for cards clicked from LogPanel, if any

Screenshot 2021-09-10 at 11 39 36 AM

### DIFF
--- a/src/components/LogPanel.vue
+++ b/src/components/LogPanel.vue
@@ -22,7 +22,7 @@
         <div class="card-panel" v-if="cards.length > 0 || globalEventNames.length > 0">
           <Button size="big" type="close" :disableOnServerBusy="false" @click="hideMe" align="right"/>
           <div id="log_panel_card" class="cardbox" v-for="card in cards" :key="card">
-            <Card :card="{name: card}"/>
+            <Card :card="{name: card, resources: getResourcesOnCard(card)}"/>
           </div>
           <div id="log_panel_card" class="cardbox" v-for="globalEventName in globalEventNames" :key="globalEventName">
             <global-event :globalEvent="getGlobalEvent(globalEventName)" type="prior"></global-event>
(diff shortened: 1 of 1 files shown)


alrusdi (approved):

src/components/LogPanel.vue
```diff
@@ -334,6 +334,16 @@ export default Vue.extend({
         current: PartyName.GREENS,
       };
     },
+    getResourcesOnCard: function(cardName: CardName) {
```
Argeento, line 337: please use shorter syntax like in every other methods
```js
 getResourcesOnCard(cardName: CardName) {
 ```

File: src/components/LogPanel.vue [hunk 2]
```diff
@@ -334,6 +334,16 @@ export default Vue.extend({
...
+      let resourceCount = undefined;
+
+      for (const player of this.players) {
+        const foundCard = player.playedCards.find((card) => card.name === cardName);
+        if (foundCard !== undefined) resourceCount = foundCard.resources;
```
Argeento, line 342:
```js
if (foundCard) return foundCard.resoruce;
```
You don't have to create `resourceCount` variable. Moreover `for` loop can stop if card is found. (If you like this level of clarity with additional variable, just add `break`)

src/components/LogPanel.vue
```diff
@@ -334,6 +334,16 @@ export default Vue.extend({
         current: PartyName.GREENS,
       };
     },
+    getResourcesOnCard: function(cardName: CardName) {
+      let resourceCount = undefined;
```
Argeento, line 338: It's
```js
let resourceCount: number | undefined;
```
not 
```
let resourceCount = undefined;
```
Now you are creating variable with `any` type. Btw `undefined` is assign to variable by default. You don't have to explicitly define it.

Argeento (changes requested):

src/components/LogPanel.vue
kberg: This is a good tip. @nwai90 mind including this?

kberg (approved):

src/components/LogPanel.vue
nwai90: Agree. I updated the code for early `return`

PR #3587, opened 2021-09-10
Make ViewModel compatible with Spectator.

kberg:
This makes the spectator much more efficient when it comes to network bandwidth.


kberg:
This depends in #3585 and #3580.

src/client/components/GameEnd.vue
```diff
@@ -41,7 +41,7 @@
                         Go to main page
                     </a>
                 </div>
-                <div v-if="!isSoloGame() || game.isSoloModeWin" class="game-end-winer-announcement">
+                <div v-if="!isSoloGame || game.isSoloModeWin" class="game-end-winer-announcement">
```
bafolts, line 44: The `isSoloGame` is still a method. Was this intentional to change to `!isSoloGame`? I think this should still be `!isSoloGame() || game.isSoloModeWin`.

kberg: Yipe! You are right. Good catch, thank you.

src/client/components/GameEnd.vue
```diff
@@ -188,14 +211,14 @@ export default Vue.extend({
       return Timer.toString(p.timer);
     },
     getSortedPlayers() {
-      this.playerView.players.sort(function(a:PublicPlayerModel, b:PublicPlayerModel) {
+      this.viewModel.players.sort(function(a:PublicPlayerModel, b:PublicPlayerModel) {
```
bafolts, line 214: Not related to this change but worth noting that `Array.prototype.sort` sorts the array in-place. This is modifying the order of `this.viewModel.players`. We should really be copying the array first before sorting.

kberg: I'll follow up with that.

kberg: Acutally I say that, but I don't actually want to go through the manual verification process. :D So, TODO?

src/client/components/WaitingFor.vue
```diff
@@ -127,7 +128,11 @@ export default Vue.extend({
               return;
             } else if (result.result === 'REFRESH') {
               // Something changed, let's refresh UI
-              root.updatePlayer();
+              if (this.playerView.id.charAt(0) === 'p') {
```
bafolts, line 131: It may be worthwhile to make a function to check if an id is for a player versus a spectator. Here a function for `isPlayerId` would return true if the string started with `p`. If this logic ever changes it will make it easier to find and update all these checks.

src/models/PlayerModel.ts
```diff
@@ -7,6 +7,12 @@ import {TimerModel} from './TimerModel';
 import {GameModel} from './GameModel';
 import {PlayerId} from '../Player';
 
+export interface ViewModel {
+  game: GameModel;
+  players: Array<PublicPlayerModel>;
+  id: string; // PlayerId or SpectatorId, as determined by the app.
```
bafolts, line 13: It may not make a difference as both `PlayerId` and `SpectatorId` both are typed as `string` but the `id` property could be of type `PlayerId | SpectatorId`. With that type in place it may make the comment irrelevant.

src/client/components/WaitingFor.vue
kberg: That's fair. THAT will be a follow-up.

src/models/PlayerModel.ts
kberg: I like that idea. Fixing that now.

bafolts:
When I run this branch locally I see a warning on the game home screen with at least a 2 player game.

```
vue.esm.js:1897 TypeError: Cannot read properties of undefined (reading 'spectatorId')
```

kberg:
> When I run this branch locally I see a warning on the game home screen with at least a 2 player game.
> 
> ```
> vue.esm.js:1897 TypeError: Cannot read properties of undefined (reading 'spectatorId')
> ```

Hmm. Can you give me a little more context? You see this in the debug view?

bafolts:
After pulling down the latest commit I no longer see the warning. When I click to start a game though I am always taken to a blank screen. Are you able to start a game on this branch?

kberg:
I'm running a fresh set of tests right now. I'll start it up and let you
know.

>

src/models/PlayerModel.ts
```diff
@@ -6,6 +6,13 @@ import {PlayerInputModel} from './PlayerInputModel';
 import {TimerModel} from './TimerModel';
 import {GameModel} from './GameModel';
 import {PlayerId} from '../Player';
+import {SpectatorId} from '../Game';
+
+export interface ViewModel {
```
Argeento, line 11: Missing I-prefix

kberg:
> I'm running a fresh set of tests right now. I'll start it up and let you know.
> […](#)

Found it. Provided an isolated PR in #3611. Sort bites us, but in the board spaces!

src/models/PlayerModel.ts
kberg: Thanks -- sorry if it wasn't clear: like the spacing, including or excluding I isn't a hard rule, so there's no need to point these out.

src/client/components/GameEnd.vue
kberg: ... and I fixed and tested it.

bafolts (approved):

PR #3589, opened 2021-09-11
Add type filter to Card view.

kberg:
I renamed the modules cards - I should have matched @bafolts 's proposed naming convention after all. :sob:


src/components/DebugUI.vue
```diff
@@ -146,33 +156,34 @@ ALL_CARD_MANIFESTS.forEach((manifest) => {
...
+const MODULE_PROMO = 'r';
+const MODULE_ARES = 'a';
+const MODULE_MOON = 'm';
 
-const ALL_MODULES = `${BASE}${CORP}${PRELUDE}${VENUS}${COLONIES}${TURMOIL}${COMMUNITY}${PROMO}${ARES}${MOON}`;
+const ALL_MODULES = `${MODULE_BASE}${MODULE_CORP}${MODULE_PRELUDE}${MODULE_VENUS}${MODULE_COLONIES}${MODULE_TURMOIL}${MODULE_COMMUNITY}${MODULE_PROMO}${MODULE_ARES}${MODULE_MOON}`;
 
 export interface DebugUIModel {
   filterText: string,
-  filterDescription: boolean | unknown[],
```
kberg, line 164: There's a bunch of these `| unknown[]` that are no longer necessary, so it's good to clean them up.

kberg:
I thought I was the big deal using `Omit<Record<>` but turns out I was not!

src/components/DebugUI.vue
```diff
@@ -333,7 +371,9 @@ export default Vue.extend({
         }
       }
 
-      switch (card?.module) {
+      if (!(this.types as any)[card.card.cardType]) return false;
```
Argeento, line 374: why `as any`? types here seem to be correct

bafolts: I don't think this `as any` is needed either. Seems to compile fine without it. I try to avoid any usage of `any` as it almost always eventually leads to a bug in the future during refactors.

bafolts (approved):
I would remove the `as any` usage prior to merging.

src/components/DebugUI.vue
kberg: Thanks both of you.  I don't know why I needed it in the first place. I agree that `any` is trouble. Looks like you're both right - it compiles fine without it.

PR #3591, opened 2021-09-11
play project after paying

### LINKED ISSUE #3579: Helion + Standard Tech not working?
From discord:

```
asimov.io: I'm seeing a bug with how hellion pays
asimov.io: You activate the effects before choosing the means of payment
asimov.io: I.e., you see the card discarded to Search for life before paying, you get the 3 rebate from std tech before etc
k9: Which server? Game link?
🦵leggup🦵: I tested on main server. Helion with Standard Technology up. I did SP Asteroid. It instantly asked how much to pay of heat vs cash. Are you talking about something else? It won't ask you if you have zero heat.
🦵leggup🦵: Ahhh I see the bug now. I'm not getting money back from Std Tech :(
🦵leggup🦵: https://terraforming-mars.herokuapp.com/player?id=p342c56885d2
🦵leggup🦵: Hit undo on the asteroid I just dropped. Pay with $14 of the $15. End up with $1. It should then rebate $3 (standard technology)
```

bafolts:
When playing as at least Helion a player should not be rewarded for playing the standard project until they have paid. This changes the ordering to run the `projectPlayed` after paying. This should close #3579


kberg (approved):
A test of some kind would be good here, any thoughts?

bafolts:
> A test of some kind would be good here, any thoughts?

I can add a test of one kind here.

PR #3596, opened 2021-09-12
Moving requirement calculation to Player.

kberg:
When looking last night at removing the proliferated code for managing the Reds policy, I noticed something mildly problematic about how `Card.canPlay` is repsonsible for requirements testing. This makes for an interesting case. Should the Reds policy code also go in there? It shouldn't - please take that for given right now. But what that means is: _sometimes a card's properties are evaluated in `player.canPlay()`, sometimes in `Card.canPlay()` and sometimes in `<CardInstance>.canPlay`._

And so, if a card currently has requirements, and also requirements in its own `canPlay` that's easy to write up, e.g. `return super.canPlay(player) && player.plants > 3;`

But then if Reds processing shouldn't be in `Card.canPlay`, then why have _some_ global card processing in `Card.canPlay` and some in `Player.canPlay`?

So this PR shows a proposal for moving the requirements processing to `Player.canPlay`. And it works great, but it means having to fix the way we test all our cards.

I can't just change the tests from 

```
it('test can play', function() {
  expect(card.canPlay(player).is.true;
});
```

to

```
it('test can play', function() {
  expect(player.canPlay(card).is.true;
});
```

Because `Player.canPlay` also calls `Player.canAfford`. And then all the tests will fail because they don't take a player's megacredit count into consideration.

One option is the one you see, where I separate `Player.canPlay` into "Can the player pay for this?" and "Can the player meet the card's specific requirements?" Right now that's done with an awkwardly-named `Player.canPlayForFree` and should be something else.


kberg:
All of this means that we can, and should, move as much reds policy costs and requirements evaluation to `Player.canPlay.` If this looks good as an idea, and we agree on a structure that `Player.canPlay` can look like, I can start by slowly moving all the tests to use a different form of Player.canPlay.

kberg:
Some alterantives:
1. change `TestPlayer` so it defaults to having 50MC, and switch all tests of `card.canPlay(player)` to `player.canPlay(card)`
2. Change `card.canPlay(player)` to `card.canPlay(player, checkCosts: boolean = true) and switch all tests of `card.canPlay(player)` to `player.canPlay(card, false)`

src/Player.ts
```diff
@@ -1806,7 +1806,19 @@ export class Player implements ISerializable<SerializedPlayer> {
...
+    if (!canAfford) {
+      return false;
+    }
+
+    return this.canPlayForFree(card);
+  }
+
+  // Verify if requirements for the card can be met, ignoring the project cost.
+  public canPlayForFree(card: IProjectCard): boolean {
+    if (card.requirements !== undefined && !card.requirements.satisfies(this)) {
```
kberg, line 1818: This is moved out of Card.

src/cards/Card.ts
```diff
@@ -88,10 +88,7 @@ export abstract class Card {
   public get reserveUnits(): Units {
     return this.properties.reserveUnits || Units.EMPTY;
   }
-  public canPlay(player: Player) {
-    if (this.properties.requirements === undefined) {
-      return true;
-    }
-    return this.properties.requirements.satisfies(player);
+  public canPlay(_player: Player) {
```
bafolts, line 91: In `IProjectCard` we should declare that `canPlay` must be defined instead of optional. This will avoid having to check if `card.canPlay === undefined` in various places.

https://github.com/bafolts/terraforming-mars/blob/main/src/cards/IProjectCard.ts#L7

bafolts (approved):
I think this is a very good change. It removes a lot of duplicated code and leads to less code we have to write in each project card. I would also modify the `IProjectCard` interface and require that `canPlay` be defined. This will allow us to remove the checks for `canPlay` being `undefined`. The `canPlay` being `undefined` is a relic from long ago.

src/Player.ts
```diff
@@ -1806,7 +1806,19 @@ export class Player implements ISerializable<SerializedPlayer> {
...
-    return canAfford && (card.canPlay === undefined || card.canPlay(this));
+    if (!canAfford) {
+      return false;
+    }
+
+    return this.canPlayForFree(card);
+  }
+
+  // Verify if requirements for the card can be met, ignoring the project cost.
+  public canPlayForFree(card: IProjectCard): boolean {
```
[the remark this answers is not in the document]
kberg: That also doesn't seem quite right, because a) cost is a card requirement, and b) this includes things like finding out if a player can place a city. `canPlayIgnoringCost`?

PR #3598, opened 2021-09-13
html escape errors before writing to HTTP response.

kberg:
If not possible, don't write the error to the HTTP response.

This addresses a security scan of the code base: https://github.com/kberg/terraforming-mars/security/code-scanning/3?query=ref%3Arefs%2Fheads%2Fmain

html-escaper is listed as an indirect dependency of our project, which `istanbul-reports` depends on, and even though the risk is low, I'd rather not write my own HTML escaper, or rely on the deprecated Javascript escaper.


bafolts:
I see a 404 response when I try to visit the link you provided. Do you have screenshot or snippet of the security flaw? If taking on this dependency we will also need to add it to `package.json` as well as adding the types file.

kberg:
Interesting, URL works for me: 
image

kberg:
> Is using `html-escaper` I would move it to `dependencies`. Otherwise if it were ever removed from the dependency where we happen to pick it up the project wouldn't run. This change is adding direct dependency to `html-escaper`.

I can't parse what you are saying here.

bafolts:
You need to add `html-escaper` to `package.json`.

If it isn't included in `package.json` and it ever stopped being a dependency of `istanbul-reports` we would run into a build error that `html-escaper` was not found.

kberg:
> You need to add `html-escaper` to `package.json`.
> 
> If it isn't included in `package.json` and it ever stopped being a dependency of `istanbul-reports` we would run into a build error that `html-escaper` was not found.

Oh! OK, done. #til about dependencies vs devDependencies.

package.json
```diff
@@ -7,6 +7,7 @@
   },
   "private": true,
   "dependencies": {
+    "@types/html-escaper": "^3.0.0",
```
bafolts, line 10: I think this only includes the type files for `html-escaper`. This will need the `html-escaper` dependency as well.

kberg: I clearly don't know what I'm doing. Huh.

bafolts (approved):

PR #3599, opened 2021-09-13
Remove `| unknown[]` from the components.

kberg:


kberg:
These were added during our first attempt at Vue type checking, and no longer apply.

nwai90 (approved):

PR #3582, opened 2021-09-09
DX improvements

Argeento:
### 1. Add ts-mocha
No more building tests

---

### 2. Remove lint from build command
Linting / testing is not required to build the app. Especially when end user builds it.

---

### 3. Remove test cover from github ci
Does anybody use it there? Imo it was only slowing down the process

---

### 4. Replace `npm i` with `npm ci` in github ci
`npm ci` is created for CI/CD environments

---

### 5. Add dev server for backend
No more `npm run compile`. Dev server will now works on pure ts files and automatically reload after changes

---

### 6. Remove types checking from templates in test command
`vti diagnostic` is moved to the `lint:sfc` command, and it's added to github ci

---

### 7. Disable sourceMaps on server
This speeds up `npm run build:server` command. There is no need for maps when we have separate development environment (I decided to leave it on client side - sometimes users send to us screenshots from the console)

---

### 8. Remove `composite` mode from tsc
The only build we have now is server for production. There is no need to composite any parts of the app (it was also required to disable sourceMap)

---

### 9. Clean up scripts 
Less chaotic scripts naming with separation of client and server

PR #3600, opened 2021-09-13
Revert "DX improvements (#3582)"

kberg:
This reverts commit 3f96ac9218280cd5ac0ccbda5546a6ef4b8cfb6e.  It also reverts the change that tests files in tests/* but one thing at a time.


kberg:
Setting DNM for the moment.

bafolts (commented):
There are a variety of ways to do this building and testing code. We could spend most of our time changing this to use new tools. I had approved the change originally as it seemed to run `ts-` more than relying on javascript first being built. I felt this might be easier for new contributors to work with. After this is reverted are there plans to introduce it again? I don't think there is a perfect solution for how all of this is built and tested but I do think what we had is good enough.

kberg:
Oh absolutely we should restore this. In fact, I'd want to start restoring
a couple of these commands immediately, one at a time.

The problem is I'm losing my productivity in terms of debugging, and
spending more time fighting with the development environment than writing
code. This would be short-term only.

But I put the Do Not Merge tag on it, so I can try one more thing.

kberg:
Thanks to @Argeento, I've added the correct configurations to https://github.com/bafolts/terraforming-mars/wiki/Working-with-VSCode which are compatible with these new commands, so I'm closing this PR. Sorry about the fire drill.

PR #3604, opened 2021-09-14
Air Scrapping Variant of the Alt Venus board

kberg:
This comes from https://boardgamegeek.com/filepage/173900/variant-venus-board, a screenshot is below.

image

Coming after will be the resource bonuses, and the colony. Sometime in there, I'll add the UI option in `CreateGameForm`.


kberg:
image

bafolts (approved):

PR #3606, opened 2021-09-15
Alt-venus track bonuses (server side)

kberg:
Depends on #3604


src/inputs/SelectResources.ts
```diff
@@ -0,0 +1,63 @@
...
+        if (sum !== this.count) {
+          throw new Error(`Select ${this.count} resources.`);
+        }
+
+        this.player.addUnits(this.units, {log: true});
+        return undefined;
+      },
+      ...SelectResources.makeOptions(count, units),
+    );
+    this.title = this.title;
```
bafolts, line 61: This should be redundant since the argument in the constructor has `public title: string`. If you look at the generated javascript this gets set for us.

bafolts (approved):
Great change.

src/inputs/SelectResources.ts
kberg: Oh for certain, it's completely redundant. "this.title = this.title", I mean...

PR #3607, opened 2021-09-15
An attempt at a canSave UI.

kberg:


kberg:
Sorry, I hit the wrong button.

PR #3608, opened 2021-09-15
Alt-Venus bonuses client-side

kberg:
Depends on the first two PRs.


src/components/SelectNumber.vue
```diff
@@ -0,0 +1,61 @@
+<template>
+  <span>
+    <Button type="minus" class="btn btn-primary" @click="delta(-1)" data-test="minus" />
+    <!-- TODO(kberg): get rid of style: width; why is this necessary when it's not in SelectHowToPay? -->
+    <input type="text" :maxlength="maxlength" class="form-input form-inline payments_input"
```
Argeento, line 5: Why type=text instead of number? You can use [native html attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#additional_attributes) like `min`, `max`, `step`. Moreover to mobile/tablet users there will show up a num-keyboard instead of alpha-numeric

src/components/SelectNumber.vue
```diff
@@ -0,0 +1,61 @@
...
+  </span>
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+import Button from '@/components/common/Button.vue';
+
+export default Vue.extend({
+  name: 'SelectNumber',
+  model: {
```
Argeento, line 21: Why modify the default names? Which are `value` and `input`

src/components/SelectNumber.vue
```diff
@@ -0,0 +1,61 @@
...
+      required: false,
+      default: 2,
+    },
+  },
+  components: {
+    Button,
+  },
+  data() {
+    return {
+      componentValue: this.value,
```
Argeento, line 50: This component does not need its own state. Pass the prop directly to the input element

src/components/SelectNumber.vue
kberg: Well, this is how all the number components work (SelectHowToPay, SelectHowToPayForProjectCard, SelectProductionToLose.) There's SelectNumber, also, which I might just drop all of this and go use now that I properly understand the rules.

src/components/SelectDistinctResources.vue
```diff
@@ -0,0 +1,112 @@
...
+      titanium: 0,
+      plants: 0,
+      energy: 0,
+      heat: 0,
+      warning: undefined,
+    };
+  },
+  methods: {
+    ...TranslateMixin.methods,
+    hasWarning() {
```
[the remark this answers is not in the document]
kberg: Since the two vue files have been removed, I'm resolving all your comments, which were otherwise excellent. Thank you!

src/client/components/Board.vue
```diff
@@ -56,6 +56,17 @@
                     :class="'global-ares-severe-dust-storms-'+aresData.hazardData.severeDustStormOxygen.threshold"></div>
                 </div>
             </div>
+
+            <div v-if="altVenusBoard" class="global-alt-venus">
+              <div class="std-wild-resource p18"/>
```
bafolts, line 61: I don't think a self-closing `div` element is valid HTML.

The starting and ending tag are both mandatory.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/div

src/client/components/PlayerHome.vue
```diff
@@ -199,6 +200,7 @@
                           :boardName ="game.gameOptions.boardName"
                           :aresExtension="game.gameOptions.aresExtension"
                           :aresData="game.aresData">
+                          :altVenusBoard="game.gameOptions.altVenusBoard"
```
bafolts, line 203: Is this a typo? The element is closed on line 202.

src/client/components/Board.vue
kberg: Will fix, but it does not seem to be a problem:

```
~/git/terraforming-mars (vulns)

$ find src/client -name "*.vue" | xargs egrep "div[^/]*/>"

$ find src/client -name "*.vue" | xargs egrep "div[^>]*/>"
src/client/components/Awards.vue:            <div class="milestone-award-price" data-test="spot-price" v-text="spotPrice" />
src/client/components/BoardSpace.vue:    <div :class="getTileClass()" :title="getVerboseTitle(space.tileType)" data-test="tile"/>
src/client/components/card/CardExtraContent.vue:     
 <div v-if="lifeFound(card)" class="little-green-men" />
src/client/components/card/CardExtraContent.vue:     
 <div v-if="isMiningTileOnMetal(card,'steel')" class="mined-metal mined-steel" />
src/client/components/card/CardExtraContent.vue:     
 <div v-if="isMiningTileOnMetal(card,'titanium')" class="mined-metal mined-titanium" />
src/client/components/card/CardRenderItemComponent.vue:    <div :class="getComponentClasses()" v-for="index in itemsToShow()" v-html="itemHtmlContent()" :key="index"/>
src/client/components/card/CardTag.vue:  <div :class="getClasses()" />
src/client/components/Colony.vue:      <div v-if="colony.name === getEuropa()" style="height: 20px; visibility: hidden;display: block;" />
src/client/components/moon/MoonSpace.vue:    <div :class="tileClass" data-test="tile"/>
src/client/components/MoonGlobalParameterValue.vue:  
  <div class="moon-tile"/>
src/client/components/overview/PlayersOverview.vue:  
          <div v-if="playerView.players.length > 1" class="player-divider" />
src/client/components/Tag.vue:  <div :class="getClasses()" />

```

src/client/components/PlayerHome.vue
kberg: Must have been a bad merge, whoops! Good catch, thanks! I'll run it again.

bafolts (approved):

PR #3610, opened 2021-09-16
Add i18n plugin

Argeento:
No more importing `TranslateMixin` and `$t` function.

Now `$t` is available by default in `this.$t` method.

---

Adding new plugin to every test was a pain, so I created one main `getLocalVue` function in separate module


bafolts:
This is a lot of removed code and great change. I am going to give this a trial run tomorrow really quick before approving and merging.

kberg (commented):
Agreed, this is very nice. Thank you!

bafolts (approved):

PR #3612, opened 2021-09-16
Defensive copies to prevent infinite loops.

kberg:


kberg:
I have verifiedf all of these changes work in spectator mode.

bafolts (approved):

PR #3613, opened 2021-09-16
StealResources didn't run callback for solo games.

kberg:


bafolts (approved):
This change will enable someone to steal from themselves in a solo game. In a multi-player game a player will not be able to steal from themselves as the execute function still filters to only other players. Along with this change it would make sense to me to allow someone to steal from themselves in multi-player games as well as solo games.

kberg:
> This change will enable someone to steal from themselves in a solo game.

How does it do that? (FYI In solo games players steal from the neutral player.)

kberg:
> > This change will enable someone to steal from themselves in a solo game.
> 
> How does it do that? (FYI In solo games players steal from the neutral player.)

Also, this only impacts the card Ancient Shipyards.

bafolts:
I forgot about the neutral player.

PR #3615, opened 2021-09-16
Disable css caching more effectively.

kberg:
This makes it possible to locally change CSS without having to restart the server.

Dev still works:

image

And production reads from the cache (set `NODE_ENV=production` and verified with a debugger)

### DIFF
--- a/src/routes/ServeAsset.ts
+++ b/src/routes/ServeAsset.ts
@@ -1,7 +1,6 @@
 import * as http from 'http';
 import * as fs from 'fs';
 import * as path from 'path';
-import * as zlib from 'zlib';
 
 import {IContext} from './IHandler';
 import {BufferCache} from './BufferCache';
@@ -11,6 +10,21 @@ import {isProduction} from '../utils/server';
 
 type Encoding = 'gzip' | 'br';
 
+export class FileAPI {
+  public static readonly INSTANCE: FileAPI = new FileAPI();
+
+  protected constructor() {}
+
+  public readFileSync(path: string): Buffer {
+    return fs.readFileSync(path);
+  }
+  public readFile(path: string, cb: (err: NodeJS.ErrnoException | null, data: Buffer) => void): void {
+    fs.readFile(path, cb);
+  }
+  public existsSync(path: string): boolean {
+    return fs.existsSync(path);
+  }
+}
 export class ServeAsset extends Handler {
   public static readonly INSTANCE: ServeAsset = new ServeAsset();
   private readonly cache = new BufferCache();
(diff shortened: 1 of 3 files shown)


src/routes/ServeAsset.ts
```diff
@@ -11,25 +10,36 @@ import {isProduction} from '../utils/server';
...
   public constructor(private cacheAgeSeconds: string | number = process.env.ASSET_CACHE_MAX_AGE || 0,
     // only production caches resources
-    private cacheAssets: boolean = isProduction()) {
+    private cacheAssets: boolean = isProduction(),
+    private fileApi: FileAPI = FileAPI.INSTANCE) {
     super();
     // prime the cache with styles.css and a compressed copy of it styles.css
-    const styles = fs.readFileSync('build/styles.css');
-    this.cache.set('styles.css', styles);
-    zlib.gzip(styles, (err, compressed) => {
```
bafolts, line 26: Would uzing `gzipSync` avoid having to run `gzip` during the `make:css`? Using the `gzip` command will require a windows build user to have `gzip` installed.

kberg: Never heard of it. And I'm using Windows FWIW.

bafolts: If you change this to using `zlib.gzipSync` the styles.css file would be compressed synchronously. I think this would solve the problem being solved by calling to `gzip` when `make:css` is called. Having to run `gzip` during the build will require users to have `gzip` installed. The `zlib` dependency comes with `npm` so users with `npm` should have that dependency.

kberg: Aha, I see.

I'd like to argue that this way is better. Tell me your thoughts:

First, if I use zlib, then it no longer readable by the in-memory API used for testing.
Second, if Windows compatibility were a problem, then we'd have had people complaining about how the component tests weren't running recently.

If asking devs to install gzip-windows is unacceptable, then I would instead just ask to use OP-agnostic gzip-cli, https://www.npmjs.com/package/gzip-cli which has 3 dependencies.

bafolts (approved):

gilberterik:
Would someone mind changing it to short flags? Busybox gzip doesn't support double-dash long name flags.
```
-    "make:css": "lessc src/styles/common.less build/styles.css && gzip -9 --force --keep build/styles.css",
+    "make:css": "lessc src/styles/common.less build/styles.css && gzip -9 -f -k build/styles.css",
```

kberg:
Not a problem. Care to submit a PR?

>

kberg:
> Would someone mind changing it to short flags? Busybox gzip doesn't support double-dash long name flags.
> 
> ```
> -    "make:css": "lessc src/styles/common.less build/styles.css && gzip -9 --force --keep build/styles.css",
> +    "make:css": "lessc src/styles/common.less build/styles.css && gzip -9 -f -k build/styles.css",
> ```

https://github.com/bafolts/terraforming-mars/pull/3644 will replace this altogether.

PR #3618, opened 2021-09-17
Optimize playerinput cases for simpler card coding.

kberg:
This introduces two optimizations:
* If an OrOptions or AndOptions contains only one element, unwrap   that one option, and put it in the queue (instead of the And/OrOptions)
* If the return from a playerinput is just a single SelectOption,   run it immediately.

Right now we do all kinds of UX optimizations on the options presented to a player depending on conditions that could just be optimized out, letting the coder just write the conditions.

This kind of optimziation does change how tests are run, and slightly changes some of the behavior, but I suspect we can be careful enough to not prevent any significant timing issues.

I wonder whether this optimization would be better placed in `DeferredQueue`.

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -623,9 +623,7 @@ export class Player implements ISerializable<SerializedPlayer> {
   }
 
   public getResourcesOnCard(card: ICard): number | undefined {
-    if (card.resourceCount !== undefined) {
-      return card.resourceCount;
-    } else return undefined;
+    return card.resourceCount;
   }
 
   public getResourcesOnCorporation():number {
@@ -1439,12 +1437,7 @@ export class Player implements ISerializable<SerializedPlayer> {
 
     // Play the card
     const action = selectedCard.play(this);
-    if (action !== undefined) {
-      this.game.defer(new DeferredAction(
-        this,
-        () => action,
-      ));
-    }
+    this.queueAction(action);
 
     // Remove card from hand
     const projectCardIndex = this.cardsInHand.findIndex((card) => card.name === selectedCard.name);
@@ -1508,16 +1501,32 @@ export class Player implements ISerializable<SerializedPlayer> {
         const foundCard = foundCards[0];
         this.game.log('${0} used ${1} action', (b) => b.player(this).card(foundCard));
         const action = foundCard.action!(this);
-        if (action !== undefined) {
-          this.game.defer(new DeferredAction(
-            this,
-            () => action,
-          ));
-        }
+        this.queueAction(action);
         this.actionsThisGeneration.add(foundCard.name);
         return undefined;
       }, 1, 1, true,
     );
+  }
+
+  private queueAction(action: PlayerInput | undefined): void {
+    if (action === undefined) return;
+    let queuedAction = action;
+    if (action instanceof OrOptions || action instanceof AndOptions) {
+      if (action.options.length === 0) {
+        return;
+      }
+      if (action.options.length === 1) {
+        queuedAction = action.options[0];
+      }
+    }
+
+    // For a single SelectOption, run it now, do not defer.
+    // TODO(kberg): consider moving this, and all the input optomizations to DeferredActionQueue
+    if (queuedAction instanceof SelectOption) {
+      queuedAction.cb();
+    } else {
+      this.game.defer(new DeferredAction(this, () => queuedAction));
+    }
   }
 
   public drawCard(count?: number, options?: DrawCards.DrawOptions): undefined {
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1512,23 +1512,7 @@ export class Player implements ISerializable<SerializedPlayer> {
 
   private queueAction(action: PlayerInput | undefined): void {
     if (action === undefined) return;
-    let queuedAction = action;
-    if (action instanceof OrOptions || action instanceof AndOptions) {
-      if (action.options.length === 0) {
-        return;
-      }
-      if (action.options.length === 1) {
-        queuedAction = action.options[0];
-      }
-    }
-
-    // For a single SelectOption, run it now, do not defer.
-    // TODO(kberg): consider moving this, and all the input optomizations to DeferredActionQueue
-    if (queuedAction instanceof SelectOption) {
-      queuedAction.cb();
-    } else {
-      this.game.defer(new DeferredAction(this, () => queuedAction));
-    }
+    this.game.defer(new DeferredAction(this, () => action));
   }
 
   public drawCard(count?: number, options?: DrawCards.DrawOptions): undefined {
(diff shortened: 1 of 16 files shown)


kberg:
Another optimization that will help a great deal and take the burden off the cards, is to simplify cases like this into one:

And more useful would be to reduce these cases into one statement:

```
const availableMicrobeCards = player.getResourceCards(ResourceType.MICROBE);
if (availableMicrobeCards.length === 1) {
  const targetMicrobeCard = availableMicrobeCards[0];
  options.push(new SelectOption('Add 3 microbes to ' + targetMicrobeCard.name, 'Add microbes', () => {
    player.addResourceTo(targetMicrobeCard, {qty: 3, log: true});
    return undefined;
  }));
} else if (availableMicrobeCards.length > 1) {
  options.push(new SelectCard('Add 3 microbes to a card',
    'Add microbes',
    availableMicrobeCards, (foundCards: Array<ICard>) => {
      player.addResourceTo(foundCards[0], {qty: 3, log: true});
      return undefined;
    }));
}
```

```
options.push(new SelectCard('Add 3 microbes to a card',
  'Add microbes',
  availableMicrobeCards, (foundCards: Array<ICard>) => {
    player.addResourceTo(foundCards[0], {qty: 3, log: true});
    return undefined;
  }, {
    singleCardOption: 'Add 3 microbes to ${0}'
  }));
```

which would rewrite the SelectCard into a SelectOption when there's only 1 card. (Which, again, would be optimized by this PR.)

src/Player.ts
```diff
@@ -623,9 +623,7 @@ export class Player implements ISerializable<SerializedPlayer> {
   }
 
   public getResourcesOnCard(card: ICard): number | undefined {
-    if (card.resourceCount !== undefined) {
-      return card.resourceCount;
-    } else return undefined;
+    return card.resourceCount;
```
kberg, line 626: * separate isolated enhancement/reduction * I can move this into a separate PR,

src/Player.ts
```diff
@@ -1508,18 +1501,34 @@ export class Player implements ISerializable<SerializedPlayer> {
...
+      }
+      if (action.options.length === 1) {
+        queuedAction = action.options[0];
+      }
+    }
+
+    // For a single SelectOption, run it now, do not defer.
+    // TODO(kberg): consider moving this, and all the input optomizations to DeferredActionQueue
+    if (queuedAction instanceof SelectOption) {
+      queuedAction.cb();
```
bafolts, line 1526: This will run the `SelectOption` synchronously instead of waiting for it to be processed in the deferred actions queue. Will this cause any problems? If a `SelectOption` needed to run in the deferred action queue order it may now run too early.

src/Player.ts
```diff
@@ -1508,18 +1501,34 @@ export class Player implements ISerializable<SerializedPlayer> {
...
         return undefined;
       }, 1, 1, true,
     );
   }
 
+  private queueAction(action: PlayerInput | undefined): void {
+    if (action === undefined) return;
+    let queuedAction = action;
+    if (action instanceof OrOptions || action instanceof AndOptions) {
+      if (action.options.length === 0) {
```
bafolts, line 1515: It would make sense to me if in `Player.ts` we still added the items to the queue and then the queue could decide what to do with the options. After `Player.ts` is already over 1000 lines I think a quick "when in doubt, pull it out" applies.

src/cards/base/ElectroCatapult.ts
```diff
@@ -45,31 +45,26 @@ export class ElectroCatapult extends Card implements IActionCard, IProjectCard {
     return player.plants > 0 || player.steel > 0;
   }
   public action(player: Player) {
-    if (player.plants > 0 && player.steel > 0) {
-      return new OrOptions(
+    const orOptions = new OrOptions();
```
bafolts, line 48: I don't know if this will have any unintended consequences. Before this change if the call to `action()` returned `undefined` this represented there was no action to take. Would need to check every place we used this return value.

src/Player.ts
bafolts: This is great change and in separate PR would make sense. This is great change because it is less code and more efficient as no runtime check is involved.

kberg: Moved to #3624

src/Player.ts
kberg: Actually DAQ is the wrong place; by then the notion of the PlayerInput is fully abstracted. Player.ts is pretty huge, I agree, and there's definitely room to optimize, but perhaps this is not the place for it.

src/cards/base/ElectroCatapult.ts
kberg: > I don't know if this will have any unintended consequences. Before this change if the call to action() returned undefined this represented there was no action to take.

That's still true. But that's because the action was already taken. Here we're just deferring the action to slightly later. It can have some ordering issue, but that's why the optimization to not defer if all there is is a single option is a good idea.

> Would need to check every place we used this return value.

Will do. Commentary put inline, so the output isn't exactly what the shell command output.

```
$ find src/ -name "*.ts" | xargs egrep "\.action\([^']"
src/cards/base/BusinessNetwork.ts:          b.action(undefined, (eb) => eb.empty().startAction.empty()).text('Action: Look at the top card and either buy it or discard it', Size.SMALL, true).br;
src/cards/base/InventorsGuild.ts:          b.action(undefined, (eb) => eb.empty().startAction.empty()).text('Action: Look at the top card and either buy it or discard it', Size.SMALL, true);
src/cards/colonies/TitanFloatingLaunchPad.ts:          b.action(undefined, (eb) => {
src/cards/community/ProjectWorkshop.ts:            cb.action(undefined, (eb) => {
src/cards/promo/Astrodrill.ts:            ce.action(undefined, (eb) => {

^--- those uses of action are for card rendering.

src/Player.ts:      (card) => card[0].action(this),

^--- this is the one you're thinking of.

src/Player.ts:      action.options.push(convertPlants.action(this));
src/Player.ts:        return convertHeat.action(this);
src/Player.ts:      action.options.push(sellPatents.action(this));

^--- these are related to standard projects


src/turmoil/TurmoilHandler.ts:            () => scientistsPolicy.action(player),
src/turmoil/TurmoilHandler.ts:            () => kelvinistsPolicy.action(player),
src/turmoil/TurmoilHandler.ts:            () => kelvinistsPolicy.action(player),
src/turmoil/TurmoilHandler.ts:            () => greensPolicy.action(player),
src/turmoil/TurmoilHandler.ts:            () => marsFirstPolicy.action(player),
src/turmoil/TurmoilHandler.ts:            () => unityPolicy.action(player),
src/turmoil/TurmoilHandler.ts:            () => unityPolicy.action(player),
src/turmoil/TurmoilHandler.ts:            () => redsPolicy.action(player),

^--- these related to Turmoil, an action of a different interface.

```

src/Player.ts
kberg: Now that I think about it, ideally the right place for this is `Player.setWaitingFor()` byt then it gets slightly more tricky when ti comes down to the cb(). 

But also, I think the right thing to do is turn this into a prototype. I've added the other optimization I think is worthwhile, by eliminating the special case for not asking a player to select cards when there's only one card to select.

PR #3620, opened 2021-09-18
Remove the phrase "Effect" for cards where it does not apply.

kberg:


assets/translations.json
```diff
@@ -1 +1 @@
...fr":"Avoir au moins 16 cartes en main lorsque vous validez cet objectif","pl":"W momencie uzyskania tego Tytułu gracz musi mieć na ręce przynajmniej 16 kart","ru":"Соберите в руке не менее 16 карт"},"Polar Explorer":{"cn":"极地探险家","de":"Polarforscher","pl":"Odkrywca Polarny","ru":"Полярник"},"Requires that you have 3 tiles on the two bottom rows":{"cn":"底图最下方的两行上拥有至少3个版块","de":"Benötigt 3 eigene Plättchen auf den unteren beiden Gebietsreihen.","pl":"Gracz posiada w sumie przynajmniej 3 obszary w dwóch dolnych rzędach pól mapy","ru":"Выложите хотя бы 3 жетона на два нижних ряда"},"Rim Settler":{"cn":"木星定居者","de":"Grenzsiedler","pl":"Osadnik Rubieży","ru":"Пионер"},"Requires that you have 3 jovian tags":{"cn":"拥有3个木星标志","de":"Benötigt mindestens 3 eigene Jupitersymbole im Spiel.","pl":"Gracz posiada w grze przynajmniej 3 symbole Jowisza","ru":"На ваших картах должно быть не менее 3 меток Юпитера"},"Specialist":{"cn":"专业人士","de":"Spezialist","pl":"Specjalista","ru":"Специалист"},"Requires that you have at least 10 in production of any resource":{"cn":"拥有至少10点任意产能","de":"Eine Produktionskette mit Wert 10.","pl":"Gracz produkuje przynajmniej 10 jednostek dowolnego zasobu","ru":"Увеличьте
```
bafolts, line 1: Is this a generated file? Should it be in `.gitignore`?

bafolts (approved):
The `assets/translations.json` shouldn't be stored under source control since it is a generated file. The existence of that file is unrelated to this change though. I may put up PR to put it in `.gitignore`.

assets/translations.json
kberg: As we discussed, it isn't. I think we should just delete this.

PR #3621, opened 2021-09-18
Clarify card behavior by dropping annotating objects with builder, but instead supplying an attribute object.

kberg:


kberg:
After looking at it, is this that much better? It removes a bunch of code in CardItemRenderer, and upon further analysis will reduce some of the other compexity in it -- this is still just a prototype for that matter.

kberg:
Once #3647 goes in, I'm going to remove RFC from this PR.

src/cards/render/CardRenderer.ts
```diff
@@ -117,208 +116,152 @@ export class CardRenderCorpBoxAction extends CardRenderer {
 
 class Builder {
   protected _data: Array<Array<ItemType>> = [[]];
+  played: any;
```
bafolts, line 119: Is this property used? Does it have to use `any` as it's type?

src/cards/render/CardRenderItem.ts
```diff
@@ -46,19 +52,36 @@ export class CardRenderItem {
...
     }
     this.size = options.size;
     if (options.amount !== undefined) {
       this.amount = options.amount;
     }
+    this.anyPlayer = options.all;
+    this.isPlayed = options.played;
+    this.secondaryTag = options.secondaryTag;
+
+    if ((options.multiplier ?? false) === true) {
```
bafolts, line 78: This conditional is identical to `if (options.multiplier === true)`. If the `multiplier` property is `undefined` it would not equal `true`. The conversion to `false` when `undefined` is unnecessary.

kberg: Thank you.

src/cards/render/CardRenderer.ts
kberg: Huh! Nice catch. Looks like it is not used. Removing it changes nothing.

bafolts (approved):

PR #3624, opened 2021-09-21
Clean up around getResourcesOnCard.

kberg:


src/Player.ts
```diff
@@ -746,7 +744,7 @@ export class Player implements ISerializable<SerializedPlayer> {
   public getResourceCount(resource: ResourceType): number {
     let count: number = 0;
     this.getCardsWithResources(resource).forEach((card) => {
-      count += this.getResourcesOnCard(card)!;
+      count += (this.getResourcesOnCard(card) ?? 0);
```
bafolts, line 747: This is good change. Prior to this change if we added up adding `undefined` to a number we would have introduced `NaN`.

src/Player.ts
```diff
@@ -635,9 +635,7 @@ export class Player implements ISerializable<SerializedPlayer> {
   }
 
   public getResourcesOnCard(card: ICard): number | undefined {
```
bafolts, line 637: Would it simplify things to have `getResourcesOnCard` return `0` instead of `undefined`? With a quick grep I don't think we use the return value of `undefined` anywhere and we always assume it is `0`.

kberg: Maybe? Mind if I do that as a next PR?

bafolts (approved):

PR #3628, opened 2021-09-22
Add more output to export_game, and survive bad fetches.

kberg:


src/tools/export_game.ts
```diff
@@ -48,10 +48,26 @@ function load(gameId: string) {
...
+      db.getGameVersion(gameId, version, (err, serialized) => {
+        if (serialized === undefined) {
+          console.log(`failed to read version ${version}: ${err}`);
+          errors++;
+        } else {
+          console.log(`Storing version ${version}`);
+          localDb.saveSerializedGame(serialized!);
+          writes++;
+        }
+        if (errors + writes === game.lastSaveId + 1) {
```
kberg, line 67: image

See, the output returns out of order, so unless I did some kind of thread join ... is there even an async thread join? Anyway, does the job.

alrusdi (approved):

PR #3629, opened 2021-09-22
Luna Archives pre-cost calculation

### LINKED ISSUE #3626: canAfford does not consider science resources from Luna Archives
Payment calculation seems to incorporate science resources from Luna Archives, but the canAfford calculations do not. This leads to a situation where a player could pay for a Moon card using Science resources but can't select it due to insufficient MC.

kberg:
Science resources on Luna Archives reduces the cost of Moon cards by 1MC per resource. SelectHowToPayForProjectCard already takes this into account, but it wasn't taken into account when choosing valid cards to play.

Fixes #3626

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1295,6 +1295,10 @@ export class Player implements ISerializable<SerializedPlayer> {
     return card.tags.includes(Tags.VENUS);
   }
 
+  private canUseScience(card: ICard): boolean {
+    return card.tags.includes(Tags.MOON);
+  }
+
   private getMcTradeCost(): number {
     return MC_TRADE_COST - this.colonyTradeDiscount;
   }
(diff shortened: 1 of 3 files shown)


puuj:
This looks like it solves the problem, but why not use the same pattern as for steel, titanium, floaters, and microbes? Luna Archives is like Psychrophiles or Dirigibles, so why not have the code pattern be the same to make it easier for someone to understand the "this is how we spend other resources to buy cards" pattern?

kberg:
> This looks like it solves the problem, but why not use the same pattern as for steel, titanium, floaters, and microbes? Luna Archives is like Psychrophiles or Dirigibles, so why not have the code pattern be the same to make it easier for someone to understand the "this is how we spend other resources to buy cards" pattern?

I forget why I chose that. I'll go back and look again. I think perhaps because it is not optimal, fwiw. But yeah I'll probably do as you suggested. Thanks.

(Can you DM me on discord so I can assocaite you with the discord, please? I think we've done that before, but it's been a while. ;))

bafolts (approved):

PR #3630, opened 2021-09-22
Minor warnings in vscode that seem to make sense.

kberg:
Though why consolidating window._translations into one space got rid of
3 vscode errors I don't know.


kberg:
argh... computer.

PR #3633, opened 2021-09-23
perf: use some and every a few places

bafolts:
This uses the `some` and `every` functions in a few popular functions. These should be slightly more efficient as they exit earlier and avoid having to compare results. These old Array functions don't get enough respect!


kberg (approved):
Nice to see some more code out of you!

PR #3637, opened 2021-09-23
Show card resource icons instead of 'RES' in the corner.

kberg:
I don't like copying the CSS. I wonder if `less` can help here.

image

### DIFF
--- a/src/styles/cards_v2.less
+++ b/src/styles/cards_v2.less
@@ -1817,3 +1817,57 @@
         top: -16px;
     }
 }
+
+.card-resources-counter {
+
+    .card-resource {
+        position: relative;
+        display: inline-block;
+        margin-bottom: 4px;
+        width: 22px;
+        height: 22px;
+        background-size: 22px;
+        vertical-align: middle;
+    }
+    .card-resource-microbe {
+        background-image: url(./assets/resources/microbe.png);
+    }
+    .card-resource-animal {
+        background-image: url(./assets/resources/animal.png);
+    }
+    .card-resource-science {
+        background-image: url(./assets/resources/science.png);
+    }
+    .card-resource-floater {
+        background-image: url(./assets/resources/floater.png);
+    }
+    .card-resource-fighter {
+        background-image: url(./assets/resources/fighter.png);
+    }
+    .card-resource-camp {
+        background-image: url(./assets/resources/camp.png);
+    }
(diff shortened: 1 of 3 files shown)


kberg:
If the CSS copying can be resolved (or accepted as a copy, _eyyuch)_, this would be good to go.

kevinb9n:
Any particular preference for "[Floater]: 5" over "5 [Floater]"?

kberg:
I agree, swapping the order is nicer. Will do.

kberg:
Done: 

image

src/styles/cards_v2.less
```diff
@@ -1817,3 +1817,57 @@
...
+    .card-resource {
+        position: relative;
+        display: inline-block;
+        margin-bottom: 4px;
+        width: 22px;
+        height: 22px;
+        background-size: 22px;
+        vertical-align: middle;
+    }
+    .card-resource-microbe {
```
alrusdi, line 1832: You can try to reduce code repetioton by using lesscss each keyword like this:
`@resource_types: microbe, animal, science;

each(@resource_types, {
    .card-resource-@{value} {
        background-image: url("./assets/resources/@{value}.png");
    }
})`

alrusdi (approved):

kevinb9n:
I love this!

Have you checked that if one of these same cards like Security Fleet is on Self-Rep Robots that it doesn't accidentally show up as being fighter resources in that situation? It should look like blank white squares.

kberg:
> I love this!
> 
> Have you checked that if one of these same cards like Security Fleet is on Self-Rep Robots that it doesn't accidentally show up as being fighter resources in that situation? It should look like blank white squares.

I have not checked SRR. I almost never play or test with promos. But it seems worth checking before applying.

bafolts:
This is what gets displayed on the Self Replicating Robot cards.

image

This seems good enough for me for time being. It is at least not showing the fighter resource.

kberg:
Yeah, this is looking decent: 

image

bafolts (approved):

kberg (commented):
> This is what gets displayed on the Self Replicating Robot cards.
> 
> image
> 
> This seems good enough for me for time being. It is at least not showing the fighter resource.

Confirmed.

kberg:
I removed the "prototype" from the definition because of satisfactory CSS reduction.

PR #3640, opened 2021-09-24
PF1: Arabia Terra Board from Pathfinders.

kberg:
image

The one visualization issue is that the energy icon should be an energy production icon. Any thoughts there?


alrusdi:
Hide tiles button looks weird.

kberg:
> Hide tiles button looks weird.

Ah yes, that issue. That's a separate issue related to viewing the map on the select initial cards view versus the player view. Happens with all the maps.

PR #3643, opened 2021-09-24
PFf3: Add the first Pathfinders card, and hook up the rest of the manifest.

kberg:
Requires PF2 and PF1


src/client/components/DebugUI.vue
```diff
@@ -79,6 +79,12 @@
                 <div class="create-game-expansion-icon expansion-icon-themoon"></div>
                 <span v-i18n>The Moon</span>
               </label><span/>
+
+              <input type="checkbox" name="pathfinders" id="pathfinders-checkbox" v-model="pathfinders">
+              <label for="pathfinders-checkbox" class="expansion-button">
+                <div class="create-game-expansion-icon expansion-icon-pathfinders"></div>
+                <span v-i18n>Pathfinders</span>
+              </label><span/>
```
bafolts, line 87: Unrelated to this change but it seems the start and end tag is required for `span`.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/span

Most browsers are probably working around this for us.

bafolts (approved):

src/client/components/DebugUI.vue
kberg: I wonder what that span/ is doing there, actually. Another day.

PR #3647, opened 2021-09-25
Clean up CardRenderer.

kberg:
There was a lot of pop and push when it wasn't necessary in CardRenderer, making the class tricky to read. That and a little cleanup makes for a completely identical output.

I know this for two reasons: the first is that I visually inspected every card, and they matched on the `/cards` path. The second is that I extracted the full card renderings, and ran diffs until they were exactly the same:

image

I've done a fair bit of analysis of the non-method elements of the rendering classes, and I am confident that the export is the complete set.

### DIFF
--- a/src/cards/render/CardRenderer.ts
+++ b/src/cards/render/CardRenderer.ts
@@ -122,85 +122,54 @@ class Builder {
     return new CardRenderer(this._data);
   }
 
-  protected _getCurrentRow(): Array<ItemType> | undefined {
-    return this._data.pop();
-  }
-
-  protected _addRowItem(item: ItemType): void {
-    const currentRow = this._getCurrentRow();
-    if (currentRow !== undefined) {
-      currentRow.push(item);
-      this._data.push(currentRow);
-    }
-  }
-
-  protected _checkExistingItem(): void {
+  protected _currentRow(): Array<ItemType> {
     if (this._data.length === 0) {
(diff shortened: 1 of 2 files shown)


kberg:
Actually I need to run `npm run test` one more time. That was a pretty confident statement to make before running the tests, I hope I don't eat crow! :D

src/cards/render/CardRenderer.ts
```diff
@@ -704,67 +577,50 @@ class Builder {
...
+    const row = this._currentRow();
+    const item = row.pop();
+    if (!(item instanceof CardRenderItem)) {
+      throw new Error('"brackets" could be called on CardRenderItem only');
+    }
+
+    if (item === undefined) {
+      throw new Error('Called "brackets" without a CardRenderItem.');
     }
+    row.push(CardRenderSymbol.bracketOpen());
```
bafolts, line 621: The `push` function takes multiple arguments. This could be written:

```
row.push(CardRenderSymbol.bracketOpen(), item, CardRenderSymbol.bracketClose());
```

Noticed this while reviewing and realize this is unrelated to this change. Should be slightly more efficient.

bafolts (approved):

src/cards/render/CardRenderer.ts
kberg: ack. That's fair.

PR #3650, opened 2021-09-27
Pre tsc as a server pretest. ts-mocha runs all tests instead of stopping when there are compiler errors.

kberg:
This change makes identifying actual compiler issues much simpler. It also means that running `npm run pretest:server` in VS Code will create a reasonable set of problem markers for quick cleanup.


kberg:
This is still failing because the changes that were introduced in #3649 also have to be applied.

alrusdi (approved):

PR #3654, opened 2021-09-27
PF4: Five Pathfinders cards.

kberg:
Depends on PF3


src/cards/pathfinders/NewVenice.ts
```diff
@@ -0,0 +1,60 @@
...
+          }).nbsp.tile(TileType.OCEAN_CITY, false, true);
+        }),
+        description: 'Requires 3 ocean tiles. Lose 2 plants. Increase your energy production 1 step and your M€ production 2 steps. ' +
+            'Place this tile on top of an existing ocean tile, IGNORING NORMAL PLACEMENT RESTRICTIONS FOR CITIES. The tile counts as a city as well as an ocean.',
+      },
+    });
+  }
+
+  public canPlay(player: Player): boolean {
+    return super.canPlay(player) && (player.plants > 2);
```
bafolts, line 39: Should this be `player.plants >= 2` ? The requirement only seems to be that 2 plants are lost. Do we need to check for 3 oceans?

src/cards/pathfinders/RareEarthElements.ts
```diff
@@ -0,0 +1,44 @@
...
+          // TODO(kberg): add new tile image.
+          b.production(((pb) => pb.megacredits(1))).slash().emptyTile().text('special');
+        }),
+        description: 'Increase your M€ production by 1 for every special tile you own on Mars.',
+      },
+    });
+  }
+
+
+  private static INVALID_TILES = [TileType.GREENERY, TileType.OCEAN, TileType.CITY];
```
bafolts, line 31: I would use `Set` here as that is how the data is being used.

bafolts (approved):

src/cards/pathfinders/NewVenice.ts
kberg: Good catch. Also oceans are checked in super.canPlay().

src/cards/pathfinders/RareEarthElements.ts
kberg: Seems like a micro-optimization given how little this will impact overall performance, but done.

Issue #3655, opened 2021-09-27
Resolve how Luna Ecumenopolis should work


kberg:
Bug reporter
>Hello, i think there is something wrong with the card Luna Ecumenopolis with the Moon Module expension, even with the two colony placed on the moon, it just add the TR with the level of the moon colony track, not more colony tile on the moon. Is it a bug or do i understand the card wrong?

Me
> Luna Ecumenopolis's placement rules is that you place two colonies adjacent to two other colony tiles. So if there is no place to do that, you can't place the tiles. It's possible the card should be rewritten so that such a card isn't even playable unless you can do it. But the rules for that are a little tricky.

Kevin
> Card instructions are generally expected to be mandatory unless you write "you may" or "if possible" (so there are 3 different, I dunno, "levels of mandatory-ness") (but if you go that way, shouldn't you only get the colony track increases for tiles actually played?) (IOW, is this not supposed to be like the relationship between greenery tiles and the oxygen track?)

PR #3656, opened 2021-09-28
Make Luna Ecumenopolis not playable unless player can place 2 colonies.

kberg:
This was surprisingly easier than I expected. What am I missing?

The algorithm takes into account the possibility that the first
colony placed is necessary to place to second colony.


kberg:
Fixes #3655

bafolts:
Wow what a complex card!

Do the two placed tiles need to be adjacent?

Each of these tiles then needs to be adjacent to two other colony tiles which are already on the board?

kberg:
> Wow what a complex card!
> 
> Do the two placed tiles need to be adjacent?

No, it's just an implied effect that the first placed tile can often be a basis for the second-placed tile.

> 
> Each of these tiles then needs to be adjacent to two other colony tiles which are already on the board?
I think the answer above clarifies that this is not the case.

kberg:
See, it's not a complex card if you're looking at a real board in front of you.

bafolts (approved):

bafolts:
This is elegant solution.

PR #3657, opened 2021-09-28
corp card doesn't show resource icon under certain circumstances.

kberg:


kberg:
Here's what I mean by "under certain circumstances"

image

PR #3659, opened 2021-09-28
PF7: Add the data structure for the planetary tracks.

kberg:
This isn't going to have tests; that'll come in PF8, which will raise the tracks, and will verify that the tracks work.

BASED ON PF6.

### DIFF
--- a/src/pathfinders/PlanetaryTrack.ts
+++ b/src/pathfinders/PlanetaryTrack.ts
@@ -1,0 +1,54 @@
+import {Reward} from './Reward';
+
+export interface PlanetaryTrack {
+  spaces: Array<PlanetaryTrackSpace>;
+}
+
+export interface PlanetaryTrackSpace {
+  everyone: Array<Reward>;
+  risingPlayer: Array<Reward>;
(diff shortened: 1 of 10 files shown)


src/pathfinders/PlanetaryTrack.ts
```diff
@@ -0,0 +1,54 @@
...
+  mostTags: Array<Reward>;
+}
+
+export class TrackBuilder {
+  private spaces: Array<PlanetaryTrackSpace>;
+  private currentSpace: number;
+
+  public constructor(size: number) {
+    this.spaces = [];
+    for (let idx = 0; idx <= size; idx++) {
```
bafolts, line 19: If this isn't an "off by one" error disregard but I am surprised this is using `idx <= size` and not `idx < size`. The caller will end up with one more item than size in the spaces array.

bafolts (approved):

src/pathfinders/PlanetaryTrack.ts
kberg: It is, so I fixed it. Thanks.

PR #3660, opened 2021-09-29
PF9: Three more Pathfinders cards

kberg:
And a minor change to New Venice.

### DIFF
--- a/src/cards/pathfinders/MartianMonuments.ts
+++ b/src/cards/pathfinders/MartianMonuments.ts
@@ -1,0 +1,37 @@
+import {IProjectCard} from '../IProjectCard';
+import {Player} from '../../Player';
+import {Card} from '../Card';
+import {CardType} from '../CardType';
+import {CardName} from '../../CardName';
+import {CardRenderer} from '../render/CardRenderer';
+import {Resources} from '../../Resources';
+import {Tags} from '../Tags';
+import {CardRequirements} from '../CardRequirements';
+
+export class MartianMonuments extends Card implements IProjectCard {
+  constructor() {
+    super({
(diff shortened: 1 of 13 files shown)


src/client/components/card/CardRenderItemComponent.vue
```diff
@@ -223,6 +223,10 @@ export default Vue.extend({
           classes.push('card-tag-earth');
         } else if (type === CardRenderItemType.BUILDING) {
           classes.push('card-tag-building');
+        } else if (type === CardRenderItemType.CITY) {
+          classes.push('card-tag tag-city');
+        } else if (this.item.type === CardRenderItemType.MARS) {
+          classes.push('card-tag tag-mars');
```
kberg, line 229: This CSS doesn't exist yet, but i'll find its way in eventually. Plus, these cards aren't used yet.

src/cards/pathfinders/MartianMonuments.ts
```diff
@@ -0,0 +1,37 @@
...
+          b.production(((pb) => pb.megacredits(1))).slash().mars(1).played;
+        }),
+        description: 'Requires that you own a city on Mars. Raise your M€ production 1 step for every Mars tag you own (including this.)',
+      },
+    });
+  }
+
+
+  public play(player: Player) {
+    const count = player.getTagCount(Tags.MARS) + 1; // +1 is the "including this"
```
bafolts, line 32: Is a TODO needed on this card to implement a `canPlay` for the requirement that you must own a city on Mars?

bafolts (approved):
Forgot there is a default satisfies that uses the requirements for canPlay.

PR #3663, opened 2021-09-29
Remove Maxwell Base's resource type requirement.

kberg:


src/Player.ts
```diff
@@ -727,7 +727,7 @@ export class Player implements ISerializable<SerializedPlayer> {
     return result;
   }
 
-  public getResourceCards(resource: ResourceType | undefined): Array<ICard> {
+  public getResourceCards(resource: ResourceType | undefined = undefined): Array<ICard> {
```
bafolts, line 730: This can be written as `public getResourceCards(resource?: ResourceType): Array<ICard>`. This is identical to what is written here for the most part but avoids having to write out `undefined`.

bafolts (approved):

nwai90 (approved):

PR #3664, opened 2021-09-30
Pf7.5: Move the venus-focused resource gaining deferred action to /deferredActions

kberg:
I'm hoping to reuse most of it for Pathfinders. Though now I'm not so certain.


kberg:
Lost confidence this is a good idea. Hang back.

PR #3667, opened 2021-10-01
Add two more oceans and define Volcanic Spaces to Arabia Terra map.

kberg:


kberg:
image

src/client/components/Board.vue
```diff
@@ -154,6 +154,36 @@
                         <tspan x="4" dy="12">Mons</tspan>
                     </text>
                 </g>
+
+                <g v-if="boardName === 'arabia terra'" id="tikhonarov" transform="translate(487, 185)">
```
bafolts, line 158: I would use `v-if="boardName === BoardName.ARABIA_TERRA"` here instead of the string.

bafolts (approved):

src/client/components/Board.vue
kberg: BoardName isn't accessible in the template. :/

Issue #3662, opened 2021-09-29
Differentiate fan board from all boards both visually and for random selection.


kberg:

kberg:
https://github.com/nwai90/terraforming-mars/commit/e67065af08ec2e9174d324c78c39f70d5b6692f7

kberg:
This is fixed.

PR #3670, opened 2021-10-01
Remove Arabia Terra from the set of randomly-selected boards.

kberg:
A bit of a hack for the time being, I admit to accommodate #3662

### DIFF
--- a/src/routes/Game.ts
+++ b/src/routes/Game.ts
@@ -60,7 +60,7 @@ export class GameHandler extends Handler {
         }
 
         if (gameReq.board === 'random') {
-          const boards = Object.values(BoardName);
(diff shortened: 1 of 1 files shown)


src/routes/Game.ts
```diff
@@ -60,7 +60,7 @@ export class GameHandler extends Handler {
         }
 
         if (gameReq.board === 'random') {
-          const boards = Object.values(BoardName);
+          const boards = Object.values(BoardName).filter((name) => name !== BoardName.ARABIA_TERRA);
```
bafolts, line 63: A TODO may be appropriate here to come up with better solution.

bafolts (approved):

src/routes/Game.ts
kberg: While I would generally agree, we have the tracking bug. I'll take that on now.

PR #3671, opened 2021-10-01
Update packages, removing 3 high vulnerabilities.

kberg:


kberg:
Created with `npm run audit --fix`

package.json
```diff
@@ -24,7 +24,7 @@
     "mnemonist": "^0.38.1",
     "pg": "^8.2.1",
     "spectre.css": "^0.5.8",
-    "sqlite3": "^5.0.0",
+    "sqlite3": "^4.2.0",
```
nwai90, line 27: Note: As discussed, this looks strange as it is a major version downgrade of the `sqlite3` package rather than an upgrade. Will check in more detail re the exact changes across these two versions

bafolts: It doesn't seem like sqlite3 has had any recent updates. It was last published 8 months ago. Using 4.2.0 should be fine as I don't think we make use of any features added in more recent versions.

bafolts (approved):

nwai90 (approved):

PR #3672, opened 2021-10-02
Make coves compatible with land claim.

kberg:


src/boards/Board.ts
```diff
@@ -209,7 +209,7 @@ export abstract class Board {
 
   public getNonReservedLandSpaces(): Array<ISpace> {
     return this.spaces.filter((space) => {
-      return space.spaceType === SpaceType.LAND &&
+      return (space.spaceType === SpaceType.LAND || space.spaceType === SpaceType.COVE) &&
```
bafolts, line 212: Would it be clever or annoying if we eventually changed `spaceType` to use bit manipulation? This could then check that the land bit were set. The cove would set both ocean and land bits. Should be less data to store and send the number versus the string for the enum as well when serializing.

bafolts (approved):

src/boards/Board.ts
kberg: I think not helpful.

PR #3674, opened 2021-10-02
lastCardPlayed does not need to be a complex type

kberg:


src/cards/base/SpecialDesign.ts
```diff
@@ -24,7 +24,7 @@ export class SpecialDesign extends Card implements IProjectCard {
     });
   }
   public getRequirementBonus(player: Player): number {
-    if (player.lastCardPlayed !== undefined && player.lastCardPlayed.name === this.name) {
+    if (player.lastCardPlayed !== undefined && player.lastCardPlayed === this.name) {
```
bafolts, line 27: Similar to the change in `IndenturedWorkers` this conditional can be simplified to `if (player.lastCardPlayed === this.name)`.

src/cards/colonies/Conscription.ts
```diff
@@ -29,7 +29,7 @@ export class Conscription extends Card implements IProjectCard {
   }
 
   public getCardDiscount(player: Player) {
-    if (player.lastCardPlayed !== undefined && player.lastCardPlayed.name === this.name) {
+    if (player.lastCardPlayed !== undefined && player.lastCardPlayed === this.name) {
```
bafolts, line 32: I would use the `if (player.lastCardPlayed === this.name)` for all of these checks.

bafolts (approved):
This is great change. Will lead to less source and should also be slightly more efficient.

src/cards/base/SpecialDesign.ts
kberg: I fixed four of them.

PR #3679, opened 2021-10-03
store all bonuses for mining cards

### LINKED ISSUE #3669: Mining Area/Rights should not "remember" your previous choice
Mining Area and Mining Rights can only raise a production of a type of metal that is printed on the board underneath the respective tile.

Now that we have Arabia Terra, there is at least one area that gives both steel and titanium, so the player (for the first time) gets to choose which kind of production they want.

However, whichever one they don't pick gets "physically" removed from the card (!) such that if they Robotic Workforce it later on, it makes them take the same production type again.

BGG consensus is they should instead be able to choose differently the second time if they want to:
https://boardgamegeek.com/thread/2663453/rule-opinions-mining-rights-robotic-workforce

bafolts:
This fixes #3669 . Stores resources which were available under the placed tile for a Mining Rights card. When two resources are available allows player to select which resource to gain. Upon playing robotic workforce player can again select the resource.

I tested this with robotic workforce and the following logs demonstrates first selecting steel and then titanium with robotic workforce.

🕑You played MINING RIGHTS
🕑You's steel production increased by 1
🕑You placed Mining Rights tile on row 3 position 7
🕑You's steel amount increased by 1
🕑You's titanium amount increased by 1
🕑You played ROBOTIC WORKFORCE
🕑You copied MINING RIGHTS production with ROBOTIC WORKFORCE
🕑You's titanium production increased by 1

### DIFF
--- a/src/cards/IProjectCard.ts
+++ b/src/cards/IProjectCard.ts
@@ -11,7 +11,7 @@ export interface IProjectCard extends ICard {
     // bonus besides the obvious ones printed on the card. Mining Rights and Mining Area are the only
     // two that use this field at the time (though don't expect this comment to be kept up to date if
     // that changes.)
-    bonusResource?: Resources | undefined;
+    bonusResource?: Array<Resources.STEEL | Resources.TITANIUM>;
 
     // Represents resources held in reserve when paying for a card.
     // Cards that require a unit of steel while playing, for instance.
--- a/src/cards/base/MiningCard.ts
+++ b/src/cards/base/MiningCard.ts
@@ -11,6 +11,9 @@ import {SelectSpace} from '../../inputs/SelectSpace';
 import {SpaceBonus} from '../../SpaceBonus';
 import {Tags} from '../../cards/Tags';
 import {TileType} from '../../TileType';
+import {DeferredAction} from '../../deferredActions/DeferredAction';
+import {SelectOption} from '../../inputs/SelectOption';
+import {OrOptions} from '../../inputs/OrOptions';
 
 export abstract class MiningCard extends Card implements IProjectCard {
   constructor(
@@ -25,7 +28,7 @@ export abstract class MiningCard extends Card implements IProjectCard {
       metadata,
     });
   }
-    public bonusResource: Resources | undefined = undefined;
+    public bonusResource?: Array<Resources.STEEL | Resources.TITANIUM>;
     public canPlay(player: Player): boolean {
       return this.getAvailableSpaces(player).length > 0;
     }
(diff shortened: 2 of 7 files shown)


src/cards/IProjectCard.ts
```diff
@@ -11,7 +11,7 @@ export interface IProjectCard extends ICard {
     // bonus besides the obvious ones printed on the card. Mining Rights and Mining Area are the only
     // two that use this field at the time (though don't expect this comment to be kept up to date if
     // that changes.)
-    bonusResource?: Resources | undefined;
+    bonusResource?: Array<Resources.STEEL | Resources.TITANIUM>;
```
kberg, line 14: Now that I've seen this three times in two separate files, consider `type BonusResource`? Maybe not.

src/cards/base/MiningCard.ts
```diff
@@ -25,7 +28,7 @@ export abstract class MiningCard extends Card implements IProjectCard {
       metadata,
     });
   }
-    public bonusResource: Resources | undefined = undefined;
+    public bonusResource?: Array<Resources.STEEL | Resources.TITANIUM>;
```
kberg, line 31: Four times in three files! Yep, please do.

Actually why bother only including the alloys?

src/cards/base/MiningCard.ts
```diff
@@ -63,22 +66,54 @@ export abstract class MiningCard extends Card implements IProjectCard {
       return TileType.MINING_AREA;
     }
 
-    public produce(player: Player) {
+    public produce(player: Player, cb: (resource: Resources.STEEL | Resources.TITANIUM) => void = () => {}) {
```
kberg, line 69: This violates OOP: ICard's definition for produce is `produce?: (player: Player) => void;`

src/client/components/card/CardExtraContent.vue
```diff
@@ -29,9 +29,9 @@ export default Vue.extend({
       const miningCard = [CardName.MINING_RIGHTS, CardName.MINING_AREA, CardName.MINING_RIGHTS_ARES, CardName.MINING_AREA_ARES];
       if (miningCard.includes(card.name)) {
         if (metal === Resources.TITANIUM) {
```
kberg, line 31: `metal` is type string. Can it be type Resources?

src/client/components/card/CardExtraContent.vue
```diff
@@ -29,9 +29,9 @@ export default Vue.extend({
       const miningCard = [CardName.MINING_RIGHTS, CardName.MINING_AREA, CardName.MINING_RIGHTS_ARES, CardName.MINING_AREA_ARES];
       if (miningCard.includes(card.name)) {
```
kberg, line 30: I don't think this check is necessary. This entire method can be simplified to `return card.bonusResource?.includes(metal) === true`

tests/cards/base/MiningRights.spec.ts
```diff
@@ -51,4 +53,25 @@ describe('MiningRights', () => {
     expect(player.getProduction(Resources.TITANIUM)).to.eq(1);
     expect(steelSpace!.adjacency?.bonus).eq(undefined);
   });
+
+  it('Should play when steel and titanium', () => {
```
kberg, line 57:
```suggestion
  it('Should play when space bonus is both steel and titanium', () => {
```

kberg (commented):
No major issues.

src/cards/base/MiningCard.ts
bafolts: On this mining card this array will only have either `Resources.STEEL | Resources.TITANIUM`. Making a separate type for it makes sense but I would name it `MiningCardBonusResource`.

src/cards/base/MiningCard.ts
bafolts: Since a default is provided for the `cb` argument this function can be called with only the `player` argument as defined in ICard's definition. This is overloading. If it were JAVA we would list out both implementations. I can write out both implementations here and avoid using the `cb` argument provided in-line if that helps with following along.

src/cards/base/MiningCard.ts
kberg: OK

src/cards/base/MiningCard.ts
kberg: Yeah, I'm a little concerned: .produce() is fairly meant, upon completion, to produce the resulting resources. This will still be confusing and unclear, and I'd like it to make a little more sense first. I'll have to stare at this some, and then I can recommend either a comment or structural change.

kberg: What do you think of this? https://github.com/kberg/terraforming-mars/commit/30521aff9d678f5d37d0ff61194e8206683405df

src/cards/base/MiningCard.ts
kberg: So, it turns out I'm going to need nearly the exact same thing for a Pathfinders card, regardless of the resource type. 
image

bafolts: Scope creep! It may make sense to move `bonusResource` down to `ICard`. With this card in mind I can have `bonusResource` allow for any `Resources` in the array and use `Array<Resources>`.

kberg: OK.

src/cards/base/MiningCard.ts
kberg: See my comment in my proposed commit, just add a comment here. Thanks for your patience.

kberg (approved):

bafolts:
I am going to make `bonusResource?: Array<Resources>` in preparation for Specialised Settlement. Will make sure that the callback added to `produce` is in a private function as well.

bafolts:
Will have to fix the build tomorrow. The change to use `Resources` in `CardExtraContent` will require using the enum and not the string value in the template.

PR #3680, opened 2021-10-03
use brotli compression for styles

bafolts:
This saves 10kb over `gzip` when loading CSS on browsers which support `br` compression.


kberg:
All checks have failed - care to look first?

kberg (approved):

kberg:
This is nice.

PR #3681, opened 2021-10-04
Move common code for filtering tiles next to no other tile.

kberg:
What do you think -- is this a good idea?


kberg:
Oh, I'll add: I'm doing this because there's a third card in Pathfinders that does this.

PR #3687, opened 2021-10-06
Make canPlay mandatory for IProjectCard.

kberg:


src/cards/community/ScienceTagCard.ts
```diff
@@ -17,6 +17,9 @@ export class ScienceTagCard implements IProjectCard {
   public get cardType() {
     return CardType.PROXY;
   }
+  public canPlay() {
+    return false;
```
bafolts, line 21: This addition alone makes this change great.

PR #3694, opened 2021-10-08
Introduce trSource used for tracking reds costs.

kberg:
This infrastructure will work for playing cards, and hopefully the `TRSource` abstraction will help with actions on cards, too.

This only tracks ocean costs, and provides a simple example with Artificial Lake.

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1796,7 +1796,7 @@ export class Player implements ISerializable<SerializedPlayer> {
 
   public canPlay(card: IProjectCard): boolean {
     const baseCost = this.getCardCost(card);
-    const redsCost = this.computeTRBump(card) * REDS_RULING_POLICY_COST;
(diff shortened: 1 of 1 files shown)


src/Player.ts
```diff
@@ -1795,15 +1795,19 @@ export class Player implements ISerializable<SerializedPlayer> {
   }
 
   public canPlay(card: IProjectCard): boolean {
+    const baseCost = this.getCardCost(card);
+    const redsCost = this.computeTRBump(card) * REDS_RULING_POLICY_COST;
```
bafolts, line 1799: We have been spelling out `TerraformRating` other places it is used within `Player`. I would name this function `computeTerraformRatingBump` to stick with that pattern.

kberg: OK, Note that there's TRSource, with a field called tr. I'm OK any way we go, hoping not to have to change those, but I just wanted to mention it.

nwai90 (approved):

PR #3696, opened 2021-10-08
Add milestones and awards for Pathfinders

kberg:
These will not appear in the random milestones and awards.


kberg:
I mean, they should eventually appear in the random milestones and awards, but not when players do not want fan expansion content.

src/awards/Coordinator.ts
```diff
@@ -0,0 +1,11 @@
+import {IAward} from './IAward';
+import {Player} from '../Player';
+import {CardType} from '../cards/CardType';
+
+export class Coordinator implements IAward {
+    public name: string = 'Coordinator';
+    public description: string = 'Having the most event cards in play'
+    public getScore(player: Player): number {
+      return player.playedCards.filter((card) => card.cardType === CardType.EVENT).length;
```
bafolts, line 9: There is a `player.getCardsByCardType` which is only being used in one project card. We could leverage it here as well.

PR #3700, opened 2021-10-08
PF13: Add Pathfinders card Secret Labs.

kberg:


src/cards/pathfinders/SecretLabs.ts
```diff
@@ -0,0 +1,63 @@
...
+          'OR Raise temperature 1 step. Gain 3 plants. ' +
+          'OR Raise oxygen level 1 step. Add 2 floaters on any card.',
+        victoryPoints: 1,
+      },
+    });
+  }
+
+  public play(player: Player) {
+    return new OrOptions(
+      new SelectOption('Place an ocean tile. Add 2 microbes on any card.', 'select', () => {
```
bafolts, line 42: With the `tr` property for red policy will this card have `ocean`, `temperature`, and `oxygen` set? This will raise the terraform rating by one.

kberg: Right. This is a tricky one. tr: doesn't really work for all the cases, and this is one where we might need something manual, but not nearly as manual as what we currently have.

PR #3704, opened 2021-10-09
Add Pathfinders card Private Security.

kberg:


kberg:
I suspect this is still incomplete. I bet there's even some room to make this more centralized

kberg:
Thanks. This is going to need more work, I fear.

>

PR #3712, opened 2021-10-12
Reds tax for Venus & Update canAfford to be more strict about the red reserve

kberg:


kberg:
This also includes a bunch of card changes while I'm at it.

PR #3716, opened 2021-10-12
Move the Cyanobacteria code to AddResourcesToCards.

kberg:
There's a Pathfinders global event that does the same thing. So might as well make it a DeferredAction.

### DIFF
--- a/src/cards/pathfinders/Cyanobacteria.ts
+++ b/src/cards/pathfinders/Cyanobacteria.ts
@@ -6,8 +6,7 @@ import {CardName} from '../../CardName';
 import {CardRenderer} from '../render/CardRenderer';
 import {Tags} from '../Tags';
 import {ResourceType} from '../../ResourceType';
-import {SelectAmount} from '../../inputs/SelectAmount';
-import {AndOptions} from '../../inputs/AndOptions';
+import {AddResourcesToCards} from '../../deferredActions/AddResourcesToCards';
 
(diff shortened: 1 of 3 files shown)


tests/cards/pathfinders/Cyanobacteria.spec.ts
```diff
@@ -40,41 +40,48 @@ describe('Cyanobacteria', function() {
...
-
-    options?.options[0].cb(1);
-    options?.options[1].cb(3);
-    options?.options[2].cb(5);
-    options?.cb();
 
-    expect(ghgProducingBacteria.resourceCount).eq(1);
-    expect(tardigrades.resourceCount).eq(3);
-    expect(ants.resourceCount).eq(5);
+    if (options instanceof AndOptions) {
```
bafolts, line 55: If `options` is not an `AndOptions` should this test fail? If this is being done for typescript I would instead assign to a local.

```
const andOptions = options as AndOptions;
```

Once typescript thinks `andOptions` is `undefined` the `?` operator no longer needs to be used as well.

bafolts (approved):

tests/cards/pathfinders/Cyanobacteria.spec.ts
kberg: It will fail because of the `expect` statement above, but the compiler doesn't know about it.

PR #3717, opened 2021-10-13
Convert a bunch of cards that place oceans to use the tr attribute.

kberg:


bafolts (approved):
All the red lines in this diff are fitting for this change for reds policy.

PR #3719, opened 2021-10-14
remove unused function Game.prototype.getCard

bafolts:
While scrolling through the `Game` class today I saw a function that I would optimize. I then saw that nothing was using the function. This removes the function.


kberg (approved):
wooo!

PR #3734, opened 2021-10-19
Use turmoil.hasAvailableDelegates, clean up canPlay, fix TempestConsultancy.

kberg:
Also, beef up Turmoil tests a little.


kberg:
@nwai90 I think you know this stuff best. Do you think the change I added to sendDelegateToParty is going to introduce other problems?

PR #3743, opened 2021-10-22
Auto-detect getVictoryPoints.

kberg:
OK, it's time for some feedback.

This changes how victory point definitions are stored, calculated, and rendered.

The major change is the introduction of a new metadata field, `victoryPoints`. It can be a number or a structured object definining how its resources or a player's tags compute that result. When `victoryPoints` is defined, the card does not need `renderData.victoryPoints` or `getVictoryPoints.` They will be aligned.

To make this happen I had to change the rule for the `getVictoryPoints` method - it is no longer optional. In those cases where it used to be optional, `getVictoryPoints` returns 0. For this change to work, resourceCount also became a required field. I'm not concerned about memory consumption in this instance.

For those cards that generate victory points, but using a special case (e.g., `Commercial District`) I have not solved that yet. You will need to include the `getVictoryPoints` method and the `metadata.victoryPoints` to describe how to draw it, though. I will also need a mechanism for making those cards Vitor-compatible, too, but I think it's not going to be terribly difficult.

It took a fair bit of work to get this to be compatible with `Vitor`, so that's reassuring. There's another test that currently needs some change; I'll get to that as long as this seems like a good net win.

Finally, I think with a change like this we can remove a majority of the tests for getVictoryPoints, since most of the time it was ensuring I called `Math.floor` #eyeroll


src/cards/Card.ts
```diff
@@ -25,8 +27,40 @@ export interface StaticCardProperties {
   cardDiscount?: CardDiscount;
   reserveUnits?: Units,
   tr?: TRSource,
+  victoryPoints?: number | 'special' | VictoryPoints,
 }
 
+function inferVictoryPoints(properties: StaticCardProperties) {
```
bafolts, line 33: I would make this a private function of the `Card` class.

src/cards/Card.ts
```diff
@@ -92,7 +130,73 @@ export abstract class Card {
...
+    case CardRenderItemType.ANIMALS:
+    case CardRenderItemType.FIGHTER:
+    case CardRenderItemType.FLOATERS:
+    case CardRenderItemType.ASTEROIDS:
+    case CardRenderItemType.PRESERVATION:
+    case CardRenderItemType.DATA_RESOURCE:
+    case CardRenderItemType.RESOURCE_CUBE:
+    case CardRenderItemType.SCIENCE:
+    case CardRenderItemType.CAMPS:
+      units = (this as unknown as IResourceCard).resourceCount ?? 0;
```
bafolts, line 186: This casting is unnecessary. This can be written as:

```
units = this.resourceCount;
```

src/cards/Card.ts
```diff
@@ -92,7 +130,73 @@ export abstract class Card {
...
+      break;
+
+    case 'jovian':
```
bafolts, line 189: This is `CardRenderItemType.JOVIAN`

src/cards/Card.ts
```diff
@@ -92,7 +130,73 @@ export abstract class Card {
...
+      units = player?.getTagCount(Tags.JOVIAN, true, false);
+      break;
+    case 'moon':
```
bafolts, line 192: This is `CardRenderItemType.MOON`

src/cards/ICard.ts
```diff
@@ -28,14 +28,28 @@ export interface IActionCard {
...
+    resourceCount: number;
     resourceType?: ResourceType;
 }
 
 export interface CardDiscount {
   tag?: Tags, // When absent, discount applies to all cards.
   amount: number
  }
+export interface VictoryPoints {
+    type: 'resource' | Tags,
```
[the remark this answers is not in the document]
bafolts: As long as we are checking the _value_ of `type` and not the type I suppose it isn't reflection. Want to avoid using `typeof` and checking type at runtime. Any place where `typeof` is used typescript can't help us! Checks at compile time typescript can help us out.

src/cards/base/PhysicsComplex.ts
```diff
@@ -27,16 +29,12 @@ export class PhysicsComplex extends Card implements IActionCard, IProjectCard, I
           }).br;
           b.vpText('2 VP for each science resource on this card.');
         }),
-        victoryPoints: CardRenderDynamicVictoryPoints.science(2, 2),
```
bafolts, line 30: Was this wrong before? It used to be 2 and 2 but now is 2 and 1.

src/cards/prelude/Vitor.ts
```diff
@@ -60,9 +60,19 @@ export class Vitor extends Card implements CorporationCard {
...
 
   public onCardPlayed(player: Player, card: IProjectCard) {
-    if (player.isCorporation(this.name) && card.getVictoryPoints !== undefined && card.getVictoryPoints(player) >= 0) {
-      player.megaCredits += 3;
+    if (!player.isCorporation(this.name)) {
+      return;
     }
+    const victoryPoints = card.metadata.victoryPoints;
+    if (victoryPoints === undefined) return;
+    if (typeof(victoryPoints) === 'number') {
```
bafolts, line 68: Having to use reflection here checking `typeof(victoryPoints)` isn't ideal. If the RHS we had a type and had say `nmber` instead of `number` typescript will be none the wiser. I don't think there is a linting rule but we will need to be diligent to ensure we don't have typos in these `typeof` checks.

bafolts: The `isFinite` function and `Number.isInteger` could help avoid needing to use `typeof` with a string comparison.

src/cards/render/CardRenderDynamicVictoryPoints.ts
```diff
@@ -10,17 +12,45 @@ export class CardRenderDynamicVictoryPoints {
...
-  public static microbes(points: number, target: number): CardRenderDynamicVictoryPoints {
-    return new CardRenderDynamicVictoryPoints(new CardRenderItem(CardRenderItemType.MICROBES), points, target);
-  }
-  public static animals(points: number, target: number): CardRenderDynamicVictoryPoints {
-    return new CardRenderDynamicVictoryPoints(new CardRenderItem(CardRenderItemType.ANIMALS), points, target);
+
+  // Remove all of the builder methods; they're no longer necessary. Keep some
+  // version of the tests, though.
+  public static resource(type: ResourceType, points: number, target: number): CardRenderDynamicVictoryPoints {
+    const map: any = new Map([
```
bafolts, line 25: This `any` is not necessary. After this line typescript is assuming the variable map is of `any` type! :warning: :skull_and_crossbones:

src/cards/render/CardRenderDynamicVictoryPoints.ts
```diff
@@ -10,17 +12,45 @@ export class CardRenderDynamicVictoryPoints {
...
+      [ResourceType.FLOATER, CardRenderItemType.FLOATERS],
+    ]);
+    const itemType = map.get(type);
+    if (itemType === undefined) {
+      throw new Error('Unknown item type ' + type);
+    }
+    return new CardRenderDynamicVictoryPoints(new CardRenderItem(itemType), points, target);
+  }
+  public static tag(type: Tags, points: number, target: number): CardRenderDynamicVictoryPoints {
+    const map: any = new Map([
```
bafolts, line 44: This any is not necessary. After this line typescript is assuming the variable map is of any type! :warning: :ghost:

bafolts (approved):
All of my suggestions are minor. I would remove the two usages of `any` in `CardRenderDynamicVictoryPoints`. Assuming those were introduced to scare me on Halloween.

src/cards/base/PhysicsComplex.ts
kberg: Not exactly? It wasn't wrong, but "2 points per 2 science" was a little strange, but it only had to do with toHTML. But now the same code is used for rendering and also point calculation, and 2 points per 2 resource doesn't make sense.

src/cards/prelude/Vitor.ts
kberg: Your evaluation is good, but it's hard to understand its worth when we have no data on what actually consumes the most CPU time. The practical cost of this typeof is that it's called once every card is played, _if_ Vitor is in play.

I'm know you know I'm all in for knowing the right way to do things. But can we get some profiling data?

kberg: image

Looks like the compiler is wiser! :D

kberg: image

bafolts: Typescript is an amazing tool. :shipit:

PR #3758, opened 2021-10-30
Implement clone tag behavior, and two cards.

kberg:
I separated out code fir serializing cards so it was easier to test. (There's really very little in the way of serialization tests. :D)


bafolts (approved):
This is an elegant way of introducing these clone tags. I like every part of this change.

PR #3762, opened 2021-11-03
PF27: Five more Pathfinders cards

kberg:


src/cards/pathfinders/NobelLabs.ts
```diff
@@ -0,0 +1,51 @@
...
+          b.action('Add 2 microbes OR 2 data OR 2 floaters to ANY card.', (eb) => {
+            eb.empty().startAction.microbes(2, {digit}).slash().data({amount: 2, digit}).slash().floaters(2, {digit});
+          });
+        }),
+        description: 'Requires 4 Science tags.',
+      },
+    });
+  }
+
+  private static RESOURCE_TYPES: Array<ResourceType | undefined> = [ResourceType.MICROBE, ResourceType.DATA, ResourceType.FLOATER];
```
bafolts, line 35: This array doesn't have `undefined` as an item. It's type should be `Array<ResourceType>`. I believe `undefined` is being included so that when the call to `NobelLabs.RESOURCE_TYPES.includes(card.resourceType)` is called typescript doesn't error. I would avoid going this route and instead check that `card.resourceType !== undefined` before calling the `includes` function.

src/cards/pathfinders/VeneraBase.ts
```diff
@@ -0,0 +1,58 @@
...
+          b.br;
+          b.vpText('1 VP per 2 Venus tags you have.');
+        }),
+        description: 'Requires Unity is ruling or that you have 2 delegates there. Raise your MC production 3 steps and place a city tile ON THE RESERVED AREA.',
+      },
+    });
+  }
+
+  public canAct(player: Player) {
+    return player.getResourceCards(ResourceType.FLOATER)
```
bafolts, line 43: Since the count isn't used this is probably another great place to use `some`.

src/cards/pathfinders/NobelLabs.ts
kberg: That's correct. I'll make the change, but I think that my implementation is slightly more elegant.

src/cards/pathfinders/VeneraBase.ts
kberg: Done.

kberg:
Thanks for the thorough review. I keep forgetting about `some`. I have to remember it.

PR #3764, opened 2021-11-04
PF29: Three more project cards

kberg:
This ends the first pass of all the Pathfinders project cards. I'm sure some are missing (Red City comes to mind) but this concludes a massive amount of what needed to get done. Corps, preludes,  global events are probably next.


src/cards/CardRequirement.ts
```diff
@@ -255,7 +255,7 @@ export class ProductionCardRequirement extends CardRequirement {
 }
 
 export class PartyCardRequirement extends CardRequirement {
-  constructor(private party: PartyName) {
+  constructor(public party: PartyName) {
```
bafolts, line 258: Since the visibility of this property is changing from `private` to `public` we could make this `public readonly` to indicate that even though it is `public` we don't expect it to be modified.

src/cards/CardRequirements.ts
```diff
@@ -34,8 +34,12 @@ export class CardRequirements {
   public hasAny(): boolean {
     return this.requirements.some((req) => req.isAny);
   }
-  public hasParty(): boolean {
-    return this.requirements.some((req) => req instanceof PartyCardRequirement);
+  public hasParty(partyName?: PartyName | undefined): boolean {
+    return this.requirements.some((req) => {
+      if (!(req instanceof PartyCardRequirement)) return false;
+      if (partyName === undefined) return true;
```
bafolts, line 40: This seems strange to me that if `partyName` is `undefined` and if there is at least one item in `this.requirements` that is an instanceof PartyCardRequirement this function returns `true`. Why does this return true and not false?

src/cards/pathfinders/Pollinators.ts
```diff
@@ -0,0 +1,52 @@
...
+      cardType: CardType.ACTIVE,
+      name: CardName.POLLINATORS,
+      cost: 19,
+      tags: [Tags.PLANT, Tags.ANIMAL],
+      resourceType: ResourceType.ANIMAL,
+      requirements: CardRequirements.builder((b) => b.tag(Tags.PLANT, 3)),
+      victoryPoints: VictoryPoints.resource(1, 1),
+
+      metadata: {
+        cardNumber: '...',
```
bafolts, line 25: This card name seems strange.

src/cards/CardRequirements.ts
bafolts: Ah I see now. We use this function client side as well. This may be worth creating a separate function for `hasAnyParty` instead of using the `partyName` argument in this way. My concern with the argument is that a caller could pass in an optional property by mistake and would end up with an unexpected result.

bafolts (approved):

src/cards/CardRequirement.ts
kberg: > Since the visibility of this property is changing from `private` to `public` we could make this `public readonly` to indicate that even though it is `public` we don't expect it to be modified.

Done.

src/cards/CardRequirements.ts
kberg: I agree with you, but also I would prefer to try to fully separate client side from server side more explicitly. I've talked about about this in other example PRs which we should look to revite. So for now, no action if you don't mind.

src/cards/pathfinders/Pollinators.ts
kberg: Card number. Yes. After things settle I want to number the unnumbered cards, but the PDF made this a little tricky to do. So I'd like to do it all at once near the end.

PR #3769, opened 2021-11-05
Prototype for purging games later than they finish.

kberg:
This is the cleanest way I can see doing this given the current database structure. It allows games to stick around long enough for me (or anyone) to do post-game forensics.


kberg:
(I've been wanting this for quite some time.)

kberg:
No feedback after 4 months, I can close this and come back to it later.

PR #3770, opened 2021-11-07
only use name when serializing IAward for Game

bafolts:
This will use less space per entry by storing only the award name instead of the entire award object. There are plans to eventually keep more records in the database and this will help use less space.

### DIFF
--- a/src/SerializedGame.ts
+++ b/src/SerializedGame.ts
@@ -21,7 +21,7 @@ import {SerializedPathfindersData} from './pathfinders/SerializedPathfindersData
 export interface SerializedGame {
     activePlayer: PlayerId;
     aresData?: IAresData;
-    awards: Array<IAward>;
+    awards: Array<IAward> | Array<string>;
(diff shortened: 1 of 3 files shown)


kberg:
This is good. Does a test verify this is backwards compatible?

kberg:
Is that worthwhile?

kberg (approved):

src/SerializedGame.ts
```diff
@@ -21,7 +21,7 @@ import {SerializedPathfindersData} from './pathfinders/SerializedPathfindersData
 export interface SerializedGame {
     activePlayer: PlayerId;
     aresData?: IAresData;
-    awards: Array<IAward>;
+    awards: Array<IAward> | Array<string>;
```
kberg, line 24: Can you add a todo indicating by when to remove the full object definition?

bafolts: Added unit test and todo.

PR #3774, opened 2021-11-09
Simplify reds calculation for standard projects.

kberg:


src/cards/base/standardActions/ConvertPlants.ts
```diff
@@ -34,10 +33,10 @@ export class ConvertPlants extends StandardActionCard {
     if (player.game.getOxygenLevel() === MAX_OXYGEN_LEVEL) {
       return true;
     }
-    if (PartyHooks.shouldApplyPolicy(player, PartyName.REDS)) {
-      return player.canAfford(REDS_RULING_POLICY_COST);
-    }
-    return true;
+    return player.canAfford(0, {
+      tr: {oxygen: 1},
+      reserveUnits: Units.of({plants: player.plantsNeededForGreenery}),
```
bafolts, line 38: Can the check on line 27 that `player.plants < player.plantsNeededForGreenery` be removed with this addition?

bafolts (approved):

src/cards/base/standardActions/ConvertPlants.ts
kberg: It can. But I also kind of am not sure that I like removing it -- it seems elegant, it seems clever, but it's confusing if you don't understand. So I'm of two minds.

kberg:
I've added a bunch of tests.

PR #3778, opened 2021-11-11
Upgrade to Typescript 4.4.4

kberg:
There are two breaking changes:
1. catch blocks have moved from defaulting from type any to unknown. The database and PlayerInput changes reflect that.
2. Some DOM methods do not exist in the interface, and somehow were compiling, and now are not. I moved them to a single location, and tested manually using a Firefox browser.

### DIFF
--- a/src/client/components/HTMLDialogElementCompatibility.ts
+++ b/src/client/components/HTMLDialogElementCompatibility.ts
@@ -1,0 +1,16 @@
+// Some bits of the HTMLDialogElement api are not generally available or functional with most browsers
+// So this just encapsulated all of the uses in one place.
+//
+
+export function windowHasHTMLDialogElement(): boolean {
+  return (window as any).HTMLDialogElement !== undefined;
+}
+
+export function hasShowModal(dialogElement: HTMLElement) {
+  const de = dialogElement as any;
+  return de.showModal !== undefined && typeof(de.showModal) === 'function';
+}
+
+export function showModal(dialogElement: HTMLElement) {
+  (dialogElement as any).showModal();
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,7 +26,7 @@
         "pg": "^8.2.1",
         "spectre.css": "^0.5.8",
         "sqlite3": "^4.2.0",
-        "typescript": "^4.1.2",
+        "typescript": "^4.4.4",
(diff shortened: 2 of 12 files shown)


src/client/components/HTMLDialogElementCompatibility.ts
```diff
@@ -0,0 +1,16 @@
+// Some bits of the HTMLDialogElement api are not generally available or functional with most browsers
+// So this just encapsulated all of the uses in one place.
+//
+
+export function windowHasHTMLDialogElement(): boolean {
+  return (window as any).HTMLDialogElement !== undefined;
```
bafolts, line 6: This can avoid using `any` by using `return HTMLDialogElement !== undefined;`

src/client/components/HTMLDialogElementCompatibility.ts
```diff
@@ -0,0 +1,16 @@
...
+}
+
+export function hasShowModal(dialogElement: HTMLElement) {
```
bafolts, line 9:
```suggestion
export function hasShowModal(dialogElement: HTMLElement): dialogElement is HTMLDialogElement {
  return dialogElement instanceof HTMLDialogElement;
}
```

src/client/components/HTMLDialogElementCompatibility.ts
```diff
@@ -0,0 +1,16 @@
...
+  const de = dialogElement as any;
+  return de.showModal !== undefined && typeof(de.showModal) === 'function';
+}
+
+export function showModal(dialogElement: HTMLElement) {
```
bafolts, line 14:
```suggestion
export function showModal(dialogElement: HTMLDialogElement) {
  dialogElement.showModal();
}
```

src/client/components/HTMLDialogElementCompatibility.ts
kberg: You might be right but I am copying existing behavior, and I am a little nervous to go that route. Follow-up?

bafolts: I can do in follow-up. These `any` better not get comfortable.

bafolts (approved):

kberg:
I'm sorry, I undid your suggestions because I could not get it to compile.

PR #3779, opened 2021-11-11
remove a usage of any

bafolts:
I throw enough shade at every usage of `any`. This removes one usage of `any`.


kberg:
Thank you

PR #3784, opened 2021-11-15
Rewrote getTagCount to take a more clear parameter.

kberg:
Now it is based on the condition you want to achieve. Mostly I'm doing
this to simplify Chimera, a new corp card; I'm not sure if this is a
good direction, and could use the opinion.

### DIFF
--- a/src/CardName.ts
+++ b/src/CardName.ts
@@ -725,7 +725,7 @@ export enum CardName {
     PLANET_PR = 'planet pr',
     AMBIENT = 'Ambient',
     RINGCOM = 'Ringcom',
-    CHIIMERA = 'Chiimera',
+    CHIMERA = 'Chimera',
     SISTEMAS_SEEBECK = 'Sistemas Seebeck',
     SPIRE = 'Spire',
     SOYLENT_SEEDLING_SYSTEMS = 'Soylent Seedling Systems',
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -767,24 +767,40 @@ export class Player implements ISerializable<SerializedPlayer> {
 
   public getAllTags(): Array<ITagCount> {
     return [
-      {tag: Tags.BUILDING, count: this.getTagCount(Tags.BUILDING, false, false)},
-      {tag: Tags.CITY, count: this.getTagCount(Tags.CITY, false, false)},
-      {tag: Tags.EARTH, count: this.getTagCount(Tags.EARTH, false, false)},
-      {tag: Tags.ENERGY, count: this.getTagCount(Tags.ENERGY, false, false)},
-      {tag: Tags.JOVIAN, count: this.getTagCount(Tags.JOVIAN, false, false)},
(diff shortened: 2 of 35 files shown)


src/Player.ts
```diff
@@ -767,24 +767,40 @@ export class Player implements ISerializable<SerializedPlayer> {
...
+   */
+  public getTagCount(tag: Tags, mode: 'default' | 'raw' | 'award' | 'vps' = 'default') {
+    switch (mode) {
+    case 'default': return this.getTagCountOld(tag, false, true);
+    case 'raw': return this.getTagCountOld(tag, false, false);
+    case 'award': return this.getTagCountOld(tag, false, false); // TODO(include Chimera)
+    case 'vps': return this.getTagCountOld(tag, true, false);
+    }
+  }
+  public getTagCountOld(tag: Tags, includeEventsTags:boolean = false, includeTagSubstitutions:boolean = true): number {
```
kberg, line 803: I need a new name for this. Or, figure out a way to merge it into getTagCount without making it impossible to read.

src/Player.ts
```diff
@@ -767,24 +767,40 @@ export class Player implements ISerializable<SerializedPlayer> {
...
+   * Get the number of tags a player has, depending on certain conditions.
+   *
+   * 'raw': count face-up tags literally.
+   * 'default': Same as raw, but include the events.
+   * 'award': Same as raw but includes special conditions for awards (Chimera)
+   * 'vps': Same as raw, but include event tags.
+   */
+  public getTagCount(tag: Tags, mode: 'default' | 'raw' | 'award' | 'vps' = 'default') {
+    switch (mode) {
+    case 'default': return this.getTagCountOld(tag, false, true);
```
kberg, line 797: Also, rename default to 'project'?

src/CardName.ts
```diff
@@ -725,7 +725,7 @@ export enum CardName {
     PLANET_PR = 'planet pr',
     AMBIENT = 'Ambient',
     RINGCOM = 'Ringcom',
-    CHIIMERA = 'Chiimera',
+    CHIMERA = 'Chimera',
```
bafolts, line 728: This will need to read the string `Chiimera` from database and map to `Chimera` for some time as we have done before. Any games with `Chiimera` saved will fail to find the card.

kberg: It's unused, fortunately. :D

src/Player.ts
bafolts: This naming seems good enough to me. Could also consider naming `raw` something similar to `baseline` as the three other types seem to be `raw` plus something else. You always get raw no matter the mode.

src/Player.ts
bafolts: This function should be `protected` and not `public`.

bafolts: Instead of naming this function `getTagCountOld` the new function could be named `getTagCountByMode`.

bafolts (approved):

PR #3785, opened 2021-11-15
Add sanity test for Sqlite and add purge table.

kberg:
(but it doesn't work, gah)

### DIFF
--- a/tests/database/SQLite.spec.ts
+++ b/tests/database/SQLite.spec.ts
@@ -1,0 +1,105 @@
+import {expect} from 'chai';
+import {Game, GameId} from '../../src/Game';
+import {TestPlayers} from '../TestPlayers';
+import {IN_MEMORY_SQLITE_PATH, SQLite} from '../../src/database/SQLite';
+import {Database} from '../../src/database/Database';
+import {restoreTestDatabase} from '../utils/setup';
+
+class TestSQLite extends SQLite {
+  public saveGamePromise: Promise<void> = Promise.resolve();
+
+  constructor() {
+    super(IN_MEMORY_SQLITE_PATH, true);
+  }
+
+  public get database() {
+    return this.db;
+  }
+
+  public saveGame(game: Game): Promise<void> {
+    this.saveGamePromise = super.saveGame(game);
+    return this.saveGamePromise;
+  };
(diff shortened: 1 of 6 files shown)


bafolts (approved):

kberg:
@bafolts take another look, please. I added another test. Am not going to change cleanSaves now. But I am testing the code I added, and how about that? It didn't actually work! :D

tests/database/SQLite.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+  constructor() {
+    super(IN_MEMORY_SQLITE_PATH, true);
+  }
+
+  public get database() {
+    return this.db;
+  }
+
+  public async saveGame(game: Game) {
+    this.saveGamePromise = super.saveGame(game);
```
bafolts, line 20: YES! As I wrote my version in javascript I realized it could be written this way! Awesome.

kberg: Look at the failed test, though.

kberg:
This is a surprising failed test which will make things flaky. Not sure what to do about this.

```
  1) SQLite
       saveIds:

      AssertionError: expected [ 0, 2, 3 ] to have the same members as [ 0, 1, 2, 3 ]
      + expected - actual

       [
         0
      +  1
         2
         3
       ]
      
      at /home/runner/work/terraforming-mars/terraforming-mars/tests/database/SQLite.spec.ts:84:28
      at step (tests/database/SQLite.spec.ts:48:23)
      at Object.next (tests/database/SQLite.spec.ts:29:53)
      at fulfilled (tests/database/SQLite.spec.ts:20:58)
```

tests/database/SQLite.spec.ts
kberg: Looks much better. Thank you.

bafolts (approved):
This is good change. We need to keep bubbling this change upward and having functions return a `Promise` that doesn't resolve until we have created this save entry. Right now a lot of functions return and leave a `Promise` that we have no pointer to that resolves at some point later.

kberg:
Thanks for making it work! I fully expect it to start failing. :( flaky tests.

PR #3788, opened 2021-11-17
remove more usages of any

bafolts:
Removes more usages of `any` from the code.

### DIFF
--- a/src/client/components/PreferencesDialog.vue
+++ b/src/client/components/PreferencesDialog.vue
@@ -59,7 +59,7 @@ export default Vue.extend({
       }
       return PreferencesManager.preferencesValues;
     },
-    updatePreferences(_evt: any): void {
+    updatePreferences(): void {
       let strVal: string = '';
       for (const k of preferences) {
         const val = PreferencesManager.preferencesValues.get(k);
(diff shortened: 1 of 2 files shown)


src/client/directives/TrimWhitespace.ts
```diff
@@ -1,8 +1,10 @@
 
-export function trimEmptyTextNodes(el: any) {
-  for (const node of el.childNodes) {
-    if (node.nodeType === Node.TEXT_NODE && node.data.trim() === '') {
+export function trimEmptyTextNodes(el: Node) {
+  for (let i = 0; i < el.childNodes.length; i++) {
```
bafolts, line 3: Can't use `for...of` on HTML collection.

kberg: > Can't use `for...of` on HTML collection.
And yet it somewhow worked?

src/client/components/PreferencesDialog.vue
```diff
@@ -59,7 +59,7 @@ export default Vue.extend({
       }
       return PreferencesManager.preferencesValues;
     },
-    updatePreferences(_evt: any): void {
```
kberg, line 62: any `any` is one `any` too many.

kberg (approved):

src/client/directives/TrimWhitespace.ts
bafolts: The generated javascript here may have gotten lucky that it was using the `.length` property which exists on an HTML collection. What that `for...of` would get wrong though is that an HTML Collection is a _live list_. As items are removed from the DOM it is updated in real time. As we were calling to `remove` we were most likely skipping items.

PR #3789, opened 2021-11-18
remove another any and add LoadGameFormModel

bafolts:
Removes another usage of `any` and adds a model for the load game form.


kberg (approved):
Nice. Thank you!

PR #3790, opened 2021-11-18
Prototype postgres test.

kberg:
@bafolts, I'm trying to do the same thing for Postgres, and you can see some of the mess I'm working with. Things aren't quite working as I expected yet, one thing at a time.

Just wondering if you have any advice for what you see so far.


kberg:
I actually don't think the postgres tests can be part of the standard test suite, and should be omitted. But we should still be able to run those tests manually from time to time. One thing is for sure: I'm not going to start risking Postgres changes without tests.

bafolts:
Running them as the standard test suite will require at least the build machine to have a postgres database it could use for the tests. The github action machines seem like they may be able to install that for us. Then we could at least run them during the build. Using a separate test command would make sense as not everyone would be able to run these locally.

kberg:
> Running them as the standard test suite will require at least the build machine to have a postgres database it could use for the tests. The github action machines seem like they may be able to install that for us. Then we could at least run them during the build. Using a separate test command would make sense as not everyone would be able to run these locally.

That might not be necessary. I'm thinking that perhaps this can be turned into an integration test only run locally. Then I can just write some dumb shell script for set-up and tear-down.

src/database/PostgreSQL.ts
```diff
@@ -8,42 +8,25 @@ import {Pool, ClientConfig, QueryResult} from 'pg';
...
-    this.client.query('CREATE INDEX IF NOT EXISTS games_i2 on games(created_time )', (err) => {
-      if (err) {
-        throw err;
-      }
-    });
   }
 
   async initialize(): Promise<void> {
-
+    this.runLoudly('initialize1', 'CREATE TABLE IF NOT EXISTS games(game_id varchar, players integer, save_id integer, game text, status text default \'running\', created_time timestamp default now(), PRIMARY KEY (game_id, save_id))')
```
bafolts, line 26: With `async` and `await` I would write this as:

```
await this.runLoudly()
await this.runLoudly()
await this.runLoudly()
await this.runLoadly()
```

src/database/PostgreSQL.ts
```diff
@@ -152,44 +135,25 @@ export class PostgreSQL implements IDatabase {
...
-      // Flag game as finished
-      this.client.query('UPDATE games SET status = \'finished\' WHERE game_id = $1', [game_id], (err2) => {
-        if (err2) {
-          console.error('PostgreSQL:cleanSaves2', err2);
-          throw err2;
-        }
-      });
-    });
-    this.purgeUnfinishedGames();
+    const cleanGame = this.runQuietly('cleanSaves', 'DELETE FROM games WHERE game_id = $1 AND save_id < $2 AND save_id > 0', [game_id, save_id]);
```
bafolts, line 143: I would write this as:

```
await this.runQuietly('cleanSaves'...);
await this.runQuietly('cleanSaves2'...);
```

This way the `cleanSaves2` query wont start until `cleanSaves` has completed. With how this is written `cleanSaves` will start and then `cleanSaves2` will start.

bafolts:
I haven't ran the postgres tests locally but I am noticing a pattern which may be source of trouble. Within any `async` function I would always use `await`. There is a difference between returning a `Promise` and using `await`. With the `await` keyboard the line after will not run until the promise has resolved. With returning a `Promise` the promise will be returned even if it hasn't resolved.

Really I did a faux pas when I used:

```
someFunction: Promise<void> {
   return Promise.resolve();
}
```

In this day and age of javascript it is usually better for everyone to _hide_ the usage of promises with async and await. I would rewrite even the stub functions I put in as:

```
async someFunction: Promise<void> {
    // this doesn't need a body, typescripts got us!
    // avoids having to write and return Promise
    // which is always good
}
```

I would steer clear of using `.then` and always use `await`. With `await` if a promise rejects this also gets thrown and can be caught with the `try...catch` syntax. With `.then` if we miss putting in a `.catch` it would only end up as an uncaught promise rejection which may error at window level but not locally.

jholm117:
Not sure if this work has been abandoned but I'd recommend adding a docker compose file to spin up a postgres container, apply any migrations you may have, then run your separate integration test suite. It will be quite simple to run this locally and in CI with just a `docker compose run` command. And since it is docker it will be nice and portable.

I'd be happy to help set that up if you wanted, but I'm not super familiar with the testing framework and would need someone to separate the test suites.

kberg:
I guess it's been abandoned (by yours truly.) Having such a test would be
so, so good.

kberg:
> I guess it's been abandoned (by yours truly.) Having such a test would be so, so good.

In fact, I just had to do emergency surgery on this thing because I didn't have such a test. >:-(

kberg:
> Not sure if this work has been abandoned but I'd recommend adding a docker compose file to spin up a postgres container, apply any migrations you may have, then run your separate integration test suite. It will be quite simple to run this locally and in CI with just a `docker compose run` command. And since it is docker it will be nice and portable.
> 
> I'd be happy to help set that up if you wanted, but I'm not super familiar with the testing framework and would need someone to separate the test suites.

John thanks for the offer. I have finally picked this back up and successfully wrote an integration test. It requires manual setup and can't be run as part of the github suite of tests, but if a docker container can do it, then yes, that would be a vast improvement.

In the meantime, this PR is going away in favor of a working test.

kberg:
> 

Brian, I don't think avoiding then all the time is good, maybe in some circumstances. But in the code itself, yes, nested thens make sense, because they can be run without blocking activity, which is kind of the point. And then await with a catch at the end is kinda nice.

Just means you have to catch when a `then` doesn't have `catch`, then.

PR #3796, opened 2021-11-21
Escape velocity

kberg:


kberg:
I'm going to do some manual testing over the next couple of days, and also a thorough code review. (Though, amusingly, someone else will have to approve it.)

src/Game.ts
```diff
@@ -121,6 +121,10 @@ export interface GameOptions {
   requiresVenusTrackCompletion: boolean; // Venus must be completed to end the game
   moonStandardProjectVariant: boolean;
   altVenusBoard: boolean;
+  escapeVelocityMode: boolean;
```
kberg, line 124: Recommend turning the whole expansion into a single object. (I wish I had done the same with some of tne Moon / Ares things. /shrug) That could also just be changed over time.

src/Player.ts
```diff
@@ -2275,6 +2300,7 @@ export class Player implements ISerializable<SerializedPlayer> {
     const player = new Player(d.name, d.color, d.beginner, Number(d.handicap), d.id);
     const cardFinder = new CardFinder();
 
+    player.actionsTakenThisGame = d.actionsTakenThisGame;
```
kberg, line 2303: @nwai90 this can be undefined for old games, and will break them. d.actionsTakenThisGame ?? 0?

src/VictoryPointsBreakdown.ts
```diff
@@ -1,13 +1,14 @@
 import {Tags} from './cards/Tags';
 
-export type VictoryPoints = 'terraformRating' | 'milestones' | 'awards' | 'greenery' | 'city' | 'moon colony' | 'moon mine' | 'moon road' | 'planetary tracks' | 'victoryPoints';
+export type VictoryPoints = 'terraformRating' | 'milestones' | 'awards' | 'greenery' | 'city' | 'escapeVelocity penalty' | 'moon colony' | 'moon mine' | 'moon road' | 'planetary tracks' | 'victoryPoints';
```
kberg, line 3: I think remove 'penalty'. It's already a negative value.

src/client/components/TagCount.vue
```diff
@@ -38,14 +38,14 @@ export default Vue.extend({
   methods: {
     getClasses(): string {
       const classes = ['tag-display'];
-      if (this.count === 0) {
+      if (this.count === 0 && this.tag !== 'escape') {
```
kberg, line 41: So why are escape tags a thing?

src/client/components/overview/PlayerTags.vue
```diff
@@ -250,6 +253,12 @@ export default Vue.extend({
       }
       return multipliers;
     },
+    isEscapeVelocityOn: function(): boolean {
```
kberg, line 256: This can be in `computed` which I hear makes things faster. Next time.

kberg (commented):
In short this looks good @nwai90.

src/Player.ts
```diff
@@ -581,6 +584,24 @@ export class Player implements ISerializable<SerializedPlayer> {
...
+    // Escape velocity VP penalty
+    if (this.game.gameOptions.escapeVelocityMode) {
+      const threshold = this.game.gameOptions.escapeVelocityThreshold;
+      const period = this.game.gameOptions.escapeVelocityPeriod;
+      const penaltyPerMin = this.game.gameOptions.escapeVelocityPenalty ?? 1;
+      const elapsedTimeInMinutes = this.timer.getElapsedTimeInMinutes();
+      if (threshold !== undefined && period !== undefined && elapsedTimeInMinutes > threshold) {
+        const overTimeInMinutes = Math.max(elapsedTimeInMinutes - threshold - (this.actionsTakenThisGame * (constants.BONUS_SECONDS_PER_ACTION/60)), 0);
+        // Don't lose more VP that what is available
+        const victoryPoints = this.getVictoryPoints();
```
kberg, line 596: Hmm! I wonder if this creates an infinite loop! What do you think, @nwai90?

nwai90: @kberg Will take a look!

kberg:
@nwai90  I have playtested this in a solo game. My feedback:

1. It works.
2. The EV icon in the final game breakdown should match the one in the player bar.
3. Something, somewhere, should describe the point loss calculation. "Why am I at -12, specifically?"

kberg:
> @nwai90 I have playtested this in a solo game. My feedback:
> 
> 1. It works.
> 2. The EV icon in the final game breakdown should match the one in the player bar.
> 3. Something, somewhere, should describe the point loss calculation. "Why am I at -12, specifically?"

So if you want to approve this, it can go in. If you want to change #2 or #3 before going anywhere else, LMK.

src/Player.ts
```diff
@@ -581,6 +584,23 @@ export class Player implements ISerializable<SerializedPlayer> {
...
 
+    // Escape velocity VP penalty
+    if (this.game.gameOptions.escapeVelocityMode) {
+      const threshold = this.game.gameOptions.escapeVelocityThreshold;
+      const period = this.game.gameOptions.escapeVelocityPeriod;
+      const penaltyPerMin = this.game.gameOptions.escapeVelocityPenalty ?? 1;
+      const elapsedTimeInMinutes = this.timer.getElapsedTimeInMinutes();
+      if (threshold !== undefined && period !== undefined && elapsedTimeInMinutes > threshold) {
+        const overTimeInMinutes = Math.max(elapsedTimeInMinutes - threshold - (this.actionsTakenThisGame * (constants.BONUS_SECONDS_PER_ACTION/60)), 0);
+        // Don't lose more VP that what is available
```
bafolts, line 595:
```suggestion
        // Don't lose more VP than what is available
```

bafolts: I didn't realize that escape velocity was an expansion. If escape velocity is an expansion then following the `EscapeVelocityExpansion.calculateVictoryPoints(this, victorPointsBreakdown)` would follow the pattern of other expansions.

src/Player.ts
```diff
@@ -2020,13 +2040,15 @@ export class Player implements ISerializable<SerializedPlayer> {
       );
       this.setWaitingFor(initialActionOrPass, () => {
         this.actionsTakenThisRound++;
+        this.actionsTakenThisGame++;
```
bafolts, line 2043: Making a private `incrementActionsTaken` function would avoid having to update both `actionsTakenThisRound` and `actionsTakenThisGame` in both places.

src/Player.ts
```diff
@@ -2275,6 +2299,7 @@ export class Player implements ISerializable<SerializedPlayer> {
     const player = new Player(d.name, d.color, d.beginner, Number(d.handicap), d.id);
     const cardFinder = new CardFinder();
 
+    player.actionsTakenThisGame = d.actionsTakenThisGame ?? 0;
```
bafolts, line 2302: This `?? 0` can be removed after some period of time. Is this only needed for while the `actionsTakenThisGame` property is being introduced. I would put one of the `TODO` comments above so we know we can come back and remove this `?? 0` at some point in future.

src/Timer.ts
```diff
@@ -55,6 +55,11 @@ export class Timer implements ISerializable<SerializedTimer> {
     this.sumElapsed += Timer.lastStoppedAt - this.startedAt;
   }
 
+  public getElapsedTimeInMinutes(): number {
+    const elapsedTimeInMin = (this.sumElapsed + (this.running ? Date.now() - this.startedAt : 0)) / (60 * 1000);
```
bafolts, line 59: Making a function for `getElapsed` would avoid duplicating some of this function's logic inside `toString`. The conversion to minutes at that point could be done by the caller. Inside `toString` we could call `d.getElapsed()` and re-use the logic.

bafolts (approved):

src/Player.ts
kberg: Agree.

src/Game.ts
```diff
@@ -121,6 +121,10 @@ export interface GameOptions {
...
+  escapeVelocityThreshold?: number;
```
kberg, line 125: What happens when someone creates a game with escapeVelocityMode = true, and no other parameters?

PR #3799, opened 2021-11-21
Make default timeout twice as fast.

kberg:


bafolts (approved):
This is going to all but double traffic to popular servers. I think they can handle it as this request is most likely going to return to wait but may be worth keeping an eye on. CPU usage runs pretty low.

PR #3803, opened 2021-11-23
Add proper logging for stealing-type cards

erisalke:
Issue: https://github.com/terraforming-mars/terraforming-mars/issues/3175
Screenshot 2021-11-29 at 19 24 18
Screenshot 2021-11-29 at 19 24 27
Screenshot 2021-11-29 at 19 29 40

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -318,9 +318,9 @@ export class Player implements ISerializable<SerializedPlayer> {
   }
 
   private logUnitDelta(
-    resource: Resources, 
-    amount: number, 
-    unitType: 'production' | 'amount', 
+    resource: Resources,
+    amount: number,
+    unitType: 'production' | 'amount',
     from: Player | GlobalEventName | undefined,
     stealing = false
   ) {
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -317,36 +317,15 @@ export class Player implements ISerializable<SerializedPlayer> {
     }
   }
 
-  private logStealing(
-    resource: Resources, 
-    amount: number, 
-    unitType: 'production' | 'amount', 
-    from: Player
-  ) {
-    const absAmount = Math.abs(amount);
-    let message = '${0} stole ${1} ${2} ' + unitType + ' from ${3}';
-    this.game.log(message, (b) => {
-      b.player(from)
-      .number(absAmount)
-      .string(resource)
-      .player(this)
-    });
-  }
-
   private logUnitDelta(
     resource: Resources, 
     amount: number, 
     unitType: 'production' | 'amount', 
     from: Player | GlobalEventName | undefined,
-    isStealing = false
+    stealing = false
   ) {
     if (amount === 0) {
       // Logging zero units doesn't seem to happen
-      return;
-    }
-
-    if (isStealing && from instanceof Player) {
-      this.logStealing(resource, amount, unitType, from);
       return;
     }
 
@@ -354,6 +333,13 @@ export class Player implements ISerializable<SerializedPlayer> {
     const absAmount = Math.abs(amount);
     // TODO(kberg): remove the ${2} for increased and decreased, I bet it's not used.
     let message = '${0}\'s ${1} ' + unitType + ' ${2} by ${3}';
+
+    if (from !== undefined) {
+      if (stealing === true) {
+        message = message + ' stolen';
+      }
+      message = message + ' by ' + ((from instanceof Player) ? '${4}' : 'Global Event');
+    }
 
     if (from !== undefined) {
       message = message + ' by ' + ((from instanceof Player) ? '${4}' : 'Global Event');
@@ -375,7 +361,7 @@ export class Player implements ISerializable<SerializedPlayer> {
     options? : {
       log?: boolean,
       from? : Player | GlobalEventName,
-      isStealing?: boolean
+      stealing?: boolean
     }) {
     this.addResource(resource, -amount, options);
   }
@@ -386,7 +372,7 @@ export class Player implements ISerializable<SerializedPlayer> {
     options? : {
       log?: boolean,
       from? : Player | GlobalEventName,
-      isStealing?: boolean
+      stealing?: boolean
     }) {
     // When amount is negative, sometimes the amount being asked to be removed is more than the player has.
     // delta represents an adjusted amount which basically declares that a player cannot lose more resources
@@ -424,7 +410,7 @@ export class Player implements ISerializable<SerializedPlayer> {
     }
 
     if (options?.log === true) {
-      this.logUnitDelta(resource, delta, 'amount', options.from, options.isStealing);
+      this.logUnitDelta(resource, delta, 'amount', options.from, options.stealing);
     }
 
     if (options?.from instanceof Player) {
@@ -441,7 +427,7 @@ export class Player implements ISerializable<SerializedPlayer> {
   public addProduction(
     resource: Resources,
     amount : number,
-    options? : { log: boolean, from? : Player | GlobalEventName, isStealing?: boolean}
+    options? : { log: boolean, from? : Player | GlobalEventName, stealing?: boolean}
   ) {
     const adj = resource === Resources.MEGACREDITS ? -5 : 0;
     const delta = (amount >= 0) ? amount : Math.max(amount, -(this.getProduction(resource) - adj));
@@ -457,7 +443,7 @@ export class Player implements ISerializable<SerializedPlayer> {
     }
 
     if (options?.log === true) {
-      this.logUnitDelta(resource, amount, 'production', options.from, options.isStealing);
+      this.logUnitDelta(resource, amount, 'production', options.from, options.stealing);
     }
 
     if (options?.from instanceof Player) {
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -322,7 +322,7 @@ export class Player implements ISerializable<SerializedPlayer> {
     amount: number,
     unitType: 'production' | 'amount',
     from: Player | GlobalEventName | undefined,
-    stealing = false
+    stealing = false,
   ) {
     if (amount === 0) {
       // Logging zero units doesn't seem to happen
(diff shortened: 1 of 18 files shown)


src/Player.ts
```diff
@@ -348,6 +375,7 @@ export class Player implements ISerializable<SerializedPlayer> {
     options? : {
       log?: boolean,
       from? : Player | GlobalEventName,
+      isStealing?: boolean
```
kberg, line 378: I think shorten this to `stealing`

src/Player.ts
```diff
@@ -317,12 +317,39 @@ export class Player implements ISerializable<SerializedPlayer> {
...
+      .player(this)
+    });
+  }
+
+  private logUnitDelta(
+    resource: Resources, 
+    amount: number, 
+    unitType: 'production' | 'amount', 
+    from: Player | GlobalEventName | undefined,
+    isStealing = false
```
kberg, line 341: short here to `stealing` too.

src/Player.ts
```diff
@@ -317,12 +317,39 @@ export class Player implements ISerializable<SerializedPlayer> {
...
+  private logStealing(
+    resource: Resources, 
+    amount: number, 
+    unitType: 'production' | 'amount', 
+    from: Player
+  ) {
+    const absAmount = Math.abs(amount);
+    let message = '${0} stole ${1} ${2} ' + unitType + ' from ${3}';
+    this.game.log(message, (b) => {
+      b.player(from)
```
kberg, line 329: While I agree that your grammar is better than what logUnitDelta has, I expect the name order will bew confusing as players scan the log. This could be made simpler by just reusing logUnitDelta.

Change 
```
if (from !== undefined) {
  message = message + ' by ' + ((from instanceof Player) ? '${4}' : 'Global Event');
}
```

to

```
if (from !== undefined) {
  if (stealing === true) {
    message = message + ' stolen';
  }
  message = message + ' by ' + ((from instanceof Player) ? '${4}' : 'Global Event');
}
```

src/Player.ts
```diff
@@ -409,7 +438,11 @@ export class Player implements ISerializable<SerializedPlayer> {
     }
   }
 
-  public addProduction(resource: Resources, amount : number, options? : { log: boolean, from? : Player | GlobalEventName}) {
+  public addProduction(
+    resource: Resources,
+    amount : number,
+    options? : { log: boolean, from? : Player | GlobalEventName, isStealing?: boolean}
```
kberg, line 444: Shorten `isStealing` here, too.

src/cards/base/AsteroidMiningConsortium.ts
```diff
@@ -34,7 +34,9 @@ export class AsteroidMiningConsortium extends Card implements IProjectCard {
   }
 
   public play(player: Player) {
-    player.game.defer(new DecreaseAnyProduction(player, Resources.TITANIUM, 1));
+    const decreaseAction = new DecreaseAnyProduction(player, Resources.TITANIUM, 1);
```
kberg, line 37: So if this were me I'd move much of the parameters to an object, just like `GainResources` The object would contain count (default to 1), title (defaulted to its current value) and stealing (default to false.) That'll be much simpler.

However, I'd probably do that in a completely separate PR or commit. If you like I'll do that.

kberg (changes requested):
Thanks for thinking about this! Let me know if you want me to knock out generalizing DecreaseProduction etc.

src/Player.ts
erisalke: Keep the same message format, sure thing.

src/cards/base/AsteroidMiningConsortium.ts
erisalke: I extracted the `Options` as follows:
```
export interface Options {
  count: number,
  stealing?: boolean
}
```
Therefore, all the previous calls changed 
from: `player.game.defer(new DecreaseAnyProduction(player, Resources.TITANIUM, 1));`
to: `player.game.defer(new DecreaseAnyProduction(player, Resources.TITANIUM, {count: 1}));`
and the stealing calls look this way:
`player.game.defer(new DecreaseAnyProduction(player, Resources.TITANIUM, {count: 1, stealing: true}));`

One thing I did't include into the `Options` was the title. For the moment all the existing calls use the default value - so it makes more sense to remove the parameter completely instead and hide it inside the class itself. And composing the "options" where the title is optional, but have a default value, and such value depends on the "count" which as well has a default value and may be optional... it's just messy.
I believe it's "good enough" for now. Maybe in future PRs if you really believe it's necessary.

src/cards/base/EnergyTapping.ts
```diff
@@ -33,7 +33,8 @@ export class EnergyTapping extends Card implements IProjectCard {
 
   public play(player: Player) {
     player.addProduction(Resources.ENERGY, 1);
-    player.game.defer(new DecreaseAnyProduction(player, Resources.ENERGY, 1));
+    player.game.defer(
+      new DecreaseAnyProduction(player, Resources.ENERGY, {count:1, stealing:true}));
```
kberg, line 37: So one thing about reformatting on the fly is that it makes reviews more complicated.

image

Now I have to read the whole thing, whereas if you had just kept it on one line I could just the latter half of the line.

Don't make any changes, because I'll have read all of them by the time you get this.

src/cards/base/AsteroidMiningConsortium.ts
```diff
@@ -34,7 +34,12 @@ export class AsteroidMiningConsortium extends Card implements IProjectCard {
   }
 
   public play(player: Player) {
-    player.game.defer(new DecreaseAnyProduction(player, Resources.TITANIUM, 1));
+    const decreaseAction = new DecreaseAnyProduction(
+      player,
+      Resources.TITANIUM,
+      {count: 1, stealing: true}
```
kberg, line 40: Is this "stealing"? It looks like it, but is it?

src/deferredActions/DecreaseAnyProduction.ts
```diff
@@ -3,24 +3,33 @@ import {Resources} from '../Resources';
...
-        public count: number = 1,
-        public title: string = 'Select player to decrease ' + resource + ' production by ' + count + ' step(s)',
-  ) {}
+    public player: Player,
+    public resource: Resources,
+    public options: DecreaseAnyProduction.Options = {
+      count: 1,
+      stealing: false
+    },
+    public title: string = 'Select player to decrease ' + resource + ' production by ' + options.count + ' step(s)',
```
kberg, line 21: I agree with your assessment of title. I think I'd still put it in options, as options tends to be the last parameter. But don't bother, this is great.

src/deferredActions/DecreaseAnyProduction.ts
```diff
@@ -41,7 +50,7 @@ export class DecreaseAnyProduction implements DeferredAction {
       this.title,
       'Decrease',
       (found: Player) => {
-        found.addProduction(this.resource, -this.count, {log: true, from: this.player});
+        found.addProduction(this.resource, -this.options.count, { log: true, from: this.player, stealing: this.options.stealing });
```
kberg, line 53: There's no `decreaseProduction` method, is there? That's fine.

src/cards/base/PowerSupplyConsortium.ts
```diff
@@ -34,7 +34,8 @@ export class PowerSupplyConsortium extends Card implements IProjectCard {
 
   public play(player: Player) {
     player.addProduction(Resources.ENERGY, 1);
-    player.game.defer(new DecreaseAnyProduction(player, Resources.ENERGY, 1));
+    player.game.defer(
+      new DecreaseAnyProduction(player, Resources.ENERGY, {count:1, stealing:true}));
```
kberg, line 38: Should there be a StealProduction deferred action? Not in this PR, just thinking.

kberg (commented):
This is a good change. Normally I'd want to see some tests, but we kind of avoid many of the logging tests. If you can manually reverify these, this will be good enough with no more changes. (Or if you want to write a unit test, I won't mind! :D)

Once this is verified, I'll merge it in. Thank you!

Thanks!

src/cards/base/AsteroidMiningConsortium.ts
kberg: As you said, I also believe this is "good enough." You saw some of my comments lamenting that the API that existed before you came in here could have been better, but those can be dealt with another time. Thank you for this!

src/cards/base/AsteroidMiningConsortium.ts
erisalke: Depends how you define "stealing" I believe if one player lost "X" and other player gains exactly the same "X" (in this case titanium production) it is best described as "stealing" :D

src/deferredActions/DecreaseAnyProduction.ts
erisalke: nope, there's only the `addProduction` used with negative values

kberg:
> This is a good change. Normally I'd want to see some tests, but we kind of avoid many of the logging tests. If you can manually reverify these, this will be good enough with no more changes. (Or if you want to write a unit test, I won't mind! :D)
> 
> Once this is verified, I'll merge it in. Thank you!
> 
> Thanks!

Thanks for the responses, etc. Just a reminder that my request above remains. Thanks again!

erisalke:
@kberg All the changes were tested and works fine. You can see the results in the screens attached on the top. I think we could merge it now.

kberg:
Those at the top are from before you made changes to the code, so it needs
another verification.

erisalke:
@kberg ah no, sorry for not being clear enough 😓 
When I said: "All the changes were tested and works fine", I meant that I've **just finished** re-testing this.
In other words, it was checked that all works fine AFTER the last commit was made. It's javascript, we take no risks :)

kberg (approved):

kberg:
AHA! Thank you. This looks good. Thank you!

PR #3815, opened 2021-11-28
Remove getHtmlContent when rendering corporation logos.

kberg:
This turns all the string literals for rendering the card corporation logo into template HTML.

image

Ignore the card without a logo in that screenshot, it's a Pathfinders corp that doesn't have a logo ... yet.

### DIFF
--- a/src/client/components/card/CardCorporationLogo.vue
+++ b/src/client/components/card/CardCorporationLogo.vue
@@ -1,5 +1,207 @@
 <template>
-  <div class="card-corporation-logo" v-html="getHtmlContent()"></div>
+  <div class="card-corporation-logo" v-if="title === CardName.APHRODITE">
+    <div class="card-aphrodite-logo">APHRODITE</div>
+  </div>
+  <div class="card-corporation-logo" v-else-if="title === CardName.ARKLIGHT">
+    <div class="card-arklight-logo">ARKLIGHT</div>
+  </div>
+  <div class="card-corporation-logo" v-else-if="title === CardName.POSEIDON">
+     <div class="card-poseidon-logo">POSEIDON</div>
+  </div>
+  <div class="card-corporation-logo" v-else-if="title === CardName.SATURN_SYSTEMS">
+    <div class="card-saturn-logo">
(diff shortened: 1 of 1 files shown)


src/client/components/card/CardCorporationLogo.vue
```diff
@@ -15,162 +217,9 @@ export default Vue.extend({
...
-      case CardName.INTRAGEN_SANCTUARY_HEADQUARTERS:
-        return '<div class="card-intragen-sanctuary-headquarters-logo"></div>';
-      case CardName.THE_ARCHAIC_FOUNDATION_INSTITUTE:
-        return '<div class="card-the-archaic-foundation-institute-logo"></div>';
-      case CardName.CURIOSITY_II:
-        return '<div class="card-curiosity-ii-logo">Curiosity II</div>';
-      }
-      return '';
+  computed: {
+    CardName(): any {
```
bafolts, line 221:
```suggestion
    CardName(): typeof CardName {
```

src/client/components/card/CardCorporationLogo.vue
```diff
@@ -1,5 +1,207 @@
 <template>
-  <div class="card-corporation-logo" v-html="getHtmlContent()"></div>
+  <div class="card-corporation-logo" v-if="title === CardName.APHRODITE">
```
bafolts, line 2: I think you can have less duplication here if you use `<template/>`

```
<div class="card-corporation-logo">
    <template v-if="title === CardName.APHRODITE">
        <div class="card-aphrodite-logo">APHRODITE</div>
    </template>
</div>

src/client/components/card/CardCorporationLogo.vue
bafolts: This suggestion may have been missed. A type of `typeof CardName` can be used here instead of `any`.

bafolts (approved):

src/client/components/card/CardCorporationLogo.vue
kberg: Oh of course. Also, I love how this allows us to use enums inside vue, which we couldn't do before.

PR #3827, opened 2021-12-01
Prototype rendering icons for card requirements

kberg:
Depends on #3826 

This feature is toggled based on the "Experimental UI" flag in #3826, and so, should be merged first. But here's a taste of the iconography. You can see it has potential, and needs work. That's why it's under the experimental UI flag.

image
image
image


src/client/components/card/CardRequirementComponent.vue
```diff
@@ -0,0 +1,146 @@
+<template>
+  <div class="card-requirement">
+      <div class="card-item-container">
+        <span v-if="requirement.isMax">MAX&nbsp;</span>
+        {{amount()}}{{suffix()}}
+        <div :class="getComponentClasses()" />
```
bafolts, line 6: The starting and end tag are mandatory for a div element.

bafolts (approved):

PR #3829, opened 2021-12-01
Cultural bug

vincentneko:
Fix a bug that prevented card to be played if you had 2 delegates in reserve.


nwai90 (approved):
Approved, but could you remove the `package-lock.json` change from this commit? It does not seem related to the bugfix

vincentneko:
> Approved, but could you remove the `package-lock.json` change from this commit? It does not seem related to the bugfix

My bad, will remove this, thanks

vincentneko:
This one is botched, will create a clean new one

PR #3833, opened 2021-12-03
Pathfinders should play card when playing a project card.

kberg:


kberg:
Verified manually. /sheepish face

nwai90 (approved):

PR #3835, opened 2021-12-04
Use static card properties for Global Events.

kberg:
This will reduce memory consumption on a high-load server.


kberg:
This is all the cards at once, but the changes are repetitive and easy to spot. That said, feel free to ask me to reduce this by a percentage.

This probably requires a little more testing but it's not as if this will get reviewed immediately. :D

bafolts (approved):
I double checked all of these.

bafolts:
> This is all the cards at once, but the changes are repetitive and easy to spot. That said, feel free to ask me to reduce this by a percentage.
> 
> This probably requires a little more testing but it's not as if this will get reviewed immediately. :D

:baby: :baby_bottle: :baby_symbol: :baby_chick: :man_feeding_baby:

PR #3844, opened 2021-12-07
Add method canAddOcean, and further refine Artifical Lake.

kberg:
Artificial Lake should not be playable if no land spaces are available.


alrusdi (approved):

kberg:
Thanks, @alrusdi!

PR #3848, opened 2021-12-08
Update red city tiles

kberg:
This includes a pretty nice image for Red City that I'll use to update the card itself.


kberg:
In the meantime, this board tile is better than the last one. Could be made _slightly better_ but it's much better. 
image

alrusdi (approved):

PR #3850, opened 2021-12-09
Remove the neutral icon from the board.

kberg:
Before:
image

After:
image


kberg:
This breaks the Pathfinders card "Martian Nature Wonders" which I'll fix in a subsequent change.

kberg:
And ... Martian Nature Wonders now looks like this:

image

Restoring the border is a little bit of trickery that I'll leave for now. It's not common.

alrusdi (approved):

PR #3852, opened 2021-12-09
More reqs viz.

kberg:
1. Fix card party requriement viz.
2. Reorder where the mars tag lives.
image
image


kberg:
I accidentally pushed another minor req, and am taking a bit of a "aw heck with it" approach. Feel free to post-reject. :D

PR #3856, opened 2021-12-10
Add Pathfinders card Vital Colony

kberg:
I don't love the way this is going. Any advice?

(Specifically the way to grant the placement bonus twice.)

### DIFF
--- a/src/cards/pathfinders/VitalColony.ts
+++ b/src/cards/pathfinders/VitalColony.ts
@@ -27,7 +27,7 @@ export class VitalColony extends PreludeCard {
         false,
         undefined,
         undefined,
-        {vitalColonyHook: true}));
+        {giveBonusTwice: true}));
     return undefined;
   }
 }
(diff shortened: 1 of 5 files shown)


src/cards/pathfinders/VitalColony.ts
```diff
@@ -0,0 +1,34 @@
...
+    });
+  }
+  public play(player: Player) {
+    player.game.defer(
+      new BuildColony(
+        player,
+        false,
+        undefined,
+        undefined,
+        {vitalColonyHook: true}));
```
bafolts, line 30: If `vitalColonyHook` is too specific could consider naming this `giveBonus` instead.

bafolts (approved):

src/cards/pathfinders/VitalColony.ts
kberg: Yeah, that's good, thanks.

PR #3864, opened 2021-12-11
Add Pathfinders corp Mars Direct.

kberg:


kberg:
image

This is a simpler version of Mars Direct which basically does the same thing but is simpler to code.

image

After all, every step means progress.

bafolts (approved):

PR #3867, opened 2021-12-13
Use an options object to define how trading works.

kberg:
Facilitates some new Pathfinders cards.


kberg:
I don't have additional tests, but one version of this code failed several tests, so ...

nwai90 (approved):

PR #3870, opened 2021-12-14
Simplify SelectColony API

kberg:
Review this by ignoring whitespace.

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -8,7 +8,6 @@ import {CardType} from './cards/CardType';
 import {ClaimedMilestone, serializeClaimedMilestones, deserializeClaimedMilestones} from './milestones/ClaimedMilestone';
 import {Colony} from './colonies/Colony';
 import {ColonyDealer, loadColoniesFromJSON} from './colonies/ColonyDealer';
-import {ColonyModel} from './models/ColonyModel';
 import {ColonyName} from './colonies/ColonyName';
 import {Color} from './Color';
 import {CorporationCard} from './cards/corporation/CorporationCard';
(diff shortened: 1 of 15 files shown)


src/Game.ts
```diff
@@ -501,23 +500,6 @@ export class Game implements ISerializable<SerializedGame> {
     this.deferredActions.push(action);
   }
 
-  public getColoniesModel(colonies: Array<Colony>) : Array<ColonyModel> {
```
kberg, line 504: This was moved to ColonyModel.

src/Player.ts
```diff
@@ -1003,11 +1002,15 @@ export class Player implements ISerializable<SerializedPlayer> {
...
+      const colonyName: ColonyName = (input[0][0]) as ColonyName;
+      if (colonyName === undefined) {
         throw new Error('No colony selected');
       }
-      this.runInputCb(pi.cb(colony));
+      const colony = this.game.colonies.find((c) => c.name === colonyName);
+      if (colony === undefined) {
+        throw new Error(`Unknown colony '${colonyName}'`);
+      }
+      this.runInputCb(pi.cb(colony!));
```
bafolts, line 1013: The exclamation mark on this `colony` should be unnecessary as on line 1010 we check that it is defined.

src/deferredActions/RemoveColonyFromGame.ts
```diff
@@ -12,16 +11,10 @@ export class RemoveColonyFromGame implements DeferredAction {
...
-          game.colonies.splice(game.colonies.indexOf(colony), 1);
-          if (game.colonyDealer === undefined) return;
-          game.colonyDealer.discardedColonies.push(colony);
-        }
-        return undefined;
-      });
+    const removeColony = new SelectColony('Select colony tile to remove', 'Remove colony', game.colonies, (colony: Colony) => {
+      game.colonies.splice(game.colonies.indexOf(colony), 1);
+      if (game.colonyDealer === undefined) return;
+      game.colonyDealer.discardedColonies.push(colony);
```
bafolts, line 17: Unrelated to this PR but I think this could be changed to:

```
game.colonyDealer?.discardedColonies.push(colony);
```

Avoids having to check if `game.colonyDealer` is defined on line prior and returning early.

bafolts (approved):
This is great change. Less code and more efficient.

src/Player.ts
kberg: Good, thanks.

src/deferredActions/RemoveColonyFromGame.ts
kberg: Good change!

PR #3874, opened 2021-12-16
Minor fixes for Archaic Foundation Institute.

### LINKED ISSUE #3873: Clarify Moon tags on The Archaic Foundation Institute corp do not count for ability?
In almost every case, tags on corporations trigger that corp ability (e.g., Saturn Systems, Point Luna, Pharmacy Union). While The Archaic Foundation Institute does not have "including this" in its description since it departs from what normally happens in the game, it is probably worth calling this out explicitly (or changing the behavior).

kberg:
1. "include this".
2. Take reds costs into account.
Fixes #3873


nwai90 (approved):
Main change LGTM, would suggest removing the unrelated change to `package-lock.json` in this PR

kberg:
> Main change LGTM, would suggest removing the unrelated change to `package-lock.json` in this PR

Thanks, sorry.

PR #3881, opened 2021-12-18
Two colonies cards for Pathfinders

kberg:


src/cards/pathfinders/HuygensObservatory.ts
```diff
@@ -0,0 +1,117 @@
...
+    }
+    if (orOptions.options.length > 1) {
+      game.defer(new DeferredAction(player, () => orOptions));
+    }
+  }
+  public canPlay(player: Player): boolean {
+    return player.hasAvailableColonyTileToBuildOn(true) || this.tradeableColonies(player).length > 0;
+  }
+
+  public play(player: Player) {
```
kberg, line 100: This class is so complicated. I would like advice on making it simpler.

src/cards/pathfinders/PathfindersCardManifest.ts
```diff
@@ -150,14 +152,14 @@ export const PATHFINDERS_CARD_MANIFEST = new CardManifest({
...
+    {cardName: CardName.HUYGENS_OBSERVATORY, Factory: HuygensObservatory, compatibility: GameModule.Colonies},
     {cardName: CardName.CASSINI_STATION, Factory: CassiniStation, compatibility: GameModule.Colonies},
     {cardName: CardName.MICROBIOLOGY_PATENTS, Factory: MicrobiologyPatents},
-    // // {cardName: CardName.COORDINATED_RAID, Factory: CoordinatedRaid, compatibility: GameModule.Colonies},       // TRICKY WORK
+    {cardName: CardName.COORDINATED_RAID, Factory: CoordinatedRaid, compatibility: GameModule.Colonies},
     {cardName: CardName.LOBBY_HALLS, Factory: LobbyHalls, compatibility: GameModule.Turmoil},
     {cardName: CardName.RED_CITY, Factory: RedCity, compatibility: GameModule.Turmoil},
     {cardName: CardName.VENERA_BASE, Factory: VeneraBase, compatibility: [GameModule.Turmoil, GameModule.Venus]},
-    // // {cardName: CardName.GATEWAY_STATION, Factory: GatewayStation, compatibility: GameModule.Turmoil},
+    // // {cardName: CardName.GATEWAY_STATION, Factory: GatewayStation, compatibility: GameModule.Turmoil}, // Place a city tile outside Mars with new adjacency??
```
kberg, line 162: Just an extra comment

src/deferredActions/BuildColony.ts
```diff
@@ -31,7 +32,8 @@ export class BuildColony implements DeferredAction {
     const openColonies = this.openColonies;
 
     return new SelectColony(this.title, 'Build', openColonies, (colony: Colony) => {
-      colony.addColony(this.player, this.options);
+      colony.addColony(this.player, {giveBonusTwice: this.options?.giveBonusTwice ?? false});
```
kberg, line 35: Because options' definition changed, I had to extract giveBonusTwice out of here.

src/Player.ts
```diff
@@ -2434,16 +2434,20 @@ export class Player implements ISerializable<SerializedPlayer> {
...
+  public hasAvailableColonyTileToBuildOn(allowDuplicate: boolean = false): boolean {
     if (this.game.gameOptions.coloniesExtension === false) return false;
 
     const availableColonyTiles = this.game.colonies.filter((colony) => colony.isActive);
-    let colonyTilesAlreadyBuiltOn: number = 0;
+    let unavailableColonies: number = 0;
 
     availableColonyTiles.forEach((colony) => {
-      if (colony.colonies.includes(this.id)) colonyTilesAlreadyBuiltOn++;
+      if (colony.colonies.length === constants.MAX_COLONIES_PER_TILE) {
```
bafolts, line 2444: Is this changing the logic for when `allowDuplicate` is `false`? Prior to this change we wouldn't have incremented `colonyTilesAlreadyBuiltOn` if `colony.colonies.length === constants.MAX_COLONIES_PER_TILE`. If `allowDuplicate` is `true` wouldn't we increment `unavailableColonies` regardless of if `colony.colonies.includes(this.id)`?

src/Player.ts
```diff
@@ -2434,16 +2434,20 @@ export class Player implements ISerializable<SerializedPlayer> {
     if (this.fleetSize > 0) this.fleetSize--;
   }
 
-  public hasAvailableColonyTileToBuildOn(): boolean {
+  public hasAvailableColonyTileToBuildOn(allowDuplicate: boolean = false): boolean {
```
bafolts, line 2437: Often times to me a boolean flag like this added to a function can be an opportunity to break a function into smaller parts. Could we do that here?

kberg: Yeah, I probably could.

src/Player.ts
kberg: No, actually, the line I removed was altogether outdated.

src/Player.ts
kberg: Reading this, is it worth doing what you suggest? I don't see the right kind of split (except for perhaps restoring the original behavior and having as completely second one for the duplicate case.)

bafolts (approved):

PR #3884, opened 2021-12-19
Fix outlier where onCardPlayed should be deferred until clone tags are resolved.

kberg:
I'm not sure this is right, yet. It certainly isn't ideal. First try.


src/Player.ts
```diff
@@ -1569,9 +1569,18 @@ export class Player implements ISerializable<SerializedPlayer> {
       this.playedCards.push(selectedCard);
     }
 
+    // See DeclareCloneTag for why.
+    if (!selectedCard.tags.includes(Tags.CLONE)) {
```
kberg, line 1573: This might be even more clean if here, when `selectedCard` actually contains `Tags.CLONE` this method defers DelclareCloneTag. At this point, ICloneTag would include a new callback which would replace the current DeclareCloneTag callback.

src/cards/pathfinders/CrewTraining.ts
```diff
@@ -41,12 +40,7 @@ export class CrewTraining extends PreludeCard implements ICloneTagCard {
   public play(player: Player) {
     player.increaseTerraformRatingSteps(2);
 
-    player.game.defer(
-      new DeclareCloneTag(
-        player,
-        this,
-        'Select a planetary track to advance 2 steps (and to clone this tag.)',
-        (tag) => PathfindersExpansion.raiseTrack(tag, player, 2)));
```
kberg, line 49: This is now handled automatically when the card's value is assigned, and onCardPlayed is called later.

src/pathfinders/DeclareCloneTag.ts
```diff
@@ -29,6 +43,9 @@ export class DeclareCloneTag implements DeferredAction {
         this.card.cloneTag = tag;
         this.player.game.log('${0} turned the clone tag on ${1} into a ${2} tag',
           (b) => b.player(this.player).card(this.card).string(tag));
+        if ([CardType.AUTOMATED, CardType.ACTIVE, CardType.EVENT, CardType.PRELUDE].includes(this.card.cardType)) {
```
kberg, line 46: Should the same occur for corps? Planet PR could be a valid case (I'm currently writing it.)

kberg:
This is ready for review. Thanks!

src/deferredActions/DeferredAction.ts
```diff
@@ -2,6 +2,7 @@ import {Player} from '../Player';
 import {PlayerInput} from '../PlayerInput';
 
 export enum Priority {
+  DECLARE_CLONE_TAG, // Used for Pathfinders. First thing a player must do before further effects.
```
bafolts, line 5: Oh no do we use the number that typescript assigns to this enum for sorting? :see_no_evil: This is right with how we have this but this is accident waiting to happen as it relies on implementation details of `enum` from `typescript`.

src/pathfinders/DeclareCloneTag.ts
bafolts: I would use a `Set` for this versus `Array` and `includes`.

src/Player.ts
bafolts: Prior to this change we were not calling `onCardPlayed` regardless of what `selectedCard.tags` included. Why do we have to call it now for tags that do not have `Tags.CLONE`?

src/deferredActions/DeferredAction.ts
kberg: It would be a problem if those values were serialized or transferred between servers, but they are not. Once the server restarts, it'll run with a new set of priorities.

src/Player.ts
kberg: `onCardPlayed` is just "the rest of this method" pushed into its own method so it can be invoked either now, or later. I think if you take a closer look it'll make sense.

bafolts: Ah yes I do see it now! The diff made that a tricky one to spot.

bafolts (approved):

PR #3885, opened 2021-12-20
Restore ability to add new colonies to the game.

### LINKED ISSUE #3882: Aridor Starting Action refuses to add colonies not in the game
At the start of the game, playing the Aridor Corp should add another colony tile. The game displays unused colony tiles correctly, but then when I choose one and click "Add Colony Tile", I get the following error (also in attached screenshot): 
Unknown colony "Europa"

Screenshot from 2021-12-18 22-41-56

with Europa being a placeholder for any {colony}.

Eventually I modified the POST and sent a colony that's in the game already, and this added a second copy of the same tile to the game, see second screenshot.

Screenshot from 2021-12-18 22-45-50

I think this has to do with the SelectColony method in:
https://github.com/terraforming-mars/terraforming-mars/blob/2813f10e9cdc8d262e73adedef64fce916ba3aca/src/Player.ts#1014

where in 1014 we get a list of the colonies in this.game, while in 1015 we check if the selected name is in that list (which it should not be for the game to proceed correctly).

Anyway, bug seems game breaking. I can PR a fix if you want, but this is my first game with this project so I'm probably not the person you want for this. (Love the rest of it btw, thank you for dev.)

kberg:
Because the issue was in handling player input, the test uses player input.

Fixes #3882

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -72,6 +72,7 @@ import {CrashSiteCleanup} from './cards/promo/CrashSiteCleanup';
 import {Turmoil} from './turmoil/Turmoil';
 import {PathfindersExpansion} from './pathfinders/PathfindersExpansion';
 import {deserializeProjectCard, serializeProjectCard} from './cards/CardSerialization';
+import {ColoniesHandler} from './colonies/ColoniesHandler';
 
 export type PlayerId = string;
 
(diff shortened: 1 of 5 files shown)


nwai90 (approved):

kberg:
Did a quick manual test to verify.

PR #3887, opened 2021-12-21
Reduce reliance on `as` for testing.

kberg:
TestingUtils.cast() verifies the type being cast, and so throws errors when the type does not match.

Also relied a bit on getTestPlayer.


kberg:
This is only a subset of the calls.

bafolts (approved):

PR #3894, opened 2021-12-23
Split out colonies trading details into handlers.

kberg:
This simplifies the code, and makes it easier to split out
Titan Floating Launch Pad. Another card for Pathfinders will
also allow trading with its resources; this will help.

### DIFF
--- a/src/cards/colonies/TitanFloatingLaunchPad.ts
+++ b/src/cards/colonies/TitanFloatingLaunchPad.ts
@@ -103,8 +103,8 @@ export class TradeWithTitanFloatingLaunchPad implements IColonyTrader {
   }
 
   public trade(colony: Colony) {
-    // TODO(kberg): find a way to remove ! or just suck it up and use an if(!undefined)
-    this.titanFloatingLaunchPad!.resourceCount--;
+    // grr I wish there was a simpler syntax.
+    if (this.titanFloatingLaunchPad !== undefined) this.titanFloatingLaunchPad.resourceCount--;
     this.player.addActionThisGeneration(CardName.TITAN_FLOATING_LAUNCHPAD);
     this.player.game.log('${0} spent 1 floater to trade with ${1}', (b) => b.player(this.player).colony(colony));
     colony.trade(this.player);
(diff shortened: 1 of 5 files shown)


kberg:
Don't merge this yet -- I'm noticing some weird behavior with Miranda colony. Can't tell if it's related.

kberg:
> Don't merge this yet -- I'm noticing some weird behavior with Miranda colony. Can't tell if it's related.

No wait, the problem wasn't the app. The problem was me.

src/cards/colonies/TitanFloatingLaunchPad.ts
```diff
@@ -83,3 +84,29 @@ export class TitanFloatingLaunchPad extends Card implements IProjectCard, IResou
...
+      !this.player.getActionsThisGeneration().has(CardName.TITAN_FLOATING_LAUNCHPAD);
+  }
+
+  public optionText() {
+    return 'Pay 1 Floater (use Titan Floating Launch-pad action)';
+  }
+
+  public trade(colony: Colony) {
+    // TODO(kberg): find a way to remove ! or just suck it up and use an if(!undefined)
+    this.titanFloatingLaunchPad!.resourceCount--;
```
bafolts, line 107: I would put in this defense prior to merging. Will cause runtime exception otherwise, May be able to leverage the `?` operator in some way.

src/colonies/ColoniesHandler.ts
```diff
@@ -12,4 +23,136 @@ export class ColoniesHandler {
...
+      }
+    });
+
+    if (howToPayForTrade.options.length === 0) {
+      return undefined;
+    }
+
+    const selectColony = new SelectColony('Select colony tile for trade', 'trade', openColonies, (colony: Colony) => {
+      if (selected === undefined) {
+        throw new Error('???');
```
bafolts, line 67: Using a clearer error message will make this easier to read and diagnose.

bafolts (approved):

src/cards/colonies/TitanFloatingLaunchPad.ts
kberg: Yeah, done. Thanks.

src/colonies/ColoniesHandler.ts
kberg: Whoops!

PR #3896, opened 2021-12-24
Use type guard for action cards.

kberg:


bafolts (changes requested):
I would change `isIActionCard` to the following:

```
export function isIActionCard(card: ICard | IActionCard): card is IActionCard {
  return card.action !== undefined && card.canAct !== undefined;                                   
} 
```

This will ensure what we are passed is what we expect and not `any`. It also couples the check with `IActionCard`. If for example we were to have a typo and use `card.actio` instead of `card.action` typescript covers us.

```
src/cards/ICard.ts:31:15 - error TS2551: Property 'actio' does not exist on type 'IActionCard | ICard'. Did you mean 'action'?
  Property 'actio' does not exist on type 'IActionCard'.
  ```

kberg:
> I would change `isIActionCard` to the following:
> 
> ```
> export function isIActionCard(card: ICard | IActionCard): card is IActionCard {
>   return card.action !== undefined && card.canAct !== undefined;                                   
> } 
> ```
> 
> This will ensure what we are passed is what we expect and not `any`. It also couples the check with `IActionCard`. If for example we were to have a typo and use `card.actio` instead of `card.action` typescript covers us.
> 
> ```
> src/cards/ICard.ts:31:15 - error TS2551: Property 'actio' does not exist on type 'IActionCard | ICard'. Did you mean 'action'?
>   Property 'actio' does not exist on type 'IActionCard'.
> ```

Will do in a couple of hours. Right now I just want to ask: why do so many stackoverflow posts recommend the style I presented?

kberg:
> Will do in a couple of hours. Right now I just want to ask: why do so many stackoverflow posts recommend the style I presented?

Ignore my question.

bafolts (approved):

PR #3898, opened 2021-12-30
Add Pathfinders corp Collegium Corpus

kberg:


src/cards/pathfinders/CollegiumCopernicus.ts
```diff
@@ -0,0 +1,112 @@
...
+      !this.player.getActionsThisGeneration().has(CardName.COLLEGIUM_COPERNICUS);
+  }
+
+  public optionText() {
+    return 'Pay 3 Data (use Collegium Copernicus action)';
+  }
+
+  public trade(colony: Colony) {
+    // TODO(kberg): find a way to remove ! or just suck it up and use an if(!undefined)
+    this.collegiumCopernicus!.resourceCount -= tradeCost(this.player);
```
bafolts, line 107: I would put this in before merging. With obfuscated javascript this error could show in logs as `unable to read property resourceCount of undefined` and we may not know line number to know which class it came from.

bafolts (approved):
Discussed offline but I would low the image quality of `corp-logo-colleguim-copernicus.png`. It is currently 168kb which is unnecessary.

src/cards/pathfinders/CollegiumCopernicus.ts
kberg: Yeah. Done.

kberg:
> Discussed offline but I would lower the image quality of `corp-logo-colleguim-copernicus.png`. It is currently 168kb which is unnecessary.

Before: 172,328
After: 13,193

Thanks for calling me out on it.

Issue #3903, opened 2021-12-31
Simplify Celstic's mechanism for identifying relevant cards


kberg:
Celestic is a tricky card. Someone should go through the expansion cards to find out if any new ones apply. Also, that can be replaced with code (especially since HTML has been replaced with a code structure [unexpected benefit, @chosta!])

kberg:
Good enough! Closing.

PR #3904, opened 2022-01-01
Update known Celestic-compatible cards.

kberg:
This includes a nice test that will ensure future cards can be included.


kberg:
Fixes #3903

nwai90 (approved):

PR #3907, opened 2022-01-03
Simplify PaymentWidgetMixin and SelectHowToPay

kberg:
This will be necessary to simplify the upcoming Soylent _something something_ project card.

The SHTP algorithm for computing initial values is now much less complicated, and does the same thing. All the tests pass against both the old version and this newer one. I left the code in for now, though, with a TODO to get rid of in in a couple of months.


nwai90 (approved):

kberg:
Oh thanks!

PR #3908, opened 2022-01-04
Refactor create game form

immdevrov:
Hello!
As @kberg told me, a lot of repeated inputs in _CreateGameForm.vue_ needs to be encapsulated.

I've introduced a component called _CreateGameToggle.vue_, which do the trick – hiding input, and give two slots for additional icons. Also, I've extracted tooltip component, I think it will be good for reusing in the future.
 
And, had to add “es2015” lib to ts.config for better autocompletion.

I actually have a lot of other thoughts about how the form can be improved, but  I think this is a proper start


immdevrov:
Oh, looks like, npm automatically changed package-lock.json lockfileVersion to 1, pls tell me if it is a critical change – I'll reverte it

src/styles/debug_ui.less
```diff
@@ -4,7 +4,9 @@
         flex-flow: row wrap !important;
         padding: 20px 10px;
         .expansion-button {
+          & label {
```
kberg, line 7: What does this do for the create game change?

.gitignore
```diff
@@ -16,6 +16,7 @@ pid
 .eslintcache
 tsconfig.tsbuildinfo
 assets/locales
+.idea
```
kberg, line 19: Moved this up a little.

src/client/components/create/CreateGameForm.vue
```diff
@@ -1,440 +1,492 @@
...
+            <CreateGameToggle v-model="aresExtension" title="Ares" class="expansion-button">
+              <template #prepend>
+                <i class="create-game-expansion-icon expansion-icon-ares" />
+              </template>
+              <template #append>
+                <Tooltip link="https://github.com/bafolts/terraforming-mars/wiki/Ares" />
+              </template>
+            </CreateGameToggle>
+
+            <CreateGameToggle v-model="promoCardsOption" title="Promos" class="expansion-button">
```
kberg, line 103: Why is promocards here, and above?

src/client/components/create/CreateGameForm.vue
```diff
@@ -1,440 +1,492 @@
...
+                <i class="create-game-expansion-icon expansion-icon-promo" />
+              </template>
+              <template #append>
+                <Tooltip link="https://github.com/bafolts/terraforming-mars/wiki/Variants#promo-cards" />
+              </template>
+            </CreateGameToggle>
+
+            <div class="create-game-subsection-label" v-i18n>Fan-made</div>
+
+            <CreateGameToggle v-model="aresExtension" title="Ares" class="expansion-button">
```
kberg, line 94: So this is a great start, because puts most of the work into the components, but this makes individual checkboxes larger, not smaller. Can this be turned into something that reduces the overall size of a component, like this?

<CreateGameToggle v-model="aresExtension" title="Ares" icon="ares" tooltip="http://....">
</CreateGameToggle>

And then icon can be expanded into the full expansion icon name, and the optional tooltip can be appended.

src/client/components/create/CreateGameToggle.vue
```diff
@@ -0,0 +1,31 @@
...
+
+export default Vue.extend({
+  props: {
+    value: {type: Boolean, required: true},
+    title: {type: String, default: ''},
+    id: {type: String, default: ''},
+  },
+  data() {
+    return {
+      inputId: 'i' + (nanoid(10)),
```
kberg, line 26: Since this is the only purpose for nanoid, I suggest removing it, and relying on the id prop only. For example,

`<label :for="id + '-checkbox'">`

Not only does that remove the change to package-lock.json, but uses the id property which for the moment is unused.

kberg (changes requested):
Great start! I have a couple other observations, which I'll add below.

kberg:
This is what the initial page looks like in production:
image

and this is what your PR generates:

image

Two things I see are that Promos is listed twice, and there are two defaults that aren't set (Corporate Era and Show Timers.)

kberg:
When clicking "Venus Next" this is what happens to "The Moon":
image

kberg:
Adding Turmoil includes "Standard Project Variant"

image

kberg:
A brief test with Turmoil, Ares and Elysium generated exactly that, which verifies that the model is doing the right things. Great!

image

I think there's not quite that much to do here, this is a great start, but needs cleanup, etc.

src/client/components/create/CreateGameToggle.vue
immdevrov: Oh, I just forgot to remove id property, in my opinion adding a new prop just overkill, we have to keep components simpler — we don't need to know this needs _id_, and should let the component decide this.
And nanoid is very VERY lightweight library (look at it like it helper function). [Its weight is about 500 bytes gzipped](https://bundlephobia.com/package/nanoid@3.1.30) and there is no way it'll cause some problems with performance.

src/client/components/create/CreateGameForm.vue
immdevrov: Yep, I have some thoughts how to reduce it, it is just a first step. I thought It will be better to divide and conquer.

For example. To reduce size of a component, we have to slice it up into few different forms, which will contain some parts of form (divide a big form into a component which control several smaller ones).
Secondly, I think there is a way for using loop render (v-for), which can help to make the form drastically smaller

src/styles/debug_ui.less
immdevrov: O, because I moved a class _.expansion-button_ from <label> to <div>, so had to change all the usages across an app.
I thought if I didn't do it, something may crack into other app parts.
Right question here – why this is using the same class as in create game form, if this is not related?

immdevrov:
A lot of misplacing happened because of not the best knowledge of how interface should look like, thanks for testing, I'll fix it.

src/client/components/create/CreateGameToggle.vue
kberg: We've adopted a pretty strong "no new components unless necessary" approach, mostly to limit depednencies and rely on active projects as best as possible. This one has no dependencies and _is_actively maintained, but. It's only being used for a form id, which could be replaced with `Math.random().toString()`

image

Don't get me wrong, it's not pretty, but it's essentially the same thing. (Not to mention our existing mechanism for generating unique numbers which you can see in `GameHandler`:

```
  public generateRandomId(prefix: string): string {
    // 281474976710656 possible values.
    return prefix + Math.floor(Math.random() * Math.pow(16, 12)).toString(16);
  }
```

Besides that, though, I'd rather keep the more readable externally-supplied id (or as I was suggesting before, id fragment), please.

src/styles/debug_ui.less
kberg: Because our CSS is a bit sloppy and that UI component is a bit sloppy. Mea culpa. OK, thank you, this makes sense!

kberg:
> 

Yes, thank you! I'm sure we can get this working.

src/client/components/create/CreateGameToggle.vue
immdevrov: If using Math.random is the common way to deal with it in the project – it is ok, I convert to it.

kberg: Sorry I wasn't clear, but the last sentence of my long comment said that actually what I'm asking for is to just supply a readable ID like what was supplied before. I wrote too much about how we didn't need the new module. I shouldn't have written so much.

immdevrov:
About changes: 
- Replaced nanoid with Math.random()
- Fixed styles (wanna keep them without attribute selector)
- Fixed placement for a little bit (div with width instead of spaces)

immdevrov:
About package-lock problem.
screen from [npm docs](https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json)

изображение

lockVersion: 2 is designed for npm v7, which is designed with node 15.

As I see in .nvmrc project is using node v14, so it is ok to  use npm v6 and lockVersion: 1 (but we can switch ofc)

src/client/components/create/CreateGameToggle.vue
immdevrov: Why do you need a readable ID? It is one more unnecessary prop here.

kberg:
> About changes:
> 
> * Replaced nanoid with Math.random()

I'm sorry if it wasn't clear: please use the ID field and have the caller supply an id. I do not want to use random, either.

> * Fixed styles (wanna keep them without attribute selector)
> * Fixed placement for a little bit (div with width instead of spaces)

immdevrov:
I am sorry, I will not do it. I think this is wrong decision and as a professional I do not wont to contribute a code which I find incorrect. I'm open to the conversation, but you have all rights to stand your ground, this is your project.

Furthermore, I found a lot of bad code in a project (bad from my perspective ofc, and I understand this is just my opinion) and now I see why. I do not want to sound disrespectful, actually I admire your work to make my favorite board game available online, and you did a great job (better than steam version), and as a user I appreciate it. But as a software developer, I believe we can not work together.

I am genially sorry, and decide to write this to leave the things clear, instead of just disappearing. So it's ok for you to close merge, or finish it by yourself if you want.

Have a nice day and good luck with the project!

kberg:
> I am sorry, I will not do it. I think this is wrong decision and as a professional I do not want to contribute a code which I find incorrect. I'm open to the conversation, but you have all rights to stand your ground, this is your project.

Reading your message, I do not think you are open to a conversation. In fact, you seem pretty closed to it.

> Furthermore, I found a lot of bad code in a project (bad from my perspective ofc, and I understand this is just my opinion) and now I see why. I do not want to sound disrespectful, actually I admire your work to make my favorite board game available online, and you did a great job (better than steam version), and as a user I appreciate it. But as a software developer I believe we can not work together.

Yes, there is a lot of messy code in this project. *A LOT.* I've made plenty of them, but keep in mind:

* The majority of people who started this project (it wasn't me - I joined over a year into its development) were not professional software developers. Many were hobbyists who did not take best practices, latency, reperformance into consideration.
* They chose Vue because they _didn't_ know it, and wanted to learn it.

And yet, with those characteristics, they still managed to create a fully-featured game. (By the time I got on board, all the main expansions were complete!) I'm not trying to say I don't write my share of bad code, but it's just not on the front-end. (I'm still embarrassed by `_AresSetup`.)

Should every PR fix every small thing? No. Should a brand new contributor get their way with what is arguably a minor offense? No.

Here's something you can think about: can we not work together because my standards are too low, or because your expectations are too high? Imagine what we could have accomplished if you ignored that one thing.

bafolts:
I was glad to see that @kberg had requested we avoid taking on a new dependency to generate an id. When I originally started the project I had a goal to use NO dependencies other than typescript. If you go back to the early code you will see a very minimal package.json file. Usually every javascript project I work in ends up bogged down by 100+ megabytes of dependencies that have to be downloaded. I think this lack of dependencies is one of the reasons we have ran over 100k games on a hobby plan heroku server with minimal downtime.

I am actually quite proud of the state of the code given it has only been worked on pro bono by strangers. It hasn't been too painful to add all the new expansions as the game expanded. This project is almost 4-5 years old now! For a project of it's age I think it is in great shape. Usually as a project gets to this age it can be difficult to change anything without breaking other parts of the system unexpectedly. The front end is definitely the weaker part of the code but can be more challenging to update as often times one would have to update it all at once. For example if we wanted to convert to react or some other front end framework besides vue-js one would have to either have a hybrid system where some components are written in vue-js and others in react or write every component in react and then cut over. Work like this can be difficult to setup pro bono as we are.

immdevrov:
This is your project and it's working and this is the only thing which matters
 Good luck!

PR #3910, opened 2022-01-06
SelectHowToPay for Soylent Seedling Systems.

kberg:
I'm breaking this into three or four pieces because without tests this
touches 25 files. That's how complicated our payment structure is; you
have to go through quite a lot of layers. I'd like to simplify this just
a little more before adding one more of these. No need for all the
hardcoding.


src/client/components/SelectHowToPay.vue
```diff
@@ -275,6 +284,14 @@ export default Vue.extend({
       <Button type="max" @click="setMaxValue('heat')" title="MAX" />
     </div>
 
+    <div class="payments_type input-group">
```
bafolts, line 287: Will this always show for a user even if paying with seeds is irrelevant for their game options?

src/client/components/SelectHowToPayForProjectCard.vue
```diff
@@ -234,6 +241,19 @@ export default Vue.extend({
...
     },
+    canUseSeeds() {
+      // FYI Seed Resources are limited to the Soylent Seedling Systems corp card, which allows spending its
+      // resources for plant cards and the standard greenery project.
+      if (this.card !== undefined && (this.playerinput.seeds ?? 0) > 0) {
+        if (this.tags.includes(Tags.PLANT)) {
+          return true;
+        }
+        if (this.card.name === CardName.GREENERY_STANDARD_PROJECT) {
+          return true;
```
bafolts, line 252: Well this worked out well. Benefit of having standard projects as cards.

kberg: Didn't it?

src/client/components/SelectHowToPay.vue
kberg: Whoops! It will. I'll fix that.

src/client/components/SelectHowToPay.vue
```diff
@@ -276,6 +285,14 @@ export default Vue.extend({
...
 
     <div class="payments_type input-group">
+      <i class="resource_icon resource_icon--seeds payments_type_icon" :title="$t('Pay by Seeds')"></i>
+      <Button type="minus" @click="reduceValue('seeds', 1)" />
+      <input class="form-input form-inline payments_input" v-model.number="seeds" />
+      <Button type="plus" @click="addValue('seeds', 1)" />
+      <Button type="max" @click="setMaxValue('seeds')" title="MAX" />
+    </div>
+
+    <div class="payments_type input-group" v-if="playerinput.canUseSeeds">
```
bafolts, line 295: I think this should be on the div for paying with divs and not on this div for paying with mega credits.

kberg: So sloppy! That's why I added a test. SRY SRY SRY

kberg: OK tests pass now.  Time is valuable sorry for wasting yours! :D

tests/components/PaymentTester.ts
```diff
@@ -41,30 +41,7 @@ export class PaymentTester {
...
-      vmVal = this.model.microbes;
-      break;
-    case 'megaCredits':
-      vmVal = this.model.megaCredits;
-      break;
-    case 'science':
-      vmVal = this.model.science;
-      break;
-    }
+    const vmVal = this.model[type];
```
bafolts, line 44: This is a great change. 👏

bafolts: I see you.

bafolts (approved):

tests/components/PaymentTester.ts
kberg: I keep trying to simplify SHTP and once in a while it works.

PR #3911, opened 2022-01-06
Second piece of Soylent Seeding Systems. 

kberg:
This is mostly structural stuff, and I don't see much for tests in here;
these are typically not tested directly, but will be part of the third
PR's tests.

**DEPENDS on #3910**


src/Player.ts
```diff
@@ -1859,7 +1881,8 @@ export class Player implements ISerializable<SerializedPlayer> {
       (canUseTitanium ? (this.titanium - reserveUnits.titanium) * this.getTitaniumValue() : 0) +
       (canUseFloaters ? this.getFloatersCanSpend() * 3 : 0) +
       (canUseMicrobes ? this.getMicrobesCanSpend() * 2 : 0) +
-      (canUseScience ? this.getSpendableScienceResources() : 0);
+      (canUseScience ? this.getSpendableScienceResources() : 0) +
+      (canUseSeeds ? this.getSpendableSeedResources() * 5 : 0);
```
bafolts, line 1885: Are seeds worth 4 or 5 mega credits?

src/deferredActions/SelectHowToPayDeferred.ts
```diff
@@ -12,10 +12,24 @@ export class SelectHowToPayDeferred implements DeferredAction {
...
+    if (this.player.canUseHeatAsMegaCredits && this.player.heat > 0) {
+      return false;
+    }
+    if (this.options.canUseSteel && this.player.steel > 0) {
+      return false;
+    }
+    if (this.options.canUseTitanium && this.player.titanium > 0) {
+      return false;
+    }
+    if (this.options.canUseSeeds && (this.player.corporationCard?.resourceCount ?? 0 > 0)) {
```
bafolts, line 25: Using `player.getSpendableSeedResources` here would avoid having to assume that if `canUseSeeds` is true that a player has the right corporation card. As of now this will work fine. If the logic for what can use seeds is updated we would most likely put that logic in `getSpendableSeedResources` and might miss putting it here as well.

bafolts (approved):

src/Player.ts
kberg: > Are seeds worth 4 or 5 mega credits?

5.

src/deferredActions/SelectHowToPayDeferred.ts
kberg: Isn't `canUseSeeds` still necessary?

kberg: Yeah it's still necessary because it's unclear what condition is necessary. `mustPayWithMegacredits` needs the parameter.

PR #3912, opened 2022-01-07
Soylent Seedling Systems party 3, the card.

kberg:


src/Player.ts
```diff
@@ -1437,6 +1441,10 @@ export class Player implements ISerializable<SerializedPlayer> {
       totalToPay += howToPay.science;
     }
 
+    if (howToPay.seeds ?? 0 > 0) {
+      totalToPay += howToPay.seeds * 5;
```
bafolts, line 1445: Might make sense to make this `5` a constant similar to `DEFAULT_STEEL_VALUE`.

src/cards/base/standardProjects/GreeneryStandardProject.ts
```diff
@@ -21,6 +21,14 @@ export class GreeneryStandardProject extends StandardProjectCard {
     });
   }
 
+  public canPayWith(player: Player) {
```
bafolts, line 24: Not sure why I am just now recommending this but we should use the `override` keyword introduced in typescript 4.3 anywhere we can. There is a flag we can turn on to require it's usage. Should help with reading and help us find places we override.
```suggestion
  public override canPayWith(player: Player) {
```

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-3.html

src/cards/pathfinders/SoylentSeedlingSystems.ts
```diff
@@ -0,0 +1,55 @@
...
+      resourceType: ResourceType.SEED,
+
+      metadata: {
+        cardNumber: 'PfC8',
+        renderData: CardRenderer.builder((b) => {
+          b.megacredits(38).seed().seed().br;
+          b.effect('When you place a greenery tile, add 1 seed resource to this card.', (eb) => {
+            eb.greenery().startEffect.seed();
+          }).br;
+          b.effect('When paying for a plant card, or the STANDARD GREENERY PROJECT, seeds here may be used as 5 M€ each.', (eb) => {
```
bafolts, line 30: The image I find online with google for this card uses 4 instead of 5.

https://cf.geekdo-images.com/_KxHiT-eLhgkQiEkmr-bdQ__imagepage/img/3lXpbS3LD7Vmfibh2fPAwpRu-lk=/fit-in/900x600/filters:no_upscale():strip_icc()/pic4267092.jpg

Not sure which is right.

bafolts (approved):

src/Player.ts
kberg: Good idea, done. Could do the same for floaters and microbes. Will follow up with that.

src/cards/base/standardProjects/GreeneryStandardProject.ts
kberg: Yeah let's.

src/cards/pathfinders/SoylentSeedlingSystems.ts
kberg: I'm going with this one image

which comes from a source resource specified by the designer. Can you point to the origin of that image? I'm curious.

kberg:
Gotcha.

PR #3923, opened 2022-01-10
Lava Flows space filter to understand Ares Hazards.

kberg:


src/cards/base/LavaFlows.ts
```diff
@@ -36,10 +36,11 @@ export class LavaFlows extends Card implements IProjectCard {
     const board = player.game.board;
     const volcanicSpaceIds = board.getVolcanicSpaceIds();
 
-    const spaces: Array<ISpace> = (volcanicSpaceIds.length > 0) ?
-      volcanicSpaceIds.map((id) => board.getSpace(id)) :
-      board.getSpaces(SpaceType.LAND, player);
-    return spaces.filter((space) => space.tile === undefined && (space.player === undefined || space.player === player));
+    const spaces = board.getAvailableSpacesOnLand(player);
+    if (volcanicSpaceIds.length > 0) {
+      return spaces.filter((space) => volcanicSpaceIds.includes(space.id));
```
bafolts, line 41: We should eventually use a `Set` for `getVolcanicSpaceIds`. It would help here with this look-up and also better reflects the data.

bafolts: I remember a bug report where supposedly `LavaFlows` allowed someone to select any space. When is `volcanicSpaceIds.length === 0` ?. Prior to this change we most likely allowed you to pick any space when `volcanicSpaceIds.length` was `0`.

bafolts (approved):

src/cards/base/LavaFlows.ts
kberg: Hellas does not have volcanic spaces. image

PR #3937, opened 2022-01-13
This fixes #3936 - Luna Ecumenopolis placement bug

### LINKED ISSUE #3936: Luna Ecumenopolis bug in computing valid second placement
Luna Ecumenopolis has faulty logic for canPlay, where it looks for second colony placements spatially after the first colony placement. However, the first colony placement can create a new valid spot spatially before the first colony. The correct logic would be to restart from the beginning and ignore the firstSpaceID spot for finding a valid second placement.

In the example below, Green was forced to play a colony on (5,1) to use Luna Ecumenopolis. However, even without that colony playing (6,2) followed by (6,1) should have been valid.

image

puuj:
Checks all spaces for the second placement, not just those that come after the first placement. Additionally adds a test for a case where the first placement allows the placement of a second colony tile with a lower space index.

### DIFF
--- a/tests/cards/moon/LunaEcumenopolis.spec.ts
+++ b/tests/cards/moon/LunaEcumenopolis.spec.ts
@@ -47,6 +47,18 @@ describe('LunaEcumenopolis', () => {
     expect(player.getPlayableCards()).does.not.include(card);
   });
 
+  it('can play when 1st placement enables 2nd placcement', () => {
+    player.cardsInHand = [card];
+    player.megaCredits = card.cost;
+
+    const moon = moonData.moon;
+    moon.getSpace('m18').tile = {tileType: TileType.MOON_COLONY};
+    moon.getSpace('m19').tile = {tileType: TileType.MOON_COLONY};
+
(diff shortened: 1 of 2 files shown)


tests/cards/moon/LunaEcumenopolis.spec.ts
```diff
@@ -47,6 +47,18 @@ describe('LunaEcumenopolis', () => {
     expect(player.getPlayableCards()).does.not.include(card);
   });
 
+  it('can play when 1st placement enables 2nd placcement', () => {
```
kberg, line 50: If I read this correctly, this works because it enables 13, and then 12? If you can confirm, I'll add a comment just clarifying that.

kberg (approved):
This is a good fix! Thank you.

tests/cards/moon/LunaEcumenopolis.spec.ts
kberg: Actually I added the comment myself. Will merge momentarily.

PR #3946, opened 2022-01-15
Patch: Artificial Lake should be playable in edge case

nwai90:


kberg (approved):
Added a minor comment. Thank you!

PR #3947, opened 2022-01-15
Isolate fetching cards to one place on the client

kberg:
This will make it far easier to find a replacement that doesn't import actual card instances. Plus by doing this all the code uses the same API, making things so much simpler! Look at how small `AllCards.ts` is. I already see new minor things that can change and improve.

### DIFF
--- a/src/client/components/SelectInitialCards.vue
+++ b/src/client/components/SelectInitialCards.vue
@@ -18,7 +18,7 @@
 import Vue from 'vue';
 
 import Button from '@/client/components/common/Button.vue';
-import {CardFinder} from '@/CardFinder';
+import {getCard} from '@/client/cards/ClientCardManifest';
 import {CardName} from '@/CardName';
 import * as constants from '@/constants';
 import {CorporationCard} from '@/cards/corporation/CorporationCard';
@@ -28,6 +28,7 @@ import SelectCard from '@/client/components/SelectCard.vue';
 import ConfirmDialog from '@/client/components/common/ConfirmDialog.vue';
 import {PreferencesManager} from '@/client/utils/PreferencesManager';
 import {Tags} from '@/cards/Tags';
+import {PreludeCard} from '@/cards/prelude/PreludeCard';
 
 export default Vue.extend({
   name: 'SelectInitialCards',
@@ -67,11 +68,12 @@ export default Vue.extend({
     getAfterPreludes() {
       let result = 0;
       for (const prelude of this.selectedPreludes) {
-        const card = new CardFinder().getPreludeByName(prelude);
-        if (card === undefined) {
+        const cam = getCard(prelude);
+        if (cam === undefined) {
           throw new Error(`Prelude ${prelude} not found`);
         }
-        result += card?.startingMegaCredits ?? 0;
+        const card = cam.card as PreludeCard;
+        result += card.startingMegaCredits ?? 0;
 
         switch (this.selectedCorporation?.name) {
         // For each step you increase the production of a resource ... you also gain that resource.
@@ -185,7 +187,7 @@ export default Vue.extend({
       this.selectedCards = cards;
     },
     corporationChanged(cards: Array<CardName>) {
-      this.selectedCorporation = new CardFinder().getCorporationCardByName(cards[0]);
+      this.selectedCorporation = getCard(cards[0])?.card as CorporationCard;
     },
     preludesChanged(cards: Array<CardName>) {
       this.selectedPreludes = cards;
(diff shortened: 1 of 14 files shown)


src/client/components/SelectInitialCards.vue
```diff
@@ -67,11 +68,12 @@ export default Vue.extend({
...
       let result = 0;
       for (const prelude of this.selectedPreludes) {
-        const card = new CardFinder().getPreludeByName(prelude);
-        if (card === undefined) {
+        const cam = getCard(prelude);
+        if (cam === undefined) {
           throw new Error(`Prelude ${prelude} not found`);
         }
-        result += card?.startingMegaCredits ?? 0;
+        const card = cam.card as PreludeCard;
```
bafolts, line 75: Should we check if this is a prelude card? If a project card ended up in `selectedPreludes` this may try to load it. Prior to this we only looked through prelude cards. Now that we look through project cards we could get a project card. Using `instanceof` should check that it is a `PreludeCard`. I think if you check `instanceof` typescript may also then let you use it as `PreludeCard` without casting.

src/client/components/create/CorporationsFilter.vue
```diff
@@ -86,25 +76,35 @@ export default Vue.extend({
...
   data() {
-    const cardsByModuleMap: Map<GameModule, Array<CardName>> =
-            new Map(ALL_CARD_MANIFESTS.map((m) => [m.module, corpCardNames(m.module)]));
+    const cardsByModule: MultiMap<GameModule, CardName> = new MultiMap();
+    getCards(byType(CardType.CORPORATION)).forEach((cam) => {
+      if (cam.card.name !== CardName.BEGINNER_CORPORATION) {
+        cardsByModule.set(cam.module, cam.card.name);
+      }
+    });
+    console.log(iterator(cardsByModule));
```
bafolts, line 90: Assuming this was for debugging. Is if iterator isn't used anywhere else I would remove it. If keeping iterator I would consider a different name. Javascript has iterators but the function named iterator returns an Array which could be confusing.

src/client/components/create/CorporationsFilter.vue
```diff
@@ -116,9 +116,9 @@ export default Vue.extend({
       return [];
     },
     getItemsByGroup(group: string): Array<CardName> {
-      if (group === 'All') return allItems.slice();
+      if (group === 'All') return Array.from(this.cardsByModule.values()).slice();
```
bafolts, line 119: The `slice` here was being used to create a copy of `allItems` so that array wouldn't be modified. Using `Array.from` should create a new `Array` so the usage of `slice` shouldn't be needed. Could also see if this array were marked `ReadonlyArray<CardName>` if anything is trying to modify this. If nothing is trying to modify then could keep it as `ReadonlyArray`.

src/client/components/create/CreateGameForm.vue
```diff
@@ -415,6 +415,7 @@
                   v-bind:promoCardsOption="promoCardsOption"
                   v-bind:communityCardsOption="communityCardsOption"
                   v-bind:moonExpansion="moonExpansion"
+                  v-bind:pathfindersExpansion="pathfindersExpansion"
```
bafolts, line 418: Was this change intentional? Seems unrelated to the other changes.

bafolts (approved):

src/client/components/create/CorporationsFilter.vue
kberg: Oh yeah it's gotta go. Good catch.

src/client/components/create/CreateGameForm.vue
kberg: You know, you're right? It is unrelated. But somehow it was absent. So I tossed it in.

src/client/components/create/CorporationsFilter.vue
kberg: Good call removing `slice.`

Let's hold off on the `ReadonlyArray` stuff which is also a little more powerful thanks to recent TS versions: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html I think that kind of stuff can have a surprise effect so I'd rather be careful.

src/client/components/SelectInitialCards.vue
kberg: Good question. Actually I'm going to advise no because `PreludeCard` will eventually go away.  The client code will no longer rely on `ICard` or any `ICard`-related stuff. That's how we're gonna get rid of all the server code on the client.

PR #3951, opened 2022-01-16
Use PlaceOceanTile. It's simpler.

kberg:


bafolts (approved):
It also uses less code!

PR #3958, opened 2022-01-18
Enforce noImplicitOverride

bafolts:
This has the remaining `override` keywords and enables the `noImplicitOverride` in `tsconfig.json`.


kberg (approved):
Wow: you're done? Awesome.

PR #3959, opened 2022-01-19
Remove some complexity from CardRenderer.

kberg:


src/cards/render/CardRenderer.ts
```diff
@@ -7,11 +7,7 @@ import {TileType} from '../../common/TileType';
 export type ItemType = CardRenderItem | CardRenderProductionBox | CardRenderSymbol | CardRenderEffect | CardRenderTile | string | undefined;
 
 export class CardRenderer {
-  constructor(protected _rows: Array<Array<ItemType>> = [[]]) {}
-
-  public get rows() {
-    return this._rows;
-  }
+  constructor(public rows: Array<Array<ItemType>> = [[]]) {}
```
bafolts, line 10: If we don't intend on these being modified I would try to use `ReadonlyArray`. If it falls into place since we are never modifying it that could help indicate we don't expect it to be changed.

bafolts (approved):

src/cards/render/CardRenderer.ts
kberg: These are about to change, so I'll catch it on the next pass.

PR #3962, opened 2022-01-20
First step moving Tags to common.

kberg:
Turns out symlinks go a long way.


src/common/cards/Tags.ts
```diff
@@ -0,0 +1 @@
+../../cards/Tags.ts
```
bafolts, line 1: What does this file do?

kberg:
It's a symlink, believe it or not.

bafolts (approved):
Does seem that the file is a symlink when I look at it locally on a mac. This should be working for windows as well since the build passed.

PR #3964, opened 2022-01-20
Upgrade to Node v16.

kberg:
Node version upgrades seem to happen rapidly. The most current version is 17.

To get this to work in an old environment you might need to run `npm rebuild`

A couple of sanity tests passed for me, including deploying to Heroku.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,7 +11,7 @@ jobs:
     - uses: actions/checkout@v1
     - uses: actions/setup-node@v1
       with:
-        node-version: '10.x'
+        node-version: '16.x'
     - run: npm ci
     - run: npm run lint
     - run: npm run build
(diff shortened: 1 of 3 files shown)


kberg:
image
Node v14 is in maintenance mode.

I tried v17, but ran into a small error that I wanted to avoid for the time being. It looked like I would have to add an SSL environment variable, and that's scary without some research.

Argeento:
Maybe node version in CI also should be updated? -> https://github.com/terraforming-mars/terraforming-mars/blob/main/.github/workflows/main.yml#L14

kberg:
> Maybe node version in CI also should be updated? -> https://github.com/terraforming-mars/terraforming-mars/blob/main/.github/workflows/main.yml#L14

Yes, that's a good idea.

kberg:
This configuration is running at https://terraforming-mars-kberg.herokuapp.com/

bafolts (approved):

PR #3967, opened 2022-01-21
Buff Mars Direct.

kberg:
Mars Direct works because of all the Mars cards in the deck. When 8
cards are missing, that reduces the card's potential. A discussion was
raised to temporarily buff the card by 1.5MC per missing card.

Some Mars cards just won't get implemented, so this buff could in theory
be permanent in some form.


tests/cards/moon/MarsDirect.spec.ts
```diff
@@ -53,5 +55,17 @@ describe('MarsDirect', () => {
     player.megaCredits = 11;
     expect(player.getPlayableCards()).has.members([powerPlant]);
   });
+
+  it('verify buff is lined up with game state', () => {
+    let count = 0;
+    PATHFINDERS_CARD_MANIFEST.projectCards.factories.forEach((factory) => {
+      if (new factory.Factory().tags.includes(Tags.MARS)) {
+        console.log(factory.cardName);
```
bafolts, line 63: Assuming this was here for debugging purposes. Remove before flight.

src/cards/pathfinders/MarsDirect.ts
```diff
@@ -10,19 +10,22 @@ import {Tags} from '../Tags';
...
 // means this result can be changed by cards like CEO's Favorite Project.
+// It also means cards with a wild tag may impact this.
 export class MarsDirect extends Card implements CorporationCard {
   constructor() {
     super({
       cardType: CardType.CORPORATION,
       name: CardName.MARS_DIRECT,
       tags: [Tags.MARS],
-      startingMegaCredits: 52,
+      startingMegaCredits: 64,
```
bafolts, line 20: Making a variable for buff and then using `52 + buff` here may make this clearer.

bafolts (approved):

tests/cards/moon/MarsDirect.spec.ts
kberg: Aye sir_

PR #3974, opened 2022-01-22
Add the Wetlands card. (part 1)

kberg:
Wetlands is currently not registered, because not all the tests are
complete, and in fact, not all the behavior is complete. (There's a
few things to do to make the greenery side of Wetlands work.)


kberg:
I'm going to add this label for the moment because #3975 and #3976 should go in first.

PR #3976, opened 2022-01-22
Centralize greenery counting, will simplify Wetlands.

kberg:
Depends on #3975


src/cards/base/EcologicalZone.ts
```diff
@@ -54,12 +55,7 @@ export class EcologicalZone extends Card implements IProjectCard, IResourceCard
 
   private getAvailableSpaces(player: Player): Array<ISpace> {
     return player.game.board.getAvailableSpacesOnLand(player)
-      .filter(
-        (space) => player.game.board.getAdjacentSpaces(space).filter(
-          (adjacentSpace) => adjacentSpace.tile !== undefined &&
-              adjacentSpace.tile.tileType === TileType.GREENERY,
-        ).length > 0,
-      );
+      .filter((space) => player.game.board.getAdjacentSpaces(space).filter(Board.isGreenerySpace).length > 0);
```
bafolts, line 58: We could change this later but since this is checking for `length > 0` we should be able to swap with `.some(Board.isGreenerySpace)`.

src/cards/base/ImmigrationShuttles.ts
```diff
@@ -29,7 +29,7 @@ export class ImmigrationShuttles extends Card implements IProjectCard {
     });
   }
   public override getVictoryPoints(player: Player) {
-    return Math.floor(player.game.getCitiesInPlay() / 3);
+    return Math.floor(player.game.getCitiesCount(player) / 3);
```
bafolts, line 32: Should this be `getCitiesCount`? The victory point is rewarded for any city tile in play.

src/turmoil/parties/Greens.ts
```diff
@@ -52,7 +52,7 @@ class GreensBonus02 implements Bonus {
 
   getScore(player: Player) {
     const boardSpaces = player.game.board.spaces;
-    const count = boardSpaces.filter((space) => space.tile && space.tile.tileType === TileType.GREENERY && space.player !== undefined && space.player.id === player.id).length;
+    const count = boardSpaces.filter((space) => Board.isGreenerySpace(space) && space.player !== undefined && space.player.id === player.id).length;
```
bafolts, line 55: I think this can be `Board.isGreenerySpace(space) && Board.isOwnedBy(player)`

bafolts (changes requested):
Pretty sure I spotted a typo in ImmigrationShuttles. Otherwise only had minor suggestions.

src/cards/base/ImmigrationShuttles.ts
kberg: Good catch.

kberg: Added additional test to ImmigrationShuttles.

src/turmoil/parties/Greens.ts
kberg: It's not that neat: it's

`Board.ownedBy(player)(space)` beause ownedBy returns a function.

kberg: Rolling it back, it's not pretty.

bafolts (approved):

PR #3978, opened 2022-01-23
Add Pathfinders card Wetlands.

kberg:


kberg:
Depends on #3976

bafolts:
Can you merge this with main? The diff on github is showing a lot of modifications that already exist in main.

kberg:
> Can you merge this with main? The diff on github is showing a lot of modifications that already exist in main.

Done. That was because 
image

kberg:
> > Can you merge this with main? The diff on github is showing a lot of modifications that already exist in main.
> 
> Done. That was because image

(should be good now.)

src/cards/base/ArcticAlgae.ts
```diff
@@ -34,7 +34,7 @@ export class ArcticAlgae extends Card implements IProjectCard {
   }
 
   public onTilePlaced(cardOwner: Player, activePlayer: Player, space: ISpace) {
-    if (space.tile?.tileType === TileType.OCEAN) {
```
bafolts, line 37: This seems like different logic. Is there anything that can place an ocean tile on the non base places? Did this have to change with this PR?

bafolts: oh ok I read this wrong. The base ocean tiles is not the places you can place the tile but the tile types themselves.

bafolts (approved):

src/cards/base/ArcticAlgae.ts
kberg: Do you have a better name for it?

PR #3979, opened 2022-01-23
Add Pathfinders corporation Bio-Sol

kberg:
image


src/styles/cards_v2.less
```diff
@@ -826,6 +826,13 @@
                 background-size: 144px 45px;
                 margin-top: 10px;
             }
+            .card-bio-sol-logo {
+                background-image: url(./assets/pathfinders/corp-logo-bio-sol.png);
```
bafolts, line 830: Is `assets/pathfinders/bio-sol.png` used anywhere? Appears to be the same image as `corp-logo-bio-sol.png`.

bafolts (approved):

src/styles/cards_v2.less
kberg: Whoops! Thanks. I'll remove it.

PR #3980, opened 2022-01-23
Add Pathfinders card Charity Donation.

kberg:


kberg:
This turned out to be surprisingly simple.

src/Game.ts
```diff
@@ -1469,6 +1469,7 @@ export class Game implements ISerializable<SerializedGame> {
     space.player = undefined;
   }
 
+  // Players returned in play order starting with first player this generation.
```
bafolts, line 1472: At some point we should consider changing the name of this function as it is so much more than `getPlayers`.

src/cards/pathfinders/CharityDonation.ts
```diff
@@ -0,0 +1,78 @@
...
+
+        const cardIdx = this.cards.indexOf(card);
+        if (cardIdx > -1) {
+          this.cards.splice(cardIdx, 1);
+        }
+
+        this.player.cardsInHand.push(card);
+        game.log('${0} drew ${1}', (b) => b.player(this.player).card(card));
+
+        const nextIndex = (this.playerIdx + 1) % this.players.length;
```
bafolts, line 67: This is very elegant and well done.

src/Game.ts
kberg: Also, it's useless most of the time. Most of the time you don't care about the order. (Or like I do, care about a different order.)

src/cards/pathfinders/CharityDonation.ts
kberg: Yeah I'm pretty happy. This was a pleasure to write.

src/cards/pathfinders/CharityDonation.ts
```diff
@@ -0,0 +1,78 @@
...
+            .br
+            .text('In player order starting with you, each player takes one card in hand. Discard the remaining card.')),
+      },
+    });
+  }
+
+  public play(player: Player) {
+    const game = player.game;
+    const players = game.getPlayers();
+    const thisIdx = players.findIndex((p) => p === player);
```
bafolts, line 34: If `thisIdx` ends up being `-1` this may end up looping indefinitely as `boundaryIndex` would be negative. If `thisIdx` it may be worth throwing an error as we expect to get a result.

src/Game.ts
bafolts: When the caller doesn't care about the order it is also very inefficient.

src/cards/pathfinders/CharityDonation.ts
kberg: OK. I'll add it.

src/cards/pathfinders/MarsDirect.ts
```diff
@@ -17,16 +17,16 @@ export class MarsDirect extends Card implements CorporationCard {
       cardType: CardType.CORPORATION,
       name: CardName.MARS_DIRECT,
       tags: [Tags.MARS],
-      startingMegaCredits: 64,
+      startingMegaCredits: 62,
```
bafolts, line 20: Take care when merging this. With another mars tag merging around this time the merge may take 62 but will need to be 60.

bafolts (approved):

src/cards/pathfinders/CharityDonation.ts
kberg: Thinking more, no. It won't infinite loop. `this.player = this.players[playerIdx]` will fail.

It certainly won't loop infinitely like a catastrophic CPU event.

src/cards/pathfinders/MarsDirect.ts
kberg: I'm so aware of this ;)

PR #3983, opened 2022-01-24
Fixes #3982 by counting all cities not player's cities

### LINKED ISSUE #3982: Regression: Immigration Shuttles should count all cities not player cities
VP count computation code was previously `return Math.floor(player.game.getCitiesInPlay() / 3);`
Recently it was updated to ` return Math.floor(player.game.getCitiesCount(player) / 3);`
That last `player` should be `undefined`.

puuj:
This fixes a regression where Immigration Shuttles awarded points only for the player's cities. The prior implementation and the intent of the card is to award points for all cities.

### DIFF
--- a/src/cards/base/ImmigrationShuttles.ts
+++ b/src/cards/base/ImmigrationShuttles.ts
@@ -29,7 +29,7 @@ export class ImmigrationShuttles extends Card implements IProjectCard {
     });
   }
   public override getVictoryPoints(player: Player) {
-    return Math.floor(player.game.getCitiesCount(player) / 3);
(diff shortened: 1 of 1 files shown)


bafolts:
Thought I spotted this in https://github.com/terraforming-mars/terraforming-mars/pull/3976. Was this merged in another PR?

bafolts (approved):

puuj:
ah, didn't think to look in that PR. I'll close this because looks like @kberg did a better job and wrote a test too.

kberg:
Yeah. Thank you, though! I wonder when this happened.

PR #3984, opened 2022-01-25
avoid synchronous file access in production

bafolts:
I am putting up a new `sw.js` file in preparation for adding a service worker. While I was looking at `ServeAsset` I saw we were always checking if files exist. This avoids the check for production. Should avoid synchronous file i/o.

### DIFF
--- a/src/routes/ServeAsset.ts
+++ b/src/routes/ServeAsset.ts
@@ -132,7 +132,7 @@ export class ServeAsset extends Handler {
       }
 
       // Return not-compressed .js files for development mode
-      if (!this.fileApi.existsSync(file)) {
+      if (!isProduction() && !this.fileApi.existsSync(file)) {
         encoding = undefined;
         file = `build/${urlPath}`;
       }
--- a/tests/routes/ServeAsset.spec.ts
+++ b/tests/routes/ServeAsset.spec.ts
@@ -35,6 +35,8 @@ describe('ServeAsset', () => {
   let res: MockResponse;
   let ctx: IContext;
   let fileApi: FileApiMock;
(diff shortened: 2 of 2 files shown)


src/routes/ServeAsset.ts
```diff
@@ -132,7 +132,7 @@ export class ServeAsset extends Handler {
       }
 
       // Return not-compressed .js files for development mode
-      if (!this.fileApi.existsSync(file)) {
+      if (!isProduction() && !this.fileApi.existsSync(file)) {
```
kberg, line 135: You can write a test for this because `ServeAsset.spec.ts` tracks calls to `existsSymc`. Of course, this also means having to read `isProduction` which means injecting the value instead of what you'er doing.

tests/routes/ServeAsset.spec.ts
```diff
@@ -88,36 +94,55 @@ describe('ServeAsset', () => {
...
+      ...primedCache,
+      readFile: 1,
+      existsSync: 1,
+    });
+  });
 
-  // Cached CSS
-  // Uncached CSS
+  it('production main.js', () => {
+    process.env.NODE_ENV = 'production';
```
kberg, line 139: This is clever. Good.

tests/routes/ServeAsset.spec.ts
```diff
@@ -35,7 +35,13 @@ describe('ServeAsset', () => {
   let res: MockResponse;
   let ctx: IContext;
   let fileApi: FileApiMock;
-
+  // The expected state of call counts in most simple cases in this test. This is a template
```
kberg, line 38: I added this comment.

kberg (approved):

PR #3987, opened 2022-01-25
More moving Tags to common (3/5)

kberg:


src/cards/pathfinders/BotanicalExperience.ts
```diff
@@ -0,0 +1,53 @@
...
+import {CardRequirements} from '../CardRequirements';
+import {all} from '../Options';
+import {Tags} from '../../common/cards/Tags';
+import {Size} from '../render/Size';
+import {ISpace} from '../../boards/ISpace';
+import {Board} from '../../boards/Board';
+import {ResourceType} from '../../common/ResourceType';
+import {IResourceCard} from '../ICard';
+
+export class BotanicalExperience extends Card implements IProjectCard, IResourceCard {
```
bafolts, line 16: Woah? This seems out of place.

bafolts (approved):
I would remove the one card this seems to be adding from this commit prior to merging.

src/cards/pathfinders/BotanicalExperience.ts
kberg: Oh! Look at that, there's also a merge conflict.

kberg: I'm sure resolving the merge conflict takes care of everything.

kberg: Yep, taht fixed it. Thanks.

PR #3996, opened 2022-01-26
Simplify Robotic Workforce test.

kberg:
Also rely on existing behavior to simulate these cards. An upcoming card
relies on the card being placed in hand before running deferred actions,
and this is pretty normal behavior. That this test did it itself, means
that it was using abnormal behavior. Not to mention it was doing more
work that way.

(Edited to add: in retrospect, this doesn't look smaller, does it? But it's certainly more standard.


nwai90 (approved):

kberg:
Thank you!

PR #3997, opened 2022-01-26
Add Pathfinders card Communication Center.

kberg:
The change to RoboticWorkforce is also in #3996, so will probably be invisible by the time this gets reviewed.


kberg:
Redrew the card.

image

src/Player.ts
```diff
@@ -736,6 +736,10 @@ export class Player implements ISerializable<SerializedPlayer> {
       card.resourceCount += count;
     }
 
+    if (typeof(options) !== 'number' && options.log === true) {
```
bafolts, line 739: May be worthwhile to store off this `typeof(options) !== 'number'` check into a constant variable in this scope. Will avoid having to type the check and run the check each time. If the type of the options argument is updated it will also be less we have to update.

src/pathfinders/PathfindersExpansion.ts
```diff
@@ -53,16 +52,16 @@ export class PathfindersExpansion {
...
-    //         (c as CommunicationCenter).bonus(p);
-    //       }
-    //     });
-    //   });
-    // }
+    // Communication Center hook
+    if (card.cardType === CardType.EVENT) {
+      player.game.getPlayers().forEach((p) => {
+        p.playedCards.forEach((c) => {
+          if (c.name === CardName.COMMUNICATION_CENTER) {
```
bafolts, line 59: If only one player can play `CommunicationCenter` it would be slightly more efficient to use a `for...of` loop and then `break` once the card is found.

bafolts (approved):

src/Player.ts
kberg: My opinion: it sacrifices readability for an optimiziation that is, for the moment, a drop in the bucket.

src/pathfinders/PathfindersExpansion.ts
kberg: Done.

PR #4000, opened 2022-01-26
Update Dockerfile to be more friendly for development

zwrss:
I did a little refactoring of the Dockerfile to make some intermediate layers to be cached, so you don't have to go through all the steps everytime you change a small thing. Put some comments to make it more clear.
The image is about 50MB larger then before - so that is a drawback.


Dockerfile
```diff
@@ -1,28 +1,39 @@
...
+WORKDIR /app
 
-WORKDIR /usr/src/app
+# Install dependencies first, to cache the image.
+COPY ["package.json", "package-lock.json", "./"]
+
+# Set node environment to 'production'. Go with 'development' for local deployments.
+# todo - not working yet # ENV NODE_ENV=production
+
+RUN npm install
```
jholm117, line 15: Would `npm ci` be better here?
https://docs.npmjs.com/cli/v8/commands/npm-ci

Dockerfile
```diff
@@ -1,28 +1,39 @@
...
 
-RUN mkdir -p /usr/src/app/db \
-  && apk add --no-cache --virtual .gyp git python3 make g++ \
-  && ln -sf python3 /usr/bin/python \
-  && npm install \
-  && npm run build \
+RUN npm run build 
+
+# Pruning.
+RUN npm prune \
```
jholm117, line 23: Instead of pruning, it might be simpler and more effective to use a multistage build.

https://docs.docker.com/develop/develop-images/multistage-build/#name-your-build-stages

This will allow you to copy only the important files from the build stage to the final stage, leaving everything else behind.

Dockerfile
kberg: That link says

> This command is similar to npm install, except it's meant to be used in automated environments such as test platforms, continuous integration, and deployment -- or any situation where you want to make sure you're doing a clean install of your dependencies.

So... yes?

jholm117: ok cool, that's what I figured. I guess we'll see what the author says but wasn't sure if there was some esoteric reason for not using it. Also just saw the CI validation is using `npm ci` as well

Dockerfile
```diff
@@ -1,28 +1,39 @@
...
 
-EXPOSE 8080
+# Prepare base image to cache.
+RUN apk add --no-cache --virtual .gyp git python3 make g++ \
+  && ln -sf python3 /usr/bin/python
 
-LABEL maintainer="bafolts" \
-      name="terraforming-mars" \
-      Version="1.0"
+WORKDIR /app
```
AndrewSav, line 7: Please do not do it. You break all exisitng docker-compose that map the old location

Dockerfile
AndrewSav: This is a good suggestion, I just was not sure which files after the build are required to run the app. If you know how to identify them this is a better approach.

Dockerfile
```diff
@@ -1,28 +1,39 @@
...
 
-WORKDIR /usr/src/app
+# Install dependencies first, to cache the image.
```
AndrewSav, line 9: Oh so that's why they were split. That makes sense, you should have commented on my PR ;)

Dockerfile
kberg: Who should `maintainer` be?

AndrewSav: Sorry, I was commenting on the change of the `WORKDIR` I have no opinion about the maintaner. What was there before the PR was fine with me.

Dockerfile
jholm117: In fact running these commands as a separate docker RUN layer will not actually reduce the size of the image. They will still exist in the previous layers.

AndrewSav: @jholm117 Have you tested it? I did because I was not sure, and I got smaller image.

jholm117: I have not tested this dockerfile in particular, but have tested in the past. That is surprising to me. How are you determining the size?

AndrewSav: As reported by `docker images`.

Dockerfile
bafolts: I don't develop with docker or use docker locally. Someone else will need to approve this file as my approval would be rubber stamping. Most all of the app runs with only node and sqlite so there isn't much setup. An `npm install` on most machines and operating systems that support node should work just the same.

kberg: I'll own the approval. I'm listening in. @AndrewSav, @jholm117, @zwrss, please be explicit about when you think the PR is good enough to go in.

zwrss: Change in workdir was not intended. Been doing a small experiment with local image. I'll revert this as well as apply some of the changes from other comments. I'll update the code and comment in other sections.

Dockerfile
zwrss: Did not think of this that time. Rebuilding image for the 5th time got me thinking.

Dockerfile
zwrss: That would prevent installing and removing packages as well. I'll try to implement this right away.

Dockerfile
zwrss: I am not a node veteran, so did not rly know about npm ci. npm install was used in the previous version of dockerfile so I just refactored. I'll change to npm ci.

zwrss:
I've added intermediate stage to Dockerfile for building the app. Final image is now free of packages used only for building and free of unnecessary files. Please have a look at the files copied from building stage to final image where I am only coping a few and may be missing something important.
Final image is now lighter (626MB) compared to original (~1GB).
Docker building is now faster as well when there is no pruning involved.

Dockerfile
```diff
@@ -1,28 +1,48 @@
...
+
+WORKDIR /usr/src/app
+
+# Copy required files.
+
+COPY ["package.json", "package-lock.json", "./"]
+
+COPY assets ./assets
+
+COPY --from=builder /usr/src/app/node_modules ./node_modules
```
jholm117, line 34: Nice job with the multi stage refactor :+1: This is a big improvement and good to merge. 

The only other thing I think you might want to consider, either now or possibly for future work, is that the `devDependencies` are still bundled into the final image even though they are not required to run the app. To reduce the image size even further you can drop them by running an `npm ci --production` I believe. Now this would require that `package.json` is properly configured, such that all runtime dependencies are listed under  `dependencies` and not `devDependencies` which will require testing that the container can run without error. 

If you choose to add this I would recommend splitting the initial `RUN npm ci` step into its own install stage. Following that you can create a `builder` stage and an `installProd` that depend on `install`. This will allow the two stages to run in parallel and you will not have to do a full install twice, only delete the `devDependencies` from the first install. 
```
FROM node:16.13.2-alpine3.15 AS install 
RUN npm ci

FROM install AS builder
RUN npm run build

FROM install AS installProd
RUN npm ci --production

FROM node:16.13.2-alpine3.15
COPY --from=installProd node_modules node_modules
```

zwrss: Hello. Changing `npm ci` to `npm ci --production` is a good point, thank you. Yet I don't quite get the idea behind adding new stages. Isn't that just going to do the whole install process twice? Why not just go with `--production` and `--prefer-offline` flags in first `npm ci` and skip the other?

jholm117: The `devDependencies` are required to run `npm run build` so if you run `npm ci --production` in the first install you will not be able to build the app.

In the example above, the `installProd` stage will inherit the filesystem from `install` so all the node modules  will already be installed and the local npm cache will be loaded. Running `npm ci --production --prefer-offline` at this point will indeed delete the `node_modules` directory and reinstall only the production dependencies *but they will come from the local cache* so you will not have to make any network requests.

zwrss:
Added intermediate stages in docker build to prepare prod dependencies to be copied to final image as suggested by @jholm117  in [comment](https://github.com/terraforming-mars/terraforming-mars/pull/4000#discussion_r795743675) . Additionally I moved adding user tfm before copying files to target image to speed up building a little more.
New image size is 309MB.

AndrewSav:
> I've added intermediate stage to Dockerfile for building the app. Final image is now free of packages used only for building and free of unnecessary files. Please have a look at the files copied from building stage to final image where I am only coping a few and may be missing something important. Final image is now lighter (626MB) compared to original (~1GB). Docker building is now faster as well when there is no pruning involved.

Currently the image as reported by `docker images` (before this PR) is 385MB, what explains the increase to 626MB?

zwrss:
> > I've added intermediate stage to Dockerfile for building the app. Final image is now free of packages used only for building and free of unnecessary files. Please have a look at the files copied from building stage to final image where I am only coping a few and may be missing something important. Final image is now lighter (626MB) compared to original (~1GB). Docker building is now faster as well when there is no pruning involved.
> 
> Currently the image as reported by `docker images` (before this PR) is 385MB, what explains the increase to 626MB?

It's 309MB now. 
It was ~1GB before recent `.dockerignore` update.

jholm117:
LGTM 👍  nice job!

andrewsav-bt:
Thank you for your work @zwrss, could you please rebase your branch on top of master, if not too much trouble, there were a few changes to .dockerignore, and I would like to test it in "final" state before we merge it. Thank you in advance. (I guess since only a single file changed in the PR I can just copy over, but we might as well follow accepted practices ;)

AndrewSav:
Thank you for your work @zwrss, could you please rebase your branch on top of master, if not too much trouble, there were a few changes to .dockerignore, and I would like to test it in "final" state before we merge it. Thank you in advance. (I guess since only a single file changed in the PR I can just copy over, but we might as well follow accepted practices ;)

zwrss:
> Thank you for your work @zwrss, could you please rebase your branch on top of master, if not too much trouble, there were a few changes to .dockerignore, and I would like to test it in "final" state before we merge it. Thank you in advance. (I guess since only a single file changed in the PR I can just copy over, but we might as well follow accepted practices ;)

Fetching upstream on my fork merged main into the branch so it's done. Hope that merging instead of rebasing is ok.

AndrewSav:
LGTM. Thanks!

AndrewSav:
>  Hope that merging instead of rebasing is ok.

I think it's a bit easier/cleaner to rebase than to merge, because when you rebase you take all your changes and apply them on the top of the current master. This way you have a liner clean history. The disadavantage is that if someone branched off your work you mess them up. Fortunately no one really branches off PRs so it is rarely a concern. However public branches (e.g master) for that reason should never be rebased (with exceptions like accidently commited credentials).

I just realised by looking at the below that in this project we are not really concerned with a clean history. In either case in context of this PR this is not a big deal at all.

kberg (approved):

PR #4003, opened 2022-01-28
Many fixes for rendering global events.

kberg:
Add title, also, compress large titles.
Remove unised 'large' for digit.
Widen the entire global event.
Realign several global events for clarity.
MC -> M€


kberg:
Global events are visible at the bottom of http://terraforming-mars-kberg.herokuapp.com/cards

bafolts (approved):

PR #4005, opened 2022-01-28
Move the policy HTML generation to a template.

kberg:
I think this is working.

image
image


src/client/components/turmoil/Agendas.vue
```diff
@@ -0,0 +1,209 @@
+<template>
+  <div :class="getClass">
+    <template v-if="id === 'mb01'">
+      <div class="resource money party-resource">1</div> /
+      <div class="resource-tag tag-building party-resource-tag"></div>
+    </template>
+    <template v-if="id === 'mb02'">
```
bafolts, line 7: Can this make use of `v-else-if`?

src/client/components/Turmoil.vue
```diff
@@ -238,68 +99,45 @@ export default Vue.extend({
...
-          break;
+          return politicalAgendas.greens.policyId;
         }
       }
-
-      if (policyId !== undefined) {
-        const policy = AGENDA_HTML.get(policyId);
-        return policy || `No ruling Policy`;
-      }
-      return '<p>' + this.$t('No ruling Policy') + '</p>';
```
bafolts, line 302: Did this get lost? I don't see if in the templates. It should be in a final `v-else`.

src/client/components/Turmoil.vue
```diff
@@ -238,68 +99,45 @@ export default Vue.extend({
...
-          bonusId = politicalAgendas.greens.bonusId;
-          break;
+          return politicalAgendas.greens.bonusId;
         }
       }
-      if (bonusId !== undefined) {
-        const bonus = AGENDA_HTML.get(bonusId);
-        return bonus || `No ruling Bonus`;
-      }
-      return `No ruling Bonus`;
```
bafolts, line 270: Did this go missing or do we not need it?

src/client/components/turmoil/Agendas.vue
kberg: It could, though it's kind of valueless here.

src/client/components/Turmoil.vue
kberg: It won't be possible. The type is now tightly defined.

src/client/components/Turmoil.vue
kberg: Same -- we don't need it.

src/client/components/turmoil/Agendas.vue
bafolts: The value would only be for performance. With this `v-if` I believe `vue` will have to check each `id` even after it gets a match. With `v-else-if` once a match is reached `vue` doesn't have to compare the rest. Could also let reader know this is used like a switch statement.

bafolts (approved):

src/client/components/turmoil/Agendas.vue
kberg: Done. Reasonable. Vue.js, where's my `v-switch`?

src/client/components/Turmoil.vue
kberg: Ah the heck with it. I'm being silly not adding a default failure condition.

PR #4009, opened 2022-01-29
Add docker build to ci

jholm117:
This adds a job to run docker build to the ci validation

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,4 +29,12 @@ jobs:
     - run: npm run lint
     - run: npm run build
     - run: npm test
+    
+  build-docker:
+    runs-on: ubuntu-latest
+    
+    steps:
+    - uses: actions/checkout@v1
+    - run: docker build .
+        
 


kberg:
I'd like to commit this, with a caveat: we don't support Docker. If it breaks, nobody here will know how to support it. On the positive side, it looks like a pretty quick build action

image

and also doesn't seem "required." Is my reading of "required" correct? Will this still permit merging even if the Docker CI test doesn't pass?

jholm117:
> and also doesn't seem "required." Is my reading of "required" correct? Will this still permit merging even if the Docker CI test doesn't pass?

in order for it to be required, it has to be added under branch protection rules in the repo settings. 

> I'd like to commit this, with a caveat: we don't support Docker. If it breaks, nobody here will know how to support it. On the positive side, it looks like a pretty quick build action

I see your point. Well you could commit it and not make it required so that it won't block anyone but at least you will know if it's broken

jholm117:
And would hopefully prevent future changes from breaking it

kberg:
> And would hopefully prevent future changes from breaking it

Yes, as long as I'm the steward of this project I'll do my best.

kberg (approved):

kberg:
@jholm117 looks as though all Docker builds are failing. See https://github.com/terraforming-mars/terraforming-mars/runs/5193543084?check_suite_focus=true which says

```
fetch https://dl-cdn.alpinelinux.org/alpine/v3.15/main/x86_64/APKINDEX.tar.gz
WARNING: Ignoring https://dl-cdn.alpinelinux.org/alpine/v3.15/main: temporary error (try again later)
fetch https://dl-cdn.alpinelinux.org/alpine/v3.[15](https://github.com/terraforming-mars/terraforming-mars/runs/5193543084?check_suite_focus=true#step:3:15)/community/x86_64/APKINDEX.tar.gz
WARNING: Ignoring https://dl-cdn.alpinelinux.org/alpine/v3.15/community: temporary error (try again later)
ERROR: unable to select packages:
```

AndrewSav:
This is a temporary error [alpine got DDOSed](https://github.com/gliderlabs/docker-alpine/issues/563#issuecomment-1039795350), that should be fine now.

kberg:
Thanks!

PR #4012, opened 2022-01-29
Fix synchronicity errors with end games.

kberg:
The use of await in Postgres seems to have changed the lastSaveID, resulting in production games' last ID being lost forever. Rather than just fix that, I resolved a couple of other synchronicity errors, too.

1. Make sure that lastSaveId is correct by reading the last save ID from
the database rather than trusting the game.
2. Don't purge games until the current game is marked as finished, so it
doesn't accidentally get deleted.
3. Don't mark the game as finished until the game is truncated. It's
more efficient.

### DIFF
--- a/src/database/IDatabase.ts
+++ b/src/database/IDatabase.ts
@@ -135,7 +135,7 @@ export interface IDatabase {
      * A maintenance task on a single game to close it out upon its completion.
      * It will:
      *
-     * * Purge all saves between `(0, save_id]`.
+     * * Purge all saves between `(0, last save]`.
      * * Mark the game as finished.
      * * It also participates in purging abandoned solo games older
      *   than a given date range, regardless of the supplied `game_id`.
@@ -144,7 +144,7 @@ export interface IDatabase {
     // TODO(kberg): rename to represent that it's closing out
     // this game. Also consider not needing the save_id, and
     // also to make the maintenance behavior a first-class method.
-    cleanSaves(game_id: GameId, save_id: number): void;
+    cleanSaves(game_id: GameId): void;
 
     /**
      * A maintenance task that purges abandoned solo games older
--- a/src/database/PostgreSQL.ts
+++ b/src/database/PostgreSQL.ts
@@ -162,11 +162,11 @@ export class PostgreSQL implements IDatabase {
   }
 
   getMaxSaveId(game_id: GameId, cb: DbLoadCallback<number>): void {
-    this.client.query('SELECT MAX(save_id) FROM games WHERE game_id = $1', [game_id], (err: Error | null, res: QueryResult<any>) => {
+    this.client.query('SELECT MAX(save_id) as save_id FROM games WHERE game_id = $1', [game_id], (err: Error | null, res: QueryResult<any>) => {
       if (err) {
         return cb(err ?? undefined, undefined);
       }
-      cb(undefined, res.rows[0]);
+      cb(undefined, res.rows[0].save_id);
     });
   }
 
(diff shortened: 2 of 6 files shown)


kberg:
I have manually tested with SQLite. I must manually test with PostgreSQL.

src/database/IDatabase.ts
```diff
@@ -144,7 +144,7 @@ export interface IDatabase {
     // TODO(kberg): rename to represent that it's closing out
     // this game. Also consider not needing the save_id, and
     // also to make the maintenance behavior a first-class method.
-    cleanSaves(game_id: GameId, save_id: number): void;
+    cleanSaves(game_id: GameId): void;
```
bafolts, line 147: I don't see where we would have incremented this value to be one too high but I think this change makes sense. With this change it won't matter if on client side we are out of sync. The `cleanSaves` will still be doing what it was intended to do when called which was to delete everything but the last row. Could we consider in later PRs moving save_id to only exist in the database :thinking:  ? This puts us on that track.

src/database/PostgreSQL.ts
```diff
@@ -161,22 +161,36 @@ export class PostgreSQL implements IDatabase {
...
+  throwIf(err: any, condition: string) {
+    if (err) {
+      console.error('PostgreSQL', condition, err);
+      throw err;
+    }
+  }
+
+  cleanSaves(game_id: GameId): void {
+    this.getMaxSaveId(game_id, ((err, save_id) => {
+      this.throwIf(err, 'cleanSaves0');
```
bafolts, line 182: With how this is written `save_id` could also be `undefined`. I would also throw an error if `save_id` were `undefined`.

src/database/SQLite.ts
```diff
@@ -148,22 +148,33 @@ export class SQLite implements IDatabase {
...
-  cleanSaves(game_id: GameId, save_id: number): void {
-    // Purges isn't used yet
-    this.runQuietly('INSERT into purges (game_id, last_save_id) values (?, ?)', [game_id, save_id]);
-    // DELETE all saves except initial and last one
-    this.db.run('DELETE FROM games WHERE game_id = ? AND save_id < ? AND save_id > 0', [game_id, save_id], function(err: Error | null) {
+  getMaxSaveId(game_id: GameId, cb: DbLoadCallback<number>): void {
+    this.db.get('SELECT MAX(save_id) AS save_id FROM games WHERE game_id = ?', [game_id], (err: Error | null, row: { save_id: number; }) => {
       if (err) {
-        return console.warn(err.message);
+        return cb(err ?? undefined, undefined);
```
bafolts, line 154: Is the `err ?? undefined` needed here? If `err` is a truthly value it should be defined and not null.

bafolts (approved):
I would avoid using `cb(err ?? undefined, undefined)`. These should be able to be written as `cb(err, undefined)`. I would also check that `res.rows[0]` is defined and error if it is `undefined`. If it is `undefined` it will return an unexpected value to the caller with no error argument. The caller is assuming the value is defined if no error.

src/database/PostgreSQL.ts
kberg: > With how this is written `save_id` could also be `undefined`. I would also throw an error if `save_id` were `undefined`.

OK. Doing that now.

src/database/SQLite.ts
kberg: > Is the `err ?? undefined` needed here? If `err` is a truthly value it should be defined and not null.

Yes. err is of type `Error | null` and this converts it to `Error | undefined`.

PR #4014, opened 2022-01-31
add empty service worker prepping for better mobile notifications

bafolts:
For notifications on android the `Notification` constructor can't be used. This adds an empty service worker to allow us to use notifications on android. Once this merges if it doesn't seem to cause any issue I will put up the PR to show notifications.

https://developers.google.com/web/ilt/pwa/introduction-to-push-notifications

Eventually the `WaitingFor` component can be fully updated to utilize these push notifications versus the polling used today.

### DIFF
--- a/src/routes/ServeAsset.ts
+++ b/src/routes/ServeAsset.ts
@@ -103,6 +103,32 @@ export class ServeAsset extends Handler {
     });
   }
 
+  private toMainFile(urlPath: string, encodings: Set<Encoding>): { file?: string, encoding?: Encoding } {
+    let file = `build/${urlPath}`;
+    let encoding: Encoding | undefined;
+    if (encodings.has('br')) {
+      encoding = 'br';
+      file += '.br';
+    } else if (encodings.has('gzip')) {
+      encoding = 'gzip';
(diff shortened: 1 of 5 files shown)


src/routes/ServeAsset.ts
```diff
@@ -121,23 +147,10 @@ export class ServeAsset extends Handler {
...
-      // Return not-compressed .js files for development mode
-      if (!isProduction() && !this.fileApi.existsSync(file)) {
-        encoding = undefined;
-        file = `build/${urlPath}`;
-      }
+      return this.toMainFile(urlPath, encodings);
 
-      return {file, encoding};
+    case 'sw.js':
+      return this.toServiceWorkerFile(urlPath);
```
kberg, line 153: Why is this better than `return {``build/src/client/${urlPath}``;}` ?

kberg (approved):

src/routes/ServeAsset.ts
bafolts: If more complexity is added at a later time the function will already be in place. If a callstack is provided during a thrown error the function name could help with debugging as well.

Issue #4015, opened 2022-02-01
Docker build broken in recent commit


AndrewSav:
This commit https://github.com/terraforming-mars/terraforming-mars/commit/ff760848a78490b0b72b006fdde92e1e73e7c4ae has broken the docker build:

![log1](https://user-images.githubusercontent.com/658865/151893897-f9fcdc33-733c-4ba6-a744-935a735c2f5a.png)

This results in missing versioning info on the site:

![log2](https://user-images.githubusercontent.com/658865/151893924-c45729a8-7a5d-4ab1-aaa2-ab253e633333.png)

Can the be reverted? @kberg @jholm117

jholm117:
Sorry about that, should be able to just take .git/ out of the dockerignore. Ill test it out one sec

PR #4019, opened 2022-02-01
Update make_static_json.js

jholm117:
This will cause the build to fail if the git repo is not available. This would have caught the regression in #4015 .

On the flipside this won't allow you to build the project without git which may or may not be an issue.


kberg:
> This will cause the build to fail if the git repo is not available. This would have caught the regression in #4015 .
> 
> On the flipside this won't allow you to build the project without git which may or may not be an issue.

Yeah I'm not sure whether the case without git would be an issue, but now that the app doesn't work without it, this change seems reasonable.

kberg (approved):

PR #4030, opened 2022-02-02
Upgrade eslint.

kberg:
This also addresses a fair number of audit warnings we've been getting.

By upgrading eslint, two interesting behaviors kicked on:
1. Indentation became more strict. (I wondered why that didn't happen sooner.)
2. Vue components had to be multi-name. I see why that's the case, but I
   wasn't about to change that here.


kberg:
**Be sure to turn on `whitespace-only changes` when reviewing.** It will simplify your review a great deal, since the vast majority of the changes fix indentation.

kberg:
(BTW audit reduced vulnerabilities from 11 moderate to 7 moderate.)

bafolts (approved):

Argeento:
@kberg 

> Indentation became more strict. (I wondered why that didn't happen sooner.)

Because you rejected prettier. Twice :V

kberg:
> @kberg
> 
> > Indentation became more strict. (I wondered why that didn't happen sooner.)
> 
> Because you rejected prettier. Twice :V

What I meant was that I looked up all the eslint versions between the one we had and the one we were moving to, and could not see a bugfix or feature change that matched that.

PR #4031, opened 2022-02-02
Fixed Dockerfile to copy assets from built application, not from sources

zwrss:
I've introduced a bug during last Dockerfile refactoring. Assets are copied from clean source files and are missing some files (such as locales directory with translations). Fixed this by copying the assets directory from image with built application and not from clean sources.
Workaround over this bug is to build docker file after building app locally, because assets directory then contains everything docker image needs. Sadly this is why I did not find this bug in the previous PR.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,13 +45,13 @@ USER tfm
 
 COPY ["package.json", "package-lock.json", "./"]
 
-COPY assets ./assets
-
 # Copy dependencies from intermediate image
 COPY --from=installProd /usr/src/app/node_modules ./node_modules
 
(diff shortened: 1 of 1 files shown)


AndrewSav:
@zwrss can you give me an example of what's broken so I could test before and after?

zwrss:
> @zwrss can you give me an example of what's broken so I could test before and after?

Translations won't work.

AndrewSav:
LGTM

kberg (approved):

kberg:
Thank you!

PR #4033, opened 2022-02-04
Add Pathfinders card Specialized Settlement

kberg:
Rather than duplicate mining card behavior, I moved some of it to its own deferred action.


kberg:
There's more work to do. I'll reopen this later.

src/cards/pathfinders/SpecializedSettlement.ts
```diff
@@ -0,0 +1,113 @@
...
+        return undefined;
+      },
+    );
+  }
+
+  public produce(player: Player) {
+    this.defaultProduce(player);
+    if (this.bonusResource && this.bonusResource.length === 1) {
+      player.addProduction(this.bonusResource[0], 1, {log: true});
+    }
```
bafolts, line 93: What if `this.bonusResource.length > 1`?

bafolts (approved):

src/cards/pathfinders/SpecializedSettlement.ts
kberg: It's not going to be. And I'd like to move bonusResource back from Array<> to single element.

PR #4037, opened 2022-02-04
Move IAresData, Award and Milestone moodels to common.

kberg:


src/common/models/FundedAwardModel.ts
```diff
@@ -0,0 +1,15 @@
+
```
kberg, line 1: This didn't move, which means losing history. Sorry about that one.

kberg (commented):
This could disrupt active games upon deploy, but Ares isn't that common.

PR #4038, opened 2022-02-04
Move SpaceModel and MoonModel to common

kberg:
This required having to break up a couple of things, which is _good._


src/common/models/MoonModel.ts
```diff
@@ -0,0 +1,8 @@
+import {SpaceModel} from './SpaceModel';
```
kberg, line 1: Another file that could have been a move, but didn't make it, especially as the majority of the code disappeared.

PR #4040, opened 2022-02-04
try to show Notification on mobile devices

### LINKED ISSUE #3899: [BUG] Android Chrome notifications
The "new Notification" constructor is deprecated on chrome Android. 
Notifications should be performed with Web Workers

bafolts:
This is beginning of better notification support for mobile devices. I start with trying to use `Notification` as we do today to not break existing functionality. If that fails, which it will on at least android, we then try to use the service worker if it is available. I was able to see a notification with this on my android device.

Situations like this can be difficult to test with unit tests. Someone else testing on android would be appreciated.

This should close #3899


src/client/components/WaitingFor.vue
```diff
@@ -113,16 +113,25 @@ export default Vue.extend({
...
+                  // ok so the native Notification doesn't work which will happen
+                  // try to use the service worker if we can
+                  if (!window.isSecureContext || !navigator.serviceWorker) {
+                    return;
+                  }
+                  navigator.serviceWorker.ready.then((registration) => {
+                    registration.showNotification(notificationTitle, notificationOptions);
+                  }).catch((err) => {
+                    // avoid promise going uncaught
+                    console.warn('Failed to display notification', err);
```
kberg, line 133: Do you want to write which notification failed to show?

kberg (approved):

src/client/components/WaitingFor.vue
bafolts: I will update this logging and then merge. I don't expect this to be perfect but I think it is a good starting point to improve as we go. The updated logging will help us find this error message and know which one failed.

PR #4042, opened 2022-02-06
Use ~~ES2021~~ ES2019

kberg:
ES2018 seems to already be supported by default.

ES2019 support is pretty broad for Browsers, except for Function.prototype.toString, which is just not likely to be a problem
ES2020 is very well supported in modern browsers.
ES2021 has a couple of features not supported in Opera, but they're WeakRef and Promise.any. It is safe to assume those features won't be run on a browser.

See https://caniuse.com/sr_es10, https://caniuse.com/sr_es11, https://caniuse.com/sr_es12. Click "Feature Support List" for each of those.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -8,7 +8,8 @@
         // ES2020 is very well supported in modern browsers.
         // ES 2021 has a couple of features not supported in Opera, but they're WeakRef and Promise.any.
         // It is safe to assume those features won't be run on a browser.
-        "lib": ["dom", "es2021"],
+        // "lib": ["dom", "es2021"],
+        "lib": ["dom", "es2019"],
         "incremental": true,
         "module": "commonjs",
         "strict": true,
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -3,7 +3,12 @@
 {
     // Compiler options are mostly for the sub-configurations
     "compilerOptions": {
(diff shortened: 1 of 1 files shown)


alrusdi (approved):

tsconfig.json
```diff
@@ -3,7 +3,12 @@
 {
     // Compiler options are mostly for the sub-configurations
     "compilerOptions": {
-        "lib": ["dom"],
+        // ES2019 support is pretty broad for Browsers, except for Function.prototype.toString, which is just not
+        // likely to be a problem
+        // ES2020 is very well supported in modern browsers.
+        // ES 2021 has a couple of features not supported in Opera, but they're WeakRef and Promise.any.
+        // It is safe to assume those features won't be run on a browser.
+        "lib": ["dom", "es2021"],
```
bafolts, line 11: I am often wrong and could be here but I don't think this changes anything functionally. These entries in `lib` provide references to `es2021` types. Until a feature of `es2021` is used it may be worth holding off on this change.

kberg:
I was going to take the time to read the 2020 and 20201 features closely
and start using them mwahahahaha

tsconfig.json
kberg: OK I removed it.

PR #4043, opened 2022-02-06
Simon's CSS tweaks

kberg:


alrusdi (approved):

kberg:
@ssimeonoff on the whole this is a nice improvement! I have two comments before merging:

1. The three moon global parmeter icons also need to grow. I could look at this afterwards.
2. This makes things a little more difficult if we want to move these to the top of the card, on the same level as the tags. What are your thoughts about the requirements going there eventually?

kberg:
image

kberg:
I fixed the Moon global parameters:

image

kberg:
I spoke privately with @ssimeonoff and he convinced me it's not worth trying to cram the requirements up top. :sob:

PR #4045, opened 2022-02-06
Make PlayerTags simpler and more generic.

kberg:
This relies on es2019 (Object.fromEntries) which is well supported on
browsers. There's another PR to move our lib to 2021, which I think is a
good move. But one thing at a time.

image


src/client/components/overview/PlayerTags.vue
```diff
@@ -129,100 +97,103 @@ export default Vue.extend({
...
+      const vps = instance?.card.victoryPoints;
+      if (vps !== undefined && typeof(vps) !== 'number' && vps !== 'special' && vps.type !== 'resource') {
+        modifiers[vps.type].points += vps.points;
+        if (vps.per === 3) modifiers[vps.type].fraction = '⅓';
+      }
+    }
+
+    if (this.playerView.game.turmoil?.ruling === PartyName.UNITY &&
+      this.playerView.game.turmoil.politicalAgendas?.unity.policyId === 'up04') {
+      // space: discount += 2;
```
bafolts, line 135: This conditional doesn't seem to do anything.

tsconfig.json
```diff
@@ -3,7 +3,7 @@
 {
     // Compiler options are mostly for the sub-configurations
     "compilerOptions": {
-        "lib": ["dom"],
+        "lib": ["dom", "es2021"],
```
bafolts, line 6: Is this intentionally changing to `es2021` from `es2019`?

kberg: Sorry, removing.

src/client/components/overview/PlayerTags.vue
kberg: Thanks, good catch. Fixed.

kberg:
Ready for another pass.

bafolts (approved):

PR #4047, opened 2022-02-06
Move CardType to common/cards

kberg:
Another big file change because CardType is ubiquitous.


kberg (commented):
I reviewed all the files.

PR #4051, opened 2022-02-07
Move CardName to common/cards

kberg:


kberg:
Compared using `git diff HEAD HEAD~1 --unified=0` and even that had too much boilerplate. But the change is localized to the import changes and file move.

Issue #4063, opened 2022-02-08
Ants doesn't show the owner of the corp card.


kberg:
![image](https://user-images.githubusercontent.com/413481/153075215-509fbda8-d94c-49b8-8984-ba7d3dfc7566.png)

PR #4064, opened 2022-02-09
Add corporation card to list of selectable cards when finding owner. (Also fix BioSol rendering)

kberg:
Show card owner when selecting targets for Ants

One issue is that the corp card has this padding that I can't find.

image


kberg:
Fixes #4063

bafolts (approved):

PR #4066, opened 2022-02-10
throw only Error, remove crufty serialization

kberg:
Two small cleanups.

### DIFF
--- a/src/CardFinder.ts
+++ b/src/CardFinder.ts
@@ -41,16 +41,13 @@ export class CardFinder {
     return this.getCardByName(cardName, (manifest) => [manifest.preludeCards]);
   }
 
-  public cardsFromJSON(cards: Array<ICard | CardName>): Array<IProjectCard> {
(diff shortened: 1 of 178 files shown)


src/CardFinder.ts
```diff
@@ -41,16 +41,13 @@ export class CardFinder {
     return this.getCardByName(cardName, (manifest) => [manifest.preludeCards]);
   }
 
-  public cardsFromJSON(cards: Array<ICard | CardName>): Array<IProjectCard> {
```
kberg, line 44: This is outdated. Everything that serializes cards uses CardName.

src/cards/base/InterstellarColonyShip.ts
```diff
@@ -22,10 +21,7 @@ export class InterstellarColonyShip extends Card implements IProjectCard {
       },
     });
   }
-  public play(player: Player) {
-    if (player.getTagCount(Tags.SCIENCE) < 5) {
```
kberg, line 26: Interesting how I guess originally requirement checking was less trusted.

PR #4072, opened 2022-02-10
provide clone game id without list

bafolts:
I still want to add unit tests for the server side stuff here but I tested this through browser and it seems to work.

This changes the dropdown for selecting a cloneable game to an input box. The value provided is then verified before we call to create the game. Right now about 1 MB is downloaded querying for the list of cloneable games. The dropdown is nearly impossible to navigate to find a game id. Bandwidth will be decreased. Games will be cloned.


src/client/components/create/CreateGameForm.vue
```diff
@@ -905,11 +888,29 @@ export default Vue.extend({
...
+            if (response.ok) {
+              return response.json();
+            }
+            if (response.status === 404) {
+              return;
+            }
+            return response.text().then((res) => new Error(res));
+          });
+        if (gameData === undefined) {
+          alert(this.$t('Game id ' + component.clonedGameId + ' not found'));
```
kberg, line 903: Is it possible to be consistent about naming? Predefined game, cloneable game, game. Follow-up is possible, though let's get the service path right.

src/database/IDatabase.ts
```diff
@@ -93,6 +93,20 @@ export interface IDatabase {
      */
     getClonableGames(cb:(err: Error | undefined, allGames:Array<IGameData>)=> void) : void;
 
+    /**
+     * Load reference to game that can be cloned. Every game is cloneable,
```
kberg, line 97: Every _completed_ game is cloneable.

src/database/IDatabase.ts
```diff
@@ -93,6 +93,20 @@ export interface IDatabase {
...
+     * this just returns the original save of the game. However, if a game's
+     * original save is pruned, say, due to {@link deleteGameNbrSaves}, it won't
+     * exist.
```
kberg, line 100: s/it won't exist/this won't return it.

src/database/LocalFilesystem.ts
```diff
@@ -79,6 +79,19 @@ export class Localfilesystem implements IDatabase {
     });
   }
 
+  getClonableGameByGameId(gameId: GameId, cb: (err: Error | undefined, gameData: IGameData | undefined) => void) {
+    this.getGames((err, gameIds) => {
```
kberg, line 83: This can be simpler. Probably close to

```
try {
  const text = fs.readFileSync(this._historyFilename(gameId, 0));
  const serializedGame = JSON.parse(text) as SerializedGame;
  cb(...)
} catch(err) {
  cb(err, ...); or something.
}
```

tests/database/SQLite.spec.ts
```diff
@@ -103,4 +102,17 @@ describe('SQLite', () => {
       expect(saveIds).has.members([0, 3]);
     }, 1000);
   });
+
+  it('gets cloneable game by id', async () => {
```
kberg, line 106: Could you add a test where the game doesn't exist?

tests/utils/setup.ts
```diff
@@ -5,6 +5,7 @@ const FAKE_DATABASE: IDatabase = {
   cleanSaves: () => {},
   deleteGameNbrSaves: () => {},
   getClonableGames: () => {},
```
kberg, line 7: Is this line still necessary?

kberg (commented):
This is good stuff.

src/routes/ApiCloneableGames.ts
```diff
@@ -10,13 +10,22 @@ export class ApiCloneableGames extends Handler {
...
-    Database.getInstance().getClonableGames(function(err, allGames) {
+    const gameId = ctx.url.searchParams.get('id');
+    if (!gameId) {
+      ctx.route.badRequest(req, res, 'id parameter missing');
+      return;
+    }
+    Database.getInstance().getClonableGameByGameId(gameId, function(err, gameData) {
       if (err) {
-        console.warn('Could not load cloneable games: ', err);
+        console.warn('Could not load cloneable game: ', err);
```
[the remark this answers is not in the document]
bafolts: If the ID doesn't exist an `undefined` will be returned. The error would happen if there were some other database error.

src/client/components/create/CreateGameForm.vue
```diff
@@ -608,19 +600,10 @@ export default Vue.extend({
...
-    fetch('/api/clonablegames')
-      .then((response) => response.json())
-      .then(onSucces)
-      .catch((_) => alert('Unexpected server response'));
   },
   methods: {
-    downloadCurrentSettings() {
-      const serializedData = this.serializeSettings();
+    async downloadCurrentSettings() {
+      const serializedData = await this.serializeSettings();
```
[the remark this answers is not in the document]
bafolts: The `async/await` is added here for the introduced call to `fetch`. The `await fetch` requires an `async` up the call stack.

tests/utils/setup.ts
bafolts: The `getClonableGames` function hasn't been removed from `IDatabase` so this is still necessary.

src/routes/ApiCloneableGames.ts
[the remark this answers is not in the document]
kberg: > If the ID doesn't exist an `undefined` will be returned. The error would happen if there were some other database error.

:+1:

kberg:
I added some naming comments; the only one I kinda care about is the path naming. The others can happen later (I'm happy to do them.) Besides that this PR is good to go.

bafolts:
> I added some naming comments; the only one I kinda care about is the path naming. The others can happen later (I'm happy to do them.) Besides that this PR is good to go.

What should we name the path?

kberg:
> > I added some naming comments; the only one I kinda care about is the path naming. The others can happen later (I'm happy to do them.) Besides that this PR is good to go.
> 
> What should we name the path?

Hmm. How about `cloneablegame`? or `cloneable`?

bafolts:
This has the build failure on windows error most likely due to node 16 but I made the naming change to `cloneablegame` for the route.

kberg (approved):

kberg:
Thanks! This is a good change.

.github/workflows/main.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     - uses: actions/checkout@v1
     - uses: actions/setup-node@v1
       with:
-        node-version: '16.x'
+        node-version: '14.x'
```
kberg, line 27: Can you roll back this file? I've dropped windows from 2022 to 2019, and that does the trick.

kberg (changes requested):
Will go back to approve after main.yml is resolved.

bafolts:
@kberg I reverted the main.yml change and builds passed thanks to #4130

kberg (approved):

PR #4073, opened 2022-02-11
Fixed polish translation of Recyclon corp

zwrss:
Polish translation wrongly stated, that you should place a microbe and increase greenery production for every played building tag.


kberg (approved):
Thank you!

Issue #3820, opened 2021-11-30
[PF] Show Terraforming Control Station's discounts in the UI.


kberg:
Feedback from tester

> Terraforming control station discounts do not show up in UI, not sure if they were applied(were applied, they just don't show up in the tags part of the ui)

kberg:
This is fixed.

PR #4077, opened 2022-02-12
Make player tags show Terraforming Control Station. Add a test.

kberg:
Discover that Venus cards have fractional VPs and adjust.


src/client/components/overview/PlayerTags.vue
```diff
@@ -187,7 +182,15 @@ export default Vue.extend({
     },
     getVPs(tag: InterfaceTagsType) {
       const modifier = this.modifiers[tag];
-      return `${modifier.points}${modifier.fraction ?? ''}`;
+      const integer = Math.floor(modifier.points);
+      const fraction = modifier.points - integer;
+      let vulgarFraction = '';
+      if (fraction === 0.5) {
+        vulgarFraction = '½';
+      } else if (Math.abs(fraction - (1/3)) < 0.001) {
```
kberg, line 190: This is just the reality of matching to floating points.

src/models/ServerModel.ts
```diff
@@ -204,13 +204,14 @@ export class Server {
 
   public static getCorporationCard(player: Player): CardModel | undefined {
     if (player.corporationCard === undefined) return undefined;
+    const card = player.corporationCard;
     return {
-      name: player.corporationCard.name,
-      resources: player.getResourcesOnCard(player.corporationCard),
+      name: card.name,
```
kberg, line 209: Most of this local change is actually due to using a local parameter for `player.corporationCard`

kberg:
Fixes #3820

kberg:
image

src/client/components/overview/PlayerTags.vue
bafolts: Is this going to eventually need to display 1/4, 1/5, 1/6... 1/10? Using a for loop would avoid having to write this out for each fraction. Using `Number.EPSILON` makes sense to me versus using `0.001`. At that point it is self explanatory what is being checked.

kberg: Probably not. And if it does, we shouldn't rely on estimating fractional amounts.

bafolts (approved):

PR #4079, opened 2022-02-12
rename getPlayers to getPlayersInGenerationOrder

bafolts:
A function way back when was hastily named `getPlayers`. This function returns the players in generation order which was needed for early usage. By now most callers don't need the players in generation order. After this change will introduce a `getPlayers` that doesn't order the players and use it where appropriate.

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -350,7 +350,7 @@ export class Game implements ISerializable<SerializedGame> {
     // Initialize each player:
     // Give them their corporation cards, other cards, starting production,
     // handicaps.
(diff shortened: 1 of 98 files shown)


kberg:
> A function way back when was hastily named `getPlayers`. This function returns the players in generation order which was needed for early usage. By now most callers don't need the players in generation order. After this change will introduce a `getPlayers` that doesn't order the players and use it where appropriate.

What if there wasn't even a `getPlayers`? Just `game.players`?

kberg:
:clap: :clap: :clap:

PR #4080, opened 2022-02-12
Remove method getResourcesOnCard.

kberg:
Not much value in this method.

```
  return card.resourceCount;
}
```


kberg:
I'm a fan of getteres and setters, but this seemed like an idea that never needed to happen.

PR #4090, opened 2022-02-13
Minor visualization fixes.

kberg:


src/cards/promo/AsteroidDeflectionSystem.ts
```diff
@@ -31,7 +31,7 @@ export class AsteroidDeflectionSystem extends Card implements IActionCard, IProj
           b.action('REVEAL AND DISCARD the top card of the deck. If it has a space tag, add an asteroid here.', (eb) => {
             eb.empty().startAction.cards(1).asterix().nbsp.space({played}).colon().asteroids(1);
           }).br;
-          b.production((pb) => pb.minus().energy(1)).text('opponents may not remove your plants', Size.SMALL, true);
+          b.production((pb) => pb.minus().plants(1)).text('opponents may not remove your plants', Size.SMALL, true);
```
nwai90, line 34: Should this remain as `energy`? The viz here is referring to the production box for `Decrease your energy production 1 step`.

kberg: You're right. This is a mistake.

PR #4103, opened 2022-02-14
Fix rendering Iapetus card discount.

### LINKED ISSUE #4093: Visual on card discounts from the Iapetus colony are broken
image

This only shows on Christian's view.

On my view it shows something different, and also broken.

image

kberg:
By accident, Iapetus card discount was applying active player's discount
to all players. That's because I used `thisPlayer` instead of `player`.

Fixes #4093


kberg:
I will add a test after #4077 gets submitted.

nwai90 (approved):

PR #4119, opened 2022-02-16
Log the proper amount removed from a card and clean up removeResourceFrom.

kberg:


kberg:
This can be merged, the check complaint is still Windows.

PR #4123, opened 2022-02-17
Add log for Philares trigger

nwai90:


kberg (approved):
I made a small tweak to the log message.

PR #4124, opened 2022-02-18
Rewrite GamesOverview

kberg:
Using a child component which simplifies quite a lot, really.
Also not sending all the requests at once (what a thing!)
Add strong typing everywhere I can.

This whole thing would be faster if the API accepted a large number of
IDs at a time. That can be next time.


src/client/components/GamesOverview.vue
```diff
@@ -1,42 +1,48 @@
...
-                        <span v-for="player in game.players" class="player_name" :class="'player_bg_color_'+ player.color" :key="player.color">
-                            <a target="blank" :href="'/player?id=' + player.id">{{player.name}}</a>
-                        </span>
-                        <a target="blank" :href="'/spectator?id=' + game.spectatorId">Spectator</a>
-                        <span v-if="isGameRunning(game.phase)">is running</span><span v-else>has ended</span>
-                    </span>
-                </li>
-            </ul>
-        </div>
+  <div id="games-overview">
```
kberg, line 2: Best not to read this in the unified view.

nwai90 (approved):

PR #4126, opened 2022-02-19
Add Pathfinders corporation Habitat Martae

kberg:
I have thoughts about whether this should be included. I mean, it works,
but it means having to redefine how to count tags. It's kinda localized
to those cards that look at science tags, but is it worth having this
special-case? Should the new Player methods be applied everywhere?
Should there be a card.getTags() method that accomplishes this for you?


kberg:
image

bafolts:
This is an interesting card effect. It makes sense why this PR is doing what it does. I can't currently think of a more straight forward way to achieve this bug want to give it some more thought before approving.

kberg:
> This is an interesting card effect. It makes sense why this PR is doing what it does. I can't currently think of a more straight forward way to achieve this but want to give it some more thought before approving.

Yes, please.

src/cards/pathfinders/MartianRepository.ts
```diff
@@ -39,7 +39,7 @@ export class MartianRepository extends Card implements IProjectCard {
   public override resourceCount = 0;
 
   public onCardPlayed(player: Player, card: ICard) {
-    const qty = card.tags.filter((tag) => tag === Tags.SCIENCE || tag === Tags.MARS).length;
+    const qty = player.cardTagCount(card, Tags.SCIENCE) + player.cardTagCount(card, Tags. MARS);
```
bafolts, line 42: Is this count what is expected? If a card has one science and one mars tag this will return 3 for Habitat Martae. The science tag would count as a science tag and a mars tag for 2 and then 1 for the mars tag that exists.

src/Game.ts
```diff
@@ -1504,7 +1504,7 @@ export class Game implements ISerializable<SerializedGame> {
         return player;
       }
     }
-    throw new Error('No player has played requested card');
+    return undefined;
```
bafolts, line 1507: I think I understood this yesterday but why does this have to change for `Habitat Martae`? Throwing here is helpful for failing fast when the player is not found. Most places we call this we expect to get a result. Is there a new place we are calling in this change where we don't expect to get a return value?

kberg: Look at src/cards/pathfinders/CollegiumCopernicus.ts. Failing is now considered a valid workable case and not just a system error. I would need to change that behavior to try/catch. And that's fine, but it's not the pattern we use in this app.

src/cards/pathfinders/MartianRepository.ts
kberg: Ah, I think you're right, now that I think about it.

kberg (commented):
Hang back on approval. I want to write some more thoughts.

src/Game.ts
bafolts: Can you link to where in `CollegiumCopernicus` this `getCardPlayer` is called? I don't see a reference to `getCardPlayer` in that file.

kberg: Sorry, you're right. I don't see a specific place where this was necessary.

Thinking back, I bet I was doing because `owner?.` is pretty small, and just as safe. Perhaps it's hiding an error, but, well, I don't know.

kberg: If you think restoring the original behavior is better, I will. But knowing so much about when this is called and why, this is a pretty safe change.

bafolts:
> Hang back on approval. I want to write some more thoughts.

Is this ready for approval?

src/Game.ts
kberg: > > Hang back on approval. I want to write some more thoughts.
> 
> Is this ready for approval?

It is now.

bafolts (approved):

kberg:
Thank you!

PR #4131, opened 2022-02-19
Upgrade corejs to eliminate unsupported version of corejs.

kberg:


bafolts:
It seems the corejs dependency was picked up for `mochapack`. Should this be a PR up for https://github.com/sysgears/mochapack/blob/master/package.json?

kberg:
> It seems the corejs dependency was picked up for `mochapack`. Should this be a PR up for https://github.com/sysgears/mochapack/blob/master/package.json?

Yeah. Perhaps closing this one makes sense.

PR #4137, opened 2022-02-21
conversion of some getPlayers

bafolts:
Converts some `getPlayersInGenerationOrder` calls to `getPlayers`.

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -1475,6 +1475,10 @@ export class Game implements ISerializable<SerializedGame> {
     space.player = undefined;
   }
 
+  public getPlayers(): ReadonlyArray<Player> {
+    return this.players;
+  }
+
   // Players returned in play order starting with first player this generation.
   public getPlayersInGenerationOrder(): Array<Player> {
     // We always return them in turn order
@@ -1529,7 +1533,7 @@ export class Game implements ISerializable<SerializedGame> {
   public someoneCanHaveProductionReduced(resource: Resources, minQuantity: number = 1): boolean {
     // in soloMode you don't have to decrease resources
     if (this.isSoloMode()) return true;
-    return this.getPlayersInGenerationOrder().some((p) => {
+    return this.getPlayers().some((p) => {
       if (p.getProduction(resource) < minQuantity) return false;
       // The pathfindersExpansion test is just an optimization for non-Pathfinders games.
       if (this.gameOptions.pathfindersExpansion && p.cardIsInEffect(CardName.PRIVATE_SECURITY)) return false;
--- a/src/models/ServerModel.ts
+++ b/src/models/ServerModel.ts
@@ -147,14 +147,12 @@ export class Server {
       const claimed = claimedMilestones.find(
         (m) => m.milestone.name === milestone.name,
       );
-      const scores: Array<IMilestoneScore> = [];
+      let scores: Array<IMilestoneScore> = [];
       if (claimed === undefined && claimedMilestones.length < 3) {
-        game.getPlayersInGenerationOrder().forEach((player) => {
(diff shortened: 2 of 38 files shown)


src/models/ServerModel.ts
```diff
@@ -147,14 +147,12 @@ export class Server {
       const claimed = claimedMilestones.find(
         (m) => m.milestone.name === milestone.name,
       );
-      const scores: Array<IMilestoneScore> = [];
+      let scores: Array<IMilestoneScore> = [];
```
kberg, line 150: I wonder if this one changes rendering on the UI. No action for the moment. This is mostly harmless if it does.

src/pathfinders/PathfindersExpansion.ts
```diff
@@ -128,14 +128,14 @@ export class PathfindersExpansion {
...
-          game.getPlayersInGenerationOrder().forEach((p) => {
+          game.getPlayers().forEach((p) => {
             PathfindersExpansion.grant(reward, p, tag);
           });
         });
         if (rewards.mostTags.length > 0) {
           const players = PathfindersExpansion.playersWithMostTags(
             tag,
-            game.getPlayersInGenerationOrder(),
+            game.getPlayers().slice(),
```
kberg, line 138: I don't think this needs a slice. Does it?

src/turmoil/globalEvents/Election.ts
```diff
@@ -29,13 +29,14 @@ export class Election extends GlobalEvent implements IGlobalEvent {
   public resolve(game: Game, turmoil: Turmoil) {
     // Solo
     if (game.isSoloMode()) {
-      if (this.getScore(game.getPlayersInGenerationOrder()[0], turmoil, game) >= 10) {
-        game.getPlayersInGenerationOrder()[0].increaseTerraformRatingSteps(2, {log: true});
-      } else if (this.getScore(game.getPlayersInGenerationOrder()[0], turmoil, game) >= 1) {
-        game.getPlayersInGenerationOrder()[0].increaseTerraformRatingSteps(1, {log: true});
+      const player = game.getPlayers()[0];
```
kberg, line 32: See proposed change.

Also, is there a bug here? >= 1 should be >= 5.

kberg (approved):

kberg:
This is great.

src/Game.ts
```diff
@@ -1475,6 +1475,10 @@ export class Game implements ISerializable<SerializedGame> {
     space.player = undefined;
   }
 
+  public getPlayers(): ReadonlyArray<Player> {
```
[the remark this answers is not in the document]
bafolts: It would preserve it's Readonly-ness but we would have to update any places which were modifying the order. Would most likely be unit tests or calls to `sort` not expecting that to change the order.

src/models/ServerModel.ts
bafolts: Some of these may change the order items were displayed on the user interface. Since the original `getPlayers` was always returning the players in generation order some players may have come to rely on this order.

src/pathfinders/PathfindersExpansion.ts
bafolts: Since `getPlayers()` is marked `Readonly` I think I added the `slice` because something must be trying to modify the array. The `ReadonlyArray` protects against anything trying to mutate the array.

src/turmoil/globalEvents/Election.ts
bafolts: This is unrelated to my change and I am not familiar enough with turmoil to know if this should be `>= 5`. It does seem the description indicates it should be 5 but the description could be wrong.

PR #4138, opened 2022-02-21
ignore tag names from before linting rule

bafolts:
Adds the tags used which are not using `multi-word-component-names`. These tags were added prior to the linting rule being enforced. Avoids having warnings on every build.


kberg (approved):
Nice compromise. Thank you.

PR #4139, opened 2022-02-22
Move card components around to decompose them into interfaces. This is key to separating out client/server.

kberg:
This is not a small change. However, fortunately it only changes card rendering which is pretty easy to verify on visual inspection. A couple of pieces could maybe be split out, I suppose. Separating client and server perhaps?


src/cards/render/CardRenderer.ts
```diff
@@ -1,53 +1,53 @@
...
+class CardRenderEffect implements ICardRenderEffect {
+  public readonly is = 'effect';
+  constructor(public rows: Array<Array<ItemType>>) {}
 
-  public static override builder(f: (builder: EffectBuilder) => void): CardRenderEffect {
+  public static builder(f: (builder: EffectBuilder) => void): CardRenderEffect {
     const builder = new EffectBuilder();
     f(builder);
-    return builder.build();
+    return builder.build().validate();
```
bafolts, line 47: This wasn't calling to `validate` prior to this change. Is this accidental?

bafolts (approved):

src/cards/render/CardRenderer.ts
kberg: It is not. Validate was called on each getter, which was just too much. Looking  at this now, having validate() return `this` is silly, but I'm gonna go ahead and commit it.

PR #4141, opened 2022-02-22
Prototype admin screen?

kberg:
Is this worth pursuing?

### DIFF
--- a/assets/admin.html
+++ b/assets/admin.html
@@ -1,0 +1,41 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <link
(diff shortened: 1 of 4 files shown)


assets/admin.html
```diff
@@ -0,0 +1,41 @@
+<!DOCTYPE html>
+<html>
+    <head>
```
bafolts, line 3: This is worth pursuing but I would use vue.js versus straight HTML.

kberg: For now it would be just this one component. What would be useful after this?

kberg: I'm beginning to see other uses. I will proceed with this but move it to Vue shortly.

PR #4142, opened 2022-02-22
Fix PORT env var name (for heroku setup)

nyddogghr:
Hi,
I've noticed that when deploying the latest version on heroku the application no longer binded to $PORT (crashed).
Maybe a dependency update made the variable name case sensitive.
This change is all that was required to work again on my own heroku instance.


kberg (approved):
Thank you!

PR #4143, opened 2022-02-23
add titles to Sidebar items

### LINKED ISSUE #4096: Provide hover text over the sidebar.
image

I know what most of those are, but not the bottom one. For instance.

bafolts:
Adds title to each sidebar item. This can close #4096

### DIFF
--- a/src/client/components/Sidebar.vue
+++ b/src/client/components/Sidebar.vue
@@ -1,10 +1,10 @@
 <template>
 <div :class="'sidebar_cont sidebar '+getSideBarClass()">
-  <div class="tm">
+  <div class="tm" :title="$t('Generation Marker')">
     <div class="gen-text">GEN</div>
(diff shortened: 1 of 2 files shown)


src/client/components/Sidebar.vue
```diff
@@ -14,32 +14,32 @@
     <global-parameter-value v-if="gameOptions.venusNextExtension" :param="this.globalParameter.VENUS" :value="this.venus"></global-parameter-value>
     <MoonGlobalParameterValue v-if="gameOptions.moonExpansion" :moonData="this.moonData"></MoonGlobalParameterValue>
```
kberg, line 15: This one can be "Moon colony, road, mining rates" (or whatever the order is in MoonGlobalParameterValue.) Mind doing that as a follow-up?

kberg (approved):

PR #4147, opened 2022-02-25
Rewrite preferences code. Make new defaults.

### LINKED ISSUE #3259: Four (or so) of the *default* values for local preferences are "backwards"
Every time I teach a new player, the first thing we have to do is go into their gear icon to turn on a few settings:

- Learner mode
- Enable sounds
- Magnify cards on hover
- Show in-game alerts

Should these be the defaults for new players?

kberg:
Enables the following preferences as defaults:
* Learner mode
* Enable sounds
* Magnify cards on hover
* Show in-game alerts

The rewrite is probably a little more code, but it relies more on
natural typing and simpler. Though it could use a little more work. I'd
like to remove PreferencesManager.INSTANCE.values() everywhere.

This fixes #3259


kberg:
@bafolts can you take a look at this please? I need to add more tests, but I really want to know what you think of this change.

src/client/components/PreferencesDialog.vue
```diff
@@ -42,52 +30,25 @@ export default Vue.extend({
...
-          } else {
-            strVal = this.$data[k] ? '1' : '0';
-          }
-          PreferencesManager.save(k, strVal);
-          PreferencesManager.preferencesValues.set(k, this.$data[k]);
-          this.setPreferencesCSS(this.$data[k], k);
-        }
+    updatePreferences() {
+      for (const k of Object.keys(this.preferencesManager.values) as Array<Preference>) {
+        const val = this[k];
```
bafolts, line 35:
```suggestion
        const val = this.preferencesManager.values[k];
```

As written this is assuming that `this.data` is set to `this.preferencesManager.values()`.

kberg: That's right. If something changes this, well, the whole thing would need to change a little bit.

kberg:
There are now tests, and I moved alot around -- again. But it's good now.

src/client/components/PlayerHome.vue
```diff
@@ -283,22 +283,23 @@ class TerraformedAlertDialog {
...
+      showActiveCards: !preferences.hide_active_cards,
+      showAutomatedCards: !preferences.hide_automated_cards,
+      showEventCards: !preferences.hide_event_cards,
       hideTiles: false,
     };
   },
   watch: {
     hide_active_cards() {
-      PreferencesManager.save('hide_active_cards', !this.showActiveCards);
+      PreferencesManager.INSTANCE.set('hide_active_cards', !this.showActiveCards);
```
bafolts, line 296: There is a getPreferences now. A setPreference may work here.

bafolts (approved):

PR #4154, opened 2022-02-26
Fix bug with Floater Technology

nwai90:
Bug: Floater Technology action should be usable (as a stall action) even with 0 targets available


kberg (approved):
Looks reasonable! Thanks.

PR #4164, opened 2022-02-27
Recent Colonies change broke ColoniesFilter when selecting distinct colonies.

kberg:
Also, whoops, Callisto is listed twice.

A test will come later. I'd like to patch.


kberg:
Manually verified.

nwai90 (approved):

PR #4166, opened 2022-02-27
Add Colonies to the debug view

kberg:
image


kberg:
Someone with CSS chops could make those stack in a row, but not me today.

bafolts (approved):

PR #4167, opened 2022-02-27
Modernize colony vue

kberg:
This depends on #4166 not just because it further changes the code, but it also is how I can spot-check all the colonies to verify they haven't changed.

### DIFF
--- a/src/client/components/DebugUI.vue
+++ b/src/client/components/DebugUI.vue
@@ -123,8 +123,20 @@
             <section class="debug-ui-cards-list">
               <h2>Global Events</h2>
               <div class="cardbox" v-for="globalEventName in getAllGlobalEvents()" :key="globalEventName">
-                <!-- <global-event :globalEvent="getGlobalEvent(globalEventName)" type="prior" :showIcons="false"></global-event> -->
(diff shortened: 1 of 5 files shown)


src/client/components/DebugUI.vue
```diff
@@ -155,8 +171,20 @@ const MODULE_COMMUNITY = '*';
 const MODULE_PROMO = 'r';
 const MODULE_ARES = 'a';
 const MODULE_MOON = 'm';
+const MODULE_PATHFINDERS = 'P';
 
-const ALL_MODULES = `${MODULE_BASE}${MODULE_CORP}${MODULE_PRELUDE}${MODULE_VENUS}${MODULE_COLONIES}${MODULE_TURMOIL}${MODULE_COMMUNITY}${MODULE_PROMO}${MODULE_ARES}${MODULE_MOON}`;
+const ALL_MODULES =
```
Argeento, line 176:
```suggestion
const ALL_MODULES = [
  MODULE_X, 
  MODULE_Y,
  ...
].join('')
```

Argeento (approved):

src/client/components/DebugUI.vue
kberg: No difference, but thank you.

bafolts (approved):

PR #4171, opened 2022-02-28
Add promo card New Partner.

kberg:
Comes from https://github.com/nwai90/terraforming-mars/commit/95b8557d3c227e38b17deba2bdfae5c2e55a7f28

### DIFF
--- a/src/cards/promo/NewPartner.ts
+++ b/src/cards/promo/NewPartner.ts
@@ -1,0 +1,50 @@
+import {Player} from '../../Player';
+import {PreludeCard} from '../prelude/PreludeCard';
+import {CardName} from '../../common/cards/CardName';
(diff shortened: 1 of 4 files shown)


kberg:
I ran a further test manually, which didn't do what I expected, but is correct.

I set up conditions where I chose two preludes: New Partner and another one that had an MC cost. I could afford both of those. But by playing New Partner, I selected another Prelude which reduced MC costs. The two reductions were more than I had.

I expected the app to be unhappy and conk out. Instead it did the right thing and didn't let me play the second prelude. But, I think the game's log is pretty unclear about what it didn't allow.

kberg:
image

nwai90 (approved):

PR #4182, opened 2022-03-04
Vastitas borealis

kberg:


src/awards/Edgedancer.ts
```diff
@@ -0,0 +1,25 @@
...
+  public getScore(player: Player): number {
+    return player.game.board.spaces
+      .filter((space) => space.player !== undefined &&
+        space.player === player &&
+        space.tile !== undefined &&
+        isHazardTileType(space.tile.tileType) === false &&
+        this.isOnEdge(space.x, space.y)).length;
+  }
+
+  private isOnEdge(x: number, y: number): boolean {
```
kberg, line 18: This is great.

src/boards/VastitasBorealisBoard.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  public getNoctisCitySpaceIds(): Array<string> {
+    return [];
+  }
+
+  public static handleBonusTemperatureFromTilePlacement(game: Game, player: Player, space: ISpace) {
+    if (space.id === SpaceName.VASTITAS_BOREALIS_NORTH_POLE && game.getTemperature() < MAX_TEMPERATURE && game.gameOptions.boardName === BoardName.VASTITAS_BOREALIS) {
+      if (player.color !== Color.NEUTRAL) {
+        game.defer(new DeferredAction(player, () => game.increaseTemperature(player, 1)));
+        game.defer(new SelectHowToPayDeferred(player, VASTITAS_BOREALIS_BONUS_TEMPERATURE_COST, {title: 'Select how to pay for placement bonus temperature'}));
```
kberg, line 97: Should payment go first, in the case of a Reds tax?

src/milestones/Irrigator.ts
```diff
@@ -0,0 +1,22 @@
...
+export class Irrigator implements IMilestone {
+  public name: string = 'Irrigator';
+  public description: string = 'Have at least 4 tiles adjacent to oceans';
+
+  public getScore(player: Player): number {
+    return player.game.board.spaces.filter((space) =>
+      space.player === player &&
+        space.tile !== undefined &&
+        isHazardTileType(space.tile.tileType) === false &&
+        player.game.board.getAdjacentSpaces(space).some((space) => Board.isOceanSpace(space)),
```
kberg, line 15: Perfect. Thank you for making it compatible with Pathfinders.

kberg (commented):
I can't approve even though @nwai90 wrote it since I added the PR. But I approve.

src/boards/VastitasBorealisBoard.ts
nwai90: Not sure if it makes a difference here. There would be some Reds-related edge cases either way (e.g. bonus ocean at 0 degrees)

nwai90 (approved):

src/boards/VastitasBorealisBoard.ts
kberg: I just realized post-merge: where is `handleBonusTemperatureFromTilePlacement` called?

kberg: Also the reds-related edge cases are handleable in the filter call, though that introduces a new problem, because of all the things that come with predicting if you can afford next sub-action costs due to placements.

But it would be to add the `tr` parameter here:

```
  private filterVastitasBorealis(player: Player, spaces: Array<ISpace>) {
    return player.canAfford(VASTITAS_BOREALIS_BONUS_TEMPERATURE_COST, {tr: {temperature: 1}}) ? spaces : spaces.filter((space) => space.id !== SpaceName.VASTITAS_BOREALIS_NORTH_POLE);
  }
```

PR #4214, opened 2022-03-17
Fixed polish translation for Directed Impactors

zwrss:
Old translation stated, that asteroid resource will be placed on this card. I also translated card name itself as it is in paper version.


Argeento (approved):

kberg (approved):
Thank you!

Issue #4212, opened 2022-03-12
[Pathfinder] Crescent Research Association Moon discount not shown


HenrikSteenbock:
The corporation "Crescent Research Association" has a moon discount of 1/moon tag played, but that discount is not displayed:
![grafik](https://user-images.githubusercontent.com/44550180/158005096-c646eb1e-e14e-487b-bef0-deb4eb8f58bc.png)

For reference, the corporation:
![grafik](https://user-images.githubusercontent.com/44550180/158005155-9a2f78d3-d822-4c26-a636-4481ffee4ff4.png)

This occured on the public server.

kberg:
Thanks for this report! I have just sent out a fix.

PR #4225, opened 2022-03-21
Show discount for Crescent Research Association.

kberg:
Since this is dynamic, it's not as easy as creating an attribute on ICard, though i could extend cardDiscount to also have type that accepts a player.

image


kberg:
Fixes #4212

PR #4227, opened 2022-03-21
Centralize most card discount processing into one method.

kberg:
This relies on the cardDiscount card property.


src/cards/pathfinders/TerraformingControlStation.ts
```diff
@@ -31,13 +30,6 @@ export class TerraformingControlStation extends Card implements IProjectCard {
     });
   }
 
-  public getCardDiscount(_player: Player, card: ICard) {
-    if (card.tags.includes(Tags.VENUS) || card.tags.includes(Tags.MARS)) {
```
bafolts, line 35: Does this card need to override? I think this change would end up returning more than 2 if a card had both mars and venus tags.

src/cards/base/EarthOffice.ts
```diff
@@ -27,7 +27,7 @@ export class EarthOffice extends Card implements IProjectCard {
     });
   }
 
-  public getCardDiscount(_player: Player, card: IProjectCard) {
+  public override getCardDiscount(_player: Player, card: IProjectCard) {
```
bafolts, line 30: Does this have to override? The base function seems like it would calculate the same value.

src/cards/base/MassConverter.ts
```diff
@@ -33,12 +33,6 @@ export class MassConverter extends Card implements IProjectCard {
   }
 
 
-  public getCardDiscount(_player: Player, card: IProjectCard) {
-    if (card.tags.includes(Tags.SPACE)) {
```
bafolts, line 37: Is the base going to award 2 per space tag on a card now versus only 2 if the card has a space tag?

src/cards/base/EarthOffice.ts
kberg: Not anymore. Removed.

src/cards/base/MassConverter.ts
kberg: Ah, you're right.

src/cards/pathfinders/TerraformingControlStation.ts
kberg: Yeah, but that's correct here. Each tag on a card contributes to the discount. "when you play a tag" seems to imply a discount per tag.

src/cards/base/MassConverter.ts
kberg: Just published a patch.

bafolts (approved):

Issue #4228, opened 2022-03-21
Loading configuration file with custom corporations and colonies does not work


andystager:
The data saved in 
"customCorporationsList": [
and
"customColoniesList": [

is not loading when importing my game configuration file. Here is my entire configuration as saved:

`{
    "players": [
        {
            "index": 1,
            "name": "You",
            "color": "red",
            "beginner": false,
            "handicap": 0,
            "first": true
        }
    ],
    "corporateEra": true,
    "prelude": true,
    "draftVariant": true,
    "showOtherPlayersVP": false,
    "venusNext": true,
    "colonies": true,
    "turmoil": false,
    "customCorporationsList": [
        "CrediCor",
        "EcoLine",
        "Helion",
        "Interplanetary Cinematics",
        "Inventrix",
        "Mining Guild",
        "PhoboLog",
        "Tharsis Republic",
        "Thorgate",
        "United Nations Mars Initiative",
        "Saturn Systems",
        "Teractor",
        "Cheung Shing MARS",
        "Point Luna",
        "Robinson Industries",
        "Valley Trust",
        "Aphrodite",
        "Celestic",
        "Manutech",
        "Morning Star Inc.",
        "Viron",
        "Aridor",
        "Arklight",
        "Polyphemos",
        "Poseidon",
        "Stormcraft Incorporated",
        "Astrodrill",
        "Factorum",
        "Pharmacy Union",
        "Recyclon",
        "Project Workshop",
        "Midas"
    ],
    "customColoniesList": [
        "Callisto",
        "Ceres",
        "Enceladus",
        "Europa",
        "Ganymede",
        "Io",
        "Luna",
        "Pluto",
        "Titan",
        "Triton",
        "Iapetus",
        "Mercury",
        "Leavitt",
        "Venus"
    ],
    "cardsBlackList": [],
    "board": "random official",
    "seed": 0.23882166678202021,
    "solarPhaseOption": true,
    "promoCardsOption": true,
    "communityCardsOption": true,
    "aresExtension": false,
    "politicalAgendasExtension": "Standard",
    "moonExpansion": false,
    "pathfindersExpansion": false,
    "undoOption": true,
    "showTimers": true,
    "fastModeOption": false,
    "removeNegativeGlobalEventsOption": false,
    "includeVenusMA": true,
    "startingCorporations": "6",
    "soloTR": true,
    "initialDraft": false,
    "randomMA": "No randomization",
    "shuffleMapOption": false,
[...]

PR #4229, opened 2022-03-22
Restore ColoniesFilter.updateCorporationsByName.

kberg:
This is a case where the method, seemingly unused, was deleted. However,
it was used in CreateGameForm, which references it with `(refs.xxx as
any)`. Getting rid of tne 'any' typing would be a big help.


kberg:
Fixes #4228

nwai90 (approved):

PR #4236, opened 2022-03-25
Introduce IColonyMetadata

kberg:
This reduces the API and also makes it easy to share that metadata with the client without the rest of the code.


kberg:
When reviewing, here's the guidance:

1. Look at how each colony now uses the same pattern as cards do when passing attributes. That's the only change to every colony.
2. Many colony fields are now only accessible via the `metadata` sub-field, fwiw.
Those two things in isolation should help interpret what is going on.

src/colonies/Colony.ts
```diff
@@ -26,28 +25,26 @@ import {Turmoil} from '../turmoil/Turmoil';
...
-    public abstract readonly tradeType: ColonyBenefit;
-    public readonly tradeQuantity: Array<number> = [1, 1, 1, 1, 1, 1, 1];
-    public readonly tradeResource?: Resources | Array<Resources>;
-    public abstract readonly colonyBonusType: ColonyBenefit;
-    public readonly colonyBonusQuantity: number = 1;
-    public readonly colonyBonusResource?: Resources;
-    public readonly shouldIncreaseTrack: ShouldIncreaseTrack = ShouldIncreaseTrack.YES;
+    public metadata: IColonyMetadata;
+
+    public constructor(metadata: IInputColonyMetadata) {
```
bafolts, line 41: I think this constructor may only need protected access level and not public.

src/common/colonies/IColonyMetadata.ts
```diff
@@ -0,0 +1,32 @@
...
+  readonly colonyBonusQuantity: number; // Default is 1
+  readonly colonyBonusResource?: Resources;
+  readonly shouldIncreaseTrack: ShouldIncreaseTrack // Default is ShouldIncreaseTrack.YES;
+}
+
+export type IInputColonyMetadata = Omit<IColonyMetadata, 'buildQuantity' |'tradeQuantity' | 'colonyBonusQuantity' | 'shouldIncreaseTrack'> & Partial<IColonyMetadata>;
+
+export function colonyMetadata(partial: IInputColonyMetadata): IColonyMetadata {
+  return {
+    buildQuantity: [1, 1, 1],
```
bafolts, line 26: This _might_ make a new `Array` in memory each time this function is called. As these properties are _readonly_ creating a constant for these would avoid having to make new instances and store in memory. If the `Array` is _readonly` I would also use `ReadonlyArray`. As written only the _properties_ are _readonly_ but the values can be modified in some cases.

bafolts (approved):

src/colonies/Colony.ts
kberg: Fixed

src/common/colonies/IColonyMetadata.ts
kberg: Fixed

PR #4243, opened 2022-03-26
Create a bridge between ICard and other card types and the client.

kberg:
This is a big step forward in separating the client and server. It's
only this simple because of the prior months of separating other
components.


Argeento:
Awesome!

src/client/components/SelectInitialCards.vue
```diff
@@ -77,14 +76,12 @@ export default (Vue as WithRefs<Refs>).extend({
...
-        const card = cam.card as PreludeCard;
-        result += card.startingMegaCredits ?? 0;
+        result += cam.card.startingMegaCredits ?? 0;
 
         switch (this.selectedCorporation?.name) {
         // For each step you increase the production of a resource ... you also gain that resource.
         case CardName.MANUTECH:
-          const productionBox = card?.productionBox;
-          result += productionBox.megacredits;
+          result += cam.card.productionBox?.megacredits || 0;
```
bafolts, line 84: Should this also use `?? 0`?

bafolts (approved):

src/client/components/SelectInitialCards.vue
kberg: Seems reasonable.

PR #4251, opened 2022-03-28
Rewrite code for giving TR.

kberg:
Current solution has raiseSteps(count) call raiseByOne() `count` times.
That raiseByOne() checks whether the player can afford each bump if reds
are ruling.

This creates a weird situation where, first of all, the player might be
faced with multiple deferred actions when only one should suffice. Also,
this is more technically correct, because if a player cannot pay all
reds costs, it couldn't implement the whole action. According to the
FAQ 1.7:

> To clarify the rulebook text further, you cannot play a card that raises
> terraforming in any way if you cannot pay the 3 MC cost per TR on that card
> while the Reds are ruling.
> This does not simply mean you can play the card in question but just ignore
> the TR boost if you lack the funds.

This never occurs in reality because most of the code already out there
handles all testing whether something can go up a number of steps. This
is just a form of defense-in-depth.

This also reduces the code repetition (which will be useful in an
upcoming PR for the pathfinders corp Aurorai.)


src/deferredActions/SelectHowToPayDeferred.ts
```diff
@@ -31,9 +31,7 @@ export class SelectHowToPayDeferred implements DeferredAction {
   public execute() {
     if (this.mustPayWithMegacredits()) {
       this.player.deductResource(Resources.MEGACREDITS, this.amount);
-      if (this.options.afterPay !== undefined) {
-        this.options.afterPay();
-      }
+      this.options.afterPay?.();
```
bafolts, line 34: Used this 💎 the other day with the `functionName?.()` in another project.

bafolts (approved):

PR #4252, opened 2022-03-29
Render card data as a JSON file and use it to load cards.

kberg:
This PR is optional based on whether 'experimental ui' is on. Once
people get a chance to try this without any issues and the flag is
removed, the front-end will no longer rely on ICard.

You can see this in action on terraforming-mars-kberg.herokuapp.com.
On that site you don't even need to set a flag, it's on by default.


src/tsconfig.json
```diff
@@ -6,5 +6,6 @@
     "include": [
         "**/*.ts",
         "genfiles/*.json",
-    ]
+    ],
+    "exclude": ["client/**/*.ts"],
```
kberg, line 10: Looks as though these .ts files get properly compiled by webpack.

package.json
```diff
@@ -69,13 +69,14 @@
   "scripts": {
     "start": "node build/src/server.js",
     "build": "npm run make:static && npm run build:server && npm run build:client",
-    "build:client": "cross-env NODE_ENV=production webpack",
+    "build:client": "npm run make:cards && cross-env NODE_ENV=production webpack",
```
kberg, line 72: This PR does change the build process. Now:

1. The server code, which contains code information, has to be built.
2. Using that server code, run it to get information to render the card as JSON.
3. Only then will the front-end compile.

Argeento: `cross-env NODE_ENV=production` probably should be the first command in this line

src/tsconfig.json
bafolts: This will decrease the build time marginally.

bafolts (approved):

PR #4256, opened 2022-03-30
Only check for multiple tags when there are many tags.

kberg:
Besides probably reducing computation time, this also fixes an edge case
with fan expansion cards that implement tag substitutions.

This will not fix an edge case that doesn't exist yet, where a card
requires a Moon and 2 Earth tags, and the player has played Earth Embassy.

### DIFF
--- a/src/cards/CardRequirements.ts
+++ b/src/cards/CardRequirements.ts
@@ -32,7 +32,7 @@ export class CardRequirements implements ICardRequirements {
         tags.push((requirement as TagCardRequirement).tag);
       }
     });
-    if (!player.checkMultipleTagPresence(tags)) {
+    if (tags.length > 1 && !player.checkMultipleTagPresence(tags)) {
       return false;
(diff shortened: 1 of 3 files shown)


src/cards/CardRequirements.ts
```diff
@@ -32,7 +32,7 @@ export class CardRequirements implements ICardRequirements {
         tags.push((requirement as TagCardRequirement).tag);
       }
     });
-    if (!player.checkMultipleTagPresence(tags)) {
+    if (tags.length > 1 && !player.checkMultipleTagPresence(tags)) {
```
bafolts, line 35: If a card had one steel tag and that happened to be the only card in play why would this skip that check if `tags.length` were `1` and looking to see if a steel tag had been played? I don't know if this is actual case in the game but don't understand why we ignore the check for the tag being present when there is only one tag versus multiple tags to check.

PR #4258, opened 2022-03-30
Remove redundancy in rendering a corp logo.

kberg:
I could remove even more, but I think this is a good compromise.

### DIFF
--- a/src/client/components/card/CardCorporationLogo.vue
+++ b/src/client/components/card/CardCorporationLogo.vue
@@ -1,15 +1,6 @@
 <template>
   <div class="card-corporation-logo">
-    <template v-if="title === CardName.APHRODITE">
-      <div class="card-aphrodite-logo">APHRODITE</div>
-    </template>
-    <template v-else-if="title === CardName.ARKLIGHT">
-      <div class="card-arklight-logo">ARKLIGHT</div>
-    </template>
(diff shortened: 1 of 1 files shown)


Argeento (approved):
Nice

src/client/components/card/CardCorporationLogo.vue
```diff
@@ -53,37 +29,19 @@
...
-      <span class="card-viron-logo">VIRON</span>
-    </template>
-    <template v-else-if="title === CardName.ARIDOR">
-      <span class="card-aridor-logo">ARIDOR</span>
-    </template>
     <template v-else-if="title === CardName.ASTRODRILL">
       <span class="card-astrodril-logo">Astrodrill</span>
     </template>
-    <template v-else-if="title === CardName.FACTORUM">
+    <!-- <template v-else-if="title === CardName.FACTORUM">
```
bafolts, line 42: Was this intended to be left commented out?

bafolts (approved):

src/client/components/card/CardCorporationLogo.vue
kberg: > Was this intended to be left commented out?

Ah that should be removed before flight. However, I'm about to change this file today, so leaving it in because _lazy_

PR #4259, opened 2022-03-30
Add Pathfinders corporation card Odyssey

kberg:
Boy things worked out well making this card simple.

### DIFF
--- a/src/cards/pathfinders/Odyssey.ts
+++ b/src/cards/pathfinders/Odyssey.ts
@@ -1,0 +1,76 @@
+import {Card} from '../Card';
+import {ICorporationCard} from '../corporation/ICorporationCard';
+import {Tags} from '../../common/cards/Tags';
+import {Player} from '../../Player';
+import {CardName} from '../../common/cards/CardName';
+import {CardType} from '../../common/cards/CardType';
+import {CardRenderer} from '../render/CardRenderer';
+import {played} from '../Options';
+import {SelectCard} from '../../inputs/SelectCard';
+import {Priority} from '../../deferredActions/DeferredAction';
+import {IProjectCard} from '../IProjectCard';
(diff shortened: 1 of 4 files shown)


src/cards/pathfinders/Odyssey.ts
```diff
@@ -0,0 +1,76 @@
...
+        player.defer(card.play(player), Priority.DEFAULT);
+        this.discard(player, card);
+        return undefined;
+      },
+    );
+  }
+
+  private discard(player: Player, card: IProjectCard) {
+    // This is copied straight out of Project Workshop.
+    const cardIndex = player.playedCards.findIndex((c) => c.name === card.name);
```
bafolts, line 66: Wish that this threw an error if `cardIndex` were `-1`. Both here and where it was copied. This feels like it belongs in a `player.discardPlayedCard` function versus being copied.

bafolts (approved):
Creating a `discardPlayedCard` function on `Player` would avoid duplicating the code from `ProjectWorkshop`. Throwing an error if `cardIndex` were `-1` would also help with failing fast.

src/cards/pathfinders/Odyssey.ts
kberg: Sigh, yeah, I'll do that in a follow-up.

kberg:
> Creating a `discardPlayedCard` function on `Player` would avoid duplicating the code from `ProjectWorkshop`. Throwing an error if `cardIndex` were `-1` would also help with failing fast.

I'm centralizing the method, but I'm having second thoughts about throwing  an error. The problem can be that it's game breaking - this situation would render a game unplayable. I wish we had a better infrastructure that allowed errors which didn't kill a game.

In the meantime I'm going to `console.log` it in an upcoming PR.

PR #4264, opened 2022-04-01
Allow the client global event manifest to read from json

kberg:


src/tools/export_card_rendering.ts
```diff
@@ -6,61 +6,91 @@ import {CardManifest} from '../cards/CardManifest';
...
+  private static processGlobalEvent(globalEvent: IGlobalEvent) {
+    const vizCardData: IClientGlobalEvent = {
+      name: globalEvent.name,
+      description: globalEvent.description,
+      revealedDelegate: globalEvent.revealedDelegate,
+      currentDelegate: globalEvent.currentDelegate,
+      renderData: globalEvent.renderData,
+    };
+
+    GlobalEventProcessor.json.push(vizCardData);
```
bafolts, line 84: Does `json` need to be of type `Array<any>`? Does it store anything besides `IClientGlobalEvent`?

kberg: Good catch. Fixed here and above.

bafolts (approved):

PR #4269, opened 2022-04-01
Rather than exclude client, make a vestigial cards.json.

kberg:
This will enable vscode to operate with the client-side Typescript.


kberg:
I think I have a better solution.

PR #4272, opened 2022-04-02
Rather than import cards.json, use require.

kberg:
This makes it possible to restore client code in tsconfig.json, which
makes those files first-class in vscode.


src/client/cards/ClientCardManifest.ts
```diff
@@ -5,8 +5,9 @@ import {ICard} from '@/cards/ICard';
 import {ICardFactory} from '@/cards/ICardFactory';
 import {GameModule} from '@/common/cards/GameModule';
 import {IClientCard} from '@/common/cards/IClientCard';
-import * as cardJson from '@/genfiles/cards.json';
```
Argeento, line 8: What's wrong with this import? I can successfully run build on main branch when I delete 
```json
"exclude": ["client/**/*.ts"],
```
from `tsconfig`

bafolts: I am also curious why this didn't work unless `require` was used.

kberg: `npm run build` does not work. I suspect @Argeento did not run `npm run clean` to remove the genfiles from the prior build.

image

kberg: (That error is due to the new global events file also having problems, revealing that this PR now actually needs that to be included. :D)

Argeento: Because there is no separation between TS config for the client and the server side -> TS requires all the code to be correct during server build

`require` is just a hidden ts-ignore

Probably the right solution will be: left `src/tsconfig.json` (where `src/client` is excluded)
and create `src/client/tsconfig.json` (make sure that webpack will use this config by adding [ts-loader option](https://github.com/TypeStrong/ts-loader#configfile) in `webpack.config.js`)

kberg: I got it. I can use import even before the file exists, as long as I prefix it with `@ts-ignore`

Still, better vscode.

Argeento: > Still, better vscode.

What do you mean? 

Can you add `// TODO -> Remove ts-ignore after server <-> client code separation`?

kberg: > > Still, better vscode.
> 
> What do you mean?

This solution makes client ts files work with vscode.

Argeento: Okey. 

The definitive solution will be server/client/tools code separation to different dirs

Argeento (approved):

bafolts (approved):

PR #4280, opened 2022-04-03
SelectInitialCards dialog wasn't showing with Prelude.

kberg:
Testing got tricky because it was not easy to find whether ConfirmDialog was visible, and, the underlying element did not have a showDialog method. This works nicely.


kberg:
Tests pass with `npm run tests:client:watch` but not `npm run tests`. Huh. I'll reopen this once it's working.

kberg:
Non-derminism FTL. This should be good now.

PR #4295, opened 2022-04-12
Update VP chart:

kberg:
Update to most recent version of chart.js.
Add minor graph lines at the 5-point mark
Use standard font


kberg:
image

nwai90 (approved):

PR #4299, opened 2022-04-13
Proxy cards should not trigger onCardPlayed.

kberg:
This is an issue with Leavitt and Vitor compatibility. If a player plays
Vitor, and then puts a colony on Leavitt, it calls onCardPlayed, which
calls Vitor's onCardPlayed, which relies on card.metadata. However the
proxy card's metadata throws an error.

This bug arose when centralizing VP management.


tests/cards/base/OlympusConference.spec.ts
```diff
@@ -75,7 +75,7 @@ describe('OlympusConference', function() {
 
   it('Triggers before Mars University', function() {
     const marsUniversity = new MarsUniversity();
-    const scienceTagCard = new ScienceTagCard();
```
kberg, line 78: ScienceTagCard is the proxy card. Using it broke these tests now. Ha.

PR #4300, opened 2022-04-13
Mons Insurance should remove MC when played in as solo.

### LINKED ISSUE #3342: rules error: Mons Insurance should lose 3 MC for any "actual" attack on the neutral opponent
(I also checked if the same happens when I have 3+ MC, and it does.)

In the solo game, even though attacks against the neutral "player" are invisible, they are always treated as having actually happened, and thus Mons is forced to pay.

In case of a resource attack, the Mons player will want to choose whether to attack or not; it may sometimes be worth the 3 MC (such as with a steal attack), sometimes not.

kberg:
Fixes #3342


src/deferredActions/DecreaseAnyProduction.ts
```diff
@@ -22,7 +23,12 @@ export class DecreaseAnyProduction implements DeferredAction {
   ) { }
 
   public execute() {
-    if (this.player.game.isSoloMode()) return undefined;
+    if (this.player.game.isSoloMode()) {
+      if (this.player.game.monsInsuranceOwner === this.player.id) {
```
kberg, line 27: I don't like this repetition. Any thoughts?

src/cards/promo/MonsInsurance.ts
```diff
@@ -42,4 +42,27 @@ export class MonsInsurance extends Card implements ICorporationCard {
     player.game.monsInsuranceOwner = player.id;
     return undefined;
   }
+
+  // When `insured` is undefined, it's the neutral player.
+  public payDebt(player: Player, insured : Player | undefined) {
+    if (player !== insured) {
+      const retribution = Math.min(player.megaCredits, 3);
+      if (insured) insured.megaCredits += retribution;
+      player.deductResource(Resources.MEGACREDITS, retribution);
```
kberg, line 51: This is a small change -- it used to be `3` instead of `retribution` which is sometimes too large.  Reported illegal state.

src/cards/promo/MonsInsurance.ts
```diff
@@ -42,4 +42,27 @@ export class MonsInsurance extends Card implements ICorporationCard {
     player.game.monsInsuranceOwner = player.id;
     return undefined;
   }
+
+  // When `insured` is undefined, it's the neutral player.
+  public payDebt(player: Player, insured : Player | undefined) {
```
kberg, line 47: Hmm. `payDebt` - got a better name?

src/deferredActions/DecreaseAnyProduction.ts
bafolts: To avoid this duplication the only thought I have is introducing a static function on `MonsInsurance. Say there were a function named `payDebt` on `MonsInsurance`. One could then call it with.

```
MonsInsurance.payDept(player, insured)
```

I don't think we need any information on the instance of the card but we could find it if need be.

bafolts: @kberg I am curious what you think of this suggestion? What you have seems fine but I think this static could avoid the duplicated code.

kberg: Hi! I basically did that, I think. `MonsInsurance` is still possibly a little cluttered. Where's the duplicated code now?

bafolts (approved):

src/deferredActions/DecreaseAnyProduction.ts
bafolts: I didn't see you had made the change.

PR #4301, opened 2022-04-13
Save every time a player takes action. And don't save when deserializing.

### LINKED ISSUE #4293: Saving after loading?
This may be specific to the `LocalFilesystem` filesystem, but logs show a game where saving occurs after loading.

```
[2022-04-10 11:29:24 EDT] Loading g2836556ac092
[2022-04-10 11:29:24 EDT] saving g2836556ac092 at position 27
[2022-04-10 11:29:24 EDT] GameLoader loaded game g2836556ac092 into memory from database
```

kberg:
Fixes #4293
Part of #3514

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1952,7 +1952,16 @@ export class Player {
     );
   }
 
-  public takeAction(): void {
+  /**
+   * Set up a player taking their next action.
+   *
+   * This method indicates the avalilable actions by setting the `waitingFor` attribute of this player.
+   *
+   * @param {boolean} saveBeforeTakingAction when true, the game state is saved. Default is `true`. This
+   * should only be false in testing and when this method is called during game deserialization. In other
+   * words, don't set this value unless you know what you're doing.
+   */
+  public takeAction(saveBeforeTakingAction: boolean = true): void {
     const game = this.game;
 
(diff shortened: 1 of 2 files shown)


src/Player.ts
```diff
@@ -1962,9 +1971,7 @@ export class Player {
 
     const allOtherPlayersHavePassed = this.allOtherPlayersHavePassed();
 
-    if (this.actionsTakenThisRound === 0 || game.gameOptions.undoOption) {
```
bafolts, line 1965: Does this still need to check that `game.gameOptions.undoOption` is `true`? There is probably no reason to save for games that don't have the `undoOption`. Typescript may let you set the default value to `Boolean(this.game.gameOptions.undoOption)` for `saveBeforeTakingAction`.

kberg: > Does this still need to check that `game.gameOptions.undoOption` is `true`?

I do not think so.If it's saving with every action, regardless of undo, then it's irrelevant.

> There is probably no reason to save for games that don't have the `undoOption`. Typescript may let you set the default value to `Boolean(this.game.gameOptions.undoOption)` for `saveBeforeTakingAction`.

I rely on it for forensics.

bafolts (approved):

PR #4302, opened 2022-04-13
Add type InputResponse to represent Array<Array<string>>

kberg:
The goal is to eventually replace Array<Array<string>> JSON. Once this
happens, not only will responses be easier to understand, but also
simplify adding more information like and nested in or, or valuable
sidechannel information.

Even if the long goal isn't met (JSON as input format) the code is
clearer and slightly easier to understand.


src/common/inputs/InputResponse.ts
```diff
@@ -0,0 +1 @@
+export type InputResponse = Array<Array<string>>;
```
bafolts, line 1: May be able to mark at least one of these as `Readonly`.

src/client/components/WaitingFor.vue
```diff
@@ -60,7 +60,7 @@ export default Vue.extend({
       }
       document.title = next + ' ' + this.$t(constants.APP_NAME);
     },
-    onsave(out: Array<Array<string>>) {
+    onsave(out: Response) {
```
bafolts, line 63: Should this be `InputResponse` versus `Response`. Do we define the type for `Response` or is this some typescript provided type?

bafolts (approved):
A few of these seem to only use `Response` versus `InputResponse`. I am guessing that is a typo. Would update those before shipping.

This is one of the last relics of early days of this project. Glad to see it go! Surprised this lasted so long.

kberg:
> A few of these seem to only use `Response` versus `InputResponse`. I am guessing that is a typo. Would update those before shipping.

Thanks, will do.

> This is one of the last relics of early days of this project. Glad to see it go! Surprised this lasted so long.

Well, there's a lot of work left. But yes.

src/common/inputs/InputResponse.ts
kberg: Yeah I think it doesn't work, but I'll try again.

Issue #4291, opened 2022-04-10
Tags are in a different order depending on where it is played


kberg:
![image](https://user-images.githubusercontent.com/413481/162624004-cd8c59a1-450d-40f8-aa0d-9056eef878ea.png)

Notice how the space tag is placed at a different index depending on which component shows it.

kberg:
Fixed by #4307

PR #4307, opened 2022-04-15
Rewrite how tags get rendered.

kberg:
This is shorter and a bit quicker than the production version, and also keeps tags in order whether or not it's a shortened list.


kberg:
Fixes #4291

bafolts (approved):

kberg:
Ignore those two references.

PR #4309, opened 2022-04-15
Rename ResourceType to CardResource.

kberg:
It is, after all, a resource on a card.


kberg:
Fairly big rename. I spot-checked it.

PR #4310, opened 2022-04-15
Remove CardAndModule.

kberg:
Now that  ICard is separate from IClientCard, CardAndModule is no longer
necessary.


kberg:
Did a testing cycle

PR #4314, opened 2022-04-16
Start moving input processing to the PlayerInputs.

kberg:
I wanted to do this about a year ago, but this kind of change didn't
work because for some reason it created a dependency cycle that went
from server to client, back to server. Huh?

Now that client and server are mostly separate, this is safe to start
doing. \o/

Getting the processing out of a single method means making more room for
robust validation, and it takes some of the work out of `Player.ts`

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -980,13 +980,11 @@ export class Player {
     return false;
   }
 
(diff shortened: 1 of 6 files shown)


kberg:
Depends on #4302

bafolts (approved):

Issue #3840, opened 2021-12-05
[PF] Show Pathfinders icons in `/help`


kberg:

Borbarad2:
What is left to close this bug?

Borbarad2:
What is left to close this bug?

kberg:
Not sure. I'm just closing this.

PR #4321, opened 2022-04-19
Add some fan content to /help, fix .disease icon

kberg:
This only addresses part of #3840 as I am bumping up against the CSS issues that came about once cards.less and cards_v2.less clashed.

### DIFF
--- a/src/styles/cards.less
+++ b/src/styles/cards.less
@@ -1062,6 +1062,9 @@ input[type="radio"]:checked + .filterDiv::after {
 .camp {
     background-image: url(./assets/resources/camp.png);
 }
+.preservation {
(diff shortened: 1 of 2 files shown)


src/styles/cards.less
```diff
@@ -1273,11 +1276,6 @@ input[type="radio"]:checked + .filterDiv::after {
     color: #bbb;
 }
 
-.disease {
```
kberg, line 1276: There are two entries for .disease. The other is on line 1072.

kberg:
image

I added Camp and a couple other official resource types. You can see that Camp is misaligned. /me lazy, I'll get to it later. Still not entirely sure this /help is the best way to go about it.

PR #4324, opened 2022-04-19
Log where neutral delegates are placed.

### LINKED ISSUE #1797: I wish there was more logging from Turmoil
The log should tell more basic info, like which party became ruling, which party got a neutral delegate, etc. 

Also the Gui components could be improved by forexample darkening the "consumed" neutral delegate parts of event cards, such that it's easier to quickly tell which party is getting a neutral delegate next time, and such.

kberg:
Fixes #1797


kberg:
image

Issue #2492, opened 2021-01-20
Minor bug: duplicate trading fleet at end game


vsrisuknimit:
![image](https://user-images.githubusercontent.com/14239220/105169205-33ed4680-5ae9-11eb-8b22-21dafe3b2fdb.png)

At end game, leave the fleets on the moons or return them. Currently they are duplicated.

Low priority bug. No gameplay effect.

PR #4327, opened 2022-04-19
Don't return trade fleets at game end.

kberg:
image

This player had two trade fleets.


kberg:
In the sample above, one of the trade fleets was used, and the other left alone. Even though it looks like the broken condition, I tested it with a single player with two trade fleets.

kberg:
Fixes #2492

bafolts (approved):

PR #4332, opened 2022-04-20
Fix SelectProductionToLose parsing for Units

Strikeskids:
This PR adds validation so that users are not able to specify negative or non-numeric unit amounts. Before this validation, players could make their production massive by "losing" a negative amount of production.

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1116,7 +1116,7 @@ export class Player {
       if (!Units.keys.every((k) => units[k] >= 0)) {
         throw new Error('All units must be positive');
       }
-      if (!this.hasUnits(units)) {
+      if (!this.canAdjustProduction(Units.negative(units))) {
         throw new Error('You do not have those units');
       }
       pi.cb(units);
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1008,6 +1008,18 @@ export class Player {
     }
   }
 
+  private parseUnitsJSON(json: string): Units {
+    try {
+      const units: unknown = JSON.parse(json);
+      if (!Units.isUnits(units)) {
+        throw new Error('not a units object');
+      }
+
+      return units;
+    } catch (err) {
+      throw new Error('Unable to parse Units input ' + err);
+    }
+  }
   protected runInput(input: InputResponse, pi: PlayerInput): void {
     if (pi instanceof AndOptions) {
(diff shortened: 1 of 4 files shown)


kberg:
```
/home/runner/work/terraforming-mars/terraforming-mars/src/common/Units.ts
Error:   50:23  error  Expected parentheses around arrow function argument  arrow-parens

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
```

You can catch these with `npm run lint`

kberg (commented):
This looks good. How have you verified that this works? I'd like to get a unit test as part of this PR. I can't tonight, but will be happy to take a crack at it over the next couple of days. Sorry I can't do this sooner.

src/Player.ts
```diff
@@ -1130,8 +1142,14 @@ export class Player {
...
       this.runInputCb(pi.cb(howToPay));
     } else if (pi instanceof SelectProductionToLose) {
-      // TODO(kberg): I'm sure there's some input validation required.
-      const units: Units = JSON.parse(input[0][0]);
+      this.checkInputLength(input, 1, 1);
+      const units: Units = this.parseUnitsJSON(input[0][0]);
+      if (!Units.keys.every((k) => units[k] >= 0)) {
+        throw new Error('All units must be positive');
+      }
+      if (!this.hasUnits(units)) {
```
kberg, line 1150: `this.hasUnits` only applies to units in your inventory, whereas this has to check against production.

How did you verify this works?

kberg (changes requested):

kberg:
Here is a sanity test that you can use. It includes changing SelectProductionToLoseDeferred to also capture that the production lost matches the count required.

Because I'm about to refactor this stuff, I'm not so worried that some of the validation is in `Player` and some in `SelectHowToPayDeferred`. Once this is done, I can merge it into one place.

kberg:
Ah wait -- **here** is the sanity test: https://github.com/terraforming-mars/terraforming-mars/commit/4a4a154e153dfc75bfef68295478d8d2dbdb1268

src/Player.ts
kberg: The call you want is `this.canPoduce`

kberg (approved):
This is nice! Thank you.

Issue #4278, opened 2022-04-03
Worms cardNumber is 130


skluyskens:
See also https://boardgamegeek.com/image/3169324/milenaguberinic

Maybe you should add a unit test to validate that all cards have unique cardNumber's?

skluyskens:
Space elevator is 13
https://www.fryxgames.se/wp-content/uploads/2016/01/TM013spaceelevator.ai_1.jpg

skluyskens:
Heather is 88
https://entropymag.org/wp-content/uploads/2017/11/IMG_6241.jpg

kberg:
> Maybe you should add a unit test to validate that all cards have unique cardNumber's?

We've talked about that. It'll be a bear of an effort because card numbers are currently corrected by best effort.

Thanks for the report.

kberg:
The card number is no longer a visual component. Closing this bug.

Issue #2775, opened 2021-02-16
"Show card numbers" for corporations (minor issues)


kevinb9n:
1. Livin' on the edge

<img src=https://user-images.githubusercontent.com/934551/108095676-fb338500-7034-11eb-8ff3-1a273e583c3d.png width=600>

2. Is there value in the id numbers given that we have to invent them?

3. The numbers chosen seem kinda random? Simon's numbering makes sense to me (https://ssimeonoff.github.io/cards-list, hit corps, numbers aren't shown but they're in that order)

@kberg, I think?

kberg:
Agreed the layout is broken for corps. I assumed I was the only one
interested in this given feedback when I wrote it. But I will prioritize
laying it out.
ID numbers are at least helpful in reporting bugs. And for me they have
been helpful in finding cards as I develop and debug them, even if some of
them are arbitrary. Since they're not keys in our app, they can change
(though of course that adds the same kind of problems that come with
renumbering highway exits.)

kevinb9n:
Cool. Well, I like the id numbers. Universal identifiers are good. There are pairs of card names that are very similar, and in other languages there might be *different* pairs of card names that are similar, etc.

I think renumbering (chrono-then-alpha as Simon has) would be "nice". Or, for corporations a three-letter code would have the advantage of being somewhat "knowable". (Other card ids are knowable only by looking at the physical copy, but at least that's something.) The only conflict so far is Teractor vs Terralabs (suggest TER vs TRR).

kberg:
The moon will not help with easy 3-letter characters. This is like naming
airports, or borg cells.

kberg:
I'm wondering whether card numbers are worthwhile any more. Going to poll Discord to ask.

kberg:
The card number is no longer a visual component. Closing this bug.

PR #4336, opened 2022-04-20
Remove the Card Number component from cards.

kberg:
It's not helping, and gets visually in the way.


kberg:
Fixes #4278
Fixes #2775

nwai90 (approved):

PR #4340, opened 2022-04-21
Fix HowToPay bugs in processInput

Strikeskids:
This PR validates the amounts specified in HowToPay so that people aren't able to generate free resources by paying negative amounts. 

It shares more of the computation of how much to pay between finding valid cards to play and actually playing the card.


kberg:
Replacing this with #4734 which solves all the merge conflicts. You get full credit though. Thank you!

Issue #3038, opened 2021-03-26
Feature request: reduce number of click for Helion


vsrisuknimit:
This suggestion

![image](https://user-images.githubusercontent.com/14239220/112699205-b4ba1d00-8e61-11eb-9dc9-1d61de041c8e.png)

or alternatively a button just for Helion such that mc will always be spent before while this button is toggled.

nwai90:
The button is a better approach if this is to be done as it avoids additional custom deck manipulation. There are also several other reasons why you would want to spend heat before MC, e.g. endgame MC tiebreaker.

kberg:
A max for the MC button is reasonablel, at the very least.

PR #4345, opened 2022-04-22
Adds a "Max" button for Megacredits.

kberg:
The SHTP widget doesn't really do much when the megacredit value
changes; it's up to players to change the other units. That can change.

But for now, this will do the bare minimum while generating feedback for
how to refine this.


kberg:
Applies to #3038

bafolts (approved):

PR #4346, opened 2022-04-22
Extra visuzliation for action cards played this generation

### LINKED ISSUE #3046: Feature request: see what cards you have used or hasnt used for putting res on them
It would be awesome if when you for example need to put floaters on your cards that among your res-order options it would be clear what cards you have used this gen (and can thu not be used this gen).

As it is now you have do keep track of it in your mind or scroll down to your played cards to see what cards are "greyed-out" or not.

kberg:
Two changes:
1. When having to select a card for adding resources, if the card was played this generation, use the dark filter.
2. When hovering over a card with a dark filter, remove the filter, making it easier to see.

This fixes #3046

### DIFF
--- a/src/client/components/SelectCard.vue
+++ b/src/client/components/SelectCard.vue
@@ -6,7 +6,7 @@
               <input v-if="isSelectOnlyOneCard()" type="radio" v-model="cards" :value="card" />
               <input v-else type="checkbox" v-model="cards" :value="card" :disabled="playerinput.maxCardsToSelect !== undefined && Array.isArray(cards) && cards.length >= playerinput.maxCardsToSelect && cards.includes(card) === false" />
             </template>
-            <Card :card="card">
+            <Card :card="card" :actionUsed="isCardActivated(card)">
               <template v-if="playerinput.showOwner">
                 <div :class="'card-owner-label player_translucent_bg_color_'+ getOwner(card).color">
                   {{getOwner(card).name}}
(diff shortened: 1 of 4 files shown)


kberg:
image
image

kberg:
And works in the Played Cards section:

image

src/client/components/SelectCard.vue
```diff
@@ -161,6 +161,10 @@ export default Vue.extend({
     buttonLabel(): string {
       return this.isSelectOnlyOneCard() ? this.playerinput.buttonLabel : this.playerinput.buttonLabel + ' ' + this.cardsSelected();
     },
+    isCardActivated(card: CardModel): boolean {
+      // Copied from PlayerMixin.
+      return this.playerView.thisPlayer.actionsThisGeneration?.includes(card.name);
```
bafolts, line 166:
```suggestion
      return this.playerView.thisPlayer.actionsThisGeneration.includes(card.name);
```

I don't think the `?` is necessary. Since this is compiling typescript must think this will never return `undefined`.

src/client/mixins/PlayerMixin.ts
```diff
@@ -44,11 +44,7 @@ export const PlayerMixin = {
       card: CardModel,
       player: PublicPlayerModel,
     ): boolean {
-      return (
-        (player !== undefined &&
-                player.actionsThisGeneration !== undefined &&
-                player.actionsThisGeneration.includes(card.name)) || card.isDisabled
-      );
+      return player?.actionsThisGeneration?.includes(card.name) || card.isDisabled;
```
bafolts, line 47: According to our types both `player` and player.actionsThisGeneration` should be defined. These checks for `undefined` are unnecessary.

bafolts (approved):

src/client/mixins/PlayerMixin.ts
kberg: They are!

src/client/components/SelectCard.vue
kberg: Fixed! Why doesn't a linter warn about this?

PR #4347, opened 2022-04-22
Remove checkbox from card selection, replace with subtle brightness shift.

### LINKED ISSUE #3211: Remove the orange bar - revert to the orange border
In cases where only one card is available to select, the orange bar occludes the text.

image

kberg:
Fixes #3211

Before
image
After
image

### DIFF
--- a/src/styles/cards.less
+++ b/src/styles/cards.less
@@ -95,26 +95,15 @@
     box-shadow: 0 0 2px 1px black;
 }
 
-input[type="checkbox"]:checked + .filterDiv,
-input[type="radio"]:checked + .filterDiv {
-    box-shadow: 0 0 0 2px #303030, 0 0 0px 6px #c97e23;
+input[type="checkbox"] + .filterDiv,
+input[type="radio"] + .filterDiv {
+    filter: brightness(0.9);
 }
 
-input[type="checkbox"]:checked + .filterDiv::after,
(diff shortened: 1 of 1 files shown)


nwai90:
@kberg I found some issues with this change, as the same UI is shared for colony and turmoil party selection, so the checkmarks are removed for those as well. We might need to define new classes for it so that we preserve the visual indicator for at least the turmoil parties:

Screenshot 2022-04-24 at 12 10 44 PM

Screenshot 2022-04-24 at 12 10 54 PM

kberg:
Ooh, thank you. Yes, that makes sense.

pauliimmonen:
I wonder if this could be the more subtle approach? I kept the checkbox for distinguishability and as small as possible above the card. Tho I had to make few pixels room between card rows. Too small checkbox would be behind the corporation tag and it looked very inconvenient in front of it. Colonies and turmoil checkboxes as they were.

project_cards
standard

kberg:
> I wonder if this could be the more subtle approach? I kept the checkbox for distinguishability and as small as possible above the card. Tho I had to make few pixels room between card rows. Too small checkbox would be behind the corporation tag and it looked very inconvenient in front of it. Colonies and turmoil checkboxes as they were.
> 
> project_cards standard

That would also be terrific.

kberg:
Closing in favor of #5018

Issue #2861, opened 2021-02-27
Solo mode is part of the Corporate Era expansion (selecting CorpEra should be mandatory)


kevinb9n:
The published rule is that solo mode only exists when using the Corporate Era expansion. The reason is that the developers feel that solo without CorpEra [makes for a bad game](https://boardgamegeek.com/thread/1663126/article/24081575#24081575).

Scanning a few threads suggests also that solo made is actually harder (!) without CorpEra, which is opposite to what players who choose this surely expect.

The app allows solo without CorpEra. I'm not sure if it is intentional. Allowing it may not be "free", because new players will be likely to assume that's how they want to play, not knowing any of the above. It is intuitively "obvious" that this would make for an easier learning curve, but surprisingly, virtually every game concept is encountered somewhere in the base deck alone anyway.

kevinb9n:
Hello thumb! I would welcome the counterarguments.

I think it's fair to frame it like this: of the infinite unofficial variants that could be supported, why should this one make the cut?

I do think the standard to meet should be a bit higher than "some people want to play this way sometimes". The fact that it's very intentionally not an official variant should have *some* amount of weight.

SliceOfBread:
What is the benefit and what is the effort to gain that benefit? In this case, the benefit is fleetingly small and the effort is not.

OTOH, if this were already the status quo, I would make the same exact argument against adding a Solo w/o Corp Era mode.

In the interest of fairness, I will try this mode out and see how bad a game it makes for.

SliceOfBread:
I played the Beginner Corp and won with 124 pts. so I don't think this will be challenging to anyone but a real beginner. I don't see any reason not to allow it though.

kevinb9n:
(this issue isn't related to Beginner Corp)

The *main* issue here isn't offering a variant that the fryxfolks intentionally didn't want; it's that unsuspecting players are getting that variant without having any idea they're going outside those bounds.

bafolts:
I have a funny story here which helps to prove @kevinb9n 's point. I didn't realize corporate era was a thing for solo games. When I first wrote this game and I would test solo mode I was wondering how anyone could ever win. Will update this to only show the corporate era option when more than one player. Jonathan Fryxelius' word is always final.

kberg:
[...]

PR #4357, opened 2022-04-24
When selecting a solo game re-enable corporate era.

kberg:


kberg:
This partially addresses #2861

nwai90 (approved):

Issue #3745, opened 2021-10-23
Don't show the moon variant options if the moon variant game option is not selected.


kberg:
![image](https://user-images.githubusercontent.com/413481/138561316-329cf42f-79e6-450a-899c-4c962654cc0a.png)

PR #4359, opened 2022-04-24
Don't show the Variant Moon standard projects if not opted in.

kberg:


kberg:
Fixes: 3745

kberg:
Fixes #3745

PR #4360, opened 2022-04-24
Properly report a game's state through the spectator ID.

kberg:
This keeps me from having to wire up game state through the preferences.
Unfortunately it means exposing the interface in App.ts. I really tried
other things. Someone more adept at Typescript is welcome to show me a
thing or two.


kberg:
**Before:**
image

**After:**
image

nwai90 (approved):

PR #4363, opened 2022-04-25
Remove the mnemonist module from the client

kberg:
It's used in one place, and reduces the (uncompressed) binary another 100K


kberg:
Relies on #4361

PR #4371, opened 2022-04-27
Make Asteroid Standard Project compatible with Aurorai

### LINKED ISSUE #4308: [pf] New Aurorai payment regression
Bug report for Aurorai: After paying the first starting data resources from Aurorai, you were not able to play and pay the standard project again which you have selected first (in our case the standard asteroid) if you have less money (than the cost of the standard project) but enough data to pay the difference.

kberg:
Fixes #4308

### DIFF
--- a/src/cards/base/standardProjects/AsteroidStandardProject.ts
+++ b/src/cards/base/standardProjects/AsteroidStandardProject.ts
@@ -9,6 +9,7 @@ export class AsteroidStandardProject extends StandardProjectCard {
     super({
       name: CardName.ASTEROID_STANDARD_PROJECT,
       cost: 14,
+      tr: {temperature: 1},
       metadata: {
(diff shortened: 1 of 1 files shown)


src/cards/base/standardProjects/AsteroidStandardProject.ts
```diff
@@ -24,9 +25,7 @@ export class AsteroidStandardProject extends StandardProjectCard {
     if (player.game.getTemperature() === constants.MAX_TEMPERATURE) {
       return false;
     }
-    return player.canAfford(this.cost, {
-      tr: {temperature: 1},
-    });
+    return super.canAct(player);
```
kberg, line 28: Basically this was due to AsteroidStandardProject not using the same pattern the other standard projects use.

nwai90 (approved):

PR #4374, opened 2022-04-28
Don't automatically show standard projects when none are enabled.

kberg:
image


kberg:
More work to do.

kberg:
Reopening.
image

nwai90 (approved):

Issue #4375, opened 2022-04-28
Red Influence Turmoil event took more money than it should


JeremyDietrich:
![TMError](https://user-images.githubusercontent.com/67653347/165678375-112a7af4-22e8-4e0d-8b61-c1b1f6e7440b.png)

This should only decrease Red's money by 15 max because of the wording saying max 5 sets, but as Red was at 59 TR it was instead decreased by 27 (9 sets).

kberg:
Thanks. A fix is out.

PR #4376, opened 2022-04-28
Red Influence doesn't observe its 'max 5' limit.

kberg:


kberg:
Fixes #4376

kberg:
Fixed #4375

Issue #3994, opened 2022-01-25
[Bug] Game setup should not allow player to select a limited set of corps when those corps aren't compatible with the game.


ap3h3ad:
Hi,

I started a game today and included Pathfinder for the first time. One of the players played Ecoline and we noticed that the player got a plant production of 6. The log states "player x played Ecoline" three times. So I think the plant production of Ecoline gehts applied three times. Since Pathfinder is the only new thing in this game we think that it most definately has to do with this expansion.

One of the players picked a Turmoil corporation although the Turmoil expansion wasn't available at all. Maybe this triggered the bug.

kberg:
Do you have a link to the game?

ap3h3ad:
Sadly the game is gone.

I was informed by another player that this bug indeed happend before so I removed the "Pathfinder" part in the title. In both cases one player chose the corporation Incite without Turmoil being present in the game. The result was in both cases that every corporation got announced three times in the log and every corporation regarding production or resources were handles three times. So this might have to do with Incite specifically.

kberg:
OK. What server?

ap3h3ad:
Hi.

I was able to reconstruct the error case so that everyone can have a look at it on his own server. I used this settings:

https://abload.de/img/settingsyejth.png

to make sure it is possible to pick Incite for one of the players.

Then I picked Incite with the one player and Thorgate with the other one. I got prompted with this error message:

https://abload.de/img/error4mk8m.png

Seems reasonable, since Turmoil was indeed not active. But then I had a look at the log:

https://abload.de/img/thorgatebrokenxtj24.png

You can see that Thorgate gets mentioned two times and that the energy production got assigned two times. That incite is played isn't mentioned at all in the log. So I think the game doesn't handle the case "Turmoil corporation gets played without turmoiil in the game" correctly.

This might be due to the fact that the game (sometimes or always?) doesn't remove the Turmoil corporations from the custom corporation list when the Turmoil option is enabled and then disabled (so the Turmoil corporations come into the game in the first place).

I hope I delivered enough detail to shoot that bug. :)

kberg:
@ap3h3ad what server is this?

ap3h3ad:
@kberg 

This happened on my private server, not the public one (heroku?). If you want to know the URL - I'm sorry but I'm uncomfortable with sharing the IP of my server in this publicly accessible part of the internet.

kberg:
[...]

Issue #2225, opened 2020-12-29
Behavior when foreign corporations are included


kevinb9n:
When I include a corporation but don't include the expansion it comes from, some strange scenarios can result.

* Valley Trust should work normally. It should still let you choose and play a Prelude card - this doesn't cause any weird situations like the other cases, and the balance is no different. Assuming that the player chose it, that is obviously what they expected to happen, and today they can only beg to restart the game.

* Four corporations essentially become dead draws: Morning Star, Aphrodite, Poseidon, Septem Tribus. You can barely even *contrive* a scenario where a player would choose one on purpose. And the danger is that a player might see the corp and not realize the expansion isn't in play, and torpedo their entire game. Recommendation: make it impossible to even click them without their expansion selected.

* I think Celestic should be put in that category as well even though it can technically work as long as *Colonies* is in.

* Aridor is still reasonably usable, and I have actually included it and played it in non-Colonies games, because it's the effect that is the interesting part. But instead of giving you a mandatory first action that does nothing, it should just not give you a mandatory first action at all.

kevinb9n:
> But instead of giving you a mandatory first action that does nothing, it should just not give you a mandatory first action at all.

This part also applies to Vitor in a solo game.

ap3h3ad:
Tbh I don't know why you would include corporations and not include the expansions that they belong to. In my opinion the game should just completely prevent that behaviour (so that you have to choose the expansions first and the corporation options change on that basis), not just for the four corporations you listed. I mean, if you only have the base game in your shelf you can't just add the Venus Next corporations and leave out the other Venus Cards and the Venus Board. 

Actually I don't understand that first point of yours. How should Valley Trust allow picking Prelude cards when the expansion is not in the game? In that case there are not Prelude cards in the game.

kevinb9n:
> Tbh I don't know why you would include corporations and not include the expansions that they belong to. In my opinion the game should just completely prevent that behaviour (so that you have to choose the expansions first and the corporation options change on that basis), not just for the four corporations you listed. I mean, if you only have the base game in your shelf you can't just add the Venus Next corporations and leave out the other Venus Cards and the Venus Board.

Can you mention an advantage of this approach?

> Actually I don't understand that first point of yours. How should Valley Trust allow picking Prelude cards when the expansion is not in the game? In that case there are not Prelude cards in the game.

[...]

PR #4394, opened 2022-05-03
Show corp compatibility requrements during game setup

kberg:
Partly addresses #3994 

image


kberg:
Just added a warning confirmation when trying to create a game with an incompatibility.

This will fix #3994 

image

kberg:
Fixes #2225

PR #4396, opened 2022-05-03
Show card discount when selling patents.

### LINKED ISSUE #2724: Always show discounts on hand cards, including when choosing what to discard
Noticed that Sell Patents and the interface for discarding a card for Sponsored Academies aren't showing the discounts.

kberg:
Fixes #2724


kberg:
image

PR #4409, opened 2022-05-05
Overhaul debug ui

kberg:
Invert for all 3 types.
Remove watch and active URL filtering. UI is much faster.
Tags buttons now only have icons.
Filtering by module applies to colonies and global events.


kberg:
Also, based on some internals, this is now a much better source for translations.

PR #4414, opened 2022-05-06
Log the number of cards bought during initial draft.

### LINKED ISSUE #3082: Feature requesto: LOG: seeing the cards bought after Corp pick
Could we get to see in the log right after seeing what Players have picked what Corporation the amount of cards they have brought from their starting hand?

As always, the greatest of work your doing guys!

kberg:
Fixes #3082


kberg:
image

PR #4418, opened 2022-05-07
add source field to settings json

AnnaMinasian:
Added `source` field to JSON with settings when user clicked download current settings.
Check for `source` field in JSON when user clicked upload settings and make an alert if sources are different.

Dependency https://github.com/terraforming-mars/terraforming-mars/issues/4399

### DIFF
--- a/src/client/components/create/CreateGameForm.vue
+++ b/src/client/components/create/CreateGameForm.vue
@@ -1,1073 +1,549 @@
 <template>
-  <div id="create-game">
-    <h1>
-      <span v-i18n>{{ constants.APP_NAME }}</span> —
-      <span v-i18n>Create New Game</span>
-    </h1>
-    <div class="create-game-discord-invite" v-if="playersCount === 1">
-      (<span v-i18n>Looking for people to play with</span>?
-      <a href="https://discord.gg/VR8TbrD" class="tooltip" target="_blank"
-        ><u v-i18n>Join us on Discord</u></a
-      >.)
-    </div>
-
-    <div class="create-game-form create-game-panel create-game--block">
-      <div class="create-game-options">
-        <div v-if="isSoloModePage">
-          <div
-            class="create-game-solo-player form-group"
-            v-for="newPlayer in getPlayers()"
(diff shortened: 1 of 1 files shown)


src/client/components/create/CreateGameForm.vue
```diff
@@ -1014,4 +1022,4 @@ export default (Vue as WithRefs<Refs>).extend({
     },
   },
 });
-</script>
+</script>
```
kberg, line 1025: Did you remove the line at the end of the file? The linter will probably be unhappy.

src/client/components/create/CreateGameForm.vue
```diff
@@ -659,6 +660,12 @@ export default (Vue as WithRefs<Refs>).extend({
         const readerResults = reader.result;
         if (typeof(readerResults) === 'string') {
           const results = JSON.parse(readerResults);
+
+          const source = results['source'];
+          if(source !== window.location.origin) {
+            alert($t('Source of uploaded settings is different, there might be some inconsistencies'));
```
kberg, line 666: This is good work, but as I think about it, the feature, as written on Github Issues probably hurts more than it helps. Here's why:

1. What's really important is not the source, but  that the parameters make sense. Right now the app will silently swallow any settings without checking that they make sense, and without warning somebody creates a game that doesn't have the parameters they expect.
2. If we change download/upload settings, the domain isn't really enough to validate that the settings should match.
3. If we change download/upload settings, the domain isn't really enough to validate that the settings should match.
4. As-is, this is going to immediately block any player that uploads settings; so an alert will be more disruptive than helpful. This code could special-case when source is empty, but that also doesn't help.

Sorry for this one. I thought it was a good starter, but now I think not. :/ Feel free to DM on discord if you would like to pick up another change. (The one specified in #1, above, might even be good.)

kberg (changes requested):

src/client/components/create/CreateGameForm.vue
kberg: I am closing the issue. If you want to do something else, all you have to do is respond.

PR #4426, opened 2022-05-08
German Turmoil translation and typo fix

pczyk:
As promised, here's the German Turmoil translation and a minor typo fix (inclusive -> inklusive).


kberg (approved):
Thank you!

PR #4428, opened 2022-05-09
Remove white background in tags.

sickerin:
Very very minor change. Maybe a more long-term fix could be removing the backgrounds of the images directly? A lot of the other icons don't have the background, it seems. Maybe removing the background would impact other places like the [tag icons](https://github.com/terraforming-mars/terraforming-mars/blob/f8487fca90c38d552c14d258097b8f83a103cd6a/src/styles/player_home.less#L152) in the main page.

Before:
image
After:
image

Another really really minor change could be to align the tag? Not sure why left-aligned.  The corporation tags still look okay with `margin: 5px 5px 5px 5px` instead of `margin: 5px 10px 5px 0px` Now the tag cards icons will be aligned center. This change isn't in the PR.

image


kberg:
This page is blatantly copying CSS from the create-game page, so the first thing I have to check is how this impacts the that page:

image

And it does.

Removing the background, or, creating a different set of UI for this page makes more sense.

image

Removing backgrounds isn't too tough.

sickerin:
Oops, you're right, I missed that small detail. I think removing the background is the better option. I can remove the bg and upload the assets.

kberg (approved):
This looks great. Thanks!

PR #4450, opened 2022-05-14
Rename cleanSaves to truncateGame.

kberg:


kberg:
Don't bother - the next bit of work actually removes the method entirely. ;D

PR #4453, opened 2022-05-15
Rewrite how sqlite purges unfinished games.

kberg:
First, the code centralizes computing time to seconds, because trusting
sqlite's date math is fine, but putting the complexity in Typescript is
better; we're better in Typescript than SQLite.

Second, type SQLite functions were not computing days ago as if it were
that many seconds ago, but rather some midnight date n days ago. That
means all unfinished games over a 24-hour period are purged at once
rather than over time.

If this same behavior exists with PostgreSQL, it could account for some
of the occasional slow operations, but I doubt it.

Finally, fixing the purge method; async sleep() is kinda better.

### DIFF
--- a/src/database/SQLite.ts
+++ b/src/database/SQLite.ts
@@ -5,6 +5,7 @@ import {IGameData} from '../common/game/IGameData';
 import {SerializedGame} from '../SerializedGame';
 
 import sqlite3 = require('sqlite3');
+import {daysAgoToSeconds} from './utils.ts';
 const path = require('path');
 const fs = require('fs');
 const dbFolder = path.resolve(process.cwd(), './db');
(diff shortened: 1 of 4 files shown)


src/database/SQLite.ts
```diff
@@ -203,10 +204,13 @@ export class SQLite implements IDatabase {
...
   }
 
-  purgeUnfinishedGames(): void {
+  purgeUnfinishedGames(maxGameDays: string | undefined = process.env.MAX_GAME_DAYS): Promise<void> {
     // Purge unfinished games older than MAX_GAME_DAYS days. If this .env variable is not present, unfinished games will not be purged.
-    if (process.env.MAX_GAME_DAYS) {
-      this.runQuietly(`DELETE FROM games WHERE created_time < strftime('%s',date('now', '-' || ? || ' day')) and status = 'running'`, [process.env.MAX_GAME_DAYS]);
+    if (maxGameDays) {
+      const dateToSeconds = daysAgoToSeconds(maxGameDays, 0);
+      return this.runQuietly(`DELETE FROM games WHERE created_time < ? and status = 'running'`, [dateToSeconds]);
```
bafolts, line 211: Does this need to consider timezone? if the timezone of the database and node runtime differ this will end up purging different games than it did before.

bafolts (approved):

src/database/SQLite.ts
kberg: No, these are all just seconds since the epoch. The prior solution probably needed to consider time zones.

PR #4454, opened 2022-05-15
Fix placeholder for 'Your name' input in game setup, enabling i18n

pczyk:
I've adjusted the placeholder for the "Your name" input. The localizations are now correctly applied to the placeholder.

Before:
before

After:
after


kberg (approved):
Thank you!

PR #4461, opened 2022-05-18
First Postgres integration test.

kberg:
This test cannot be run as part of the regular suite. I mean, perhaps it
could, but not in its current state.


src/database/PostgreSQL.ts
```diff
@@ -7,44 +7,35 @@ import {SerializedGame} from '../SerializedGame';
 import {Pool, ClientConfig, QueryResult} from 'pg';
 
 export class PostgreSQL implements IDatabase {
-  private client: Pool;
+  protected client: Pool;
 
-  constructor() {
-    const config: ClientConfig = {
+  constructor(
+    config: ClientConfig = {
```
kberg, line 13: This makes the connection overridable.

src/database/PostgreSQL.ts
```diff
@@ -7,44 +7,35 @@ import {SerializedGame} from '../SerializedGame';
...
-    this.client.query('CREATE INDEX IF NOT EXISTS games_i2 on games(created_time )', (err) => {
-      if (err) {
-        throw err;
-      }
-    });
   }
 
-  async initialize(): Promise<void> {
-
+  public async initialize(): Promise<void> {
```
kberg, line 25: Moving those calls to initialize (and using promise chaining) means the tests won't start until the tables are all set up.

src/database/PostgreSQL.ts
```diff
@@ -266,21 +257,17 @@ export class PostgreSQL implements IDatabase {
     });
   }
 
-  saveGame(game: Game): Promise<void> {
+  async saveGame(game: Game): Promise<void> {
```
kberg, line 260: An async saveGame is pretty important. I _think_ I understand them. Your job is to understand them. ;)

src/database/PostgreSQL.ts
bafolts: The less done in constructor is usually directly proportional to how easy the class is to test. 👍

src/database/PostgreSQL.ts
```diff
@@ -7,44 +7,35 @@ import {SerializedGame} from '../SerializedGame';
...
-      if (err) {
-        throw err;
-      }
-    });
   }
 
-  async initialize(): Promise<void> {
-
+  public async initialize(): Promise<void> {
+    return this.client.query('CREATE TABLE IF NOT EXISTS games(game_id varchar, players integer, save_id integer, game text, status text default \'running\', created_time timestamp default now(), PRIMARY KEY (game_id, save_id))')
```
bafolts, line 26:
```suggestion
    await this.client.query('CREATE TABLE IF NOT EXISTS games(game_id varchar, players integer, save_id integer, game text, status text default \'running\', created_time timestamp default now(), PRIMARY KEY (game_id, save_id))')
```

With this now inside an `async` function making use of `await` will spare us from the `then` ladders of yore.

src/database/PostgreSQL.ts
bafolts: Prior to this change the query to save the game would run in the background. I think what you have done here is the right thing. We are going to see an increase in response time after this change since we will have to wait for this query to run. I mention this so no one is surprised if we see increase in response time once this merges.

src/database/PostgreSQL.ts
```diff
@@ -266,21 +257,17 @@ export class PostgreSQL implements IDatabase {
     });
   }
 
-  saveGame(game: Game): Promise<void> {
+  async saveGame(game: Game): Promise<void> {
     const gameJSON = game.toJSON();
-    this.client.query(
+    return this.client.query(
```
bafolts, line 262: With `es6` #4467 and `async` here I would use `await` versus `then` and `catch`. A general rule of thumb I practice is to only use `then` and `catch` in very rare situations. Even then I wouldn't use them.

```suggestion
    try {
        await this.client.query()
        game.lastSaveId++;
     } catch(err) {
         console.error();
     }
```

tests/integration/PostgreSQL.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  // Tests can wait for saveGamePromise since save() is called inside other methods.
+  public override saveGame(game: Game): Promise<void> {
+    this.saveGamePromise = super.saveGame(game);
+    return this.saveGamePromise;
+  }
+
+  public async getSaveIds(gameId: GameId): Promise<Array<number>> {
+    return this.client.query('SELECT distinct save_id FROM games WHERE game_id = $1', [gameId]).then(
```
bafolts, line 35:
```suggestion
    const res = await this.client.query('SELECT distinct save_id FROM games WHERE game_id = $1', [gameId]);
```

tests/integration/PostgreSQL.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+    expect(game.lastSaveId).eq(1);
+
+    await db.saveGame(game);
+    await db.saveGame(game);
+    await db.saveGame(game);
+
+    expect(await db.getSaveIds(game.id)).has.members([0, 1, 2, 3]);
+
+    db.cleanSaves(game.id);
+    await sleep(500);
```
bafolts, line 107: If this is here because `cleanSaves` is synchronous I would leave a TODO to come back and make `cleanSaves` asynchronous as well and wait on it here.

bafolts (approved):

src/database/PostgreSQL.ts
kberg: Tests seem to run in milliseconds, so I'm not particularlly worried.

Besides won't it block only when calls to saveGame are blocked?

src/database/PostgreSQL.ts
kberg: Yes, this makes sense. Should this also remove all the `then` calls and just call await four times in a row?

PR #4466, opened 2022-05-18
More postgresql tests and more promises.

kberg:


src/database/PostgreSQL.ts
```diff
@@ -7,44 +7,30 @@ import {SerializedGame} from '../SerializedGame';
...
-        throw err;
-      }
-    });
   }
 
-  async initialize(): Promise<void> {
-
+  public async initialize(): Promise<void> {
+    return this.client.query('CREATE TABLE IF NOT EXISTS games(game_id varchar, players integer, save_id integer, game text, status text default \'running\', created_time timestamp default now(), PRIMARY KEY (game_id, save_id))')
+      .then(() => this.client.query('CREATE TABLE IF NOT EXISTS game_results(game_id varchar not null, seed_game_id varchar, players integer, generations integer, game_options text, scores text, PRIMARY KEY (game_id))'))
```
bafolts, line 27: Does this need to be re-based? The changes in this PR look like the last commit.

kberg: No, this is correct. It's just slightly simpler.

bafolts: I haven't read comments since last PR but I would use only await and no `.then` or `.catch`. Mixing `await` with `.then` and `.catch` reads strange.

src/database/PostgreSQL.ts
```diff
@@ -182,51 +176,35 @@ export class PostgreSQL implements IDatabase {
...
-    });
-  }
-
-  throwIf(err: any, condition: string) {
-    if (err) {
-      console.error('PostgreSQL', condition, err);
-      throw err;
-    }
+  private async getMaxSaveId(game_id: GameId): Promise<number> {
+    return this.client.query('SELECT MAX(save_id) as save_id FROM games WHERE game_id = $1', [game_id])
```
bafolts, line 180: This function returns a `Promise` and doesn't need to be marked `async`.

src/database/PostgreSQL.ts
```diff
@@ -182,51 +176,35 @@ export class PostgreSQL implements IDatabase {
...
     const days = Number.isInteger(envDays) ? envDays : 10;
-    this.client.query('DELETE FROM games WHERE created_time < now() - interval \'1 day\' * $1', [days], function(err?: Error, res?: QueryResult<any>) {
-      if (res) {
-        console.log(`Purged ${res.rowCount} rows`);
-      }
-      if (err) {
-        return console.warn(err.message);
-      }
-    });
+    return this.client.query('DELETE FROM games WHERE created_time < now() - interval \'1 day\' * $1', [days])
```
bafolts, line 205: This function doesn't use `await` so doesn't need to be marked `async`. Only a function that uses `await` needs to be `async.

bafolts (approved):

src/database/PostgreSQL.ts
kberg: I hope I got it right now.

src/database/PostgreSQL.ts
kberg: Fixed, PTAL. This actually came about due to a vscode recommendation.

kberg:
Please peek again, sorry. I clearly don't know what I'm doing.

src/database/PostgreSQL.ts
bafolts: The `async` keyword is only needed if function has an `await` keyword inside it. There is a linting rule one can enable for this as well. 

https://palantir.github.io/tslint/rules/no-async-without-await/

Since this function is only using `.then` and `.catch` it doesn't need the `async` keyword.

src/database/IDatabase.ts
```diff
@@ -42,7 +42,7 @@ export interface IDatabase {
     /**
      * Creates any tables needed
      */
-    initialize(): Promise<void>;
+    initialize(): Promise<any>;
```
bafolts, line 45: Does this need to use `any`? Could it use `unknown` instead?

bafolts (approved):

PR #4467, opened 2022-05-19
Target es6.

kberg:
From https://www.typescriptlang.org/tsconfig#target:

> Modern browsers support all ES6 features, so ES6 is a good choice. You might choose to set a lower target if your code is deployed to older environments, or a higher target if your code is guaranteed to run in newer environments.


bafolts (approved):
The comment by typescript is very optimistic. Worst case if we see enough issues show up we can revert this. I don't think we will know who this is going to impact until we change it.

PR #4468, opened 2022-05-19
This one trick to lower memory for our LogMessage will surprise you.

bafolts:
This change avoids setting the `playerId` property on the `LogMessage` object when the argument is `undefined`. With less properties used per `LogMessage` less memory will be used. As JSON doesn't store properties with values of `undefined` this should also reflect what we have stored in JSON within the database. I added a regression test to ensure the generated javascript no longer adds the property.

## Before
```js
class LogMessage {
    constructor(type, message, data, playerId) {
        this.type = type;
        this.message = message;
        this.data = data;
        this.playerId = playerId;
        this.timestamp = Date.now();
    }
}
 ```

## After
```js
var LogMessage = (function () {
    function LogMessage(type, message, data, playerId) {
        this.type = type;
        this.message = message;
        this.data = data;
        this.timestamp = Date.now();
        if (playerId !== undefined) {
            this.playerId = playerId;
        }
    }
    return LogMessage;
}());
 ```

### DIFF
--- a/src/common/logs/LogMessage.ts
+++ b/src/common/logs/LogMessage.ts
@@ -5,11 +5,19 @@ import {PlayerId} from '../Types';
 
 export class LogMessage implements Message {
   public timestamp = Date.now();
+  public playerId?: PlayerId;
   constructor(
     public type: LogMessageType,
     public message: string,
     public data: Array<LogMessageData>,
     // When set, this message is private for the specifed player.
     // Always filter messages so they're not sent to the wrong player.
-    public playerId?: PlayerId) {}
+    playerId?: PlayerId) {
+    // setting in body to avoid setting property when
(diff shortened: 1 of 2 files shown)


kberg (approved):
When this game gets serialized/deserialized, does it retain this? I might suggest that we do something in serialize/deserialize to make sure someone doesn't add it back.

bafolts:
The serialize and de-serialize should maintain this. JSON doesn't have a an
undefined so the playerId isn't stored when serialized. This change will
only decrease memory when we create LogMessage instances on the server.
Even with 1 byte saved, with thousands of LogMessage you have soon saved
kilobytes of memory on the server. I plan to do some more small tweaks such
as this in more PRs. With local testing all the minor changes saved me
10-20kb for the game with the largest save id value.

PR #4473, opened 2022-05-21
only store property when not LogMessageType.DEFAULT

bafolts:
This change only stores the `type` property when the value is not `LogMessageType.DEFAULT`. The only messages that have a different type are the new generation log messages. This should remove this property from a large percentage of log messages. With the property not applied less memory will be consumed on the server as well as in the database. Performed quick sanity check locally and still saw generation messages appear and survive server reboot.


kberg (approved):
This is good! We should definitely do some pre-post measurements when the time comes.

PR #4474, opened 2022-05-21
refactor: remove unused conditionals for src/turmoil

bafolts:
I want to enable the [no-unnecessary-condition](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-unnecessary-condition.md) linting rule as it can spot code we can remove. These places usually pop up during re-factors when properties that were once optional gain default values. Going to do this change in small pieces to make it easier to review before enabling the rule. The changes in this PR were fixed after eslint found uncessary conditions.

This PR fixes the files in src/turmoil subdirectory.


kberg (approved):
I like this.

bafolts:
I plan to keep doing these until they are all fixed. At that point I will
enable the rule.

PR #4477, opened 2022-05-22
remove unnecessary pretest:server script

bafolts:
Since we are using `ts-mocha` we shouldn't need to transpile the typescript to javacript anymore. This will speed up the build.


kberg:
I use pretest sometimes to verify that test changes, well, they compile, without launching tests. Let's see if I still need it after this.

kberg (approved):

PR #4500, opened 2022-05-27
Add stats() method to IDatabase. Show at startup.

kberg:
Ideally this should work with an admin web page, but one thing at a time.

Example output:

```
Connecting to Postgres database.
node_modules/console-stamp/index.js:38
{
  "type": "POSTGRESQL",
  "pool-total-count": 1,
  "pool-idle-count": 1,
  "pool-waiting-count": 3,
  "size-bytes-games": "32 kB",
  "size-bytes-game-results": "16 kB",
  "size-bytes-database": "8065 kB"
}
node_modules/console-stamp/index.js:38
Starting ea600e87f, built at Wed, 25 May 2022 23:43:37 -0400
node_modules/console-stamp/index.js:38
Starting server on port 8080
```

### DIFF
--- a/src/database/PostgreSQL.ts
+++ b/src/database/PostgreSQL.ts
@@ -8,6 +8,7 @@ import {Pool, ClientConfig, QueryResult} from 'pg';
 
 export class PostgreSQL implements IDatabase {
   protected client: Pool;
+  private databaseName: string | undefined = undefined; // Use this only for stats.
 
   constructor(
     config: ClientConfig = {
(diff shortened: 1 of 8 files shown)


src/database/PostgreSQL.ts
```diff
@@ -19,20 +20,26 @@ export class PostgreSQL implements IDatabase {
...
+      } catch (e) {
+        console.log(e);
+      }
+    }
+    // Configuration stats saved for
     this.client = new Pool(config);
   }
 
-  public initialize(): Promise<void> {
+  public initialize(): Promise<QueryResult<any>> {
```
kberg, line 38: Now that I'm writing a test that uses game_results, I see that wow, the current promise infra doesn't actually create the second table. So, at least now I know it does.

bafolts (approved):

PR #4501, opened 2022-05-28
Move getGameVersion to a promise.

kberg:
This one only impacts the export tool, so it's fortunately low risk.


src/database/PostgreSQL.ts
```diff
@@ -167,17 +167,16 @@ export class PostgreSQL implements IDatabase {
...
-    });
+  getGameVersion(game_id: GameId, save_id: number): Promise<SerializedGame> {
+    return this.client.query('SELECT game game FROM games WHERE game_id = $1 and save_id = $2', [game_id, save_id])
+      .then((res) => {
+        if (res.rowCount === 0) {
+          throw new Error(`Game ${game_id} not found at save_id ${save_id}`);
+        }
+        return JSON.parse(res.rows[0].game);
+      }).catch((err) => {
+        throw err;
```
bafolts, line 178: I have been seeing this often and it seems to be showing up more often. I am pretty sure this doesn't do anything. If we didn't have this `catch` the error would get thrown just the same.

bafolts (approved):

src/database/PostgreSQL.ts
kberg: OK.

PR #4504, opened 2022-05-28
remove setting properties to undefined for Player

bafolts:
Avoids setting properties to `undefined` on the `Player` class. Unnecessary and takes up small amount of memory and CPU.


kberg:
It'll probably also reduce disk space

kberg (approved):

PR #4510, opened 2022-05-29
performance improvements to getAdjacentSpaces

bafolts:
[One of the unit tests](https://github.com/terraforming-mars/terraforming-mars/blob/main/tests/boards/Board.spec.ts#L302) can sometimes run over 2 seconds when I run `test:server`. This test creates 4000 board instances. I started to look into how this could be optimized. 

Noticed a minor change to our check for computing adjacent spaces that should help a bit. We would check for the x and y coordinate of the space last when that is the check that is more likely to fail. Moving these up front helped this unit test in question run 30% faster for me when ran in isolation with `it.only`.

The other change avoids duplicating the array and instead uses `ReadonlyArray` so that readonly is enforced at compile time versus having to copy the array each time using CPU.

### DIFF
--- a/src/boards/Board.ts
+++ b/src/boards/Board.ts
@@ -86,8 +86,8 @@ export abstract class Board {
       const spaces: Array<ISpace> = [];
       for (const [x, y] of coords) {
         const adj = this.spaces.find((adj) =>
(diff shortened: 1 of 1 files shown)


kberg (approved):
This is a nice change. Feel free to toss a few more ReadonlyArrays around if you like.

PR #4516, opened 2022-06-01
Move getSaveIds from test method to DB method.

kberg:
It simplifies export_game by removing erroneous reads.


kberg:
I'm gonna automerge this because it is completely isolated, doesn't touch anything but the export_game tool.

PR #4519, opened 2022-06-02
Remove method from setup.ts.

kberg:


kberg:
This is the types issue. :(

PR #4520, opened 2022-06-02
Load finished games at startup.

kberg:
After server restart, old games don't show up again. This will load them
up for access.

We could optimize this later to conditionally load finished games.


src/database/PostgreSQL.ts
```diff
@@ -63,7 +63,7 @@ export class PostgreSQL implements IDatabase {
 
   getGames(cb: (err: Error | undefined, allGames: Array<GameId>) => void) {
     const allGames: Array<GameId> = [];
-    const sql: string = 'SELECT games.game_id FROM games, (SELECT max(save_id) save_id, game_id FROM games WHERE status=\'running\' GROUP BY game_id) a WHERE games.game_id = a.game_id AND games.save_id = a.save_id ORDER BY created_time DESC';
+    const sql: string = 'SELECT distinct game_id FROM games';
```
bafolts, line 66: This query had `ORDER by created_time` to ensure that the most recent games are loaded first. Does this change also need to remove this ordering?

kberg: Well I did miss that, but before I restore it, why does order matter? The method is blocked until it's all loaded.

kberg: I did restore it, but hey, lemme know. :D

bafolts: We [load games](https://github.com/terraforming-mars/terraforming-mars/blob/main/src/database/GameIds.ts#L42) in the `GameIds` class. This should end up populating the `games` and `participantIds` map sequentially based on the order of this query. It seems this was broken [here](https://github.com/terraforming-mars/terraforming-mars/pull/3475/files) so we are no longer loading games while we are waiting to load every game id. Probably doesn't matter if this is ordered anymore then. We used to allow users to load games while we were waiting for `getGames` to populate all game ids. In production we can have thousands of game ids we have to load while the server is restarting.

Considering this new information this change is going to impact how long it takes until players can take an action while the server is restarting. When the server is restarting we are going to get a lot of players "waiting in line" while we are running these queries. We use to process these while loading game ids and the ordering was a trick to hopefully load games that were being played without having to wait for every game to load.

kberg: Well, that sounds like a problem them. Do you want to try to restore that behavior?

I didn't restore the query yet beacuse I'm slightly stubborn, and honestly, I think at this point perhaps we should shelve this and fix the async loading, which I also thought was happening.

What do you think?

PR #4534, opened 2022-06-05
Make card selection translucent

StaticBR:
Reduce the opacity of the card selection, to be able to read all information of the selected card.
This is especially helpful if there is only one card which is directly selected.

### DIFF
--- a/src/styles/cards.less
+++ b/src/styles/cards.less
@@ -117,7 +117,6 @@ input[type="radio"]:checked + .filterDiv::after {
     font-size: 36px;
     background: linear-gradient(to right, #c97e23bb, #e28c22bb, #c97e23bb);
     color: black;
-    border-bottom: 1px solid #222;
 }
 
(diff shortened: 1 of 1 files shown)


StaticBR:
Here a screenshot of the new card selection: grafik

kberg:
Thanks for this! Another PR (#4347) also tried this, but was rebutted because it turns out that bit of CSS interferes with colony and party selection. Take a look at [this specific comment](https://github.com/terraforming-mars/terraforming-mars/pull/4347#issuecomment-1107705507) for example. How does your change impact those components?

StaticBR:
@kberg 
Colonies  and Turmoil are not effected by this change:
grafik
grafik

kberg:
Thanks. This is terrific. At some point, perhaps we can do the more subtle approach, but I'm just not going to work on it, so this is great.

kberg (approved):

PR #4535, opened 2022-06-05
Rising z-index of top bar

StaticBR:
Rise z-index of top-bar  top 105 in order to be higher than the z-index of a hovered card (100).
This takes care that the hovered card does not show over the top-bar
Old behavior:
grafik

### DIFF
--- a/src/styles/player_home.less
+++ b/src/styles/player_home.less
@@ -938,7 +938,7 @@
     position: sticky;
     margin: -40px 0 0 auto;
     margin-bottom: 15px;
-    z-index: 20;
+    z-index: 105;
     top: -12px;
     display: flex;
 


kberg:
Thanks! Can you give a screencap of the new behavior?

src/styles/player_home.less
```diff
@@ -938,7 +938,7 @@
     position: sticky;
     margin: -40px 0 0 auto;
     margin-bottom: 15px;
-    z-index: 20;
+    z-index: 105;
```
kberg, line 941: How did you come up with 105? What elements were at 104?

StaticBR:
> Thanks! Can you give a screencap of the new behavior?
The hovered card is behind the top-bar:
grafik

src/styles/player_home.less
StaticBR: 100 is z-index of the card hover. So 101 would be enough, but i left some buffer. So it ended up being 105

kberg:
Oh, I see.. this is when the selection is scrolled up real high:

image

Approved. Thanks for this!

kberg (approved):

Issue #4097, opened 2022-02-13
[Colonies] Fix Market Manipulation visual


kberg:
![image](https://user-images.githubusercontent.com/413481/153770808-6eaf8d0d-8adc-408d-b40d-6a2af0376903.png)

Replace with showing the colonies

kberg:
Fixed!

PR #4544, opened 2022-06-07
Market Manipulation Action Changes

justin-dupre:
Prompt the user using the `SelectColony` input instead of the checkbox text options.

Issue https://github.com/terraforming-mars/terraforming-mars/issues/4541

Before: 
Screen Shot 2022-06-07 at 1 24 24 AM

After: Prompts the user to select which colony to increase, then which colony to decrease.
Screen Shot 2022-06-07 at 12 51 54 AM

Screen Shot 2022-06-07 at 12 52 01 AM

### DIFF
--- a/src/cards/colonies/MarketManipulation.ts
+++ b/src/cards/colonies/MarketManipulation.ts
@@ -63,33 +63,38 @@ export class MarketManipulation extends Card implements IProjectCard {
   }
 
   public play(player: Player) {
-    const increasableColonies = this.getIncreasableColonies(player.game);
+    let increasableColonies = this.getIncreasableColonies(player.game);
     const decreasableColonies = this.getDecreasableColonies(player.game);
+    // if there is only one decreaseable colony and it is an increaseable colony, don't allow increase of that colony.
+    if (decreasableColonies.length === 1 && increasableColonies.some((colony) => colony.name === decreasableColonies[0].name)) {
+      increasableColonies = increasableColonies.filter((colony) => colony.name !== decreasableColonies[0].name);
+    }
     const increaseColonyTrack = new SelectColony(
       'Select which colony tile track to increase',
       'Increase',
       increasableColonies,
-      (colony: IColony) => {
-        player.game.log('${0} increased ${1} track', (b) => b.player(player).string(colony.name));
-        colony.increaseTrack();
+      (increasedColony: IColony) => {
+        increasedColony.increaseTrack();
+        player.game.log('${0} increased ${1} track', (b) => b.player(player).string(increasedColony.name));
+        const decreaseColonyTrack = new SelectColony(
+          'Select which colony tile track to decrease',
+          'Decrease',
+          decreasableColonies.filter((decreaseableColony) => decreaseableColony.name !== increasedColony.name),
+          (colony: IColony) => {
+            player.game.log('${0} decreased ${1} track', (b) => b.player(player).string(colony.name));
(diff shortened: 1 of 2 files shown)


src/cards/colonies/MarketManipulation.ts
```diff
@@ -22,7 +22,11 @@ export class MarketManipulation extends Card implements IProjectCard {
       metadata: {
         cardNumber: 'C23',
         renderData: CardRenderer.builder((b) => {
-          b.text('Increase one colony tile track 1 step. Decrease another colony tile track 1 step.', Size.SMALL, true);
+          b.text(
```
kberg, line 25: Keep in mind that autoformatting makes reviews much more difficult.

src/cards/colonies/MarketManipulation.ts
```diff
@@ -34,50 +38,59 @@ export class MarketManipulation extends Card implements IProjectCard {
...
-            },
-          );
-
-          selectColonies.options.push(colonySelect);
-        }
-      });
-    });
-
-    return selectColonies;
+    const increaseColonyTrack = new SelectColony(
```
kberg, line 68: TL;DR: doing option 1 below will remove some subtle issues with this approach.

This is overall a good change, but introduces some subtle issues, which can be confirmed by the the card's entry in the unofficial FAQ:

image

The change as-is doesn't adhere to requiring the two colonies being different.

I see two options:

1. Instead of deferring both actions at once, defer one, and then in the callback for the first, defer the second.
This one introduces a new edge case bug: imagine only one colony can be decreased, but it's chosen as the colony to increase, so...
1a. Add this check: if an increasable colony is the _sole decreasable colony_ do not include it as an increasble colony.

2. return AndOptions with two entries [SelectColony, SelectColony]. This has the added benefit of letting a player choose both at once. It has the added cost of requiring some extra steps in the callback to act upon the add/remove operations.

This one has some subtle retry edge cases that make it slightly more complicated because the validation has to occur on the server side.

src/cards/colonies/MarketManipulation.ts
```diff
@@ -34,50 +38,59 @@ export class MarketManipulation extends Card implements IProjectCard {
...
+        colony.increaseTrack();
+        return undefined;
+      },
+    );
+    const decreaseColonyTrack = new SelectColony(
+      'Select which colony tile track to decrease',
+      'Decrease',
+      decreasableColonies,
+      (colony: IColony) => {
+        player.game.log('${0} decreased ${1} track', (b) => b.player(player).string(colony.name));
```
kberg, line 83: image

By changing the log messages, these two translations will be out of sync. Nothing to do here - this is a good change, and it only impacts two localizations that have translated this string.

kberg (commented):
LMK if you want some help here. I see exactly how to do it, so if you struggle with it lmk.

src/cards/colonies/MarketManipulation.ts
justin-dupre: Thanks for the input. I did miss this rule and edge case and went with your option 1 to fix it. Let me know if that's what you were thinking.

src/cards/colonies/MarketManipulation.ts
justin-dupre: Apologies for this; I will keep this in mind for future contributions

kberg (approved):
Nice work. Thank you!

kberg:
This also Fixes #4097

PR #4546, opened 2022-06-08
Measure PostgreSQL save operations.

kberg:
This includes measuring when an insert results in an udpate (using xmax = 0)
and if or when it occurs during an undo operation.

All of this is to measure whether both upserts _actually occur_ and also
whether it's exclusively the purvue of undo operations.

### DIFF
--- a/src/database/PostgreSQL.ts
+++ b/src/database/PostgreSQL.ts
@@ -205,7 +205,8 @@ export class PostgreSQL implements IDatabase {
 
   restoreGame(game_id: GameId, save_id: number, cb: DbLoadCallback<Game>): void {
     // Retrieve last save from database
-    this.client.query('SELECT game game FROM games WHERE game_id = $1 AND save_id = $2 ORDER BY save_id DESC LIMIT 1', [game_id, save_id], (err, res) => {
+    logForUndo(game_id, 'restore to', save_id);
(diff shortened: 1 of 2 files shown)


src/database/PostgreSQL.ts
```diff
@@ -9,6 +9,14 @@ export class PostgreSQL implements IDatabase {
   protected client: Pool;
   private databaseName: string | undefined = undefined; // Use this only for stats.
 
+  // I don't want this public; it's for testing. :/
+  public statistics = {
```
bafolts, line 13: Can you make it protected and provide access from tests?

src/database/PostgreSQL.ts
```diff
@@ -232,15 +240,35 @@ export class PostgreSQL implements IDatabase {
 
   async saveGame(game: Game): Promise<void> {
     const gameJSON = game.toJSON();
+    this.statistics.saveCount++;
```
bafolts, line 243: I would do this within `TestPostgreSQL` is possible if this is only for tests.

bafolts (approved):

src/database/PostgreSQL.ts
kberg: It's not for tests, it's meant to be measured outside.

PR #4549, opened 2022-06-10
Make getPlayerCount async

kberg:
This required a couple of external changes. For instance, introducing an
Async version of a handler, so tests could wait for results.

Also fixed an irritating issue, looks like I did a poor job of
saving and restoring the fake database test. Ha.

### DIFF
--- a/src/routes/Handler.ts
+++ b/src/routes/Handler.ts
@@ -44,13 +44,28 @@ export abstract class Handler implements IHandler {
     }
   }
 
-  get(req: http.IncomingMessage, res: http.ServerResponse, ctx: IContext): void {
+  get(req: http.IncomingMessage, res: http.ServerResponse, ctx: IContext): void | Promise<void> {
     ctx.route.notFound(req, res);
   }
-  put(req: http.IncomingMessage, res: http.ServerResponse, ctx: IContext): void {
+  put(req: http.IncomingMessage, res: http.ServerResponse, ctx: IContext): void | Promise<void> {
     ctx.route.notFound(req, res);
   }
-  post(req: http.IncomingMessage, res: http.ServerResponse, ctx: IContext): void {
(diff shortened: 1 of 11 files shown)


src/routes/Handler.ts
```diff
@@ -44,13 +44,28 @@ export abstract class Handler implements IHandler {
     }
   }
 
-  get(req: http.IncomingMessage, res: http.ServerResponse, ctx: IContext): void {
+  get(req: http.IncomingMessage, res: http.ServerResponse, ctx: IContext): void | Promise<void> {
```
bafolts, line 47: With the `void | Promise<void>` syntax one can use `await handler.get()`. Either will work with `void | Promise<void>`. This also avoids having to return `Promise.resolve` in source. Was this change needed?

kberg: This is a completely fair question.

My goal is to move all the handlers to Promise<void> but this was the only way I could think of doing it in transit. What do you think?

bafolts: Using `void | Promise<void>` should be fine. When we are using a `Promise` I would use `await` versus `.then`. If using `await` there is more flexibility as you can `await` something that isn't a `Promise` and a `Promise` will be created. If a route doesn't do anything asynchronous it could then not have to have `return Promise.resolve()` at the end. This is less code we have to write and maintain.

I don't think this is a huge deal but we will end up with a lot of functions having `return Promise.resolve()` when they could still be `void`. The caller should be using `await` so it doesn't matter if we return void.

bafolts (approved):

src/routes/Handler.ts
kberg: I think I'm still missing how these things all work together in a fundamental mechanical way. I'm still for the most part pattern matching and hoping the tests work. Do you have a recommendation for something that can help me really internalize all of this?

PR #4550, opened 2022-06-10
Drop dedicated solo page, replace with "How to play" link.

kberg:
image


kberg:
Links to https://github.com/terraforming-mars/terraforming-mars/wiki/Rulebooks which has further links to rulebooks and videos.

nwai90 (approved):

kberg:
It's not done yet, it seems.. The multiplayer page hides the player name / color selector, and also the beginner options that were recently discussed.

PR #4558, opened 2022-06-16
Work in progress: Make restoreGame async

kberg:


kberg:
@bafolts I need help here. For some reason `PlayerInput.spec.ts` does not pass. When I run the tests in a debugger, it gets all the way to `await post;` and then fails. The debugger's way of dealing with it is to just shut down the test, though running `npm run test:server` does seem to just complain about tests taking too long.

kberg:
In addition, `PlayerInput.processInput`, which handles the events, don't fire:

 ```
      req.on('data', async (data) => {
*        body += data.toString();
      });
      req.once('end', async () => {
        try {
*          const entity = JSON.parse(body);
          if (this.isWaitingForUndo(player, entity)) {
            await this.performUndo(req, res, ctx, player);
          } else {
```

I have breakpoints at the lines with `*` and they never fire.

File: tests/routes/PlayerInput.spec.ts [hunk 1]
```diff
@@ -35,18 +37,21 @@ describe('PlayerInput', function() {
...
     options.options.push(new UndoActionOption());
-    scaffolding.ctx.gameLoader.restoreGameAt = function(_gameId: string, _lastSaveId: number, cb: (game: Game | undefined) => void) {
-      cb(undo);
-    };
-    scaffolding.post(PlayerInput.INSTANCE, res);
-    scaffolding.req.emit('data', JSON.stringify([[String(options.options.length - 1)], ['']]));
-    scaffolding.req.emit('end');
+    scaffolding.ctx.gameLoader.restoreGameAt = (_gameId: string, _lastSaveId: number) => Promise.resolve(undo);
+    const post = scaffolding.asyncPost(PlayerInput.INSTANCE, res);
+    req.emit('data', JSON.stringify([[String(options.options.length - 1)], ['']]));
```
bafolts, line 42: This emission is happening too early before we are listening. This is happening because `Promise.resolve()` ands up running on the next _tick_. 

This snippet may help show the point:

```
Promise.resolve().then(() => console.log("B"));
console.log("A");
```

This outputs "A" and then "B" since `Promise.resolve()` doesn't run until the next tick.

With this in mind I suggest the following solution.

```suggestion
const emit = Promise.resolve().then(() => {
    req.emit('data', JSON.stringify([[String(options.options.length - 1)], ['']]));
    req.emit('end');
});

try {
   await Promise.all([emit, post]);
}
```

kberg: Thank you!

kberg:
Creating a fresh PR

PR #4587, opened 2022-06-23
Upload solar phase

bradwerth:
This ensures that the Solar Phase Option is *really* set after all the other settings, because the watch triggers change the results structure (somehow).


src/client/components/create/CreateGameForm.vue
```diff
@@ -626,6 +626,10 @@ export default (Vue as WithRefs<Refs>).extend({
           component.showColoniesList = results['customColoniesList'].length > 0;
           component.showCardsBlackList = results['cardsBlackList'].length > 0;
 
+          // Capture the solar phase option since several of the other results will change
```
kberg, line 629: This is nice. Do think there are other fields that should have the same behavior applied to it?

kberg (approved):

src/client/components/create/CreateGameForm.vue
bradwerth: It's the only one I've noticed is unset in the upload. I confess I don't quite understand the mechanism of it being changed -- I assume by the watch code but I'm not sure.

kberg: Well this is still a good change, thank you.

PR #4588, opened 2022-06-23
Corporations draft

pierrehilbert:
A new option is available: being able to have a big open hand draft corporations.
The mechanism is quite simple: 
- Everyone can see what are corporations available for this game
- Last player is picking first a corp then the one before the last, etc.
- When first player pick a corp, he will pick a second one immediately and we draft in the other way
- If more than two corps, doing the same when arrive to the last player.
After this corp draft is finished, we continue in the normal way: draft if selected or research phase.


src/common/models/GameModel.ts
```diff
@@ -36,4 +37,5 @@ export interface GameModel {
   turmoil: TurmoilModel | undefined;
   undoCount: number;
   venusScaleLevel: number;
+  corporationsToDraft: Array<CardModel>;
```
kberg, line 40: You don't need a full card model, do you? CardName ought to do.

src/common/models/PlayerModel.ts
```diff
@@ -69,6 +69,7 @@ export interface PlayerViewModel extends ViewModel {
   dealtCorporationCards: Array<CardModel>;
   dealtPreludeCards: Array<CardModel>;
   dealtProjectCards: Array<CardModel>;
+  draftedCorporations: Array<CardModel>;
```
kberg, line 72: Same here: drafted cards don't have resources, so CardModel is probably not necessary.

kberg (commented):
I've completely ignored the meat of your PR so far. I'll get to it next. My hope is that we can find a way to remove some of the repetition or something.

Let's figure out a way to add some good tests.

src/Game.ts
```diff
@@ -955,6 +984,37 @@ export class Game {
...
+      } else {
+        this.getPlayerBefore(player)!.runDraftCorporationPhase(this.getPlayerBefore(player)!.name, cards);
+      }
+      this.draftRound++;
+      return;
+    }
+
+    // Push last card to last player depending of the way we are drafting
+    if (this.corporationDraftToNext) {
+      this.getPlayerAfter(player)!.draftedCorporations.push(...cards);
```
bafolts, line 1005: The `getPlayerAfter` and `getPlayerBefore` returns `undefined` if there is no result. This means the caller has to use `!` assuming the item is going to exist. I would have `getPlayerAfter` and `getPlayerBefore` throw an error if we don't find a player. Is there a case where when those functions return `undefined` that represents some expected game state?

kberg: I'm OK with that. Also, we're removing almost all uses of `!` to tell the compiler to ignore concerns that a value is undefined.

kberg (commented):
Update: I didn't get a look yesterday, but I'm starting this morning. Busy for the next few hours, but what I see is fine as a first step, and now that we have FOUR TYPES OF DRAFT (initial, initial prelude, corp, and intergenerational) next step will be to consolidate them. That doesn't have to happen here. But I'll still give everything a good stare until it makes sense.

kberg:
@pierrehilbert #4598 adds a unit test for drafting and can serve as a template for testing corp drafting.

The only problem here is that in my example, I get to set the deck before initiating the draft. I don't see a way to do that without adding more game options or something. That doesn't seem so good. Any thoughts?

src/Game.ts
```diff
@@ -397,17 +403,38 @@ export class Game {
...
+      game.startGame();
     }
 
     return game;
   }
 
+  // Function use to properly start the game: with project draft or with research phase
+  public startGame(): void {
+    // Initial Draft
+    if (this.gameOptions.initialDraftVariant) {
```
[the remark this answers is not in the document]
pierrehilbert: I like this idea!

kberg:
#4598 now has an initial card draft. Basically you can create a determinstic draw by mocking out `Dealer.shuffle` and using this snippet:

```
    const shuffle = Dealer.shuffle;
    let game: Game;
    try {
      // This keeps the cards in their original order. If necessary, this deck could be front-loaded instead.
      Dealer.shuffle = <T> (cards: Array<T>) => cards;
      game = newTestGame(2, {draftVariant: true, initialDraftVariant: true});
    } finally {
      Dealer.shuffle = shuffle;
    }
    const player = getTestPlayer(game, 0);
    const otherPlayer = getTestPlayer(game, 1);
```

@pierrehilbert if you are still stuck on this, let me know and I will help out. At least now I'm super confident a reasonable corporate drafting test is not unreasonable.

src/Game.ts
```diff
@@ -397,17 +403,38 @@ export class Game {
 
     game.log('Generation ${0}', (b) => b.forNewGeneration().number(game.generation));
 
-    // Initial Draft
-    if (gameOptions.initialDraftVariant) {
-      game.phase = Phase.INITIALDRAFTING;
-      game.runDraftRound(true);
+    // Do we draft corporations or do we start the game?
+    if (gameOptions.corporationsDraft) {
```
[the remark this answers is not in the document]
kberg: Well, let's start with whatever is simplest. :D

src/common/models/GameModel.ts
kberg: Just a reminder this is pending.

src/common/models/PlayerModel.ts
kberg: Just a reminder this is pending.

kberg:
This is really nice work. I can't wait to get it on the server.

OK so at this point I'm waiting on Pierre to fix a couple of things. We can go forward without a test, because I'm fairly confident that existing games won't be negatively impacted. (But I kinda want one. :D)

I'm going to push this to terraforming-mars-kberg@ to solicit feedback.

pierrehilbert:
Hey,
After some tests, I made some little mistakes in the way it's working so I will try to correct them today!
I think it will be a cool feature to add then.
Pierre

pierrehilbert:
Normally it should be okay now!

src/common/models/GameModel.ts
kberg: Can you do this?

kberg (commented):
Couple of things pending. Can you look at these?

kberg (approved):
Proceeding without the final bits I asked for.

PR #4589, opened 2022-06-24
Update ko translations.

derornos:
- promo 2022 summer (merger)
- venus-next, colonies, turmoil cards
- global event cards


kberg (approved):

kberg:
Thanks!

PR #4590, opened 2022-06-24
remove non-null assertion from GlobalEventDealer

bafolts:
Saw non-null assertions being used in #4588 . Was surprised to see them! Then realized we didn't have a linting rule in place preventing them. Hoping to enable [no-non-null-assertion](https://typescript-eslint.io/rules/no-non-null-assertion/). There are 20 or so places we are using these in `/src`. This removes them from the `GlobalEventDealer`.


kberg (approved):
Thanks

PR #4592, opened 2022-06-24
Fix market manipulation log

justin-dupre:
Noticed while playing the log said the same colony was increased and decreased. This was just an issue with the log. Changed variable name from `colony` to `decreasedColony` to make it more clear.


kberg (approved):
Nice, thank you.

PR #4605, opened 2022-06-28
Make Random SeededRandom with an interface.

kberg:
The goal will be to make testing the Dealer simpler, which will apply to
making it possible to test the initial draft.


kberg:
I don't know if I like this. I mean, it'll be good to have a reliable draw for tests. Somehow I'm worried this will cause issues with cloning games. I don't see much other reason to dislike it though.

To be honest, anything that removes non-determinism, blah blah blah.

(I also found an acceptable workaround.)

bafolts (approved):

PR #4615, opened 2022-07-01
Update description of card "Declaration of Independence"

ni0ki:
The current text description of the card is a bit misleading.
Align it with other cards that explicitly state the requirement is based only on player's tags (such as Luna Political Institute).


nwai90:
@ni0ki Note that the text also needs to be updated in translation files, such as `src/locales/de/pathfinder_cards.json`

ni0ki:
Thanks for the advice @nwai90. I updated it. Seems no other language implement the translations for pathfinders.

kberg (approved):

PR #4630, opened 2022-07-05
Make purging unfinished games async.

kberg:


src/database/IDatabase.ts
```diff
@@ -144,7 +144,7 @@ export interface IDatabase {
      * * In Sqlite, it doesn't purge
      * * This whole method is ignored in LocalFilesystem.
      */
-    purgeUnfinishedGames(maxGameDays?: string): void;
+    purgeUnfinishedGames(maxGameDays?: string): Promise<void>;
```
bafolts, line 147: I am surprised typescript allowed this to compile prior to this change since `PostgreSQL` was returning `Promise<void>` and not `void`. :thinking: . Will need to _learn_ why that was. What is here is right though.

bafolts (approved):

src/database/IDatabase.ts
kberg: I believe it's because any type is a valid subtype of void.

PR #4640, opened 2022-07-05
Narrow the types for {Game,Player,Spectator}Id.

kberg:


src/database/LocalFilesystem.ts
```diff
@@ -13,7 +13,7 @@ export class LocalFilesystem implements IDatabase {
   private readonly historyFolder: string;
   constructor(dbFolder: string = defaultDbFolder) {
     this.dbFolder = dbFolder;
-    this.historyFolder = path.resolve(defaultDbFolder, 'history');
+    this.historyFolder = path.resolve(dbFolder, 'history');
```
kberg, line 16: This is just a discovery of an issue with the integration test. :D

PR #4646, opened 2022-07-06
Fix undo, maybe.

kberg:
This is still a prototype, so only provide feedback.

I believe what is happening is that game.save() happens more than it should. AND, it happens in an asynchronous manner that seemed super weird. And eventually what happens is that in between player turns the game gets saved TWICE, yet somehow one of the saves doesn't occur, but the save ID increments. So that's strange, right? I think that has to do with game.save not being synchronous.

This also attempts a simpler version of undo (which doesn't delete the last save, but again, this is a prototype.)

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1852,7 +1852,7 @@ export class Player {
    * words, don't set this value unless you know what you're doing.
    */
   // @ts-ignore saveBeforeTakingAction is unused at the moment.
-  public takeAction(saveBeforeTakingAction: boolean = true): void {
+  public async takeAction(saveBeforeTakingAction: boolean = true): Promise<void> {
     const game = this.game;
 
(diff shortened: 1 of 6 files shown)


src/Player.ts
```diff
@@ -1862,8 +1862,8 @@ export class Player {
 
     const allOtherPlayersHavePassed = this.allOtherPlayersHavePassed();
 
-    if (this.actionsTakenThisRound === 0 || game.gameOptions.undoOption) game.save();
-    // if (saveBeforeTakingAction) game.save();
+    // if (this.actionsTakenThisRound === 0 || game.gameOptions.undoOption) game.save();
```
kberg, line 1865: I don't know if this is the actual problem.

kberg:
Got a better fix. Closing this one.

PR #4658, opened 2022-07-08
add linting rule require-await

bafolts:
Adding the [require-await](https://eslint.org/docs/latest/rules/require-await) lint rule. This rule can help developers new to async and await. Also useful during refactors to catch usages that become unnecessary.


kberg (approved):
This is great. Thank you.

PR #4659, opened 2022-07-08
Slice instance loading by 1000s. Prevents runtime failure.

kberg:


kberg:
Output reads like this
```
Loaded 37000 to 38000 of 43088
Loaded 38000 to 39000 of 43088
Loaded 39000 to 40000 of 43088
Loaded 40000 to 41000 of 43088
Loaded 41000 to 42000 of 43088
Loaded 42000 to 43000 of 43088
Loaded 43000 to 43088 of 43088
```

Tested against a copy of the production database which demonstrated how this would fail.

src/database/GameIds.ts
```diff
@@ -23,10 +23,16 @@ export class GameIds extends EventEmitter {
     }
   }
 
-  private getAllInstances(allGameIds: Array<GameId>) : Promise<void[]> {
-    return Promise.all(allGameIds.map((x) => {
-      return this.getInstance(x);
-    }));
+  private async getAllInstances(allGameIds: Array<GameId>): Promise<void> {
+    Metrics.INSTANCE.mark('game-ids-get-all-instances-started');
+    const sliceSize = 1000;
```
bafolts, line 28: I made up 1000 as it sounded like a good number and as humans we lean towards base 10. This was the only number I tried as well. May be worth including that in a comment. It worked and I was short on time so went with it.

bafolts (approved):

src/database/GameIds.ts
kberg: I want to make this a runtime environment variable and do some timings, but right now I'm more interested in survival. But yeah.

kberg:
Thanks!

PR #4662, opened 2022-07-09
remove a few unnecessary null chains

bafolts:
Removes some unnecessary null chains. These were spotted with the @typescript/eslint rule of no-unnecessary-conditions.


kberg (approved):

kberg:
Thanks!

PR #4672, opened 2022-07-10
Remove vti.

kberg:
It hasn't been updated in 8 months, and prevents us from upgrading to a
newer versions of Typescript.

vti the command-line companion to Vetur, the VSCode IDE tool. Vetur used
to be the recommended IDE tool but has been taken out of favor for
Volar.

While `vti diagnostics` was vital while moving to single file
components, it doesn't generate that many errors now. This is a good
compromise, and I continue to look for a replacement.

Thank you `vti`. You helped out when we needed you.


kberg:
Putting Do Not Merge on this. It looks as though vti got its update last night. I can test it out.

kberg:
VTI upgrade works. Closing this.

PR #4674, opened 2022-07-11
Speed up ID initialization.

kberg:
This creates a new table to track game IDs. It speeds up load time immensely. Loading up a copy of the production database running on my workstation:

old: 36 seconds
new: ~ 1s.

Keep in mind that the same operation on Heroku takes over 10 minutes, so this change
is even more dramatic. (10 minutes to est. 20 seconds?)

This isn't ready for general use, though. It's missing a very seamless way for new people to use it (Not to mention there's no SQLite implementation.)

But as is, this seems the right direction.


src/database/GameLoader.ts
```diff
@@ -1,7 +1,8 @@
 import {Database} from './Database';
 import {Game} from '../Game';
 import {PlayerId, GameId, SpectatorId, isGameId} from '../common/Types';
-import {GameIdLedger, IGameLoader} from './IGameLoader';
+import {IGameLoader} from './IGameLoader';
+import {GameIdLedger} from './IDatabase';
```
kberg, line 5: GameIdLedger became really useful and was moved over.

src/database/IDatabase.ts
```diff
@@ -2,6 +2,8 @@ import {Game, GameOptions, Score} from '../Game';
 import {GameId, PlayerId, SpectatorId} from '../common/Types';
 import {SerializedGame} from '../SerializedGame';
 
+export type GameIdLedger = {gameId: GameId, participants: Array<PlayerId | SpectatorId>}
```
kberg, line 5: GameIdLedger also had its first field renamed from id to gameId, which is why you'll see a few minor changes here and there.

src/database/PostgreSQL.ts
```diff
@@ -89,23 +91,9 @@ export class PostgreSQL implements IDatabase {
...
-        FROM games
-        WHERE save_id = 0 AND CAST(game AS JSON)->>'spectatorId' = $1`;
-    } else {
-      throw new Error(`id ${id} is neither a player id nor spectator id`);
-    }
-
     try {
-      const res = await this.client.query(sql, [id]);
+      // TODO(kberg): Add a Postgres GIN index on ids.
+      const res = await this.client.query('select game_id from game_ids where $1 in ids', [id]);
```
kberg, line 96: This method is now much simpler (and fortunately way way faster.)

src/database/PostgreSQL.ts
```diff
@@ -199,9 +189,8 @@ export class PostgreSQL implements IDatabase {
     this.statistics.saveCount++;
     if (game.gameOptions.undoOption) logForUndo(game.id, 'start save', game.lastSaveId);
     try {
-      // By pre-computing the next game ID we avoid certain race conditions where
-      // saveGame is called twice in a row.
-      const nextSaveId = game.lastSaveId + 1;
+      // Holding onto a value avoids certain race conditions where saveGame is called twice in a row.
+      const thisSaveId = game.lastSaveId;
```
kberg, line 193: This makes interacting with `game_ids` a little easier.

src/tools/create_game_ids_table.ts
```diff
@@ -0,0 +1,25 @@
+require('dotenv').config();
+
+import {GameLoader} from '../database/GameLoader';
+import {Database} from '../database/Database';
+
+const db = Database.getInstance();
+
+async function main() {
```
kberg, line 8: This is a useful successful tool that when run will populate the game_ids table. Someone could migrate simply by running this. I just want to make the migration a little more painless.

tests/utils/setup.ts
```diff
@@ -12,14 +12,15 @@ const FAKE_DATABASE: IDatabase = {
   getGameIds: () => Promise.resolve([]),
   getSaveIds: () => Promise.resolve([]),
   initialize: () => Promise.resolve(),
-  restoreGame: () => {
-    throw new Error('game not found');
-  },
+  restoreGame: () => Promise.reject(new Error('game not found')),
```
kberg, line 15: Not important, but it's more correct.

src/database/LocalFilesystem.ts
```diff
@@ -10,30 +10,30 @@ const defaultDbFolder = path.resolve(process.cwd(), './db/files');
 
 export class LocalFilesystem implements IDatabase {
   protected readonly dbFolder: string;
-  private readonly historyFolder: string;
+  readonly #historyFolder: string;
   constructor(dbFolder: string = defaultDbFolder) {
     this.dbFolder = dbFolder;
-    this.historyFolder = path.resolve(dbFolder, 'history');
+    this.#historyFolder = path.resolve(dbFolder, 'history');
```
kberg, line 16: This `#` means private is nice.

src/database/PostgreSQL.ts
```diff
@@ -89,23 +93,8 @@ export class PostgreSQL implements IDatabase {
   }
 
   public async getGameId(id: string): Promise<GameId> {
-    let sql = undefined;
```
kberg, line 92: This method is only used in export_game.ts, and is now much simpler thanks to the participants table.

src/database/PostgreSQL.ts
```diff
@@ -172,7 +161,9 @@ export class PostgreSQL implements IDatabase {
       const selectResult = await this.client.query('SELECT DISTINCT game_id FROM games WHERE created_time < $1', [dateToSeconds]);
       const gameIds = selectResult.rows.map((row) => row.game_id);
       const deleteResult = await this.client.query('DELETE FROM games WHERE game_id in ANY($1)', [gameIds]);
-      console.log(`Purged ${deleteResult.rowCount} rows`);
+      console.log(`Purged ${deleteResult.rowCount} rows from games`);
+      const idsResult = await this.client.query('DELETE FROM participants WHERE game_id in ANY($1)', [gameIds]);
```
kberg, line 165: When purging a game, have to purge the participants table.

src/database/SQLite.ts
```diff
@@ -24,34 +26,16 @@ export class SQLite implements IDatabase {
...
-              reject(err3);
-              return;
-            }
-            resolve();
-          });
-        });
-      });
-    });
+  async initialize(): Promise<void> {
+    await this.asyncRun('CREATE TABLE IF NOT EXISTS games(game_id varchar, players integer, save_id integer, game text, status text default \'running\', created_time timestamp default (strftime(\'%s\', \'now\')), PRIMARY KEY (game_id, save_id))');
```
kberg, line 30: this.asyncRun makes SQLite so much easier to deal with. No more nested garbage. Thanks for teaching me, @bafolts !

src/database/GameIds.ts
```diff
@@ -9,16 +9,25 @@ export class GameIds extends EventEmitter {
...
       if (game.spectatorId !== undefined) {
         this.participantIds.set(game.spectatorId, gameId);
+        participantIds.push(game.spectatorId);
       }
       for (const player of game.players) {
         this.participantIds.set(player.id, gameId);
+        participantIds.push(player.id);
+      }
+      try {
+        this.db.storeParticipants({gameId, participantIds});
```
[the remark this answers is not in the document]
bafolts: If a query were ran one time could the `participants` table be populated ahead of time? I would think one query could handle this update versus having to run the separate queries from javascript.

src/database/LocalFilesystem.ts
bafolts: I am not familiar with `#` for private variables. Which version of node.js first had support for them? This could impact node.js minimum supported runtime.

src/database/GameIds.ts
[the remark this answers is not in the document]
kberg: A single query could probably handle it but it's not necessary. There is also a command-line tool below that will do this, though not as a single operation.

Keep in mind that to do that requires SQL that pulls apart JSON strings, and does it reliably. That's why I didn't go down that path.

bafolts (approved):

bafolts:
For both SQLite and Postgres the migration tool could be one query with an INSERT INTO SELECT. This seems tailored for local file system which couldn't leverage that query which is strange to me but should work.

src/database/LocalFilesystem.ts
kberg: I'm removing it. I'm using ES2019, but for some reason it says it's required for ES2022. No idea what's happening here.

kberg:
> For both SQLite and Postgres the migration tool could be one query with an INSERT INTO SELECT. This seems tailored for local file system which couldn't leverage that query which is strange to me but should work.

I think you're missing the point. This has nothing to do with local filesystem (which doesn't need a participants table.)

Take a look at the code written for searching the table by player ID or spectator ID:

```
    let sql = undefined;
    if (id.charAt(0) === 'p') {
      sql =
        `SELECT game_id
          FROM games, json_array_elements(CAST(game AS JSON)->'players') AS e
          WHERE save_id = 0 AND e->>'id' = $1`;
    } else if (id.charAt(0) === 's') {
      sql =
        `SELECT game_id
        FROM games
        WHERE save_id = 0 AND CAST(game AS JSON)->>'spectatorId' = $1`;
    } else {
```

```
    // Default sql is for player id;
    let sql: string = 'SELECT game_id from games, json_each(games.game, \'$.players\') e where json_extract(e.value, \'$.id\') = ?';
    if (id.charAt(0) === 's') {
      sql = 'SELECT game_id from games where json_extract(games.game, \'$.spectatorId\') = ?';
    }
```

How would you propose this one-time operation would work? What kind of testing would you need to do? How fast do you think it would run? (I can tell you from personal experience it's very slow.)

Then there's how to deploy and migrate -- think through how a clean migration would work. How would you design the migration such that games aren't disrupted, or that don't require shutting down the server for maintenance so as to not disrupt existing games?

That's the beauty of this migration path. If you need to do it, it's once. It's slow, but it's not meaningfully slower than what you've already got. The next time you run, it's super fast, and might pick up any stragglers that might have been skipped the first time. Once you try to be efficient in that way, it's flexible for a static system, not for a running system.

PR #4680, opened 2022-07-12
Provide a default cache size? Allow caching during development.

kberg:


src/routes/ServeAsset.ts
```diff
@@ -1,3 +1,4 @@
+require('dotenv').config();
```
bafolts, line 1: Does this need to be required here?

kberg:
I think I want to abandon this.

PR #4685, opened 2022-07-14
Configure removing finished games from the running server.

kberg:


.env.sample
```diff
@@ -32,3 +32,38 @@
...
+
+## Specifies how the server removes games from memory after they're completed.
+##
+## Specified as a series of semicolon-separated <key>=<value> pairs, there are
+## three possible keys. All of them are optional, as is this environment
+## variable.
+##
+## sweep: (default='manual')
+##  Can be 'auto' or 'manual'. When 'auto', the sweeper will run on a regular
+##  basis to find and evict compmleted games. When 'manual' the server doesn't
```
bafolts, line 44:
```suggestion
##  basis to find and evict completed games. When 'manual' the server doesn't
```

src/database/GameIds.ts
```diff
@@ -87,4 +105,42 @@ export class GameIds extends EventEmitter {
...
+        this.evictionSchedule.delete(gameId);
+      }
+    }
+    console.log('Finished sweep');
+  }
+
+  private evict(gameId: GameId) {
+    const game = this.games.get(gameId);
+    if (game === undefined) return;
+    game.getPlayers().forEach((p) => (p as any)._game = undefined);
```
bafolts, line 131:
```suggestion
    game.getPlayers().forEach((p) => p.game = undefined);
```

This will involve updating the setter for `game` to take `Game | undefined` as well. Will avoid using `any` and we allow `game` to be `undefined` on `Player`.

src/database/GameIds.ts
```diff
@@ -87,4 +105,42 @@ export class GameIds extends EventEmitter {
...
+    this.games.set(gameId, undefined); // Setting to undefied is the same as "not yet loaded."
+  }
+}
+
+function scheduleSweep(cache: GameIds, sleepMillis: number) {
```
bafolts, line 136: If this is called manually we may need to call to `clearTimeout`. I would store the timeout from `setTimeout` and when this function is called then clear it. This will ensure subsequent calls to `scheduleSweep` don't start multiple sweepers.

src/database/GameLoader.ts
```diff
@@ -21,27 +31,37 @@ const initialize = new prometheus.Gauge({
...
+      GameLoader.instance = new GameLoader(config, new Clock());
     }
     return GameLoader.instance;
   }
 
+  public static newTestInstance(config: CacheConfig, clock: Clock): GameLoader {
+    return new GameLoader(config, clock);
+  }
+
+  public resetForTesting(): void {
```
bafolts, line 60: Other places we have leveraged a class within the unit test that extends the class to have methods for testing only. Here `idsContainer` could be made `protected` and the unit test class could override `reset`.

src/database/GameLoader.ts
```diff
@@ -101,4 +126,31 @@ export class GameLoader implements IGameLoader {
...
+  public mark(gameId: GameId) {
+    this.idsContainer.mark(gameId);
+  }
+
+  public sweep() {
+    this.idsContainer.sweep();
+  }
+}
+
+function parseConfigString(stringValue: string): CacheConfig {
```
bafolts, line 139: Is the format of this configuration string a standard for duration? Taking in seconds would avoid having this string parsing and could be more straight forward. Instead of passing in 15m users could pass in 900.

bafolts (changes requested):
This looks good and should allow the terraforming-mars.herokuapp.com server to go more than 24 hours without running out of memory 🥳 

I have a few minor comments. Requesting changes mostly for the `setTimeout` that doesn't have a corresponding `clearTimeout`. Manual users could eventually have multiple sweepers running which I don't think is intentional.

src/database/GameIds.ts
kberg: Yeah, sorry. I fixed it. PTAL. Making game take undefined will be a huge maintenance issue. A proper fix is to define Game before Player, and not the other way around.

There's another possibility which relies on `GameLoader.getGame` but that's slower and messier. A proper refactor seems the right step forward.

kberg: The solution I will push shortly seems the right compromise.

src/database/GameIds.ts
kberg: It's not exported, so it won't be called unnecessarily. Is this still an issue?

src/database/GameLoader.ts
kberg: Hi -- first, this is the `reset` method from above moved down and given a new name for clarity. This is a fine idea, but can be its own isolated change.

src/database/GameLoader.ts
kberg: You can pass '15m' or '900s' or '10m300s' or '.25h'.

src/Player.ts
```diff
@@ -162,6 +162,13 @@ export class Player {
     public handicap: number = 0,
     id: PlayerId) {
     this.id = id;
+    // This seems pretty bad. The game will be set before the Player is actually
+    // used, and if that doesn't happen, well, it's a worthy error.
+    // The alterantive, to make game type Game | undefined, will cause compilation
+    // issues throughout the app.
+    // Ideally the right thing is to invert how players and games get created.
+    // But one thing at a time.
+    this.game = undefined as unknown as Game;
```
bafolts, line 171: Before I recommended the change I updated the `set game` to take `Game | undefined` as it's argument and I didn't run into compilation issues throughout the app. The type of `_game` is already `Game | undefined`.

src/database/GameLoader.ts
bafolts: This looks close to https://en.wikipedia.org/wiki/ISO_8601#Durations. Was asking as I was curious if there was some other duration standard.

src/Player.ts
kberg: Yes, that's right. I removed the setter and getter.

kberg: Yes, that's right. I removed the setter and getter.

src/database/GameLoader.ts
kberg: It is close to that! There's no duration API until a later standard. And it also includes some irritating prefixes that I was hoping to avoid.

kberg: It is close to that! There's no duration API until a later standard. And it also includes some irritating prefixes that I was hoping to avoid.

bafolts (approved):

PR #4688, opened 2022-07-14
More version updates, based on `npm outated`

kberg:


kberg:
Huh, that worked locally. More later.

PR #4691, opened 2022-07-14
Clarify corporationDraftPlayerToNext

kberg:
@pierrehilbert this seems like a good change for readability.


src/Game.ts
```diff
@@ -197,7 +197,7 @@ export class Game {
   private initialDraftIteration: number = 1;
   private unDraftedCards: Map<PlayerId, Array<IProjectCard>> = new Map();
   // Used for corporation global draft: do we draft to next player or to player before
-  private corporationsDraftToNextPlayer: boolean = false;
+  private corporationsDraftDirection: 'before' | 'after' = 'after';
```
kberg, line 200: Actually this should probably be 'before'

bafolts (approved):

PR #4694, opened 2022-07-16
increase checks if card is playable

StaticBR:
Introduces new game methods in order to check if Oxygen and temperature can be Increased.
- Use this methods where reasonable
- Added Checks for a few action cards if they can be played or if the global condition has already been maxed out.

**Detail:**
- AquiferPumping / WaterImportFromEuropa = Only available if oceans are not maxed
- GHGProducingBacteria & RegolithEaters = Do not display options if scale is maxed.
- VenusMagnetizer = Only available if venus scale is not maxed.

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -500,8 +500,8 @@ export class Game {
   }
 
   public marsIsTerraformed(): boolean {
-    const oxygenMaxed = this.oxygenLevel >= constants.MAX_OXYGEN_LEVEL;
-    const temperatureMaxed = this.temperature >= constants.MAX_TEMPERATURE;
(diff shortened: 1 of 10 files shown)


kberg:
What's up? Why did you close this, @StaticBR?

StaticBR:
@kberg 
Because I wasn't aware that  it is intentional that you are able to play AquiferPumping / WaterImportFromEuropa even after the global scales are maxed out. So i have to do a clean rebase of my branch an remove this commits.

kberg:
That's fair. That gets people all the time. I wouldn't know it myself I
weren't hearing about it so often. :D

PR #4707, opened 2022-07-19
Simplify card callbacks.

kberg:
1: rely on [x] as an lvalue to simplify the code a bit.
2: changing the name `foundCards` to cards as most of these aren't
found.
3: remove the type declarations.


bafolts (approved):
Now this is some advanced typescripting. 👏

PR #4710, opened 2022-07-19
Start moving to a multi-corporation environment.

kberg:
This introduces `getCorporation` which, despite the eventual Merger
prelude is turning out to be a pretty useful slice of code.

`corporations` as a getter is a starting point for eventually moving to
a multiple corporation setup. It's possible this might not stay the
same name (might even move away from corporations to all played
cards. But it reduces ther reliance on a single `corporationCard`


src/Player.ts
```diff
@@ -750,17 +751,16 @@ export class Player {
...
   public getCardsWithResources(resource?: CardResource): Array<ICard & IResourceCard> {
-    let result: Array<ICard & IResourceCard> = this.playedCards.filter((card) => card.resourceType !== undefined && card.resourceCount && card.resourceCount > 0);
-    if (this.corporationCard !== undefined &&
-          this.corporationCard.resourceType !== undefined &&
-          this.corporationCard.resourceCount !== undefined &&
-          this.corporationCard.resourceCount > 0) {
-      result.push(this.corporationCard);
-    }
+    let result: Array<ICard & IResourceCard> = [
+      ...this.playedCards.filter((card) => card.resourceType !== undefined && card.resourceCount && card.resourceCount > 0),
```
bafolts, line 761: Seems now we may eventually want some function that lets us filter through both `playedCards` and `corporations`. Looking like it is going to be common to have to filter through both of them.

bafolts (approved):

src/Player.ts
kberg: Oh yes, that's coming. I mentioned it in https://github.com/terraforming-mars/terraforming-mars/issues/4698

PR #4716, opened 2022-07-25
Step 2 of moving to multi-corporation environment

kberg:


src/Player.ts
```diff
@@ -2250,26 +2267,32 @@ export class Player {
...
 
-    // Rebuild corporation card
-    if (d.corporationCard !== undefined) {
-      player.corporationCard = cardFinder.getCorporationCardByName(d.corporationCard.name);
-      if (player.corporationCard !== undefined) {
-        if (d.corporationCard.resourceCount !== undefined) {
-          player.corporationCard.resourceCount = d.corporationCard.resourceCount;
+    // Rebuild corporation cards
+    let corporations = d.corporations;
+    if (corporations === undefined && d.corporationCard !== undefined) corporations = [d.corporationCard];
```
bafolts, line 2272: I think this could have comment to remove this at certain point. At some point I don't think we would need this mapping anymore.

bafolts (approved):

src/Player.ts
kberg: There's a comment in serializedplayer for that. :)

PR #4734, opened 2022-07-30
Fix HowToPay bugs in processInput

kberg:
@Strikeskids gets full credit for #4340 . This just became a pain to merge after all the changes.

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -993,22 +993,19 @@ export class Player {
     }
   }
 
+  public isHowToPay(u: unknown): u is HowToPay {
+    if (typeof u !== 'object') return false;
+    if (!u) return false;
+    const h = u as {[key in keyof HowToPay]?: any};
+    return HowToPay.keys.every((key) =>
+      h.hasOwnProperty(key) && typeof h[key] === 'number' && !isNaN(h[key]));
+  }
+
   public parseHowToPayJSON(json: string): HowToPay {
-    const defaults: HowToPay = {
-      steel: 0,
-      heat: 0,
-      titanium: 0,
-      megaCredits: 0,
-      microbes: 0,
-      floaters: 0,
-      science: 0,
-      seeds: 0,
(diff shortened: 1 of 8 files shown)


src/Player.ts
```diff
@@ -993,22 +993,19 @@ export class Player {
     }
   }
 
+  public isHowToPay(u: unknown): u is HowToPay {
```
bafolts, line 996: Have to say. This is pretty crafty handle little pattern for checking type of interface. :clap:

src/Player.ts
```diff
@@ -993,22 +993,19 @@ export class Player {
...
+    if (typeof u !== 'object') return false;
+    if (!u) return false;
+    const h = u as {[key in keyof HowToPay]?: any};
```
bafolts, line 999:
```suggestion
    const h = u as {[key in keyof HowToPay]: number};
```

This would avoid a usage of `any`.

src/Player.ts
```diff
@@ -1761,46 +1732,80 @@ export class Player {
...
+      microbes: DEFAULT_MICROBES_VALUE,
+      floaters: DEFAULT_FLOATERS_VALUE,
+      science: 1,
+      seeds: constants.SEED_VALUE,
+      data: constants.DATA_VALUE,
+    };
+
+    const usable: {[key in keyof HowToPay]: boolean} = {
+      megaCredits: true,
+      steel: options?.steel ?? false,
```
bafolts, line 1771:
```suggestion
      steel: Boolean(options?.steel),
```

This is styling preference of mine and not important but I prefer `Boolean` in these situations. Reads better to me than all the `? ? ??? ?`. My preference is probably because `??` is a newer operator though.

bafolts (approved):

src/Player.ts
kberg: Let's stick with the other form for consistency?

PR #4757, opened 2022-08-05
insert cards while sorting versus swapping

### LINKED ISSUE #4731: Make drag and drop insert cards rather than swap.
From Discord:

> recently i realised that you can re-arrange the cards in hand.
> 
> tho, it only lets you switch places between 2 cards.
> would it be possible to change this so when i drag and drop a card, i instead place it between the 2 cards where i hold it?
> 
> this would make for easier planning and overview.

bafolts:
Updates the card sorting to _insert_ the card versus swapping the position. Also updated the card magnification to avoid magnifying cards while dragging and dropping.

This closes #4731

### DIFF
--- a/src/client/components/SortableCards.vue
+++ b/src/client/components/SortableCards.vue
@@ -64,8 +64,8 @@ export default Vue.extend({
           max = Math.max(max, this.cardOrder[key]);
         }
         this.cardOrder[this.dragCard] = max + 1;
-      // place it ahead of the card
       } else {
+        // place it ahead of the card
         const temp = this.cardOrder[source];
         const keys = Object.keys(this.cardOrder);
         for (const key of keys) {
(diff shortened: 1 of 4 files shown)


src/client/components/SortableCards.vue
```diff
@@ -53,21 +53,41 @@ export default Vue.extend({
...
+      }
+      // put the card at the end of the list
+      if (source === 'end') {
+        let max = 0;
+        const keys = Object.keys(this.cardOrder);
+        for (const key of keys) {
+          max = Math.max(max, this.cardOrder[key]);
+        }
+        this.cardOrder[this.dragCard] = max + 1;
+      // place it ahead of the card
```
kberg, line 67: Is this mean to be a comment of the block below? Can the comment go inside the else? Or is this a style thing?

kberg (approved):
Wow. This looks nice.

src/client/components/SortableCards.vue
bafolts: I don't think it is a styling thing of this repository. Must have been a styling thing in some repository I worked with at one point. Move this inside the else block.

PR #4758, opened 2022-08-05
Add a new debug view mode for showing and hiding tiles

kberg:
This mode is hidden behind a 'debug ui' preference.

image

### DIFF
--- a/src/client/components/BoardSpace.vue
+++ b/src/client/components/BoardSpace.vue
@@ -1,17 +1,20 @@
-<template>
+TileView<template>
   <div :class="getMainClass()" :data_space_id="space.id">
     <board-space-tile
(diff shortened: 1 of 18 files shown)


kberg:
You can review the two PRs separately, which might help a little.

src/client/components/BoardSpace.vue
```diff
@@ -1,17 +1,20 @@
-<template>
+TileView<template>
```
bafolts, line 1: What does this `TileView` do here?

bafolts (approved):

src/client/components/BoardSpace.vue
kberg: Typo. Thanks!

Issue #4763, opened 2022-08-07
Merger compatibility errors


kberg:
~~* Helion & Robinson Industries: Robinson Industries will trigger if there's heat but no money. But it will not use the heat.~~

nwai90:
Helion & Stormcraft: Helion is not able to use floaters as 2 heat each

kberg:
Unsubstantiated: Manutech + Odyssey, using Project Inspection to use Odyssey's Corpration

kberg:
unsubstantiated: 
> I played Curiosity with merger and didn't get those 2 steel productions from the corp

kberg:
Luna First Incorporated assigns steel and titanium like so:

    player.steel = 2;
    player.titanium = 2;

which means it overrides any steel and titanium gained before the Merger prelude came out.

kberg:
All the known Merger issues have been addressed, so I'm going to close this. We can open new ones as is necessary, of course.

PR #4766, opened 2022-08-08
Merger Compatibility: Helion + Robinson Industries.

kberg:
SelectHowToPayDeferred can be much simpler to use. So introducing that.

### DIFF
--- a/src/Player.ts
+++ b/src/Player.ts
@@ -1325,6 +1325,11 @@ export class Player {
       data: this.canUseData(selectedCard),
     };
   }
+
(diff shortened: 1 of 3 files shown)


kberg:
Part of #4763

nwai90 (approved):
LGTM, I wondered about putting the unit test in `RobinsonIndustries.spec.ts` instead of `Merger.spec.ts` but it's fine either way imo

kberg:
I'm thinking the same thing. I've got two more tests I'm about to send out,
let's think about this.

PR #4769, opened 2022-08-08
Fix Protected Valley having too many "on"s

norm:


kberg (commented):
This is great, but will break translated components:

image

I'd be grateful if you can catch those. If not I'll pick this up tomorrow night.

norm:
Like so?

kberg:
> Like so?

Perfect.

kberg (approved):

PR #4771, opened 2022-08-11
A series of changes related to top-level components.

kberg:
* The main app itself is now accessible from `window.app`
* BugReportDialog is now invoked through `window.app.showBugReportDialog`.
* removed a couple of 'as unknown' type conversions.


kberg:
(I'm not sure if this is a great idea yet or not. .I wanted to remove BugReportDialog from the preferences component, so that's nice. We shall see.)

kberg:
Oh wait, bug reporting is broken. So, wait on this.

PR #4777, opened 2022-08-13
Move server code to server/

kberg:
This is a huge change. Doing this by files would have been pretty messy.

Once this is done we can
1. separate client, common, and server tsconfig
2. client, common, and server lint rules
3. enforce some package boundaries.

It looks like everything got caught up as git moves, which is nice.


bafolts (approved):

kberg:
wooof

PR #4784, opened 2022-08-15
Lost the player tokens on the game-end Mars map.

kberg:


kberg:
Before: 
image

After: image

PR #4800, opened 2022-08-21
Rename Tags to Tag

kberg:
It's a genreal convention that enums are named singular.
I'm about to create a class that should be called `Tags` which will
conflict pretty hard with this `Tags`.


kberg:
Did a visual review. This is good.

PR #4806, opened 2022-08-23
Make Production a first-class object. Leverage Units and index by name.

kberg:


kberg:
I reviewed the code, and ~30% of the tests. The tests were all repetitive, and resolved by a series of replace operations.

PR #4811, opened 2022-08-23
Remove field no longer in use.

kberg:


kberg:
How about that -- the TODO was wrong.

PR #4818, opened 2022-08-25
Rewrite how intitial cards are selected.

### LINKED ISSUE #1400: Initial cards selection: please consider allowing all cards to be selected
Currently, the interface permits too few cards to be selected -- giving an error if you try to start -- but will not permit too many cards to be selected -- for example 3 preludes or 2 corporations.

However, for me, and I *think* for many other players, our process of deciding our initial cards is much more a process of progressively ruling OUT cards until the right number are left. That is, I would strongly prefer to begin by selecting ALL corps/preludes/projects, and then keep unclicking them until I'm down to the right number.

If I'm unusual in doing it this way, then of course there's no need to change anything here!

kberg:
This allows a player to make more choices, and tinker with the cards,
viz bug #1400.

Fixes #1400


kberg:
image
image
image

PR #4831, opened 2022-08-28
Make Timer.lastStoppedAt an instance variable rather than static

### LINKED ISSUE #2231: Fix timers
The `lastStoppedAt` should be moved from `Timer.lastStoppedAt` to `game.lastStoppedAt`. Currently this variable is one for all games which is unintentional.

kberg:
Fixes #2231


bafolts (approved):

kberg:
This didn't work. Rolling it back.

PR #4833, opened 2022-08-28
Show the colonies in the log panel.

kberg:
image

There's room for improvement: it doesn't use colony data from the game itself.

### DIFF
--- a/src/client/components/LogPanel.vue
+++ b/src/client/components/LogPanel.vue
@@ -20,13 +20,16 @@
           </div>
           <div class='debugid'>(debugid {{step}})</div>
         </div>
-        <div class="card-panel" v-if="cards.length > 0 || globalEventNames.length > 0">
+        <div class="card-panel" v-if="cardNames.length + globalEventNames.length + colonyNames.length > 0">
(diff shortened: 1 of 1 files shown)


src/client/components/LogPanel.vue
```diff
@@ -20,14 +20,17 @@
           </div>
           <div class='debugid'>(debugid {{step}})</div>
         </div>
-        <div class="card-panel" v-if="cards.length > 0 || globalEventNames.length > 0">
+        <div class="card-panel" v-if="cardNames.length + globalEventNames.length + colonyNames.length > 0">
```
kberg, line 23: Renamed to cardNames for consistency.

src/client/components/LogPanel.vue
```diff
@@ -80,10 +86,31 @@ const translatableMessageDataTypes = new Set([
   LogMessageDataType.TILE_TYPE,
   LogMessageDataType.GLOBAL_EVENT]);
 
+class ToggleSet<T> {
```
kberg, line 89: Rather than repeat everything below. I tried subclassing `Set` but Set isn't reactive in Vue.

PR #4840, opened 2022-08-29
Fix sqlite3 async purgeUnfinishedGames.

kberg:
purgeUnfinishedGames wasn't finished for SQLite, so this finishes it.

### DIFF
--- a/src/server/database/PostgreSQL.ts
+++ b/src/server/database/PostgreSQL.ts
@@ -161,10 +161,10 @@ export class PostgreSQL implements IDatabase {
       const dateToSeconds = daysAgoToSeconds(maxGameDays, 10);
       const selectResult = await this.client.query('SELECT DISTINCT game_id FROM games WHERE created_time < $1', [dateToSeconds]);
       const gameIds = selectResult.rows.map((row) => row.game_id);
-      const deleteResult = await this.client.query('DELETE FROM games WHERE game_id in ANY($1)', [gameIds]);
-      console.log(`Purged ${deleteResult.rowCount} rows from games`);
-      const idsResult = await this.client.query('DELETE FROM participants WHERE game_id in ANY($1)', [gameIds]);
-      console.log(`Purged ${idsResult.rowCount} rows from participants`);
+      const deleteGamesResult = await this.client.query('DELETE FROM games WHERE game_id in ANY($1)', [gameIds]);
+      console.log(`Purged ${deleteGamesResult.rowCount} rows from games`);
+      const deleteParticipantsResult = await this.client.query('DELETE FROM participants WHERE game_id in ANY($1)', [gameIds]);
+      console.log(`Purged ${deleteParticipantsResult.rowCount} rows from participants`);
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/server/database/SQLite.ts
```diff
@@ -256,7 +265,7 @@ export class SQLite implements IDatabase {
 
   private asyncGet(sql: string, params?: any): Promise<any> {
     return new Promise((resolve, reject) => {
-      this.db.get(sql, params, (err: Error | null, row: any) => {
+      this.db.get(sql, params, function(err: Error | null, row: any) {
```
kberg, line 268: There was some concern that I should use `function` instead of `()` but I don't think that changed anything meaningfully. I could keep this, or restore the original. Doesn't much matter to me.

src/server/database/PostgreSQL.ts
```diff
@@ -161,10 +161,10 @@ export class PostgreSQL implements IDatabase {
       const dateToSeconds = daysAgoToSeconds(maxGameDays, 10);
       const selectResult = await this.client.query('SELECT DISTINCT game_id FROM games WHERE created_time < $1', [dateToSeconds]);
       const gameIds = selectResult.rows.map((row) => row.game_id);
-      const deleteResult = await this.client.query('DELETE FROM games WHERE game_id in ANY($1)', [gameIds]);
-      console.log(`Purged ${deleteResult.rowCount} rows from games`);
-      const idsResult = await this.client.query('DELETE FROM participants WHERE game_id in ANY($1)', [gameIds]);
-      console.log(`Purged ${idsResult.rowCount} rows from participants`);
+      const deleteGamesResult = await this.client.query('DELETE FROM games WHERE game_id in ANY($1)', [gameIds]);
```
kberg, line 164: This is just renaming for clarity.

kberg, line 164: This is just renaming for clarity.

src/server/database/SQLite.ts
```diff
@@ -36,7 +36,7 @@ export class SQLite implements IDatabase {
         game_id varchar not null,
         last_save_id number not null,
         completed_time timestamp not null default (strftime('%s', 'now')),
-        PRIMARY KEY (game_id)`);
+        PRIMARY KEY (game_id))`);
```
kberg, line 39: This was actually broken. And since `purges` isn't used, and the error reporting was backwards, I didn't know.

src/server/database/SQLite.ts
bafolts: We should turn on linting rule to enforce this one way or the other to avoid user preference. I would always use `=>` as it keeps scope to the `thisArg` for us. If using `function` when one types `this` it may not be what they expect.

https://eslint.org/docs/latest/rules/prefer-arrow-callback

bafolts (approved):

src/server/database/SQLite.ts
kberg: Turns out that `function` is correct in this limited case. And apparently () => is not ideal in Mocha, though I don't recall why.

But I think keep this as-is for now.

PR #4847, opened 2022-08-29
Provide Card2, bespokeCanPlay, & bespokePlay.

kberg:
This is going to be cleaner and easier to manage than the current
approach. This removes most uses of `migrated`. The rest will come
shortly.

### DIFF
--- a/src/server/cards/Card.ts
+++ b/src/server/cards/Card.ts
@@ -13,6 +13,9 @@ import {CardRenderDynamicVictoryPoints} from './render/CardRenderDynamicVictoryP
 import {CardRenderItemType} from '../../common/cards/render/CardRenderItemType';
 import {IVictoryPoints} from '../../common/cards/IVictoryPoints';
 import {IProjectCard} from './IProjectCard';
+import {MoonExpansion} from '../moon/MoonExpansion';
+import {PlayerInput} from '../PlayerInput';
+import {isICorporationCard} from './corporation/ICorporationCard';
 
 export interface StaticCardProperties {
   adjacencyBonus?: AdjacencyBonus;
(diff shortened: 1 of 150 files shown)


src/server/cards/Card.ts
```diff
@@ -230,3 +241,53 @@ export abstract class Card {
...
+    if (this.requirements?.satisfies(player) === false) {
+      return false;
+    }
+    if (this.productionBox && !player.production.canAdjust(this.productionBox)) {
+      return false;
+    }
+    return this.bespokeCanPlay(player);
+  }
+
+  public bespokeCanPlay(_player: Player): boolean {
```
bafolts, line 275: Does this need to be public access level? If intentions are for cards to override this then protected may make more sense.

bafolts (approved):

src/server/cards/Card.ts
kberg: That's an interesting though. There's a whole pipeline of these changes. Let me see how it looks far down the line.

BTW, I'm now prototyping automating as much behavior as possible, making `bespoke` a real thing.

e.g.

```
      behavior: {
        production: {heat: 3, plants: 1},
        addResourceToAnyCards: {type: CardResource.MICROBE, count: 2},
      },
```

and

```
      actions: [
        {
          drawCard: 2,
        },
      ],
```

```
      actions: [
        {
          production: {energy: -1},
          tr: 1,
        },
      ],
```

These should be able to get more sophisticated, so I'm really prototyping carefully.

The point here is that so much of the code should be replaced with code that can better determine whether you can play something. (This is really about getting edge cases right. But also it'll remove **so much code**.)

PR #4850, opened 2022-08-30
Add "Download game log". Also clean up some translations.

### LINKED ISSUE #1110: Feature Request: Game log download

kberg:
Fixes #1110

Download is available at game end.

image

The result isn't pretty, but we can work on that.

```
Drew and discarded House Printing (cost 10) to place a 2
Drew and discarded Rover Construction (cost 8) to place a 0
Drew and discarded Indentured Workers (cost 0) to place a 2
Drew and discarded Fueled Generators (cost 1) to place a 0
The id of this game is gf82fe669cf81
Good luck red!
Generation 1
red drew 1 card(s)
You drew Space Hotels
red played Point Luna
red kept 7 project cards
red played Metals Company
red played Research Network
red drew 3 card(s)
You drew Earth Office, Ore Processor and Sabotage
red played Nuclear Zone
red drew 1 card(s)
You drew Windmills
red placed Nuclear Zone tile on row 4 position 5
red's plants amount increased by 2
red played Earth Office
red drew 1 card(s)
You drew Artificial Lake
red played Space Hotels
red drew 1 card(s)
You drew Fuel Factory
red passed
Generation 2
red bought 3 card(s)
You drew Cartel, Invention Contest and Peroxide Power
red played Peroxide Power
red played Invention Contest
red drew 1 card(s)
You drew Power Plant
red played Power Plant
red played Sabotage
red passed
Generation 3
red bought 1 card(s)
You drew Sponsors
red played Sponsors
red drew 1 card(s)
You drew Mass Converter
red played Cartel
```


src/server/routes/GameLogs.ts
```diff
@@ -25,34 +25,22 @@ export class GameLogs {
...
+      return game.gameLog.filter(messagesForPlayer).slice(-50);
     } else { // pull all logs for generation
-      logs = this.getLogsForGeneration(game.gameLog, Number(generation)).filter(messagesForPlayer);
+      return this.getLogsForGeneration(game.gameLog, Number(generation)).filter(messagesForPlayer);
+    }
+  }
+
+  public getLogsForGameEnd(game: Game): Array<string> {
+    if (game.phase !== Phase.END) {
+      throw new Error('Game is not over');
```
bafolts, line 41: Will this need to be caught so that we return a bad request response? Otherwise this may end up as an unhandled exception.

kberg: It's only going to happen when someone tries to assemble a URL and access the long before the game is over, so it's not something I was concerned with. Do you have any other thoughts?

bafolts: Uncaught exceptions could stop the node process. We happen to prevent this in node where we listen for `uncaughtException` [here](https://github.com/terraforming-mars/terraforming-mars/blob/main/src/server/server.ts#L20). I think we should only use this for places we accidentally end up with an unhandled exception. I think we should catch this error and return a ctx bad request. Otherwise I am not sure res.end would ever be called.

kberg: Aye. Fair point. I'll have it done in about ten minutes.

kberg: Update: we handle uncaught exceptions on the request path:

image

PR #4867, opened 2022-09-02
Add board pins to fan boards

kberg:


kberg:
For some reason this is a bit of a mess.

PR #4876, opened 2022-09-03
Improve translatable string in GameEnd view

nicobubulle:
I have added a few strings being able to be translated in the Game End view and improved render of some translation.

New Strings :
    "Milestones points": "Milestones points",
    "Awards points": "Awards points",
    "won!": "won!",
    "Download game log": "Download game log",
    "Gen:": "Gen:",

### DIFF
--- a/src/client/components/GameEnd.vue
+++ b/src/client/components/GameEnd.vue
@@ -1,6 +1,6 @@
 <template>
   <div id="game-end" class="game_end_cont">
-      <h1>{{ constants.APP_NAME }} - Game finished!</h1>
+      <h1  v-i18n>{{ constants.APP_NAME }} - Game finished!</h1>
       <div class="game_end">
           <div v-if="isSoloGame()">
               <div v-if="game.isSoloModeWin">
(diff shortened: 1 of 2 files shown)


src/client/components/GameEnd.vue
```diff
@@ -42,17 +42,17 @@
...
-              <h2 v-i18n>Victory points breakdown after<span> {{game.generation}} </span>generations</h2>
+              <h2><span v-i18n>Victory points breakdown after</span> {{game.generation}} <span v-i18n>generations</span></h2>
               <table class="table game_end_table">
                   <thead>
                       <tr v-i18n>
                           <th><div class="card-delegate"></div></th>
                           <th><div class="tr"></div></th>
-                          <th><div class="m-and-a" title="Milestones points">M</div></th>
-                          <th><div class="m-and-a" title="Awards points">A</div></th>
+                          <th><div class="m-and-a tooltip tooltip-top" :data-tooltip="$t('Milestones points')">M</div></th>
```
kberg, line 54: Curious -- why did you have to add the additional class selectors?

kberg (approved):

src/client/components/GameEnd.vue
nicobubulle: When I have translated them, it didn't work. If I use the tooltip class, the translation is working. I'm not sure to fully understand how the translation system works and why some string can't be translated.

kberg: Got it.

Did you intentionally not supply the additional translations?

nicobubulle: Yes I'm updating French Translation and there is a lot of work ! I will pull it when it's finished (80% yet). For information, I'm using a selfhosted instalation of weblate to do that. I will maybe later share it to help to improve all translations. But to do this we should add an english translation in locales as reference for all the translation.

PR #4891, opened 2022-09-05
Make Helion compatible with Stormcraft Incorporated for Merger

kberg:
Right now I'm asking for feedback. Three things remain:

~~1. Unit tests.~~
~~2. More manual testing~~
3. Making all of this work with Dirigibles. (There is no end. ;)) (This can be resolved on its own.)

Part of #4763


nwai90 (approved):
Tested, LGTM!

kberg:
woooo

PR #4904, opened 2022-09-07
Add increasing terraform rating to Behavior.

kberg:


src/server/behavior/Behaviors.ts
```diff
@@ -37,5 +38,19 @@ export class Behaviors {
         player.drawCard(options.count, {tag: options.tag, resource: options.resource, cardType: options.type});
       }
     }
+
+    if (behavior.global !== undefined) {
+      const global = behavior.global;
```
bafolts, line 43: I would be cautious with naming a variable or property _global_. In some javascript environments _global_ will be an object that this shadows.

https://developer.mozilla.org/en-US/docs/Glossary/Global_object

kberg: Oh that's fair. Do you think I should rename the `behavior` attribute?

bafolts (approved):

PR #4908, opened 2022-09-08
Rewrote creating games for tests yet again.

kberg:
I have mixed feelings about this approach. It introduces a fake form of extension that doesn't really accomplish what I set out to do, which is to provide an equivalent to TestPlayer. The difference is that TestPlayers' parent, Player, has an accessible constructor, but Game does not.

So, just not clear.

### DIFF
--- a/tests/TestGame.ts
+++ b/tests/TestGame.ts
@@ -3,7 +3,9 @@ import {GameOptions} from '../src/server/GameOptions';
 import {setCustomGameOptions} from './TestingUtils';
 import {TestPlayer} from './TestPlayer';
 
-export function newTestGame(count: number, customOptions?: Partial<GameOptions>, idSuffix = ''): Game {
+export type TestGame = Game & {testPlayers: Array<TestPlayer>};
+
+export function newTestGame(count: number, customOptions?: Partial<GameOptions>, idSuffix = ''): TestGame {
   const players = [
     TestPlayer.BLUE.newPlayer(false, idSuffix),
     TestPlayer.RED.newPlayer(false, idSuffix),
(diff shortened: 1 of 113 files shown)


kberg:
@bafolts I'd really like your thoughts about this approach.

tests/TestGame.ts
```diff
@@ -18,13 +20,9 @@ export function newTestGame(count: number, customOptions?: Partial<GameOptions>,
   const options: GameOptions | undefined = customOptions === undefined ?
     undefined :
     setCustomGameOptions(customOptions);
-  return Game.newInstance(`game-id${idSuffix}`, players, players[0], options);
-}
-
-export function getTestPlayer(game: Game, idx: number): TestPlayer {
-  return game.getPlayers()[idx] as TestPlayer;
-}
+  const game = Game.newInstance(`game-id${idSuffix}`, players, players[0], options);
```
bafolts, line 23:
```suggestion
  const game: TestGame = Game.newInstance(`game-id${idSuffix}`, players, players[0], options);
  game.testPlayers = players;
  return game;
```

Save an _any_ and be proud!

bafolts (approved):
This approach seems fine. With my suggestion the _any_ can be avoided.

tests/TestGame.ts
kberg: Unfortunately this doesn't work.
image

bafolts: When I make the change locally the build works for me. Is this error seen while building? I suggested the change after being able to build locally.

kberg:
I'm gonna close this one for now.

Issue #4917, opened 2022-09-10
Copernicus Tower shouldn't increase megacredits production


elynnlee:
Copernicus Tower's actions read: "Add 1 Science resource here, or spend 1 Science resource here to raise your TR 1 step." However, when activating the card, my megacredit production also increases by 1.

Here's an image from the log, happy to send game link as well if helpful: 
<img width="409" alt="image" src="https://user-images.githubusercontent.com/2634759/189503337-f26a2a58-5e90-488d-a091-be3e251192d4.png">

Repro steps:
- Play Copernicus Tower
- Add a science resource to it
- Take the action to remove 1 resource and raise TR by 1 step
- Observe that megacredits production also increase

kberg:
While you're still playing, hit the gear icon, select "Report a Bug". Can
you copy that information into this bug, please?

kberg:
Never mind. I see it. Huh! Why is it doing that?

kberg:
This was fixed two days ago. Thank you.

PR #4919, opened 2022-09-11
Copernicus Tower has a production benefit it shouldn't.

kberg:


kberg:
Fixes #$917

kberg:
Fixes #4917

Issue #4746, opened 2022-07-31
Merger: initially discarded corporations were shuffled back in?


kevinb9n:
Sorry, I know this is extremely minor.

I think in an over-the-table game players would not shuffle initially-discarded corporations back in to the corp deck before playing the Merger prelude, unless/until the corporation deck was actually empty. So you wouldn't have to redraw a corp you had already discarded once.

I could have missed a ruling about this though.

Again, small, sorry.

motoyugota:
I would agree that this is how it should work. Same goes for Valley Trust and New Partner with Preludes.

kberg:
This is fixed.

kberg:
Agreed. Thanks.

Issue #4741, opened 2022-07-30
Corporation draft with limited corporations might not work


kberg:
From Discord:
> FoxRageQc: HEllo, when i try to create a game today, it says this: Internal server error: No more corporation cards for game g1d08898d0861
> k9: @FoxRageQc did you try corporations draft with a limimted set of corporations?
> FoxRageQc: Hmmm yes i think so

pauliimmonen:
I did a little investigation of the issue. The error message derives from corporations draft mode. The error happens because as we initialize each player the corporations are dealt to them. So as we start drafting "minimum corporations required" amount is already popped out of the deck. As a result you need double the minimum to get the corporations draft going.

kberg:
Thanks, @pauliimmonen. I think #4928 will fix this as it gives you a deck of corporations no matter how many you start with.

kberg:
This is fixed.

PR #4928, opened 2022-09-13
Remove Dealer, replace with Deck. Things happen.

### LINKED ISSUE #4920: New Partner selects duplicate preludes
Game ID: ge6fd824eb788

Screenshot_20220909-233031_Chrome

In the screenshot, you can see that i already played Aquifier Turbines but it has come up again as an option with New Partner.

kberg:
**Don't review this until #4924  is in and done.**

It replaces `Dealer` with 3 instances of `Deck`. One for project cards, one for corporation cards, and one for preludes. The code now actively discards unused corps and preludes, and might even shuffle them back in.

It also treats customPrelude and customCorps differently, not by eliminating all the other cards from the game, but by creating a standard deck of cards that the game would contain, then taking those custom cards and adding them to the top of the deck so they're drawn first. This way if customPrelude or customCorps aren't enough (because of a bug, or because of New Partner or Merger) the code will have a read-made deck to draw from.

This should fix #4920, #4746 and #4741.

As I'm typing this I can see other tests worth adding:

1. Verifying that discarded corps and preludes get tossed in their discard piles.
2. Verifying that after deserialization, `Dealer.draw` and `Deck.draw` grab the same cards.


src/server/Game.ts
```diff
@@ -305,26 +287,17 @@ export class Game implements Logger {
         gameOptions.turmoilExtension ||
         gameOptions.initialDraftVariant) {
         for (let i = 0; i < gameOptions.startingCorporations; i++) {
-          const corpCard = corporationCards.pop();
-          if (corpCard !== undefined) {
-            player.dealtCorporationCards.push(corpCard);
-          } else {
-            throw new Error('No corporation card dealt for player');
-          }
+          player.dealtCorporationCards.push(corporationDeck.draw(game));
```
pauliimmonen, line 290: Here you should not draw cards if you have corporationsDraft game option

kberg: > Here you should not draw cards if you have corporationsDraft game option

Yes, you're definitely right. Since this change is already pretty heavy, what I'd like to do is fix that in a PR all its own after this goes in. It should be easy to write a test for this one, so I'm happy to also do that later, too. Seem reasonable?

src/server/GameCards.ts
```diff
@@ -92,18 +94,33 @@ export class GameCards {
...
-    if (preludes.length > 0) {
-      return preludes;
+    if (preludes.length === 0) {
+      preludes = this.instantiate(PRELUDE_CARD_MANIFEST.preludeCards);
     }
-    return this.instantiate(PRELUDE_CARD_MANIFEST.preludeCards);
+    return this.addCustomCards(preludes, this.gameOptions.customPreludes);
+  }
+
+  private addCustomCards<T extends ICard>(cards: Array<T>, customList: Array<CardName> | undefined): Array<T> {
```
bafolts, line 112: This avoids having to check if `customList` is defined and creating `list` var. Typescript can generate this code.

```suggestion
  private addCustomCards<T extends ICard>(cards: Array<T>, customList: Array<CardName> = []): Array<T> {
```

src/server/GameCards.ts
```diff
@@ -92,18 +94,33 @@ export class GameCards {
...
+    const list = (customList || []);
+    for (const cardName of list) {
+      const idx = cards.findIndex((c) => c.name === cardName);
+      if (idx === -1) {
+        const card = this.cardFinder.getCardByName(cardName);
+        if (card !== undefined) {
+          cards.push(<T> card);
+        }
```
bafolts, line 120: Is this worth logging a warning if `card` is not defined? In this situation the server may not be aware of the custom card user is attempting to add.

src/server/utils/shuffle.ts
```diff
@@ -0,0 +1,20 @@
+// in-place shuffle using the Fisher-Yates algorithm..
+// See https://bost.ocks.org/mike/shuffle/ for details.
+
+import {Random} from '../Random';
+
+export function inplaceShuffle(array: Array<any>, rng: Random) {
```
bafolts, line 6:
```suggestion
export function inplaceShuffle(array: Array<unknown>, rng: Random) {
```

src/server/utils/shuffle.ts
```diff
@@ -0,0 +1,20 @@
...
+  let last = array.length;
+
+  // While there remain elements to shuffle…
+  while (last > 0) {
+    // Pick a remaining element…
+    const idx = rng.nextInt(last);
+    last--;
+
+    // And swap it with the current element.
+    const temp = array[last];
```
bafolts, line 16:
```suggestion
   
    [array[last], array[idx]] = [array[idx], array[last]];
```

Not important but in ES6 this can be done with [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment).

bafolts (approved):

src/server/GameCards.ts
kberg: It might be worth throwing if that's the case. It's not clear if that error makes it all the way back to the caller, though. So, for now, logging a warning makes sense. Thanks.

src/server/utils/shuffle.ts
kberg: Nice!

src/server/GameCards.ts
kberg: This is nice.

src/server/Game.ts
kberg: This is #4965

PR #4935, opened 2022-09-14
Clarify community card Political Uprising.

kberg:
"Place 4 delegates" does not really describe how this card works, since each delegate may go into its own party. The existing cards tend to say "in 1 party" and so it's possible to assume the absence of "in 1 party" is enough, but I recommend this is explicit.


kberg:
@nwai90 does this make sense to you?

nwai90:
> @nwai90 does this make sense to you?

Looks fine to me, or alternatively "Place 4 delegates into any party / parties" could work too

nwai90 (approved):

PR #4936, opened 2022-09-14
Hi-tech lab unnecessary choose when using 1 energy

pauliimmonen:
When using Hi-tech lab with one energy the game asks you to choose a card from one option which is unnecessary. Replaced draw one keep one simply by drawing a card.


kberg:
This seems reasonable. How did you ensure it works?

kberg:
Thanks!

kberg (approved):

Issue #4941, opened 2022-09-15
User report: Think Helion might have a bug double spending heat when playing local heat trapping card at a discount


kberg:
Reference: https://discord.com/channels/737945098695999559/742721510376210583/1019088448759545948

kberg:
Fixed by #4943

PR #4943, opened 2022-09-15
Local Heat Trapping is double-deducting.

kberg:
Moving reserveUnits caused a little trouble in this one place where resources weren't meant to be deducted.


kberg:
Fixes #4941

PR #4945, opened 2022-09-16
Replace resourceType from with isSelfReplicatingRobotsCard.

kberg:
resourceType is no longer necessary.


kberg:
SRR cards still show correctly (the 4M€ discount is due to an additional general -2 cost for all cards.)

image

Unplayed cards look correct

image

Card with resources on them look correct (you *wish* you had two resources on Physics Complex)

image

SRR card looks correct while offering to double it:

image

PR #4962, opened 2022-09-19
updated viz for 4 global events.

kberg:


kberg:
image
image
image
image

PR #4963, opened 2022-09-19
Add initial Countability to Behavior.

kberg:
This initial change will only impact three cards that have a benefit based on tags in play. Most cards that count tags for behavior increase production and that's a little more complicated. Better to show this simple case first, applied to just a couple of cards.


src/server/awards/amazonisPlanitia/Engineer.ts
```diff
@@ -10,15 +9,20 @@ export class Engineer implements IAward {
...
     // TODO(kberg): should Engineer include events?
     const score = player.tableau.filter((card) => {
-      if (card.produce !== undefined) return true;
-      if (card.behavior?.production !== undefined && Units.isEmpty(card.behavior.production)) return true;
       if (Engineer.productionCards.includes(card.name)) return true;
+
+      if (card.produce !== undefined) return true;
+      const production = card.behavior?.production;
+      if (production !== undefined) {
+        return Object.keys(production).length > 0;
```
bafolts, line 17: I would be cautious with using `Object.keys` for this check. We don't have [exactOptionalProperties](https://www.typescriptlang.org/play?ts=4.4.2&exactOptionalPropertyTypes=true#example/exact-optional-properties) set. Until we have this set I think this also needs to check that if there is a key on production that it's value is _undefined_. As long as _production_ has one key that has a value that isn't _undefined_ I think that check would work for time being. This check is losing type safety.

src/server/behavior/Counter.ts
```diff
@@ -0,0 +1,91 @@
...
+   *
+   * The other part of why includeCard is interesting is sometimes this counting is related to a card
+   * being replayed via Robotic Workforce. In that case, the card is already played.
+   *
+   * Counter is not meant to be kept between game states, so dispose of this after making calcuations.
+   */
+  private cardIsUnplayed: boolean;
+
+  public constructor(private player: Player, private card: ICard) {
+    this.cardIsUnplayed = card !== undefined && !player.cardIsInEffect(card.name);
```
bafolts, line 28: The _card_ argument's type doesn't indicate it can be _undefined_. Is the check for _undefined_ necessary?

bafolts (approved):

Issue #4970, opened 2022-09-21
Corroder Suits allows adding resource to non-venus tag card


andystager:
First time I've noticed this behavior, see my game log-

![image](https://user-images.githubusercontent.com/72101041/191538142-f57e9482-2be7-4ba2-8660-edd290be6c3e.png)

kberg:
Whoops! A fix is on its way.

PR #4972, opened 2022-09-22
Whoops! Actually make tag filtering work when adding resources to cards

kberg:


kberg:
Fixed #4970

kberg:
Fixed #4970

PR #4977, opened 2022-09-22
Language constants should be in native language.

kberg:
image

But then I saw that the languages themselves were being translated.

image

Hrm.


kberg:
These language translations shouldn't occur. The purpose is not to inform players what languages are available in their tongue, but to make it easy for a player to find their language without having to read English.

kberg:
(Fixed the Chinese string thanks to @nwai90

PR #4978, opened 2022-09-22
Translation Improvement

nicobubulle:


src/client/components/GameEnd.vue
```diff
@@ -22,9 +22,9 @@
               <div v-else>
                   <div class="game_end_fail">
                       <h2 v-i18n>Sorry, you lose.</h2>
-                      <div class="game_end_notice" v-i18n>
-                          Next time you will get more luck!<br>
-                          Also, take into account these small hints to win:
+                      <div class="game_end_notice">
+                        <span v-i18n>Next time you will get more luck!</span><br>
```
kberg, line 26: I have been thinking about rewriting these to be more clear, and which point your translations might be out of date. Would it be a problem if you translated twice?

kberg (commented):
Looks great. I just have that one question.

src/client/components/GameEnd.vue
nicobubulle: No worry about that, I still have a lot of work with French translation. I will translated it again.

kberg (approved):

src/client/components/GameEnd.vue
kberg: Great. Thanks for this!

PR #4979, opened 2022-09-23
Make Earth Embassy compatible with Martian Zoo

kberg:


kberg:
Fixes a bug reported on Discord.

PR #4983, opened 2022-09-23
Add method for counting cities not on Mars

kberg:


kberg:
Tests do support the different counting strategies. These count operations ought to move to Board, at which point perhaps add direct tests.

PR #4994, opened 2022-09-25
Add Moon corporation Luna Trade Federation.

kberg:
image


bafolts:
132kb for the logo is probably unnecessary. I would upload a smaller logo.

src/client/mixins/PaymentWidgetMixin.ts
```diff
@@ -61,12 +61,22 @@ export const PaymentWidgetMixin = {
     },
     getMegaCreditsMax(): number {
       const model = this.asModel();
-      return Math.min(model.playerView.thisPlayer.megaCredits, model.$data.cost);
+      return Math.min(model.playerView.thisPlayer.megaCredits, model.cost);
+    },
+    canUseTitanium(): boolean {
+      throw new Error('Should be overridden');
```
bafolts, line 67: This pattern feels like runtime accident waiting to happen. Is there a way to enforce this at compile time?

kberg:
> 132kb for the logo is probably unnecessary. I would upload a smaller logo.

Good catch. I thought I had already fixed that.

src/client/mixins/PaymentWidgetMixin.ts
kberg: Tests will fail.

kberg: Since the only thing keeping this from being submitted is a question about a problem revealed either at compile time or runtime, and given the review latency, I'm going to submit the PR.

PR #5002, opened 2022-09-26
Removing a lot of unnecessary capitalizations.

kberg:


kberg:
I think I'm going to rewrite this in smaller pieces in order to be more confident about the change.

PR #5007, opened 2022-09-28
Add Community corporation United Nations Mission One

kberg:
This is a near-direct copy of the Community server's version of this corp.

image

### DIFF
--- a/src/server/cards/community/UnitedNationsMissionOne.ts
+++ b/src/server/cards/community/UnitedNationsMissionOne.ts
@@ -1,0 +1,50 @@
+import {ICorporationCard} from '../corporation/ICorporationCard';
+import {Tag} from '../../../common/cards/Tag';
+import {Card} from '../Card';
+import {CardName} from '../../../common/cards/CardName';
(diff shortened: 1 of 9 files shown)


src/server/cards/community/UnitedNationsMissionOne.ts
```diff
@@ -0,0 +1,50 @@
...
+      },
+    });
+  }
+
+  public override bespokePlay(player: Player): undefined {
+    player.increaseTerraformRating();
+    return undefined;
+  }
+
+  // TODO(kberg): Since UNMO can generate MC for raising TR, that MC can offset reds costs, can't it?
```
nwai90, line 42: Good question! There is no ruling on it from the designer yet. We could go either way on that

nwai90 (approved):

src/server/cards/community/UnitedNationsMissionOne.ts
kberg: No matter what, it makes reds costs even harder to predict.

PR #5017, opened 2022-09-30
More big decaps.

kberg:


kberg:
This PR stepped on Italian translations (Animale became animale.) Whether that's allowable, it's still inconsistent.

kberg:
... so I'll patch this PR shortly.

PR #5025, opened 2022-10-03
Solar Probe card draw was showing what you drew to other players.

kberg:
Also add sanity test for Solar Probe.


kberg:
I tested this manually, because I want to get the fix out. I'll put together an automated test tomorrow.

PR #5047, opened 2022-10-12
Adjust community Venus colony.

kberg:
First change rendering so it shows that it requires venus cards with resources. Second, change the behavior to match. Finally, add a test!


kberg:
image

PR #5093, opened 2022-10-22
Remove problematic reduced brigness in inputs

pauliimmonen:
Remove reduced brightness from unselected cards. Selected cards slightly less distinguishable. I think it is either this or full roll back to old visualization.

no_brightness_reduced

Fixes the problems escpecillay with in initial card selection. 

init


kberg (approved):
Thank you

PR #5096, opened 2022-10-24
Update ui.json

Borbarad2:
typo
adding translation for all missing mielstones and awards

There is an inconsistency: After all awards there is not a full stop behind the description except for the moon awards. I took the full stop with the moon to not block that part of the translation. The full stop must be removed, but then it need to remove in translation files as well.


kberg:
This has an error:

image

Borbarad2:
Yes, right: the semicolon";" should be a colon","
How can i correct that?
Or could you do it quickly for me?

Borbarad2:
Yes, right, the semicolon";" should be a colon","
I have corrected it.

PR #5117, opened 2022-11-06
Replace array of arrays with JSON responses.

kberg:
Having this type of structure will allow some features currently not possible, including PlayerInputs nested more than one level deep, adding a top-level interrupt-type response (think 'undo') and also adding a unique request/response ID, which just adds a little bit of safety for knowing that the response is genuinely for the PlayerInput that player is waiting for.

This will improve comprehension at the cost of network bandwidth, but I believe it is an acceptable tradeoff.

It's possible these validations are a little strict (so what if additional fields are part of the input?) But we can loosen those.

Still pending are some new tests, and also some serious manual testing.


kberg:
Still a few tests broken, but I wanted this to be in the queue.

bafolts:
As this is rolling out we may need to support the array of array and the JSON response on the server. At some date can then remove support for array of array. This will allow rolling out without requiring users to have to restart their browser.  As this is a hobby project perhaps any players playing while server is updating isn't the worst thing.

The original array of array was not put in place to save bandwidth. Was writing this quickly for fun and used array of array for simplicity!

kberg:
> As this is rolling out we may need to support the array of array and the JSON response on the server. At some date can then remove support for array of array. This will allow rolling out without requiring users to have to restart their browser. As this is a hobby project perhaps any players playing while server is updating isn't the worst thing.
> 
> The original array of array was not put in place to save bandwidth. Was writing this quickly for fun and used array of array for simplicity!

Yeah, I think it isn't the worst thing. I tried ways of implementing this in pieces, and it never came together.

Issue #4276, opened 2022-04-03
"Prototype" font does not support Polish characters


Argeento:
![image](https://user-images.githubusercontent.com/13007891/161425221-fe709fb8-ee6f-498c-ab4e-caa6dbc23aab.png)

kberg:
Is the written text wrong, or is it just a matter of how they are drawn? It looks like the latter but it's good to confirm.

kberg:
Roboto 700 Bold might do the job.

![image](https://user-images.githubusercontent.com/413481/162590442-26640266-53a7-4fc2-90b9-25d74203e237.png)

It's one of the fonts on https://www.whatfontis.com/Prototype.similar

kberg:
Play is nice too ![image](https://user-images.githubusercontent.com/413481/162590509-ff72364f-ef3e-4835-897e-0607452aaab3.png)

kberg:
![image](https://user-images.githubusercontent.com/413481/162590568-def13172-4e23-42a8-9e08-6b3aa43bdc20.png)

kberg:
These are all free Google fonts.

Argeento:
The glyphs are missing. 

Mostly we need cover for ["European Latin"](https://maximilian.schalch.de/2018/05/complete-list-of-european-special-characters/) (every font: ✔️)

`Exo` is missing [Slavic Cyrillic](https://en.wikipedia.org/wiki/Cyrillic_script)

Imo `Play` looks good 🚀

kberg:
@Argeento if this is something you'd like to implement, feel free. I could get to this in the coming weeks.

Could be an automatic selection based on having selected Polish. Or could be part of the experimental UI (still while lang === polish).

Pyzdra:
[Prototype.zip](https://github.com/terraforming-mars/terraforming-mars/files/9950127/Prototype.zip)
I found Prototype font with polish characters on forum: https://www.gry-planszowe.pl/forum/viewtopic.php?f=55&t=74422&start=200.
Could you replace the font with the one provided? (It's zip cause github does not accept .ttf as attachement).

rafalszampera:
Based on the font licensing: "Copyright 2001, free for personal use, freely distributed with Prototype.txt" So the license is the same as the original Prototype font.
@Pyzdra Your font is resolving problems with a lack of all European Latin chars, I create PR with this font: https://github.com/terraforming-mars/terraforming-mars/pull/5118

kberg:
[...]

PR #5118, opened 2022-11-07
#4276 "Prototype" font does not support Polish characters

rafalszampera:
Adding the Pr for https://github.com/terraforming-mars/terraforming-mars/issues/4276
The source of the font: https://drive.google.com/file/d/15KlRdhCYwS2zLSmdhgFah1_3RQvRCJis/view
Based on the font licensing: "Copyright 2001, free for personal use, freely distributed with Prototype.txt" So the license is the same as the original Prototype font.
@Pyzdra Your font is resolving problems with a lack of all European Latin chars, I create PR with this font


kberg:
Thank you! A couple of observations: first, doesn't that license say it has to be distributed with the text file? And second, it seems generally safer to allow the client to choose this alternate font, in case it's not a straight replacement. What do you think about that second one?

rafalszampera:
> Thank you! A couple of observations: first, doesn't that license say it has to be distributed with the text file? 
Based on the prototype licensing I found the font before my changes have a similar licensing or a more restrictive one: "Free for personal use  only"
Here license with txt file: https://www.1001fonts.com/prototype-font.html
Here for non commercial use: https://www.dafont.com/prototype.font

> And second, it seems generally safer to allow the client to choose this alternate font, in case it's not a straight replacement. What do you think about that second one?
I think that the decision about the font used should be taken in the code, different fonts take different spaces and look not so futuristic. 
So if the usage of prototype font is a grey area, I would prefer to use google fonts, as you mentioned in the original ticket.

kberg:
> > Thank you! A couple of observations: first, doesn't that license say it has to be distributed with the text file?
> 
> Based on the prototype licensing I found the font before my changes have a similar licensing or a more restrictive one: "Free for personal use only" Here license with txt file: https://www.1001fonts.com/prototype-font.html Here for non commercial use: https://www.dafont.com/prototype.font

Please take a minute to read the file Prototype.txt inside the .zip file. It has, among other things, this term:
> Please include this Prototype.txt file with any redistribution.

Which this PR currently does not.

> > And second, it seems generally safer to allow the client to choose this alternate font, in case it's not a straight replacement. What do you think about that second one?
> 
> I think that the decision about the font used should be taken in the code, different fonts take different spaces and look not so futuristic. So if the usage of prototype font is a grey area, I would prefer to use google fonts, as you mentioned in the original ticket.

It's a gray area only because I don't know whether this font is usable for all the other current applications. Does it work with Korean, for example.

We could let the code choose the font, and if we do, then we could at least limit this font's usage to Polish for the time being. That's where I think this should go.

rafalszampera:
Ok, points taken. So what I'll do is:
I will upload txt file, and close this PR. - thanks to that if this PR will be a point in a future discussion it will be somehow "complete enough".
After that, I'll move your comment into the ticket and I'll try to prepare the change for that - let's check if I know the application enough to do so. :)

PR #5122, opened 2022-11-11
Add Merger Variant (twoCorpsVariant)

d-little:
## What does this PR do?
Implements the Merger Variant, which Everyone gets the Merger Prelude.  It's given post-drafting (if enabled).
- A few visual changes to New Game menu were made to make it a little 'easier'.
- Based off the wonderful work in [tfm-community](https://github.com/nwai90/terraforming-mars)

## Notes for code reviewer

- You can very easily run out of corporations if minimal expansions are added.  
  - A check should be done prior to game start to ensure more than enough corporations have been selected.
  - I'm unsure on how to approach doing this just yet

(My apologies for the number of superfluous commits as I try to figure out how this language likes to concat strings.. recommending squashing if PR is accepted.)

---

### DIFF
--- a/src/client/components/create/CreateGameForm.vue
+++ b/src/client/components/create/CreateGameForm.vue
@@ -997,6 +997,25 @@ export default (Vue as WithRefs<Refs>).extend({
         }
       }
 
+      // Check custom corp. count
+      if (customCorporations.length > 0) {
+        if (!this.twoCorpsVariant)
+          let neededCorpsCount = players.length * startingCorporations;
+          if (this.prelude && this.promoCardsOption) neededCorpsCount += 4;
+        } else {
+          // Add an additional 4 for the Merger prelude
+          // Everyone-Merger needs an additional 4 corps per player
+          //  NB: This will not cover the case when no custom corp list is set!
+          //  It _can_ come about if  the number of corps included in all expansions is still not enough.
+          let neededCorpsCount = (players.length * startingCorporations) + (players.length * 4);
+        }
+
+        if (customCorporations.length < neededCorpsCount) {
+          window.alert(translateTextWithParams('Must select more corporations (${0}/${0})', [customCorporations.length, neededCorpsCount.toString()]));
+          return;
+        }
+      }
+
       if (players.length === 1 && corporateEra === false) {
         const confirm = window.confirm(translateText(
           'We do not recommend playing a solo game without the Corporate Era. Press OK if you want to play without it.'));
--- a/src/client/components/create/CreateGameForm.vue
+++ b/src/client/components/create/CreateGameForm.vue
@@ -999,15 +999,16 @@ export default (Vue as WithRefs<Refs>).extend({
 
       // Check custom corp. count
       if (customCorporations.length > 0) {
+        let neededCorpsCount = 0
         if (!this.twoCorpsVariant) {
-          let neededCorpsCount = players.length * startingCorporations;
+          neededCorpsCount = players.length * startingCorporations;
           if (this.prelude && this.promoCardsOption) neededCorpsCount += 4;
         } else {
           // Add an additional 4 for the Merger prelude
           // Everyone-Merger needs an additional 4 corps per player
           //  NB: This will not cover the case when no custom corp list is set!
           //  It _can_ come about if  the number of corps included in all expansions is still not enough.
-          let neededCorpsCount = (players.length * startingCorporations) + (players.length * 4);
+          neededCorpsCount = (players.length * startingCorporations) + (players.length * 4);
         }
 
         if (customCorporations.length < neededCorpsCount) {
--- a/src/client/components/create/CreateGameForm.vue
+++ b/src/client/components/create/CreateGameForm.vue
@@ -258,8 +258,24 @@
                         </div>
 
                         <div class="create-game-page-column" v-if="playersCount > 1">
+
                             <h4 v-i18n>Multiplayer Options</h4>
 
+                            <input type="checkbox" v-model="randomFirstPlayer" id="randomFirstPlayer-checkbox">
+                            <label for="randomFirstPlayer-checkbox">
+                                <span v-i18n>Random first player</span>
+                            </label>
+
+                            <input type="checkbox" name="showOtherPlayersVP" v-model="showOtherPlayersVP" id="realTimeVP-checkbox">
+                            <label for="realTimeVP-checkbox">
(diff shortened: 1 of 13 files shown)


kberg:
Thanks! I'll give this a proper look on Saturday.

kberg:
Hi @d-little . Thanks again for this PR! Note that there are some lint failures. You should be able to find those with `npm run lint`, but also, they're here: https://github.com/terraforming-mars/terraforming-mars/actions/runs/3443012788/jobs/5748527525

src/client/components/create/CreateGameForm.vue
```diff
@@ -258,8 +258,24 @@
                         </div>
 
                         <div class="create-game-page-column" v-if="playersCount > 1">
+
                             <h4 v-i18n>Multiplayer Options</h4>
 
+                            <input type="checkbox" v-model="randomFirstPlayer" id="randomFirstPlayer-checkbox">
```
kberg, line 264: Hi! Can you say more about why you reorganized these as part of this PR?

src/client/components/create/CreateGameForm.vue
```diff
@@ -323,15 +336,19 @@
...
+                            <div class="create-game-subsection-label" v-i18n>Multiplayer Variants</div>
+
+                            <template v-if="prelude">
+                              <input type="checkbox" v-model="twoCorpsVariant" id="twoCorps-checkbox">
+                              <!-- We really need a warning here if there arent enough corps for everyone in the game.
+                                  eg: pCount * (startingCorporations+4) > 'Number of Corps in all activated Expansions' - 'Excluded Corps'
+                              -->
+                              <label for="twoCorps-checkbox" title="Always gain the Merger Prelude card (will be given post-draft)">
+                                    <div class="create-game-expansion-icon expansion-icon-prelude"></div>
+                                    <span v-i18n>Merger</span>&nbsp;<a href="https://github.com/terraforming-mars/terraforming-mars/wiki/Variants#Merger" class="tooltip" target="_blank">&#9432;</a>
```
kberg, line 348: We'll need to update the wiki to include this. I'll take care of that after this is done.

src/server/Game.ts
```diff
@@ -623,7 +623,15 @@ export class Game implements Logger {
 
   private gotoInitialResearchPhase(): void {
     this.phase = Phase.RESEARCH;
+    
+    // As each player who doesn't have Merger is dealt Merger in SelectInitialCards.ts,
+    // remove it from the deck to avoid possible conflicts (e.g. Valley Trust / New Partner)
+    if (this.gameOptions.twoCorpsVariant) {
```
kberg, line 629: This is a good catch, however, this is probably not the place for this change, as this occurs after the game object has been fully created. Can you move this to `GameCards.ts`?

Instead of `return this.addCustomCards(...)` store that result in a local variable, and filter on _that_.


image

src/client/components/create/CreateGameForm.vue
```diff
@@ -279,11 +295,8 @@
                             <label for="corporationsDraft-checkbox">
                                 <span v-i18n>Corporations Draft</span>
                             </label>
-                            <input type="checkbox" v-model="randomFirstPlayer" id="randomFirstPlayer-checkbox">
-                            <label for="randomFirstPlayer-checkbox">
-                                <span v-i18n>Random first player</span>
-                            </label>
 
+                            <div class="create-game-subsection-label" v-i18n>Milestones/Awards</div>
```
kberg, line 299: I can't add this comment in the right place, because GH won't let me, but update the check around line 987 of this file to include this check: https://github.com/nwai90/terraforming-mars/blob/ux_improvements/src/components/create/CreateGameForm.ts#L537

kberg (commented):
Thank you for this.

If you could add a test to `GameCards.spec.ts` that ensures that Merger does not appear in the twoCorp variant, I'd be grateful. Let me know if you need help figuring that out.

src/client/components/create/CreateGameForm.vue
d-little: I think I made these changes early, and then kinda forgot I included them.  Part of it was a minor reshuffle to try to figure out where to even put this menu option.. 

I'm happy to scrap all of these changes as a part of the PR and resubmit a minor shuffle of menu items for discussion.  I'll just include the Merger option underneath the existing 'Multiplayer Options'

src/client/components/create/CreateGameForm.vue
d-little: Can do, but this might actually need even more additional checks.  In the edge case that Mergers is enabled, and no custom corp list is enabled, we still need to check to ensure that all of the expansions included have the number of corps required.  I'm not currently sure how to go about checking that!

src/server/inputs/SelectInitialCards.ts
```diff
@@ -31,6 +34,16 @@ export class SelectInitialCards extends AndOptions implements PlayerInput {
       ),
     );
 
+    // Give each player Merger in this variant, or an extra prelude if they already have Merger
```
kberg, line 37: Shouldn't the change to GameCards prevent the "or an extra prelude" condition?

src/server/Player.ts
```diff
@@ -902,10 +902,16 @@ export class Player {
       cards = passedCards;
     }
 
-    const message = cardsToKeep === 1 ?
+    let tmpmsg = cardsToKeep === 1 ?
       'Select a card to keep and pass the rest to ${0}' :
       'Select two cards to keep and pass the rest to ${0}';
 
+    if (this.game.gameOptions.twoCorpsVariant) {
```
kberg, line 909: I don't recommend this. First of all, this will probably appear during normal game drafts, too. Even if you fixed this, I don't think this belongs here, but in SelectInitialCards.vue.

If you remove this change from the PR, I'll merge it in and get it out Monday. The change to SelectInitialCards can be dealt with as a fast follow.

kberg (commented):
This looks good! Once the two changes go in, we can move forward.

src/server/Player.ts
d-little: Agreed, the more I explore the codebase the better I understand where things go.  Reverted this

src/server/inputs/SelectInitialCards.ts
d-little: Yup, leftover debt from before we moved the remove-Merger step.  Good catch

kberg (approved):

kberg:
Good work. Once the tests run one last time, I'll merge it in, and push it in the morning. Thank you!

d-little:
> 

Ugh can you tell I only really know python?

I'll get the linting fixed

PR #5125, opened 2022-11-14
Translation fix

nicobubulle:
Another translation fix.

### DIFF
--- a/src/client/components/Awards.vue
+++ b/src/client/components/Awards.vue
@@ -2,16 +2,7 @@
   <div class="awards_cont" v-trim-whitespace>
     <div class="awards">
       <div class="ma-title">
-        <a
-          @click.prevent="toggleList"
-          class="ma-clickable awards-padding"
-          href="#"
-          v-i18n
-          data-test="toggle-awards"
-        >
-          Awards
-        </a>
-
+        <a @click.prevent="toggleList" class="ma-clickable awards-padding" href="#" data-test="toggle-awards" v-i18n>Awards</a>
         <span
           v-for="award in fundedAwards"
           :key="award.name"
--- a/src/client/components/colonies/Colony.vue
+++ b/src/client/components/colonies/Colony.vue
@@ -38,13 +38,13 @@
       <div v-if="colony.name === ColonyName.VENUS" class="resource wild" style="margin:15px 10px 10px 20px;">?<div class="card-icon tag-venus" style="color: white;margin-top: -36px;margin-left: 16px;"></div></div>
 
       <div v-if="colony.name === ColonyName.PALLAS" style="display:inline-block">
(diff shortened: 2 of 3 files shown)


kberg:
Lint is complaining: https://github.com/terraforming-mars/terraforming-mars/actions/runs/3461566578/jobs/5779337004#step:5:66

src/client/components/Awards.vue
```diff
@@ -2,16 +2,7 @@
...
-          @click.prevent="toggleList"
-          class="ma-clickable awards-padding"
-          href="#"
-          v-i18n
-          data-test="toggle-awards"
-        >
-          Awards
-        </a>
-
+        <a @click.prevent="toggleList" class="ma-clickable awards-padding" href="#" data-test="toggle-awards" v-i18n>Awards</a>
```
kberg, line 5: This isn't actually a change, is it?

src/client/components/colonies/Colony.vue
```diff
@@ -86,30 +82,22 @@
...
-      </span>
-      <span v-if="colony.name === ColonyName.HYGIEA" class="colony-background-color" style="margin-left: 3px;">
-        Trade Income: Steal 3 indicated resources
-      </span>
-      <span v-if="colony.name === ColonyName.LEAVITT" class="colony-background-color" style="margin-left: 3px;">
-        Trade Income: Draw X cards and keep 1
-      </span>
+      <span v-if="colony.name !== ColonyName.EUROPA && colony.name !== ColonyName.PLUTO && colony.name !== ColonyName.MERCURY && colony.name !== ColonyName.IAPETUS && colony.name !== ColonyName.HYGIEA && colony.name !== ColonyName.TITANIA && colony.name !== ColonyName.LEAVITT && colony.name !== ColonyName.PALLAS" class="colony-background-color"
+      v-i18n>Trade Income</span>
+      <span v-if="colony.name === ColonyName.PLUTO" class="colony-background-color" style="position:relative; top:-3px"
```
kberg, line 87: I appreciate the reformatting, but it makes reading the meaningful differences tricky. If all you did is add v-i18n in these places, can you just keep it as it was?

kberg (changes requested):

src/client/components/colonies/Colony.vue
nicobubulle: I noticed that if the string is not inside the tag, the translation is not parsed :
v-i18n>
Trade Income
</span>
Don't work but
v-i18n>Trade Income</span>
works

src/client/components/Awards.vue
nicobubulle: same issue here

kberg (approved):

PR #5149, opened 2022-11-27
Disable the revised corporation counting algorithm.

kberg:
This will at least allow further development to continue. This does mean that someone who custom-selects corporations AND the two corps variant might have issues during gameplay, but one thing at a time.


kberg:
FYI @d-little

PR #5150, opened 2022-11-27
Translation fix for cards with duplicated names

emittmann:
On 11/24/2022 luxidream reported in #bug-reports that "magnetic field generators shows as its english name in the log when playing in spanish"

This error is able to be replicated when playing `Magnetic Field Generators:promo` but not when playing `Magnetic Field Generators`. This can be fixed by removing the `:promo` from the English text to translate.

Given that the English logs only log the card name and not if it is `:promo` or not, this PR attempts to do the same for other languages. If translation fails, and it identifies a card extension of `:promo`, `:SP`, `:ares`, or `:Pathfinders` it removes the card extension and retries translation.

If there are better ways to fix this I'm happy to go with something else - I'm new to this repo so I easily could have tried to fix it in the wrong spot/ method.


kberg:
Thank you for making a commit! This is a good thing to fix, but unfortunately, this isn't the place for the change. Most of the suffix work is done inside the UI itself.

Is there a reason `Magnetic Field Generators:promo` can't be translated?

emittmann:
We could add a new translation for `Magnetic Field Generators:promo` and other promo cards in [promo.json](https://github.com/terraforming-mars/terraforming-mars/blob/main/src/locales/es/promo.json), which will be very roughly (10 cards that have a suffix)*(10 languages to translate) so on the order of 100 new lines of translation. If you think that is the best path I'm happy to do it; my gut reaction had just been that there might be a better way to do this given that we already have the non-suffixed cards translated. Let me know if you think this is the desired path?

I'll also keep poking at the UI files and learning about the translation package, as I'd like to understand them better. And then maybe I'll be able to see if there is something in [LogPanel.vue](https://github.com/terraforming-mars/terraforming-mars/blob/main/src/client/components/LogPanel.vue) or similar I can edit to fix this!

kberg:
Looking  at LogPanel.vue, it's ... it's one of the parts of the code where HTML is generated by typescript, which isn't ideal here.

But ... ah .. I think I found it.

```
    cardToHtml(cardType: CardType, cardName: string) {
      const cardNameString = this.$t(cardName);
      const suffixFreeCardName = cardNameString.split(':')[0];
      const className = cardTypeToCss[cardType];

      if (className === undefined) {
        return suffixFreeCardName;
      }
      return '<span class="log-card '+ className + '">' + suffixFreeCardName + '</span>';
    },
```

See how the first thing it does is translate the card name, and _then_ removes the suffix?

Try this.

```
    cardToHtml(cardType: CardType, cardName: string) {
      const suffixFreeCardName = cardName.split(':')[0];
      const className = cardTypeToCss[cardType];

      if (className === undefined) {
        return suffixFreeCardName;
      }

      return '<span class="log-card '+ className + '">' + $t(suffixFreeCardName) + '</span>';
    },
```

emittmann:
Neat, that does seem like a much better place to solve this issue, and with one small additional `this.` works for the three suffixed cards I've been testing on!

kberg (approved):

PR #5167, opened 2022-12-03
Extract function for pulling player ID from RPCs.

kberg:


kberg:
The tests are proving too brittle.

PR #5197, opened 2023-01-04
Fix Postgres purge.

kberg:
This has been broken for a while. Two things:

1. The early return was broken, so the operation never worked.
2. Even worse, the SQL was wrong.

### DIFF
--- a/src/server/database/PostgreSQL.ts
+++ b/src/server/database/PostgreSQL.ts
@@ -159,7 +159,7 @@ export class PostgreSQL implements IDatabase {
   async purgeUnfinishedGames(maxGameDays: string | undefined = process.env.MAX_GAME_DAYS): Promise<void> {
     const dateToSeconds = daysAgoToSeconds(maxGameDays, 10);
     const selectResult = await this.client.query('SELECT DISTINCT game_id FROM games WHERE created_time < to_timestamp($1)', [dateToSeconds]);
-    const gameIds = selectResult.rows.map((row) => row.game_id);
+    const gameIds = selectResult.rows.slice(0, 1000).map((row) => row.game_id);
     console.log(`${gameIds.length} games to be purged.`);
     if (gameIds.length > 1000) {
       gameIds.length = 1000;
--- a/src/server/database/PostgreSQL.ts
+++ b/src/server/database/PostgreSQL.ts
@@ -157,15 +157,19 @@ export class PostgreSQL implements IDatabase {
 
   // Purge unfinished games older than MAX_GAME_DAYS days. If this environment variable is absent, it uses the default of 10 days.
   async purgeUnfinishedGames(maxGameDays: string | undefined = process.env.MAX_GAME_DAYS): Promise<void> {
-    if (maxGameDays) {
-      const dateToSeconds = daysAgoToSeconds(maxGameDays, 10);
-      const selectResult = await this.client.query('SELECT DISTINCT game_id FROM games WHERE created_time < $1', [dateToSeconds]);
-      const gameIds = selectResult.rows.map((row) => row.game_id);
-      const deleteGamesResult = await this.client.query('DELETE FROM games WHERE game_id in ANY($1)', [gameIds]);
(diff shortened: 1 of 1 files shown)


kberg:
I tested this manually a few times, and in order to expedite the cleanse, I ran a manual purge. It leaves a few games behind, but will do the trick. h/t to @bafolts for getting that started.

```
terraforming-mars::DATABASE=> select count( distinct game_id) from games where created_time < '2022-12-20';
 count  
--------
 407541
(1 row)

terraforming-mars::DATABASE=> delete from participants where game_id in (select distinct game_id from games where created_time < '2022-12-20');
DELETE 407541
terraforming-mars::DATABASE=> delete from games where game_id in (select distinct game_id from games where created_time < '2022-12-20');
```

src/server/database/PostgreSQL.ts
```diff
@@ -157,15 +157,19 @@ export class PostgreSQL implements IDatabase {
...
-      const gameIds = selectResult.rows.map((row) => row.game_id);
-      const deleteGamesResult = await this.client.query('DELETE FROM games WHERE game_id in ANY($1)', [gameIds]);
-      console.log(`Purged ${deleteGamesResult.rowCount} rows from games`);
-      const deleteParticipantsResult = await this.client.query('DELETE FROM participants WHERE game_id in ANY($1)', [gameIds]);
-      console.log(`Purged ${deleteParticipantsResult.rowCount} rows from participants`);
+    const dateToSeconds = daysAgoToSeconds(maxGameDays, 10);
+    const selectResult = await this.client.query('SELECT DISTINCT game_id FROM games WHERE created_time < to_timestamp($1)', [dateToSeconds]);
+    const gameIds = selectResult.rows.map((row) => row.game_id);
+    console.log(`${gameIds.length} games to be purged.`);
+    if (gameIds.length > 1000) {
```
bafolts, line 164: I would do this before `selectResults.rows.map`. That way we only map 1000 rows at most versus millions.

kberg: You know, I'm not quite confident I can change queryResult.rows.length.

For what it's worth, we're talking hundreds of thousands right now. :D

bafolts: The selectResult.rows is an Array like any other Array. If concerned about setting the length property you could use slice(0, 1000) to take the first 1000. This wouldn't modify the selectResult.rows array but still only map the first 1000. My concern here is memory related. The hundreds of thousands of rows are going to take up memory. The call to map will also have to iterate through hundreds of thousands of rows which is going to block the thread.

src/server/database/PostgreSQL.ts
```diff
@@ -157,15 +157,19 @@ export class PostgreSQL implements IDatabase {
...
-      const dateToSeconds = daysAgoToSeconds(maxGameDays, 10);
-      const selectResult = await this.client.query('SELECT DISTINCT game_id FROM games WHERE created_time < $1', [dateToSeconds]);
-      const gameIds = selectResult.rows.map((row) => row.game_id);
-      const deleteGamesResult = await this.client.query('DELETE FROM games WHERE game_id in ANY($1)', [gameIds]);
-      console.log(`Purged ${deleteGamesResult.rowCount} rows from games`);
-      const deleteParticipantsResult = await this.client.query('DELETE FROM participants WHERE game_id in ANY($1)', [gameIds]);
-      console.log(`Purged ${deleteParticipantsResult.rowCount} rows from participants`);
+    const dateToSeconds = daysAgoToSeconds(maxGameDays, 10);
+    const selectResult = await this.client.query('SELECT DISTINCT game_id FROM games WHERE created_time < to_timestamp($1)', [dateToSeconds]);
+    const gameIds = selectResult.rows.map((row) => row.game_id);
```
bafolts, line 162:
```suggestion
    const gameIds = selectResult.rows.slice(0, 1000).map((row) => row.game_id);
```

src/server/database/PostgreSQL.ts
kberg: Thanks. Accepted.

bafolts (approved):

PR #5205, opened 2023-01-07
Don't draw the Restricted Area card before paying.

### LINKED ISSUE #5191: Helion payment timing allowing early information
When playing as Helion, and when paying for Restricted Area's action, the card is drawn before the payment method is decided. In most cases the payment method's timing is not an issue as long as the player is aware what they wish to spend, but in this case they can react to the card they draw before choosing a resource to pay with, which is not intended.

kberg:
Fixes #5191


kberg:
Manually verified.

Issue #5206, opened 2023-01-07
Turn order incorrect and inconsistent with draft information


mwestphal:
```
URL: https://terraforming-mars.herokuapp.com/spectator?id=s5eec58fc5eb6&noredirect=
Player color: yellow
Step: 616
Version: 71d0a3b, built at Fri, 06 Jan 2023 15:08:58 GMT
Browser: Firefox 108
```

During the whole game, the turn order was incorrect and drafting information was incorrect (because of it ? ).

https://terraforming-mars.herokuapp.com/game?id=gc0f80521837

I'm Yellow, supposed to be the third player but in the log we can see:

https://terraforming-mars.herokuapp.com/the-end?id=p6538486fb773

That I played second on first generation.

The drafting information was also incorrect as it said that "You are giving your card to B" but In reality I was giving to "C".
Same for all other players.

PR #5208, opened 2023-01-08
Revert "Rewrite playerIsFinishedTakingActions."

kberg:
Reverts terraforming-mars/terraforming-mars#5174


kberg:
This is the most likely cause for #5206

Issue #5183, opened 2022-12-12
Possibly broken translations


kberg:
> I now have 3 different examples of translations that are blocked from being translated even though the translation has already worked. The translations are also not mine, so I cannot have created the error. What could have gone wrong here? Was the layout/structure changed? Has the English text changed?

1.  "${0} raised the ${1} planetary track 1 step": "${0} erhöhte die ${1}-Planeten-Wertung um 1", 
2.  "Gain 1 M€ for each heat production you have": "Erhalte 1 M€ für jeden Punkt deiner Wärme-Produktion",
3. "Gain 1 heat for each heat production you have": "Erhalte 1 Wärme für jeden Punkt deiner Wärme-Produktion", 
4. "${0} moved 1 animal from Bioengineering Enclosure to ${1}": "${0} verschob 1 Tier aus dem Bioengineering-Gehäuse nach ${1}",

alrusdi:
Let me try to fix it :-)

Borbarad2:
Thank you very much looking forward to these.
Maybe there are some few more examples, which i don't know yet. So if these examples has an issue in common, then there is a generall problem.

kberg:
> Let me try to fix it :-)

Thanks! Happy to have you digging into it.

Borbarad2:
> Let me try to fix it :-)

What is the issue here? Did you find anything?

alrusdi:
I don't found any general problems yet. We have some changes in source strings like dot in the end of 
`${0} moved 1 animal from Bioengineering Enclosure to ${1}.`
That changes should be reflected in all translations to work properly. 
And I still digging what happened with others.

Borbarad2:
With the latest update, issue #2, #3, #4 are solved.
Issue #4 "${0} moved 1 animal from Bioengineering Enclosure to " is solved as well.
Only issue #1 the pathfinder log message that the planetary tracks are raised is not solved with the last issue. This remains.

For me we could close that issue #5183: Possibly broken translations. (we could instead open a new issue just for that case as reminder, up to you, guys)
But keep in mind that something still blocks some translation from being translated.

alrusdi:
Great! I'll give #1 another try

Borbarad2:
I will check with my next pathfinder game if it is working or not. Then we can close that issue.

kberg:
Is this fixed? @Borbarad2, @alrusdi

alrusdi:
It chould be fixed

[...]

PR #5210, opened 2023-01-08
issue-5183 fix planetary track increase log message translation

alrusdi:


Borbarad2:
Thank you very much. 
Looking forward to see that.

kberg (approved):
Thanks. While I think the first version is slightly easier to read, this is appropriate.

kberg:
@alrusdi is #5183 completely fixed?

alrusdi:
I hope so :-)

PR #5232, opened 2023-01-15
CEO PR#3 - Ender, Floyd, HAL9000, Karen

d-little:
4 easy CEOs that do not modify any other files.

* Ender
* Floyd
* HAL9000
* Karen

Tests are written and working as expected


src/server/cards/leaders/Ender.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  public canAct(player: Player): boolean {
+    return player.cardsInHand.length > 0 && this.isDisabled === false;
+  }
+
+  public action(player: Player): PlayerInput | undefined {
+    const max = Math.min(player.cardsInHand.length, player.game.generation * 2);
+
+    return new SelectAmount(
```
kberg, line 42: Interesting. This would be better as a SelectCard input, instead of a SelectAmount input. While that can be done later, add a TODO for this?

// TODO(d-little): Replace with SelectCard.

src/server/cards/leaders/HAL9000.ts
```diff
@@ -0,0 +1,65 @@
...
+  public override play() {
+    return undefined;
+  }
+
+  public canAct(): boolean {
+    return this.isDisabled === false;
+  }
+
+  public action(player: Player): PlayerInput | undefined {
+    const prd = player.production;
```
kberg, line 41: I bet this will do the same with less.

```
  public action(player: Player): PlayerInput | undefined {
    for (const type of Units.keys) {
      const adjustment = Units.of({});
      adjustment[type] = -1;
      if (player.production.canAdjust(adjustment)) {
        player.production.adjust(adjustment, {log: true});
        adjustment[type] = 4;
        player.addUnits(adjustment, {log: true});
      }
    }

    this.isDisabled = true;
    return undefined;
  }
```

I don't like how it initializes, but I'm pretty sure this ought to work.

kberg: Again, this might make you waste time if I'm wrong, but it's something to keep in mind.

src/server/cards/leaders/Karen.ts
```diff
@@ -0,0 +1,66 @@
...
+  constructor() {
+    super({
+      name: CardName.KAREN,
+      cardType: CardType.LEADER,
+      metadata: {
+        cardNumber: 'L11',
+        renderData: CardRenderer.builder((b) => {
+          b.opgArrow().text('X').prelude().asterix();
+        }),
+        description: 'Once per game, draw Prelude cards equal to the current generation number and choose one to play.',
```
kberg, line 23: ", choose one to play, and discard the rest."

That might be worth adding to the community server, if you feel like it.

src/server/cards/leaders/Karen.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  public canAct(): boolean {
+    return this.isDisabled === false;
+  }
+
+  public action(player: Player): PlayerInput | undefined {
+    const cardsDrawn: Array<IProjectCard> = [];
+    const game = player.game;
+    for (let i = 0; i < player.game.generation; i++) {
```
kberg, line 41: `i < game.generation`. You do create the local variable `game` after all.

src/server/cards/leaders/Karen.ts
```diff
@@ -0,0 +1,66 @@
...
+      cardsDrawn.push(game.preludeDeck.draw(game));
+    }
+
+    cardsDrawn.forEach((card) => {
+      if (card.canPlay !== undefined && card.canPlay(player) === false ) {
```
kberg, line 46: I am pretty sure this can be shortened to `card.canPlay?.(player) === false`. But if the tests don't support this, then you'll have to manually verify.

(I actually strive for tests that thorough if I can. If you want help with some of those, let me know.)

src/server/cards/leaders/Karen.ts
```diff
@@ -0,0 +1,66 @@
...
+        cardsDrawn.splice(cardsDrawn.indexOf(card), 1);
+        player.game.log('${0} was discarded as ${1} could not afford to pay for it', (b) => b.card(card).player(player));
```
kberg, line 48: It's not necessarily that the player couldn't afford it. The card might just not be playable (For instance you might have -5 MC production, and the prelude requires you lose 1 MC production. So, consider `... as ${1} could not play it`,

src/server/cards/leaders/Karen.ts
```diff
@@ -0,0 +1,66 @@
...
+      player.game.log('${0} drew no playable prelude cards', (b) => b.player(player));
+      return undefined;
+    }
+
+    return new SelectCard('Choose prelude card to play', 'Play', cardsDrawn, ([card]) => {
+      if (card.canPlay === undefined || card.canPlay(player)) {
+        this.isDisabled = true;
+        return player.playCard(card);
+      } else {
+        throw new Error('You cannot pay for this card');
```
kberg, line 62: Why is this additional conditional necesary? Perhaps @nwai90 can say.

tests/cards/leaders/HAL9000.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+  let card: HAL9000;
+  let player: TestPlayer;
+  let player2: TestPlayer;
+
+  beforeEach(() => {
+    card = new HAL9000();
+    player = TestPlayer.BLUE.newPlayer();
+    player2 = TestPlayer.RED.newPlayer();
+    Game.newInstance('gameid', [player, player2], player);
+    player.production.add(Resources.PLANTS, 0);
```
kberg, line 18: Can be shortened to `player.production.adjust({plants: 0, megaCredits: -1, steel: 1, titanium: 1, energy: 1, heat: 1});`

Perhaps it's `production.adjust(Units.of({...});`

tests/cards/leaders/HAL9000.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+    player.production.add(Resources.MEGACREDITS, -1);
+    player.production.add(Resources.STEEL, 1);
+    player.production.add(Resources.TITANIUM, 1);
+    player.production.add(Resources.ENERGY, 1);
+    player.production.add(Resources.HEAT, 1);
+  });
+
+  it('Takes OPG action', function() {
+    expect(player.production.plants).eq(0);
```
kberg, line 27: `expect(player.production.asUnits()).deep.eq({plants: 0, megaCredits: -1, steel: 1, titanium: 1, energy: 1, heat: 1});`

kberg (commented):
Nothing major.

src/server/cards/leaders/Karen.ts
d-little: This style of comparison tastes so bad to me, but hopefully I'll catch on and start using it :)

kberg: I've been there. Hopefully you can, too. There's a lot of great stuff that Typescript does for us.

kberg (commented):
Broken test, and one more simplification.

kberg (approved):

tests/cards/leaders/Karen.spec.ts
```diff
@@ -0,0 +1,70 @@
+import {expect} from 'chai';
```
kberg, line 1: Something's up with this card:

```
1) Karen
       Takes action in Generation 4:

      AssertionError: expected [ Array(3) ] to have a length of 4 but got 3
      + expected - actual

      -3
      +4
      
      at Context.<anonymous> (tests\cards\leaders\Karen.spec.ts:47:34)

```

And since it didn't happen in both environments, that means something non-deterministic is happening.

Do you know how to run a single test class? I can look later.

kberg: Digging into this, the problem is at the start of the test. I added the two lines at the bottom.

```
    for (let i = 0; i < 3; i++) {
      game.deferredActions.runAll(() => {});
      forceGenerationEnd(game);
    }
    console.log(game.generation);
    expect(game.generation).eq(4);
```

You would expect game.generation to be 4, but it's 3.

This test fails ~9-10% of the time.

kberg: I think I know what's going on. This has to do with Turmoil. What's going on is that the Global Event deck is shuffled. When Aquifer Released By Public Council appears, it means the generation doesn't _actually_ end. No, the game is still waiting for someone to do some Ocean-type action.

This is fixable. I wont' block this PR any further.

PR #5235, opened 2023-01-16
Update the Leader tests.

kberg:


tests/cards/leaders/Ender.spec.ts
```diff
@@ -32,15 +29,15 @@ describe('Ender', function() {
     player.cardsInHand.push(new Research(), new MicroMills());
     expect(card.canAct(player)).is.true;
 
-    const selectAmount = card.action(player) as SelectAmount;
+    const selectAmount = cast(card.action(player), SelectAmount);
```
kberg, line 32: Use `cast()` instead of `as`. cast is used only for testing, but it's just a little bit better. `as` forces us to believe that `card.action` return an instance of `SelectAmount`, but `cast` verifies it. This means that once in a while the `cast` call will throw an error because the expectation of the forced typecast from `as` was wrong. And that's a good thing.

tests/cards/leaders/Karen.spec.ts
```diff
@@ -17,11 +17,11 @@ describe('Karen', function() {
 
   beforeEach(() => {
     card = new Karen();
-    player = TestPlayer.BLUE.newPlayer();
-    player2 = TestPlayer.RED.newPlayer();
-    const gameOptions = setCustomGameOptions({preludeExtension: true});
-    game = Game.newInstance('gameid', [player, player2], player, gameOptions);
+    const game = newTestGame(2, {preludeExtension: true});
```
kberg, line 20: `setCustomGameOptions` enables too many features. It enables Turmoil by default. And that's what caused the issue in your test.

Now: why should a test enable so many features, even one that surprises us? So instead of creating games and players as these normally go, swap those with newTestGame() and getTestPlayer(). It's a little bit of work, but I hope they're a little more reliable.

It's at least the standard that the tests all have now.

kberg:
@d-little FYI

alrusdi (approved):

PR #5251, opened 2023-01-20
Allow temperature reduction at maximum temperature

amulder87:


kberg:
I appreciate a well-written test. But as I mentioned in https://github.com/terraforming-mars/terraforming-mars/issues/5250 I believe this is working as intended.

amulder87:
Agreed, it is working as expected, thank you for the quick response!

PR #5263, opened 2023-01-26
CEO PR#4.1 - CardFinder.ts

d-little:
_PR#1 in a set of card+deck related PRs_

### CardFinder.ts

I pretty much just tried to replicate the code as-is for Preludes, however there are a few oddities I'm not 100% sure on.  

For starters, `getLeaderByName`+`getPreludeByName` are only ever referenced in this `CardFinder.ts` in their respective `leadersFromJSON`+`preludesFromJSON` blocks in this same file.  Not sure why they'd need to be split out?

I've also included `leaderCards` in the generic `getCardByName`, but again, not sure this needed

Prior to this I'd just used the generic `cardsFromJSON` and it worked Fine, but would rather get this right if necessary.

(This whole PR for CardFinder.ts may be unnecessary, but will want to get it ironed out before moving on to `Deck.ts`+`Game.ts`

### DIFF
--- a/src/server/CardFinder.ts
+++ b/src/server/CardFinder.ts
@@ -4,6 +4,7 @@ import {CardManifest, ModuleManifest} from './cards/ModuleManifest';
 import {CardName} from '../common/cards/CardName';
 import {ICorporationCard} from './cards/corporation/ICorporationCard';
 import {IPreludeCard} from './cards/prelude/IPreludeCard';
+import {ILeaderCard} from './cards/leaders/ILeaderCard';
 import {ALL_MODULE_MANIFESTS} from './cards/AllCards';
 
 const CARD_RENAMES = new Map<string, CardName>([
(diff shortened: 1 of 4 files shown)


src/server/CardFinder.ts
```diff
@@ -75,6 +80,23 @@ export class CardFinder {
     return result;
   }
 
+  public leadersFromJSON(cards: Array<CardName>): Array<ILeaderCard> {
```
kberg, line 83: This should be reduced to one main function now that it's repeated three times. But it's OK.

kberg (commented):
This PR is fine. Add the test. :D

src/server/CardFinder.ts
d-little: > Add the test. :D

What test ? :o

I can add in some cardfinder tests when there are some cards to find!  But, at the moment, I'm not sure where to add more tests for this new block

kberg:
https://github.com/terraforming-mars/terraforming-mars/blob/restart-turn/tests/CardFinder.spec.ts

kberg:
It's not in the `server` directory for historical reasons.

kberg (approved):

PR #5267, opened 2023-01-28
CEO PR#4.4 - Deck

d-little:
_PR#4 in a set of card+deck related PRs_

### [src/server/cards/Deck.ts](https://github.com/terraforming-mars/terraforming-mars/compare/main...d-little:leaders-Deck?expand=1#diff-954dcd314b4559b0fed3b2df781925caee54ea9853546a1ac95cbb591020810e)

Another easy one.  Just a clone of Preludes really

### [tests/cards/Deck.spec.ts](https://github.com/terraforming-mars/terraforming-mars/compare/main...d-little:leaders-Deck?expand=1#diff-20d0e550025dab8dee2ba255f5d9542ec3a33059956384e57db92112164f4a50)

This one is trickier.  I'm not 100% sure why the Preludes test is hard coded, but.. it is.  I don't really want to update this test every single time we push in a new CEO during deployment so I'm leaving it dynamic.  We are testing to see if the drawing works as expected, and the serialisation works after that. So maybe the hard coded length isn't ever necessary.

I _could_ add a test that we draw a custom list of Leaders, and work with that.  A deck size of 5 leaders maybe?  But I _think_ the test intentions are met as-is

I'm also not sure what the `customCorporationsList: [CardName.MERGER]` on the Prelude test is doing? I didn't replicate it in any case

### DIFF
--- a/tests/cards/Deck.spec.ts
+++ b/tests/cards/Deck.spec.ts
@@ -1,5 +1,5 @@
 import {expect} from 'chai';
-import {PreludeDeck} from '../../src/server/cards/Deck';
+import {PreludeDeck, LeaderDeck} from '../../src/server/cards/Deck';
 import {GameCards} from '../../src/server/GameCards';
 import {DEFAULT_GAME_OPTIONS} from '../../src/server/GameOptions';
 import {CardName} from '../../src/common/cards/CardName';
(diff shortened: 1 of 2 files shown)


tests/cards/Deck.spec.ts
```diff
@@ -72,3 +72,42 @@ describe('PreludeDeck', function() {
...
 });
+
+describe('LeaderDeck', function() {
+  const random = new UnseededRandom();
+
+  it('serialization compatibility', () => {
+    const deck = new LeaderDeck(new GameCards(
+      {...DEFAULT_GAME_OPTIONS,
+        leadersExtension: true,
+        preludeExtension: true,
```
kberg, line 83: I don't mind, but why enable prelude? What happens otherwise?

kberg (approved):
Looks good, but I have a question.

tests/cards/Deck.spec.ts
d-little: Good question;  I think I had some expansion-specific tests here but then realised it wasn't the right place.  We have enough CEOs in the game now that this can be removed (We have 4 non-expansion CEOs, we need at least 3 for this test).

I can add in the Prelude tests or we can remove this game option

kberg: I think I prefer removing the option, at your leisure.

PR #5273, opened 2023-01-29
CEO PR# 5.2 - CEO Group 2 - Asimov

d-little:
## CEO Asimov

This 'group' is just Asimov alone.  There's enough going on here that he deserves to be alone.

### Asimov.ts

A couple of private functions for Asimov, hopefully self explanatory.  The main takeaway is that we grab _all_ awards from all expansions, remove awards currently in the game + any award that requires an expansion to use.  Then offer those up to be played.

### Asimov.spec.ts

A lot of tests, hopefully fully covered all of the use cases.  Nothing really unusual

### [IAward.ts](https://github.com/terraforming-mars/terraforming-mars/compare/main...d-little:leaders-CEOAsimov?expand=1#diff-727f6191936fa70c1a0fca045ee76644095333434dfdeec7de66b8e5ba54f820)

IAward.ts has a new function getAdditionalScore.  It currently only checks to see if the player has Asimov in play.

This means that every single other Award has been updated to all return the same line:  `return score + getAdditionalScore(player);`

Many of the awards I needed to do some minor changes to do this, but I wanted a consistent method of returning that exact same line every time.  Where previously one might have used `return someValue + something;` it will now `const score = someValue + something;`, followed by the above return.

### DIFF
--- a/src/common/constants.ts
+++ b/src/common/constants.ts
@@ -53,6 +53,9 @@ export const DEFAULT_ESCAPE_VELOCITY_THRESHOLD = 30;
 export const DEFAULT_ESCAPE_VELOCITY_PERIOD = 2;
 export const DEFAULT_ESCAPE_VELOCITY_PENALTY = 1;
 
+// Leaders/CEOs
+export const ASIMOV_AWARD_BONUS = 2;
+
 export const LANGUAGES2 = {
   en: 'English',
   de: 'Deutsch',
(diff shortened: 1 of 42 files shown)


kberg:
https://github.com/terraforming-mars/terraforming-mars/pull/5274 would significantly simplify this.

d-little:
Closing pending #5274, will reorganize and either reopen or send a new PR

PR #5276, opened 2023-02-01
Restart turn

kberg:
This is still a work in progress. It is meant to allow a player to restart the action they're currently taking.


d-little (commented):
Yeah, way beyond my knowledge.  Will need to poke memories90 to review

PR #5283, opened 2023-02-02
Update AncientShipyards.ts

Borbarad2:
Adjustment for ancient shipyards for solo mode only. The orginal effect is back and it is now a multiple "Indentured Workers" card, thus balanced as +8M€ for -1VP multiple times if you wish. So its fair for solo,

This would be the way how it should be copied over to the Darkside of the Moon Syndicate corp for solo mode only.

### DIFF
--- a/src/server/cards/moon/AncientShipyards.ts
+++ b/src/server/cards/moon/AncientShipyards.ts
@@ -45,7 +45,7 @@ export class AncientShipyards extends Card {
       p.stealResource(Resources.MEGACREDITS, 2, player);
     }
     if (game.isSoloMode()) {
-      player.addResource(Resources.MEGACREDITS, 2);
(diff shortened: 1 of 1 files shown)


kberg:
This change breaks tests: https://github.com/terraforming-mars/terraforming-mars/actions/runs/4075378006/jobs/7021693386

```
  1) AncientShipyards
       act solo:

      AssertionError: expected 18 to equal 12
      + expected - actual

      -18
      +12
      
      at Context.<anonymous> (tests/cards/moon/AncientShipyards.spec.ts:68:32)
```

Borbarad2:
Could you, please fix the test for me to get that PR right? thanks, Robert. 
It is just a number/value to adjust, nothing complicate.

PR #5292, opened 2023-02-05
Fix deployment to work in any path

Hietamaki:
Changed all links (that I could find) to relative instead of absolute form in client. 

With this users can deploy to any path on their server, not just the root directory.

So address like:
https://own-home-server.com/path/to/mars/

works too.

Tests ran all ok. Also tested locally and remotely.


kberg (commented):
Thanks for this. Let's talk about paths.

src/client/components/LoadGameForm.vue
```diff
@@ -26,15 +26,15 @@ export default Vue.extend({
       const gameId = this.$data.gameId;
       const rollbackCount = this.$data.rollbackCount;
       const xhr = new XMLHttpRequest();
-      xhr.open('PUT', paths.LOAD_GAME);
+      xhr.open('PUT', paths.LOAD_GAME.substring(1));
```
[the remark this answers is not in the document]
Hietamaki: That was my first thought as well but it broke the server and I couldn't figure out where the problem was so I kind of gave up on it and went the easy road here.

src/styles/players_overview.less
```diff
@@ -158,7 +158,7 @@
         }
         .points-per-tag {
             position: absolute;
-            background: url(/assets/board/mars.png);
+            background: url(assets/board/mars.png);
```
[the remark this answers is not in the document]
Hietamaki: Thanks for the feedback. I'll look into it.

src/client/components/LoadGameForm.vue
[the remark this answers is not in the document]
Hietamaki: Okay I figured out where I went wrong. Should work now.

Hietamaki:
The issues mentioned should be fixed now. Tested to be working.

src/client/components/App.ts
```diff
@@ -144,24 +144,24 @@ export const mainAppSettings = {
               window.location.search.includes('&noredirect') === false
             ) {
               app.screen = 'the-end';
-              if (currentPathname !== '/the-end') {
+              if (currentPathname !== paths.THE_END) {
```
kberg, line 147: This is reasonable. Thank you.

src/client/components/GameHome.vue
```diff
@@ -101,13 +101,15 @@ export default Vue.extend({
...
+        return `spectator?id=${playerId}`;
       }
-      return `/player?id=${playerId}`;
+      return `player?id=${playerId}`;
     },
     copyUrl(playerId: ParticipantId | undefined): void {
       if (playerId === undefined) return;
-      copyToClipboard(window.location.origin + this.getHref(playerId));
+
+      const path = window.location.href.replace(/game\?id=.*/, "");
```
kberg, line 111: What is this regex doing?

src/server/requestProcessor.ts
```diff
@@ -70,12 +70,13 @@ export function processRequest(
   }
 
   const url = new URL(req.url, `http://${req.headers.host}`);
+  const pathname = url.pathname.substring(1); // Remove leading '/'
```
kberg, line 73: I was able to directly edit your repo, and added the comment.

src/client/components/App.ts
```diff
@@ -119,12 +119,12 @@ export const mainAppSettings = {
     getVisibilityState(targetVar: string): boolean {
       return (this as unknown as MainAppData).componentsVisibility[targetVar] ? true : false;
     },
-    update(path: '/player' | '/spectator'): void {
-      const currentPathname = window.location.pathname;
+    update(path: typeof paths.PLAYER | typeof paths.SPECTATOR): void {
```
kberg, line 122: Nice.

kberg (commented):
In general, this looks good.

What have you done to confirm that this works? Have you tested the admin paths? Have you possibly tested the clone feature? (If you haven't, I plan to. But I'd like to know what testing you've done.)

src/common/app/paths.ts
```diff
@@ -1,22 +1,22 @@
...
-export const GAMES_OVERVIEW = '/games-overview' as const;
-export const HELP = '/help' as const;
-export const LOAD = '/load' as const;
-export const LOAD_GAME = '/load_game' as const;
-export const NEW_GAME = '/new-game' as const;
-export const PLAYER = '/player' as const;
-export const PLAYER_INPUT = '/player/input' as const;
-export const SPECTATOR = '/spectator' as const;
-export const THE_END = '/the-end' as const;
+export const ADMIN = 'admin' as const;
```
kberg, line 1: I hate to tell you this, but you might find yourself with some merge conflicts. Hopefully they're not too much. WaitingFor.vue might be the worst.

Hietamaki:
Thanks for the warning, I managed to merge the main. I ran the automated tests and did some user testing (played few mp games). Everything seems to work, also the admin paths. I'm not sure about the cloning feature, I don't know what that is.

kberg:
@Hietamaki this is failing lint checks. Run `npm run lint` and adjust accordingly. Thank you!

Hietamaki:
Fixed

assets/index.html
```diff
@@ -5,7 +5,8 @@
             href="https://fonts.googleapis.com/css?family=Ubuntu&display=swap"
             rel="stylesheet"
         />
-        <link rel="stylesheet" type="text/css" media="all" href="/styles.css" />
+        <link rel="stylesheet" type="text/css" media="all" href="styles.css" />
+        <link rel="shortcut icon" href="favicon.ico" type="image/x-icon" />
```
kberg, line 9: nb this isn't about fixing the deployment, but adding something new. Keep that in mind in future PRs to keep things easy to manage.

src/client/components/App.ts
```diff
@@ -239,3 +239,8 @@ export const mainAppSettings = {
     }
   },
 };
+
+function getCurrentPath() {
```
kberg, line 243: The path is considered `/the/whole/path`, so the name `getCurrentPathname` is misleading. This is returning the last fragment of the path. I realize this is not your doing, but how about naming this `getLastPathSegment`?

This also introduces a minor issue, it forces future paths that might appear here to only be one level deep. Is there any way to instead know the base url, and remove that from the front?

src/client/components/GameHome.vue
```diff
@@ -101,13 +101,15 @@ export default Vue.extend({
...
+        return `spectator?id=${playerId}`;
       }
-      return `/player?id=${playerId}`;
+      return `player?id=${playerId}`;
     },
     copyUrl(playerId: ParticipantId | undefined): void {
       if (playerId === undefined) return;
-      copyToClipboard(window.location.origin + this.getHref(playerId));
+      // Get current location path without game?id=xxxxxxx
+      const path = window.location.href.replace(/game\?id=.*/, '');
```
kberg, line 111: `replace(/\/[^\/]*$/\/)`? Perhaps not.

kberg: Your version is slightly problematic of the parent URL contains 'game' anywhere.

src/server/routes/Route.ts
```diff
@@ -58,7 +58,7 @@ export class Route {
   public downgradeRedirect(_req: http.IncomingMessage, res: http.ServerResponse, ctx: Context): void {
     const url = new URL(ctx.url); // defensive copty
     url.searchParams.set('serverId', ctx.ids.statsId);
-    res.writeHead(301, {Location: url.toString()});
+    res.writeHead(301, {Location: url.pathname + url.search});
```
kberg, line 61: Have you tested this case?

kberg (commented):
You know how it goes. The more I read, the more I understand, the more questions. :D Hopefully we're close. Thanks.

src/client/components/GameHome.vue
Hietamaki: It only matches when it finds `game?id=` so it shouldn't be a problem as far as I can tell.

src/client/components/App.ts
Hietamaki: That's actually a good point. The match could be more specific since we know the current page name.

src/server/routes/Route.ts
Hietamaki: Turns out this change actually did nothing. The redirect doesn't work like I thought it would. I'm out of easy ideas how to fix it properly. This is imo. difficult function to understand its intention. If it's only used for the Stats API I could fit it for that or just leave it as it is for now, it's not major issue since it's no breaking any existing functionality.

src/client/components/App.ts
Hietamaki: Actually nvm what I just said. I think the only proper way would be to use environment variable for the base url.

Hietamaki:
Those two things you pointed out are a bit trickier to solve properly and this change set is becoming too big to maintain. I'd like to get majority of this merged to minimize the surface area for potential conflicts. I suggest we simplify this and convert this PR into preparatory work for the the feature. I can convert the getCurrentPathname/getLastPathSegment to its original implementation and I'll get back to this a bit later with better time.

src/client/components/GameHome.vue
kberg: Good. Point.

src/client/components/App.ts
kberg: Alright, we can let this go for now. Mind adding a comment saying that it forces future paths that might appear here to only be one level deep? It can be fixed later.

kberg: I'll add the comment.

kberg:
> Those two things you pointed out are a bit trickier to solve properly and this change set is becoming too big to maintain. I'd like to get majority of this merged to minimize the surface area for potential conflicts. I suggest we simplify this and convert this PR into preparatory work for the the feature if that's okay. I can convert the getCurrentPathname/getLastPathSegment to its original implementation and I'll get back to this a bit later with better time. All the changes made would be then internal and they don't affect any functionality.

Yes, I agree with the general sentiment to do this in pieces if necessary. I just broke one of my PRs at work into five smaller ones because it turned out I was overreaching. This is less so, and I wouldn't quite call it overreaching, but you know, we learn the scope as we work, sometimes.

kberg (approved):

kberg:
Thank you!

PR #5293, opened 2023-02-06
CEO#4.7 - DeferredAction PlaceHazardTile 

d-little:
### [src/server/deferredActions/PlaceHazardTile.ts](https://github.com/terraforming-mars/terraforming-mars/compare/main...d-little:ceoPlaceHazardTile?expand=1#diff-e243312c29bfacbd52cc95772f1bf0f3c85eade8c2cad715c2d5a5a3a1ca95f8)

Hopefully fairly obvious.  Currently nothing uses this, but Caesar will.  After that there are a handful of Community Cards that interact with Hazards in Ares.

### DIFF
--- a/src/server/Game.ts
+++ b/src/server/Game.ts
@@ -1376,6 +1376,17 @@ export class Game implements Logger {
     });
   }
 
+  public addHazardTile(
+    player: Player, space: ISpace, hazardType: TileType): void {
+    this.addTile(
+      player,
+      space,
+      {
+        tileType: hazardType,
+      },
+    );
+  }
+
   public canAddOcean(): boolean {
     return this.board.getOceanCount() < constants.MAX_OCEAN_TILES;
   }
--- a/src/server/Game.ts
+++ b/src/server/Game.ts
@@ -1376,17 +1376,6 @@ export class Game implements Logger {
     });
   }
 
-  public addHazardTile(
-    player: Player, space: ISpace, hazardType: TileType): void {
-    this.addTile(
-      player,
-      space,
-      {
-        tileType: hazardType,
-      },
-    );
-  }
-
   public canAddOcean(): boolean {
     return this.board.getOceanCount() < constants.MAX_OCEAN_TILES;
   }
--- a/src/server/deferredActions/PlaceHazardTile.ts
+++ b/src/server/deferredActions/PlaceHazardTile.ts
@@ -1,0 +1,41 @@
+import {Player} from '../Player';
+import {SelectSpace} from '../inputs/SelectSpace';
+import {ISpace} from '../boards/ISpace';
+import {DeferredAction, Priority} from './DeferredAction';
+import {_AresHazardPlacement} from '../ares/AresHazards';
+import {TileType} from '../../common/TileType';
+import {PlacementType} from '../boards/PlacementType';
+
+
+export class PlaceHazardTile extends DeferredAction {
+  constructor(
+    player: Player,
+    private options?: {
+      on?: PlacementType,
+      title?: string,
+      hazardType?: TileType,
+    }) {
+    super(player, Priority.DEFAULT);
+  }
+
+  public execute() {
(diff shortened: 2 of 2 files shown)


src/server/Game.ts
```diff
@@ -1376,6 +1376,17 @@ export class Game implements Logger {
     });
   }
 
+  public addHazardTile(
```
kberg, line 1379: This isn't really how a hazard tile gets placed. See `_AresHazardPlacement.putHazardAt`. Perhaps this is obvious to me, and not to someone else. But a hazard isn't _really_ treated like a tile in the traditional sense.

Anyway, use that method, even though _AresHazardPlacement is meant to be "package private". I'll deal with that later.

kberg: In fact, remove this method, and call .putHazardAt in PlaceHazardTile.

src/server/deferredActions/PlaceHazardTile.ts
```diff
@@ -0,0 +1,41 @@
...
+import {PlacementType} from '../boards/PlacementType';
+
+
+export class PlaceHazardTile extends DeferredAction {
+  constructor(
+    player: Player,
+    private options?: {
+      on?: PlacementType,
+      title?: string,
+      hazardType?: TileType,
```
kberg, line 16: Typescript fun! Change `TileType` to `TileType.DUST_STORM_MILD | TileType.EROSION_MILD`

src/server/Game.ts
kberg: By the way, this has the additional effect of giving someone placement bonuses for a hazard, which should not apply. (FYI @nwai90 in case you want to consider that for the Community server.)

src/server/deferredActions/PlaceHazardTile.ts
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  public execute() {
+    // TODO: Confirm, would this ever NOT be land?
+    const type = this.options?.on || 'land';
+    const availableSpaces = this.player.game.board.getAvailableSpacesForType(this.player, type);
+    if (availableSpaces.length === 0) {
+      return undefined;
+    }
+    const hazardType = this.options?.hazardType || Math.floor(Math.random() * 2) === 0 ? TileType.DUST_STORM_MILD : TileType.EROSION_MILD;
```
kberg, line 28: This is a little janky. First, I think the card doesn't say it's a randomly-selected hazard type. Second, why would that matter? It only matters somewhat thematically, but I think it's worth considering. The different global parameters conditions for Ares makes a) certain hazard types technically unavailable b) certain hazard types severe or c) intentionally removed.

The simplest thing, I think, is to merely say that the card places a mild erosion. That way it doesn't get removed when the temperature goes above a certain amount, and can still be upgraded to severe. This is a thematic problem I've had with  placing a hazard tile for some time.

src/server/deferredActions/PlaceHazardTile.ts
```diff
@@ -0,0 +1,41 @@
...
+      on?: PlacementType,
+      title?: string,
+      hazardType?: TileType,
+    }) {
+    super(player, Priority.DEFAULT);
+  }
+
+  public execute() {
+    // TODO: Confirm, would this ever NOT be land?
+    const type = this.options?.on || 'land';
```
kberg, line 23: No. Hazard spaces are only playable on available land spaces. It should be reasonable to remove this.options.on.

kberg (changes requested):

src/server/deferredActions/PlaceHazardTile.ts
d-little: Makes sense.  I've made the hazardType mandatory, but I imagine _most_ cards will place mild erosion.  I'll need to eyeball the other community cards (if we add them) to see if they should add anything other than this.

kberg (approved):

PR #5296, opened 2023-02-07
CEO PR# 5.4 - CEOs Xavier, Yvonne, Zan

d-little:
### Xavier

Pretty straight forward, a hook in [src/server/player/Tags.ts](https://github.com/terraforming-mars/terraforming-mars/compare/main...d-little:ceosXavierYvonneZan?expand=1#diff-589997eb1b5bf44b74bdb5b6ca8fab4d59981d4253d0123f66ecb4a890fe26dd) + [CeoExtension.ts](https://github.com/terraforming-mars/terraforming-mars/compare/main...d-little:ceosXavierYvonneZan?expand=1#diff-2a96143b4b90a5ed4f588de70afbb47f4e2672f268a1bff867f8e0a00d59b325)

### Yvonne

Also straightforward, let me know if this was not the correct method to give bonuses to one player though:
`player.game.defer(new SimpleDeferredAction(player, () => colony.giveColonyBonus(player.game.getPlayerById(owner))));`

### Zan

Zan will need some rendering items in the near future, for now it's text.

One hook via [src/server/turmoil/parties/PartyHooks.ts](https://github.com/terraforming-mars/terraforming-mars/compare/main...d-little:ceosXavierYvonneZan?expand=1#diff-5f4b9996aa4ee9c97b175b0352aa5ecb668f21accd8dfd39b41700bdd5ddaf5a) to ignore Reds negatives

### DIFF
--- a/src/server/cards/ceos/Zan.ts
+++ b/src/server/cards/ceos/Zan.ts
@@ -8,7 +8,6 @@ import {CeoCard} from './CeoCard';
 import {PartyName} from '../../../common/turmoil/PartyName';
 
 import {Turmoil} from '../../turmoil/Turmoil';
-import {SimpleDeferredAction} from '../../deferredActions/DeferredAction';
 
 export class Zan extends Card implements CeoCard {
   constructor() {
--- a/src/server/cards/ceos/Zan.ts
+++ b/src/server/cards/ceos/Zan.ts
@@ -1,0 +1,58 @@
+import {CardName} from '../../../common/cards/CardName';
+import {CardType} from '../../../common/cards/CardType';
+import {Player} from '../../Player';
+import {PlayerInput} from '../../PlayerInput';
+import {Card} from '../Card';
+import {CardRenderer} from '../render/CardRenderer';
+import {CeoCard} from './CeoCard';
+import {PartyName} from '../../../common/turmoil/PartyName';
+
+import {Turmoil} from '../../turmoil/Turmoil';
+import {SimpleDeferredAction} from '../../deferredActions/DeferredAction';
+
+export class Zan extends Card implements CeoCard {
+  constructor() {
+    super({
+      name: CardName.ZAN,
+      cardType: CardType.CEO,
+      metadata: {
+        cardNumber: 'L26',
+        renderData: CardRenderer.builder((b) => {
+          b.br.br;
+          // b.redsInactive().asterix();
+          b.text('(NoRedsNeg)').asterix();
+          b.br.br;
+          // b.opgArrow().text('ALL').delegates(1).colon().nbsp.nbsp.party(PartyName.REDS);
+          b.opgArrow().text('ALL').delegates(1).colon().text('REDS');
+        }),
+        description: 'You are immune to Reds\' ruling policy. Once per game, place all your delegates in Reds.',
+      },
+    });
+  }
+
+  public isDisabled = false;
+
+  public override play() {
+    return undefined;
+  }
+
+  public canAct(): boolean {
+    return this.isDisabled === false;
+  }
+
+  public action(player: Player): PlayerInput | undefined {
+    const game = player.game;
+    const turmoil = Turmoil.getTurmoil(game);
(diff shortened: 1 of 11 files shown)


src/server/cards/ceos/Zan.ts
```diff
@@ -0,0 +1,58 @@
...
+        cardNumber: 'L26',
+        renderData: CardRenderer.builder((b) => {
+          b.br.br;
+          // b.redsInactive().asterix();
+          b.text('(NoRedsNeg)').asterix();
+          b.br.br;
+          // b.opgArrow().text('ALL').delegates(1).colon().nbsp.nbsp.party(PartyName.REDS);
+          b.opgArrow().text('ALL').delegates(1).colon().text('REDS');
+        }),
+        description: 'You are immune to Reds\' ruling policy. Once per game, place all your delegates in Reds.',
```
kberg, line 28: "place all your available delegates" i assume? and not just all of the delegates.

src/server/cards/ceos/Zan.ts
```diff
@@ -0,0 +1,58 @@
...
+    return this.isDisabled === false;
+  }
+
+  public action(player: Player): PlayerInput | undefined {
+    const game = player.game;
+    const turmoil = Turmoil.getTurmoil(game);
+    // const reserveDelegates = turmoil.delegateReserve.filter((d) => d === player.id).length;
+    const reserveDelegates = turmoil.getAvailableDelegateCount(player.id);
+    for (let i = 0; i < reserveDelegates; i++) {
+      game.defer(new SimpleDeferredAction(player, () => {
```
kberg, line 49: Use `player.defer()`

src/server/cards/ceos/Zan.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  public canAct(): boolean {
+    return this.isDisabled === false;
+  }
+
+  public action(player: Player): PlayerInput | undefined {
+    const game = player.game;
+    const turmoil = Turmoil.getTurmoil(game);
+    // const reserveDelegates = turmoil.delegateReserve.filter((d) => d === player.id).length;
+    const reserveDelegates = turmoil.getAvailableDelegateCount(player.id);
```
kberg, line 47: Actually, you can simplify this by putting all of it in the deferred action:
```
  while (turmoil.getAvailableDelegateCount(player.id) > 0) {
    turmoil.sendDelegateToParty(player.id, PartyName.REDS, game);
  }
```

src/server/turmoil/parties/PartyHooks.ts
```diff
@@ -38,6 +39,9 @@ export class PartyHooks {
         policyId = rulingParty.policies[0].id;
       }
 
+      // Hook for CEO Leader Zan's effect (Skip all Reds Policy effects)
+      if (player.cardIsInEffect(CardName.ZAN) && policyId.startsWith('r')) return false;
```
kberg, line 43: Perhaps this is better: `if (partyName === PartyName.REDS && player.cardIsInEffect(...))`.

First, it puts the party comparison first, which will be slightly faster, and it uses the party rather than policy. (If they ever don't match, that's a real problem.)

kberg (commented):
Basically seems good.

Off-topic: I'm starting to think about how to optimize some of the constant calculations we run. We must look at tag counts and card is in effect calls zillions of times, and optimizing those might have a noticeable speed improvement.

But that's for later.

src/server/cards/ceos/Zan.ts
d-little: This reminds me, does getAvailableDelegateCount get the Lobby delegate?  I've forgotten, but I think it does... and Lobby is just displayed in rendering

kberg: It does. The original code (and community server) separated them, but the code to manage the two was unwieldy, ubiquitous, and unnecessary.

kberg:
`src/server/cards/ceos/Zan.ts(11,1): error TS6133: 'SimpleDeferredAction' is declared but its value is never read.`

kberg (approved):

PR #5297, opened 2023-02-07
Remove CeoCard as an interface.

kberg:

PR #5298, opened 2023-02-07
Add CeoCard as a support class like the other card types.

kberg:
This follows on from #5297. It reduces repetition, and also uses the StaticCard pattern which significantly reduces memory consumption.


d-little:
Amazing work



Any reason for the difference in approach for Naomi and Yvonne canAct overrides?  is the canAct() better, performance wise?  Or negligible.  They're in opposite evaluating order


Naomi:
```ts
  public override canAct(player: Player): boolean {
    if (!super.canAct(player)) {
      return false;
    }
    return openColonies.length > 0;
```

Yvonne :
```ts
  public override canAct(player: Player): boolean {
    return player.game.gameOptions.coloniesExtension === true && super.canAct(player);
  }
```

Should they be:

```ts
  public override canAct(player: Player): boolean {
    if (!super.canAct(player)) return false;
    return openColonies.length > 0;
```

```ts
  public override canAct(player: Player): boolean {
    if (!super.canAct(player)) return false;
    return player.game.gameOptions.coloniesExtension;
```

src/server/cards/ceos/Naomi.ts
```diff
@@ -29,15 +25,12 @@ export class Naomi extends Card implements ICeoCard {
     });
   }
 
-  public isDisabled = false;
-
-  public override play() {
-    return undefined;
-  }
-
-  public canAct(player: Player): boolean {
```
d-little, line 38: Any reason for the difference in approach for Naomi and Yvonne canAct overrides?  is the canAct() better, performance wise?  Or negligible.  They're in opposite evaluating order


Naomi:
```ts
  public override canAct(player: Player): boolean {
    if (!super.canAct(player)) {
      return false;
    }
    return openColonies.length > 0;
```

Yvonne :
```ts
  public override canAct(player: Player): boolean {
    return player.game.gameOptions.coloniesExtension === true && super.canAct(player);
  }
```

Should they be:

```ts
  public override canAct(player: Player): boolean {
    if (!super.canAct(player)) return false;
    return openColonies.length > 0;
```

```ts
  public override canAct(player: Player): boolean {
    if (!super.canAct(player)) return false;
    return player.game.gameOptions.coloniesExtension;
```

kberg: It's a little arbitrary, but I'm using these rules.

1. Do faster evaluations first if possible.
2. Make it readable.

kberg: Rules/Guidelines

d-little (approved):

d-little:
Looks good, assuming these tests pass, send it

PR #5319, opened 2023-02-09
CEO#4.9 - Game.ts

d-little:
### Game.ts

No drafting, that can come later. Just the deck + hand dealing so we can get testing first.

### DIFF
--- a/src/server/SerializedGame.ts
+++ b/src/server/SerializedGame.ts
@@ -20,6 +20,7 @@ export type SerializedGame = {
     aresData?: AresData;
     awards: Array<string>;
     board: SerializedBoard;
+    ceoDeck: SerializedDeck;
     currentSeed: number;
     claimedMilestones: Array<SerializedClaimedMilestone>;
     clonedGamedId?: string;
--- a/src/server/SerializedGame.ts
+++ b/src/server/SerializedGame.ts
@@ -20,7 +20,8 @@ export type SerializedGame = {
     aresData?: AresData;
     awards: Array<string>;
     board: SerializedBoard;
-    ceoDeck: SerializedDeck;
+    // TODO(dl): Remove conditional on ceoDeck after 2023-03-20
+    ceoDeck?: SerializedDeck;
     currentSeed: number;
     claimedMilestones: Array<SerializedClaimedMilestone>;
     clonedGamedId?: string;
--- a/tests/cards/ceo/Bjorn.spec.ts
+++ b/tests/cards/ceo/Bjorn.spec.ts
@@ -17,7 +17,9 @@ describe('Bjorn', function() {
 
   beforeEach(() => {
     card = new Bjorn();
-    game = newTestGame(4, {ceoExtension: true});
+    // Todo: Prelude extension is only activated here as we didnt have enough CEOs in early testing
+    // In the future, when more base-game CEOs are added, we can remove prelideExtension from here
+    game = newTestGame(4, {ceoExtension: true, preludeExtension: true});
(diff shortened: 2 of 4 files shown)


src/server/SerializedGame.ts
```diff
@@ -20,6 +20,7 @@ export type SerializedGame = {
     aresData?: AresData;
     awards: Array<string>;
     board: SerializedBoard;
+    ceoDeck: SerializedDeck;
```
kberg, line 23: For many games, this will be undefined to start with. Without care, this kind of thing makes it possible for the server to throw errors while loadingr legacy games.

The standard way we've dealt with that is by changing this to `ceoDeck?: SerializedDeck;`. And then in `deserialize` or here, add a comment "Remove the conditional by..." and pick a date at least 2 weeks from now. Maybe 4.

tests/cards/ceo/Bjorn.spec.ts
```diff
@@ -17,7 +17,9 @@ describe('Bjorn', function() {
 
   beforeEach(() => {
     card = new Bjorn();
-    game = newTestGame(4, {ceoExtension: true});
+    // Todo: Prelude extension is only activated here as we didnt have enough CEOs in early testing
```
kberg, line 20: nit: capitalize TODO. Especially since you're going to change code anyway.

tests/cards/ceo/Will.spec.ts
```diff
@@ -21,7 +21,7 @@ describe('Will', function() {
 
   beforeEach(() => {
     card = new Will();
-    game = newTestGame(4, {ceoExtension: true});
+    game = newTestGame(1, {ceoExtension: true});
```
kberg, line 24: Why the change from 4 to 1?

kberg (commented):
Much easier to review without the draft. Much easier.

tests/cards/ceo/Will.spec.ts
d-little: We only have one test player here.

CEOs cannot currently support 4 players without expansions, we dont have enough in the base game.  Cleaner to switch to one player.

Now that I think about it, does this matter for Solo vs. Multiplayer game testing?  For this example I dont think it should, but some CEOs might care if there are no opponents

src/server/SerializedGame.ts
d-little: Remove Conditional By?  Or After date x?

kberg: Doesn't matter. We'll know.

tests/cards/ceo/Will.spec.ts
kberg: /shrug you tell me!

src/server/SerializedGame.ts
d-little: Ah, making it optional freaks out CeoDeck.deserialize() because it could come through as undefined.

I _think_ it should be handled with some temporary code that looks like this?  Maybe?  I'm just not sure what to put in the `{}`, it needs to be type SerializedDeck

```ts
    // TODO(dl): remove ?? {} by 2023/03/20
    const ceoDeck = CeoDeck.deserialize(d.ceoDeck ?? {}, rng);

```

kberg: SerializedDeck is of this type:

```
export type SerializedDeck = {
  drawPile: Array<CardName>;
  discardPile: Array<CardName>;
}
```

So the empty deck is `{drawPile: [], discardPile: []}`

kberg (approved):

PR #5329, opened 2023-02-09
Optimize the intersection check

aloisdg:
Hi!

This is a minor modification to optimize a very narrow part of the code. It does not matters much but here it goes :)

Thanks a lot for your work


tests/ma/MilestoneAwardSelector.spec.ts
```diff
@@ -6,12 +6,22 @@ import {chooseMilestonesAndAwards, LIMITED_SYNERGY, maximumSynergy, verifySynerg
 import {IMilestone} from '../../src/server/milestones/IMilestone';
 import {RandomMAOptionType} from '../../src/common/ma/RandomMAOptionType';
 import {testGameOptions} from '../TestingUtils';
-import {intersection} from '../../src/common/utils/utils';
 
 function toNames(list: Array<IMilestone | IAward>): Array<string> {
   return list.map((e) => e.name);
 }
 
+/**
```
kberg, line 14: This would be better off in utils, for the same reason that hasIntersection is better in utils  instead of Counter.

kberg (changes requested):
Minor change. Thank you!

kberg:
Also, this looks like it isn't going to build:

```

> terraforming-mars@1.0.0 build:server
> tsc --build src/tsconfig.json

src/common/utils/utils.ts(27,3): error TS2322: Type 'boolean' is not assignable to type 'T[]'.
The command '/bin/sh -c npm run build' returned a non-zero code: 1
Error: Process completed with exit code 1.
```

Please make sure to run `npm install`, `npm run build` and `npm run test`, and then `npm run lint`. Thanks!

tests/ma/MilestoneAwardSelector.spec.ts
aloisdg: Not sure how the code is deployed, but I guessed (and I a may be wrong), that the tests folder where not a part of the deploy-able. Since `intersection` is not needed in the client nor the server, is there a specific reason to move it under `src`?

aloisdg:
I didn't try the code jet. I will do it later on.

tests/ma/MilestoneAwardSelector.spec.ts
kberg: I wasn't thinking about tests/utils, but I'd really rather it be back where it was. The additional cost on the client-side download isn't going to be meaningful. (Having just cut the download size in half.)

Utility methods in the `tests` package are typically for the testing API only.

Leave it where you put it, but I will move it back there. The speed optimization is very welcome, though!

aloisdg:
Error come from the linter. I will have to clone the code after all... :)

tests/ma/MilestoneAwardSelector.spec.ts
aloisdg: I am ok with that. Done.

kberg (approved):

kberg:
Thank you for this!

aloisdg:
You are welcome! Thank your for the help!

PR #5330, opened 2023-02-09
 CEO#4.8.1 - SelectInitialCards.vue - Add missing this.hasCeo to saveData

d-little:
###

As it says on the box.  Required when turn on CEOs


kberg (approved):

kberg:
Let's get tests in there for some of this. Good reminder.

PR #5333, opened 2023-02-10
CEO cards should not be counted as Blue Actions

d-little:
Creating this PR with an error, it'll fail tests.

This is open for discussion, but CEOs on community have their own 'action item'.  Currently in the CEO code as-is the CEO actions show up in both Blue Card Actions and its own CEO Action.

This PR is an attempt to remove the CEO Action from the 'Blue Card Actions' list.  However, it's mad that CEO cards dont have defined actions.

Im also unsure of the impact on removing CEO's from `getAvailableBlueActionCount()`, but from my brief check I'm pretty sure the only place that's used is on the game-screen to show how many actions each player has remaining for that generation.  I suggest CEO's arent counted here at all, and instead maybe a new item is added in the bar indicating whether the CEO is:
* OPG has not been used
* OPG has been activated and has an ongoing effect this generation
* OPG has been used, and is inactive


src/server/Player.ts
```diff
@@ -780,15 +780,21 @@ export class Player {
   public getPlayableActionCards(): Array<ICard & IActionCard> {
     const result: Array<ICard & IActionCard> = [];
     for (const playedCard of this.tableau) {
-      if (isIActionCard(playedCard) && !this.actionsThisGeneration.has(playedCard.name) && playedCard.canAct(this)) {
+      if (isIActionCard(playedCard) && !this.actionsThisGeneration.has(playedCard.name) && playedCard.canAct(this) && !isCeoCard(playedCard)) {
```
kberg, line 783: Put isCeoCard before canAct since it's probably faster.

kberg:
OK, this looks good. Let's get a test for this.

kberg (changes requested):
Yeah, test is pending.

kberg (approved):

kberg (approved):

PR #5335, opened 2023-02-10
Use partition instead of two filters

aloisdg:
Hello,

another small upgrade. Calling two `filters` to split an array is akin to using [`partition`](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-collections-listmodule.html#partition). Alas `partition` is not standard yet in JavaScript... The good news is that it is really easy to implement with reduce. Hopefully `partition` will be added by the [tc39](https://es.discourse.group/t/array-prototype-partition-like-filter-but-two-arrays-for-true-and-false/447). Then, we will have to replace `partition(a, f)` with `a.partition(f)`.

On more thing is a small word on performance. Theoretically `O(N)` and `O(2N)` are the same. But practically, `O(N)` will definitely have a shorter running time, but not significant. If N is large enough, the running time of both will be identical. Well there is that.

[Try it online](https://tio.run/##XZHPasMwDMbPy1OIXupCCGzs1K6FvkNvJgPXUVdDagfZhpbQZ89k5w/dcrH0Sd9PchweHXpNpgvDcIlWB@MsdIqCSdHX6SC8i6RxC0ci9WChhI6wMVoFFkXYwmkD@wOcnWtR2c3cOLcfoC@AP8IQycJIywpAxaCoUQhCH9tQArZ4QxsysJ@akjVV5TJWLG3f8F5XXfTXRdq9uPLA0TzLz3IKpKxLkHU9GZ5FUWhnfQCtqPGwB1m89Vbd@I6ri3MrNi75WdGf/H6/p7yeCI4aJCb0z12RY7lmwrpmKdoGL8Zis1SYlSufH7t5A3kz3qMv4WqCT7XlOUReroR85p80UlJapWW4e/8yhS@XkPwwVet@xMj9J6YhLA3DLw) or [Benchmark it against filters](http://jsben.ch/nJZub)

Note that I didn't try to run the code either yet.

Cheers,


kberg:
Thanks for this, too. We'll see how the tests go. I agree that a nice `partition` method is handy. These optimizations are nice. I really just wish some of them could be measured.

[lint doesn't like what you've done](https://github.com/terraforming-mars/terraforming-mars/actions/runs/4146891468/jobs/7175849839), so hopefully that'll get fixed.

kberg:
Still not building, please try `npm run build && npm run test && npm run lint` (or the Windows equivalent.)

aloisdg:
@kberg I am on linux (manjaro) and it is not working...

aloisdg:
Well I am a bit puzzle here:

aloisdg:
Well...

image

aloisdg:
I have a lead: https://stackoverflow.com/questions/17526805/chai-test-array-equality-doesnt-work-as-expected (`eq` vs `eql`)

aloisdg:
I love CI/CD :)

src/client/utils/utils.ts
```diff
@@ -0,0 +1,16 @@
+/**
```
kberg, line 1: Thanks for this. Can you put this in src/common/utils/utils.ts, and the test in tests/common/utils/utils.spec.ts?

You've inspired me to beef up tests in utils,ts by the way.

aloisdg: Thank you for the kind word. I move them. Glad to be a part of this great project. I have some idea to dig too, but I will start by helping with the french translation (it is my mother tongue).

kberg (approved):
This is great. Thank you.

aloisdg:
Thank you for your reviews

PR #5336, opened 2023-02-10
Remove all direct access to this.$root.

kberg:
This cleans up a lot of code, and a lot of unnecessary unsafe casting, putting it in a single controlled location.


src/client/components/overview/PlayerTags.vue
```diff
@@ -186,6 +186,21 @@ export default Vue.extend({
     tooltipCss(): string {
       return 'tooltip tooltip-' + (this.isTopBar ? 'bottom' : 'top');
     },
+    tags(): Array<TagDetail> {
```
kberg, line 189: While trying to move to vueRoot, this component gave me some trouble. I had hoped turning into a computed value would help, but it didn't. No matter, I kept it in.

PR #5337, opened 2023-02-11
Promo Card #44 - 16 Psyche

d-little:
Per https://boardgamegeek.com/thread/3026060/16-psyche-promo

(Where I wasnt allowed to start the object name with 16 I have simply dropped it.  This means that _some_ objects are named 16Psyche, some are just named Psyche.  I can standardise all of them to Psyche if desired)


kberg (approved):
Wow, nice!

PR #5346, opened 2023-02-12
CEO - Init Faraday

d-little:
### Faraday

A lot of strange moving parts, hopefully they're put together well


src/server/cards/ceos/Faraday.ts
```diff
@@ -0,0 +1,98 @@
...
+      //  this wont work if someone makes a card with > 5 tags of one type, but...
+      if (playerTagCountPrePlay%5 + cardTagCount >= 5) {
+        player.game.defer(new SimpleDeferredAction(player, () => this.effectOptions(player, tag)));
+      }
+    });
+  }
+
+  public effectOptions(player: Player, tag: Tag) {
+    return new OrOptions(
+      new SelectOption(`Pay 2 M€ to draw a ${tag} card`, 'Confirm', () => {
```
kberg, line 61: I wonder if we can make this easier to internationalize, though I don't know how much it'll matter.

src/server/cards/ceos/Faraday.ts
```diff
@@ -0,0 +1,98 @@
...
+        player.game.defer(new SelectPaymentDeferred(player, 2, {title: 'Select how to pay for action'}));
+        player.drawCard(1, {tag: tag});
```
kberg, line 63: use `SelectPaymentDeferred.options.afterPay` to draw the card only once the ability is paid for.

src/server/cards/ceos/Faraday.ts
```diff
@@ -0,0 +1,98 @@
...
+        return undefined;
+      }),
+      new SelectOption('Do nothing', 'Confirm', () => {
+        return undefined;
+      }),
+    );
+  }
+
+  private getAllValidTags(player: Player): Array<Tag> {
```
kberg, line 72: Use `player.tags.getAllTags()`

kberg (commented):
This looks pretty good!

tests/cards/ceo/Faraday.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+    expect(player.megaCredits).to.eq(8);
+    expect(player.cardsInHand).has.length(1);
+    expect(player.cardsInHand[0].tags.includes(Tag.SCIENCE)).is.true;
+  });
+
+  it('Can choose to do nothing when reaching a multiple of 5 for a tag', function() {
+    player.playedCards.push(fakeCard({tags: [Tag.SCIENCE, Tag.SCIENCE, Tag.SCIENCE, Tag.SCIENCE]}));
+    player.playCard(fakeCard({tags: [Tag.SCIENCE]}));
+
+    const orOpations = cast(game.deferredActions.pop()!.execute(), OrOptions);
```
kberg, line 57: orOpations

tests/cards/ceo/Faraday.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+    expect(game.deferredActions).has.lengthOf(0);
+    expect(player.cardsInHand).has.length(0);
+    expect(player.megaCredits).to.eq(10);
+  });
+
+
+  it('Can draw a card when reaching a multiple of 5 for a tag', function() {
+    player.playedCards.push(fakeCard({tags: [Tag.SCIENCE, Tag.SCIENCE, Tag.SCIENCE, Tag.SCIENCE]}));
+    // 4 tags: Not sufficient
+    expect(game.deferredActions).has.lengthOf(0);
```
kberg, line 40: What I do instead of counting deferred actions this way is calling

```
runAllActions();
expect(player.getWaitingFor()).is.undefined;
```

This has a strange side effect of dealing with how games are initialized. (It requires some special behavior which I won't describe tonight because I'm tired. If you just do what I wrote above, you'll hit the problem I'm describing, so in many ways, this is just fine.

tests/cards/ceo/Faraday.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+    // 5 tags: Draw a card with a Science tag
+    player.playCard(fakeCard({tags: [Tag.SCIENCE]}));
+
+    expect(game.deferredActions).has.lengthOf(1);
+    const orOptions = cast(game.deferredActions.pop()!.execute(), OrOptions);
```
kberg, line 45: Same here. Instead of counting deferred actions.

```
runAllActions();
const orOptions = cast(player.popWaitingFor(), OrOptions);
```

tests/cards/ceo/Faraday.spec.ts
d-little: I've copyed the way `HabitatMarte.spec.ts` works, put `player.popWaitingFor()` into `beforeEach`

src/server/cards/ceos/Faraday.ts
d-little: I avoided this at first, and I cannot remember why... I added it in but it might be more complicated than I needed to make it, but I'm not sure how efficient find is vs. associative arrays in TS!

kberg: Given that this is only called when events are played when someone has Faraday, this is OK. As I've said I'm still kind of desperate to measure some of the areas of poor performance.

kberg: By this, I mean "getAllTags"

src/server/cards/ceos/Faraday.ts
```diff
@@ -0,0 +1,87 @@
...
+      return acc;
+    }, {} as { [key in Tag]: number });
+
+    const uniqueTagsOnCard = new Set(card.tags);
+    uniqueTagsOnCard.forEach((tag) => {
+      if (invalidTags.includes(tag)) return;
+      const playerTagCount = countOfTags[tag];
+      const cardTagCount = card.tags.filter((cardTag) => cardTag === tag).length;
+      const playerTagCountPrePlay = playerTagCount - cardTagCount;
+      // Modulo 5 what the tag count was before the card was played.
```
kberg, line 66: If a player meets the requirement for two tag types due to a single card play, they a) aren't told it, and b) not given an option for the order they are resolved.

kberg (changes requested):
OK, let me know when this is updated. Thanks!

src/server/cards/ceos/Faraday.ts
d-little: Would this look something like:

```ts
return new OrOptions(
  if tagsToDraw.length > 0 {
    for (tag in tagsToDraw){ 
      new SelectOption('Pay 2MC to draw card with ${tag} tag', 'Draw card', () => { 
        // pay 2mc
        // draw card
        // return to this loop
      }),
     }
     new SelectOption('Do nothing', 'Confirm', () => {
       // exit the whole loop
     }
  }
);
```

kberg: I had something else in mind, but let's revisit this.

src/server/cards/ceos/Faraday.ts
d-little: Done, I think :)

d-little:
Shall I wait on the Tags work before finalising this?

kberg (approved):

PR #5349, opened 2023-02-12
CEO - Init Lowell, Musashi, Neil

d-little:
### Lowell

Obvious enough, but I dont know why Prelude is required for Lowell.  Maybe minimum CEO count at some point?

### Musashi

Not much to see, but I've renamed Sweeney to Musashi to stay in the alphabet theme

### Neil

Moon man


kberg:
What do you think about keeping the original name for Musashi to keep the two servers in sync? If you don't have a strong reason, I'd rather keep them together.

src/server/cards/ceos/CeoCardManifest.ts
```diff
@@ -58,17 +58,18 @@ export const CEO_CARD_MANIFEST = new ModuleManifest({
     // [CardName.INGRID]: {Factory: Ingrid},
     // [CardName.JANSSON]: {Factory: Jansson},
     [CardName.KAREN]: {Factory: Karen, compatibility: 'prelude'},
-    // [CardName.LOWELL]: {Factory: Lowell, compatibility: 'prelude'},
+    // TODO(dl): Why is Lowell limited to Prelude?
```
kberg, line 61: Ask @memories90, aka @nwai90.

src/server/cards/ceos/Lowell.ts
```diff
@@ -0,0 +1,61 @@
...
+      if (card.canPlay?.(player) === false) {
+        cardsDrawn.splice(cardsDrawn.indexOf(card), 1);
+        game.log('${0} was discarded as ${1} could not play it,', (b) => b.card(card).player(player));
+      }
+    });
+
+    player.game.defer(new SelectPaymentDeferred(player, 8, {title: 'Select how to pay for action'}));
+    this.isDisabled = true;
+
+    return new SelectCard('Choose leader card to play', 'Play', cardsDrawn, (foundCards: Array<IProjectCard>) => {
```
kberg, line 55: You probably don't need to use `afterPay` here as long as the deferred action priorities don't get in the way of each other. But it's worth noting.

src/server/cards/ceos/Musashi.ts
```diff
@@ -0,0 +1,71 @@
...
+      name: CardName.MUSASHI,
+      metadata: {
+        cardNumber: 'L28',
+        renderData: CardRenderer.builder((b) => {
+          b.opgArrow().minus().cards(1, {secondaryTag: Tag.EARTH}).colon().cards(1, {secondaryTag: Tag.SPACE}).titanium(1).asterix();
+          b.br;
+          b.titanium(6);
+          b.br.br;
+        }),
+        description: 'Once per game, discard any number of Earth cards to draw/gain that many Space cards and titanium. Gain 6 titanium.',
```
kberg, line 29: "draw that many space cards, and gain that many units of titanium, plus 6." (also space is lower case.)

src/server/cards/ceos/Musashi.ts
```diff
@@ -0,0 +1,71 @@
...
+      },
+    });
+  }
+
+  public action(player: Player): PlayerInput | undefined {
+    const game = player.game;
+    const eligibleCards = player.cardsInHand.filter((card) => card.tags.includes(Tag.EARTH));
+    const max = eligibleCards.length;
+
+    return new SelectAmount(
```
kberg, line 39: if (eligibleCards.length === 0) {
  game.log('${0} has no Earth cards", ...);
  return undefined;
}

kberg, line 39: Drop the `SelectAmount` piece of this and just use SelectCard, min: 0, max: cards.length.

src/server/cards/ceos/Musashi.ts
```diff
@@ -0,0 +1,71 @@
...
+                  game.projectDeck.discard(card);
+                }
+                return undefined;
+              },
+              {min: amount, max: amount});
+          },
+        );
+        player.game.defer(discardEarthCards, Priority.DISCARD_AND_DRAW);
+        player.game.defer(DrawCards.keepAll(player, amount, {tag: Tag.SPACE}));
+        player.addResource(Resources.TITANIUM, 6 + amount);
```
kberg, line 63: log this.

src/server/cards/ceos/Neil.ts
```diff
@@ -0,0 +1,50 @@
...
+    for (const tag of card.tags) {
+      if (tag === Tag.MOON) {
+        player.game.getCardPlayer(this.name).addResource(Resources.MEGACREDITS, 1, {log: true});
+      }
+    }
+  }
+
+  public action(player: Player): PlayerInput | undefined {
+    const game = player.game;
+    const moonData = game.moonData as IMoonData;
```
kberg, line 40: MoonExpansion.ifMoon(game, (moonData) => {
  const lowestRate = ...
  ...
  ...
});
this.isDisabled = true;
return undefined;

kberg (commented):
I'll read the tests later. Good night!

src/server/cards/ceos/Musashi.ts
d-little: They'll still get the 6 titanium though, so cannot just return here

kberg:
> > What do you think about keeping the original name for Musashi to keep the two servers in sync? If you don't have a strong reason, I'd rather keep them together.
> 
> Entirely personal, just dont really like the man +1

I don't either, but there are two reasons for keeping it. The first is thematic. Elon Musk converting earth tags for space tags? It's kind of on the mark. One of the thematic elements of CEO is a light, if not overt, relationship between the leader and the behavior.

The other reason is that it makes the community server incompatible with the main server. Right now that isn't a problem, but when we start allowing players to choose CEOs like corporations (and one day that'll happen,) it will be.

src/server/cards/ceos/Musashi.ts
```diff
@@ -26,46 +23,35 @@ export class Musashi extends CeoCard {
...
-              {min: amount, max: amount});
-          },
-        );
-        player.game.defer(discardEarthCards, Priority.DISCARD_AND_DRAW);
-        player.game.defer(DrawCards.keepAll(player, amount, {tag: Tag.SPACE}));
-        player.addResource(Resources.TITANIUM, 6 + amount);
-        this.isDisabled = true;
+    if (eligibleCards.length === 0) {
+      game.log('${0} has no Earth cards', (b) => b.player(player));
+      player.addResource(Resources.TITANIUM, 6, {log: true});
```
kberg, line 38: Since players can resolve effects in any order, and card draw has nothing to do with resources in hand, you can add the 6 titanium in one place.

d-little: Over two instances of player.addResource?  One for the initial 6, and another for all of the discarded cards?

kberg: Oh I see. OK.

PR #5352, opened 2023-02-12
Seperate code from Playerhome to own components

Hietamaki:
Doing refactor on PlayerHome to make it more manageable.

### DIFF
--- a/src/client/components/App.ts
+++ b/src/client/components/App.ts
@@ -64,7 +64,7 @@ export const mainAppSettings = {
     settings: raw_settings,
     isServerSideRequestInProgress: false,
     componentsVisibility: {
-      'milestones_list': true,
+      'milestones': true,
       'awards_list': true,
       'tags_concise': false,
       'pinned_player_0': false,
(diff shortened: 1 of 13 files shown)


kberg:
Thanks for the draft. Are you asking for someone to look at it?

Hietamaki:
Sure I'd appreciate to hear your opinion on it.

kberg:
So I'm scanning it, and in general this is a nice reuse of stuff. So thank you for that.

A couple of favors, please:
1. Could you split out any functional changes? You're looking to add a new keyboard shortcut. That involves changing some of our documentation, for example, one bit of which is an image showing keystrokes.
2. Can we do this one component per PR? That would make reading the review much simpler.
3. I notice you name components in kebab-case, whereas we have been using PascalCase. Is that the standard or just your preference? Let's keep them consistent unless you have a strong case for changing them.

kberg:
I'm going to close this out only because I can't review something this large this easily. But this is a good change.

Hietamaki:
Thanks for the feeback, I'll look into it.

PR #5354, opened 2023-02-13
Custom corporations list by player

bli17:
Hi! This is my first pull request on GitHub ever. I'm not that familiar with web development and my code might be clumsy -- I appreciate any suggestions!

This implements the ability to define custom corporations by player. One example use case is a corporation round robin tournament where each player plays every corporation once over a season. As the season progresses, players cannot play corporations they have previously played, which means every player will have a different list of legal corporations they can play.

Some notes about behavior:
- There is a bug where, after downloading or uploading settings with a non-random player order, the settings will not allow changing the first player.
- The corporation deck will be populated with the union of cards in the overall custom corporations list and every player's custom corporations list.
- The custom corporations by player does not work with corporations drafting. I'm not even sure what the intended behavior should be here.
- If a player would be dealt a corporation card but has none remaining in his custom corporations list, they are dealt a Beginner Corporation.

One additional thing this does is rewrite the code for randomizing player order. The new version was my way of tracking how the player order changed so the custom corporations by player array can be re-ordered accordingly. However, I think the old version of the random player order code had a bug and did not fully randomize the player order. IIRC it kept first player first too often, and also kept neighbors together too often. If that problem did exist, the new code should fix that.

There is also a minor wording change to clarify the "End Turn" and "Pass for the Generation" options.


kberg:
Thank you for this contribution. Before we go any further, I think it's important to discuss a few things, some of which are in https://github.com/terraforming-mars/terraforming-mars/blob/main/CONTRIBUTING.md. You might want to read that.

1. This isn't complete. You need to change `SerializedGame`, `game.serialize` and `game.deserialize`.
2. This needs tests. There's a lot of code in the client side that could use some test, and I won't ask you to start testing client code that isn't already tested. But there's server code that is changing, and those need tests. Most importantly, you're adding new behavior. How are we supposed to know that it works? Tests. And I might write a couple of tests around the area you're changing so it becomes a little easier for you to add tests. We'll discuss this.
3. Please address all the errors keeping this from being compiled: https://github.com/terraforming-mars/terraforming-mars/actions/runs/4159100489/jobs/7195239541. These come from `npm run lint`. So before your next contribution, please run `npm run build; npm run test; npm run lint`. (If you're using Windows, it's a slightly different command, I suppose.)
4. This is the tricky part. See, I appreciate the contribution. But, you're the only person I know of so far that wants this. So, is it worth adding the extra load to the server? If people want it, then I'm not opposed to it. But we still have to address a few areas of reliability and correctness. In recent months I've accepted two changes from two well-meaning contributors. But either the tests were incomplete, or they introduced new issues. Who had to fix them? I had to fix them. The contributors were gone. So my point is, helping you here doesn't necessarily mean we should accept it.

As the contributor's guide says:

> Our codebase is growing and changing all the time, and with each change we ask ourselves if it introduces too much risk, resource consumption or unclear code. All of this is to make sure the project remains nimble and easy to maintain, enhance and understand. This is why we adhere to some degree of quality and standards and comprehension, so please do not be surprised if our first response is to ask for clarification or possibly significant changes.

But also:

> We commit to respectfully working with you to ensure your change can be accepted, and also, that it be clear, correct, and well-tested.

kberg:
I'm going to close this pull request. It seems you have added some more features to this branch which are not related to the pull request description. Perhaps you didn't realize it. Anyway, if you want to talk further, you can use this PR (I'll get notified) or Discord.

PR #5355, opened 2023-02-13
Update log_messages.json

Borbarad2:
Add missing log messages

Add work around for gaining M€. If there is a better solution please, tell me


src/locales/de/log_messages.json
```diff
@@ -322,6 +324,16 @@
...
+    "gain 3 M€": "3 M€ erhalten",
+    "gain 6 M€": "6 M€ erhalten",
+    "gain 9 M€": "9 M€ erhalten",
+    "gain 12 M€": "12 M€ erhalten",
+    "gain 15 M€": "15 M€ erhalten",
+    "gain 18 M€": "18 M€ erhalten",
+    "gain 21 M€": "21 M€ erhalten",
+    "gain 24 M€": "24 M€ erhalten",
+    "gain 27 M€": "27 M€ erhalten",
+    "gain 30 M€": "30 M€ erhalten",
```
kberg, line 336: Where do these come from? I'd fix the code so the translation on line 326 is sufficient.

Borbarad2: Yes, please fix the code, then these line are not useful anymore. 
Because line 326 was not working properly, these lines were added and should work as **work around**. 
Of course if line 326 is working these line are not necassary anymore and could be deleted.

Borbarad2: By the way, could you fix the code for "Remove X plants from ... ", because these lines are also not tranlateable and the translation code is blocked by something. 

And for the promo prelude card "Corporate Archives" something/unknown character is in between the two sentences, so that these lines could not be translated. 

Thank you so much for your support and help, making these great app even greater :)

kberg: > Yes, please fix the code, then these line are not useful anymore. Because line 326 was not working properly, these lines were added and should work as **work around**. Of course if line 326 is working these line are not necassary anymore and could be deleted.

I'm asking where in a running instance you saw this happening.

Borbarad2: I saw it in connection with the "Sulphur-Eating Bacteria". That's why the 3M€ jumps.

kberg: I'll fix it in Sulphur-Eating Bacteria, and the code that  removes plants. Closing this PR. Thanks for the report!

Borbarad2: Thank you, too. Thanks for fixing the problem in the code, and not solving it via the tramslation. Perfect. 

No problem for closing the PR although there are more changes in it, which i will put in an other PR any time in future together with other change, too.

PR #5359, opened 2023-02-14
Add array of all tags on the server.

kberg:


d-little:
I like it

Is it worth adding a Game.ts function for the _possible_ tags in the game?  Removing Mars from games not using Pathfinder for example.

It's not a very common use case I'll admit, but it's come up I _think_ more than once :)

kberg:
> I like it
> 
> Is it worth adding a Game.ts function for the _possible_ tags in the game? Removing Mars from games not using Pathfinder for example.
> 
> It's not a very common use case I'll admit, but it's come up I _think_ more than once :)

I looked, and there doesn't seem to be many uses. Plus, client-side uses can't use code in `Game`.

d-little (approved):

Issue #4846, opened 2022-08-29
feature request: strong differentiation of 'pass for this generation' and 'end turn'


waxoid:
Apologies if this is a well-traveled topic! didn't find prior discussion... and first, thank you to this community for this amazing online version.

We just played an epic five player game with a photo finish and I blew the game and final turn by accidentally selecting "pass for this generation" instead of "end turn" on the final round. User error of course, but game decisive and pain level high for the entire board.

As it stands both 'pass' and 'end for this generation' are mixed in with other actions in the menu. Perhaps this is some logic around showing most frequently used commands?

Idea (1) – differentiate 'end for this generation' by placing it last in the list after a separation of some kind, perhaps change color or otherwise indicate the finality of the option.

Idea (2) – warn if ending for this generation but still having ability to play standard projects, cards or card actions.

kberg:
My first instinct is that this is generally a safe thing to do, but it is also a little funny: should available actions count, even when some of the actions are often worthless? (eg 10MC to raise heat production 2 steps)? If the player can still sell patents, should they be warned? (I suppose they should.)

Just thinking this through.

waxoid:
My thought was cards you can both afford and meet requirements, card actions (you could afford), or basically if you have 11 M€ (standard projects) for something that isn't maxed (met terraform goals). And prompt with a warning icon or something.

But it's certainly a fair concern that something happening in some situations but not others could cause new (maybe more minor) confusion.

Another idea would be a game setting at setup 'Ask to confirm when passing' or more generically 'Ask to confirm important actions' if there are a number of these that could be controlled. Naturally a simpler coding option would be to just always prompt on pass – I think that would still have value and help in the scenario, given it would be a contrast with 'ending turn' (which wouldn't prompt).

kberg:
Could keep it simple and provide that warning as it's know to be your last
action of the game (modulo greeneries)

kberg:
From #5360 by @d-little (though summarized here)

Possible solutions include:
> tooltip message clarifing the action,
Mobile users don't have tooltips, but I'd be happy to find a way to incorporate tooltips into the server in general. :D

> Clearer messaging:
> * "Pass for the _remainder_ of this generation"
> * "Pass and end all actions for Generation #{{currentGeneration}}"
Similar to the wording in https://github.com/terraforming-mars/terraforming-mars/wiki/Variants#fast-mode
> A confirmation if the number of available actions > 0 (less trivial)

[...]

PR #5360, opened 2023-02-14
Make "Pass" option more visually distinct

bli17:
Make "Pass for this generation" option more visually distinct to reduce chance of confusion with "End turn" option.

### DIFF
--- a/src/server/Player.ts
+++ b/src/server/Player.ts
@@ -1461,7 +1461,7 @@ export class Player {
   }
 
   private passOption(): PlayerInput {
-    return new SelectOption('Pass for this generation', 'Pass', () => {
+    return new SelectOption('~~!! Pass for this generation !!~~', 'Pass', () => {
       this.pass();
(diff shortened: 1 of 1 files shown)


d-little:
This is related to #4846

I'm not sure this is a huge issue, but do feel that there's an approach here that's a better solution over unusual iconography.

I've:
* Tooltip message clarifing the action,
* Clearer messaging:
  * "Pass for *the remainder of* this generation"
  * "Pass and end all actions for Generation #{{currentGeneration}}"
  * Similar to the wording in https://github.com/terraforming-mars/terraforming-mars/wiki/Variants#fast-mode
* A confirmation if the number of available actions > 0 (less trivial)

Thankyou for prompting the change though :)

bli17:
For a separate private server, I changed the options to "Skip 2nd Action (End Turn)" and "PASS FOR THE GENERATION" (in all caps). It has been enough to avoid confusion. But that requires more work in translations, hence the iconography proposal.

PR #5368, opened 2023-02-15
Centralize the list of tags in the game.

kberg:
This required a little bit of organization in how clone tags were declared, and I added a nice test to go with it.

### DIFF
--- a/src/server/Game.ts
+++ b/src/server/Game.ts
@@ -68,6 +68,7 @@ import {TheNewSpaceRace} from './cards/pathfinders/TheNewSpaceRace';
 import {CorporationDeck, PreludeDeck, ProjectDeck, CeoDeck} from './cards/Deck';
 import {Logger} from './logs/Logger';
 import {addDays, dayStringToDays} from './database/utils.ts';
+import {ALL_TAGS, Tag} from '../common/cards/Tag';
 
 export interface Score {
   corporation: String;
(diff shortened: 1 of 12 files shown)


src/server/Game.ts
```diff
@@ -180,6 +183,14 @@ export class Game implements Logger {
       player.game = this;
       if (player.isCorporation(CardName.MONS_INSURANCE)) this.monsInsuranceOwner = player.id;
     });
+
+    this.tags = ALL_TAGS.filter((tag) => {
+      if (tag === Tag.VENUS) return gameOptions.venusNextExtension;
+      if (tag === Tag.MOON) return gameOptions.moonExpansion;
+      if (tag === Tag.MARS) return gameOptions.pathfindersExpansion;
+      if (tag === Tag.CLONE) return gameOptions.pathfindersExpansion;
+      return true;
```
d-little, line 192: Completely minor, but I recommend splicing Venus in between Jupiter and Plants tags.

There's a few cards that use the ALL_TAGs that display a prompt, it would be good to ensure that this prompt is consistent with the playerview tag bar:

image

d-little: @kberg

kberg: Ah, that ordering comes from someplace else:

image

d-little: Ah! OK. So, deal with sorting when we display the information

d-little (approved):

PR #5378, opened 2023-02-16
CEO : Rename Musk tests so they actually run

d-little:
Problem with Musk; the test name was wrong and I never caught it before being pushed.

There's also now a problem with Musk himself, so this test will fail.

He's not drawing the cards after discarding.

### DIFF
--- a/src/server/cards/ceos/Musk.ts
+++ b/src/server/cards/ceos/Musk.ts
@@ -44,12 +44,12 @@ export class Musk extends CeoCard {
       'Discard',
       eligibleCards,
       (cards) => {
+        player.addResource(Resources.TITANIUM, cards.length + 6, {log: true});
         for (const card of cards) {
           player.cardsInHand.splice(player.cardsInHand.indexOf(card), 1);
           game.projectDeck.discard(card);
         }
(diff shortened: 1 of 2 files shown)


kberg:
broken!

d-little:
So I couldnt figure out how to make Musk gain Titanium _after_ drawing cards, but, at least now the card works as _intended_, if not as _written_

src/server/cards/ceos/Musk.ts
```diff
@@ -44,12 +44,12 @@ export class Musk extends CeoCard {
       'Discard',
       eligibleCards,
       (cards) => {
+        player.addResource(Resources.TITANIUM, cards.length + 6, {log: true});
         for (const card of cards) {
           player.cardsInHand.splice(player.cardsInHand.indexOf(card), 1);
           game.projectDeck.discard(card);
         }
-        DrawCards.keepAll(player, cards.length, {tag: Tag.SPACE});
```
kberg, line 51: Oh yeah, DC.ka is a weird bit of code that could be cleaned made more obvious. (Adding .execute()) at the end would have done it.

kberg (approved):

PR #5384, opened 2023-02-16
PlayerHome refactor: main component

Hietamaki:


kberg:
I'm sorry if this wasn't clear. But in one commit, move a single component out, and remove the same component from PlayerHome and SpectatorHome.  Does that make sense?

d-little:
> 

This whole set of changes looks _really good_, but I agree it needs to be done piecemeal.  More hassle, but less chance things go kaboom

Hietamaki:
All I did in PlayerHome was to remove the code that was specific to each respective component I moved. I can appreciate that smaller PRs make the review process faster, but I'm sorry it's kind of pain in the butt to split it afterwards.

While not ideal, I'd argue the diff should be intelligible as is. For now I'm limited on time so I leave it up to your consideration whether you want to use the refactor or not.

kberg:
I understand if you can't spend the time.

PR #5393, opened 2023-02-18
Add 'ares' to list of expansions for card compatibility

d-little:
This is required for CEOs Caesar and Gaia, and anyone else in the future who needs to add ares compatibility :)

The two CEOs:
```ts
    // [CardName.CAESAR]: {Factory: Caesar, compatibility: 'ares'},
    // [CardName.GAIA]: {Factory: Gaia, compatibility: 'ares'},
```

PR #5394, opened 2023-02-18
CEO Caesar

d-little:
Requires #5393 

### Caesar

image

I'm not confident about the way the Tests were implemented. 

Required a new Hazard Tile asset, _but_, we're only ever placing EROSION_MILD.  I am very tempted to remove the new asset entirely and just use the EROSION_MILD tiles


src/server/cards/ceos/Caesar.ts
```diff
@@ -0,0 +1,61 @@
...
+      },
+    });
+  }
+
+  public override canAct(player: Player): boolean {
+    if (!super.canAct(player)) {
+      return false;
+    }
+    const game = player.game;
+    return game.board.getAvailableSpacesOnLand(player).length >= game.generation;
```
kberg, line 35: What a weird constraint. That's how it works, huh?

kberg: If you care, inline `game`.

src/server/cards/ceos/Caesar.ts
```diff
@@ -0,0 +1,61 @@
...
+    super({
+      name: CardName.CAESAR,
+      metadata: {
+        cardNumber: 'L33',
+        renderData: CardRenderer.builder((b) => {
+          b.opgArrow().text('X').hazardTile(1, {size: Size.LARGE}).br;
+          b.minus().production((pb) => pb.wild(1, {all})).asterix();
+          b.br;
+        }),
+        description: 'Once per game, place X hazard tiles. Each opponent loses 1 unit of production, or 2 units if there are 6 or more hazard tiles.',
```
kberg, line 25: A unit of production of their choice?

src/server/cards/ceos/Caesar.ts
```diff
@@ -0,0 +1,61 @@
...
+    for (let i = 0; i < game.generation; i++) {
+      game.defer(new PlaceHazardTile(player, TileType.EROSION_MILD));
+    }
+
+    const otherPlayers = game.getPlayers().filter((p) => p.id !== player.id);
+
+    game.defer(new SimpleDeferredAction(player, () => {
+      const hazardTileCount = game.board.spaces.filter((space) => space.tile && HAZARD_TILES.has(space.tile.tileType)).length;
+      otherPlayers.forEach((opponent) => {
+        if (hazardTileCount < 6) {
```
kberg, line 49:
````
const units =  hazardTimeCount < 6 ? 1 : 2;
game.defer(new SelectProductionToLoseDeferred(opponent, units));
```

kberg (approved):
Minor changes before flight.

Ceaser is a real jerk.

src/server/cards/ceos/Caesar.ts
d-little: Yeah, there's no 'up to' or 'may', you gotta place every tile.

src/server/cards/ceos/Caesar.ts
d-little: Makes sense.

Also, the original card had this:
`(Once per game, place X hazard tiles. Each opponent loses 1 unit of production, or 2 units if there are more than 5 hazard tiles.)`

I'm big on clear values.  'More than 5' doesnt make it clear that 6 is the magic number. At least, it doesn't to me!

kberg (approved):

PR #5399, opened 2023-02-18
CEO CreateGameForm :: select starting card count

d-little:
Allow players to choose how many CEOs they'd like to be dealt.

`startingCeos` has been added to all the GameOption locations where `startingCorporations` lives

My playgroup has been using 4 instead of the default 3, sometimes its more interesting.

Output display:

image

image

### DIFF
--- a/src/client/components/create/CreateGameForm.vue
+++ b/src/client/components/create/CreateGameForm.vue
@@ -139,6 +139,13 @@
                                 <div class="create-game-expansion-icon expansion-icon-ceo"></div>
                                 <span v-i18n>CEOs (BETA)</span>&nbsp;<a href="https://github.com/terraforming-mars/terraforming-mars/wiki/CEOs" class="tooltip" target="_blank">&#9432;</a>
                             </label>
+                            <template v-if="ceoExtension">
+                              <label for="startingCEONum-checkbox">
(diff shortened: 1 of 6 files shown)


d-little:
I forgot, I also modified 

`src/common/game/NewGameConfig.ts`'s `twoCorpsVariant: boolean;`

(it previously ended in a comma?  Nothing seemed to care, but updating it to be in line with the rest of the file)

kberg:
I'd argue this belongs in Options: 
image

d-little:
I dont hate moving it, but we probably need to move the two Moon options over.

I propose adding in the expansion icon though:
image

Just to make it a little more obvious that it's tied to CEOs.

If this looks good we can repeat for all of the other Game Options

kberg:
> I dont hate moving it, but we probably need to move the two Moon options over.
> 
> I propose adding in the expansion icon though: image
> 
> Just to make it a little more obvious that it's tied to CEOs.
> 
> If this looks good we can repeat for all of the other Game Options

Sure. And after thinking, I think your first option was more than acceptable.

kberg (approved):
There's a pending comment which I cannot find. But it'll appear momentarily, I guess.

PR #5400, opened 2023-02-18
Fix default card list view, show CEOs.

kberg:


d-little:
Im impressed you figured this out!  I had no idea where to start

PR #5401, opened 2023-02-18
CEO Gordon

d-little:
### Gordon

image

There probably needs to be discussions around how this plays with Ares and other special tiles.

Either we need to be strict in the definition, or work through special tiles.

If we rework Gordon from:
* 'Ignore placement restrictions for greenery and city tiles on Mars. Gain 2 M€ when you place a greenery or city tile on Mars.'
To:
* 'Ignore **normal** placement restrictions for greenery and city tiles on Mars. Gain 2 M€ when you place a greenery or city tile on Mars.',

We can say that special tile placement restrictions override even Gordon. If so, does Gordon override placement rules/get 2MC for:

|Tile|Placement Rules|2MC Bonus if the player places it?|
--|--|--
**Red City**|Gordon _can_ place next to Red City|Player **does** gain 2MC|
**Martian Natural Wonders**| Gordon can **not** place a City/Greenery on the cube'd space|N/A|
**Ocean Cities**|Gordon does not override placement behaviour; Ocean Cities are _not_ City Tiles.|Player **does** gain 2MC|
**Wetlands**|Gordon does not override placement behaviour; Wetlands is _not_ a Greenery Tile.|Player **does** gain 2MC|

(I'm sure there are more I'm missing)

### DIFF
--- a/src/server/cards/ceos/Gordon.ts
+++ b/src/server/cards/ceos/Gordon.ts
@@ -1,0 +1,43 @@
+import {CardName} from '../../../common/cards/CardName';
+import {Player} from '../../Player';
+import {CardRenderer} from '../render/CardRenderer';
+import {CeoCard} from './CeoCard';
+
+import {Board} from '../../boards/Board';
+import {ISpace} from '../../boards/ISpace';
+import {GainResources} from '../../deferredActions/GainResources';
+import {Resources} from '../../../common/Resources';
+import {SpaceType} from '../../../common/boards/SpaceType';
+import {BoardType} from '../../boards/BoardType';
+import {Phase} from '../../../common/Phase';
+
+export class Gordon extends CeoCard {
+  constructor() {
+    super({
+      name: CardName.GORDON,
+      metadata: {
+        cardNumber: 'L07',
+        renderData: CardRenderer.builder((b) => {
+          b.greenery().city().colon().megacredits(2).asterix();
+          b.br.br;
+        }),
+        description: 'Ignore placement restrictions for greenery and city tiles on Mars. Gain 2 M€ when you place a greenery or city tile on Mars.',
+      },
+    });
+  }
+
+  public override canAct(): boolean {
+    return false;
+  }
+
+  public onTilePlaced(cardOwner: Player, activePlayer: Player, space: ISpace, boardType: BoardType) {
+    if (cardOwner.id !== activePlayer.id) return;
+    if (boardType !== BoardType.MARS || space.spaceType !== SpaceType.LAND) return;
+    if (cardOwner.game.phase === Phase.SOLAR) return;
+
+    if (Board.isCitySpace(space) || Board.isGreenerySpace(space)) {
+      cardOwner.game.defer(new GainResources(cardOwner, Resources.MEGACREDITS, {count: 2}));
+    }
+    return;
+  }
+}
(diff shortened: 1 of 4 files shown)


src/server/cards/ceos/Gordon.ts
```diff
@@ -0,0 +1,43 @@
...
+    return false;
+  }
+
+  public onTilePlaced(cardOwner: Player, activePlayer: Player, space: ISpace, boardType: BoardType) {
+    if (cardOwner.id !== activePlayer.id) return;
+    if (boardType !== BoardType.MARS || space.spaceType !== SpaceType.LAND) return;
+    if (cardOwner.game.phase === Phase.SOLAR) return;
+
+    if (Board.isCitySpace(space) || Board.isGreenerySpace(space)) {
+      cardOwner.game.defer(new GainResources(cardOwner, Resources.MEGACREDITS, {count: 2}));
```
kberg, line 39: `{count: 2, log: true}`, I think?

tests/cards/ceo/Gordon.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+    game = newTestGame(2);
+    player = getTestPlayer(game, 0);
+    player2 = getTestPlayer(game, 1);
+
+    player.playedCards.push(card);
+  });
+
+  it('Can place greenery tile on any available land space, not just adjacent to exising greenery', function() {
+    addGreenery(player, '35');
+    expect(game.board.getAvailableSpacesForGreenery(player).length).greaterThan(6);
```
kberg, line 30: That's clever.

tests/cards/ceo/Gordon.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+  });
+
+  it('Can place cities next to other cities', function() {
+    const initialAvailableSpaces = game.board.getAvailableSpacesForCity(player).length;
+    addCityTile(player, '35');
+    expect(game.board.getAvailableSpacesForCity(player).length).eq(initialAvailableSpaces - 1);
```
kberg, line 36: This test is a little close to the edge. What about

```
const space = board.getSpace('35');
addCityTile(player, space.id);

const availableSpacesForCity = board.getAvailableSpacesForCity(player);
const spacesNextToCity = board.getAdjacentSpaces(space);
expect(intersection(spacesNextToCity, availableSpacesForCity).deep.eq(spacesNextToCity); // intersect preserves order of first element.
```

tests/cards/ceo/Gordon.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+    game.addTile(player, game.board.spaces.find((space) => space.spaceType === SpaceType.COLONY)!, {
+      tileType: TileType.CITY,
+    });
+    runAllActions(game);
+    expect(player.megaCredits).to.eq(0);
+  });
+
+  it('Does not gain MC when opponent places city or greenery tile', function() {
+    player.megaCredits = 0;
+    addGreenery(player2, '35');
```
kberg, line 59: Include `runAllActions`

tests/cards/ceo/Gordon.spec.ts
kberg: If this is messy and doesn't work after a few minutes. Let me know.

kberg (commented):
Nicely done.

I made a small edit to Board.ts that wasn't relevant, but I noticed it.

kberg (approved):

tests/cards/ceo/Gordon.spec.ts
d-little: I took an alternative approach, hopefully it's OK.  Just, place a city, see if I can place one directly adjacent to it ([0])

kberg: Yes,. thanks.

kberg (approved):

Issue #5406, opened 2023-02-20
[CEO] New Van Allen CEO from custom expansion can't claim milestones without $8MC


nabby101:
Sorry, forgot to take screenshots, but this is a pretty straightforward one, although I am unsure whether it's a bug or a slightly unclear rule.
I had $0MC and was unable to claim the Tycoon milestone despite having 15 green/blue cards.
My interpretation of Van Allen is that the milestone is free, not reimbursed, but it's possible that "you must still meet the requirements" applies to having the $8MC as well.

URL: https://terraforming-mars.herokuapp.com/spectator?id=sc03350ebb86e
Player color: orange
Step: 74
Version: bc70f33, built at Fri, 17 Feb 2023 05:58:58 GMT
Browser: Firefox 110

d-little:
Its a bug, good catch.  Thankyou for the report

d-little:
Resolved in #5408

PR #5408, opened 2023-02-20
Van Allen should claim for free

d-little:
Fix for #5406

Improved tests.

### DIFF
--- a/src/server/Player.ts
+++ b/src/server/Player.ts
@@ -1843,7 +1843,8 @@ export class Player {
       'Take your first action' : 'Take your next action';
     action.buttonLabel = 'Take action';
 
-    if (this.canAfford(MILESTONE_COST) && !this.game.allMilestonesClaimed()) {
+    // VanAllen can claim milestones for free:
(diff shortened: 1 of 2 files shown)


src/server/Player.ts
```diff
@@ -1843,7 +1843,8 @@ export class Player {
       'Take your first action' : 'Take your next action';
     action.buttonLabel = 'Take action';
 
-    if (this.canAfford(MILESTONE_COST) && !this.game.allMilestonesClaimed()) {
+    // VanAllen can claim milestones for free:
+    if ((this.canAfford(MILESTONE_COST) && !this.game.allMilestonesClaimed()) || this.cardIsInEffect(CardName.VANALLEN)) {
```
kberg, line 1847: Van Allen can claim milestones beyond the third one? That's how this reads, but it's not how Van Allen reads.

d-little: You are correct, good catch. Fixed up the order

kberg (approved):

PR #5412, opened 2023-02-20
Remove double extension

aloisdg:
Hi!

Just notice that a file have two extensions. Since I can't see why? I guess it is a bug. (If not close this MR).

This MR rename "server/database/utils.ts.ts" into "server/database/utils.ts".

Cheers,


kberg (approved):
Nice catch, thanks.

PR #5425, opened 2023-02-23
CEO Oscar

d-little:
### Oscar

image


src/server/cards/ceos/Oscar.ts
```diff
@@ -0,0 +1,57 @@
...
+          b.br.br;
+          b.opgArrow().chairman();
+        }),
+        description: 'You have +1 influence. Once per game, replace the Chairman with one of your delegates.',
+      },
+    });
+  }
+
+  public override play(player: Player) {
+    const turmoil = player.game.turmoil;
```
kberg, line 26: Is it possible to discard a CEO? If so, we'll need an onDiscoard.

src/server/cards/ceos/Oscar.ts
```diff
@@ -0,0 +1,57 @@
...
+      return false;
+    }
+    const turmoil = Turmoil.getTurmoil(player.game);
+    return turmoil.hasDelegatesInReserve(player.id) && turmoil.chairman !== player.id;
+  }
+
+  public action(player: Player): PlayerInput | undefined {
+    const turmoil = Turmoil.getTurmoil(player.game);
+    // This action does not grant TR, so we dont use turmoil.setNewChairman()
+    turmoil.delegateReserve.remove(player.id);
```
kberg, line 42: Should there be code which moves the chairman to the reserve?

kberg (commented):
Nice work.

src/server/cards/ceos/Oscar.ts
d-little: Yeah.  

Might I suggest adding options to turmoil.setNewChairman(), with a flag for gainTR (default true). That way we keep all of this code in one place?

src/server/cards/ceos/Oscar.ts
d-little: Lowell discards himself.  Not entirely sure what that relates to here though

src/server/cards/ceos/Oscar.ts
kberg: Please do. Feel free to do that in this PR or another, your call.

d-little:
Ready for review after the Turmoil turmoil resolved itself

kberg (approved):
Looks great.

PR #5426, opened 2023-02-24
Turmoil Bugfixes

d-little:
### 14 Neutrals

Per official rulebook.  

### Exiting Chairman Return to Reserve

Moved out of `setRulingParty()` and into `setNewChairman()`, so that Vote of No Confidence returns Neutral Chairman to Reserve. (and Oscar will too)

### Additional tests for Vote Of No Confidence

Just making sure Neutrals get their Chairman Delegate back to reserve

### DIFF
--- a/src/common/constants.ts
+++ b/src/common/constants.ts
@@ -30,7 +30,7 @@ export const TITANIUM_TRADE_COST = 3;
 
 // Turmoil
 export const DELEGATES_PER_PLAYER = 7;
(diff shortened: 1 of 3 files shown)


kberg (commented):
Given the screenshot I showed, it looks like there really is another delegate. I bet the initial chairman doesn't come from the reserve, but from thin air.

kberg (approved):
Nice! Thanks.

Issue #4929, opened 2022-09-13
[Community] Playwrights allows replaying cards placing special tiles which should not be allowed


HenrikSteenbock:
Events that place a tile (Deimos Down, Metallic Asteroid, etc.) can be re-played using the Playwrights corporation action. The special tile should only be allowed to be placed once, and playing it should be mandatory, so those cards should not be allowed to be re-played.

HenrikSteenbock:
I imagine it will be the same case for the corporation Odyssey (Pathfinder).

kberg:
Yes, it would be.

So, why shouldn't it be allowed? I see your point, but why not allow it in a web game?

Some references:

https://discord.com/channels/737945098695999559/737945098695999562/869601868056244286: 
>  Playwright allows double playing some special-event tiles. That is really not a problem  with web app and it is intuitively correct, but would obiviously be different in real life play since you don't have 2 copies of the tiles

https://discord.com/channels/737945098695999559/742721510376210583/843924586425090108
> apart from whether it's OP it's just not the way the card would work in a real life game, no one's printing out second copies of their tiles -- and note it would literally break the game if Mohole Area just happened to be an event

That second argument is the strongest for your point. Fair point about Mohole Area.

HenrikSteenbock:
I can only think of 2 events that place special tiles (Deimos Down and Metallic Asteroid). For Metallic Asteroid it wouldn't be an issue to have just another asteroid, but at least for Deimos Down it thematically doesn't make sense.

In general, while I absolutely see the advantages a web app has over the physical copy and I know we are outside of official territory anyway, being consistent with the physical copy seems more appropriate to me.

kberg:
Yeah. As you can see I'm on the fence here, and Discord no longer has the community to discuss it in detail. But I am starting to lean toward your perspective, except that the other perspective means no work for me. ;)

There's another option: play the card, deny the additional special tile. What do you think about that?

HenrikSteenbock:
Personally, I don't like the idea of playing the card without placing the tile because other cards placing tiles (city/special) MUST place that tile (except oceans) or may not be played. Having tile placement not mandatory only is inconsistent with the rules.
I'd rather play the event twice and "only" violate the thematic background. Afterall, it's only a niche case that isn't game breaking.

kevinb9n:
[...]

PR #5427, opened 2023-02-24
CEO Buck

d-little:
### Buck

image

@nwai90 added this recently, figured I'd bring it over almost entirely as-is because I _really_ like the idea :)

### DIFF
--- a/src/common/cards/CardName.ts
+++ b/src/common/cards/CardName.ts
@@ -757,6 +757,7 @@ export enum CardName {
     // CEOs
     ASIMOV = 'Asimov',
     BJORN = 'Bjorn',
+    BUCK = 'Buck',
     CLARKE = 'Clarke',
     DUNCAN = 'Duncan',
     ENDER = 'Ender',
--- a/src/common/cards/render/AltSecondaryTag.ts
+++ b/src/common/cards/render/AltSecondaryTag.ts
@@ -16,6 +16,9 @@ export enum AltSecondaryTag {
   NO_PLANETARY_TAG = 'no_planetary_tag',
   WILD_RESOURCE = 'wild-resource',
 
+  // CEOs:
+  // used in Buck CEO
+  GREEN = 'green',
   // used in Faraday CEO
   DIVERSE = 'diverse',
 }
--- a/src/server/behavior/Executor.ts
+++ b/src/server/behavior/Executor.ts
@@ -57,6 +57,13 @@ export class Executor implements BehaviorExecutor {
         if (player.game.board.getAvailableSpacesForType(player, behavior.city.on ?? 'city').length === 0) {
           return false;
         }
+      } else {
+        const space = player.game.board.getSpace(behavior.city.space);
+        if (space.player !== undefined) {
+          // The Space tile has already been played.
+          // Somehow (probably Buck) the Green Automation card has returned to a players hand.  We cannot allow this card to be replayed.
+          return false;
+        }
       }
     }
 
--- a/src/server/cards/base/NoctisCity.ts
+++ b/src/server/cards/base/NoctisCity.ts
@@ -34,7 +34,7 @@ export class NoctisCity extends Card implements IProjectCard {
   }
 
   public override bespokeCanPlay(player: Player): boolean {
-    if (player.game.board.getNoctisCitySpaceId !== undefined) {
+    if (player.game.board.getNoctisCitySpaceId() !== undefined) {
       return true;
     } else {
       return player.game.board.getAvailableSpacesForCity(player).length > 0;
--- a/src/server/cards/ceos/Buck.ts
+++ b/src/server/cards/ceos/Buck.ts
@@ -34,6 +34,7 @@ export class Buck extends CeoCard {
   }
 
   public action(player: Player): PlayerInput | undefined {
+    this.isDisabled = true;
     return new SelectCard(
       'Select card to take back into hand',
       'Select',
@@ -41,7 +42,6 @@ export class Buck extends CeoCard {
       (foundCards: Array<IProjectCard>) => {
         const selectedCard = player.playedCards.splice(player.playedCards.indexOf(foundCards[0]), 1)[0];
         player.cardsInHand.push(selectedCard);
-        this.isDisabled = true;
         return undefined;
       },
     );
--- a/src/server/cards/ceos/Buck.ts
+++ b/src/server/cards/ceos/Buck.ts
@@ -1,0 +1,62 @@
+import {CardName} from '../../../common/cards/CardName';
+import {Player} from '../../Player';
+import {PlayerInput} from '../../PlayerInput';
+import {CardRenderer} from '../render/CardRenderer';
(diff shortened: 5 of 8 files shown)


kberg:
Buck brings up an issue: https://github.com/terraforming-mars/terraforming-mars/issues/4929 which addresses the issue of making cards replayable if they place special tiles.

This includes Bio-Fertilizer Facility, Capital, Commercial District, Great Dam, Luna Mining Hub, Luna Train Station, Mare Imbrium Mine, Mare Nectaris Mine, Mare Serenatis Mine, Mining Area, Mohole Area, Natural Preserve, New Venice, Noctis City, Nuclear Zone, Ocean City, Ocean Farm, Solar Farm, and Wetlands.

There are at least four of those cards that are specifically tricky, that they place a tile in a very specific spot.

(Actually, this excludes all the tiles that place space colonies, like Stanford Torus, not to mention that Stanford Torus played twice, by the rules, is  kinda tricly.)

d-little:
> Buck brings up an issue: #4929 which addresses the issue of making cards replayable if they place special tiles.
> 
> This includes Bio-Fertilizer Facility, Capital, Commercial District, Great Dam, Luna Mining Hub, Luna Train Station, Mare Imbrium Mine, Mare Nectaris Mine, Mare Serenatis Mine, Mining Area, Mohole Area, Natural Preserve, New Venice, Noctis City, Nuclear Zone, Ocean City, Ocean Farm, Solar Farm, and Wetlands.
> 
> There are at least four of those cards that are specifically tricky, that they place a tile in a very specific spot.
> 
> (Actually, this excludes all the tiles that place space colonies, like Stanford Torus, not to mention that Stanford Torus played twice, by the rules, is kinda tricly.)

Maybe we need to expand the scope, exlude _any_ card that places a special tile?  Or even any tile?  Seems bad for cars that place regular forests and oceans

@nwai90 any thoughts?

nwai90:
> > Buck brings up an issue: #4929 which addresses the issue of making cards replayable if they place special tiles.
> > This includes Bio-Fertilizer Facility, Capital, Commercial District, Great Dam, Luna Mining Hub, Luna Train Station, Mare Imbrium Mine, Mare Nectaris Mine, Mare Serenatis Mine, Mining Area, Mohole Area, Natural Preserve, New Venice, Noctis City, Nuclear Zone, Ocean City, Ocean Farm, Solar Farm, and Wetlands.
> > There are at least four of those cards that are specifically tricky, that they place a tile in a very specific spot.
> > (Actually, this excludes all the tiles that place space colonies, like Stanford Torus, not to mention that Stanford Torus played twice, by the rules, is kinda tricly.)
> 
> Maybe we need to expand the scope, exlude _any_ card that places a special tile? Or even any tile? Seems bad for cars that place regular forests and oceans
> 
> @nwai90 any thoughts?

Perhaps it needs to expand to exclude cards that place tiles on reserved areas (e.g. space cities), and cards that place greenery tiles on ocean spaces

kberg:
Sorry, if I wasn't clear, there's already a ruling: #4929 was resolved with the decision that cards that place special tiles _can_ be replayed, and I'd like this to be consistent. The issue was resolved with a comment in the card description that makes this explicit.

image

And I think these should be the same. However, this makes it complicated for Noctis City, Stanford Torus, et cetera. And now that I've had a minute to consider it, here's what I think about those:

If you're playing with CEOs, then you're not a beginner. If you pull Noctis City back it's because you know what you are doing. Noctis City should not be replayable since you can't do everything on the card, and so the only think you can do with it is sell the patent. *But* it also  means that all those cards  (the ones above, and also Stratopolis, Ganymede Colony, etc) need to be changed so their canPlay doesn't let them get played if the tile is already there.

But it's not like this kind of problem is isolated to this condition. After all, it's a waste to pull back Archebacteria if the temperature is too high:

image

And just to think about the future: if you have two CEOs, Faraday is also not prepared for this.

d-little:
> Noctis City 

I'm pretty sure if the Noctis City space is is not available the game lets you place a city anywhere.

Looking at the code, this seems to be the way it's implemented:
```ts
  public override bespokeCanPlay(player: Player): boolean {
    if (player.game.board.getNoctisCitySpaceId !== undefined) {
      return true;
    } else {
      return player.game.board.getAvailableSpacesForCity(player).length > 0;
    }
  }
  public override bespokePlay(player: Player) {
    const noctisCitySpaceId = player.game.board.getNoctisCitySpaceId();
    if (noctisCitySpaceId !== undefined) {
      const space = player.game.board.getSpace(noctisCitySpaceId);
      player.game.addCityTile(player, space);
      return undefined;
    }
    return new SelectSpace('Select space for Noctis city', player.game.board.getAvailableSpacesForCity(player), (space: ISpace) => {
      player.game.addCityTile(player, space);
      return undefined;
    });
  }

```

`Stratopolis` does not behave this way, it's always assumed the Space tile is available.  Is there a possibility for a blanket ruling?  If a card would place a city tile on a Space location, but that location is not available (missing, occupied, or any other reason), instead you may place the city per normal city placement rules?

kberg:
> > Noctis City
> 
> I'm pretty sure if the Noctis City space is is not available the game lets you place a city anywhere.

Only on maps that don't have a Noctis City space on their maps.

`getNoctisCitySpaceId()` returns the space ID for the map. Tharsis has one. Hellas does not. The rules for Hellas are that NC allows you to place a city anywhere. So, playing Noctis City on Tharsis a second time will still try to place the tile on the reserved space. Which it cannot because a city is already there.

> Looking at the code, this seems to be the way it's implemented:

By the way, small bug. Hah!

> 
> ```ts
>   public override bespokeCanPlay(player: Player): boolean {
>     if (player.game.board.getNoctisCitySpaceId !== undefined) {
      ^--- bug. This should be getNoctisCitySpaceId().
>       return true;
>     } else {
>       return player.game.board.getAvailableSpacesForCity(player).length > 0;
>     }
>   }
>   public override bespokePlay(player: Player) {
>     const noctisCitySpaceId = player.game.board.getNoctisCitySpaceId();
>     if (noctisCitySpaceId !== undefined) {
>       const space = player.game.board.getSpace(noctisCitySpaceId);
>       player.game.addCityTile(player, space);
>       return undefined;
>     }
>     return new SelectSpace('Select space for Noctis city', player.game.board.getAvailableSpacesForCity(player), (space: ISpace) => {
>       player.game.addCityTile(player, space);
>       return undefined;
>     });
>   }
> ```
> 
> `Stratopolis` does not behave this way, it's always assumed the Space tile is available. Is there a possibility for a blanket ruling? If a card would place a city tile on a Space location, but that location is not available (missing, occupied, or any other reason), instead you may place the city per normal city placement rules?

Replaying Stanford Torus means getting a city on the map for 12MC. Dawn City gives you a city + Titanium production for 15. Great for greenery bonuses. Also, I don't think it makes sense for a space tag card to be used to place a city on the surface, paying with Titanium.

kberg:
What's the current status?

d-little:
Closing this, will likely reopen when I revisit Buck

PR #5436, opened 2023-02-28
Ingrid bugfixes for #5434

### LINKED ISSUE #5434: [CEO] Ingrid tile placement bonus doesn't include ocean tiles (possibly intended)
"Once per game, activate this ability: when you take an action that places a tile on Mars THIS GENERATION, draw a card."

Activating the once-per-game ability of the CEO Ingrid does not draw a card when you place oceans on Mars during the turn.

I'm not sure if this is a bug or intended gameplay because as far as I can recall, the mechanic doesn't exist in any of the real expansions. Looking at the card list, apparently there are a few custom cards that have a similar effect (the corporation Curiosity II and the blue card Geological Expedition), but I haven't played with the fan-made expansions much, so I'm not sure if this behaviour is consistent with those cards as far as counting an ocean as "placing a tile on Mars."

My interpretation of this ability as worded is that it should include ocean tiles, otherwise I would expect it to follow the wording convention on cards like Land Claim ("non-reserved area").

URL: https://terraforming-mars.herokuapp.com/spectator?id=s6f969a4a6c32
Player color: orange
Step: 22
Version: 7441c56, built at Mon, 20 Feb 2023 03:35:06 GMT
Browser: Firefox 110

d-little:
### Bugfix

Should resolve  #5434

### DIFF
--- a/src/server/cards/ceos/Ingrid.ts
+++ b/src/server/cards/ceos/Ingrid.ts
@@ -36,7 +36,8 @@ export class Ingrid extends CeoCard {
 
   public onTilePlaced(cardOwner: Player, activePlayer: Player, space: ISpace, boardType: BoardType) {
     if (this.opgActionIsActive === false) return;
-    if (boardType !== BoardType.MARS || space.spaceType !== SpaceType.LAND) return;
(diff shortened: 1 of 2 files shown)


kberg (approved):
Nice fix.

PR #5437, opened 2023-02-28
Simplify CardRequirement a little bit.

kberg:


d-little (approved):
Makes sense, but also makes me wonder why it was passing everything by reference prior?

Just a design decision?

kberg:
> Makes sense, but also makes me wonder why it was passing everything by reference prior?
> 
> Just a design decision?

Probably just happened organically.

Issue #47, opened 2019-12-08
Frustrating solo game final


alrusdi:
Just finished solo game successfully and got no reward :-)
Can we do special win/lose page for solo mode?
If player wins we can show him image like this https://upload.wikimedia.org/wikipedia/commons/thumb/6/62/Terraformed_Mars.jpg/800px-Terraformed_Mars.jpg
and suggest to make another attempt to try get more Victory Points.
If player loses suggest him to try again with some hints from BGG and rules.

bafolts:
A separate page would probably be better. What is currently there was the fastest to implement as I was rushing to get this working.

alrusdi:
Merged

PR #5438, opened 2023-03-01
Promo #47 - Robotic Pollinators

d-little:
### Robotic Pollinators

... not a particularly good card, but cute lil robees

image

Source: https://i.redd.it/70vv0axhf2la1.jpg

### DIFF
--- a/src/server/cards/promo/RobotPollinators.ts
+++ b/src/server/cards/promo/RobotPollinators.ts
@@ -5,8 +5,6 @@ import {CardType} from '../../../common/cards/CardType';
 import {CardRenderer} from '../render/CardRenderer';
 import {played} from '../Options';
 import {CardRequirements} from '../CardRequirements';
-import {Player} from '../../Player';
-import {Resources} from '../../../common/Resources';
 import {Tag} from '../../../common/cards/Tag';
 
 export class RobotPollinators extends Card implements IProjectCard {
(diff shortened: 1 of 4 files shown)


src/server/cards/promo/RobotPollinators.ts
```diff
@@ -0,0 +1,38 @@
...
+export class RobotPollinators extends Card implements IProjectCard {
+  constructor() {
+    super({
+      cardType: CardType.AUTOMATED,
+      name: CardName.ROBOT_POLLINATORS,
+      cost: 9,
+      requirements: CardRequirements.builder((b) => b.oxygen(4)),
+      behavior: {
+        production: {plants: 1},
+        stock: {titanium: 3},
```
kberg, line 21: Gain 3 titanium?

You can instead do `stock: {plants: {tag: Tag.PLANT}` and remove bespokePlay.

kberg (approved):
One change. Well, two changes?

src/server/cards/promo/RobotPollinators.ts
d-little: Whoops!  I'll blame being distracted from real actual work

d-little: .. I also really should start doing deeper sanity checks, for all production/resources, not just what I'm looking for.  Dont want to end up like Jurassic Park

PR #5452, opened 2023-03-06
Fix: CEO Floyd OPG Discounts

d-little:
May need to update tests too

Should resolve:

> Floyd CEO bug - would not allow me to use the discount against a card that cost more than the discount to be given.  Generation 3, should be a discount of 13 + 6, would not let me play Phobos Space Haven when I had 9 MC in hand and 3 titanium


kberg:
```
  1) Floyd
       Takes action:

      AssertionError: expected { Object (insertId, queue) } to have a length of 2 but got 1
      + expected - actual

      -1
      +2
      
      at Context.<anonymous> (tests/cards/ceo/Floyd.spec.ts:29:38)
```

kberg (approved):

PR #5462, opened 2023-03-09
CEO Gaia

d-little:
### Gaia

I'm almost certain I've not implemented this correctly. The original had a lot of custom work, I figured I could just use the Ares' own `earnAdjacencyBonuses`... I hope I was right

image


src/server/cards/ceos/Gaia.ts
```diff
@@ -0,0 +1,40 @@
...
+    });
+  }
+
+  public action(player: Player): PlayerInput | undefined {
+    this.isDisabled = true;
+    const board = player.game.board;
+    // For every tile placed on the board, grant all the adjacency bonuses for that tile.
+    // Owners and types of the tiles do not matter.  All the space needs a tile, including Ocean Tiles.
+    const tilesOnMars = board.spaces.filter((space) =>
+      space.tile?.tileType !== undefined && space.spaceType !== SpaceType.COLONY,
```
kberg, line 31: I'm not positive, but I suspect it would do well to add `space.player !== undefined` and `!isHazardType(space.tile)`. What do you think?

src/server/cards/ceos/Gaia.ts
```diff
@@ -0,0 +1,40 @@
...
+    );
+    tilesOnMars.forEach((space) => {
+      AresHandler.ifAres(player.game, (aresData) => {
+        AresHandler.earnAdjacencyBonuses(aresData, player, space);
```
kberg, line 35: This might improperly count the Ares Networker milestone. Consider adding an optional parameter to `earnAdjacencyBonuses` that doesn't track the Networker.

kberg (commented):
Seems pretty reasonable. A couple of minor changes?

src/server/cards/ceos/Gaia.ts
d-little: Wouldnt `space.player !== undefined` remove the bonuses from Ocean tiles?

Hazard tiles is curious.  Gaia does say _all_ tiles... I'm not sure what the design decision there would be. @nwai90 ?

src/server/cards/ceos/Gaia.ts
kberg: ping!

src/server/ares/AresHandler.ts
```diff
@@ -33,15 +33,12 @@ export class AresHandler {
     }
   }
 
-  public static earnAdjacencyBonuses(aresData: AresData, player: Player, space: ISpace) {
+  public static earnAdjacencyBonuses(aresData: AresData, player: Player, space: ISpace, isTileBeingPlaced = true) {
```
kberg, line 36: This is fine, but I think the better option is to split out the milestone stuff into its own function. I can do that post-merge.

src/server/cards/ceos/Gaia.ts
kberg: Wow, Oceans. OK? /shrug

kberg (approved):

kberg:
Nice work.

src/server/cards/ceos/Gaia.ts
kberg: From @nwai90:
image

kberg (changes requested):
(this is a vestigial comment to change the review state back to 'request changes')

src/server/cards/ceos/Gaia.ts
d-little: I'd have to go back to the Ares rules, but do Ocean tiles not grant adjacency bonuses on placement?  I was pretty sure they did

kberg: They do, but apparently according to @nwai90 they don't apply. /shrug?

nwai90: Ocean tiles do grant adjacency bonuses on initial placement. The reason they were excluded initially is for balance reasons. The effect is potentially a lot stronger if oceans were also included, especially because it will apply to adjacencies that would not have been originally granted (i.e. placing an Ares tile where there are already adjacent tiles, suddenly those neighbor tiles will apply for this effect)

d-little: > The effect is potentially a lot stronger if oceans were also included, especially because it will apply to adjacencies that would not have been originally granted (i.e. placing an Ares tile where there are already adjacent tiles, suddenly those neighbor tiles will apply for this effect)

That would apply here anyway, right?  How does excluding Oceans change that behaviour?

I'm happy to update with **Once per game, gain the Ares adjacency bonuses of all non-ocean tiles placed on Mars.',**

src/server/cards/ceos/Gaia.ts
```diff
@@ -28,7 +28,7 @@ export class Gaia extends CeoCard {
     // For every tile placed on the board, grant all the adjacency bonuses for that tile.
     // Owners and types of the tiles do not matter.  All the space needs a tile, including Ocean Tiles.
     const tilesOnMars = board.spaces.filter((space) =>
-      space.tile?.tileType !== undefined && space.spaceType !== SpaceType.COLONY,
+      space.tile?.tileType !== undefined && space.spaceType !== SpaceType.OCEAN && space.spaceType !== SpaceType.COLONY,
```
kberg, line 31: spaceType !== ocean will omit counting Mangrove. You probably want space.tile.tileType.

image

This will also count hazard tiles. Consider

space.tile !== undefined && (space.tile.tileType  !== TileType.OCEAN && !isHazardTile(tile.tileType)) && space.spaceType !== SpaceType.COLONY.

d-little: It grants Ares tiles too, I think?

@nwai90 Maybe Gaia can be updated to only grant adjacency bonuses to tiles with 'owners', those with marker-cubes on them.  That would remove the Oceans and Ares tiles, but keep all other tiles in play?

kberg (approved):

PR #5464, opened 2023-03-10
Ares - Great Dam Tile

d-little:
### Great Dam (Ares)

A few tiles are missing Ares replacements as the extension came before said cards was released.  Great Dam in particular is missing.

I figured I'd have a go at converting one of them.  Final product:
image

Taken from this fanmade promo:

image

And the three Great Dams side-by-side:
image


Borbarad2:
Wow, finally. Thank you very much. A long request feature become true.

kberg:
OK, so, nice work, but the app doesn't quite fit this well. See this is where things get a little slippery. Take a look at `cardsToRemove` in the manifests. `PromoCardManifest` removes `GreatDam`. `AresCardManifest` needs to remove both `GreatDam` and `GreatDamPromo`.

Which also leads to this next question: what happens in a game with Ares and without Promo? In that case, there shouldn't even be a Great Dam ares version, and without a more sophisticated mechanism for `cardsToRemove` you'll get a game that doesn't have Great Dam at all.

I'm afraid it's just not this simple.

d-little:
> 

As discussed, Im proposing that we just include the Great Dam tile as a 'standard' part of Ares.  That way we just sweep under the rug any issue of non-Promo Great Dam + Ares being in the same game  :)

I'll admit I totally forgot that some people might not play with Promo all the time... it's a core expansion in my head!

I'll also be submitting PRs for the other missing tiles, Deimos Down and.. the one I cannot remember the name of.  If they have similar problems, well.. we'll see

tests/cards/ares/GreatDamAres.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  it('Requirements', function() {
+    maxOutOceans(player, 4);
+    card.play(player);
+    runAllActions(game);
+    const action = cast(player.popWaitingFor(), SelectSpace);
+    const space = action.availableSpaces[0];
+    action.cb(space);
+
+    expect(space.tile && space.tile.tileType).to.eq(TileType.GREAT_DAM);
```
kberg, line 42: `expect(space.tile?.tileType)`

kberg (approved):

PR #5475, opened 2023-03-15
Show All Corporation Names (Merger etc)

d-little:
Display all corporation names under the players name on both player screen and endgame.

This kinda breaks the placement of the First Player token, but I think the utility outweighs the issue.

PlayerView:
image

Endgame:
image

### DIFF
--- a/src/client/components/GameEnd.vue
+++ b/src/client/components/GameEnd.vue
@@ -283,7 +283,7 @@ export default Vue.extend({
       const corporationCards = cards
         .filter((card) => getCard(card.name)?.type === CardType.CORPORATION)
         .map((card) => card.name);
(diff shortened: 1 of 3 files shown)


kberg:
A rough guess suggests that it's possible to address this by moving  the first player icon into PlayerStatus and changing `margin-top` to 8px but I haven't checked it fully.

image
image
image

kberg:
> A rough guess suggests that it's possible to address this by moving the first player icon into PlayerStatus and changing `margin-top` to 8px but I haven't checked it fully.
> 
> image image image

Do you need some help with this?

d-little:
I had to put it in another div to get it to be in the right place:

image

Issue #5480, opened 2023-03-17
Special Design proxy card should be excluded from the Community deck?


puuj:
We started a game (on a personal server) that used the Community expansion. One of the cards dealt during the initial draft was the Special Design proxy (from commit #5445 ). The behavior was that players would see one fewer card in their draft selection, as the proxy card does not render. I'm not sure what the expected behavior is, but it seems that there should be some sort of logic to exclude Special Design proxy from the deck and then sub it in during the Playwrights action? I manually modified the game state to replace it with Special Design to prevent us from having to restart.

kberg:
Yes, I need help fixing this, and I would love a copy of your game.

Is it on Heroku? Is it SQLite or Postgres?

puuj:
Sent on Discord, but since I have a bunch of other stuff in my build it might not work automatically. I think the root cause is that if it's in the manifest, it ends up in the deck. Not sure if the filter or replace functionality is the right way to approach this, or there should be a class of cards that are known but not included in the Deck. This would be helpful for some other fun ideas like creating a "Happy Birthday" card that gets inserted into a player's hand on their birthday but otherwise does not exist in the deck.

kberg:
Thanks for this. Fixed!

Issue #5463, opened 2023-03-09
Missing card


theeagle1968:
When playing a game today, we were on the 4th generation draft. One of our players in the initial cards was only dealt 4 cards. In an effort to keep the game moving, he picked a card, and passed the rest on. The missing card continued through all players, aka, player he passed to (me) received 2 instead of 3, I passed 1 instead of two, and the 4th player passed to received nothing. 

URL: https://terraforming-mars.herokuapp.com/spectator?id=s522ee8b8ba1c#cards
Player color: green
Step: 103
Version: 382f3f4, built at Sun, 05 Mar 2023 16:07:45 GMT
Browser: Chrome 110

PR #5481, opened 2023-03-17
Do not instantiate proxy cards as part of the game cards.

kberg:


kberg:
Fixes #5463
Fixes #5480

Issue #5304, opened 2023-02-08
Urbanist award might not take all city VP accordingly


kberg:
Source: https://discord.com/channels/737945098695999559/742721510376210583/1072721212049993808

In short, Urbanist award rewards "Most city tile adjacencies on Mars".

![image](https://user-images.githubusercontent.com/413481/217432632-7020c023-f1d4-4813-b7f2-5585eb5af491.png)

Does that include Commercial District? Does it include Capital City? Does Red City count? Right now, none of them do. It only counts greeneries next to cities.

Borbarad2:
To give an answer from my point of view: Capitel City( Greeneries and ocean) and Red City(empty spaces) should be counted for that award, as they are cities in respect to that award. Comercial District should not count as it is not a city. It is getting VPs from Cities, but this is not part of that award.

PR #5514, opened 2023-03-27
Make Urbanist support Capital and Red City.

kberg:
@nwai90 just getting started on this. Does this seem reasonable? I'll write up a test.


nwai90:
@kberg Looks good! I did it differently here: https://github.com/nwai90/terraforming-mars/commit/599955c4f73fe70cbbff4600d3669a1ac04ca464 but the outcome is the same

kberg:
Fixed #5304

Issue #5302, opened 2023-02-08
Possible Economic Help regression?


kberg:
Unconfirmed report from Discord:
> Economic Help does not correctly raise a planetary influence track when the relevant (lowest) non-completed track is higher than any other already completed track. Example: A non-completed Moon track will not be raised if it is higher than a completed Jovian track. If this is intended behavior the wording of the project is quite misleading.

PR #5516, opened 2023-03-28
Economic Help was not correctly identifying completed tracks.

kberg:


kberg:
Fixed #5516

kberg:
Fixed #5302

PR #5525, opened 2023-03-31
Pathfinder Corp - Mars Maths

d-little:
### Mars Maths

Oh Boy!

image
image

I took the path of least resistance and did not change the behaviour of Luna Project Office.  Currently, if you're Mars Maths, and you play Luna Project Office, you'll still only draw 5 cards during draft but will, for two generations, be allowed to buy 5 cards


src/server/Player.ts
```diff
@@ -999,14 +1006,23 @@ export class Player {
...
   public runResearchPhase(draftVariant: boolean): void {
     let dealtCards: Array<IProjectCard> = [];
-    if (!draftVariant) {
-      this.dealForDraft(LunaProjectOffice.isActive(this) ? 5 : 4, dealtCards);
-    } else {
+    let cardsToKeep = 4;
+    if (draftVariant) {
       dealtCards = this.draftedCards;
       this.draftedCards = [];
+    } else {
```
kberg, line 1013: Reminder to review this more carefully.

src/server/Player.ts
```diff
@@ -1447,7 +1463,7 @@ export class Player {
 
   private endTurnOption(): PlayerInput {
     return new SelectOption('End Turn', 'End', () => {
-      this.actionsTakenThisRound = 1; // Why is this statement necessary?
+      this.actionsTakenThisRound = 1;
```
kberg, line 1466: Did you remove the comment because you understand why the statement is necssary?

src/server/deferredActions/DrawCards.ts
```diff
@@ -86,14 +86,25 @@ export class DrawCards<T extends undefined | SelectCard<IProjectCard>> extends D
...
     if (options.paying) {
       const spendableMegacredits = player.spendableMegacredits();
       const affordableCards = Math.floor(spendableMegacredits / player.cardCost);
       max = Math.min(max, affordableCards);
+      if (max === 0) {
+        msg = 'You cannot afford any cards';
+      } else if (max < cards.length) {
+        // We're being offered more cards than we're able to buy
+        // So we should be specific on maximum number of cards
+        msg = `Select up to ${max} card(s) to buy`;
```
kberg, line 99: Changes like this have to be coordinated with any translations.

kberg: image
This one looks wrong, but a couple others seem to match, though probably because of other cards. Kind keeping "keep" for now?

d-little: > Changes like this have to be coordinated with any translations.

The numbers here are going to be very dynamic.  I'm really hands-off on the Translation stuff, if there's a better way of putting in logs like these I'm happy to adjust.

> image This one looks wrong, but a couple others seem to match, though probably because of other cards. Kind keeping "keep" for now?

Keep is only used when `options.paying` != true.  In this instance you need to buy the cards, and you do not need to buy ${max} cards.  So I think it is important to keep them distinct

kberg: Ack. This is fine for now. In the future, this is better for translation:

`log('Select ${0} card(s) to keep', (b) => b.number(max));`

kberg (approved):

PR #5530, opened 2023-04-01
Ares deimos down

d-little:
### Deimos Down for Ares

image


d-little:
I'm not 100% on a few things, but firstly:

* I've had to add a new space-bonus type, Asteroid.
  * This doesnt seem to be a problem in testing Deimos Down, but I'm not sure if adding a new element to that array in `src/common/boards/SpaceBonus.ts` will affect anything

kberg (approved):

PR #5561, opened 2023-04-05
WIP: disallow loading and saving purged games.

kberg:
This change removes a case where games can continue to exist after their base records are purged. This is done by starting to add a layer that hides the database and the GameLoader together.

Tests needed.


kberg:
Resolved in another PR.

PR #5573, opened 2023-04-08
Make sure Van Allen is in the game before actually checking the status

d-little:
So it turns out while I _did_ run manual tests, and tried to include a lot of tests, I never actually made sure that this works when players arent using CEOs at all...

Huge mistake, will find out how to prevent it.

I'm not even 100% sure that this will resolve the issue, I ran a lot of manual tests and added some custom automated tests but am unsure of the coverage, since it wasnt caught beforehand!


d-little:
Closed in preference to #5574

Issue #5572, opened 2023-04-07
"No Player has played Van Allen" when trying to claim Specailist


nimbusscale:
When trying to claim the Specialist milestone there was a pop-up saying "No Player has played Van Allen". After clicking the link nothing happened and I tried to claim the milestone again with the same results. I then played another card and ended my turn. After I ended my turn I saw that the milestone was claimed. It didn't take the 8 MC for purchasing the milestone. The claim milestone action, not the second action taken did not show up in log. And looking in the log looks of a bit as off in general. Some log entries from Generation 1 shows up in Generation 2. 

It looks like Van Allen is related to the CEO expansion, but that expansion was not selected.We abandoned the game after that.

See screenshots here: https://imgur.com/a/KW47OmY

Player game link: https://terraforming-mars.herokuapp.com/player?id=p5d2f059b24bc

pavel-saladukha:
Same happened here - https://terraforming-mars.herokuapp.com/the-end?id=s82f2c4942c42

nimbusscale:
Playing another game and we ran into the same situation with the Planner, so it's not specific to just the Specialist milestone.

JacobORiordanRoss:
Similar situation here: https://terraforming-mars.herokuapp.com/the-end?id=p9e96d1b9eb0c

lizthegrey:
The problem is here:

https://github.com/terraforming-mars/terraforming-mars/blob/aeb048caba64cc038725848bfbaa4f9e36f0189f/src/server/Player.ts#L1399-L1400

`getCardPlayer()` is assumed to return `undefined` if the card has not been played, but instead it throws the exception if the card is not being played; either the callsite needs a catch() around it or the function needs to return undefined if the card has not been played.

kberg:
Thanks for all the reports. A fix went out a few days ago, with a second update today that fixed existing games.

PR #5574, opened 2023-04-08
Fix Van Allen's CEO by making getCardPlayer explicit.

kberg:
h/t @d-little

### DIFF
--- a/src/server/Game.ts
+++ b/src/server/Game.ts
@@ -1454,16 +1454,29 @@ export class Game implements Logger {
     return ret;
   }
 
-  public getCardPlayer(name: CardName): Player {
+  /**
+   * Returns the Player holding this card, or throws.
+   */
+  public getCardPlayerOrThrow(name: CardName): Player {
+    const player = this.getCardPlayerOrUndefined(name);
+    if (player === undefined) {
+      throw new Error(`No player has played ${name}`);
+    }
+    return player;
+  }
+
+  /**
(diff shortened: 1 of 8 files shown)


kberg:
Fixes #5572

src/server/Game.ts
```diff
@@ -1454,16 +1454,29 @@ export class Game implements Logger {
     return ret;
   }
 
-  public getCardPlayer(name: CardName): Player {
+  /**
+   * Returns the Player holding this card, or throws.
+   */
+  public getCardPlayerOrThrow(name: CardName): Player {
+    const player = this.getCardPlayerOrUndefined(name);
```
lizthegrey, line 1461: avoid duplicating the code by calling getCardPlayerOrUndefined, doing the undefined check, and throwing if undefined.

kberg: How is that different from what's in the PR? What am I missing?

kberg (commented):
Thanks for the drive-by

src/server/Game.ts
lizthegrey: I, uh, didn't have my coffee yet, that's exactly what you did. *facepalm*

lizthegrey (approved):

src/server/Game.ts
kberg: Thanks for the extra confirmation! :coffee:  :clinking_glasses:

d-little (approved):

Issue #4204, opened 2022-03-08
Some toggles aren't sticky


kberg:
> [](https://cdn.discordapp.com/attachments/742721510376210583/950885495636905994/unknown.png)Bug report about the Toggle card filter: The setting you are set are non persistent, i.e. after the website is refreshed or the next action took place, you lose your filter setting, which you choose. So the Toggle card filter function is not useful at all. But it would be great to have that function functional.
![image](https://user-images.githubusercontent.com/413481/157343769-47737bb3-beb6-43e5-bfd2-23b606336ce6.png)

Borbarad2:
What is left to close this bug?

kberg:
Quite a lot, really. This is only implemented for awards and milestones, it seems.

PR #5585, opened 2023-04-11
Fix toggles not working over page refreshes (#4204)

velit:
Watch functions were not annotated with the state they were watching on. This fixes card filters not persisting over page refreshes


velit:
I added another commit that adds award and milestone collapse toggles to be persistent over page refreshes

src/client/components/PlayerHome.vue
```diff
@@ -337,13 +337,13 @@ export default Vue.extend({
     };
   },
   watch: {
-    hide_active_cards() {
+    showActiveCards: function hide_active_cards() {
```
kberg, line 340: Why is this `showActiveCards: function hide_active_cards()`? It's got two verbs that are the opposite of each other.

src/client/utils/PreferencesManager.ts
```diff
@@ -5,6 +5,8 @@ export type Preferences = {
   show_alerts: boolean,
   hide_hand: boolean,
   hide_awards_and_milestones: boolean,
+  collapse_milestones: boolean,
```
kberg, line 8: There's already an option to hide the milestones and awards. What do yout hink about reusing them?

kberg (commented):
This looks like a decent change.  `Milestones.spec.ts`, etc are probably pretty straightforward for adding some tests to confirm this behavior. Especially by injecting Preferences as a property.

src/client/components/Milestones.vue
```diff
@@ -48,8 +48,9 @@ export default Vue.extend({
     },
   },
   data() {
+    const preferences = getPreferences();
```
kberg, line 51: Instead of this, add a property: 

```
    preferences: {
      type: Object as () => Readonly<Preferences>,
      default: () => PreferencesManager.INSTANCE.values(),
    },
```

src/client/components/PlayerHome.vue
velit: You'd have to ask the previous developer who had implemented the infrastructure for this feature. Do you want me to refactor it? Both the state and the toggle function already existed I just fixed the watch section to follow the right state.

src/client/utils/PreferencesManager.ts
velit: This was the least destructive way to do it from a user perspective. Some people might enjoy disabling milestone / awards from being shown completely. The change I made just makes the collapsing persistent.

Ie. from this
image

To this
image

src/client/components/Awards.vue
```diff
@@ -67,14 +67,16 @@ export default Vue.extend({
     },
   },
   data() {
+    const preferences = getPreferences();
     return {
-      showAwards: true,
+      showAwards: !preferences.collapse_awards,
```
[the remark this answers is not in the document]
velit: I followed the existing convention of having hides / collapses in the preferences. Do you want me to switch it for these two?

src/client/components/Milestones.vue
velit: I followed the existing way it was done in PlayerHome.vue. Do you want me to refactor it there also?

src/client/components/PlayerHome.vue
velit: It is a function but it already existed, I only annotated the function with the correct state parameter so the watch function runs when the state parameter changes which it wasn't doing previously. Or that's how I understood it. I changed the function names to toggle_ instead of hide_.

src/client/components/Awards.vue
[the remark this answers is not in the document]
velit: I'd actually prefer to keep it collapse in the preferences because that describes the behaviour of collapsing the component as opposed to totally hiding/removing the component. I'll switch the component state itself to better represent what it is doing.

src/client/utils/PreferencesManager.ts
velit: I didn't add these to the explicit user shown settings in the page, so the vars here are just for the persistence on the page. Ie. they're only dev visible and allow the functionality for users to persist the award/milestone state if they have the feature enabled or if other users prefer to not have the awards/milestones not visible at all it's possible to do it

velit:
Added collapse persistence tests for awards. I didn't add tests for milestones because the spec file is in poor shape. There's a very simple mock and nothing else.

velit:
Improved the code based on suggestions and added tests

src/client/components/Awards.vue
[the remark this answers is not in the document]
kberg: I can't convince you?

src/client/components/PlayerHome.vue
kberg: This comment disappeared. I think because you force-pushed? (Force pushing isn't something to worry about, I'm going to squish everything into a single commit.)

What happens if this is changed from

`showActiveCards: function hide_active_cards() {` to `showActiveCards: function () {`

kberg (commented):
Thank you for the test! :100:  I hope they weren't to difficult to figure out.

src/client/utils/PreferencesManager.ts
```diff
@@ -30,6 +32,8 @@ const defaults: Preferences = {
 
   hide_hand: false,
   hide_awards_and_milestones: false,
+  collapse_milestones: false,
```
kberg, line 35: Sorry -- I'm still asking for these to be show instead of collapse

.gitignore
```diff
@@ -12,6 +12,7 @@ dist/
 *.swp
 .nyc_output/
 .vscode/
+.idea/
```
kberg, line 15: Thanks!

kberg (approved):
This is a good change, and thank you for the tests.

PR #5605, opened 2023-04-20
CEO Prelude : Co Leadership

d-little:
### Co Leadership

A prelude from the community server's CEO implementation

image


src/server/cards/ceos/CoLeadership.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  public override bespokePlay(player: Player) {
+    const game = player.game;
+    const cardsDrawn: Array<ICeoCard> = [
+      game.ceoDeck.draw(game),
+      game.ceoDeck.draw(game),
+      game.ceoDeck.draw(game),
+    ];
+
+    cardsDrawn.forEach((card) => {
```
kberg, line 31: Given play order, this CEO will be played before the one you already chose. Is that what players will want? I think it won't be what they expect.

src/server/cards/ceos/CoLeadership.ts
```diff
@@ -0,0 +1,45 @@
...
+      if (card.canPlay?.(player) === false) {
+        cardsDrawn.splice(cardsDrawn.indexOf(card), 1);
```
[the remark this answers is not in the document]
d-little: inplaceRemove*

That wont be logged though, right? I'd prefer if it was announced.  Actually, should it be announced?  That's information given on a private draw

src/server/cards/ceos/CoLeadership.ts
d-little: I dont mind adding it to hand to be played with the other CEO.  Currently CEOs are just automatically played, because there's was no expectation that they'll interact with each other.

Xu changed that though, it has a Tag... so I probably need to update that code to have the player choosee

src/common/cards/CardName.ts
```diff
@@ -799,4 +798,6 @@ export enum CardName {
     RYU = 'Ryu',
     SHARA = 'Shara',
     XU = 'Xu',
+    // CEO Preludes
+    CO_LEADERSHIP = 'Co Leadership',
```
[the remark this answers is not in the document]
d-little: I'll take the thumbsup as agreement :)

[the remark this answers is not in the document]
kberg: image
It seems that the problem might be how it was copied over?

d-little:
@kberg ready for re-review.  I'd also like confirmation that the way I'm discarding CEOs is 'correct', if it is I'll roll it out to other instances of CEOs being discarded to ensure the discard deck is accurate

src/server/cards/ceos/CoLeadership.ts
```diff
@@ -0,0 +1,56 @@
...
+    });
+
+    if (cardsDrawn.length === 0) {
+      game.log('${0} drew no playable CEO cards', (b) => b.player(player));
+      return undefined;
+    }
+
+    return new SelectCard('Choose CEO card', 'Take', cardsDrawn, (([chosenCeo]) => {
+      // Discard unchosen CEOs
+      inplaceRemove(cardsDrawn, chosenCeo);
```
kberg, line 46: Instead of `inplaceRemove` you could just as well do this:

```
cardsDrawn.filter((c) => c !== chosenCeo).forEach((c) => game.ceoDeck.discard(c)));
```

The following might also work, but I wonder whether expressing merely the function loses whatever the correct definition of 'this' is at execution time.
```
cardsDrawn.filter((c) => c !== chosenCeo).forEach(game.ceoDeck.discard));
```

kberg (approved):
Since you asked about discarding.

src/server/cards/ceos/CoLeadership.ts
d-little: Thanks!  Second one failed, I like the first method though

kberg:
Well done.

PR #5609, opened 2023-04-21
Simplify Duncan VP counting.

kberg:
(Also simplify CEO card definition.)

To do this, though, means including another icon on the card. Which isn't so bad, I suppose?

image


kberg:
Some tests are failing. I'm figuring it out.

PR #5611, opened 2023-04-21
Standardize language on milestones.

kberg:


kberg:
image

kberg:
Standardizing on "at least" might also make sense,fwiw.

d-little:
If we're taking about standardizing these, I'd personally prefer: **Have _N_ or more _nouns_.** across them all.

Means changing the original wording but 'at least N' is kinda clunky

kberg:
> If we're taking about standardizing these, I'd personally prefer: **Have _N_ or more _nouns_.** across them all.
> 
> Means changing the original wording but 'at least N' is kinda clunky

I had some goals:
1. Standardize language
2. Remove needless words

I think if we're going to add text back, perhaps it's best to stick to that familiar refrain, applying it across the board. Does that seem like a reasonable compromise?

d-little:
> 

I'd like them all to be identical first and foremost, we're on the same page with that.  But new players _will_ ask if "Have 3 Jovian tags" means exactly 3, or if they can have more than 3.  I guess they ask that once and know, and it removes a _lot_ of words if we drop it.

I think we should remove all the 'at least' entries too and standardise which ones need 'in play', which need 'own'.

* **Tags**: "Have X {{tag}} tags".  Drop the 'at least' and 'in play' on the remaining items. eg: 'Have 8 building tags'
* **Resources**: "Have X {{Resource}}." Drop all other text. eg: "Have 20 heat."
* **Tiles**: "Own X {{tile}} tiles.". Drop everything else, eg: "Own 3 city tiles"

Is it a good/bad idea to move all of these strings into some kind of global? :) To compare them... I guess it only matters this once.

Examples of just the top row:

* **Terraformer**: Have a tertaform rating of 35 or 26 with Turmoil.
* **Mayor**: Own 3 city tiles
* **Gardener**: Own 3 greenery tiles
* **Planner**: Have 16 cards in your hand
* **Builder**: Have 8 building tags
* **Generalist**: Have increased all 6 productions by 1 step
* **Specalist**: Have 10 in production of any one resource
* **Ecologist**: Have 4 bio tags (eg)
* **Tycoon**: Have 15 project cards (blue and green cards)
* **Legend**: Have played 5 events

kberg:
Updated:

image

kberg:
(new prototype, not pushed.)

image

kberg:
Possible final proposal: 
image

PR #5617, opened 2023-04-25
CEO: Add Discards where necessary

### LINKED ISSUE #5546: CEO Lowell does not work if the CEO deck is empty/reshuffled during that action
Please include:
1. Lowell fizzles and does not count as an action.
2. Start a new game with 4 CEOs and 5 Players. Use Lowell (used gen 7, but not sure if gen matters).
3. image

URL: https://terraforming-mars.herokuapp.com/spectator?id=secd792a86185
Player color: yellow
Step: 383
Version: 648b01c, built at Sat, 01 Apr 2023 02:28:32 GMT
Browser: Chrome 111

### LINKED ISSUE #5607: CEOs are not being discarded
Related to #5546, but might become more of an issue with Prelude/Corps drawing CEOs.  They're not being discarded, but instead are just being cast into void after selection.

Expected behaviour is that they'd be put into the ceoDiscard pile, and reshuffled into draw if needed

d-little:
This should cover all our bases

Fixes #5607
Fixes #5546

Also standardised some of the wording/language on these items

### DIFF
--- a/src/server/cards/ceos/Lowell.ts
+++ b/src/server/cards/ceos/Lowell.ts
@@ -35,27 +35,33 @@ export class Lowell extends CeoCard {
   public action(player: Player): PlayerInput | undefined {
     this.isDisabled = true;
     const game = player.game;
-    const cardsDrawn: Array<ICeoCard> = [
+    const ceosDrawn: Array<ICeoCard> = [
       game.ceoDeck.draw(game),
       game.ceoDeck.draw(game),
       game.ceoDeck.draw(game),
     ];
 
-    cardsDrawn.forEach((card) => {
-      if (card.canPlay?.(player) === false) {
-        cardsDrawn.splice(cardsDrawn.indexOf(card), 1);
-        game.log('${0} was discarded as ${1} could not play it,', (b) => b.card(card).player(player), {reservedFor: player});
+    ceosDrawn.forEach((ceo) => {
+      if (ceo.canPlay?.(player) === false) {
+        game.ceoDeck.discard(ceo);
+        ceosDrawn.splice(ceosDrawn.indexOf(ceo), 1);
+        game.log('${0} was discarded as ${1} could not play it,', (b) => b.card(ceo).player(player), {reservedFor: player});
       }
     });
 
     player.game.defer(new SelectPaymentDeferred(player, 8, {title: 'Select how to pay for action'}));
 
-    return new SelectCard('Choose CEO card to play', 'Play', cardsDrawn, (([card]) => {
-      const cardIndex = player.playedCards.findIndex((c) => c.name === this.name);
-      player.playedCards.splice(cardIndex, 1);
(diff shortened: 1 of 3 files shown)


src/server/cards/ceos/Lowell.ts
```diff
@@ -35,27 +35,33 @@ export class Lowell extends CeoCard {
...
     ];
 
-    cardsDrawn.forEach((card) => {
-      if (card.canPlay?.(player) === false) {
-        cardsDrawn.splice(cardsDrawn.indexOf(card), 1);
-        game.log('${0} was discarded as ${1} could not play it,', (b) => b.card(card).player(player), {reservedFor: player});
+    ceosDrawn.forEach((ceo) => {
+      if (ceo.canPlay?.(player) === false) {
+        game.ceoDeck.discard(ceo);
+        ceosDrawn.splice(ceosDrawn.indexOf(ceo), 1);
```
kberg, line 47: Another good use of `inplaceRemove`

src/server/cards/ceos/Lowell.ts
```diff
@@ -35,27 +35,33 @@ export class Lowell extends CeoCard {
...
+      ceosDrawn.filter((c) => c !== chosenCeo).forEach((c) => game.ceoDeck.discard(c));
+      // Remove Lowell from played cards
+      const lowellIndex = player.playedCards.findIndex((c) => c.name === this.name);
+      player.playedCards.splice(lowellIndex, 1);
+      // Add Lowell to Discard pile
       game.ceoDeck.discard(this);
-
-      return player.playCard(card);
+      // Play the new CEO
+      player.playCard(chosenCeo);
```
kberg, line 63: CEOs playCard returns void?

kberg (approved):
Minor stuff. Good PR

src/server/cards/ceos/Lowell.ts
d-little: `undefined`, its the same mechanism as project cards

src/server/cards/ceos/Lowell.ts
d-little: This raises a question for me, does modifying a list which is being iterated over cause problems in JS?

Im guessing not.  Python _hates_ it, never ever modify a container you're iterating on!

src/server/cards/ceos/Lowell.ts
kberg: Then I recommend returning it. That way if it returns something that's not undefined, it can get processed.

src/server/cards/ceos/Lowell.ts
kberg: Oh right, uh... good question, don't know. Perhaps forEach helps out. Perhaps it's wrong in this case.

d-little: I got rid of the loop entirely, it felt.. better

(For future readers: Yes, iterating over a container and removing items will cause it to skip indexes and migraines)

d-little:
> Minor stuff. Good PR

Ready for rereview!

PR #5619, opened 2023-04-25
Fix "End Turn" with Mars Maths and other 'increased/decreased actions per round'

### LINKED ISSUE #5618: Mars Maths and End Turn allow for unlimited actions
[04:30]Borbarad: Still the bug with Mars Maths: I could do several turns a row:
image
[04:31]Borbarad: URL: http://terraforming-mars.herokuapp.com/spectator?id=sb0f14522daaa
Player color: green
Step: 103
Version: 0292905, built at Sun, 16 Apr 2023 04:39:12 GMT
Browser: Firefox 112
[04:41]Borbarad: Although i end turn, i could do two actions more:
image
[05:00]Borbarad: I abused the bug and could do as many actions i wanted, due to I end my turn in the third action, then i do an action, then I end my turn, then I could do an other action, then I end my turn, and so on and so. 
image
image
[05:15]Borbarad: And making undo in between the three action is also not working, and result in an interrupt of my turn!
[05:17]Borbarad: happen in the game mentioned above in gen7

d-little:
This should fix #5618

### DIFF
--- a/src/server/Player.ts
+++ b/src/server/Player.ts
@@ -1470,7 +1470,7 @@ export class Player {
 
   private endTurnOption(): PlayerInput {
     return new SelectOption('End Turn', 'End', () => {
-      this.actionsTakenThisRound = 1;
+      this.actionsTakenThisRound = this.actionsThisRound; // This allows for variable actions per turn, like Mars Maths
       this.game.log('${0} ended turn', (b) => b.player(this));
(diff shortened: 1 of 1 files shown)


kberg (approved):
I'm going to rename actionsThisRound to make that statement less ambiguous. Can be done in a separate PR.

PR #5634, opened 2023-04-29
Change hide cards in hand functionality to be similar to card filters

velit:
In other words add a toggle button similar to other card filters next to the Cards in hand section that allows hiding cards in hand without going into preferences.

Vis when toggled off:
image

Vis when toggled on:
image


kberg (approved):
This is pretty good, and can go in as-is. NB that I was able to make a change to your branch. It matches the text that accompanies the other toggles. (I hate all of them, but I'd rather have consistency first, and solve that problem second.)

image

PR #5635, opened 2023-04-29
Updating russian translation

tltneon:


src/locales/ru/ui.json
```diff
@@ -2,10 +2,13 @@
     "Terraforming Mars": "Покорение Марса",
     "Create New Game": "Новая игра",
     "№ of Players": "Кол-во игроков",
+    "Solo": "Соло",
+    "solo": "Соло",
```
kberg, line 6: Did you mean to duplicate this?

src/locales/ru/ui.json
```diff
@@ -213,15 +245,31 @@
     "Remove negative Global Events": "Убрать негативные события Кризиса",
     "All": "Все",
     "Filter": "Фильтры",
+    "show": "показать",
+    "concise": "компактно",
+    "CONCISE": "КОМПАКТНО",
+    "full": "полностью",
+    "FULL" :"ПОЛНОСТЬЮ",
```
kberg, line 252: I'm pretty sure the language is not case sensitive, which means this might not work the way you hope. Let's see. :crossed_fingers:

kberg (commented):
Very minor comments. This is a good change. Thank you!

tltneon:
Hello, @kberg !
Let's think that it was just a try from me :)
I looked at another locale files to get most used lines.

src/locales/ru/ui.json
tltneon: I took lines mostly from German file so I thought that make sence
image

tltneon:
As I found language is case sensitive
image
image


But I think this is necessary for some reasons like using e.g. 3 word for 3 different places in entire project. It can be same word in one language but it also can be different words in other languages.

src/locales/ru/ui.json
kberg: Yeah, I think you're right.

kberg (approved):
Thank you!

PR #5639, opened 2023-04-30
[Experimental] Show player cube on used action cards.

### LINKED ISSUE #5603: Represent used action cards without dimming the full card?
It's useful to know which action cards have already been used this generation. Currently that's shown by dimming the entire card. This has a couple negatives:

* when choosing where to put resources, it makes it look like you can't pick that card (but just removing that graying would lose useful information)
* I'm still hopeful that #1607 could one day happen, which would block on this

imho the ideal way to depict a used action card is to put a player-cube on it. That has the side benefit that when someone plays the physical game for the first time they will already know what they're supposed to do.

Second best would be something that dims only the actions themselves, or the full interior of the card, but doesn't dim the tags.

kberg:
Fixes #5603


kberg:
image

Valmonzo:
I agree with this idea !

kberg:
Updated the prototype with a larger cube. 
image

PR #5640, opened 2023-04-30
Russian translation (again) & little visual improvement

tltneon:
I replaced original Prototype font with new one with supporting Cyrillic symbols (for russian, ukrainian, bulgarian, serbian an so on)

image

And also edit CSS rules that sets name exactly at the middle of screen.

Next one. I made little feature: when card has a huge of description, it can be scrolled.
It will look like (pointed scroll at the right side)
image image

Needs almost for every locale
image image
image image

PR #5655, opened 2023-05-06
Adding scrolling for cards with huge description

tltneon:
Splitted from #5640

When card has a huge of description, it can be scrolled.
It will look like (pointed scroll at the right side)
image image

Needs almost for every locale
image image
image image

### DIFF
--- a/src/client/components/card/Card.vue
+++ b/src/client/components/card/Card.vue
@@ -7,7 +7,7 @@
               <CardTags :tags="getTags()" />
           </div>
           <CardTitle :title="card.name" :type="getCardType()"/>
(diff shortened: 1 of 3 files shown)


kberg:
Thanks for this one. I'm about to test it locally.

kberg:
Some observations. I'll add each in a separate comment:

The space taken on the right side does have a small impact on large cards:

image
image

kberg:
It doesn't seem to work with cards with extended effects:

image

But for those if you hover, the effect covers the resource count.

kberg:
It has a small impact on global events (note how the Influence icon is partially obscured)

image
image

kberg:
> It has a small impact on global events (note how the Influence icon is partially obscured)

For the example, the obscured icon is due to the new feature.

kberg:
Some other UX thoughts: it's a bit tough to grab the scroll bars. But it is still a huge improvement.

kberg (approved):
This is a good change.

PR #5697, opened 2023-05-15
Added ability to translate for geological places

tltneon:
How it is now
image

And how it will be (some string shifts)
image
image


kberg (approved):
This is a good change.

kberg:
This has broken the Tharsis map image

I'll check the remaining maps. I might roll back part of this PR for a while.

PR #5712, opened 2023-05-21
Style fixes

tltneon:
Fixed https://github.com/terraforming-mars/terraforming-mars/pull/5655#issuecomment-1537427478
image
Link now looks like a link
image
When mouse over
image


kberg:
Thanks!

PR #5713, opened 2023-05-21
Fully translated Help page

tltneon:


src/locales/ru/help_iconography.json
```diff
@@ -4,16 +4,19 @@
     "Tiles": "Тайлы",
     "Global Parameters": "Глобальные параметры",
     "Others": "Другие",
-    "Standard Resources": "Стандартные ресурсы",
-    "Resources on Cards": "Ресурсы на картах",
+    " Standard Resources ": "Стандартные ресурсы",
```
kberg, line 7: This shouldn't be necessary because incoming text is supposed to be trimmed.  I'm going to add a change which removes the leading and trailing spaces, and fix it in code.

kberg (approved):

src/locales/ru/help_iconography.json
tltneon: I also thought about that and I come to opinion that it can broke some logic both for original and translated strings

PR #5698, opened 2023-05-15
Tooltips for player resources

tltneon:
image
image

PR #5720, opened 2023-05-22
Tooltips for player's tags

tltneon:
similar to #5698
image
image
image


kberg (approved):
This is a good change. Thank you.

PR #5724, opened 2023-05-26
Update start_screen.json

Borbarad2:
It is not working without these.
See russian translations, there it is working like this.

### DIFF
--- a/src/locales/de/start_screen.json
+++ b/src/locales/de/start_screen.json
@@ -12,5 +12,5 @@
     "Looking for people to play with": "Suchst du Mitspieler",
     "version": "Version",
     "deployed": "Installiert",
-    "Not affiliated with FryxGames, Asmodee Digital or Steam in any way.": "In keinster Weise verbunden mit FryxGames, Asmodee Digital oder Steam."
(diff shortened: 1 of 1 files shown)


src/locales/de/start_screen.json
```diff
@@ -12,5 +12,5 @@
     "Looking for people to play with": "Suchst du Mitspieler",
     "version": "Version",
     "deployed": "Installiert",
-    "Not affiliated with FryxGames, Asmodee Digital or Steam in any way.": "In keinster Weise verbunden mit FryxGames, Asmodee Digital oder Steam."
+    " Not affiliated with FryxGames, Asmodee Digital or Steam in any way. ": "In keinster Weise verbunden mit FryxGames, Asmodee Digital oder Steam."
```
kberg, line 15: Please don't add spaces like this. I'll fix the offending code if necessary.

kberg (changes requested):

src/locales/de/start_screen.json
Borbarad2: Yes please, then correct these lines and remove the space, then it will be correct in all language at the same time. And then this PR is obsolete.

PR #5741, opened 2023-06-01
Remove BETA tag from CEOs

d-little:
All the core components are done, remaining issues are external from the CEO application itself.

Wiki is up to date: https://github.com/terraforming-mars/terraforming-mars/wiki/CEOs

All future issues/bugs can be reported as regular App bugs, not part of the CEO Beta


kberg (approved):
:clap:

PR #5743, opened 2023-06-01
New promo card Supercapacitors

kberg:
image

### DIFF
--- a/src/server/Player.ts
+++ b/src/server/Player.ts
@@ -67,6 +67,7 @@ import {calculateVictoryPoints} from './game/calculateVictoryPoints';
 import {IVictoryPointsBreakdown} from '..//common/game/IVictoryPointsBreakdown';
 import {YesAnd} from './cards/requirements/CardRequirement';
 import {PlayableCard} from './cards/IProjectCard';
+import {Supercapacitors} from './cards/promo/Supercapacitors';
 
 
 const THROW_WAITING_FOR = Boolean(process.env.THROW_WAITING_FOR);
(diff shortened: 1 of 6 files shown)


src/server/Player.ts
```diff
@@ -740,13 +741,23 @@ export class Player {
...
+    } else {
+      this.heat += this.energy;
+      this.energy = 0;
+      this.finishProductionPhase();
+    }
+  }
+
+  public finishProductionPhase() {
     this.megaCredits += this.production.megacredits + this.terraformRating;
-    this.heat += this.energy;
```
kberg, line 744: This is just reordering.

src/server/Player.ts
```diff
@@ -740,13 +741,23 @@ export class Player {
...
-    this.heat += this.production.heat;
-    this.energy = this.production.energy;
```
kberg, line 746: But this is replaced with this.energy +=, below.

d-little (approved):
Fast work, great job!

src/server/cards/promo/Supercapacitors.ts
```diff
@@ -0,0 +1,53 @@
...
+          b.br;
+          b.production((pb) => pb.megacredits(1));
+        }),
+        description: 'Increase M€ production 1 step.',
+      },
+    });
+  }
+
+  public static onProduction(player: Player) {
+    if (player.energy === 0) {
```
nwai90, line 35: This is incorrect IMO. If a player happens to have positive energy production but 0 energy at the end of the round (i.e. they spent all their energy during the generation),  they will never produce any resources as `player.finishProductionPhase()` will never be called

nwai90 (commented):
It seems there is a bug in this implementation

PR #5753, opened 2023-06-02
Moon card Road Piracy

kberg:
image

### DIFF
--- a/src/server/cards/moon/RoadPiracy.ts
+++ b/src/server/cards/moon/RoadPiracy.ts
@@ -1,0 +1,120 @@
+import {CardName} from '../../../common/cards/CardName';
+import {Player} from '../../Player';
+import {CardType} from '../../../common/cards/CardType';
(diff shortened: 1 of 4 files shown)


alrusdi (approved):

src/server/cards/moon/RoadPiracy.ts
```diff
@@ -0,0 +1,120 @@
...
+    }
+    if (selectAmounts.length === 0) {
+      return undefined;
+    }
+
+    const cb = () => {
+      const total = sum(ledger.map((e) => e[1]));
+      if (total > limit) {
+        // throw new Error(newMessage('You may only steal up to ${0} ${1} from all players', (b) => b.number(limit).string(resource)));
+        throw new Error(`You may only steal up to ${limit} ${resource} from all players`);
```
nwai90, line 70: This is bugged - the ledger should be emptied / reset before throwing the error, otherwise the player can never complete the action even after correcting their input values

PR #5756, opened 2023-06-03
Reset the Road Piracy ledger on error.

kberg:
This prevents an endless loop.


nwai90 (approved):
Looks good!

kberg:
Thank you.

Issue #5353, opened 2023-02-12
Add Description field to Colonies


d-little:
## Why?

Primarily I can see this being used in the same way as the Description fields for Awards+Milestones.

Prompted by this post in discord: https://discord.com/channels/737945098695999559/787457316009803786/1074449627497123912

## How

I imagine we can extend `IColony` to include `description: String`, and put these on each colony.

Then on each of the player/cardlist, duplicate the behaviour of Awards+Milestones.

 

### Cards List Behaviour

In the Cards List, Milestones look something like:

![image](https://user-images.githubusercontent.com/2050250/218343063-6c8fdeba-ea53-4b4c-8fcb-ef3c35f5fc00.png)

I can see the Colony descriptions looking something like this quick hack:

![image](https://user-images.githubusercontent.com/2050250/218343236-a10b2594-4e40-4998-ad3f-f0712925797e.png)

### PlayerView Behaviour

I like the Milestone/Award description implementation;  click on them and they show up, click again and they go away:

![image](https://user-images.githubusercontent.com/2050250/218343391-40a8a1f5-5e90-4d39-a2f1-3577c3608083.png)

I can see the same thing for the PlayerView of Colonies:

![image](https://user-images.githubusercontent.com/2050250/218343503-0440519e-7afa-44d5-956d-e142b11ae4a9.png)

kberg:
Yeah this makes sense.

PR #5764, opened 2023-06-06
Add description for colonies

kberg:


kberg:
This is using a hover label.
image

This isn't ideal, though.

image

kberg:
Fixes #5353. Not perfectly, but enough to get started.

PR #5794, opened 2023-06-13
Use a stronger webpack hash function

cjwatson:
Building on Debian 12 failed with `error:0308010C:digital envelope routines::unsupported`, because `webpack` defaults to MD4 which is disallowed by recent versions of OpenSSL.  Use a stronger hash function instead.  This required first upgrading `webpack`, which in turn required upgrading `vue-loader`.


kberg (approved):
Thank you! This is a good PR.

PR #5810, opened 2023-06-20
MA naming based on #5685

### LINKED ISSUE #5685: Award and milestone description suggestions
Okay so I'm gonna do my thing where I offer my lil nitpicks about something, and please just ignore any you don't feel like doing and no need to explain.

Yes, they're very minor. You've seen this movie before...

### General

When we discussed how the game uses "in play", I overlooked something. I believe it *is* helpful for the case of distinguishing *cards* in play as opposed to cards *in hand*. It's almost like a "cardinplay" is a different noun from a "cardinhand". For most of these "Play N cards" (or "Have played N cards?") M&As, I think "Have N cards in play" would be less ambiguous. This applies to Excentric too, which was actually officially retconned to add "in play" (to exclude SRR resources).

"After final production" could be misleading, suggest "after production and greenery placement", or since people are aware awards get judged at the end, the clarification this is really trying to make might be just "(note there is a final production)".

For "Have played N events" I would suggest "have N cards in your events pile", only because of PharmU and Law Suit and that one fan corp...

Colony things: curious why it's "have built" and not just "have"? If nothing can ever remove them then it makes no difference, but still...

This one is not necessarily wrong as-is, but... I have heard numerous experienced players insist that corporation cards are "active" and/or that they're "blue" (imho, they're neither). So wording that relies on those terms might be misinterpreted... maybe "blue project cards"?

### Specific

Tourist: game always uses "area" instead of "space" (which is good because those without good English might really confuse what "space" means in the context of a game like this). Suggest "Have the most areas with no tile adjacent to your own tiles" because it's never been ruled whether communities (i.e. LandClaim/Arcadian) make an area nonempty.

A. Zoologist: I think "own" is normally used just for tiles and the rest are "have"? And it's just because they're in a communal place with your "ownership marker" on them...

Biologist: The phrase "count as bio tags" just seems unnecessary

Politician: "Have" the most delegates? Can't remember whether that includes party leaders or the Chairman, but that might be common knowledge for Turmoil players.

Urbanist: unclear whether it counts oceans next to Capital or not. I think that technically those are card VPs and if a fan card could steal your Capital card you'd lose them.

Warmonger: suggest '...that can reduce resources/production of "any player"'  .. the "can" is, of course, because it's irrelevant whether you *did*...

Terraformer: this seems like a special case where "35 *or higher*" may be worth it. I'm not sure everyone thinks of TR as a quantity of things you have, so this might really sound like "your TR must BE 35".

Generalist: I think "have increased" solves some misunderstandings but maybe causes others (i.e., that the game has a memory of whether you ever *did* increase it). What would be great is if it could swap between "have at least 1 of each" or "have at least 2 of each" depending on CorpEra... in which case Terraformer could do that too.

Specialist: suggest "of any one resource", as it sorta sounds like you can add them up.

Networker: I'm not clear to me whether tiles next to oceans count. They do, precisely, grant adjacency bonuses, right?

Land specialist: how about "brown or Capital", "brown or white"?

Collector: I think some people get confused about "sets" so I was wondering about "have at least 3 active cards, 3 automated cards, and 3 cards in your event pile"

Tradesman: unclear if this includes SRR cards. Annoyingly, the official ruling is that those resources "have no type" (pffft)

(Cultivator: just noticed the icon has the oxygen symbol on it and shouldn't)

kberg:
Fixes #5685


d-little:
All makes sense to me, good changes

Issue #3707, opened 2021-10-10
[Moon] The Archaic Foundation Institute's 'automatic' behavior introduces some issues


kberg:
This card automatically raises TR 1 step once 3 cubes are accumulated. This introduces 2 problems:
1. That makes it tricky to enforce the reds tax if the player does not have enough money. When does the tax get paid? When does the player gain their TR? After every action?
2. How does the automatic computation get processed when it comes to cards like CEO's favorite project, which might put a cube on this card?

kberg:
Partially addressed by #4078

kberg:
> What else is to do here? I think everything is done. Automatic behavior is addressed, treated and running okay.

I changed the card to be reds-compatible. It's not ideal, but it'll work.

PR #5830, opened 2023-07-02
Update Archaic Foundation Institute to be more resilient with Reds.

kberg:
This isn't quite enough, because it should ideally reduce as many
as possible as soon as possible.


kberg:
Fixes #3707

Issue #5840, opened 2023-07-06
When selected card is changed, reserveUnits are not updated. 


puuj:
Found this bug in a game. The player has a card with reserved titanium as the first in the sort order. Switching to a different card preserves the reserved titanium preventing them from spending all their titanium. Solution would be to add
`this.reserveUnits = this.card.reserveUnits ?? Units.EMPTY;` in the cardChanged() callback.

https://github.com/terraforming-mars/terraforming-mars/blob/a98b5e3dbf9b74e8f6ab1186826e5e1aeeb2817c/src/client/components/SelectProjectCardToPlay.vue#L267

kberg:
Fixed.

PR #5841, opened 2023-07-06
Update reserveUnits when a new card is selected (#5840)

puuj:
Currently, the reserveUnits variable for playing a card is fixed to the first card selected and when the selected card is changed, the variable is not updated. This may allow a user to try and play a card they cannot afford or be unable to play a card that they can afford. The bug is fixed by updating reserveUnits as the selected card changes.


kberg:
You're the best.

kberg (approved):

PR #5845, opened 2023-07-08
Fix placement log to correspond to coordinates shown on map

arseniybanayev:
We played [this game](https://terraforming-mars.herokuapp.com/the-end?id=p80534bc4bdc9) and noticed that when placing tiles, the logs did not match the coordinates we saw on the game board. The coordinates displayed on the game board correctly correspond the `Space` instances (where row is `y` and position from the left on that row is `x`), so I changed the log message to be consistent. I removed the `offset` logic (which I can't rationalize) and I removed the `+ 1` because although humans like 1-based indexing, the coordinates displayed on the map use 0-based indexing. I could be convinced that both the coordinates displayed on the map and the log should use 1-based indexing, though.

### DIFF
--- a/src/server/LogHelper.ts
+++ b/src/server/LogHelper.ts
@@ -39,12 +39,8 @@ export class LogHelper {
     // Skip solo play random tiles
     if (player.name === 'neutral') return;
 
-    const offset = Math.abs(space.y - 4);
-    const row = space.y + 1;
-    const position = space.x - offset + 1;
-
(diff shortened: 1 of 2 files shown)


kberg:
Thanks for this PR. The I added visual component showing row and column because, well, rows and columns are stored differently than shown. I needed it for my own development.

In other words, what is in the log is correct. I'll roll back the visual component showing a different row/column structure. Like I said, it was just for me.

PR #5854, opened 2023-07-16
Make Escape Velocity bonus seconds configurable 

ejao1:
According to https://github.com/terraforming-mars/terraforming-mars/wiki/Escape-Velocity, the number of seconds you get back per action should be configurable. This PR makes it so and also fixes some Escape Velocity translations.


src/client/components/GameSetupDetail.vue
```diff
@@ -112,6 +113,13 @@ export default Vue.extend({
     boardColorClass(): string {
       return boardColorClass[this.gameOptions.boardName];
     },
+    escapeVelocityDescription(): string {
+      const {escapeVelocityThreshold, escapeVelocityPenalty, escapeVelocityPeriod} = this.gameOptions ?? {};
```
kberg, line 117: clever, #til

src/client/components/GameSetupDetail.vue
```diff
@@ -47,7 +47,7 @@
 
             <li v-if="gameOptions.escapeVelocityMode">
               <div class="create-game-expansion-icon expansion-icon-escape-velocity"></div>
-              <span>After {{gameOptions.escapeVelocityThreshold}} min, reduce {{gameOptions.escapeVelocityPenalty}} VP every {{gameOptions.escapeVelocityPeriod}} min.</span>
+              <span>{{escapeVelocityDescription}}</span>
```
kberg, line 50: Does this work instead?
`<span v-i18n>After {{gameOptions.escapeVelocityThreshold}} min, reduce {{gameOptions.escapeVelocityPenalty}} VP every {{gameOptions.escapeVelocityPeriod}} min.</span>`

src/client/components/create/CreateGameForm.vue
```diff
@@ -945,6 +952,7 @@ export default (Vue as WithRefs<Refs>).extend({
       const requiresVenusTrackCompletion = this.requiresVenusTrackCompletion;
       const escapeVelocityMode = this.escapeVelocityMode;
       const escapeVelocityThreshold = this.escapeVelocityMode ? this.escapeVelocityThreshold : undefined;
+      const escapeVelocityBonusSeconds = this.escapeVelocityBonusSeconds ? this.escapeVelocityBonusSeconds : undefined;
```
kberg, line 955: Hm. I realize you copied from the line above, but I wonder what the value of the `x ? x : undefined` pattern is. I guess it resolves 0 and null to undefined (or something close to that.)

kberg (commented):
Thanks for this thorough PR. Some minor comments.

src/client/components/create/CreateGameForm.vue
ejao1: Yep it coalesces falsy values to undefined and yes I did copy it from the surrounding lines :P

src/client/components/GameSetupDetail.vue
kberg: OK. Thanks.

kberg (approved):
This is a good PR.

PR #5858, opened 2023-07-18
Added Hungarian language support for the official expansions

ligvigfui:
I plan to add the unofficial expansions too but right now I don't have time to do that.
If I messed something up (in the translation) you can reach me at ligvigfui@gmail.com


kberg:
Wow, this looks terrific. Just a couple of files you need to remove.

ligvigfui:
Ok, tell me what do I do? This is my first PR on github. How do I remove what and why?

PR #5868, opened 2023-07-29
Pathfinders minor css tweaks

ssimeonoff:
Changes:
- numbers track is gone (it is useless for the gameplay)
- all empty cells have the same width (only 1 card in the game care which is the lowest track and it is not that hard to check anyway)
- more padding
- more prominent highlight of the current level track cells

Top: Before
Bottom: After
Untitled

### DIFF
--- a/src/client/components/pathfinders/PlanetaryTrack.vue
+++ b/src/client/components/pathfinders/PlanetaryTrack.vue
@@ -1,6 +1,6 @@
 <template>
     <tr>
-      <td>{{icon}}</td>
+      <div class="track-icon">{{icon}}</div>
       <td v-for="idx in range" :key="idx" :class="getClass(idx)">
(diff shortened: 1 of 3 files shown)


src/client/components/pathfinders/PlanetaryTrack.vue
```diff
@@ -1,6 +1,6 @@
 <template>
     <tr>
-      <td>{{icon}}</td>
+      <div class="track-icon">{{icon}}</div>
```
kberg, line 3: Interesting! A div in a table #til

kberg (approved):
This is a good PR.

src/client/components/pathfinders/PlanetaryTrack.vue
ssimeonoff: couldnt think of an easier cheat

PR #5874, opened 2023-08-19
Fix Secret Labs card

alrusdi:


kberg (approved):
This is really good. If you're interested, the same issue will happen with temperature and oxygen. Would you like to fix those also?

alrusdi:
Raising temperature and oxygen is not blocking UI like oceans do. But i'll check it

PR #5898, opened 2023-09-03
Fix Ares' Magnetic Field Generator rendering

### LINKED ISSUE #5769: Non-ares tile Magnetic Field Generators shown
image

Although Ares expansion was enable and the Card MFG was the Ares style promo card, the tile plced was the none Ares tile.

kberg:
Fixes #5769

image


kberg:
Fixes https://github.com/terraforming-mars/terraforming-mars/issues/5769

PR #5911, opened 2023-09-07
Update promo_cards.json

Borbarad2:
Added new promo card Aqueduct Systems


src/locales/de/promo_cards.json
```diff
@@ -11,6 +14,10 @@
   
   "16 Psyche": "16 Psyche",
   "Increase titanium production 2 steps. Gain 3 titanium.": "Erhöhe deine Titan-Produktion um 2. Erhalte 3 Titan.",
+
+  "Applied Science": "Angewandte Wissenschaft",
```
kberg, line 18: This should probably get put in a new file. But this is fine for now.

kberg (approved):

PR #5931, opened 2023-09-11
remove util fallback from webpack.config.js

bafolts:
I want to setup a mode where single player games can be run without the server! This would make the game more portable for enthusiasts of the single player games. Was taking a look again at the webpack config and noticed we provide a util fallback for the client testing. Updating sinon to most recent version avoided the dependency.

This change updates sinon to latest and makes the webpack config file simpler!


kberg (approved):
Good to see you boss

PR #5932, opened 2023-09-11
Render starting production and megacredits for Tycho Magnetics

puuj:
Currently the Tycho Magnetics card does not show the starting production or megacredits. This change adds that information and makes it similar to other corporation cards.


kberg (approved):
Thanks, @puuj!

PR #5942, opened 2023-09-13
allow for optional database providers

bafolts:
This change allows for pg or sqlite3 to fail to load. I am working on a mode where the server will run in a web worker. Within web worker will not use pg or sqlite3 so need these dependencies to be optional. This change should also help users who struggle to get sqlite3 to install. The file system database is now used as default if we fail to setup any other database.

### DIFF
--- a/src/server/database/PostgreSQL.ts
+++ b/src/server/database/PostgreSQL.ts
@@ -50,7 +50,7 @@ export class PostgreSQL implements IDatabase {
   }
 
   public async initialize(): Promise<void> {
-    const Pool = (await require('pg')).Pool;
+    const {Pool} = await import('pg');
     this._client = new Pool(this.config);
     await this.client.query('CREATE TABLE IF NOT EXISTS games(game_id varchar, players integer, save_id integer, game text, status text default \'running\', created_time timestamp default now(), PRIMARY KEY (game_id, save_id))');
     await this.client.query('CREATE TABLE IF NOT EXISTS participants(game_id varchar, participants varchar[], PRIMARY KEY (game_id))');
--- a/src/server/database/PostgreSQL.ts
+++ b/src/server/database/PostgreSQL.ts
@@ -1,15 +1,14 @@
+import type * as pg from 'pg';
 import {IDatabase} from './IDatabase';
 import {IGame, Score} from '../IGame';
 import {GameOptions} from '../game/GameOptions';
 import {GameId, ParticipantId} from '../../common/Types';
 import {SerializedGame} from '../SerializedGame';
-import {Pool, ClientConfig, QueryResult} from 'pg';
 import {daysAgoToSeconds} from './utils';
 import {GameIdLedger} from './IDatabase';
 import {difference} from '../../common/utils/utils';
 
 export class PostgreSQL implements IDatabase {
-  protected client: Pool;
   private databaseName: string | undefined = undefined; // Use this only for stats.
 
   protected statistics = {
@@ -18,9 +17,17 @@ export class PostgreSQL implements IDatabase {
     saveConflictUndoCount: 0,
     saveConflictNormalCount: 0,
   };
+  private _client: pg.Pool | undefined;
+
+  protected get client(): pg.Pool {
+    if (this._client === undefined) {
+      throw new Error('attempt to get client before intialized');
+    }
+    return this._client;
+  }
 
   constructor(
-    config: ClientConfig = {
+    private config: pg.ClientConfig = {
       connectionString: process.env.POSTGRES_HOST,
     }) {
     if (config.connectionString?.startsWith('postgres')) {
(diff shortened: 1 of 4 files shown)


src/server/database/Database.ts
```diff
@@ -12,15 +13,27 @@ export class Database {
     if (!Database.instance) {
       if (process.env.POSTGRES_HOST !== undefined) {
         console.log('Connecting to Postgres database.');
-        Database.instance = new PostgreSQL();
+        try {
+          Database.instance = new PostgreSQL();
+        } catch (err) {
+          console.warn('unable to start PostgreSQL', err);
```
kberg, line 19: This should be a throw.

src/server/database/Database.ts
```diff
@@ -12,15 +13,27 @@ export class Database {
...
       } else if (process.env.LOCAL_FS_DB !== undefined) {
         console.log('Connecting to local filesystem database.');
         Database.instance = new LocalFilesystem();
       } else {
         console.log('Connecting to SQLite database.');
-        Database.instance = new SQLite();
+        try {
+          Database.instance = new SQLite();
+        } catch (err) {
+          console.warn('unable to start SQLite', err);
```
kberg, line 29: Same here.

kberg (commented):
By and large most of the console.warns should be throws. If someone defines POSTGRES_HOST and pg isn't installed, that's a case for throwing. Same for Sqlite, even though that's the fallback.

kberg:
> By and large most of the console.warns should be throws. If someone defines POSTGRES_HOST and pg isn't installed, that's a case for throwing. Same for Sqlite, even though that's the fallback.

I say most, but it could also be all of them.

src/server/database/Database.ts
bafolts: That works. For the LocalStorageDatabase I will set an environment variable like LOCAL_FS_DB when that is to be used.

src/server/database/PostgreSQL.ts
```diff
@@ -1,15 +1,23 @@
+import type * as pg from 'pg';
 import {IDatabase} from './IDatabase';
 import {IGame, Score} from '../IGame';
 import {GameOptions} from '../game/GameOptions';
 import {GameId, ParticipantId} from '../../common/Types';
 import {SerializedGame} from '../SerializedGame';
-import {Pool, ClientConfig, QueryResult} from 'pg';
+let Pool: typeof pg.Pool | undefined;
+
+try {
```
kberg, line 9: Why not let this throw?

src/server/database/SQLite.ts
```diff
@@ -4,8 +4,16 @@ import {GameOptions} from '../game/GameOptions';
 import {GameId, ParticipantId} from '../../common/Types';
 import {SerializedGame} from '../SerializedGame';
 
-import sqlite3 = require('sqlite3');
-import {RunResult} from 'sqlite3';
+import * as sqlite3 from 'sqlite3';
+
+let Database: typeof sqlite3.Database | undefined;
+
+try {
```
kberg, line 11: Why not let this throw?

src/server/database/PostgreSQL.ts
bafolts: If you locally run a `rm -rf node_modules/pg` that may show the problem I am working around here. This will remove the postgres dependency leaving you in situation of a user that was unable to install postgres. If you then build and start the server with the logging here you will see the info logged.

```
[2023-09-16 12:41:16 MDT] postgres not installed, use another database or npm install pg
```

If instead this were to throw the server would be unable to start as the exception would be thrown once we required the PostgreSQL module which in turn will throw this exception.

Another option would be to transition towards dynamic imports versus the static imports we use today. At that point we could only call to require PostgreSQL if the user defined the environment variable to use PostgreSQL. Changing to dynamic imports would require us to introduce an await somewhere so we had time to load the module before using it.

kberg: Perhaps the app should only try to load Postgres when the environment variable is set.  Or when the supplied class is instantiated.

My concern about an ignorable warning  is that it puts users who _want_ postgres in a position where the app will silently fall back on a database they don't want.

kberg: Perhaps the app should only try to load Postgres when the environment variable is set.  Or when the supplied class is instantiated.

My concern about an ignorable warning  is that it puts users who _want_ postgres in a position where the app will silently fall back on a database they don't want.

kberg: Perhaps the app should only try to load Postgres when the environment variable is set.  Or when the supplied class is instantiated.

My concern about an ignorable warning  is that it puts users who _want_ postgres in a position where the app will silently fall back on a database they don't want.

bafolts: I will re-factor postgres and sqlite to dynamically require their dependencies in the initialize function.

kberg:
This is starting to look kinda complicated. Is the dynamic import mechanism what you're looking for? https://v8.dev/features/dynamic-import

bafolts:
That dynamic import syntax works too. What I had prior works but is using the old AMD style method. I have updated this to use the newer dynamic import syntax.

kberg:
Thank you! This is a good PR.

kberg (approved):

PR #5954, opened 2023-09-16
Introducing LocalStorage as an IDatabase.

bafolts:
This allows for one to use LocalStorage for IDatabase. When one eventually runs games with the server running in a worker this can be used for the database implementation. May also prove useful for those looking for something that doesn't need file system access.


kberg (approved):
Nice!

PR #5971, opened 2023-09-18
Update promo_corporations.json

Borbarad2:
Added new promo Corp Kuiper Cooperative


kberg:
Merged as one commit in #6002

PR #5972, opened 2023-09-18
Update prelude_corporations.json

Borbarad2:
Added the new prelude2 corps


kberg:
Merged as one commit in #6002

PR #5973, opened 2023-09-18
Update pathfinder_preludes.json

Borbarad2:
Adapt Anti-deserfication Techniques


kberg:
Merged as one commit in #6002

PR #5974, opened 2023-09-18
Update promo_cards.json

Borbarad2:
cleaning up promo cards with out prelude cards


kberg:
Merged as one commit in #6002

PR #5975, opened 2023-09-18
Update prelude.json

Borbarad2:
Adding new prelude cards
sorting prelude cards in sections


kberg:
Merged as one commit in #6002

PR #5976, opened 2023-09-18
Update prelude_cards.json

Borbarad2:
adding in the new prelude2 cards


kberg:
Merged as one commit in #6002

PR #5977, opened 2023-09-18
Update log_messages.json

Borbarad2:
Added new log_messages from promo card Astra Mechanica


kberg:
Merged as one commit in #6002

PR #5988, opened 2023-09-20
Star Wars fan expansion

kberg:
This is my star wars fan expansion, a series of 9 cards. Is now the time to add this? What the heck.

I'll take some time getting this reviewed, though. I need to test this a few more times.


Borbarad2:
Wow, great, finally. Great stuff.
We played them already in a rl game. Great.
Where do you put these cards? Community or extra section?

PR #5996, opened 2023-09-21
CEO Jane (LFC, no tests)

d-little:
Before I commit to writing tests for this, does it look 'right'?  Had to introduce some new concepts that I'd like to confirm before putting down the tests.

Basic human-run tests came back Okay!  Including Prelude and Corporation Actions, which is important.

image


kberg:
My LFC response is that I think this could be very easily overpowered. I would (like I think most people) wait until the last generation, and then tap the CEO. IMO the best CEOs require you to think about the right time to tap them.

kberg:
Feel free to reopen this when there's some traction.

PR #5997, opened 2023-09-21
Fix Solar Logistics Price

d-little:
Should be 20, not 14.  

(I'm assuming that was a copy/paste error from Tesalaract)


kberg (approved):

kberg:
Wow, thanks!

PR #5998, opened 2023-09-21
add: log additions to venusian animals

bobertstark:
Small QOL addition to the log

Sidenote: Logging across the board seems arbitrary to me, is there some criteria?


kberg (approved):
Looks great. Thanks!

PR #6000, opened 2023-09-21
abstract away http

bafolts:
I want to add support for an offline mode and allow the current server javascript to run in a worker. Currently we rely on http.IncomingMessage and http.ServerResponse for taking in a request and serving response. We don't use many parts of these types yet require the full type. This change creates a Request and Response type which picks from http for the parts we use. If nothing else this makes unit testing simpler as we only have to create the smaller Request and Response objects versus the full http.IncomingMessage and http.ServerResponse.

### DIFF
--- a/src/server/Request.ts
+++ b/src/server/Request.ts
@@ -1,6 +1,10 @@
 
 import type * as http from 'http';
 
+/**
+ * Limits Request to what we use. This is used to ease in unit testing
+ * and to allow for javascript without 'http' to run the server code.
+ */
 export type Request = Pick<http.IncomingMessage, 'headers' | 'method' | 'url'> & {
   once: (type: 'end', func: () => void) => void;
   on: (type: 'data', func: (dat: Buffer) => void) => void;
--- a/src/server/Request.ts
+++ b/src/server/Request.ts
@@ -1,0 +1,15 @@
+
+import type * as http from 'http';
+
+export type Request = Pick<http.IncomingMessage, 'headers' | 'method' | 'url'> & {
+  once: (type: 'end', func: () => void) => void;
+  on: (type: 'data', func: (dat: Buffer) => void) => void;
+  socket: {
+    address(): string | {
+      address: string;
+      family: string;
+      port: number;
+    };
+  }
+};
+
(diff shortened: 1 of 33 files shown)


tests/routes/HttpMocks.ts
```diff
@@ -1,29 +1,23 @@
...
     if (content) {
       this.content += content;
     }
   }
   public writeHead(statusCode: number): void {
     this.statusCode = statusCode;
   }
-
-  // MockResponse isn't really of type ServerResponse,
-  // and this makes it seem like it is so we can sneak
```
kberg, line 24: ooooo nice

tests/routes/PlayerInput.spec.ts
```diff
@@ -11,16 +10,17 @@ import {RouteTestScaffolding} from './RouteTestScaffolding';
...
 
 describe('PlayerInput', function() {
   let scaffolding: RouteTestScaffolding;
-  let req: EventEmitter;
+  let req: Request;
   let res: MockResponse;
 
   beforeEach(() => {
-    req = new EventEmitter();
+    req = new EventEmitter() as unknown as Request;
```
kberg, line 21: What happens if EventEmitter isn't forced into being Request?

src/server/Request.ts
```diff
@@ -0,0 +1,15 @@
+
+import type * as http from 'http';
+
+export type Request = Pick<http.IncomingMessage, 'headers' | 'method' | 'url'> & {
```
kberg, line 4: Can you add something close to this, please? Repeat for Response.

```
// An abstraction of http.IncomingMessage, exposing only those methods the app needs.
//
// This abstraction enables launching a serverless in-browser copy of the game.
```

kberg (approved):
Modulo a question and a request for some code comments, this looks great.

tests/routes/PlayerInput.spec.ts
bafolts: I can update this so it doesn't have to be forced to Request and make a real Request instance. Without this forced to Request typescript compilation fails with a compile time error.

kberg: Let's merge this now. We can fix this later. It's for testing after all.

bafolts: It made it in. Ended up being MockRequest and it fell into place pretty nicely.

https://github.com/terraforming-mars/terraforming-mars/blob/main/tests/routes/HttpMocks.ts#L5

PR #6001, opened 2023-09-22
Ceo balance update

d-little:
Couple of small balance changes to bring these into line with the tfm-community branch

### DIFF
--- a/src/server/cards/ceos/Bjorn.ts
+++ b/src/server/cards/ceos/Bjorn.ts
@@ -14,10 +14,10 @@ export class Bjorn extends CeoCard {
       metadata: {
         cardNumber: 'L02',
         renderData: CardRenderer.builder((b) => {
-          b.opgArrow().text('STEAL').megacredits(0, {multiplier}).asterix();
+          b.opgArrow().text('STEAL').megacredits(0, {multiplier}).megacredits(2).asterix();
           b.br;
         }),
-        description: 'Once per game, steal X M€ from each player that has more M€ than you, where X is the current generation number.',
+        description: 'Once per game, steal X+2 M€ from each player that has more M€ than you, where X is the current generation number.',
       },
     });
   }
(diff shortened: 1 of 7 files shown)


src/server/cards/ceos/Bjorn.ts
```diff
@@ -14,10 +14,10 @@ export class Bjorn extends CeoCard {
       metadata: {
         cardNumber: 'L02',
         renderData: CardRenderer.builder((b) => {
-          b.opgArrow().text('STEAL').megacredits(0, {multiplier}).asterix();
+          b.opgArrow().text('STEAL').megacredits(0, {multiplier}).megacredits(2).asterix();
           b.br;
         }),
-        description: 'Once per game, steal X M€ from each player that has more M€ than you, where X is the current generation number.',
+        description: 'Once per game, steal X+2 M€ from each player that has more M€ than you, where X is the current generation number.',
```
kberg, line 20: or as much as possible

src/server/cards/ceos/Xavier.ts
```diff
@@ -15,8 +18,10 @@ export class Xavier extends CeoCard {
           b.br.br;
           b.text('GAIN').nbsp.wild(2, {played});
           b.br.br;
+          b.cards(1, {secondaryTag: AltSecondaryTag.REQ}).colon().megacredits(-2);
+          b.br.br;
         }),
-        description: 'Gain 2 wild tags for THIS GENERATION.',
+        description: 'Once per game, gain 2 wild tags for THIS GENERATION. When playing a card with a requirement, you pay 1 M€ less for it AFTER this action has been used.',
```
kberg, line 24: I thought there was an OPG vis option

kberg (approved):

src/server/cards/ceos/Xavier.ts
d-little: Description vs. Iconography I guess.  Not all of the text has the 'Once per game,' in `description`, sometimes cut to keep things short.  Maybe it should?

For this one I think we need to be more specific due to the two effects working off each other

src/server/cards/ceos/Bjorn.ts
d-little: Does any card use language like that?  (This isnt a no, this card does steal a lot more than normal, just after similarly used language)

kberg: You are correct. Please ignore.

src/server/cards/ceos/Xavier.ts
kberg: Sorry I didn't see the opg rendering component.

kberg (approved):

PR #6012, opened 2023-09-22
update jsdom, remove jsdom-global and mutation-observer

bafolts:
Doing some spring cleaning in the fall. Here updating jsdom to the latest version and removing jdsom-global and mutation-observer as they can be replaced with a few lines of javascript as of the most recent jsdom version.

I removed a unit test that was spying on window.location.reload. The window.location object in jsdom makes it difficult to spy on. I didn't think this test was providing much value so decided to remove it. I can add it back but we wont be able to spy on window.location.reload and will have to spy on a function that calls to window.location.reload instead.


kberg (approved):
This is nice!

PR #6022, opened 2023-09-24
Nuclear

maserion:
Part9
Another set of cards


kberg:
So I see you pushed this to the wrong repo, but hey, how's this expansion going?

maserion:
Hello,
unfortunately recently I had very little time to work on that. It is more
likely I will start working on that closer to the spring season. If I
recall correctly around 60 projects cards are already implemented (w/o
tests), but yellow cards are not started at all.

Regards,
M

czw., 14 gru 2023 o 20:23 Robert Konigsberg ***@***.***>
napisał(a):

PR #6031, opened 2023-09-24
St. Joseph + Helion: Fix payment with heat

matipau:
The city owner should be able to use heat to pay 2 M€ for 1 card iff she/he is Helion (and not iff the player building the cathedral is Helion).


kberg:
Thanks for this. Unfortunately this is the wrong spot because this will apply to any other card like this. I'll fix it tonight.

PR #6032, opened 2023-09-24
Neptunian Power Consultants + Helion: Allow paying with heat

matipau:
As Helion, it should be possible to pay the 5 M€ using heat when the effect triggers.


kberg (commented):
Thanks for this. Unfortunately this is the wrong spot because this will apply to any other card like this. I'll fix it tonight.

PR #6033, opened 2023-09-25
Restore Helion and Luna Trade Fed when paying for non-cards.

kberg:


kberg:
This is a more robust fix than the proposed #6031 and #6032. But I wouldn't have known about it without those, so thank you.

matipau:
Thanks for looking into it and for the more general fix. While this resolves #6032, I think that the fix for #6031 is incomplete because in [`StJosephOfCupertinoMission.ts`](https://github.com/terraforming-mars/terraforming-mars/blob/bdb9c71a8bc8655085ae7600ebcefcccb58268fc/src/server/cards/promo/StJosephOfCupertinoMission.ts) there is still an explicit
```typescript
{heat: player.canUseHeatAsMegaCredits}
```
passed to the `SelectPayment` constructor, so this overwrites the default `paymentOptions` intoduced in 826f380e4d23fe7fc8ff6cdbf8c182c404939741. Therefore, the bug is still present unfortunately, i.e. it is checked whether the player placing the cathedral is Helion, and not whether the city owner is Helion. One should just remove the heat option in [`StJosephOfCupertinoMission.ts`](https://github.com/terraforming-mars/terraforming-mars/blob/bdb9c71a8bc8655085ae7600ebcefcccb58268fc/src/server/cards/promo/StJosephOfCupertinoMission.ts).

kberg:
The supplied dictionary actually doesn't impact the code below.
```
      paymentOptions: {
        heat: player.canUseHeatAsMegaCredits,
        lunaTradeFederationTitanium: player.canUseTitaniumAsMegacredits,
        ...this.paymentOptions,
      },
```

This doesn't change if this.paymentOptions has the same heat declaration.
You are right though that the right thing is to remove that assertion that
a player can spend heat to draw a card.

matipau:
I'm pretty sure that it did impact the code in `SelectPayment.ts`, since `player` refers (hopefully) to the correct player there, i.e. the one doing the payment, while `player` in `StJosephOfCupertinoMission.ts` is the one performing the card action. If the constructor would have been called with `{heat: spaceOwner.canUseHeatAsMegaCredits}` (as in #6031), then it would indeed not make a difference. But the problem is that `player` in `StJosephOfCupertinoMission.ts` is not necessarily the same as `player` in `SelectPayment.ts`. So the heat payment rule removed in 9b3bc331d1cab7b3e42a608dc2490b9ed7e638aa was not just unnecessary, it was in fact even wrong because it referred to the wrong player (unless the card owner decides to build the cathedral on one of its own cities).
Anyway, thanks for the quick fix and this great project :-)

kberg:
Thanks for your input! I think the final result looks good. I made some
massive changes, and they came with a couple of bumps. Things are
definitely cleaner, though.

PR #6036, opened 2023-09-25
:broom: Less logging during unit test runs

bafolts:
Updated the language switcher component with a function we can stub prior to calling window.location.reload. In jsdom that call to reload was leading to console logging during test runs with an error. While I was in there I removed sinon since we were only using it in the language switcher component and I am not a huge fan of dependencies. There were some other tests logging to console as well which were removed.


kberg (approved):
Nice.

Reminds me of a proverb I just learned about: [A little copying is better than a little dependency.](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=9m28s)

It seems like a proverb you can get behind.

Source: https://go-proverbs.github.io/

PR #6091, opened 2023-10-09
Update log_messages.json

Borbarad2:
Many new log messages from promo, prelude2 and star wars cards.


src/locales/de/log_messages.json
```diff
@@ -353,6 +360,9 @@
     "gain 24 M€": "24 M€ zu erhalten",
     "gain 27 M€": "27 M€ zu erhalten",
     "gain 30 M€": "30 M€ zu erhalten",
+    "gain 33 M€": "33 M€ zu erhalten",
```
kberg, line 363: Which card does this apply to? Because I don't think this will happen anymore.

Borbarad2: It happenend in our last game, that why I extended the range. 
Its the Venus project card "Sulphur-Eating Bacteria" 
grafik

Borbarad2: If you are really sure, it will not happenend and needed anymore, then i will remove the whole block next time i update log_messages. And try it out, if nit is not working, then i would add the block again.

kberg: This change was made 8 months ago. I don't think Sulphur-eating-bacteria is the cause. Can you share the game link?

```
    player.game.log('${0} removed ${1} microbes from ${2} to gain ${3} M€', (b) =>
      b.player(player).number(amount).card(this).number(megaCreditsGained));
```

Borbarad2: Okay, nice, i will change that, and then let's see if still working. Fine.
Thanks.

kberg (approved):

PR #6093, opened 2023-10-09
Update game_end.json

Borbarad2:
Added new feature translation to end screen


alrusdi (approved):

Borbarad2:
@alrusdi: Thanks for approving and merging.
Could it be possible, that you may check following translations, because those does not work.
Custom Polotics and agendas
Awards listed here at the end_screen, if my latest change according to that does not work. 
(If you know it, you could do it for your own language and I would transfer your solution to my Language).

PR #6127, opened 2023-10-17
Modified secondary tags and added new requirements

EthanDobbs:
Secondary Tags were not centered with the edge of the resource, but that is now fixed. Card resources can now had the class "card-resource--has-secondary-tag" to give them a wider right-side margin.


src/common/cards/RequirementType.ts
```diff
@@ -9,6 +9,8 @@ export enum RequirementType {
     TAG = 'tag',
     PRODUCTION = 'production',
     REMOVED_PLANTS = 'Removed plants',
+    SUM_TAGS = 'Sum tags',
```
kberg, line 12: My mistake, these shouldn't be part of the PR since they're not being used in the app

EthanDobbs:
Removed new requirements

kberg (approved):
I made a couple of modifications, but this is

kberg:
... but this is good.

PR #6163, opened 2023-10-31
Make Sagitta work correctly with wild tags. Add test

### LINKED ISSUE #6149: Possible bug related to wild tags and Sagitta
Hello,

Reporting a possible bug related to one of the new corporations, "Sagitta Frontier Services."

When playing a card with a wild tag, the second ability of Sagitta triggers (gain $1) rather than the first ability (gain $4). I think this is incorrect, because wild tags are not active until **after** you play the card (see text below) - for instance, they don't benefit from tag-specific cost reduction effects like Earth Office.

Note that I have only seen this with the wild tag prelude cards; I haven't had the opportunity to play the card Research Coordination to see if it has the same issue, but I would expect it does.

I was not able to find any rules clarification on BGG or anywhere concerning this interaction, although I know that Community Services includes cards with wild tags when counting cards with no tags.

**Sagitta card text:**
Effect: When you play a card with no tags, including this, gain 4 M€.)
Effect: When you play a card with EXACTLY 1 TAG, gain 1 M€.)
You start with 28 M€. Increase energy production 1 step and M€ production 2 steps. Draw a card with no tags.

**Wild tag text** (from Research Network and Research Coordination):
After being played, when you perform an action, the wild tag counts as any tag of your choice.

**Bug Report Dialog:**
  "url": "https://terraforming-mars.herokuapp.com/spectator?id=s509c5766c696",
  "color": "orange",
  "step": 18,
  "version": "18a0713",
  "builtAt": "Thu, 19 Oct 2023 15:09:46 GMT",
  "browser": "Firefox 119",
  "language": "en",
  "experimental_ui": false

kberg:
Fixes #6149


kberg:
I added some notes about this at https://github.com/terraforming-mars/terraforming-mars/wiki/Card-Details#sagitta-frontier-services

Now I wonder whether a card with just a wild tag counts as drawing a card with "no tags"

PR #6188, opened 2023-11-05
[Bugfix] Comet for Venus and Warmonger award

ap3h3ad:
This commit fixes a bug where the Comet for Venus card would not count as attack card for the Warmonger award.


kberg:
Comet For Venus doesn't have a remove plants option.

kberg:
Also, Comet For Venus was already included in Warmonger: 
image

PR #6194, opened 2023-11-06
Update ui.json

Borbarad2:
wrong and misleading translation


src/locales/de/ui.json
```diff
@@ -739,11 +755,47 @@
     "Don't remove M€ from adjacent player": "Entferne keine M€ vom angrenzenden Spieler",
     "Do not remove resource": "Entferne keine Ressourcen",
     "Skip removal": "Überspringe Entfernung",
+    "Remove ${0} M€ from": "Entferne ${0} M€ von",
+    "Remove ${0} M€ from ${1}": "Entferne ${0} M€ von ${1}",
+    "Remove 1 M€ from": "Entferne 1 M€ von",
+    "Remove 2 M€ from": "Entferne 2 M€ von",
```
kberg, line 761: Rather than continuing to add these broad translations, can you tell me what cards triggered these effects?

Borbarad2: All Asteroid cards for plants: 
grafik
grafik

Borbarad2:
Did remove the remove part again, because the Sabotage card was working. You could now merge and push that Update/PR.

kberg:
> Did remove the remove part again, because the Sabotage card was working. You could now merge and push that Update/PR.

There are still plenty of those in here. If there's a basis for those changes, I will make them. But I don't see them.

kberg (approved):

PR #6213, opened 2023-11-09
Japanese Translation

fujiken0201:
Japanese translation is being added. Translations of official goods other than termoiles have been added.
Since only the above translations are sufficient for my playing environment, I would like to merge them in this state.

If the number of Japanese players increases, I will consider adding translations for Turmoil and Fan Expansion.
Thank you for your time and approval.


kberg:
Thank you for this!

Can you please remove the change to `constants.ts`? If you want, I can do that for you.

We seem to be using country codes (e.g. cn) and not language codes (e.g. zh.) So can you please rename the `ja` directory to `jp`?

fujiken0201:
Thanks for the confirmation!
Can I ask you to change the folder name since I'm on the road for a bit? Thank you in advance.

kberg:
> Thanks for the confirmation! Can I ask you to change the folder name since I'm on the road for a bit? Thank you in advance.

Sorry, I can't for a few days.

fujiken0201:
> Sorry, I can't for a few days.

Changed "src/locales/en" to "src/locales/jp".
Please confirm.

kberg (approved):

kberg:
image
Create start!

fujiken0201:
I was relieved that it could be displayed without any problems!  If there is no problem, please merge it to main.

kberg:
Done! Thank you for your hard work

PR #6217, opened 2023-11-10
Text fix magnetic field generators promo

JTBringe:
Card lets you place a special tile, but description does not state that.
Updated description to match that of the official promo card.


kberg:
Is this the official card text?

JTBringe:
I don't have the promo variant myself, but I googled some screenshots and found some pictures of the promo variant with that exact text.
Here's one example: https://boardgamegeek.com/image/6163312/terraforming-mars-big-box

kberg:
Thanks for doing the extra effort, and for your contribution!

PR #6222, opened 2023-11-12
Global events rework

ssimeonoff:
**AC**
- events css rework
- cleaned up multiple linebreaks 
- the event context is displayed in a 275x100px div, vertically aligned

Untitled


kberg:
Looks nice! A couple of cards don't quite fit.

image

We discussed that the name could move to the bottom, but perhaps that's a mistake? I think it'll confuse people. What do you think?

kberg:
I'll do a proper review in the next day or two.

ssimeonoff:
I knew some might not fit perfectly, but with the given height, the best I could do was the vertically align it. Technically we can make it bigger height, but I think we should not because it looks good for 95% of the events. 
I think we should manually fix the remaining 5%.

kberg (approved):

PR #6227, opened 2023-11-15
Update base_cards.json

Borbarad2:
Adapt text, thanks to 🤖  Text repair thanks to @JTBringe 🙏 🎉

(Besides removed unnecassary brackets.)


JTBringe:
Thanks for the mention :)
Keep in mind that it was the promos variant of Magnetic Field Generators that I made changes to, whereas you've changed the base card, which does not place a special tile.

JTBringe:
I'm working on Norwegian translations, and as a starting point I copied all of the german translations, so this good for me to know this as well :D

kberg (approved):

PR #6229, opened 2023-11-15
Turmoil Css updates

EthanDobbs:
Fixed some Issues with the turmoil css


src/styles/cards_v2.less
```diff
@@ -2396,16 +2396,18 @@
 
 .card-party {
   display: flex;
-  position: absolute;
-  transform: scale(1.35);
```
ssimeonoff, line 2400: why would you remove css from the common class only to duplicate it in two other classes?

src/styles/turmoil.less
```diff
@@ -89,22 +89,23 @@
 .global-event {
     display: inline-block;
     vertical-align: top;
-    width: 330px;
-    height: 170px;
+    width: 332px;
```
ssimeonoff, line 92: this breaks the events alignment

src/styles/turmoil.less
```diff
@@ -89,22 +89,23 @@
...
 
 .global-event-card-content {
-    width: 275px !important;
-    height: 100px !important;
+    width: 276px !important;
+    height: 104px !important;
     justify-content: center;
     margin: 0 auto;
+    top: 10px;
+    overflow-x: hidden;
```
ssimeonoff, line 108: this hides some info

ssimeonoff: Untitled

ssimeonoff: Untitled

src/styles/cards_v2.less
EthanDobbs: Yes, because the card-party id is used for turmoil requirements. Adding the css to card-party breaks requirements.

src/styles/turmoil.less
EthanDobbs: Yes it does, oh no

src/styles/turmoil.less
EthanDobbs: I'll see what I can do about election and revolution

src/styles/cards_v2.less
ssimeonoff: oh, yeah

kberg:
What's pending?

ssimeonoff (approved):

PR #6245, opened 2023-11-19
Prelude 2 corp logos

kberg:
The Spire log is awful, would be happy to get someone to do a better job. #hint


kberg:
h/t @nwai90 for the others

PR #6250, opened 2023-11-20
Update turmoil_events.json

Borbarad2:
Adapt to new Turmoil Global Event Design


Borbarad2:
Could be closed, as I add this commit to an other commit

PR #6251, opened 2023-11-20
Update log_messages.json

Borbarad2:
Add missing new resource to the list: graphene


Borbarad2:
Could be closed, because i add that commit to another commit in one PR.

PR #6266, opened 2023-11-27
Global events css polish

ssimeonoff:
**AC**
- css changes - removed the 0.9 scaling, the vertical scrolling and added vertical alignment
The "drawback" is that only 2 events ELECTION and REVOLUTION have iconography on two rows and slightly go off the desired borders
- minor events polish 

events


kberg (approved):
Thank you!

PR #6274, opened 2023-11-30
Fixed a typo with Why? on the game setting screen

JTBringe:
Typo caused i18n to not work properly, now it works as intended.


kberg (approved):
Whoops! Thanks.

PR #6277, opened 2023-11-30
Help iconology translation fix

JTBringe:
Added some divs around some category names
This fixes an issue with i18n which made the translation not show up.

### DIFF
--- a/src/client/components/help/HelpIconology.vue
+++ b/src/client/components/help/HelpIconology.vue
@@ -49,7 +49,6 @@
                 <div class="resource-tag tag-city"></div>
                 <div class="help-icon-label" v-i18n>City</div>
             </div>
-            <br>
             <div>
                 <div class="resource-tag tag-wild"></div>
                 <div class="help-icon-label" v-i18n>Wild</div>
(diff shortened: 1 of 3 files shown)


src/client/components/help/HelpIconology.vue
```diff
@@ -49,7 +49,6 @@
                 <div class="resource-tag tag-city"></div>
                 <div class="help-icon-label" v-i18n>City</div>
             </div>
-            <br>
```
kberg, line 52: Why did you remove this?

JTBringe: Screenshot_20231201_072704_Chrome

I figured the space between City and Wild was not supposed to be there if they all belong to the same category. None of the other categories have this kind of space.

JTBringe: If you want I can put the <br> back :)

kberg: I've been away. Yes, the separation is intentiona.

JTBringe: No worries, I'll put it back later today 👍

kberg:

PR #6283, opened 2023-12-05
Update IcyImpactors.ts

Borbarad2:
Correct a typo on the new promo card: PLACER --> PLAYER


kberg:
Thanks. This is already handled in #6284

PR #6292, opened 2023-12-14
Few minor css changes

ssimeonoff:
**AC**
- EV icon fix - double margin right has been removed and the width is proper now 
(top: current state, bottom: previous state)
Untitled

- "other player cards" fixes:
-- on mobile and desktop the width matches the width of the player boards
-- player colour, label and close button are in a header - the backgroud colour and text colour now matches the colours of the player board
-- background colour is now 222 - identical to the background of the backlog
-- vertical scrolling now longer hides/scrolls away the close button 
Untitled

### DIFF
--- a/src/client/components/OtherPlayer.vue
+++ b/src/client/components/OtherPlayer.vue
@@ -53,32 +53,36 @@ export default Vue.extend({
 });
 </script>
 <template>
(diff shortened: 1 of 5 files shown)


kberg (approved):
Thank you!

PR #6295, opened 2023-12-16
Undo the card-local data.

kberg:
There are some issues. For now, undo is the right choice.


kberg:
This is going to break one generation for existing games with Pristar and UNMI. :/

Issue #6305, opened 2023-12-27
nodejs version required by package.json is EOL


harkovs:
Nodejs 16 is no longer being maintained. The projects package.json file contains an entry for engine requirements:
`"engines": {
        "node": ">=14.x <=16.x"
      },`
Would it be possible to move to node 18?

I've tested running in node 18 by changing the DockerFile. It seems to run albeit somewhat slower.

kberg:
I install v18, which has 3 moderate severity vulnerabilities.

It also fails tests:

```
  1) PlayerInput
       sends 400 on server error:

      AssertionError: expected '{"message":"Unexpected token \'}\', \\"}{\\" is not valid JSON"}' to equal '{"message":"Unexpected token } in JSON at position 0"}'
      + expected - actual

      -{"message":"Unexpected token '}', \"}{\" is not valid JSON"}
      +{"message":"Unexpected token } in JSON at position 0"}
      
      at Suite.<anonymous> (tests/routes/PlayerInput.spec.ts:99:25)
      at Generator.next (<anonymous>)
      at fulfilled (tests/routes/PlayerInput.spec
```

I installed v20, which introduced high severity vulnerabilities. Those might not be a real issue, but the other issue is that tests won't pass:
```
  1) PlayerInput
       sends 400 on server error:

      AssertionError: expected '{"message":"Unexpected token \'}\', \\"}{\\" is not valid JSON"}' to equal '{"message":"Unexpected token } in JSON at position 0"}'
      + expected - actual

      -{"message":"Unexpected token '}', \"}{\" is not valid JSON"}
      +{"message":"Unexpected token } in JSON at position 0"}
      
      at Suite.<anonymous> (tests/routes/PlayerInput.spec.ts:99:25)
      at Generator.next (<anonymous>)
      at fulfilled (tests/routes/PlayerInput.spec.ts:5:58)
```

kberg:
So now that 18 and 20 are available, I think it's safe to close this issue. Even though 16 is EOL, it's worth keeping around so users aren't jostled around quite so much.

kberg:
Saw your thumb, @harkovs. Closing.

PR #6307, opened 2023-12-28
Disable Node 14, enable Node 18 and Node 20

kberg:
#6305 points out that 16 is about to EOL. Adding 18 and 20, and removing 14, which is farther out of support than 16. Looks like 18 and 20 should work.


bafolts:
This makes sense to me. Only thing I would consider is if it impacts heroku
build in some way.

PR #6315, opened 2023-12-31
Create new Deck.draw that returns undefined when draw pile is empty

lukenofurther:
Renames existing `draw` method to `drawLegacy` while we switch references over to the new `draw` in stages.

The new `draw` returns undefined when the draw pile is empty, rather than throwing an error that's generally not handled properly.

~~Switches `AsteroidDeflectionSystem` and `SearchForLife` over to the new `draw` with some light refactoring to tidy them up a bit.~~

### DIFF
--- a/src/server/cards/Deck.ts
+++ b/src/server/cards/Deck.ts
@@ -57,6 +57,22 @@ export class Deck<T extends ICard> {
     }
   }
 
+  public draw(logger: Logger, source: 'top' | 'bottom' = 'top'): T | undefined {
+    if (this.drawPile.length === 0) {
+      logger.log(`The ${this.type} discard pile has been shuffled to form a new deck.`);
+      this.shuffle();
+    }
+
+    const card = source === 'top' ? this.drawPile.pop() : this.drawPile.shift();
+
+    if (this.drawPile.length === 0) {
+      logger.log(`The ${this.type} discard pile has been shuffled to form a new deck.`);
+      this.shuffle();
+    }
+
+    return card;
+  }
+
   public drawLegacy(logger: Logger, source: 'top' | 'bottom' = 'top'): T {
     return this.drawOrThrow(logger, source);
   }
--- a/src/server/cards/Deck.ts
+++ b/src/server/cards/Deck.ts
@@ -57,7 +57,11 @@ export class Deck<T extends ICard> {
     }
   }
 
-  public draw(logger: Logger, source: 'top' | 'bottom' = 'top'): T {
+  public drawLegacy(logger: Logger, source: 'top' | 'bottom' = 'top'): T {
+    return this.drawOrThrow(logger, source);
+  }
+
+  public drawOrThrow(logger: Logger, source: 'top' | 'bottom' = 'top'): T {
     const card = source === 'top' ? this.drawPile.pop() : this.drawPile.shift();
 
     if (card === undefined) {
@@ -81,7 +85,7 @@ export class Deck<T extends ICard> {
         logger.log(`discarded every ${this.type} card without a match`);
         break;
       }
-      const projectCard = this.draw(logger);
+      const projectCard = this.drawLegacy(logger);
       if (include(projectCard)) {
         result.push(projectCard);
       } else {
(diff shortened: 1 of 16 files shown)


tests/cards/Deck.spec.ts
```diff
@@ -111,3 +114,118 @@ describe('CeoDeck', function() {
     expect(deck.discardPile).to.deep.eq(deserialized.discardPile);
   });
 });
+
+describe('draw()', function() {
+  let deck: ProjectDeck;
```
kberg, line 119: THANK YOU

tests/cards/Deck.spec.ts
```diff
@@ -1,10 +1,13 @@
 import {expect} from 'chai';
-import {PreludeDeck, CeoDeck} from '../../src/server/cards/Deck';
+import {PreludeDeck, CeoDeck, ProjectDeck} from '../../src/server/cards/Deck';
 import {GameCards} from '../../src/server/GameCards';
 import {DEFAULT_GAME_OPTIONS} from '../../src/server/game/GameOptions';
 import {CardName} from '../../src/common/cards/CardName';
 import {ConstRandom, UnseededRandom} from '../../src/common/utils/Random';
 import {ICard} from '../../src/server/cards/ICard';
+import { Game } from '@/server/Game';
```
kberg, line 8: Replace `@` with `../../../src`, please. I love and hate the `@`

Also, the linter is going to complain about the spaces in `{ Game }`

src/server/cards/base/SearchForLife.ts
```diff
@@ -49,17 +49,44 @@ export class SearchForLife extends Card implements IActionCard, IProjectCard {
...
-          player.game.log('${0} found life!', (b) => b.player(player));
-        }
-
-        player.game.projectDeck.discard(topCard);
-      });
+      .andThen(() => this.revealCard(player));
 
     return undefined;
   }
+
```
kberg, line 56: So, one of the reasons I thought it would be good to address these one at a time is because it's actually not clear whether drawing a card is a mandatory effect, or an optional one, and so each card might be handled differently.

I don't think it's clear. I did a couple of searches in TfM rules on bgg, and referred to the official FAQ and unofficial FAQ. (https://github.com/terraforming-mars/faq/wiki#what-cardaction-effects-or-requirements-are-mandatory-and-which-are-not)

This may be pedantic, but would you mind removing the changes to SFL and ADS? OR keep them, and we can have a discussion about it. It'll just slow down moving this PR in.

kberg (commented):
This looks good. Couple of things to resolve.

kberg:
Yep, linter is sad:

image

kberg:
I've raised the question on bgg https://boardgamegeek.com/thread/3217206/empty-deck-discard-pile-during-action-phase and discord https://discord.com/channels/737945098695999559/742826825922904225/1191049266509119508

src/server/cards/base/SearchForLife.ts
lukenofurther: Totally makes sense - I'll remove those commits and put them on branches that can be merged later if necessary

src/server/cards/Deck.ts
```diff
@@ -57,7 +57,27 @@ export class Deck<T extends ICard> {
...
+    }
+
+    return card;
+  }
+
+  public drawLegacy(logger: Logger, source: 'top' | 'bottom' = 'top'): T {
+    return this.drawWithError(logger, source);
+  }
+
+  public drawWithError(logger: Logger, source: 'top' | 'bottom' = 'top'): T {
```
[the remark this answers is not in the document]
lukenofurther: Happy to amend and stick with the existing convention

tests/cards/Deck.spec.ts
lukenofurther: My VS Code added the `@` automatically and I forgot to replace, or lint my changes! I know what you mean about the root alias. I've used named aliases before but it's probably better to use more explicit relative imports. Anyway, have replaced with relative imports.

kberg:
Wow, thanks again! Not much time today, so this might not get attention until tomorrow. If you're on the discord, that might be an easy way to discuss details.

kberg (approved):
This is a good change. I haven't read the tests too carefully, but they look good. Thank you!

PR #6320, opened 2024-01-01
Reflect modified global requirements limits in the UI

### LINKED ISSUE #6302: Modify displayed requirements when Inventrix / Adaptation Technology are in play
In the same way that cost discounts are displayed on cards:

Screenshot 2023-12-23 at 20 27 07

If Inventrix or Adaptation Technology is in play, it would be great to show that global requirements limits are also modified, rather than needing to calculate the difference myself.

Currently, just the "for everyone" global requirement is shown not factoring in that one player may have different requirements.

illicitonion:
There are a couple of major questions/issues here:
1. Should we implement this generally like this (passing the
   requirements modifiers from the server to the client, and asking the
   client to compute the applicable modifications), or should the server
   pre-compute the relevant modifications and tell the client about
   them?
2. Assuming we pass the requirement modifications to the client, how
   should we do so? This is currently doing so with an ugly
   `Array<RequirementModifierEntry>` because `Map` doesn't
   JSON-serialize. Is this fine? Should we try to force `Map` to JSON
   serialize? Is there a different `Map`-like type we should be using
   which is more JSON friendly?

Fixes #6302

### DIFF
--- a/src/common/cards/RequirementType.ts
+++ b/src/common/cards/RequirementType.ts
@@ -1,3 +1,6 @@
+import {match} from 'ts-pattern';
+import {GlobalParameter} from '../GlobalParameter';
+
 export enum RequirementType {
     OXYGEN = 'O2',
     TEMPERATURE = 'C',
@@ -31,4 +34,35 @@ export enum RequirementType {
     // Underworld
     EXCAVATION = 'Excavation',
     CORRUPTION = 'Corruption',
+}
+
+export function requirementTypeFromGlobalParameter(globalParameter: GlobalParameter): RequirementType {
+  return match(globalParameter)
+    .with(GlobalParameter.OCEANS, () => RequirementType.OCEANS)
+    .with(GlobalParameter.OXYGEN, () => RequirementType.OXYGEN)
+    .with(GlobalParameter.TEMPERATURE, () => RequirementType.TEMPERATURE)
+    .with(GlobalParameter.VENUS, () => RequirementType.VENUS)
+    .with(GlobalParameter.MOON_HABITAT_RATE, () => RequirementType.HABITAT_RATE)
+    .with(GlobalParameter.MOON_MINING_RATE, () => RequirementType.MINING_RATE)
+    .with(GlobalParameter.MOON_LOGISTICS_RATE, () => RequirementType.LOGISTIC_RATE)
+    .exhaustive();
+}
+
+export function globalParameterForRequirementType(requirementType: RequirementType): GlobalParameter | null {
+  switch (requirementType) {
+  case RequirementType.TEMPERATURE:
+    return GlobalParameter.TEMPERATURE;
+  case RequirementType.OXYGEN:
+    return GlobalParameter.OXYGEN;
+  case RequirementType.VENUS:
+    return GlobalParameter.VENUS;
+  case RequirementType.HABITAT_RATE:
+    return GlobalParameter.MOON_HABITAT_RATE;
+  case RequirementType.MINING_RATE:
+    return GlobalParameter.MOON_MINING_RATE;
+  case RequirementType.LOGISTIC_RATE:
+    return GlobalParameter.MOON_LOGISTICS_RATE;
+  default:
+    return null;
+  }
 }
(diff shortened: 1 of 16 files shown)


kberg:
Hey, thanks!

> There are a couple of major questions/issues here:
> 
> 1. Should we implement this generally like this (passing the
>    requirements modifiers from the server to the client, and asking the
>    client to compute the applicable modifications), or should the server
>    pre-compute the relevant modifications and tell the client about
>    them?

Client and server representations of cards are sufficiently different, and in fact, no behavior is passed down to the client. I don't think there's enough information to reasonably compute this on the client, and if there was, that seems like duplicat ed behavior.

> 2. Assuming we pass the requirement modifications to the client, how
>    should we do so? This is currently doing so with an ugly
>    `Array<RequirementModifierEntry>` because `Map` doesn't
>    JSON-serialize. Is this fine? Should we try to force `Map` to JSON
>    serialize? Is there a different `Map`-like type we should be using
>    which is more JSON friendly?

Unsure, I will look.

> Fixes #6302

src/common/cards/RequirementType.ts
```diff
@@ -32,3 +35,34 @@ export enum RequirementType {
     EXCAVATION = 'Excavation',
     CORRUPTION = 'Corruption',
 }
+
+export function requirementTypeFromGlobalParameter(globalParameter: GlobalParameter): RequirementType {
```
kberg, line 39: I think these two functions can be replaced with a mnemonist/BiMap.

e.g.

```
export const mapping: BiMap<GlobalParameter, RequirementType> = BiMap.from({
  [GlobalParameter.OCEANS]: RequirementType.OCEANS,
  [GlobalParameter.OXYGEN]: RequirementType.OXYGEN,
  [GlobalParameter.TEMPERATURE]: RequirementType.TEMPERATURE,
  [GlobalParameter.VENUS]: RequirementType.VENUS,
  [GlobalParameter.MOON_HABITAT_RATE]: RequirementType.HABITAT_RATE,
  [GlobalParameter.MOON_MINING_RATE]: RequirementType.MINING_RATE,
  [GlobalParameter.MOON_LOGISTICS_RATE]: RequirementType.LOGISTIC_RATE,
});

mapping.get(GlobalParameter.OCEANS);
mapping.inverse.get(RequirementType.OCEANS);
```

I admit never using a BiMap, so I don't know it's edge cases (for instance, what happens if you call `mapping.inverse.get(RequirementType.TAG)` but that's also solvable.

src/common/constants.ts
```diff
@@ -1,3 +1,5 @@
+import {GlobalParameter, getGlobalParameterLimits} from './GlobalParameter';
```
kberg, line 1: I would like constants to have no imports.  It's a top-level definition. IOW, keep the constants here, and use them to define your new data structures.

kberg (commented):
I've only done a _partial cursory look_, sorry you deserve more time, but people are about to arrive. (I should be cleaning instead of reviewing  code. But I love one more than the other. :D) Anyway, more to come. Thank you!

kberg:
ps `npm run build:tests` will help avoid things like this

image

kberg:
OK now I'm really taking care of local things. Thanks again. I'll look at this again no later than tomorrow.

kberg:
I lied. Will look in no later than the weekend. That's a reliable commitment.

src/common/models/CardModel.ts
```diff
@@ -3,14 +3,23 @@ import {Units} from '../Units';
...
+
 export interface CardModel {
     name: CardName;
     resources: number | undefined;
     calculatedCost?: number;
     isSelfReplicatingRobotsCard?: boolean,
     discount?: Array<CardDiscount>,
+    // TODO: Ideally this would be a Map<RequirementType, number> but Maps don't JSON serialize.
+    // We should probably polyfill a toJSON and fromJSON into Map rather than storing this as an Array.
+    requirementsModifiers: Array<RequirementModifierEntry>,
```
kberg, line 22: Oh, I see. This can be simpler by replacing `Array<RME>` with `Partial<Record<RM, number>>`

kberg (changes requested):
This is a good change. Let's get the types sorted, etc, and I'll run it locally.

kberg:
Hi @illicitonion, any thoughts?

kberg:
I'm closing this just to clear the queue. But if you respond here, I'll see it and reopen it.

kberg:
Sounds great. I'll be around.

kberg:
Well, just thought I'd check in. This is still a good thing to make happen.

kberg:
Hi, are you still interested in finishing this?

PR #6336, opened 2024-01-04
Side by side game view to minimize scrolling

kevinmachstudio:
## Description: 

- Related to https://github.com/terraforming-mars/terraforming-mars/issues/6335
- With these changes, the map and other player's boards can be viewed side by side with the game log / actions / cards in hand

image


kberg (changes requested):
It looks like the file's been reformatted, which means I cannot tell what changes were meaningful, and which were purely syntactic.

Most important: we have a variety of users, some of whom don't have wide screens, many of whom are on phones. (One person apparently love to play in the bath.) How will this work with smaller screens like that? I'll be able to more consider this more when the syntax part is addressed.

kberg:
Thanks for this! I hope we can make it work out.

kevinmachstudio:
Thanks for the quick feedback @kberg! That's a great point. I'll play around with the screen sizes and see what I come up with to support all users.

Borbarad2:
And please, consider, check and lock at how it looks like if Turmoil is on or off, if Moon is on or off, and/or if Pathfinder is on or off. Because the side you are use right now is reserved for those (especially Turmoil) if they are enable.

kberg:
Hi @kevinmachstudio - any thoughts here?

kevinmachstudio:
Hi @kberg @Borbarad2 Sorry I haven't gotten much time to work on this but i'm curious to hear what you think about this:

For users who don't have wide screens (in which a side by side view would be useful), a keyboard shortcut may be a better option to make it easier to access the map and boards, something like CMD-M / CTRL-B, and then it could appear as a slideout with the background tinted. User could just click to exit this view

image

image

- with this approach, web users on 13" screens would not have to constantly scroll up and down or click on the map/cards icons on the left side
- for tablet users without a keyboard they may have to stick with scrolling or finger pressing on the the map/cards icons on the left side
- for mobile users i'm not sure if the experience can be improved (i don't play on mobile as it's too small)

Having this shortcut option could be a separate PR but better option for a bigger number of users

kberg:
People on phones will not have wide screens, and so won't have keyboard shortcuts.

kevinmachstudio:
@kberg Yes it wouldn't be for mobile users. The shortcut would be for users with keyboards

kberg:
I'm closing this just to clear the queue. Plus, this is moving from a code review to a design discussion, which we can have in discord or on an issue. But if you respond here, I'll see it and reopen it.

PR #6338, opened 2024-01-04
Fix canAct behaviour in AquiferPumping

s-alexey:
Resolved an issue where canAct incorrectly returned true even when all ocean tiles were already placed.

### DIFF
--- a/src/server/cards/base/AquiferPumping.ts
+++ b/src/server/cards/base/AquiferPumping.ts
@@ -29,7 +29,7 @@ export class AquiferPumping extends Card implements IActionCard, IProjectCard {
   }
 
   public canAct(player: IPlayer): boolean {
-    return player.canAfford({cost: OCEAN_COST, steel: true, tr: {oceans: 1}});
+    return player.game.canAddOcean() && player.canAfford({cost: OCEAN_COST, steel: true, tr: {oceans: 1}});
   }
   public action(player: IPlayer) {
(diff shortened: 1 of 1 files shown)


kberg:
Thanks for the submission. One of our tests thankfully caught what is an appropriate issue:

```
  1) AquiferPumping
       Can act if can pay even after oceans are maxed:

      AssertionError: expected false to be true
      + expected - actual

      -false
      +true
      
      at Context.<anonymous> (tests/cards/base/AquiferPumping.spec.ts:91:35)
      at processImmediate (node:internal/timers:476:21)
```

This is saying that it's valid to use the action, even though it doesn't place an ocean. Some details about the rule can be seen here: https://github.com/terraforming-mars/faq/wiki#can-you-perform-an-action-that-raises-a-global-parameter-that-has-already-been-maxed-out

These are known as _stall actions_, ways to take a single action that has no real effects, but makes it possible for you to wait out other players so you can execute a plan that requires several repeated steps.

kberg:
For example, I started toying with what is a legal move: spending 8 heat even though temperature has reached its maximum, as a stall action: https://github.com/terraforming-mars/terraforming-mars/pull/5836 I stopped working on it, but it could be in there.

The appropriate fix, I believe, is not the approach you are taking, but providing a warning to the player that the action won't have the expected effect.

Closing this submission for now, but you are welcome to respond in this issue and I'll read the comments. Thank you, once again.

PR #6357, opened 2024-01-09
Add property to exclude card from draw pile based on chosen board

kevinmachstudio:
Noctis City card can only be plased on Tharsis, so for any other boards we should exclude it from the draw pile

### DIFF
--- a/src/server/Game.ts
+++ b/src/server/Game.ts
@@ -222,7 +222,7 @@ export class Game implements IGame, Logger {
     const board = GameSetup.newBoard(gameOptions, rng);
     const gameCards = new GameCards(gameOptions);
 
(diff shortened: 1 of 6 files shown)


Borbarad2:
No, I would be against that PR/Change. Because Noctis City can be placed on the other boards as well, because there the restriction for the reserved space is lifted and you can place Noctis City where ever you want.

d-little:
Per the Hellas ruleset, this is working as intended:

> The Hellas map lacks volcanos and the Noctis region, so the tiles from the cards Noctis City and Lava Flows lose their placement restrictions and may be placed on any non-ocean area.

kevinmachstudio:
@Borbarad2 @d-little thanks for the clarification, I was not aware of this rule, will close this PR

Issue #6409, opened 2024-02-12
"Advertising" promo card not increasing M€ production as expected when playing a 20+M€ card.


Oliv-zibuth:
Please include:
1. "Advertising" promo card does not increase M€ production as expected when playing a 20+M€ card.

2. Play "Imigrant cities" then "Avertising" then "Field-capped City"
Expected behavior : M€ production should increase twice (once for "Advertising" after playing the card and once for "Immigrant Cities" after placing the city tile on Mars)
Observed behavior : M€ production increased only once, after placing the city on Mars
Same behavior observed for : "Capital"

Then play "Sub-crust measurements".
Expected behavior : M€ production should increase after playing the card
Observed behavior : M€ production did not increase
Same behavior observed for : "Earth Catapult"; "Imported Nitrogen"; Big Asteroid"; 

3. Screenshots
a)Playing "Immigrant city" and "Advertisting"
![image](https://github.com/terraforming-mars/terraforming-mars/assets/95378406/aedb62a4-df56-490b-8f49-ac060a1b9637)
![image](https://github.com/terraforming-mars/terraforming-mars/assets/95378406/e0f7cdad-b321-43e9-bcd0-5c01975b431b)

b) For "Field-Capped city"
![image](https://github.com/terraforming-mars/terraforming-mars/assets/95378406/4bdc4645-02f6-4138-aad0-375c3a9d849e)

c) For "Sub-crust measurements"
![image](https://github.com/terraforming-mars/terraforming-mars/assets/95378406/a799b20d-c3bb-4b18-8abb-470a271a8277)

d)For "Earth Catapult"
![image](https://github.com/terraforming-mars/terraforming-mars/assets/95378406/53717419-d0db-42fc-8f77-13e085f0d69f)

e) For "Capital"
![image](https://github.com/terraforming-mars/terraforming-mars/assets/95378406/df83cabe-0e8a-4928-adc1-7e03c449173c)

4. Any information from  the Bug Report Dialog
{
  "url": "http://terraforming-mars.herokuapp.com/spectator?id=s695384ecb234",
  "color": "red",
  "step": 137,
  "version": "b584fef",
  "builtAt": "Sun, 11 Feb 2024 10:10:55 GMT",
  "browser": "Firefox 122",
  "language": "en",
  "experimental_ui": false
}

Borbarad2:
[...]

Issue #6408, opened 2024-02-12
Media group not providing 3M€ as expected when playing an event card.


Oliv-zibuth:
Please include:
1. Media group card does not provide 3M€ as expected when playing an event card.

2. Play Media group card then Permafrost extraction.
Expected behavior : The player should receive 3M€ after playing Permafrost extraction.
Observed behavior : The player does not receive 3M€ after playing permafrost extraction.
Also observed after playing Big Asteroid.
Maybe linked to issue : Optimal Aerobraking was played just before Media Group (Third screenshot) and was working correctly after Big Asteroid.

3. Screenshots : 
![image](https://github.com/terraforming-mars/terraforming-mars/assets/95378406/a0bfada4-4e94-42fb-b98d-dc5525840966)
![image](https://github.com/terraforming-mars/terraforming-mars/assets/95378406/4a8ad1c4-1d45-47da-9cfd-bb43686e0f47)
![image](https://github.com/terraforming-mars/terraforming-mars/assets/95378406/7c5efd0a-d671-433f-89cf-b0de86bfbe9f)

4. Any information from  the Bug Report Dialog
{
  "url": "http://terraforming-mars.herokuapp.com/spectator?id=s695384ecb234",
  "color": "red",
  "step": 137,
  "version": "b584fef",
  "builtAt": "Sun, 11 Feb 2024 10:10:55 GMT",
  "browser": "Firefox 122",
  "language": "en",
  "experimental_ui": false
}

Borbarad2:
Media Gruop is working perfectly. The M€ amount you get from Media Group is not logged, that's why you might have not notice it. 
And yes, it should be logged to avoid that kind of bug messages.

Issue #6405, opened 2024-02-10
Show that Mons Insurance was played prior to the MC being drained


kberg:
![image](https://github.com/terraforming-mars/terraforming-mars/assets/413481/baf8b2c0-c47a-4a07-a53c-e5ab67e2659a)

PR #6424, opened 2024-02-20
3 minor fixes

kberg:


kberg:
Fixes #6405 
Fixes #6408 
Fixes #6409

PR #6515, opened 2024-03-28
Fixes parameter mismatch in translation

Unihedro:
Occured in game: "BUG: Unparseable message: ${0}がカード${2}枚を売却, ((2) pink, (1) 5)"

This patch may or may not fix the bug


kberg:
Thank you for the contribution information

src/locales/jp/log_messages.json
```diff
@@ -134,7 +134,7 @@
     "${0} converted plants into a greenery": "${0}が植物を支払い、緑地タイルを配置",
     "${0} traded with ${1}": "${0}が${1}と交易",
     "${0} used ${1} standard project": "${0}が標準プロジェクト：${1}を実行",
-    "${0} used ${1} standard project,sold ${2} card(s)": "${0}がカード${2}枚を売却",
+    "${0} used ${1} standard project,sold ${2} card(s)": "${0}が${1}によりカード${2}枚を売却",
```
kberg, line 137: Actually, this line isn't the one that needs to change. See line 175.

kberg (changes requested):

src/locales/jp/log_messages.json
kberg: I'm going to delete line 137 entirely in #6516, so can you either delete the line, or restore the original behavior?

kberg: It wouldn't pollute the main branch; I'd do a squash commit.

kberg (approved):

PR #6535, opened 2024-03-29
Add reds tests for some of the cards.

kberg:
These cards may have their card.tr field removed, so these tests are good, but also will verify that the refactoring is safe. Already, some of the cards aren't actually working correctly after removing card.tr, so there's some research to do.


kberg:
I'm going to keep adding tests for a while, so perhaps don't review right now? Your choice.

kberg:
I can write several PRs after all

PR #6552, opened 2024-04-03
DeferredActionsQueue Refactoring

EthanDobbs:


src/server/Game.ts
```diff
@@ -463,7 +463,7 @@ export class Game implements IGame, Logger {
     return ids.map((id) => this.getPlayerById(id));
   }
 
-  public defer<T>(action: DeferredAction<T>, priority?: Priority): AndThen<T> {
+  public defer<T>(action: DeferredAction<T>, priority?: Priority): DeferredAction<T> {
```
kberg, line 466: Help me understand why this is better.

src/server/deferredActions/DeferredActionsQueue.ts
```diff
@@ -3,98 +3,78 @@ import {GiveColonyBonus} from './GiveColonyBonus';
...
 
 export class DeferredActionsQueue {
-  private insertId: number = 0;
+  /** Array of Deferred Actions sorted in order of when to be executed */
   private queue: Array<IDeferredAction<any>> = [];
 
-  get length(): number {
-    return this.queue.length;
+  /** Gets if the Queue is empty */
+  public get IsEmpty(): boolean {
```
kberg, line 10: Name as `isEmpty` please, and keep it as a function instead of a getter, please.

src/server/deferredActions/DeferredActionsQueue.ts
```diff
@@ -3,98 +3,78 @@ import {GiveColonyBonus} from './GiveColonyBonus';
...
+    return this.queue.length === 0;
   }
 
+  /** Adds a deferred action to the queue, then sorts the queue. */
```
kberg, line 14: No need to describe that it sorts the queue, since that is an implementation detail. The queue delivers messages in order.

src/server/deferredActions/DeferredAction.ts
```diff
@@ -2,30 +2,31 @@ import {IPlayer} from '../IPlayer';
...
     public priority: Priority = Priority.DEFAULT,
   ) {}
 
+  /** Execute the deferred action and possibly return a player input */
   public abstract execute(): PlayerInput | undefined;
-  // TODO(kberg): Make protected again.
-  public cb: (param: T) => PlayerInput | undefined | void = () => {};
-  private callbackSet = false;
 
+  /** Executed a callback function with a parameter determined by the child class of DeferredAction */
```
kberg, line 29: Thanks for adding a description. This description talks about how something is executed. I suggest instead:

> Stores `cb`, to be called after this deferred action completes.

src/server/deferredActions/DeferredActionsQueue.ts
```diff
@@ -3,98 +3,78 @@ import {GiveColonyBonus} from './GiveColonyBonus';
...
-  // The following methods are used in tests
-  public peek(): IDeferredAction<any> | undefined {
-    return this.queue[this.nextItemIndex()];
-  }
-
-  public pop(): IDeferredAction<any> | undefined {
-    return this.queue.splice(this.nextItemIndex(), 1)[0];
   }
 
+  // This function should be moved to the DeferredAction class, so GiveColonyBonus can override the function, instead using of this hook
```
kberg, line 45: Prefix with TODO(ethandobbs): please

src/server/deferredActions/DeferredActionsQueue.ts
```diff
@@ -3,98 +3,78 @@ import {GiveColonyBonus} from './GiveColonyBonus';
...
+  /** Resolves a single action. */
   public run(action: IDeferredAction, cb: () => void): void {
-    // Special hook for trade bonus deferred actions
-    // So that they happen for all players at the same time
+    // GiveColonyBonus hook so that all players can give a response at the same time.
```
kberg, line 48: The original comment was just fine, and contained more information.

src/server/deferredActions/DeferredActionsQueue.ts
```diff
@@ -3,98 +3,78 @@ import {GiveColonyBonus} from './GiveColonyBonus';
...
     if (action instanceof GiveColonyBonus) {
       action.andThen(cb);
       action.execute();
       return;
     }
-
     const input = action.execute();
-    if (input !== undefined) {
+    if (input) {
```
kberg, line 55: the original was fine, thanks.

src/server/deferredActions/DeferredActionsQueue.ts
```diff
@@ -3,98 +3,78 @@ import {GiveColonyBonus} from './GiveColonyBonus';
...
       action.player.setWaitingFor(input, cb);
     } else {
       cb();
     }
   }
 
+  // The following methods are used in tests
```
kberg, line 62: I'd rather not add functions just for test unless the behavior is really encapsulated.

kberg (changes requested):
It's still hard for me to tell what is going on. There's a series of behavioral refactorings, some semantic ones, and some seem to be stylistic preferences. I can't tell the motivations behind a few of them. Deferred actions, and their ordering, is a pretty critical part of the system so I want to be careful here.

src/server/Game.ts
EthanDobbs: There is no direct benefit, other than I had trouble understanding what was going on and this change made more sense to me. I can understand though why the interface existed tho. If you insisted on re-adding it, I would add comments as to why it exists.

src/server/deferredActions/DeferredActionsQueue.ts
EthanDobbs: Yeah this change was result of wanting to delete the hook, then deciding that was a job for a later time.

src/server/deferredActions/DeferredActionsQueue.ts
EthanDobbs: Yeah, it's an aesthetic preference of mine. I'll opt for 'if (x !== undefined)' instead of 'if (x)' in cases like this where readability is more important than being concise (so, nearly every case).

src/server/Game.ts
kberg: By providing a more narrow API, it informs callers that its use is to only attach an andThen.

kberg:
Feel free to reopen this and/or create a new PR when you are ready.

PR #6553, opened 2024-04-03
Addition of Japanese translation

nasubi-land:
(DeepL)
Addition of Japanese translation, focusing on colonies and corporations.
The Japanese translation is based on the official one, but some of the translations are  original.
More translations will be added in the future.

(原文)
コロニーと企業を中心に日本語訳の追加。
公式を基準に日本語訳していますが、一部独自の翻訳になっています。
今後も翻訳を追加していく予定です。


nasubi-land:
The translation of "Venus" in Colonies applies to "UI.json" instead of "colonies.json". What should I do?

Image Word: "金星タグ1つ以上"
Expected Word: "ヴィーナス"

nasubi-land:
After "Venus" is "2 Venus", "3 Venus"... So it seems to be an area that translates the number of tags.
It is actually translated as "タグ" (meaning tag).
I am not sure where this translation is used, but it might be better to translate only "Venus" as "金星" or "ヴィーナス".
Also, the file name was "UI_cards.json", not "UI.json".

Issue #6509, opened 2024-03-26
Mining Guild doesn't work on the moon 


Titozzz:
Mining Guild passive doesn't active for placement on the moon.
It might be too strong if it worked but it should be specified that it doesn't as it's unexpected and plan breaking 😂 
Can make a PR if needed to document the card

Debug data:
```
{
  "url": "https://terraforming-mars.herokuapp.com/spectator?id=s8436c865b526",
  "color": "yellow",
  "step": 22,
  "version": "8cb9a67",
  "builtAt": "Wed, 20 Mar 2024 15:41:02 GMT",
  "browser": "Chrome 123",
  "language": "fr",
  "experimental_ui": true
}
```

Borbarad2:
There is a rule for it:

Compatibility issues

    Philares and Mining Guild do not trigger when tiles are placed on the moon. This is an intentional decision to prevent them from being overpowered.

https://github.com/terraforming-mars/terraforming-mars/wiki/The-Moon#compatibility-issues

Titozzz:
Alright, should I add a comment on the card text ? Or I can just close that issue

kberg:
I am very reluctant to add text to official cards. I'm thinking of an alternative.

Borbarad2:
I think, just close the issue.

Titozzz:
Feel free to close. Happy to help if needed.

kberg:
Like I said, I think I have a solution, but it will take a while.

PR #6558, opened 2024-04-04
Updating the help icon viz & behavior

kberg:
image

1. Shows on hover
2. Smaller
3. Move out of experimental
4. Add more cards
5. Move behavior for deciding whether to include the icon to the cards themselves.


kberg:
This fixes #6509

PR #6561, opened 2024-04-06
Map Vastitas Borealis Novus

kberg:
image


src/client/components/Board.vue
EthanDobbs, line 1: I wish there was a better way to go about these labels, perhaps auto-generating them based on map data? This is fine for now though.

EthanDobbs (approved):

PR #6562, opened 2024-04-06
Addition of Japanese translation

nasubi-land:
Addition of Japanese translation, focusing on Venus, Communities and CEOs.


kberg (approved):
This is a good contribution. Thank you!

PR #6574, opened 2024-04-08
An attempt to remove dependencies of Self-Replicating Robots

EthanDobbs:
This is not quite ready to merge, as you currently can't see what cards are attached to an opponent's instance of SRR without digging through the game logs. This PR was opened so others can suggest changes.

SRR appears in many places of the code where it shouldn't. If I want to modify what happens when you play a card, I shouldn't have to worry about breaking SRR. The goal of this PR is to improve maintainability in regards to the many dependencies of SRR by relocating relevant code to the SRR file.

This PR will change card serialization, so implementing this code on your server will break existing games.


src/client/components/OtherPlayer.vue
```diff
@@ -21,14 +21,6 @@
                 <stacked-cards :cards="getCardsByType(player.tableau, [CardType.EVENT])" :player="player"></stacked-cards>
             </div>
         </div>
-        <div v-if="player.selfReplicatingRobotsCards.length > 0" class="player_home_block">
```
kberg, line 24: Why is this better? SRR cards should really be visually separate from other cards.

src/server/cards/promo/SelfReplicatingRobots.ts
```diff
@@ -8,18 +8,22 @@ import {SelectCard} from '../../inputs/SelectCard';
...
+import {CardModel} from '../../../common/models/CardModel';
+import {newProjectCard} from '../../../server/createCard';
 
 export class SelfReplicatingRobots extends Card implements IProjectCard {
   constructor() {
     super({
       type: CardType.ACTIVE,
       name: CardName.SELF_REPLICATING_ROBOTS,
       cost: 7,
+      resourceType: CardResource.RESOURCE_CUBE,
```
kberg, line 26: This will make SRR susceptible to cards like CEO's favorite project, etc.

kberg:
What's the fundamental mechanical change? Can you describe it? It will help guide me through the review.

(I bet there's a way to preserve existing games, let's see.)

src/client/components/OtherPlayer.vue
EthanDobbs: This is part of the PR that isn't finalized yet and why I changed it to a draft.

src/server/cards/promo/SelfReplicatingRobots.ts
EthanDobbs: True, this is intentional.

PR #6575, opened 2024-04-08
Simplify the testRedsCosts API in advance of simplifying purchase code.

kberg:


EthanDobbs (approved):
I should in general familiarize myself more with the tests, especially in regards to my branch of Think Tank refactoring.

PR #6582, opened 2024-04-10
Better SRR refactoring

EthanDobbs:
This new branch removes the 'RobotCard' interface and achieves better abstraction through the use of 'RobotProxyCard' when selecting cards. The serialized value on Self-Replicating Robots have been substituted for the data property that cards have. Cards attached to SRR are no longer separate from the player's hand.


kberg:
Feel free to reopen this and/or create a new PR when you are ready.

PR #6585, opened 2024-04-12
Addition of Japanese translation 3

nasubi-land:
Addition of Japanese translation, focusing on Ares, Turmoil and Agendas.

More accurate translation.
greenery：緑地
greenery tile(s)：緑地タイル
city：都市
city tile(s)：都市タイル


kberg (approved):

kberg:
Thank you!

PR #6586, opened 2024-04-13
Fix the included cards functionality by only adding them to project c…

rvangrinsven:
…ards

### DIFF
--- a/src/client/components/create/CardsFilter.vue
+++ b/src/client/components/create/CardsFilter.vue
@@ -38,7 +38,6 @@ const allItems: Array<CardName> = [
   ...getCards(byType(CardType.AUTOMATED)),
   ...getCards(byType(CardType.ACTIVE)),
   ...getCards(byType(CardType.EVENT)),
-  ...getCards(byType(CardType.PRELUDE)),
   ...getCards(byType(CardType.CEO)),
 ].map(toName)
   .sort((a, b) => a.localeCompare(b));
--- a/src/server/GameCards.ts
+++ b/src/server/GameCards.ts
@@ -16,6 +16,7 @@ import {ICard} from './cards/ICard';
 import {isCompatibleWith} from './cards/ICardFactory';
 import {GameOptions} from './game/GameOptions';
 import {ICorporationCard} from './cards/corporation/ICorporationCard';
+import {isIProjectCard} from './cards/corporation/IProjectCard';
 import {IProjectCard} from './cards/IProjectCard';
 import {IStandardProjectCard} from './cards/IStandardProjectCard';
 import {newCard} from './createCard';
@@ -81,12 +82,7 @@ export class GameCards {
       cards,
       this.gameOptions.includedCards,
     );
-    return cardsWithIncludedCards.filter((card) => {
-      return
-        card.type === CardType.AUTOMATED &&
-        card.type === CardType.ACTIVE &&
-        card.type === CardType.EVENT;
-    });
+    return cardsWithIncludedCards.filter(isIProjectCard);
   }
   public getStandardProjects() {
     return this.getCards<IStandardProjectCard>('standardProjects');
--- a/src/server/GameCards.ts
+++ b/src/server/GameCards.ts
@@ -11,13 +11,11 @@ import {PATHFINDERS_CARD_MANIFEST} from './cards/pathfinders/PathfindersCardMani
 import {CEO_CARD_MANIFEST} from './cards/ceos/CeoCardManifest';
 import {CardManifest, ModuleManifest} from './cards/ModuleManifest';
 import {CardName} from '../common/cards/CardName';
-import {CardType} from '../common/cards/CardType';
 import {ICard} from './cards/ICard';
 import {isCompatibleWith} from './cards/ICardFactory';
 import {GameOptions} from './game/GameOptions';
 import {ICorporationCard} from './cards/corporation/ICorporationCard';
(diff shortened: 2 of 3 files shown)


rvangrinsven:
@kberg I also noticed the issue you mentioned in another pull request. I think this solves all problems.

kberg:
Thanks. Let's get a test in there that proves this fixes the problem. Hang out, I'll put something together.

kberg:
Actually, a little manual testing and it's clear this doesn't quite fix the feature. I can add any kind of card, even a Prelude

image

image

The included cards feature should only apply to project cards (since other types of cards have their own exclusion/inclusion mechanism.) Make these two tests pass. (Give them better test names, please.)

```
  it('test name', function() {
    const gameOptions: GameOptions = {
      ...DEFAULT_GAME_OPTIONS,
      corporateEra: true,
      includedCards: [CardName.DONATION],
    };
    expect(() => new GameCards(gameOptions)).to.throw(/Donation is not a project card/);
  });

  it('test name', function() {
    const gameOptions: GameOptions = {
      ...DEFAULT_GAME_OPTIONS,
      corporateEra: true,
      includedCards: [CardName.VENUSIAN_INSECTS],
    };
    const names = new GameCards(gameOptions).getCorporationCards().map((c) => c.name);
    expect(names).to.not.contain(CardName.VENUSIAN_INSECTS);
  });
```

There are no tests for the UI, but if you can look there I would be grateful.

kberg:
I am disabling the feature from the UI for the time being at https://github.com/terraforming-mars/terraforming-mars/pull/6588. I suspect once the issues I pointed out are fixed, the feature can be re-enabled. I can do that in a separate PR (just by rolling back #6588)

kberg (changes requested):
Changing the state to "changes requested"

src/client/components/create/CardsFilter.vue
```diff
@@ -38,7 +38,6 @@ const allItems: Array<CardName> = [
   ...getCards(byType(CardType.AUTOMATED)),
   ...getCards(byType(CardType.ACTIVE)),
   ...getCards(byType(CardType.EVENT)),
-  ...getCards(byType(CardType.PRELUDE)),
```
rvangrinsven, line 41: I think removing Preludes from both the included cards and excluded cards select box makes sense, since for prelude we have a separate functionality anyway. This has caused unintended behavior for us in the past anyway when combining these.

tests/GameCards.spec.ts
```diff
@@ -95,5 +95,65 @@ describe('GameCards', function() {
...
+    const gameOptions: GameOptions = {
+      ...DEFAULT_GAME_OPTIONS,
+      corporateEra: true,
+      includedCards: [CardName.VENUSIAN_INSECTS],
+    };
+    const names = new GameCards(gameOptions).getCorporationCards().map((c) => c.name);
+    expect(names).to.not.contain(CardName.VENUSIAN_INSECTS);
+  });
+
+  it('should not include corporation cards in the included cards', function () {
```
rvangrinsven, line 129: I didn't let it throw an error as in your example, but just filter out all non project cards that people might have manually added in a settings json file.

rvangrinsven:
@kberg I've adjusted the PR. Only thing still missing is UI tests, if necessary maybe you could give a starting point?

src/client/components/create/CardsFilter.vue
kberg: Let me think through the impact: it's going to be in the case where someone has a JSON file which uses the "exclude cards" behavior. But I think that may be acceptable. If it causes problems, I'll own it instead of kicking it back to you.

src/server/GameCards.ts
```diff
@@ -75,7 +76,22 @@ export class GameCards {
...
+    const cardsWithIncludedCards = this.addCustomCards(
+      cards,
+      this.gameOptions.includedCards,
+    );
+    return cardsWithIncludedCards.filter((card) => {
+      if (!card) {
+        return false;
+      }
+      return (
+        card.type !== CardType.CORPORATION &&
```
kberg, line 89: How about  instead be inclusive rather than exclusive? e.g.

```
return card.type === CardType.AUTOMATED || card.type === CardType.ACTIVE || card.type === CardType.EVENT;
```

If a new type, like CEO, comes in, this is the safer approach.

tests/GameCards.spec.ts
```diff
@@ -95,5 +95,65 @@ describe('GameCards', function() {
     const names = new GameCards(gameOptions).getProjectCards().map((c) => c.name);
     expect(names).to.contain(CardName.VENUSIAN_INSECTS);
   });
+
+  it('should not include the included cards in the standard projects', function() {
```
kberg, line 99: Ooh more and more tests. Thank you!

tests/GameCards.spec.ts
kberg: OIK good

kberg:
> @kberg I've adjusted the PR. Only thing still missing is UI tests, if necessary maybe you could give a starting point?

Don't do UI tests, it's not fair. I think manual testing there is enough.

kberg (commented):
This is a good change. I will run it again locally, but I think this is good to go.

kberg:
> This is a good change. I will run it again locally, but I think this is good to go.

Modulo that minor change I requested on how you filter card types.

kberg (approved):
Took a few tries, but I fixed that last bit. Good work.

PR #6591, opened 2024-04-14
Italian translations for start and game setup screens

ryoku:
Added most missing translations to Italian in start screen and game setup screen


kberg (approved):
Thank you!

PR #6599, opened 2024-04-16
Norwegian localization and various translations fixes

JTBringe:


src/client/components/GameEnd.vue
```diff
@@ -205,6 +205,7 @@ import {Color} from '@/common/Color';
 import {CardType} from '@/common/cards/CardType';
 import {getCard} from '@/client/cards/ClientCardManifest';
 import {GlobalParameter} from '@/common/GlobalParameter';
+import {$t} from '@/client/directives/i18n';
```
kberg, line 208: It is not allowed to import `client` code into `common`, sorry.

src/client/components/OtherPlayer.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div v-show="isVisible()">
     <div :class="'player_translucent_bg_color_' + player.color" class="other_player_header">
-      <div class="player_name"> {{ player.name }} played cards </div>
+      <div class="player_name" v-i18n> {{ player.name }} played cards </div>
```
kberg, line 4: This can't really be translated since the player name is not parameterized.

src/client/components/help/HelpIconology.vue
```diff
@@ -49,7 +49,6 @@
                 <div class="resource-tag tag-city"></div>
                 <div class="help-icon-label" v-i18n>City</div>
             </div>
-            <br>
```
kberg, line 52: This is an intentional break. Is there a reason why you removed it?

src/styles/language_hacks.less
```diff
@@ -178,6 +178,18 @@
     }
 }
 
+.language-de {
```
kberg, line 181: What is this doing?

src/styles/language_icon.less
```diff
@@ -1,5 +1,7 @@
 .language-switcher {
   text-align: center;
+  position: relative;
```
kberg, line 3: What is this doing?

kberg (changes requested):
Thanks for all of this! I typically ask that changes like this are isolated. For instance, there are some changes you make that are not isolated to translations, and could be their own PR. The one that changes the `common` code, for instance, is now a point of debate which delays getting all the rest of it in.

src/client/components/help/HelpIconology.vue
JTBringe: I removed it in the beginning of December, but then I reverted the change a few days later.
For some reason it was removed in the version I pushed now.
Anyhow, I'll fix it.

src/styles/language_icon.less
JTBringe: Default is position: static, which prevents z-index from having an effect.

src/styles/language_hacks.less
JTBringe: It changes "VP" on the player info to "SP", to align with Norwegian and German translations.
When the player changes languages the <body> gets a class that reflects the language, such as language-nb.
I've corrected a mistake with the classname for Norwegian (from -no to -nb), pushing it soon.

src/client/components/OtherPlayer.vue
JTBringe: You're right. For some reason I thought I saw it working when doing local testing, but it must've been something with a similar wording. I'll remove it.

src/client/components/GameEnd.vue
JTBringe: GameEnd.vue is a file in the client folder, and there's several imports from both common and client into that file.

kberg: I mistakenly _this file_ was in `/common`. My mistake, sorry.

kberg (approved):
Looks great. It looks like by the commit history that you've been working on this for some time. Thanks once more. :rocket:

kberg:
Dangit, merge conflicts: 
image

LMK if you're stuck.

PR #6611, opened 2024-04-17
Fix strings and add some Polish translations

ololuki:
Fix translations - remove trailing spaces.
Fix typo in Polish translation.
Add some new Polish translations. String order based on de translation.


src/client/components/help/HelpPhases.vue
```diff
@@ -6,7 +6,10 @@
           <label for="initial-draft" class="first"><span v-i18n>Initial Drafting (optional variant)</span></label>
             <ul>
               <li v-i18n>Draw 3 corporation cards</li>
-              <li v-i18n>Draw 4 prelude cards <div class="expansion-icon expansion-icon-prelude"></div></li>
+              <li>
```
kberg, line 9: Is this a meaningful change?

kberg (commented):
Hi -- thanks for the contribution. I just have a question about the change to `HelpPhases.vue` because I can't tell if it's an intentional change.

Thanks!

src/client/components/help/HelpPhases.vue
ololuki: It is to fix translation in languages other than de and ru - otherwise I would need to add trailing space to pl and 6 other languages like: "Draw 4 prelude cards " so I just fixed it using convention already existing in this file.

kberg: Thanks, I see.

kberg (approved):

Issue #5488, opened 2023-03-20
tile-placing cards and Mars University


kevinb9n:
In the vein of #4313, it would be great if when playing Natural Preserve we could place the tile (possibly drawing a card) before handling the discard-and-draw. Could affect Research Outpost too, and something similar with Research Colony (& Pluto).

kberg:
Gotcha. This will be on the list of things to deal with when thinking about cards and play order.

TarunTheo13:
Is this still a problem?

PR #6616, opened 2024-04-19
Deprioritize robotic workforce and mars univeristy.

kberg:
These got deprioritized at the same time as part of addressing #5488. By and large, these two behaviors don't need to be priority options; discard after you get all the cards you can get! Implement robotic workforce after gaining all the required production you can get!

Though these are close-to-the-edge cases, I'm surprised that Robotic Workforce had such high priority for this long.


kberg:
This fixes #5488

PR #6620, opened 2024-04-22
Made New Venice compatible with Martian Lumber Corp

### LINKED ISSUE #6605: New Venice is not compatible with Martian Lumber Corp
Martian Lumber Corp can turn plants into 5MC when paying for building tags.
New Venice has a building tag, and you must spend 2 plants.

This is what `reserveUnits` is for.

ryoku:
Fixes #6605


kberg (approved):
Thank you for this change. Nice fix!

Issue #6365, opened 2024-01-16
Shuffle moon tiles


kberg:
### Discussed in https://github.com/terraforming-mars/terraforming-mars/discussions/6363

<div type='discussions-op-text'>

<sup>Originally posted by **Negrolyte** January 12, 2024</sup>
It would be nice to randomize the moon tiles, similar to randomizing the different board tiles.</div>

PR #6635, opened 2024-04-25
Randomisation of Moon tiles

ryoku:
Implements #6365

Shuffling of Moon tiles is enabled through a specific option in the Game setup screen


src/server/boards/BoardBuilder.ts
```diff
@@ -104,6 +104,7 @@ export class BoardBuilder {
     return this.spaces;
   }
 
+  /*
```
kberg, line 107: What's the plan here?

src/client/components/create/CreateGameForm.vue
```diff
@@ -260,6 +260,14 @@
                                     <span v-i18n>Randomize board tiles</span>&nbsp;<a href="https://github.com/terraforming-mars/terraforming-mars/wiki/Variants#randomize-board-tiles" class="tooltip" target="_blank">&#9432;</a>
                             </label>
 
+                            <template v-if="moonExpansion">
```
kberg, line 263: I appreciate the extra effort, but let's just use the shuffle parameter already available. I don't feel that there's a strong reason to have one shuffled but not the other. At least, it's not worth the conceptual and maintenance weight.

tests/boards/BoardBuilder.spec.ts
```diff
@@ -95,4 +97,23 @@ describe('BoardBuilder', function() {
       expect(board.getSpaces(SpaceType.RESTRICTED, TestPlayer.BLUE.newPlayer())[0].bonus).is.empty;
     }
   });
+
+  it('Randomizes map for Moon', () => {
```
kberg, line 101: This behavior doesn't have anything to do with BoardBuilder. It probably belongs someplace else.

kberg (changes requested):
I'll have to look at the algorithm more closely. So this is an initial pass.

src/server/boards/BoardBuilder.ts
ryoku: No specific plan, I spent time analysing this and in the end I realised it is not used, so I think it could be removed to reduce cognitive load. If you are ok I will remove it rather than commenting. If there is specific reason that I do not see to keep the function I will remove the comments.

src/client/components/create/CreateGameForm.vue
ryoku: Actually this was my first implementation. I will revert back to it.

ryoku: You are right! Moved to MoonBoard.spec.ts

tests/boards/BoardBuilder.spec.ts
ryoku: You are right! Moved to MoonBoard.spec.ts

src/client/components/create/CreateGameForm.vue
ryoku: It's done

Borbarad2:
@ryoko: Are you available on discord, too? What is your name there?

src/server/boards/BoardBuilder.ts
kberg: You know what? Leave it for a bit. The board API needs one more cleanup phase. Thanks for identifying it though!

src/server/moon/MoonBoard.ts
```diff
@@ -31,22 +29,29 @@ export class MoonBoard extends Board {
...
     b.row(1).mine(TITANIUM, TITANIUM).mine(/* Mare Imbrium */).land(STEEL).land().land();
     b.row(0).mine().land(STEEL).land(STEEL, TITANIUM).mine(/* Mare Serenatis*/).mine(TITANIUM).land(STEEL, STEEL);
     b.row(0).land(STEEL).land().land().mine(TITANIUM).mine(TITANIUM);
     b.row(0).land().mine(TITANIUM).mine(/* Mare Nubium */).land().mine(/* Mare Nectaris */).land(STEEL);
     b.row(1).land().land(STEEL).land(STEEL).land(DRAW_CARD, DRAW_CARD).land(STEEL);
     b.row(2).land(DRAW_CARD, DRAW_CARD).mine(TITANIUM).mine(TITANIUM, TITANIUM).land();
-    b.colony();
-    return new MoonBoard(b.spaces);
+
+    if (gameOptions.shuffleMapOption!== undefined && gameOptions.shuffleMapOption) {
```
kberg, line 46: Can be shortened to `if (gameOptions.shuffleMapOption === true)`

tests/moon/MoonBoard.spec.ts
```diff
@@ -3,21 +3,64 @@ import {SpaceType} from '../../src/common/boards/SpaceType';
...
 
 describe('MoonBoard', function() {
   let board: MoonBoard;
 
   beforeEach(function() {
-    board = MoonBoard.newInstance();
+    board = MoonBoard.newInstance(DEFAULT_GAME_OPTIONS, new SeededRandom(0));
   });
 
+  it('sanity', () => {
```
kberg, line 16: I added this sanity test which ensures that the default board doesn't actually change based on the rewrite. And that's enough for me to approve the PR (once the tests finish.)

If they don't pass it means something's funny with the new setup code.

kberg (approved):

kberg:
Looks like it's blocked now on a merge conflict. :/

ryoku:
> @Ryoko: Are you available on discord, too? What is your name there?

Yes, here my nick: ryoku5049

kberg (approved):
Thank you!

kberg:
Still got merge conflicts. :/

kberg:
Nope never mind, it's good

PR #6645, opened 2024-04-28
Addition of Japanese translation 5

nasubi-land:
Addition of Japanese translation, focusing on Pathfinders.


kberg (approved):
Thank you for improving translations!

PR #6719, opened 2024-06-07
Update MarsFrontierAlliance and other misc

Borbarad2:
Adapt to consistent language. Making it consistent for translation.


kberg (approved):
Thanks!

PR #6728, opened 2024-06-10
Typo fix: SearchForLife.ts

derornos:
- search for life: action


Borbarad2:
Top. Indeed that typo blocks all other translation. Thanks for fixing it.

kberg (approved):
Thanks. I don't know how I did that.

kberg:
> Thanks. I don't know how I did that.

However, I just fixed it in another, larger, commit.

PR #6756, opened 2024-06-21
Fixes #6755 by returning correct object in initial card selection client side

### LINKED ISSUE #6755: Initial draft returning AndOptionsResponse instead of SelectInitialCardsResponse
Initial draft is returning an AndOptions object rather than a SelectInitialResponse object on the latest main branch. Since the rewrite of SelectInitialResponse hasn't made it to Heroku, can't reproduce on the official server. 
Specifically I am getting a "Not a valid SelectInitialCardsResponse" popup. Tracking that down and figuring what the response is via a debug statement: 

```
 if (!isSelectInitialCardsResponse(input)) {
      console.log(input);
      throw new InputError('Not a valid SelectInitialCardsResponse');
    }
```

yields
```
[2024-06-21 10:54:07 PDT] {
  type: 'and',
  responses: [
    { type: 'card', cards: [Array] },
    { type: 'card', cards: [Array] },
    { type: 'card', cards: [Array] }
  ]
}
```

I believe the type should be `initialCards` so this is breaking down somewhere along the way.

puuj:
Updated the front end to send the right object type.

### DIFF
--- a/src/client/components/SelectInitialCards.vue
+++ b/src/client/components/SelectInitialCards.vue
@@ -43,7 +43,7 @@ import SelectCard from '@/client/components/SelectCard.vue';
 import ConfirmDialog from '@/client/components/common/ConfirmDialog.vue';
 import {getPreferences, Preferences, PreferencesManager} from '@/client/utils/PreferencesManager';
 import {Tag} from '@/common/cards/Tag';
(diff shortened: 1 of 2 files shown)


kberg (approved):

kberg:
Thank you!

kberg (approved):

PR #6807, opened 2024-07-13
8 Prelude Cards

maserion:
[Prelude 2]: Envoys from Venus
[Prelude 2]: Colonial Representation
[Prelude 2]: Jovian Envoys
[Prelude 2]: Microgravity Nutrition
[Prelude 2]: Soil Studies
[Prelude 2]: SpecialPermit
[Prelude 2]: StratosphericExpedition
[Prelude 2]: UnexpectedApplications


kberg:
image
image
image
image
image

@maserion this looks great!

kberg:
So the lack of a substantial test for Microgravity Nutrition means I definitely can't merge it right now. But we can do that.

src/server/cards/prelude2/MicrogravityNutrition.ts
```diff
@@ -0,0 +1,35 @@
...
+        description: 'Increase your M€ production 1 step for each colony you have.',
+        cardNumber: 'P79',
+        renderData: CardRenderer.builder((b) => {
+          b.production((pb) => pb.megacredits(1).slash().colonies(1, {size: Size.SMALL})).br;
+        }),
+      },
+    });
+  }
+
+  public override bespokePlay(player: IPlayer) {
```
kberg, line 30: I think this can be reduced to

`production: {megacredits: {colonies: {colonies: {}}}},`

src/server/cards/prelude2/SoilStudies.ts
```diff
@@ -0,0 +1,38 @@
...
+      behavior: {
+        stock: {
+          megacredits: {
+            tag: [Tag.VENUS, Tag.PLANT],
+            colonies: {colonies: {}},
+          },
+        },
+      },
+
+      metadata: {
```
kberg, line 27: Reduce to 

```
      metadata: {
        cardNumber: 'P81',
          b.plants(1).slash().tag(Tag.VENUS).tag(Tag.PLANT).colonies(1).br;
        }),
        description: 'Requires that temperature is -4 C or lower. Gain 1 plant per Venus tag, plant tag, and colony you have.',
      },
```

src/server/cards/prelude2/MicrogravityNutrition.ts
maserion: this part is taken fgrom EcologyResearch.ts
```
  public override bespokePlay(player: IPlayer) {
    const coloniesCount = player.getColoniesCount();
    player.production.add(Resource.PLANTS, coloniesCount, {log: true});
 }
```

kberg:
Do you need some help running things locally?

Also, how about let's just get five in at a time. The more you add, the more everything has to be right first.

maserion:
> Do you need some help running things locally?
> 
> Also, how about let's just get five in at a time. The more you add, the more everything has to be right first.

the issue is with one test, all other looks fine

src/server/cards/prelude2/MicrogravityNutrition.ts
kberg: Yeah, it should change there too. (NB, custom code like that makes the card incompatible with the Engineer milestone.) It should be OK.

src/server/cards/prelude2/UnexpectedApplication.ts
```diff
@@ -0,0 +1,39 @@
...
+          b.minus().cards(1).venus(1);
+        }),
+        description: 'Discard 1 card to terraform Venus 1 step.',
+      },
+    });
+  }
+  public override bespokeCanPlay(player: IPlayer) {
+    return player.cardsInHand.length > 0;
+  }
+  public override bespokePlay(player: IPlayer) {
```
kberg, line 35: This is a little bit out of order because the card has to be played before raising Venus 1 step.

But doing that makes the whole thing more complicated: code has to identify if Reds are in power (using the canAfford behavior.) Let's address this by adding custom spend behavior, so the card looks like this:

```
behavior: {
  spend: {cards: 1},
  global: {venus: 1},
```

Don't do that, I'll do it. For now, this will do.

kberg (changes requested):
The changes to package.json and package-lock.json have to go before merge.

Still needs 3 more tests. As I was suggesting earlier, perhaps remove the 3 new cards to a new PR.

kberg (approved):
Looks great. Thank you!

PR #6814, opened 2024-07-14
[Prelude 2]: VenusOrbitalSurvey

eratchia:
Not sure whether the chosen bought cards should be known (if one of two non-venus cards is bought). For now I left it unknown because it was easier this way.

### DIFF
--- a/src/server/cards/prelude2/VenusOrbitalSurvey.ts
+++ b/src/server/cards/prelude2/VenusOrbitalSurvey.ts
@@ -1,0 +1,76 @@
+import {Tag} from '../../../common/cards/Tag';
+import {CardName} from '../../../common/cards/CardName';
+import {CardRenderer} from '../render/CardRenderer';
+import {IPlayer} from '../../IPlayer';
+import {Card} from '../Card';
+import {CardType} from '../../../common/cards/CardType';
+import {IActionCard} from '../ICard';
+import {DrawCards} from '../../deferredActions/DrawCards';
+import {IProjectCard} from '../IProjectCard';
+import {LogType, keep} from '../../deferredActions/ChooseCards';
+import {ChooseCards} from '../../deferredActions/ChooseCards';
+import {Size} from '../../../common/cards/render/Size';
+import {oneWayDifference} from '../../../common/utils/utils';
+
+export class VenusOrbitalSurvey extends Card implements IActionCard {
+  constructor() {
+    super({
+      name: CardName.VENUS_ORBITAL_SURVEY,
+      type: CardType.ACTIVE,
+      tags: [Tag.VENUS, Tag.SPACE],
+      cost: 18,
+
+      metadata: {
+        cardNumber: 'P88',
+        renderData: CardRenderer.builder((b) => {
+          b.action(undefined, (ab) =>
+            ab.empty().startAction.empty()).text('Action: Reveal the top 2 cards, take any venus cards to hand for free. Any other card you either buy or discard', Size.TINY, true);
+        }),
+      },
+    });
+  }
+
+  public canAct(player: IPlayer): boolean {
+    if (!player.game.projectDeck.canDraw(2)) {
+      this.warnings.add('deckTooSmall');
+    }
+    return true;
+  }
+
+  public action(player: IPlayer) {
+    const isVenus = (card: IProjectCard) => player.tags.cardHasTag(card, Tag.VENUS);
+
+    player.game.defer(new DrawCards(player, 2, {}).andThen((cards) => {
+      let message = '${0} revealed ';
+      if (cards.length === 0) {
+        message += 'no cards';
+      } else {
+        message += '${1}';
+        if (cards.length === 2) {
+          message += ' and ${2}';
+        }
+      }
+      player.game.log(message, (b) => {
+        b.player(player);
+        for (const card of cards) {
+          b.card(card);
+        }
+      }, {});
+
+      const venus = cards.filter(isVenus);
+      if (venus.length > 0) {
+        keep(player, venus, [], LogType.DREW_VERBOSE);
+      }
+
+      // The cards bought are private here
+      const rest = oneWayDifference(cards, venus);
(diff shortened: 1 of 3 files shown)


kberg:
Thank you! I'll look at this shortly.

src/server/cards/prelude2/VenusOrbitalSurvey.ts
```diff
@@ -0,0 +1,76 @@
...
+        renderData: CardRenderer.builder((b) => {
+          b.action(undefined, (ab) =>
+            ab.empty().startAction.empty()).text('Action: Reveal the top 2 cards, take any venus cards to hand for free. Any other card you either buy or discard', Size.TINY, true);
+        }),
+      },
+    });
+  }
+
+  public canAct(player: IPlayer): boolean {
+    if (!player.game.projectDeck.canDraw(2)) {
```
kberg, line 34: I can't recall if there's been discussion about what should happen in this case (or even if some solutions have been implemented) but I believe the answer is, if the cards can't be drawn, the action can't be taken.

Save me from doing research, I suppose? :D

src/server/cards/prelude2/VenusOrbitalSurvey.ts
```diff
@@ -0,0 +1,76 @@
...
+      this.warnings.add('deckTooSmall');
+    }
+    return true;
+  }
+
+  public action(player: IPlayer) {
+    const isVenus = (card: IProjectCard) => player.tags.cardHasTag(card, Tag.VENUS);
+
+    player.game.defer(new DrawCards(player, 2, {}).andThen((cards) => {
```
kberg, line 43: , {} is not required.

src/server/cards/prelude2/VenusOrbitalSurvey.ts
```diff
@@ -0,0 +1,76 @@
...
+      let message = '${0} revealed ';
+      if (cards.length === 0) {
```
kberg, line 45: Yes, cards should be revealed.

No need for custom logging, use LogHelper.logDrawnCards (and don't set privateMessage)

kberg (commented):
This is nice work.

Cards being public should simplify the second half of the function too.

src/server/cards/prelude2/VenusOrbitalSurvey.ts
eratchia: I think the distinction is important here as not all of the revealed cards land in the players hand here.

src/server/cards/prelude2/VenusOrbitalSurvey.ts
eratchia: I was looking at it like something between drawing cards and things like search for life/asteroid deflection system that reveal cards. Both seem to be available with lacking cards with the latter giving a warning.

kberg:
Thanks for this. At this point, I'm trying to just simplify your implementation. I'm not going to force  those changes, but I do think we'll wind up down the road removing that behavior from SearchForLife, for instance.

eratchia:
Well in that case, I think allowing for the action anyways without any warning is the way to go.

kberg:
> Well in that case, I think allowing for the action anyways without any warning is the way to go.

OK. What I mean though is eventually Search For Life's action won't be playable if the draw and discard piles are empty. (e.g. you can't use it as a stall action.)

eratchia:
> > Well in that case, I think allowing for the action anyways without any warning is the way to go.
> 
> OK. What I mean though is eventually Search For Life's action won't be playable if the draw and discard piles are empty. (e.g. you can't use it as a stall action.)

Oh, ok. But I assume this action should be still usable with one card right?

kberg:
> Oh, ok. But I assume this action should be still usable with one card right?

I don't think so. Here's why:

It is generally known (though I didn't really internalize it) that there are only three things that are optional in an action:
1. If a global parameter is at its maximum, you can take an action that raises a global parameter without having to do so. So, spending 8 heat when temperature is already maxxed is legal, but used as a stall tactic.
2. Actions that remove resources from any player is optional: you do not have to remove resources. Side-note: removing production _is_ mandatatory. So if an action removes plant production and you're the only one with plant production, guess whose plant production gets removed.
3. Actions that add resources of a certain type to your cards are allowed, even if you don't have any cards of that type.

ref: https://boardgamegeek.com/thread/1705209/official-faq-rules-clarifications

I tried to get a statement about running out the deck, but my question was not taken very seriously: https://boardgamegeek.com/thread/3217206/empty-deck-discard-pile-during-action-phase/page/1

You can see that my original thought was to house rule to allow the action, but that's just because the app hasn't taken a definitive stance yet.

So like I say, leave it be if you like. But if you want to keep the code simpler, take it out.

Either way, at least you understand where I'm coming from.

eratchia:
Huh, thanks for the detailed explanation. I didn't previously understand drawing to be manatory.

I do think card draw being required does make low deck gameplay very weird but I can definitely understand that the readability and rule consistency shouldn't suffer because of such an edge case.

src/server/cards/prelude2/VenusOrbitalSurvey.ts
```diff
@@ -31,31 +31,17 @@ export class VenusOrbitalSurvey extends Card implements IActionCard {
...
-        }
-      }
+      const message = '${0} revealed ${1} and ${2}';
       player.game.log(message, (b) => {
-        b.player(player);
-        for (const card of cards) {
-          b.card(card);
-        }
-      }, {});
+        b.player(player).card(cards[0]).card(cards[1]);
```
kberg, line 43: And add the {tags: true} attribute for both cards, just like SearchForLife. It'll be useful.

`player.game.log('${0} revealed and discarded ${1}', (b) => b.player(player).card(card, {tags: true}));`

kberg (approved):
Add that last bit if you don't mind. This will be good to deploy tomorrow. Thank you!

kberg:
Thanks for sticking with me.

PR #6824, opened 2024-07-16
Multiple expansion icons fix

EthanDobbs:
Fixed Corporation Cards and Underworld Icons


kberg:
I think we got a good solution in #6853. Teamwork!

PR #6829, opened 2024-07-16
Update prelude2 translations

Borbarad2:


src/locales/de/prelude.json
Borbarad2, line 1: Reallocate Base Planning here, and adapt the modification

kberg (approved):
Thank you!

PR #6831, opened 2024-07-16
[Prelude 2]: MainBeltAsteroids

eratchia:
Screenshot_20240716_120424


kberg (approved):
This is a good change

PR #6840, opened 2024-07-19
Update Engineer Award to include Prelude 2 Bespoke Production Cards

poofytoo:
Reviewed the list of currently implemented prelude 2 cards and added relevant cards to be included in the engineer award count


kberg (approved):
Thanks!

Issue #6839, opened 2024-07-19
Server Crashing while drafting causes hard lock game state


jsmith613:
Please include:
1. A clear and concise description of the bug.
When in the first drafting round of a new generation, the game server crashed.  Upon reload, both players were shown no cards to draft.  We are shown the 'keep' button, but upon pressing the button, an error occurs.
2. Clear steps to reproduce, and the expected behavior.
- Start a new generation with drafting enabled
- Crash the server
- Reload the game
- Shown no cards to draft
- Hit Keep button
- Shown input error
3. Any screenshots
![Screenshot(18)](https://github.com/user-attachments/assets/33ed77ba-a2c9-4d1c-a4e2-cf8e63b2ab33)
![Screenshot(19)](https://github.com/user-attachments/assets/0ce93e31-091d-43f1-bbf2-b4b9fa9941b8)
4. Any information from  the Bug Report Dialog
{
  "url": "https://terraforming-mars.herokuapp.com/spectator?id=s7593ad0c9913",
  "color": "blue",
  "step": 64,
  "version": "20792f5",
  "builtAt": "Tue, 16 Jul 2024 22:53:26 GMT",
  "browser": "Firefox 128",
  "language": "en",
  "experimental_ui": false
}

kberg:
Thank you. I'm glad this was fixable.

Issue #6810, opened 2024-07-13
Glitch during card draft, only able to purchase one card instead of four


Erostal:
This issue occurred while playing a two player game, during the second generation when drafting cards. While selecting the second card for the draft, both players experienced a freeze/hang on locking in the second card. After 5-10 seconds, both of the players refreshed the browser, which then jumped them to the card buying stage of the draft, however there was only one card available to purchase (see attached photo). The card was the first card selected from the draft, but the other 3 were effectively gone/discarded for both players. This issue did not recur in the following generations.

Debug Info:
{
  "url": "https://terraforming-mars.herokuapp.com/spectator?id=sa6dcfc26d2e2&noredirect=",
  "color": "orange",
  "step": 396,
  "version": "d8e9413",
  "builtAt": "Fri, 12 Jul 2024 19:55:34 GMT",
  "browser": "Firefox 127",
  "language": "en",
  "experimental_ui": false
}

[TFM Game Log July 13, 2024.txt](https://github.com/user-attachments/files/16203645/TFM.Game.Log.July.13.2024.txt)
![Gen 2 Draft Issue](https://github.com/user-attachments/assets/843d1744-3452-4185-904d-be6a7032f572)

kberg:
Thank you for your very detailed response. I'll take a look today.

PR #6844, opened 2024-07-19
Fix: draft save behaviour + in memory database saving

eratchia:
This is where tinkering with #6839 led me. I'm not completely sure of this solution yet so I'm opening it as a draft PR.

This change makes it so that the game is always saved after draft cards are dealt. That should mean that both:
1. Previous "hack" that fixed the behavior of #6810 isn't needed because when the game saves after dealing the cards players needToDraft properties are properly saved.
2. #6839 no longer occurs because partial information that lead to a bad restore isn't saved.

This change additionally fixes a problem with InMemoryDatabase which would freeze on saving a game with a higher lastSaveId because it would use push() which adds no new elements.

### DIFF
--- a/src/server/Draft.ts
+++ b/src/server/Draft.ts
@@ -77,7 +77,7 @@ export abstract class Draft {
     const players = this.game.getPlayers();
 
     // When restoring drafting, it might be that nothing was dealt yet.
-    if (players.every((p) => p.needsToDraft === undefined)) {
+    if (!players.some((p) => p.needsToDraft !== undefined)) {
       this._startDraft();
       return;
     }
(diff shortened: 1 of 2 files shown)


kberg:
Thanks. I'll spend time on this now, and reference this while debugging.

kberg:
@eratchia thank you! The eventual fix is very similar to yours, mostly only structurally different, and includes a test. Do you still want the inmemory database change? I'd be happy to incorporate it.

kberg:
If you want to keep the IMD fix, I'll be happy to reopen this PR or accept a new one.

eratchia:
@kberg Nice! 

The in-memory database change isn't super important as it is in test but it is very annoying to find while debugging. I think fixing it in a similar way could save someone time at some point.

PR #6853, opened 2024-07-21
Adjust Corporation Card Expansion Icon

poofytoo:
The Icon on the Corporation Cards that shows the expansion would occasionally overlap the card. There's no real great place to place this icon, so I tucked it in the top right hand corner, slightly proud of the card to match the card's tag.

### Before: 
image

### After: 
Regular Cards:
image
Small Cards:
image

### DIFF
--- a/src/styles/cards_v2.less
+++ b/src/styles/cards_v2.less
@@ -1,8 +1,8 @@
 @import "./variables.less";
 .card-corporation-expansion {
     position: absolute;
-    top: 75px;
-    right: 33px;
+    top: 38px;
+    right: -11px;
 }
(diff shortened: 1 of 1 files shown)


kberg:
I might have to get used to it along the edge like that, but it's completely better than what was already there. Thank you!

kberg (approved):

Borbarad2:
Thanks, much better now. But still the icon are offset (to the right) over the cards themself. If the icon could be moved more left onto the card, it would be even better and looks nicer. Thus now all icons are over the left edge of the corporation card.

Issue #6868, opened 2024-07-28
Soil Bacteria giving too many plants


IceTyp:
When playing Soil Bacteria, one expects to obtain 3 plants and 1 plant from the triggered effect. However, one obtains a total of 5 plants (4 + 1) instead.

![image](https://github.com/user-attachments/assets/ea6a9acb-42b6-49ad-8231-432e7041ca89)

It happened in this game:
{
  "url": "https://terraforming-mars.herokuapp.com/spectator?id=sc6c5ce081922#cards",
  "color": "orange",
  "step": 17,
  "version": "d79227e",
  "builtAt": "Sat, 27 Jul 2024 02:28:20 GMT",
  "browser": "Firefox 128",
  "language": "en",
  "experimental_ui": false
}

PR #6870, opened 2024-07-29
[Prelude 2] Soil Bacteria fix

maserion:
Update SoilBacteria.ts to give 3 plants initlaly instead of 4 and then 1 plant from own tag
Update SoilBacteria.spec.ts to check for 3 and 4 plants


kberg (approved):
Oh thanks! Looks like this was my fault. Sorry for making you look at it, but thank you for fixing it.

kberg:
Fixes #6870

kberg:
Fixes #6868

PR #6874, opened 2024-08-01
[Prelude 2] Rise To Power and High Circles

maserion:
[Prelude 2] Rise To Power
[Prelude 2] High Circles


kberg (approved):
Nice work. I might tighten up the tests a little sometime later, but the ones you wrote are quite good.

PR #6905, opened 2024-08-06
Fix Deimos, which wasn't eroding tiles after all.

kberg:


kberg:
Fixes https://github.com/terraforming-mars/terraforming-mars/issues/6902

PR #6919, opened 2024-08-09
[Prelude 2] Colonial Envoys

maserion:
[Prelude 2] Colonial Envoys


src/server/cards/prelude2/ColonialEnvoys.ts
kberg, line 1: Oh goodness no wonder you had difficulty with it. I'll approve this but there's probably a simpler way to do this.

kberg (approved):
Nice!

PR #6922, opened 2024-08-11
More Polish Translations

yhnavein:
- Fix broken translations in Pathfinder expansion
- Add missing translations for Prelude 2 and Promo Cards expansions
- Some small adjustments and fixes of PL translation

As always - tested on my local instance and it was working fine.


kberg (approved):

kberg:
Thank you!

PR #6938, opened 2024-08-20
Update Floating Refinery

maserion:
It should be usable regardless of number of floaters on card
https://github.com/terraforming-mars/terraforming-mars/issues/6935


kberg (approved):
I'd have added a test here, but this is fine. Good work. Thanks for fixing this!

PR #6954, opened 2024-08-30
Update auto pass message

nyddogghr:
Hi,
Here is a suggestion to improve the auto pass message. Based on what I've seen in the code, it only passes once on your next action when you perform an action with the checkbox checked.
If that's correct, I suggest this change. I was confused myself when I saw it for the first time and I kinda feared checking it, in case it just passed until the end of the game.
If I'm wrong about this feature then I apologize.


kberg (approved):

PR #6958, opened 2024-09-01
Update promo_cards.json

Borbarad2:
Add new autumn promo 2024


kberg (approved):

kberg:
image
Which of these is correct?

Borbarad2:
Perfect, thanks so far:

For  Update promo_cards.json #6958 this is correct:
  "Hospitals": "Krankenhäuser",
  "Effect: Each time a city is placed, gain a disease here.": "EFFEKT: Jedes Mal, wenn eine Stadt platziert wird, erhälst du hier eine Krankheit.",
  "Action: Remove a disease from ANY OF YOUR cards to gain 1 M€ per city in play.": "AKTION: Entferne eine Krankheit von EINER DEINER Karten und erhalte 1 M€ pro Stadt im Spiel.",
  "Decrease your energy production 1 step.": "Reduziere deine Energieproduktion um 1.",

The patch-37 is correct. --> two lines, separate "Effect: " and "Action: "

PR #6966, opened 2024-09-04
Improve auto pass message again

nyddogghr:
Hi, after testing the auto pass I realised this was actually immediate, and not waiting for our next action to pass.
Here is a suggestion to udpate (again, sorry) the message to make it clear it will pass immediately after doing the action.


kberg (approved):
Thank you!

PR #6990, opened 2024-09-25
Update PostgreSQL reporting and PostgreSQL testing.

kberg:


kberg:
There's one remaining Postgresql test failing, but this is good. And necessary.

PR #6991, opened 2024-09-25
Remove everything but the last 10 rows in PostgreSQL.

kberg:
This should drop database sizes pretty quickly. Estimated drop in the amount of stored characters is approximately 85%. If that matches estimated database size drop, it would put us far within storage limits.


kberg:
After some manual work, it looks like approx 83% of rows have been removed. rows != characters, and neither of those != MB, but I'm looking forward to the maintenance `VACUUM` later tonight.

PR #7003, opened 2024-10-02
Custom Prelude starting number

maserion:
Number of Prelude Cards in starting hand can be adjusted now


src/client/components/create/CreateGameForm.vue
```diff
@@ -191,6 +191,14 @@
                                 <span v-i18n>Starting Corporations</span>
                             </label>
 
+                            <template v-if="prelude">
+                              <label for="startingPreludeENum-checkbox">
+                              <div class="create-game-expansion-icon expansion-icon-prelude"></div>
+                              <input type="number" class="create-game-corporations-count" value="4" min="4" :max="8" v-model="startingPreludes" id="startingPreludeNum-checkbox">
```
kberg, line 197: These can still be constants. Can you restore the constant in `common`? This comes from `GamesOverview`: 

```
  computed: {
    constants(): typeof constants {
      return constants;
    },
    serverId(): string {
      return (new URL(location.href)).searchParams.get('serverId') || '';
    },
  },
```

src/server/game/GameOptions.ts
```diff
@@ -54,6 +54,7 @@ export type GameOptions = {
   customPreludes: Array<CardName>;
   customCeos: Array<CardName>;
   startingCeos: number;
+  startingPreludes: number;
```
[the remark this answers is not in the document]
kberg: Also add a comment above it: // TODO(maserion): Remove '?' by 2025-01-01

src/client/components/create/CreateGameForm.vue
kberg, line 1: Have you tested that downloading and uploading work as expected? (Also, what happens if you upload a JSON file that doesn't have the new attribute?)

kberg (commented):
This is a great start. The bits that are left are mostly cleanup and verification of the different bits that should be verified.

src/server/game/GameOptions.ts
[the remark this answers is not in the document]
maserion: According to my understanding adding '?' will make that optional - that can be number or undefined. Making that undefined won't break current games (as this is undefined for all games played right now)/ However for all games created in the future, that will be already number, right?

src/client/components/create/CreateGameForm.vue
maserion: constant restored, but to be honest I am nto catching why :/

src/client/components/create/CreateGameForm.vue
maserion: yes, I checked that right now
when I am trying to upload old json (without startingPrelude), value is set automatically to 4
when I am trying to download new josn (with startingPrelude) everything looks fine
when I am trying to upload new json (with startingPrelude), value is imported correctly

src/server/game/GameOptions.ts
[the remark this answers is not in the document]
kberg: That's right. That's why sometime in the future (which I suggested as 2025-01-01) all the games will have the value, and we can remove the '?'.

src/client/components/create/CreateGameForm.vue
kberg: Thanks.

src/client/components/create/CreateGameForm.vue
kberg: Nice! We really should have repeatable tests, but for now this is terrific, thank you.

kberg (approved):
Thank you1

kberg:
I mean thank you!

PR #7014, opened 2024-10-08
Preludes Fix

maserion:


src/server/Game.ts
```diff
@@ -351,11 +351,15 @@ export class Game implements IGame, Logger {
           player.dealtProjectCards.push(...projectDeck.drawN(game, 10));
         }
         if (gameOptions.preludeExtension) {
-          if (gameOptions.startingPreludes !== undefined) {
-            player.dealtPreludeCards.push(...preludeDeck.drawN(game, gameOptions.startingPreludes));
+          if (gameOptions.startingPreludes === undefined || gameOptions.startingPreludes < 2) {
+            gameOptions.startingPreludes = 4;
```
kberg, line 355: Please use the constant values (and let's keep it simple and not go below 4. In retrospect my test for 3 was not a good one.)

This can be shortened to 

`gameOptions.startingPreludes = Math.min(gameOptions.startingPreludes ?? 0, constants.WHATEVER_THAT_CONST_IS_NAMED)`

Same for CEOs.

kberg (commented):
Just a little cleanup please. Thank you!

kberg (approved):

PR #7066, opened 2024-10-29
Auto select preludes in create game form

TarunTheo13:
CEO and Underworld preludes now auto-select when those expansions are selected in the Create Game Form. Also, titles for CEOs, Underworld and Prelude 2 are now present in the Custom Preludes list, as they are for all other expansions.


kberg (approved):
Thanks! Sorry I missed this one.

kberg (approved):

PR #7071, opened 2024-11-02
Rename milestones and awards from fan maps.

kberg:
This makes way for the modular milestones and awards to have their proper names.


kberg:
@maserion there are four separate commits in here. Three are cleanups, which at least help keep the main work a little easier to follow.

kberg:
Still not done. 
image

PR #7083, opened 2024-11-11
Fix message for Faraday so tag is specified

TarunTheo13:


kberg (changes requested):
Thanks for the suggestion. While that would work for English, it would not work for any other translation language unless they wrote a custom translation for every single tag.

I believe that the correct fix is to restore the original code, and then replace `${1}` with `${0}`.

TarunTheo13:
Okay will implement tomorrow :)

kberg (approved):
Thank you!

PR #7085, opened 2024-11-18
Pathfinders aesthetic changes

TarunTheo13:
Changes production box on Early Expedition, so it is visually in line with other city cards; capitalises the word 'any' on Martian Culture


kberg (approved):
This is a good change.

kberg:
image

PR #7092, opened 2024-11-24
Update ImportedHydrogen.ts

Borbarad2:
Increase readibility
(There is plenty of space avaialbale on the card itself)


kberg:
The preferred approach is to use `language_hacks.less`. That requires using CSS, which I realize you may not be familiar with. This is described in https://github.com/terraforming-mars/terraforming-mars/wiki/Translations#language-hacks-css

PR #7126, opened 2024-12-13
Include events in Tags.distinctCount with Odyssey

kberg:
It involved rewriting a fair bit of tags.


src/server/player/Tags.ts
```diff
@@ -183,6 +182,27 @@ export class Tags {
...
+   *
+   * This is also the maximum value that distinctTagCount can return.
+   */
+  public tagsInGame(): number {
+    if (this._tagsInGame === 0) {
+      this._tagsInGame = 10;
+      const game = this.player.game;
+      if (game.gameOptions.venusNextExtension) this._tagsInGame++;
+      if (game.gameOptions.moonExpansion) this._tagsInGame++;
+      if (game.gameOptions.pathfindersExpansion) this._tagsInGame++;
```
Borbarad2, line 201: Is gameOptions.underworldExpansion forgotten here? Underworld games does also add Mars tag

kberg: Yes.

PR #7153, opened 2025-01-02
Add test to Suitable Infrastructure to ensure it's effect works with Manutech

ryannewton:
This is in response to an [issue report](https://github.com/terraforming-mars/terraforming-mars/issues/7140) that Suitable Infrastructure's effect didn't work with Manutech.

I attempted to repro and the effect worked. Nonetheless, I thought a test would add protection for this scenario.


tests/cards/prelude2/SuitableInfrastructure.spec.ts
```diff
@@ -87,4 +88,19 @@ describe('SuitableInfrastructure', () => {
     expect(player.production.megacredits).eq(-1);
     expect(player.megaCredits).eq(0);
   });
+
+  it('works when player has other cards with onProductionGain()', () => {
```
ryannewton, line 92: Let me know if you think this type of test even makes sense. I could see this being a more fundamental test around game engine behavior.

I just saw [this issue](https://github.com/terraforming-mars/terraforming-mars/issues/7140) and figured adding a test would be a good place to start.

kberg (approved):

PR #7164, opened 2025-01-12
Document what the service worker does

kberg:


Borbarad2:
What is a service worker at all?

PR #7165, opened 2025-01-12
Have some special units spent greedily.

kberg:


kberg:
This also has a couple of useful simplifications that help remove some of the mystery behind this stuff.

Issue #7167, opened 2025-01-13
Cyberia Systems: should it be playable when I have only one suitable card (with building tag and production box)?


kevinb9n:
Since the card doesn't use the phrase "up to two", I would assume that I shouldn't be allowed to play it unless I have two such cards down first. Currently the game is letting me - I have just one option for the "first" card to copy, and then it just finishes the action.

I couldn't find a mention of this on the BGG forums tho.

kberg:
You're right, it shouldn't be playable. I am in a position where I have to either choose to err on being permissive or restrictive. Imagine the case where the first playable card gives the player an energy production which the second card would remove. Or the case where both cards remove 1 unit of energy production when there is only 1 energy production to give.

It's certainly possible to start to dig into that, and make something that works.

I continue to want to make an abstraction representing the player's purse (or production purse) and make it something easy to copy as a way of iterating through a multitude of cases. :/

kberg:
Alright, it'll now only be playable if it at least one card playable right now, and at least two cards with the tags and production boxes. Not ideal, but cheap and slightly better.

PR #7171, opened 2025-01-14
Partially adjust how Cyberia Systems works.

kberg:
Now it will just count the number of cards with building tags and production boxes, and hope there are 2. Are they playable? Can the player afford any costs? Unknown.


kberg:
Related to #7167

PR #7173, opened 2025-01-15
Add director icon.

kberg:
image


Borbarad2:
The text: "This is not the director icon yet" should now be deleted.

PR #7195, opened 2025-02-02
Add handicap to the player view model

Stasmo:
I am starting to collect the player's game data at the end of the match to help determine the best handicap for my friends when we play. Can we add their handicap to the response data please? :)

Reordered a few things because alphabet.

### DIFF
--- a/src/common/models/PlayerModel.ts
+++ b/src/common/models/PlayerModel.ts
@@ -46,6 +46,7 @@ export type PublicPlayerModel = {
   energyProduction: number;
   excavations: number,
   fleetSize: number;
+  handicap: number;
   heat: number;
   heatProduction: number;
   id: PlayerId | undefined;
--- a/src/common/models/PlayerModel.ts
+++ b/src/common/models/PlayerModel.ts
@@ -46,7 +46,7 @@ export type PublicPlayerModel = {
   energyProduction: number;
   excavations: number,
   fleetSize: number;
-  handicap: number;
+  handicap: number | undefined;
   heat: number;
   heatProduction: number;
   id: PlayerId | undefined;
--- a/src/server/models/ServerModel.ts
+++ b/src/server/models/ServerModel.ts
@@ -89,7 +89,14 @@ export class Server {
   public static getPlayerModel(player: IPlayer): PlayerViewModel {
     const game = player.game;
 
-    const players: Array<PublicPlayerModel> = game.getPlayersInGenerationOrder().map(this.getPlayer);
+    const players: Array<PublicPlayerModel> = game.getPlayersInGenerationOrder().map(this.getPlayer)
+
+    if (players.some(p => p.handicap && p.handicap > 0)) {
+      players.forEach(p => {
+        p.handicap = p.handicap || 0;
+      })
+    }
+
     const thisPlayerIndex = players.findIndex((p) => p.color === player.color);
     const thisPlayer: PublicPlayerModel = players[thisPlayerIndex];
 
--- a/src/server/models/ServerModel.ts
+++ b/src/server/models/ServerModel.ts
@@ -89,12 +89,12 @@ export class Server {
   public static getPlayerModel(player: IPlayer): PlayerViewModel {
     const game = player.game;
 
-    const players: Array<PublicPlayerModel> = game.getPlayersInGenerationOrder().map(this.getPlayer)
+    const players: Array<PublicPlayerModel> = game.getPlayersInGenerationOrder().map(this.getPlayer);
 
-    if (players.some(p => p.handicap)) {
-      players.forEach(p => {
+    if (players.some((p) => p.handicap)) {
+      players.forEach((p) => {
         p.handicap = p.handicap || 0;
-      })
+      });
     }
 
     const thisPlayerIndex = players.findIndex((p) => p.color === player.color);
--- a/src/server/models/ServerModel.ts
+++ b/src/server/models/ServerModel.ts
@@ -221,7 +221,7 @@ export class Server {
       energy: player.energy,
       energyProduction: player.production.energy,
       fleetSize: player.colonies.getFleetSize(),
-      handicap: player.handicap,
+      handicap: player.handicap || undefined,
       heat: player.heat,
       heatProduction: player.production.heat,
       id: game.phase === Phase.END ? player.id : undefined,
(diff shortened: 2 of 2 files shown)


src/common/models/PlayerModel.ts
```diff
@@ -76,20 +77,20 @@ export type PublicPlayerModel = {
...
   autopass: boolean;
   cardsInHand: ReadonlyArray<CardModel>;
+  ceoCardsInHand: ReadonlyArray<CardModel>;
+  dealtCeoCards: ReadonlyArray<CardModel>;
   dealtCorporationCards: ReadonlyArray<CardModel>;
   dealtPreludeCards: ReadonlyArray<CardModel>;
   dealtProjectCards: ReadonlyArray<CardModel>;
-  dealtCeoCards: ReadonlyArray<CardModel>;
   draftedCards: ReadonlyArray<CardModel>;
+  handicap: number;
```
Stasmo, line 92: This is one of the 2 actual changes I made

src/server/models/ServerModel.ts
```diff
@@ -94,22 +94,23 @@ export class Server {
...
       cardsInHand: cardsToModel(player, player.cardsInHand, {showCalculatedCost: true}),
       ceoCardsInHand: cardsToModel(player, player.ceoCardsInHand),
+      dealtCeoCards: cardsToModel(player, player.dealtCeoCards),
       dealtCorporationCards: cardsToModel(player, player.dealtCorporationCards),
       dealtPreludeCards: cardsToModel(player, player.dealtPreludeCards),
-      dealtCeoCards: cardsToModel(player, player.dealtCeoCards),
       dealtProjectCards: cardsToModel(player, player.dealtProjectCards),
       draftedCards: cardsToModel(player, player.draftedCards, {showCalculatedCost: true}),
       game: this.getGameModel(player.game),
+      handicap: player.handicap,
```
Stasmo, line 106: This is the other change I made.

kberg:
Hi. That sounds like a nice project you have. Two thoughts, though:

1. Is this the best way to do this? You only need this information at the end of the game, for instance. Should this just be shown at the end?

2. Your PR has many changes that are not related to the work you actually want. I think you have meant to only make small changes. Can you limit your change to meaningful stuff only? If you have some automatic formatter, please turn it off for this project.

Borbarad2:
What kind of handicap do you want to add?
How do you measure the data (and what kind of data) to figure out your handicap?

Stasmo:
The other changes I made are just reordering large lists of properties to be in alphabetical order for better readability and fixing a spelling error. It's just janitorial work, nothing special. If you prefer it not to be in alphabetical let me know and I can undo those changes.

@Borbarad2  what I mean by handicap is that I want to be able to see the TR Boost that a player gets at the beginning of the game in some sort of summary.

@kberg Maybe it makes more sense to see the handicap [here](https://github.com/terraforming-mars/terraforming-mars/blob/main/src/server/models/ServerModel.ts#L43-L47) instead?

kberg:
Let's take a step back and talk about how you plan to access this information. Right now you're asking me to put data into the model that isn't being shown. How are you going to read it?

Stasmo:
Oh yeah I totally forgot to explain that. I just wanted to grab the JSON data from the API for the player or for the game with a  `GET /api/player` or `GET /api/game` and maybe process the data and save it to a Google Sheet or just save the JSON file somewhere to process it later or something.

We're just looking to track who won, how much they won by and what was their handicap, but the handicap data is not sent back to client at any time.

Stasmo:
Also I know a little Vue and can add something to the `GameEnd.vue` so that players can see what their handicap was at the end of the match too.

kberg:
Do you have your own instance of the server running?

Stasmo:
Starting that up now. Reading through the wiki. Is there some seed data file available somewhere for the db?

kberg:
Wait a sec, I'm still trying to understand your requirements. So, do you host your own server? If you do, then you can get the information right out of the database. Let's start there.

Stasmo:
I am not hosting an instance of the app.
I just wanted to be able to see the players handicap via the API at the heroku app endpoint.

kberg:
Okay. Let me think about it.

Borbarad2:
Ah, you don't want to introduce a handicap, you just want to display the given handicap during play on screen somewhere next to the player stats and maybe display it additionally at the endgame screen. 
Okay, understood.

kberg:
Alright. I have a non-disruptive visual solution for including the handicap. Let's add this PR, but please clean it up so it only contains the changes. (Or close this one and open a new one.)

kberg:
It'll look like this

image

And ideally should only appear when the attribute is set. That way in games without it, they won't see it.

Stasmo:
OK I updated the PR. I think it makes more sense to put the handicap in the `PublicPlayerModel` instead of the `PlayerViewModel` as this information is not really a secret. Also it is important for my ego to see that I lost _only_ because my friends got a high handicap.

kberg:
That's good. Here's the one change I will ask you to make, though: Only set
the handicap if at least one player has a non-zero handicap. Otherwise let
it be undefined. So, handicap could be declared `handicap: number |
undefined` and at the start of getPlayer, add `const useHandicap =
game.getPlayers().some((p) => p.handicap !== 0)`. Then line 224 can be:

`player: useHandicap ? player.handicap : undefined`

Stasmo:
Oh sorry, I misunderstood. Will fix.

kberg:
I knew you would get there.

src/server/models/ServerModel.ts
```diff
@@ -90,6 +90,13 @@ export class Server {
     const game = player.game;
 
     const players: Array<PublicPlayerModel> = game.getPlayersInGenerationOrder().map(this.getPlayer);
+
+    if (players.some((p) => p.handicap)) {
```
kberg, line 94: With the change I proposed, `getPlayer` returns an instance of the player model fit for use. In this proposal, `getPlayer` does not. My recommendation was specific and precise for a reason.

Unless you have discovered a bug with my logic. Did you?

Stasmo: Seemed kind of wasteful to do the check for handicap at each `getPlayer` call but I see now that `getPlayer` is called in a bunch of places and I am sensing that you might want all of those function calls to return the handicap if any player has a handicap. Will fix.

kberg (approved):

PR #7236, opened 2025-02-16
Milestones and Awards for fan maps for Vastias Borealus (Novus)

kberg:
image


Borbarad2:
Great!!! Well done. 
Will the MAs for the Utopia map, for the Terra Cimmeria Novus map and the MAs for the Underworld expansion also be coming?!?

PR #7240, opened 2025-02-22
Update pathfinder_cards.json

Borbarad2:
Adapt to new clarifying text. Add the clarification to the german text as well


kberg:
I hope to deal with all of these next week.

PR #7241, opened 2025-02-22
Update .env.sample

PetrGasparik:
taking Discord Id from env

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -109,4 +109,4 @@
 # SESSION_DURATION=30m
 
 # The root URL of this server. Used for auth redirects
-# URL_REDIRECT=http://localhost:8080
+# URL_ROOT=http://localhost:8080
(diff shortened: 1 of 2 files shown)


kberg (approved):
Good catch!

PetrGasparik:
sorry, found another one from Discord era :)

kberg:
Yeah, I think this doesn't work. The environment variable is on the server, not the client.

PetrGasparik:
What do you mean?

kberg:
Well, tell me. Does it work?

PetrGasparik:
the URL_ROOT do work

kberg:
I'm skeptical. If it does work, it's because the browser is on the same
machine as the server. But I'll look some more

kberg:
I'll assemble something for you. This PR can probably be closed since most
of the work won't be necessary. But, thank you for identifying it!

Actually, if you roll back the .vue change, and leave the `.env.sample`
change, I"ll merge it.

PetrGasparik:
Done and thanks!

kberg (approved):
Thanks! I haven't forgotten about the other work

PR #7253, opened 2025-03-04
Update promo_cards.json

Borbarad2:
Add new promo card


Borbarad2:
Merge it into an other commit.

PR #7283, opened 2025-04-05
Visual cue that Labor Trafficking's effect was used

kberg:
Fixes https://github.com/terraforming-mars/terraforming-mars/issues/7280


kberg:
image


Also, slightly cleaned the card so it fits

PR #7316, opened 2025-04-30
Trim english strings before i18n

alrusdi:


kberg (approved):

Borbarad2:
Great it works now better.

Issue #7340, opened 2025-05-13
Missing Action arrow in FloatingTradeHub card


JestemStefan:
We noticed that card  is missing Action arrow which caused confusion during the game until we pulled original card out of the internet.

![Image](https://github.com/user-attachments/assets/dea1a8cd-c463-48ee-9901-5612e40fc4a4)

I'm not familiar with the code, but I checked and it seems that line

` b.action('Add 2 floaters to ANY card.', (ab) => ab.startAction.resource(CardResource.FLOATER, 2).asterix()).br;`

is missing `.empty()` function call, because all other similar cards have it.

Link to the code:

https://github.com/terraforming-mars/terraforming-mars/blob/324d4758b68ffe0035c4928da41f79538e7ef6df/src/server/cards/prelude2/FloatingTradeHub.ts#L25

kberg:
Good work. Thank you!

PR #7341, opened 2025-05-13
Fix Action arrow rendering in Floating Trade Hubs card

JestemStefan:
Fix for #7340


kberg (approved):
Thank you! Good catch, and good fix!

PR #7344, opened 2025-05-15
Update ares_cards french

Borbarad2:
Advance french ares cards on behalf of @ramdog


Borbarad2:
Probably this and further translations are not needed anymore, due to this larger PR7347. Thanks.

PR #7347, opened 2025-05-17
French translation update from DE translation

nicobubulle:
Phase 1 = Cards from offcial game


kberg:
Thank you!

> Phase 1 = Cards from offcial game

It looks like you included much more than the cards from the official game.

nicobubulle:
Yes I have added all strings from DE translation but only translated/checked cards from the official game. Other translated string are from previous translation. When no translation were available, I left the string in English. It's easer to me for tracking what I have to translate.
In phase 2, I will translate UI and phase 3, community cards.

kberg:
OK. Thank you again. This is a good PR.

kberg (approved):

Issue #6613, opened 2024-04-19
Corporation cards description text is hidden when the corp box is too big


kberg:
Here's a case where Celestic's description is hidden.

![image](https://github.com/terraforming-mars/terraforming-mars/assets/413481/7ab635e3-3021-4eb4-af30-218c34235e92)

Here's a case where Project Workshop's text is completely not visible.

![image](https://github.com/terraforming-mars/terraforming-mars/assets/413481/712f36bb-8cab-4032-9609-57967901065d)

Look, with a mild HTML hack that removes part of thew corp box, it appears: 
![image](https://github.com/terraforming-mars/terraforming-mars/assets/413481/85bd69bd-222f-4ce3-80e9-8adc56a8bbfa)

kberg:
Fixed by #7366

PR #7366, opened 2025-05-31
#6613

katieyang:
Problem:
Some corporation cards get their description text hidden.

Result of fix looks like this:
BEFORE
image

AFTER (Now it scrolls)
image
image

BEFORE
image

AFTER (Now it scrolls)
image
image

Here is a rough description of the proposed fix.

1.	Moved card-description between card-rows. I added something to make this only apply to corporation cards, because I didn't want to accidentally mess up anything else. 
2.	Made card-corporation box position: relative. Got rid of the left: 8px and bottom: 8px.
3.	To stop the overlap of content with the card-title, I changed the height of the card-title to 40px. I added something to make this only apply to corporation cards, because I didn't want to accidentally mess up anything else.
4.	To not have the content overflow out of the bottom of the card, added padding-bottom: 0px to .card-content.

I'm relatively new to contributing so if I did anything wrong just let me know and am happy to fix/change.

### DIFF
--- a/src/client/components/card/CardContent.vue
+++ b/src/client/components/card/CardContent.vue
@@ -1,8 +1,10 @@
 <template>
   <div :class="getClasses()">
-    <CardRequirementsComponent v-if="requirements.length > 0" :requirements="requirements"/>
-    <CardRenderData v-if="metadata.renderData" :renderData="metadata.renderData" />
-    <CardDescription v-if="hasDescription" :item="metadata.description" />
+    <CardRequirementsComponent v-if="requirements.length > 0" :requirements="requirements"/>      
+    <CardRenderData v-if="metadata.renderData && metadata.renderData.rows.length > 0" :renderData="{ rows: [metadata.renderData.rows[0]] }" />
+    <CardDescription v-if="isCorporation && hasDescription" :item="metadata.description"/>
+    <CardRenderData v-if="metadata.renderData.rows.length > 1" :renderData="{ rows: metadata.renderData.rows.slice(1) }" />
+    <CardDescription v-if="!isCorporation && hasDescription" :item="metadata.description"/>
     <CardVictoryPoints v-if="metadata.victoryPoints" :victoryPoints="metadata.victoryPoints" />
     <div class="padBottom" v-if="padBottom" style="padding-bottom: 22px;"></div>
   </div>
--- a/src/client/components/card/CardTitle.vue
+++ b/src/client/components/card/CardTitle.vue
@@ -1,5 +1,5 @@
 <template>
-  <div :class="getMainClasses()">
+  <div :class="[getMainClasses(), { 'is-corporation': isCorporation() }]">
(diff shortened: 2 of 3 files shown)


kberg:
Thanks!

src/client/components/card/CardContent.vue
```diff
@@ -1,8 +1,10 @@
 <template>
   <div :class="getClasses()">
-    <CardRequirementsComponent v-if="requirements.length > 0" :requirements="requirements"/>
-    <CardRenderData v-if="metadata.renderData" :renderData="metadata.renderData" />
-    <CardDescription v-if="hasDescription" :item="metadata.description" />
+    <CardRequirementsComponent v-if="requirements.length > 0" :requirements="requirements"/>  
+    <CardRenderData v-if="metadata.renderData && metadata.renderData.rows.length > 0" :renderData="{ rows: [metadata.renderData.rows[0]] }" />
```
kberg: What exactly is in `rows[0]` that merits having it move up like that? Eventually I could see updating the `renderData` structure such that `rows[0]` is its own field.

src/client/components/card/CardTitle.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div :class="getMainClasses()">
+  <div :class="[getMainClasses(), { 'is-corporation': isCorporation() }]">
```
kberg, line 2: Why is this a better change than updating `getMainClasses`?

kberg (commented):
Looks terrific. I have a couple of questions that probably shouldn't block submission, but let's get them answered first.

src/client/components/card/CardContent.vue
katieyang: Ah I can explain this. Rows[0] is the "cost" of the card. What I'm trying to do here is pushing the card description below the cost but above the next card row, which is card-corporation-box. So it's not rows[0] moving up, but rather moving up card-description to just below rows[0]. Otherwise, it won't be positioned properly.

It would indeed have been easier/more convenient to do this if it was its own field. The code would be more readable. I just wasn't sure how to do it.

image

src/client/components/card/CardTitle.vue
katieyang: No specific reason, this was just how it crossed my mind to do it. If updating getMainClasses() is preferred, I can do that instead while keeping the logic the same, just let me know.

kberg: Thank you, no, leave it for now.

src/client/components/card/CardContent.vue
kberg: Thanks. I will look at this some more.

kberg (approved):

kberg:
The reason the code won't compile is because of Typescript. Yay! Also, grr.

On the positive side, I made things type-compatible, and also arguably easier to read. (At least, the hard work is in Typescript and not embedded in the HTML. Please fetch the latest version and see what you think.

Issue #7374, opened 2025-06-05
Server: The parameter .env > GAME_CACHE > sweep_freq is ignored


kselaenne:
**When** I set `GAME_CACHE=sweep=auto;eviction_age=24h;sweep_freq=1h` in `.env`
**Then** sweeping still takes place according to the default value of 5m (Log: _Sweeper sleeping for 300000ms_).

**Expected:** Sweeping takes place once an hour (Log: _Sweeper sleeping for 3600000ms_).

**Reason:** Inverted condition in [GameLoader.ts#L191](https://github.com/terraforming-mars/terraforming-mars/blob/main/src/server/database/GameLoader.ts#L191)

**Pull request:** #7375

kberg:
Fixed.

PR #7375, opened 2025-06-05
Fix condition to take GAME_CACHE's sweep_freq into account when it is set

kselaenne:
Issue #7374


src/server/database/GameLoader.ts
```diff
@@ -188,6 +188,6 @@ function parseConfigString(stringValue: string): CacheConfig {
   const evictMillis = durationToMilliseconds(parsed.eviction_age);
   if (!isNaN(evictMillis)) options.evictMillis = evictMillis;
   const sleepMillis = durationToMilliseconds(parsed.sweep_freq);
-  if (isNaN(sleepMillis)) options.sleepMillis = sleepMillis;
+  if (!isNaN(sleepMillis)) options.sleepMillis = sleepMillis;
```
kberg, line 191: How embarrassing. Thank you.

kberg (approved):
I'd love to write a test for this. I'll do it later since it's slightly harder than trivial.

Issue #7376, opened 2025-06-05
The parameter .env > HOST is ignored


kselaenne:
**When** I set `HOST=localhost` in `.env`
**Then** the server still listens to all traffic on PORT.

**Expected:** The server only listens to localhost:PORT.

**Use case:** When using Nginx as reverse proxy, I want to expose the server only via Nginx, but the current behavior also exposes it globally as an HTTP server on PORT.

**Reason:** `process.env.HOST` is not referenced anywhere.

PR #7377, opened 2025-06-05
Add HOST .env variable to server.listen(port, host)

kselaenne:
Issue #7376 

- Maintain existing behavior by using '::' as default host to listen to all IPv4 and IPv6 addresses
  - As far as I understand, on the vast majority of systems '::' also encompasses '0.0.0.0', but this is not 100% given. On systems where that is not the case, the existing behavior with the unspecified host in `server.listen(port)` would _not_ be preserved.
- Add localhost as suggested HOST to .env.sample
  - However, this is not the default value, so this could be confusing? Would it be better to leave it empty?
- `server.listen(port)` with `port = process.env.PORT` is also present in [analyze_ma.ts](https://github.com/terraforming-mars/terraforming-mars/blob/main/src/server/tools/analyze_ma.ts#L40)
  - I'm not sure what analyze_ma.ts is used for, but it could be that the HOST should also be applied there analogously

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -7,7 +7,8 @@
 # PORT=8080
 
 ## What hostname to use
-# HOST=
+## When this is unspecified, the server listens to all network interfaces
+# HOST=localhost
 
 ## Your TLS certificate path (=> `fullchain.pem` created by certbot)
 # CERT_PATH=
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -99,12 +99,17 @@ async function start() {
   }
   GameLoader.getInstance().maintenance();
 
-  const port = process.env.PORT || 8080;
-  const host = process.env.HOST || '::';
   console.log(`Starting ${raw_settings.head}, built at ${raw_settings.builtAt}`);
-  console.log(`Starting server listening to ${host} on port ${port}`);
 
-  server.listen({'port': port, 'host': host});
+  const port = process.env.PORT || 8080;
+  const host = process.env.HOST;
+  if (host) {
+    console.log(`Starting server listening to ${host} on port ${port}`);
+  } else {
+    console.log(`Starting server on port ${port}`);
+  }
+
+  server.listen({port: port, host: host});
 
   if (!process.env.SERVER_ID) {
     console.log(`The secret serverId for this server is ${ansi.style.bold}${serverId}${ansi.style.reset}.`);
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -104,7 +104,7 @@ async function start() {
   console.log(`Starting ${raw_settings.head}, built at ${raw_settings.builtAt}`);
   console.log(`Starting server listening to ${host} on port ${port}`);
 
-  server.listen({"port": port, "host": host});
+  server.listen({'port': port, 'host': host});
 
   if (!process.env.SERVER_ID) {
     console.log(`The secret serverId for this server is ${ansi.style.bold}${serverId}${ansi.style.reset}.`);
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -100,10 +100,11 @@ async function start() {
   GameLoader.getInstance().maintenance();
 
   const port = process.env.PORT || 8080;
+  const host = process.env.HOST || '::';
   console.log(`Starting ${raw_settings.head}, built at ${raw_settings.builtAt}`);
-  console.log(`Starting server on port ${port}`);
+  console.log(`Starting server listening to ${host} on port ${port}`);
 
-  server.listen(port);
+  server.listen({"port": port, "host": host});
 
   if (!process.env.SERVER_ID) {
     console.log(`The secret serverId for this server is ${ansi.style.bold}${serverId}${ansi.style.reset}.`);


.env.sample
```diff
@@ -7,7 +7,8 @@
 # PORT=8080
 
 ## What hostname to use
-# HOST=
+## When this is unspecified, the server listens to all network interfaces
```
kberg, line 10: Listens to all network interfaces? I'm reading the documentation at https://nodejs.org/api/net.html#serverlistenport-host-backlog-callback which says

> If host is omitted, the server will accept connections on the [unspecified IPv6 address](https://en.wikipedia.org/wiki/IPv6_address#Unspecified_address) (::) when IPv6 is available, or the [unspecified IPv4 address](https://en.wikipedia.org/wiki/0.0.0.0) (0.0.0.0) otherwise.

kberg (commented):
Thank you. I did not realize this has been removed. (Though I might have been the one to remove it.)

Here's a question: is this value still useful? I assume it is. If your bug report describes why it is useful,ignore this question.

kberg:
> Issue #7376
> * Add localhost as suggested HOST to .env.sample
>   * However, this is not the default value, so this could be confusing? Would it be better to leave it empty?

As long as `server.listen({host: ""})` is the same as not supplying it at all.

> * `server.listen(port)` with `port = process.env.PORT` is also present in [analyze_ma.ts](https://github.com/terraforming-mars/terraforming-mars/blob/main/src/server/tools/analyze_ma.ts#L40)
>   * I'm not sure what analyze_ma.ts is used for, but it could be that the HOST should also be applied there analogously

It's OK to ignore it. It starts a local web server that is used to test random milestone and award combinations.

.env.sample
kselaenne: Yeah, that's what I basically refer to in the comment, but I wasn't sure how to word it well for the configuration.

When HOST is not specified (`process.env.HOST === undefined`), I take '::' as the value to pass to `server.listen` (which also seems to work for IPv4 equivalently to '0.0.0.0'), and thus the behavior should be the same as it has been until now with the omitted host parameter in `server.listen`.

However, now that I think of it, it could be slightly safer to restructure the code to omit the host parameter passed to `server.listen` altogether when `process.env.HOST === undefined`. That would avoid having to attempt to "reimplement" the default behavior of `server.listen` when the host is omitted to maintain backward compatibility.

kselaenne: I took a closer look at [net.js#L1062,L1065](https://github.com/nodejs/node/blob/v24.1.0/lib/net.js#L1062,L1065) and did some tests, and it doesn't seem to make a difference whether you omit host or pass a falsy value (= including undefined). On the other hand, passing '::' is already more specific and should be avoided, I think, so I'll remove that fallback.

kselaenne:
> As long as `server.listen({host: ""})` is the same as not supplying it at all.

This is indeed the case and extends to all falsy values.

I was just wondering whether there is a convention in `.env.sample` to only have default values as commented presets except for explicitly labeled examples (as for GAME_CACHE). If there is no such convention, then I think it makes sense to have localhost as the "default custom value" available as soon as the line is uncommented.

.env.sample
kberg: I trust your judgment here.

kberg:
> > As long as `server.listen({host: ""})` is the same as not supplying it at all.
> 
> This is indeed the case and extends to all falsy values.
> 
> I was just wondering whether there is a convention in `.env.sample` to only have default values as commented presets except for explicitly labeled examples (as for GAME_CACHE). If there is no such convention, then I think it makes sense to have localhost as the "default custom value" available as soon as the line is uncommented.

That's fine. Thanks.

.env.sample
kberg: But ... wait, sorry, I still don't understand: what does "listens to all network interfaces" mean?

Did you already say, above, and I'm just daft?

kselaenne: It's not you, it's probably poorly worded and should be changed to something more understandable.

Basically the problem is that with no specified host, the server is globally accessible on PORT regardless of what address is used to access it. Say I host it on port 8080 on a machine having some IP address 123.456.789.123, to which I also have a DNS entry "mydomain.com". Consequently, I can access it externally via http://123.456.789.123:8080 and http://mydomain.com:8080, and on the machine itself, it's of course also accessible via http://localhost:8080 or http://127.0.0.1:8080. Now my use case (as briefly described in the bug report) is having Nginx act as a reverse proxy relaying traffic from my chosen subdomain https://sub.mydomain.com:443 to http://localhost:8080. Therefore, it becomes superfluous (and increases attack surface) to have the server _also_ respond to the aforementioned two HTTP addresses on port 8080. It should only listen to localhost, which is served by Nginx.

Would the following make more sense, or do you have a better idea?
`## When this is unspecified, the server listens to all addresses`
Technically the hostname behavior comes 1:1 from NodeJS and would not need further clarification for users familiar with NodeJS, but not all users necessarily know how NodeJS works (I didn't before setting up this server), so I think it's good to have some kind of hint trying to explain what the parameter does (even if in the end, one may have to figure it out through trial and error).

kberg:
Thanks. Thinking about this some, there's a problem with defaulting to
localhost, which is that the current implementation is that HOST is not
required, or, as you noticed, even used. My implementation on Heroku does
not set HOST, and I worry that adding a default could break it.

Actually: does this help? https://gemini.google.com/share/21119f1a8a63

kselaenne:
Yes, I've considered that. That's why my intention would be to leave HOST in .env _commented_, so that it is unspecified by default, leading to the same behavior as until now and thus maintaining backward compatibility. What I meant with "default" is having localhost as a _suggested_ value for HOST _if_ the line is simply uncommented. Currently there is no suggested value (`HOST=`), and I think adding an example/suggestion would illustrate what it might be useful for. I hope this clarifies what I meant!

kselaenne:
> Actually: does this help? https://gemini.google.com/share/21119f1a8a63

That actually also very well describes my issue with the missing host in `server.listen`:
> This default behavior is often convenient for development, as it means you don't have to worry about specifying the exact IP address of your machine. It allows you to **access the server from localhost, your machine's local IP address, or any other IP address assigned to your machine**, provided the client can route to it.

> You might want to explicitly supply the host parameter in scenarios like:
> - **Security:** If you only want your server to be accessible from localhost (127.0.0.1) for local development or **internal communication [kselaenne: such as reverse proxy]**, you would specify server.listen(port, '127.0.0.1') or server.listen(port, 'localhost'). **This prevents external access.**

kberg (approved):

PR #7379, opened 2025-06-05
Communication center solar storm interaction

TarunTheo13:
Communication Center still automatically draws a card before Solar Storm has the opportunity to remove data, have created the PR anyhow so that you can point me in the right direction.


kberg:
Can you clarify what problem you are actually solving? You're solving an interaction between Communication Center and Solar Storm. What's the problem?

Also, it looks like this PR has a lot of commits that are unrelated (though related to your prior PR.) Do you think you could clean it up a bit, to make the PR easier to review?

src/server/cards/pathfinders/CommunicationCenter.ts
```diff
@@ -40,13 +40,18 @@ export class CommunicationCenter extends Card implements IProjectCard {
...
-    while (this.resourceCount >= 3) {
-      this.resourceCount -= 3;
-      player.drawCard(1);
-      player.game.log('${0} automatically removed 3 data from ${1} to draw a card.', (b) => {
-        b.player(player).card(this);
-      });
-    }
+
+    // 
+    const priority = (playedCard.name === CardName.SOLAR_STORM) ? Priority.BACK_OF_THE_LINE : Priority.DEFAULT;
```
kberg: I think instead, the change goes in `onCardPlayedByAnyPlayer`.

if `playedCardOwner !== thisCardOwner`, reduce the priority.

You might be tempted to make the current `const priority` a longer conditional, but instead use a `let` with a default priority that is changed based on an `if`/`elseif` that are well documented.

Before you make this change, write the test, and verify it fails. There's a ready a test that shows how this is compatible with CEO's favorite Project. Let me know if you want help with that.

TarunTheo13:
> Can you clarify what problem you are actually solving? You're solving an interaction between Communication Center and Solar Storm. What's the problem?
> 
> Also, it looks like this PR has a lot of commits that are unrelated (though related to your prior PR.) Do you think you could clean it up a bit, to make the PR easier to review?

Cleaned it up now

src/server/cards/pathfinders/CommunicationCenter.ts
TarunTheo13: I can do that, but I'm not sure I can accomplish the initial plan through the `onCardPlayedByAnyPlayer ` function, because:

The initial plan was to be able to 'freeze' Communication Center on three data and then remove it with Solar Storm before automatically removing it to draw a card. So that logic would have to happen in `onResourceAdded`, because otherwise I can only delay the adding of data.

kberg:
Here is a test you can use

```
  it('Works with Solar Storm', () => {
    // Solar storm removes data from another player. Since the other player can
    // control order of operations, they get to remove data from Communication Center
    // before this player can remove them.
    card.resourceCount = 2;
    player.playedCards = [card];

    const solarStorm = new SolarStorm();
    player2.playCard(solarStorm);

    runAllActions(game);

    cast(player.popWaitingFor(), undefined);
    const orOptions = cast(player2.popWaitingFor(), OrOptions);
    const selectCard = cast(orOptions.options[0], SelectCard);
    selectCard.cb([card]);
    orOptions.cb(undefined);

    expect(card.resourceCount).eq(0);

    runAllActions(game);

    expect(player.cardsInHand).has.length(0);
    expect(card.resourceCount).eq(1);
  });

```

TarunTheo13:
> Here is a test you can use
> 
> ```
>   it('Works with Solar Storm', () => {
>     // Solar storm removes data from another player. Since the other player can
>     // control order of operations, they get to remove data from Communication Center
>     // before this player can remove them.
>     card.resourceCount = 2;
>     player.playedCards = [card];
> 
>     const solarStorm = new SolarStorm();
>     player2.playCard(solarStorm);
> 
>     runAllActions(game);
> 
>     cast(player.popWaitingFor(), undefined);
>     const orOptions = cast(player2.popWaitingFor(), OrOptions);
>     const selectCard = cast(orOptions.options[0], SelectCard);
>     selectCard.cb([card]);
>     orOptions.cb(undefined);
> 
>     expect(card.resourceCount).eq(0);
> 
>     runAllActions(game);
> 
>     expect(player.cardsInHand).has.length(0);
>     expect(card.resourceCount).eq(1);
>   });
> ```

`solarStorm.play(player2);` does not trigger a resource to be added to Communication Center (which is why there was a false positive
`player2.onCardPlayed(solarStorm);` means `player2.getWaitingFor()` returns `undefined` instead of an instance of `OrOptions`
What's the work-around/solution for dealing with this?

TarunTheo13:
> That's why my example uses `player2.playCard(solarStorm);`

Ahh okay I see now, sorry I missed that. I'm getting an acceptable failing test now. Do you have any direction on how to get the `PRIORITY` working as intended?

kberg:
From Communication Center

```
const priority = (card.name === CardName.CEOS_FAVORITE_PROJECT) ? Priority.BACK_OF_THE_LINE : Priority.DEFAULT;
```

Basically the same thing, if it's Solar Storm, put it in the back.

kberg:
Look at the last commit. https://github.com/terraforming-mars/terraforming-mars/pull/7379/commits/ba0037660af8a40caf098de20a397da59535e5e9 I think this does it.

kberg:
I see. I guess we have different ideas about what "automatically removing 3 data" is like. In my mind, as soon as a third data is put on the card, the effect triggers immediately. But I can see that you think "automatically" is not right away, but just an additional effect which the opponent should control. Does that sum it up?

I am probably biased by intimately understanding how the engine works.

That said, what do you think about calling this good enough? The Solar Storm player gets to prevent the card draw. Here's a compromise: commit this, and then start another PR that changes onResourceAdded that goes from

```
    while (this.resourceCount >= 3) {
      this.resourceCount -= 3;
      player.drawCard(1);
      player.game.log('${0} automatically removed 3 data from ${1} to draw a card.', (b) => {
        b.player(player).card(this);
      });
    }
```

to something like

```
  if (this.resourceCount >= 3) {
   game.defer(() => {
    while (this.resourceCount >= 3) {
      this.resourceCount -= 3;
      player.drawCard(1);
      player.game.log('${0} automatically removed 3 data from ${1} to draw a card.', (b) => {
        b.player(player).card(this);
      });
    }
  }), Priority.SOME;
 }
```

TarunTheo13:
It does some it up. Part of the problem is that there is no other card that automatically removes resources from itself so it is unprecedented functionality.

I have another idea actually for optimal functionality (while respecting resources being removed automatically). I'll commit it as a separate branch and you can review

PR #7392, opened 2025-06-10
Take VP out of the scroll.

kberg:
This requires moving some of the card CSS out some of the class heirarchy. It generally looks OK but there may still be issues.

image


kberg:
image

PR #7406, opened 2025-06-14
Fix: Corporation Visualizations

katieyang:
Before fix:
image

After fix:
image

Brief description of fix:
- part of the issue was that there were empty card rows with margin, so I made those have 0 margin
- also gave the card-corporation-box a bit more top margin to match the rest of the card

Note: This fix is not specific to this card, but a more general fix that will impact all corporation cards. Credicor is just an example of the issue.

### DIFF
--- a/src/client/components/card/CardCorporationLogo.vue
+++ b/src/client/components/card/CardCorporationLogo.vue
@@ -87,7 +87,7 @@
       </div>
     </template>
     <template v-else-if="title === CardName.UNITED_NATIONS_MARS_INITIATIVE">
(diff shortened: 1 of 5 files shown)


katieyang:
> Hi, Perfect what you do! Do you have discord account?

Same name as my GitHub. I'm in the TM Web App Server.

katieyang:
Also fixed UNMI. 

Before fix:
image

After fix:
image

katieyang:
Also fixed this one.

Before fix:
image

After fix: 
image

PR #7416, opened 2025-06-18
Warning for playing microbe tags as PharmU

EthanDobbs:
Simple check added in the player's canPlay method


kberg (approved):
This is nice. Thank you.

PR #7417, opened 2025-06-18
Removed Beta and "New" symbols for official expansions

EthanDobbs:
Also moved the prelude 2 expansion to after turmoil in the create-game menu, to be in chronological order of release. This has always bothered me

### DIFF
--- a/src/client/components/create/CreateGameForm.vue
+++ b/src/client/components/create/CreateGameForm.vue
@@ -40,12 +40,6 @@
                                 <span v-i18n>Prelude</span>
                             </label>
 
-                            <input type="checkbox" name="prelude2" id="prelude2-checkbox" v-model="expansions.prelude2">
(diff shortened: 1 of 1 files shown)


src/client/components/create/CreateGameForm.vue
```diff
@@ -64,10 +58,16 @@
                                 <span v-i18n>Turmoil</span>
                             </label>
 
+                            <input type="checkbox" name="prelude2" id="prelude2-checkbox" v-model="expansions.prelude2">
```
kberg: The reason they're in that order is because they are so closely correlated. I agree that for the most part, temporal ordering makes sense, this seems like a useful exception. Would you mind putting it back?

Borbarad2:
Same could be done for the Star Wars Expansion: Removed B(beta) and new

Issue #7412, opened 2025-06-17
Carbon Nanosystems does not trigger when playing corporations


kberg:
Carbon Nanosystems gains resources whenever a science tag is played.

There are a couple of cases where Carbon Nanosystems gets played before a corp with a science icon (e.g. Inventrix.):
1. Play prelude Eccentric Sponsors (play a card at a discount.)
2.  Use this to play Carbon Nanosystems.
3. Play Merger.
4. Choose Inventrix.

Another way:
1. Play Carbon Nanoystems during a game
2. Use Board of Directors to draw preludes
3. Use that to draw and play Merger
4. Choose Inventrix.

Inventrix has a science tag but Carbon Nanosystems is not informed.

This is an infrastructural problem, which is once again tied to the strange way corporation cards seem distinct.

PR #7421, opened 2025-06-20
Card API Rewrite

kberg:
This is the first working version. More tests are necessary as is more documentation.

This cleans up some API inconsistencies which will further simplify the code base, and allow for new optimizations.
This WILL break forked implementations with custom cards, and require small migration steps.


kberg:
Fixes #7412

PR #7424, opened 2025-06-20
Ceos drafting

TarunTheo13:


src/server/Game.ts
```diff
@@ -1805,13 +1805,15 @@ export class Game implements IGame, Logger {
...
         case 1:
           newInitialDraft(game).restoreDraft();
-          case 2:
-            newInitialDraft(game).restoreDraft();
-          case 3:
-            newPreludeDraft(game).restoreDraft();
-          case 4:
-            newCEOsDraft(game).restoreDraft();
-
+          break;
```
kberg, line 1768: Oof, this must have caused some problems.

src/server/IGame.ts
```diff
@@ -240,9 +240,8 @@ export interface IGame extends Logger {
    * Drafting before the first generation goes through 3 iterations:
    * 1. first 5 project cards,
    * 2. second 5 project cards
-   * 3. [optional] preludes.
-   *
-   * This works, but makes it hard to add a CEO draft.
+   * 3. [optional] preludes (or CEOs, if draft option is selected, but not for preludes)
```
kberg: The word [optional] in front carries the weight of the entire additional content in parenthesis. In other words, let's keep the content short.

src/server/IGame.ts
```diff
@@ -240,9 +240,8 @@ export interface IGame extends Logger {
    * Drafting before the first generation goes through 3 iterations:
```
kberg: This should be "4" iterations now. Also can you please change it specifically to "up to 4"?

TarunTheo13:
CEOs drafting is working when prelude drafting is not selected. Can you check if you're happy with the approach, I think I went for the clearest option.

Also let me know if you want me to write any tests for it (or if anything else needs doing). If not, we should be ready to merge :)

src/server/IGame.ts
TarunTheo13: ✔️

src/server/IGame.ts
TarunTheo13: ✔️

src/server/Game.ts
TarunTheo13: What caused the problems was the `draftRound` variable. I didn't realise it needed to be reset to 1 and spent an hour and a half trying to work out why the CEO cards weren't showing when it was time to draft them 😅

kberg: OK, good one.

TarunTheo13:
> > CEOs drafting is working when prelude drafting is not selected. Can you check if you're happy with the approach, I think I went for the clearest option.
> > Also let me know if you want me to write any tests for it (or if anything else needs doing). If not, we should be ready to merge :)
> 
> There are tests in drafting.spec.ts, and there should be at least one test for CEOs with no other drafting (e.g. no project card, or prelude draft.)
> 
> Why isn't CEO drafting working with prelude drafting?

Sorry, I should have been more clear, it's working with prelude drafting and without prelude drafting :)

kberg:
There is now a test. This can go in today. (The new logging stuff can go in tomorrow.) I'd rather not deploy both changes on the same day.

PR #7433, opened 2025-06-22
Removing non-translations for de.

kberg:


Borbarad2:
Please don't do these PR for all language. 
These non-translations have and make sense: These are for completness and consistency and structuring. These are for find all the non-translations via debug mode, and when you remove all these, then these will be shown in the debug mode as non-translate, which make then finding real not translate strings impossible. 
You can remove duplicates or not used strings in english, and then the translations string as well, but not all in generall. Please don't merge these non-translations PR for all languages.

Borbarad2:
Or even with names: it makes a difference in German if you write names in capital letters or not:
as example:
    "tharsis": "Tharsis",
    "hellas": "Hellas",
    "elysium": "Elysium",
    "t. cimmeria": "T. Cimmeria",
    "tharsis": "",
    "hellas": "",
    "elysium": "",
    "t. cimmeria": "",

Please do not remove theses!!!!!!

PR #7456, opened 2025-06-25
New logging for resources and production.

kberg:
Much more natural.

image
image

### DIFF
--- a/src/server/LogHelper.ts
+++ b/src/server/LogHelper.ts
@@ -115,11 +115,11 @@ export class LogHelper {
     player.game.log('${0} stole ${1} ${2} from the neutral player', (b) => b.player(player).number(amount).string(resource));
   }
 
-  public static logUnitDelta(
+  static logUnitDelta(
     player: IPlayer,
     resource: Resource,
     amount: number,
-    unitType: 'production' | 'amount',
+    production: boolean,
     from: From | undefined,
     stealing = false,
   ) {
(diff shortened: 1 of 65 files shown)


src/server/LogHelper.ts
```diff
@@ -113,4 +114,74 @@ export class LogHelper {
   static logStealFromNeutralPlayer(player: IPlayer, resource: Resource, amount: number) {
     player.game.log('${0} stole ${1} ${2} from the neutral player', (b) => b.player(player).number(amount).string(resource));
   }
+
+  static logUnitDelta(
```
TarunTheo13, line 118: Does this cover stealing microbes/animals/other resources?

kberg: No, just standard resources.

src/common/utils/utils.ts
```diff
@@ -184,3 +184,14 @@ export function toName<T>(item: {name: T}): T {
 export function toID<T>(item: {id: T}): T {
   return item.id;
 }
+
+export function getEnumKeys<T extends Record<string, any>>(enumObject: T): Array<T[keyof T]> {
```
Argeento:
```suggestion
export function getEnumKeys<T extends Record<string, any>>(enumObject: T): Array<keyof T> {
```
`Array<T[keyof T]>` returns the enum’s values instead of its keys

src/common/utils/utils.ts
```diff
@@ -184,3 +184,14 @@ export function toName<T>(item: {name: T}): T {
...
+  return Object.keys(enumObject) as Array<T[keyof T]>;
+}
+
+export function getEnumStringValues<T extends Record<string, string>>(enumObject: T): Array<string> {
```
Argeento:
```suggestion
export function getEnumStringValues<T extends Record<string, string>>(enumObject: T): Array<T[keyof T]> {
```

Using `Array<T[keyof T]>` preserves the literal enum value types (e.g. 'RED' | 'GREEN' | 'BLUE') instead of a general string[] 

(or is the type-narrowing here intentionally sacrificed? 🤔)

src/common/utils/utils.ts
kberg: These wound up unused, so I'm removing them altogether.

PR #7457, opened 2025-06-25
Fix polish translations

Argeento:
Fix translation issues:
- amount vs. number
- anglicism-related translations
- various typos


kberg:
Note that with https://github.com/terraforming-mars/terraforming-mars/pull/7456, resource and production changes will log differently. So for instance, the change you just made to  ` "${0}\u2019s ${1} production ${2} by ${3} by ${4}": "Ilość produkcji ${0} została zmieniona o ${1} za sprawą ${3}",` will need another update. I'm not sure what to do other than let the translations become invalid.

PR #7473, opened 2025-06-30
More German updates

Borbarad2:
Made a mistake myself, correct it now.
More corrections

### DIFF
--- a/src/locales/de/ui.json
+++ b/src/locales/de/ui.json
@@ -1025,7 +1025,7 @@
   "7 oceans": "7 Ozeane",
   "8 oceans": "8 Ozeane",
   "oceans": "Ozeane",
-  "°C": "°C",
+  "°C": "",
   "3% oxygen": "3% Sauerstoffgehalt",
   "4% oxygen": "4% Sauerstoffgehalt",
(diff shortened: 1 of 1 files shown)


src/locales/de/underworld_cards.json
```diff
@@ -7,8 +7,7 @@
 "Requires that you have no more than 0 corruption. ALL players lose 2 corruption each.": "Benötigt, dass du nicht mehr als 0 Korruption hast. ALLE Spieler verlieren jeweils 2 Korruption.",
 
 "Artesian Aquifer": "Artesischer Grundwasserleiter",
-"// Place an ocean, then excavate the underground resource in its space, if possible.": "Platziere einen Ozean und grabe dann, wenn möglich, die unterirdische Ressource in seinem Feld aus.",
```
kberg, line 10: I'll accept these but they're the last underworld translation changes I'm going to accept for a while.

Borbarad2:
Thanks anyway. I just get along with your underworld 2 changes.
If you not accept underworld 2 changes anymore, I need to collect them somewhere else, otherwise I would get lost for your many changes you need to do along with the underworld 2 development.

kberg (approved):

PR #7488, opened 2025-07-03
Adding Corporation Boxes to Some Corporations

katieyang:
Before 
image

After
image

Before 
image

After
image

Planning to do this for the remaining corps that don't have boxes still, so if you keep this PR open I'll just keep adding to it.


kberg:
One of the reasons the corp action / effect boxes isn't included is it pushes too much content into the overflow. So, I'm not sure every one of those cards deserves the change, you know?

I've wondered if there was a way to expand a card to show it fully, rather than just scroll the card. But do that without changing the size of cards in play.

PR #7499, opened 2025-07-04
Add effect action corp box and implement it on Palladin Shipping

TarunTheo13:
Opening this PR for discussion on this. The change works (as shown below) and does not create an overflow on Palladin Shipping. Should I attempt a refactor on how corp boxes are implemented?
Palladin Shipping

### DIFF
--- a/src/client/components/card/CardRowComponent.vue
+++ b/src/client/components/card/CardRowComponent.vue
@@ -16,8 +16,10 @@ import {isIDescription} from '@/common/cards/render/ICardRenderDescription';
 import {
   ICardRenderCorpBoxAction,
   ICardRenderCorpBoxEffect,
+  ICardRenderCorpBoxEffectAction,
   isICardRenderCorpBoxAction,
   isICardRenderCorpBoxEffect,
+  isICardRenderCorpBoxEffectAction,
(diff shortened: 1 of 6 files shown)


TarunTheo13:
Palladin shipping also doesn't scroll for @katieyang. Maybe we can make the cards with lots of text much larger on hover? I also wondered if you thought I should do a general refactor of corp boxes?

kberg:
I don't know what a general refactor would do.

TarunTheo13:
> I don't know what a general refactor would do.
Tidy things up a bit but happy to leave things as is. Is it worth merging what I've done so far so we have the functionality there when we find a solution for the corps with lots of text?

katieyang:
I think this update makes a lot of sense for corps that have both actions and effect. Before you could only list it as one or the other on the corp box label, so this way of doing things makes more sense in those cases. 

This doesn't solve the scrolling issue (if corp boxes were to be implemented consistently for other corps) but at least Palladin Shipping seems to have little enough text that that's fine. I think this should be merged.

kberg:
Thanks for the help, everyone. In it goes!

kberg (approved):

PR #7500, opened 2025-07-05
Sync fork

Mizkei1:


kberg:
Let me know if you have issues merging, because there's a big change to how
cards behave, particularly corporations.

Mizkei1:
Thanks, just accidentally tried to merge in the main here instead of merging from here to my fork

src/server/cards/community/HiredEarthRaiders.ts
```diff
@@ -0,0 +1,59 @@
...
+  public override bespokePlay(player: IPlayer) {
+    if (player.game.isSoloMode()) {
+      return undefined;
+    }
+
+    const availableActions = new OrOptions();
+
+    player.getOpponents().forEach((target) => {
+      if (target.megaCredits > 0) {
+        const difference = player.tags.countAllTags()[Tag.EARTH] - target.tags.countAllTags()[Tag.EARTH];
```
kberg, line 41: This should be `player.tags.count(Tag.EARTH) - target.tags.count(Tag.EARTH, 'raw')`. First of all, that's a little bit more efficient and cleaner, but also it gives the card player, the advantage of their wild tags or any tag substitutions (Like Earth Embassy).

src/server/cards/community/PirkkaLuna.ts
```diff
@@ -0,0 +1,52 @@
+import {Tag} from '../../../common/cards/Tag';
+import {IPlayer} from '../../IPlayer';
+import {ICorporationCard} from '../corporation/ICorporationCard';
+import {CorporationCard} from '../corporation/CorporationCard';
+import {ICard} from '../ICard';
+import {CardName} from '../../../common/cards/CardName';
+import {CardRenderer} from '../render/CardRenderer';
+
+export class PirkkaLuna extends CorporationCard {
+    private generationMap: Record<number, boolean> = {};
```
kberg, line 10: Think you already know this, but if your server restarts this generation map will disappear. Is that something you want to preserve?

src/server/cards/community/ColonialConquest.ts
```diff
@@ -0,0 +1,48 @@
...
+            b.effect('When you trade, decrease trade step by one.', (eb) => {
+              eb.trade().startEffect.text('-1', Size.LARGE);
+            }).br;
+            b.colonies(2);
+          }),
+          description: 'Place two colonies. The second one can be placed to colonies where you already have a colony',
+        },
+      });
+    }
+    public override bespokeCanPlay(player: IPlayer) {
```
kberg, line 38: bespokeCanPlay that always returns true can be removed entirely.

kberg (commented):
It was fun to read your cards!

src/server/cards/community/HiredEarthRaiders.ts
Mizkei1: Thanks for the info, need to fix this at some point then 👍

src/server/cards/community/PirkkaLuna.ts
Mizkei1: Good point. Tbh, does not matter 99% in my case as this fork is only for very limited use, but might be something that I'll fix just in case at some point

PR #7526, opened 2025-07-11
Communication Center II: Attack of the Solar Storm

TarunTheo13:
Let me know what you think, and if I should change this to a traditional if statement.


TarunTheo13:
They are relatively short tests and Communication Center may need more work in the future as new cards are inevitably added. So probably doesn't hurt to leave them in for visibility

kberg (approved):
This is a good PR.

PR #7527, opened 2025-07-11
Defer adding science resource to Olympus Conference for Hyperspace Drive Prototype

### LINKED ISSUE #7505: Hyperspace Drive Prototype adds science resource to Olympus Conference after being triggered by tag
A rare edge case, but it would be nice to be able to add a science resource to Olympus Conference and then take it off via the science tag on HDP. It could be accomplished by changing OC's priority value to 0 and then deferring that action on HDP to -1. But I know -1 is a legacy value so I wanted to ask about it. Will do the work for this one, just want your approval.

TarunTheo13:
Fixes #7505

I have chosen the Priority values so that if a player also had Mars University, this interaction would take place before discard and draw. I will add a test(/s) and manual test worked as expected

### DIFF
--- a/src/server/cards/underworld/HyperspaceDrivePrototype.ts
+++ b/src/server/cards/underworld/HyperspaceDrivePrototype.ts
@@ -40,7 +40,6 @@ export class HyperspaceDrivePrototype extends Card implements IProjectCard {
       player.game.log('${0} has no fighter resource cards and gained 1 titanium.', (b) => b.player(player));
       player.stock.titanium += 1;
     }
-    // If the player has Olympus Conference, its effect should take place after this gets a resource on it.
     const scienceCards = player.getResourceCards(CardResource.SCIENCE);
     if (scienceCards.length > 0) {
       player.game.defer(new AddResourcesToCard(player, CardResource.SCIENCE), Priority.HYPERSPACE_DRIVE_PROTOTYPE);
--- a/src/server/cards/underworld/HyperspaceDrivePrototype.ts
+++ b/src/server/cards/underworld/HyperspaceDrivePrototype.ts
@@ -7,6 +7,7 @@ import {Tag} from '../../../common/cards/Tag';
 import {IPlayer} from '../../IPlayer';
 import {CardResource} from '../../../common/CardResource';
 import {AddResourcesToCard} from '../../deferredActions/AddResourcesToCard';
+import {Priority} from '../../deferredActions/Priority';
 
 export class HyperspaceDrivePrototype extends Card implements IProjectCard {
   constructor() {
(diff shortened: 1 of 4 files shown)


kberg (commented):
What if instead of all this, `OlympusConference`'s priority changed to `GAIN_RESOURCE_OR_PRODUCTION`? I'm loathe to use `SUPERPOWER` if it can be avoided.

This needs a test so someone doesn't undo it.

TarunTheo13:
> What if instead of all this, `OlympusConference`'s priority changed to `GAIN_RESOURCE_OR_PRODUCTION`? I'm loathe to use `SUPERPOWER` if it can be avoided.

`OlympusConference` effect needs to happen before `Mars University` at 12. Do you have an alternative suggestion?

kberg:
As far as I can tell, these are the interesting science resource / card draw behaviors. 
image

The four on the right are not interesting. They operate independently since they're actions or sufficiently isolated so as to not impact this pipeline.

I don't know why Olympus Conference is quite so low, so let's do this instead:

image

Change Priority like so. Don't use SUPERPOWER, and in fact, this gets rid of it.

If it turns out something else is part of the problem, then we can discuss it.

kberg:
e.g. 

```
  /** Any effect from one of your opponent's card that triggers during your turn. */
  OPPONENT_TRIGGER,

  /**
   * Resolve Hyperspace Drive Prototype before Olympus Conference
   *
   * https://docs.google.com/drawings/d/1VXfVmoJWU_QmMDwZ-liVh5ZvWVmmpl2kuo_3ANz4omY/edit?usp=sharing
   */
  HYPERSPACE_DRIVE_PROTOTYPE,
  /**
   * Resolve Olympus Conference before Sponsored Academies and Mars U.
   */
  OLYMPUS_CONFERENCE,
  /**
   * Sponsored Academies should come before Mars U.
   */
  SPONSORED_ACADEMIES,
  DRAW_CARDS,
  BUILD_COLONY,
```

src/server/cards/underworld/HyperspaceDrivePrototype.ts
```diff
@@ -39,10 +40,10 @@ export class HyperspaceDrivePrototype extends Card implements IProjectCard {
       player.game.log('${0} has no fighter resource cards and gained 1 titanium.', (b) => b.player(player));
       player.stock.titanium += 1;
     }
-    // TODO(kberg): If the player has olympus conference, its effect should take place after this gets a resource on it.
+    // If the player has Olympus Conference, its effect should take place after this gets a resource on it.
```
kberg: Remove this comment.

kberg (approved):
Thank you!

PR #7539, opened 2025-07-14
L1 Trade Terminal will not automatically add resource to Ancient Shipyards

TarunTheo13:
Not sure this is the correct way to write it but this fixes a particular edge case that no-one has complained about, yet.

### DIFF
--- a/src/server/cards/prelude2/L1TradeTerminal.ts
+++ b/src/server/cards/prelude2/L1TradeTerminal.ts
@@ -7,6 +7,7 @@ import {IPlayer} from '../../IPlayer';
 import {PlayerInput} from '../../PlayerInput';
 import {SelectCard} from '../../inputs/SelectCard';
 import {ICard} from '../ICard';
+import {AncientShipyards} from '../moon/AncientShipyards';
 
 export class L1TradeTerminal extends Card {
   constructor() {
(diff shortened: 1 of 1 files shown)


kberg:
What's the actual problem? Let's find a solution that doesn't hard code like that.

TarunTheo13:
Yes but we still wouldn't want it to be added automatically

kberg:
Looking at the cards, it would be good to also apply this to Vermin.
image

kberg:
What's your next step?

TarunTheo13:
Have written another commit, have done manual tests - it does not automatically add to Vermin & it _does_ automatically add when it's meant to. Let me know if you want me to write a test

src/server/cards/prelude2/L1TradeTerminal.ts
```diff
@@ -55,7 +55,8 @@ export class L1TradeTerminal extends Card {
 
   public override bespokePlay(player: IPlayer): PlayerInput | undefined {
     const cards = this.getEligibleCards(player);
-    if (cards.length <= 3) {
+    const includesAncientShipyardsOrVermin = (cards.some(card => card.name === CardName.ANCIENT_SHIPYARDS || card.name === CardName.VERMIN));
```
kberg: This makes sense.

Here's a way we can make it more permanent if other cards show up with negative VP:

```
    const cardsWithNegativeVP = cards.filter((card) => {
      const vp = card.victoryPoints;
      if (typeof vp === 'object' && vp.resourcesHere !== undefined && (vp.per ?? 0) < 0) {
        return false;
      }
      return true;
    });
```

 That doesn't work if future cards have type victoryPoints type `special` but wdyt

TarunTheo13: I think with future cards there will be future unexpected interactions but this is still a better solution

TarunTheo13: It doesn't work now because vermin does have victoryPoints type `special`

kberg (approved):
I added a small comment.

kberg:
Thanks for the continued perserverence.

PR #7561, opened 2025-07-22
Support i18n of select card to play warnings

alrusdi:


src/client/components/SelectProjectCardToPlay.vue
```diff
@@ -10,11 +10,11 @@
 
   <section v-trim-whitespace>
     <template v-if="card.additionalProjectCosts">
-      <div v-if="card.additionalProjectCosts.thinkTankResources" class="card-warning">
-        Playing {{card.name}} consumes {{card.additionalProjectCosts.thinkTankResources}} data from Think Tank
+      <div v-if="card.additionalProjectCosts.thinkTankResources" class="card-warning" v-i18n="[$t(card.name), card.additionalProjectCosts.thinkTankResources, $t('Think Tank')]">
```
kberg, line 13: Wow, this is good. This also renders in English, I assume?

kberg (approved):
This is a good PR. Thanks for the work lining up the translations.

src/client/components/SelectProjectCardToPlay.vue
alrusdi: Yes it does not affect English text

PR #7568, opened 2025-07-23
Update CardList.vue

Borbarad2:
make new cardlist button  v-i18n friendly


kberg (approved):
I removed the card rendering changes.

PR #7589, opened 2025-08-04
Maintain player discount from Indentured Workers (and similar) after it is used through Playwrights action

### LINKED ISSUE #7080: When Player A takes the Playwrights action after Player B plays Indentured Workers, the card discount appears for Player B
In this game:
Player A is the Playwrights corp. 
Player B plays indentured workers, and ends turn. 
Player A uses the Playwrights action on Player B's Indentured workers
Player B's discount disappears

Affected Game: https://terraforming-mars.herokuapp.com/the-end?id=pa958571188c7

Notes about how to (possibly) fix:

In the same way that there is a special check for special design and lawsuit, a proxy card might be needed for indentured workers (and maybe conscription as well?)

https://github.com/terraforming-mars/terraforming-mars/blob/5eba889a903c6d0360acaa101a832ab13c9ee3e2/src/server/cards/community/Playwrights.ts#L75C1-L78C17

TarunTheo13:
Fixes #7080

This is done, in part, by saving the original player of the card to the discarded card. I cannot see any downsides of doing it this way but it is using Javascript functionality I am not familiar with so please evaluate if there are any unforeseen consequences

### DIFF
--- a/src/server/Player.ts
+++ b/src/server/Player.ts
@@ -707,6 +707,20 @@ export class Player implements IPlayer {
       }
     });
 
+    // Check for discounts from other players' removedFromPlayCards
+    // This allows the original player to still use discounts from cards that Playwrights replayed
+    for (const otherPlayer of this.game.players) {
+      if (otherPlayer === this) continue;
+      otherPlayer.removedFromPlayCards.forEach((removedFromPlayCard) => {
+        if (removedFromPlayCard.getCardDiscount !== undefined) {
+          const originalOwner = (removedFromPlayCard as any).originalOwner;
+          if (originalOwner === this.id) {
+            cost -= removedFromPlayCard.getCardDiscount(this, card);
+          }
+        }
+      });
+    }
+
     // TODO(kberg): put this in a callback.
     if (card.tags.includes(Tag.SPACE) && PartyHooks.shouldApplyPolicy(this, PartyName.UNITY, 'up04')) {
       cost -= 2;
--- a/src/server/cards/community/Playwrights.ts
+++ b/src/server/cards/community/Playwrights.ts
@@ -60,6 +60,9 @@ export class Playwrights extends CorporationCard implements ICorporationCard {
         ([card]) => {
           const selectedCard: IProjectCard = card;
 
+          // Save the original owner before removing the card
+          const originalOwner = players.find((originalPlayer) => originalPlayer.playedCards.get(selectedCard.name))?.id;
+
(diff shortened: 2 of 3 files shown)


src/server/Player.ts
```diff
@@ -707,6 +707,20 @@ export class Player implements IPlayer {
       }
     });
 
+    // Check for discounts from other players' removedFromPlayCards
+    // This allows the original player to still use discounts from cards that Playwrights replayed
+    for (const otherPlayer of this.game.players) {
```
kberg, line 712: Use `this.opponents.`

src/server/Player.ts
```diff
@@ -707,6 +707,20 @@ export class Player implements IPlayer {
...
+      if (otherPlayer === this) continue;
+      otherPlayer.removedFromPlayCards.forEach((removedFromPlayCard) => {
+        if (removedFromPlayCard.getCardDiscount !== undefined) {
+          const originalOwner = (removedFromPlayCard as any).originalOwner;
```
kberg, line 716: Gonna have to find a way to do that other than 'as any'

src/server/cards/community/Playwrights.ts
```diff
@@ -71,6 +74,14 @@ export class Playwrights extends CorporationCard implements ICorporationCard {
             .andThen(() => {
               player.playCard(selectedCard, undefined, 'nothing'); // Play the card but don't add it to played cards
               player.removedFromPlayCards.push(selectedCard); // Remove card from the game
+              // Store ownership information in the card itself for now
+              if (originalOwner) {
+                Object.defineProperty(selectedCard, 'originalOwner', {
```
kberg, line 79: If you need a property, then define it. If it's got to survive serialization, then there are ways to deal with it.

kberg (commented):
I still want to know why this is the right behavior. If the event is out of play, should it count?

TarunTheo13:
https://boardgamegeek.com/thread/2907313/fan-expansion-tm-underworld-2/page/6

I wonder if your conversation R.E. Astra Mechanica with the new Underworld expansion sets a precedent? Of course, it's a completely different expansion but it would be more consistent if we implemented these changes.

kberg:
Maybe a player property "discarded" which tracks cards the player has discarded. It sits outside the tableau. I think that complicates code in alot of places, /shrug

kberg:
There's a simpler solution. Going to implement it.

PR #7608, opened 2025-08-20
Several minor changes

kberg:


kberg:
Fixes https://github.com/terraforming-mars/terraforming-mars/issues/7607

PR #7675, opened 2025-10-05
dcep93/custom middleware

dcep93:
Per https://github.com/terraforming-mars/terraforming-mars/issues/7673

This is my first contribution to this repo. I just wanted to get something written down so we could have stuff to point to.

This PR adds an easy capability for devs to use the live site with custom middleware. With this configuration, the middleware will allow the user to intercept all fetch requests and responses. Also fires when there is a DOM update.

I'm linking a video of myself using this.

The major change is conversion of XMLHttpRequest to fetch, hopefully this is not a problem.


src/client/components/App.ts
```diff
@@ -248,6 +242,9 @@ export const mainAppSettings = {
       app.screen = 'start-screen';
     }
   },
+  updated() {
+    runCustomMiddleware({app:this})
```
dcep93: Run custom middleware providing the vue context after every update.

src/client/components/WaitingFor.vue
```diff
@@ -98,38 +98,68 @@ export default Vue.extend({
...
+              responseType: "json",
+              response,
+            }))
+            .catch((err) => ({
+              status: resp.status,
+              err,
+            }))
+        )
+        .then((xhr) => {
+          this.loadPlayerViewResponse(xhr);
```
dcep93: This function is used in two places with similar context. I kept the xhr signature, but maybe we want to fully move away from that framework, and make a new helper.

src/client/main.ts
```diff
@@ -13,6 +14,7 @@ declare global {
 }
 
 async function bootstrap() {
+  await customMiddleware()
```
dcep93: load custom middleware during bootstrap

src/client/plugins/customMiddleware.ts
```diff
@@ -0,0 +1,53 @@
+// uses url query param customMiddlewaresSrc
+// to download custom middleware
+// and run them before fetch, after fetch,
+// and on completion of rendering
+
+// this powerful tool enables arbitrary frontend
+// workloads without needing to merge to
+// the main repo, as long as the src is visible
+
+const originalFetch = window.fetch;
```
dcep93, line 10: intercept window.fetch

src/client/plugins/customMiddleware.ts
```diff
@@ -0,0 +1,53 @@
...
+
+declare global {
+  interface Window {
+    customMiddlewareF: (ctx: {
```
dcep93, line 14: clients defined by customMiddlewaresSrc should assign to this global field

dcep93:
The attached video shows how it might be used by a dev like myself. The "disregard" tooling is clunky, but that's not the point, and that's not what I'm trying to merge. The focus of this PR is that users can provide a src url to perform arbitrary work and manipulate the DOM (or manipulate a fetched response). Super cool!

dcep93:
https://github.com/user-attachments/assets/f285f332-dd02-43fe-8a7c-625a9530cafe

kberg:
I guess, is this really necessary? I do local development with `npm run client:dev` and `npm run server:dev` (and `npm run css:watch`). What does this feature do that those don't?

kberg:
> This feature allows for arbitrary features to be included for everyone who includes the middleware as a url parameter. This will work for users who don't know how to perform `npm run` and also for mobile users. Specifically, my friends have asked for the "[time warp](https://github.com/terraforming-mars/terraforming-mars/pull/7695)" feature, and this custom middleware would allow that usage, even before time warp is landed.
> 
> No worries if we decide it's not worth it!

I appreciate that you have that perspective. I understand your use case a bit now, but not fully. I'm inclined to not move forward, but ... not yet. I'll keep this open and get back to the others.

PR #7685, opened 2025-10-11
Dcep93/hand jive

dcep93:
https://github.com/user-attachments/assets/a080472a-1185-4d57-bedc-819b1676c2d6

Existing sortable card behavior is tricky, and doesn't work on mobile. We may want to tweak the specific UI rendering, but this feature will help immensely!

### DIFF
--- a/src/client/components/SortableCards.vue
+++ b/src/client/components/SortableCards.vue
@@ -1,6 +1,6 @@
 <template>
 <div>
-  <div>
+  <div v-if="experimentalUI()">
     <label>
       <input type="checkbox" v-model="showReorder" />Reorder cards
     </label>
@@ -24,6 +24,7 @@ import Card from '@/client/components/card/Card.vue';
 import {CardName} from '@/common/cards/CardName';
 import {CardModel} from '@/common/models/CardModel';
 import {CardOrderStorage} from '@/client/utils/CardOrderStorage';
+import {getPreferences} from '@/client/utils/PreferencesManager';
 
 type DataModel = {
   /** When true use the point-and-click reorder UI */
@@ -136,6 +137,9 @@ export default Vue.extend({
           }
         }
       }
+    },
+    experimentalUI(): boolean {
+      return getPreferences().experimental_ui;
     },
   },
 });


src/client/components/SortableCards.vue
```diff
@@ -1,13 +1,25 @@
 <template>
+<div>
+  <div>
+    <label>
+      <input type="checkbox" v-model="handJiveIsChecked" /> hand jive
```
kberg: Let's give this a more obvious name: "Reorder cards"?

kberg: Also, add a `v-i18n` attribute to this HTML element. That makes the string translateable.

src/client/components/SortableCards.vue
```diff
@@ -1,13 +1,25 @@
...
+    </label>
+  </div>
   <div class="sortable-cards">
     <div ref="draggers" :class="{ 'dragging': Boolean(dragCard) }" v-for="card in getSortedCards()" :key="card.name" draggable="true" v-on:dragend="onDragEnd()" v-on:dragstart="onDragStart(card.name)">
       <div v-if="dragCard" ref="droppers" class="drop-target" v-on:dragover="onDragOver(card.name)"></div>
-      <div ref="cardbox" class="cardbox">
-        <Card :card="card"/>
+      <div ref="cardbox" class="cardbox" @click="clickMethod" style="position:relative">
```
kberg: Replace the styles with CSS. Oh, I wish we could use Vue 3's local styles.

src/client/components/SortableCards.vue
kberg: By the way, I love the name "hand jive" because I've watched Grease like a hundred times. But it's probably a little too colloquial.

src/client/components/SortableCards.vue
```diff
@@ -47,6 +59,10 @@ export default Vue.extend({
       }
     }
     return {
+      handJiveIsChecked: false,
+      handJiveCardStyle: Object.freeze({
```
kberg: Why freeze?

src/client/components/SortableCards.vue
kberg: Why "position:relative"?

src/client/components/SortableCards.vue
dcep93: ha! fair enough, happy to rename this - "reorder cards" sounds plenty descriptive. I can look at adding the v-i18n, and I'll have chatgpt translate it perhaps?

src/client/components/SortableCards.vue
dcep93: Cool, I can look at using the proper css file.

src/client/components/SortableCards.vue
dcep93: good question, I don't remember actually... in any case, it seems that we'd prefer to have a class selector, instead of a style selector, and move this "pointerEvents: none" to a css file - I can get on that

src/client/components/SortableCards.vue
dcep93: I believe this is necessary for the pattern with the red bars - since that uses `position:absolute`

I can add this style to a css file, but is a problem to use `position:relative`?

kberg: No problem, I'm just asking.

src/client/components/SortableCards.vue
kberg: No worries about adding the translations. That gets done bit by but on its own, most of the time.

kberg:
@dcep93 I have made some changes.

I recommend at this point that you change the color of the banners to something much more subtle. I suggest perhaps something like

```
| |                  | |
|«|     card         |»|
| |                  | |
|«|                  |»|
| |                  | |
```

Otherwise it's kinda loud and in the way of playing.

Also, don't show the respective banner if the card is at either end.

kberg:
Also, please ask for feedback from the community before it gets checked in.

PR #7686, opened 2025-10-11
Dcep93/remote server

dcep93:
It's now easier to develop locally and showcase feature changes by using local assets and remote server.

https://github.com/user-attachments/assets/96287ee1-f58e-4b1b-8f99-ce4ffae60a3c


dcep93:
Hmm in retrospect, maybe this isn't so useful - I'll close it.

It was nice to be able to use the remote server instead of the local server, so that I could showcase a feature with identical game state. Though, I don't think that pattern will be valuable very often.

PR #7687, opened 2025-10-11
dcep93/anchorRadioAction

dcep93:
Personally, it's annoying for me when I click a radio button and then I have to scroll again to select the submit button. This makes the radio button sticky!

https://github.com/user-attachments/assets/eb726e7c-9564-40ce-ab63-a68172410bcc

### DIFF
--- a/src/client/components/OrOptions.vue
+++ b/src/client/components/OrOptions.vue
@@ -30,6 +30,7 @@
 
 import Vue from 'vue';
 import AppButton from '@/client/components/common/AppButton.vue';
+import {isHTMLElement} from '@/client/utils/vueUtils';
 import {PlayerViewModel, PublicPlayerModel} from '@/common/models/PlayerModel';
 import {OrOptionsModel, PlayerInputModel} from '@/common/models/PlayerInputModel';
 import {getPreferences} from '@/client/utils/PreferencesManager';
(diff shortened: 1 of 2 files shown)


src/client/components/OrOptions.vue
```diff
@@ -87,7 +87,50 @@ export default Vue.extend({
       selectedOption,
     };
   },
+  watch: {
+    selectedOption(newOption: PlayerInputModel | undefined) {
+      if (typeof window === 'undefined' || newOption === undefined) {
```
kberg: I don't think we ever bother checking if window is undefined.  Help me understand why this is a good idea?

src/client/components/OrOptions.vue
```diff
@@ -87,7 +87,50 @@ export default Vue.extend({
       selectedOption,
     };
   },
+  watch: {
```
kberg, line 91: What do you think of this adaptation?
```
  watch: {
    selectedOption(newOption: PlayerInputModel) {
      // Clicking the option can shift elements on the page.
      // This preserves the location of the option button the user just clicked by
      // tracking where it was on the screen, where it moved, and then repositioning it.
      const anchorTop = this.getTop(newOption);
      this.$nextTick(() => {
        const newTop = this.getTop(newOption);
        if (anchorTop !== undefined && newTop !== undefined) {
          const delta = newTop - anchorTop;
          if (Math.abs(delta) > 0.5) {
            window.scrollBy(0, delta);
          }
        }
      });
    },
  },
  methods: {
    getTop(option: PlayerInputModel | undefined): number | undefined {
      const element = this.getOptionLabelElement(option);
      return element?.getBoundingClientRect().top;
    },
...
```

kberg: (I tossed in a comment.)

kberg: Also, we almost never use `null`, almost exclusively we use `undefined` unless an API requires it.

dcep93:
Thanks so much! I’m really excited for this project, feel free to make any changes to any of the PRs. 

I’m still planning to hop into discord, but it’s been tough to find time these days.

kberg (approved):
I tried writing a test but it doesn't look hopeful.

PR #7688, opened 2025-10-11
dcep93/easyActionScroll

dcep93:
I don't want to have to scroll to the action button when I select a card to play. Autoscroll!
Let's discuss before merging - behavior isn't so good for multi-select.

https://github.com/user-attachments/assets/44569107-e06d-432f-8df8-e065ad72b42f


kberg:
This is tricky, because there are many times when selecting a card is not the end of the operation. For instance, selecting the starting hand, or selling patents.

I applaud the idea, but perhaps a better solution is to always show the action button. _But only perhaps._

PR #7689, opened 2025-10-11
Dcep93/fetch

dcep93:
This should be a noop that converts xhr to fetch. Sort of risky, if we get something wrong, behavior will break for an entire endpoint!

This unlocks https://github.com/terraforming-mars/terraforming-mars/pull/7675

Discussion here should be for the safety of this specific change.
Discussion of the desire for the parent PR should be kept in #7675

### DIFF
--- a/src/client/components/App.ts
+++ b/src/client/components/App.ts
@@ -153,7 +153,7 @@ export const mainAppSettings = {
             app.screen = 'the-end';
             if (currentPathname !== paths.THE_END) {
               window.history.replaceState(
-                xhr.response,
+                model,
                 `${constants.APP_NAME} - Player`,
                 `${paths.THE_END}?id=${model.id}`,
               );
@@ -166,7 +166,7 @@ export const mainAppSettings = {
             }
             if (currentPathname !== path) {
               window.history.replaceState(
-                xhr.response,
+                model,
                 `${constants.APP_NAME} - Game`,
                 `${path}?id=${model.id}`,
               );


src/client/components/App.ts
```diff
@@ -209,25 +204,28 @@ export const mainAppSettings = {
...
       app.screen = 'game-home';
-      const xhr = new XMLHttpRequest();
-      xhr.open('GET', paths.API_GAME + window.location.search);
-      xhr.onerror = function() {
-        alert('Error getting game data');
-      };
-      xhr.onload = function() {
-        if (xhr.status === statusCode.ok) {
+
+      const url = paths.API_GAME + window.location.search;
```
kberg, line 208: This one looks good. App.ts has a few of these, so I'm looking at them one at a time.

src/client/components/LoadGameForm.vue
```diff
@@ -51,14 +49,21 @@ export default Vue.extend({
...
-      xhr.onerror = function() {
-        alert('Error loading game');
-      };
-      xhr.onload = () => {
-        if (xhr.status === statusCode.ok) {
-          const response = xhr.response as SimpleGameModel;
+      fetch(paths.LOAD_GAME, {
+        method: 'PUT',
+        headers: {'Content-Type': 'application/json'},
+        body: JSON.stringify({
```
kberg: This removed the type. The purpose of the type definition was to ensure that in case the type for the LoadGameForm API changes, this will have a compiler error.

src/client/components/logpanel/LogPanel.vue
```diff
@@ -110,20 +109,26 @@ export default Vue.extend({
...
+      fetch(url, {signal: controller.signal})
+        .then((resp) => {
+          if (!resp.ok) {
+            console.error(`error updating messages, response code ${resp.status}`);
+            return null;
+          }
+          return resp.json();
+        })
+        .then((data) => {
+          if (!data) return;
```
kberg, line 130: This file's change does not look complete. Line 133 is still `messages.push(...xhr.response);`. Did you run this to see if it works?

src/client/components/LoadGameForm.vue
dcep93: good point! fixed here
https://github.com/terraforming-mars/terraforming-mars/pull/7689/commits/f58efccf5b2585d76c91440514054db4e2c756fb

src/client/components/logpanel/LogPanel.vue
dcep93: oh that's embarrassing, fixed now here, along with everything else the typechecker didn't like

https://github.com/terraforming-mars/terraforming-mars/pull/7689/commits/af4e89ec68c7122051afd43eafb62b6101f74fc4

kberg:
I have merged this with `main`, which removes a couple of the updates I made manually. Before I move forward, how many of these did you manually test?

How do you know they work?

dcep93:
tl;dr
1. thanks!
2. I tested manually on a previous parallel-ish version, but not on this PR
3. want me to split up the PRs? Or do you have a good workflow there?
4. want me to add tests?

Thanks for that! I really appreciate you doing the heavy lifting.

It was a while ago at this point, but I remember testing the wider [customMiddleware](https://github.com/terraforming-mars/terraforming-mars/pull/7675/files#diff-089d55e47d393e1997a143e3345fd6fba8e3472ac43fbd08d6c89accf51c77dc) PR. I see that had the corrected `messages.push(...xhr.response);` removal. Though as I recall, I made all those customMiddleware changes, tested a 1-player and 2-player game, and since things appeared to work, I packaged it up to present to the community. So it's possible that things "silently" failed without me knowing.

I like your plan/process for splitting these out into smaller changes, and it looks like we have fewer than 10 left. Do you think it'd be worthwhile for me to pick it up and make the remaining smaller PRs? Happy to let you keep on that track if you already have an efficient workflow. Also, you might have an easier time than I would doing the testing.

Speaking of testing, I'm completely unfamiliar with our unit tests. Should they be expanded? Until my [fixes](https://github.com/terraforming-mars/terraforming-mars/pull/7689/commits/af4e89ec68c7122051afd43eafb62b6101f74fc4) commit was applied, indeed the typechecker complained, but it might be nice to actually run end to end tests with seeded data, which would have exposed the bogus ...xhr.response bug. I also see that `npx cross-env NODE_ENV=production webpack` is failing due to a bad change of mine, I think let's look at that on the specific change PR.

src/client/components/App.ts
```diff
@@ -211,25 +206,28 @@ export const mainAppSettings = {
...
     } else if (currentPathname === paths.GAME) {
       app.screen = 'game-home';
-      const xhr = new XMLHttpRequest();
-      xhr.open('GET', paths.API_GAME + window.location.search);
-      xhr.onerror = function() {
-        alert('Error getting game data');
-      };
-      xhr.onload = function() {
-        if (xhr.status === statusCode.ok) {
+
```
kberg: NB For some reason this change introduced some console errors related to timing. I had to move the `app.screen = 'game-home'` down to the second `.then` to eliminate the errors.

kberg: No response required, just pointing it out.

kberg:
There is only one WaitingFor left.

PR #7690, opened 2025-10-11
Dcep93/hand preview

dcep93:
The UI is rough, and should be iterated on before merging. This opens the conversation and shows the kind of thing I'm imagining. You can already click cards in the log to preview them. What if cards were always previewed?

https://github.com/user-attachments/assets/bae2895a-7b42-404d-a73f-b66ac66359c5


kberg:
I gave this a look over. It seems ... interesting? As before, see what other people say. Perhaps there's something useful but I can't say for sure.

PR #7691, opened 2025-10-11
dcep93/mobileCardSelect

dcep93:
I can do a screen recording if that's valuable. Currently, quickly tapping cards on mobile is pretty finicky. Let's make it more consistent!


kberg:
This looks pretty safe. Can you tell me: have you verified this doesn't impact desktop?

kberg (approved):
Sorry for the long delay. Thanks for this.

kberg:
I added a test.

ssimeonoff:
@kberg @dcep93 
this broke mobile UX in a bad way...
Was doing "missclicks" for days until it finally hit me today:
- use mobile
- touch over a selectable card
- swipe up or down (to trigger scroll event)
- lift finger
Observed: the card is selected
Expected: click event not to be registered and the card not to be selected.

kberg:
Rolling back.

kberg:
@dcep93 if you can fix this we can get it in.

dcep93:
Argh, sorry for the trouble, glad you found the cause of the issue! I think we should just leave this reverted. Let me know if you want me to try for another attempt, but it feels too risky with little/undefined upside.

kberg:
Would you like to try to fix it?

dcep93:
Perhaps, it's on my radar, but not sure I'll find the time :/

PR #7692, opened 2025-10-11
dcep93/tagCountTrace

dcep93:
Hmm I can't seem to reproduce this consistently. I'd seen a console error when we didn't pass the right type to TagCount.vue. Feel free to ignore this until I get a recording perhaps?


kberg:
Hi @dcep93. What was the error? (It's probably documented elsewhere, lemme know. Perhaps I can find it.)

PR #7693, opened 2025-10-11
Dcep93/waiting for jump

dcep93:
If I'm looking through my hand, and it becomes my turn, my screen jumps to a random place. Let's fix that!

### DIFF
--- a/src/client/components/WaitingFor.vue
+++ b/src/client/components/WaitingFor.vue
@@ -52,6 +52,10 @@ type DataModel = {
   playersWaitingFor: Array<Color>
   suspend: boolean,
   savedPlayerView: PlayerViewModel | undefined;
+  /**
+   * The size of this element before it's mounted. Compared after it is mounted,
+   * without which the page jars suddenly when the WaitingFor appears
+   */
   beforeMountSize: number | null;
 }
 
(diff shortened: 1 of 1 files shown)


src/client/components/WaitingFor.vue
```diff
@@ -240,12 +242,41 @@ export default Vue.extend({
     showRefresh(): boolean {
       return this.suspend === true && this.savedPlayerView !== undefined;
     },
+    getRef(): HTMLElement | null {
+      return document.querySelector('[data-ref="waiting-for-ref"]') as HTMLElement | null;
+    },
+    isAboveViewportBottom(element: HTMLElement): boolean {
+      const rect = element.getBoundingClientRect();
+      const windowHeight = window.innerHeight || document.documentElement.clientHeight;
```
kberg: windowHeight is unused.

src/client/components/WaitingFor.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div>
+  <div data-ref="waiting-for-ref">
```
kberg: Instead of a data-ref, how about just giving this an `id` and search for that? In this case, I wonder if there is some best practice you're following that I don't know about.

kberg:
ping

dcep93:
Thanks for the ping, not sure when I'll get to this, but I'll have it nice and clean!

PR #7697, opened 2025-10-13
Update promo cards and more

Borbarad2:
Adapt to new official German translation from the publisher themselves


kberg (approved):

Borbarad2:
Thanks for approving.

PR #7699, opened 2025-10-16
Fix problem of clipped science tag in log message

ilevit:
The science tag is the only tag that currently sets a custom background position.

To reproduce, have a card like Search For Life and draw a card with science tag (as the top card) after activating it.


kberg:
Oh, good catch!

kberg:
Thank you!

Any chance you have a screenshot of the before and after?

PR #7700, opened 2025-10-16
Fix own victory points not showing

### LINKED ISSUE #7265: Own VP not visible in 3 players game
I don't see my VP while playing with 2 other players (it works in a solo game now -thank you!)

Image

{
  "url": "https://terraforming-mars.herokuapp.com/spectator?id=s1dfce1494a15",
  "color": "green",
  "step": 82,
  "version": "d7b9204",
  "builtAt": "Tue, 11 Mar 2025 14:28:40 GMT",
  "browser": "Firefox 136",
  "language": "en",
  "experimental_ui": false
}

nyddogghr:
Hi,
This is a suggestion to fix #7265 . I've tested it locally and it does allow me to see my own victory points but not other's at all if "show other players VP" is not checked.
Let me know if you want to make changes before merging or just do it another way.

### DIFF
--- a/src/server/models/ServerModel.ts
+++ b/src/server/models/ServerModel.ts
@@ -87,7 +87,7 @@ export class Server {
   public static getPlayerModel(player: IPlayer): PlayerViewModel {
     const game = player.game;
 
-    const players: Array<PublicPlayerModel> = game.playersInGenerationOrder.map((p) => this.getPlayer(p, player.color));
+    const players: Array<PublicPlayerModel> = game.playersInGenerationOrder.map((p) => this.getPlayer(p, p.color === player.color));
 
     const thisPlayerIndex = players.findIndex((p) => p.color === player.color);
     const thisPlayer: PublicPlayerModel = players[thisPlayerIndex];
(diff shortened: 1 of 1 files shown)


kberg:
Thanks for the code. I haven't forgotten about this, and I'll come look at it later. I haven't stared at it too closely, but it feels a little hacky, and I'm not sure why. Like I suggested, this just requires a little focus.

src/server/models/ServerModel.ts
```diff
@@ -91,6 +91,13 @@ export class Server {
 
     const thisPlayerIndex = players.findIndex((p) => p.color === player.color);
     const thisPlayer: PublicPlayerModel = players[thisPlayerIndex];
+    if (thisPlayer.victoryPointsBreakdown.total === 0) {
```
kberg: I'm actually not confident about the following assertion. But here's what I _think_:

OK, so yeah this restores the original problem. There are two player views: A public view (how many plants you have) and private views (what cards are in your hand.) `thisPlayer` is the public view, which means anybody can see it.

Now, the UI takes care of some of this by knowing whether the game shows other players the VP, but if you can view your browser's network responses, this will still get delivered regardless of the game options.

Perhaps your feature works because it only applies to a single player?  Ah, I think I see why this looks wonky: it changes `thisPlayer` which could in theory still be passed down to other players.

A unit test would go a long way to verify that it works as you expect.

kberg:
It'll probably be a little while before I look but thanks again.

kberg:
OK I see now. This is a good change. I think there's a small way to clean this up which I will supply and then merge it. Nice work.

tests/models/ServerModel.spec.ts
kberg, line 1: Tests. AMAZING. Thank you!

kberg (approved):

PR #7706, opened 2025-10-23
Add ability to disable animating titlebar

### LINKED ISSUE #7703: Add ability to disable title animations
This is both a feature request and an "if I sent in a patch for this, would it be mergable?"

Currently, I have to disable my screen reader when using the game, because the title animations are constantly announced on my turns, because, as far as I can tell, `document.title` updates are always announced.

It'd be nice to not have to do that, but that also seems like a niche toggle.

krzysz00:
This commit adds the ability to disable the title animation to prevent it from interfering with screen readers.

Fixes #7703

### DIFF
--- a/src/client/components/PreferencesDialog.vue
+++ b/src/client/components/PreferencesDialog.vue
@@ -74,6 +74,15 @@
 
       <div class="preferences_panel_item">
         <label class="form-switch">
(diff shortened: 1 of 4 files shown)


krzysz00:
(Yes, I realize this is an extremely specific feature, and I'm open to other approaches here, but I figured I might as well get a patch in to start the discussion somewhere concrete.)

PR #7717, opened 2025-11-03
Milestones and Awards updates for Terra Cimmeria Novus

ilevit:
Based on the official expansion for Terra Cimmeria.

All awards were defined, just needed to be specified for this board.

For milestones, a few needed to be added.  For cases with small variations, defined alternatives within the same module as alternate exported symbols.  Introduced the notion of a TagBasedMilestone which many other milestones also fall under.  This helps eliminate defining new classes with no real behavioral/logic differences.  Using Function.prototype.bind(), it's pretty easy to define ctor-like invocations, effectively turning a parameterized constructor into a parameterless one to define new milestones.


kberg:
Thanks for the PR. Looks like there's a failing test. I might peek at it now, I might have to wait a few days.

src/server/milestones/TagBasedMilestone.ts
```diff
@@ -0,0 +1,17 @@
+import {Tag} from '@/common/cards/Tag';
```
kberg, line 1: The @ is actually not going to work at runtime. The `@` only works on the client side, and when some tests run. it's a bit irritating, but there you go. Tell me, have you tried running this? No trouble?

Let me know if you want some help with that.

src/server/milestones/modular/Forester.ts
```diff
@@ -1,15 +1,19 @@
...
+      `Have ${threshold} plant production`,
+      threshold);
   }
 
   public getScore(player: IPlayer): number {
     return player.production.plants;
   }
 }
+
+// Forester variant from Terra Cimmeria Novus.
```
kberg, line 15: I'd prefer keeping the same style throughout, which is that this would be its own class. The amount of code does not repeat so much, and nothing else uses `.bind`.

src/server/milestones/TagBasedMilestone.ts
kberg: (odd, the app runs.) /shrug

kberg (commented):
The test failure is probably also confusing. I can look at taking on that part for you. It's a bit messy.

Borbarad2:
Wow, great PR, very nice, long expected. 
Could you please don't forget to do the same for the Utopia Planitia Map as well, please.
There is no need to define/code new milestones or awards for this map, just the set of milestones/awards is needed to compile and set/fix it to Utopia Planitia Map. This would be great, too, and very welcomed. 
Thanks, if you could do that as well. 

@ilevit: Do you have a discord contact as well?

src/server/milestones/TagBasedMilestone.ts
ilevit: Yep, it runs so I figured it's OK.  Would be nice to move everything to use @ to make moving files around more painless.

kberg: Believe me I agree. I think this approach was chosen because it's compile time. I still don't understand why it worked for you.

kberg: Believe me I agree. I think this approach was chosen because it's compile time. I still don't understand why it worked for you.

src/server/milestones/modular/Forester.ts
ilevit: To me, it felt like the codebase in general could use some cleanup with a smaller number of actual classes and then some factory functions that just bind a few values to them since there's so much structural repetition (e.g,. many milestones are have X tags of type Y, etc.)

That said, I can change it to just follow the existing pattern for now

src/server/milestones/TagBasedMilestone.ts
ilevit: Resolving this as long as it's not causing issues

src/server/milestones/modular/Forester.ts
ilevit: Updated

ilevit:
> The test failure is probably also confusing. I can look at taking on that part for you. It's a bit messy.

I think this is the main blocking issue now.  It's easy enough to tweak the test to pass (with a different board), but if you have a better idea of how it should be updated, please go ahead

src/server/milestones/TagBasedMilestone.ts
kberg: Amazing.

Borbarad2:
> > @ilevit: Do you have a discord contact as well?
> 
> I'm in the Discord chat as tts_user333, just been lurking for a week or 2
I can't find you there!?!

ilevit:
> OK. I've been away from this one but let's look at it some more. TL;DR, let's have only one milestone that's like Forester / Irrigator. But it's tricky work, and I'll work with you.
> 
> Here's my question: is it meaningfully important that there are two milestones that are almost exactly the same except for their count? Unless we do some work, using pure random milestones can bring both into the same game.
> 

Great catch

> If we did that, which is better? For instance, the Forester from the fan expansion or the new one? I don't have a strong preference, but if we choose the old one, it might confuse some people who don't know there are fan maps outside of the ones from the recent Kickstarter. If we choose the new one, we have to make sure to not break existing games. And we would have to go through a safe renaming (which I can help with.)

My preference would be to settle on the new/"official" version as the canonical one.  Good point on the in-flight games and games that use the fan version of the board.

This raises a question about the old/fan versions in general of how long do we want to support a very similar version with small differences or do we want to consolidate on the newer versions entirely and deprecate the old boards so they stop showing up as options for new games.

As long as we do want to support the old board versions, sounds like we can introduce a 'deprecated' field and update the random milestone/award logic to filter out all deprecated M/As from new games that use random.  We'll need to continue to support the old versions for the old board versions indefinitely until we decide to sunset those boards entirely.

kberg:
> Actually, thinking about it a little more and looking a bit more closely at the current state, is this already an existing problem? We already have Spacefarer and V. Spacefarer. So do some boards randomly get both?

Good point! Argh, I'm sorry. This really needs some focus.

image


I don't know that V Spacefarer is used anywhere.

image

ilevit:
> I don't know that V Spacefarer is used anywhere.

I have another PR coming where I'm going to use it for Vastitas Borealis Novus, since 4 tags is the official version.  So I guess both should be kept around if we want the other one for Terra Cimmeria (the fan/non-official version)

kberg:
I resolved merge conflicts with `Milestones.ts` and `Awards.ts`. I hope they are correct.

src/server/ma/MAManifest.ts
```diff
@@ -6,6 +6,9 @@ type MAManifestSpec<V> = {
   // Creates a new instance of this Milestone or Award.
   Factory: new () => V;
   compatibility?: Expansion | undefined;
+  // Should not be included in any new games when picking randomly.  Can only be set to true if set at all to avoid
+  // 3-value problem (unset, true, false).
+  deprecated?: true;
```
kberg, line 11: This is a nice attribute and documentation

kberg:
Thank you for your patience.

PR #7718, opened 2025-11-03
Explicitly specify --experimental-transform-types for ts-mocha server test, since it otherwise fails to build at least on Windows

ilevit:


kberg:
The reason Windows no longer did CI tests was to reduce build and test time on pulls. Manually running `npm run test` doesn't break the test for you? How odd.

ilevit:
The new flag doesn't hurt Linux but it does seem to fix Windows.  The tests build and run on Windows with this.

ilevit:
Updated slightly because a small change snuck in before (changing the exclude pattern to an --exclude flag)

PR #7721, opened 2025-11-03
Update Utopia Planitia with the official awards and milestones

ilevit:
Introduced new variant of Smith for UP with the official 6 value.

Updated UI to now consider Utopia Planitia an official board (as official as Hellas, Elysium anyway).

Also, updated the logic in game creation and generally cleaned up the logic there to be more straightforward.

"Suburbian" and "Edgedancer" seem like the same award (by a different name) but Edgedancer is better-supported (e.g., has an image asset) so went with that.


src/client/components/create/CreateGameForm.vue
```diff
@@ -184,6 +183,7 @@
                                     &nbsp;<a :href="boardHref(boardName)" class="tooltip" target="_blank">&#9432;</a>
                                   </template>
                               </label>
+                              <div v-if="boardName==='random official'" class="create-game-subsection-label" v-i18n>Fan-made</div>
```
kberg: I appreciate what you're trying to do here. However, I'm pretty sure the three new maps are considered fan-made, and not official. If things have changed and they are official, then not just Utopia Planitia goes in, but the two maps with the Novus suffix.

Rather than have two changes in one PR, please remove this particular feature from here, and we can discuss that separately.

NB https://boardgamegeek.com/boardgameexpansion/365385/terraforming-mars-terra-cimmeria-big-board-edition is called a "fan-made" board.

kberg (commented):
I'll do more review in a little while.

src/client/components/create/CreateGameForm.vue
ilevit: > I appreciate what you're trying to do here. However, I'm pretty sure the three new maps are considered fan-made, and not official. If things have changed and they are official, then not just Utopia Planitia goes in, but the two maps with the Novus suffix.

I consider this official since this is listed on the official FryxGames page with Jacob Fryxelius listed as the designer: https://fryxgames.se/product/terraforming-mars-utopia-cimmeria/

They may have started as fan-made, but at this point there are at least official versions of them.  I don't actually know whether the fan/non-official versions are worth maintaining at this point.

> 
> Rather than have two changes in one PR, please remove this particular feature from here, and we can discuss that separately.

I agree that both Terra Cimmeria and Amazonis Planitia are both also now commercial/official per the pages https://fryxgames.se/product/terraforming-mars-amazonis-vastitas/ and the above URL.  The reason I didn't move them into that category yet is because, per the Discord, there are other differences that need to be patched up before we sync to the official version; see comment from snakelauncher on 9/2/2025 summarizing the differences in detail.  I think once we sync the maps, awards, and milestones, we can eliminate some of the duplication/confusion by consolidating on a single version.

> 
> NB https://boardgamegeek.com/boardgameexpansion/365385/terraforming-mars-terra-cimmeria-big-board-edition is called a "fan-made" board.

That page does acknowledge that there's also a commercially-published version ("Replaced with Fundraiser 12 M€ production, in the commercially-published version")

ilevit: I've split up the change into 2 separate commits, we can discuss if we want to handle the official change as a separate pull request entirely

kberg: @ilevit what I mean is a separate PR, where it could be discussed and decided upon separately,  because it holds up resolving the other feature.

Either way, my take is that they're not official maps. I'm not saying that can't change, but let's discuss it separately, so the milestones themselves can be managed on their own.

The other delay is that I'm significantly reducing the time I spend on this project, so that, too, will take as long as it takes. I hope you understand.

ilevit: > @ilevit what I mean is a separate PR, where it could be discussed and decided upon separately, because it holds up resolving the other feature.

Updated to remove the second change from this PR entirely

> 
> Either way, my take is that they're not official maps. I'm not saying that can't change, but let's discuss it separately, so the milestones themselves can be managed on their own.
>

SG, I'll reach out separately
 
> The other delay is that I'm significantly reducing the time I spend on this project, so that, too, will take as long as it takes. I hope you understand.

Sorry to hear that, though of course I understand

kberg:
The comment I made about milestones and awards that already exist applies here too.

src/common/ma/MilestoneName.ts
```diff
@@ -75,6 +75,9 @@ export const milestoneNames = [
   // Ares Extreme
   'Purifier',
 
+  // Utopia Planitia
```
kberg: Thanks for all this work, but I think we should simplify it. Instead of creating a new U. Smith, just change Smith to the new value, and we can move on. Yes, it could disrupt existing games, but in this case, I'll live with it.

kberg (commented):
It will at least vastly simplify things.

src/common/ma/MilestoneName.ts
Borbarad2: Yes, please!!!
Same could be done for forester, its just the value (take the new one)
Same could be done for Costguard and Irrigator, its just the value (take the new one: costguard)
Same could be done for suburbian and Edgedancer, its just the name both are the same anyway, use the new one suburbian and take over the image.
Same could be done for Spacefarer and V.Spacefarer. Just use the new modular milestone, and adapt the other one, and call it just Spacefarer. 
Delete the old smith, forester, Irrigator, Edgedancer and Spacefarer from the code. 
For all fan made map with these MAs, it is good if they are adapted anyway and nobody would care if the old one are gone. Everyone would be happy if the new same modular one's are used. The old MAs were the inspiration for the new one. So just adapt them!

kberg:
Please let us focus on the change at hand.

src/common/ma/MilestoneName.ts
ilevit: > Thanks for all this work, but I think we should simplify it. Instead of creating a new U. Smith, just change Smith to the new value, and we can move on. Yes, it could disrupt existing games, but in this case, I'll live with it.

OK updated as requested.  I'll do a separate PR at this point for the synergies fix since it's no longer related to this change at all.

PR #7724, opened 2025-11-05
Additions and fixes to the Finnish translation

samuelmr:
Some fixes based on another played game.


src/locales/fi/ui.json
```diff
@@ -1019,6 +1019,24 @@
   "Save 0": "Tallenna 0",
   "Save 1": "Tallenna 1",
   "Save 2": "Tallenna 2",
+  "Save 3": "Tallenna 3",
```
kberg, line 1022: Do you remember why these were added to begin with? I don't see why they're here.

PR #7733, opened 2025-11-13
Update German translation UI due to recent changes

Borbarad2:
Lots of UI changes due to recent changes and forgotten ui


src/locales/de/ui.json
```diff
@@ -1021,6 +1029,24 @@
   "Save 0": "Speichern 0",
   "Save 1": "Speichern 1",
   "Save 2": "Speichern 2",
+  "Save 3": "Speichern 3",
```
kberg: What feature prints this?

Borbarad2: You approved it as well in: "Additions and fixes to the Finnish translation #7724", where he did the same!

I'm not sure exactly what feature this has, but I can remember it happening several times, e.g. with .... (I will look it up).
If not necessary, then of course I will take it out again, but then I wonder why the Finnish translation need it.

kberg: My guess is it's not necessary. I suggest removing it for now. I'll do the same with the Finnish translation.

Borbarad2: Okay, I am fine with that, but I will remove it myself.

Borbarad2 (commented):
Okay, I am fine with that, but I will remove it myself.

src/locales/de/ui.json
kberg: Yes, let's remove it. I'm going to make those message-translations, and there's already text above for it.

PR #7758, opened 2025-11-26
Vastitas Borealis Novus official

ilevit:
Sync to official version of the Vastitas Borealis Novus board.  Update temperature bonus tile and Milestones.


kberg (approved):
Thanks for doing this. I appreciate the extra work. I considered moving both maps to 4MC, but was too lazy.

This is a good change.

ilevit:
> Thanks @ilevit for doing this! Just wanted to point out that the official 5th award is Blacksmith (highest production of steel and titanium combined), not Miner (total steel and titanium resources). Reference: https://fryxgames.se/product/terraforming-mars-amazonis-vastitas/

Thank you for flagging.  Put out https://github.com/terraforming-mars/terraforming-mars/pull/7775 to address the accidental inclusion of the wrong award.

PR #7773, opened 2025-12-07
Optimize logic in DarksideObservatory to not bother asking for input...

ilevit:
…when there's only 1 option available to add resources to


kberg (approved):
This is a good PR

PR #7775, opened 2025-12-09
Follow-up fix for Vastitas Borealis, introducing the Blacksmith award…

ilevit:
… (most titanium + steel production) and assigning it to the board.  This is a follow-up to https://github.com/terraforming-mars/terraforming-mars/pull/7758#issuecomment-3629684879 which points out the incorrect award previously assigned.


kberg (approved):
This looks good.

PR #7780, opened 2025-12-14
dcep93/customMiddleware2

dcep93:
Now that we're using the `fetch` api (thanks @kberg ), we can more simply and cleanly define `customMiddleware.ts`

I'm working on creating an example of a good utilization of this, but [here's](https://github.com/terraforming-mars/terraforming-mars/pull/7675#issuecomment-3368636242) an old video for reference.


kberg:
Hi @dcep93 -- I understand now what this is doing. However, this is not an appropriate feature. It allows _anybody_ to send a malicious URL with arbitrary Javascript. If you would like to discuss this further, I will be happy to do so. In the meantime I am closing this issue.

PR #7806, opened 2026-01-06
French Translation for Promo pack & Log Messages

cv-side-projects:
Hi,

Some translations for french locales, if there is any French proofreader I'll be glad to have a review!

Some comments or assumptions I made:
- 16 PSYCHE -> 16 PSYCHÉ : https://fr.wikipedia.org/wiki/(16)_Psych%C3%A9
- Carbon Nanosystems: I used PSYCHROPHILE card as an example
- Missed 3 cards: Weather Balloons, Vermin, Static Harvesting. I added them in the json file (was there a particular reason they were missing?)
- Some typos in log_messages: 
a. 1 steel and 1 plants
b. 1 graphene(s)
c. "// ${0} moved 1 animal from Bioengineering Enclosure to ${1}." <- is this intended?
d. "${0} placed a ${1} tile at $2" <- is this intended?
- What is Neutral Player in log_message line 347 ? I just wanted to have more information about it to be sure I translated well
Additionally, I've found a complete French translation for Moon extension here (only for a next PR I'll make in the future): https://drive.google.com/drive/folders/1o2wk2sDcsHVbXuKC895hfrmYML4QSGVp

Thanks a lot!

### DIFF
--- a/src/locales/fr/promo_cards.json
+++ b/src/locales/fr/promo_cards.json
@@ -35,7 +35,7 @@
   "Action: Gain 1 M€ for each CITY AND SPECIAL TILE adjacent to an ocean.": "Action : Gagnez 1 M€ pour chaque TUILE SPÉCIALE OU CITÉ adjacente à un Océan.",
   "Requires 4% oxygen.": "Nécessite au moins 4% d'oxygène.",
   "Neptunian Power Consultants": "Consultants en énergie Neptunienne",
-  "Effect: When any ocean is placed, you MAY pay 5 M€ (steel may be used), to raise energy production 1 step and add 1 hydroelectric resource to this card.": "Effet : Lorsqu'une tuile Océan est placée, vous POUVEZ 5 M€ (vous pouvez utilisé l'acier) pour augmenter votre production d'énergie de 1 et ajouter une ressource hydroélectrique sur cette carte.",
+  "Effect: When any ocean is placed, you MAY pay 5 M€ (steel may be used), to raise energy production 1 step and add 1 hydroelectric resource to this card.": "Effet : Lorsqu'une tuile Océan est placée, vous POUVEZ payer 5 M€ (vous pouvez utiliser l'acier) pour augmenter votre production d'énergie de 1 et ajouter une ressource hydroélectrique sur cette carte.",
   "1 VP per hydroelectric resource on this card": "1 PV par ressource hydroélectrique sur cette carte.",
   "Mars Nomads": "Nomades de Mars",
   "Action: MOVE THE NOMADS to an adjacent, non-reserved empty area and collect THE PLACEMENT BONUS as if placing a special tile there. No tiles may be placed on the Nomad area.": "Action : DÉPLACEZ LES NOMADES à une tuile adjacente vide et non reservée et collectez le BONUS DE PLACEMENT comme si vous y placiez une tuile spéciale. Aucune tuile ne peut être placée sur les Nomades.",


src/locales/fr/log_messages.json
```diff
@@ -188,252 +188,252 @@
...
-  "${0} doubled resources on ${1} from ${2} to ${3}": "",
-  "${0} linked ${1} with ${2}": "",
-  "${0} turned ${1} face down to gain 3 TR and lost ${2} M€": "",
-  "${0} added a disease to ${1} and lost ${2} M€": "",
-  "${0} added ${3} diseases to ${1} and lost ${2} M€": "",
+  "${0} found life!": "${0} a chatté de la vie !",
+  "${0} gained ${1} energy": "${0} a gagné ${1} énergies",
+  "${0} decreased energy production 1 step to gain 8 M€": "${0} a diminué sa production d'énergie de 1 pour gagner 8 M€",
+  "${0} stole ${1} ${2} from the neutral player": "${0} a volé ${1} {2} d'un joueur neutre",
+  "Neutral player received ${0} M€ from ${1} owner (${2})": "Le joueur neutre a reçu ${0} M€ du propriétaire ${1} (${2})",
```
kberg, line 347: In this specific case, the neutral player is still the fake opponent in solo. It's used for Mons Isurance. When the solo player plays Mons Insurance and removes resources from the neutral player, the solo player "pays" the neutral player. In reality of course, the money is just disposed of.

src/locales/fr/log_messages.json
```diff
@@ -188,252 +188,252 @@
...
+  "${0} blackmailed the neutral player and was paid 10 M€.": "${0} a fait chanté le joueur neutre et a été payé 10 M€.",
+  "${0} blackmailed ${1} and was paid 10 M€.": "${0} a fait chanté ${1} et a été payé 10 M€.",
+  "${0} sued ${1} who had ${2} M€.": "${0} a poursuivi ${1} qui avait ${2} M€.",
+  "${0} selected Global Event ${1} for the current global event": "${0} a sélectionné l'Évènement Global ${1} comme Évènement Global actuel",
+  "Neutral delegate added to ${0}": "Délégué neutre ajouté sur ${0}",
+  "${0} searched for ${1} tags": "${0} a recherché des badges ${1}",
+  "earth": "terre",
+  "venus": "vénus",
+  "jovian": "jovien",
+  "mars": "mars",
```
kberg: Remove this translation, or return it to "". That's why the translation won't work. It rejects when someone just repeats the english translation. (I suggest returning it to "")

src/locales/fr/promo_cards.json
```diff
@@ -1,109 +1,109 @@
 {
-  "New Holland": "",
-  "Requires 4 city tiles ON MARS. Increase your M€ production 3 steps. Place a city tile on top of an already placed ocean tile, FOLLOWING NORMAL CITY PLACEMENT RESTRICTIONS. The tile counts as a city and an ocean.": "",
-  "Protected Growth": "",
-  "Oxygen must be 7% or less. Gain 1 plant per power tag you have.": "",
-  "Public Baths": "",
-  "Requires 6 oceans. Gain 6 M€.": "",
+  "New Holland": "Néo Pays-Bas",
```
kberg, line 2: Why not "Hollande"?

src/locales/fr/log_messages.json
cv-side-projects: Ok, I think then the translation is on point!

src/locales/fr/promo_cards.json
cv-side-projects: In French, we use Pays-Bas, I don't think Hollande exists. Plus, it's a former president name :')

src/locales/fr/log_messages.json
cv-side-projects: changed to "mars": "" as suggested. 

I also briefly checked I didn't put any other duplicates like this. Should I create a pipeline that checks if there is no duplicate at all?

src/locales/fr/promo_cards.json
cv-side-projects: Also the term "Néo" feels more appropriate, we often use it in sci-fi novels in futuristic cities like Néo Paris

kberg (approved):

PR #7817, opened 2026-01-17
Update server sources to use @ for TS imports

ilevit:
Makes moving files around much simpler than when relative paths are used

Updated various settings.json scripts command to support these changes


ilevit:
> I would love to have a similar solution. There is one thing this breaks, which is running the server, and unit tests, with the integrated VSCode debugger. This PR breaks those and once it doesn't this can go in. I will provide two examples, below.

Ah thanks for the heads up on this.  I don't use VSCode so that's probably why I didn't catch this.  I'll look for a solution and then circle back to this once I find something.

kberg:
Or, I'll take a look. I mean, I'm the one using it.

kberg:
Update: I think everything is working.

PR #7845, opened 2026-02-06
Update to Synergy Matrix v7.8

ilevit:
Includes new synergy matrix (exported to a CSV file and then transformed into .json format) from Borbarad

Including the parsing/transforming tool (turns a .csv into .json), generated by Gemini AI (first as python then migrated to TS).  Paths are currently hardcoded, wouldn't be too hard to update in the future.

The inline "record" generated by this was too large for the TS compiler (it couldn't "infer type" anymore) and it generally makes sense to treat this as data (which is easy to import).


kberg:
Sorry, but I am not accepting this PR. I have already spoken with borbarad that I am not taking milestone/award synergy contributions anymore, so I guess he thought going another path would work.

But also, my intention is to make sure the current synergies are based entirely on a spreadsheet. This loses that association.

I realize you might have done a lot of work here. Sorry.

Feel free to DM me on discord if you have further questions.

PR #7853, opened 2026-02-13
Translation Italian: fixes

d4nyl0:
There was a rendering problem creating a new game: "Primo giocatore" didn't fit into 40px css declares. Instead of taking into account a css hack, I prefer to translate "1st" to a short italian form "1°".
There was an error on the number for iron resources on a card.
I renamed gamesetting.json to game_setting.json as in the de locale directory.

Please, feel free to edit what you want.


src/locales/it/game_setting.json
```diff
@@ -0,0 +1,31 @@
+{
+	"Game Id:${0}":"Identificativo partita:${0}",
+  "Instructions: To start the game, separately copy and share the links with all players, and then click on your name.": "Istruzioni: per iniziare la partita, copia e condividi i link separatamente per ciascun giocatore, dopodiché fai clic sul tuo nome.",
+  "Save this page in case you or one of your opponents loses a link.": "Salva questa pagina nel caso uno dei giocatori perda il link.",
+  "copy": "copia",
+	"1st":"1°",
```
kberg, line 6: The same string in two different files will have unpredictable results. Remove the six entries here, please.

d4nyl0: I see, you're right. I wanted to rename the gamesetting.json file to game_setting.json, but both are present.
I'll close this PR, fix the commit (deleting one of the two files), and submit a new PR.
Thanks for the careful review.

PR #7864, opened 2026-02-21
Fix faraday issue with multiple same tags one card

TarunTheo13:
If a player has 4 science tags and plays Research (2 science tags), `this.data.counts[science]` would be set to 11 (before this change). With this change, that will no longer happen.


kberg:
Do you think you can add a test for this? Would you like some help?

PR #7865, opened 2026-02-21
Select all for public plans and log cards revealed concisely

TarunTheo13:
This PR creates a button so that all cards in hand can be selected at once (this is useful for quickly selecting all or most cards in a large hand). It also changes how the revealed cards are logged so they are shown in one log message and are therefore easier for opponents to see.

The 'Select All/Deselect All' toggle button is currently added only to `Public Plans` but could be added to other actions like `Sell Cards`

### DIFF
--- a/tests/client/components/SelectCard.spec.ts
+++ b/tests/client/components/SelectCard.spec.ts
@@ -20,6 +20,7 @@ describe('SelectCard', () => {
           showOnlyInLearnerMode: false,
           selectBlueCardAction: false,
           showOwner: false,
+          showSelectAll: false,
         },
         onsave: () => {},
         showsave: true,
--- a/tests/client/components/SelectInitialCards.spec.ts
+++ b/tests/client/components/SelectInitialCards.spec.ts
@@ -168,6 +168,7 @@ function createComponent(corpCards: Array<CardName>, projectCards: Array<CardNam
     showOnlyInLearnerMode: false,
     selectBlueCardAction: false,
     showOwner: false,
+    showSelectAll: false,
   }, {
     type: 'card',
     title: titles.SELECT_PROJECTS_TITLE,
@@ -178,6 +179,7 @@ function createComponent(corpCards: Array<CardName>, projectCards: Array<CardNam
     showOnlyInLearnerMode: false,
     selectBlueCardAction: false,
     showOwner: false,
+    showSelectAll: false,
   }];
 
   if (preludeCards) {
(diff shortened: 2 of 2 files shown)


src/server/cards/promo/PublicPlans.ts
```diff
@@ -34,10 +34,23 @@ export class PublicPlans extends Card implements IProjectCard {
     return new SelectCard('Select cards to reveal', 'Reveal', player.cardsInHand, {
       min: 0,
       max: player.cardsInHand.length,
+      showSelectAll: true,
     }).andThen((cards) => {
       player.stock.add(Resource.MEGACREDITS, cards.length, {log: true, from: {card: this}});
-      for (const card of cards) {
-        player.game.log('${0} revealed ${1}', (b) => b.player(player).card(card));
+      if (cards.length > 0) {
```
kberg: Hmm. This becomes a pain to translate. This could be debated in its own PR.

src/client/components/SelectCard.vue
```diff
@@ -17,6 +17,7 @@
         <div v-if="hasCardWarning()" class="card-warning" v-i18n>{{ warning }}</div>
         <warnings-component :warnings="warnings"></warnings-component>
         <div v-if="showsave === true" class="nofloat">
+            <AppButton v-if="showSelectAll" @click="toggleSelectAll" type="submit" :title="allSelected ? $t('Deselect All') : $t('Select All')" />
```
kberg, line 20: If allSelected is computed, how is that button dynamic?

kberg: I just realized I NEVER published this question. Wow.

src/server/cards/promo/PublicPlans.ts
TarunTheo13: I could drop that commit, I don't know so much about how the translations are structured - is there any documentation on them that I could consult?

src/client/components/SelectCard.vue
TarunTheo13: `allSelected` automatically re-evaluates whenever this.cards changes

src/server/cards/promo/PublicPlans.ts
kberg: https://github.com/terraforming-mars/terraforming-mars/wiki/Translations but it might not answer your question. At least, let's get this into its own PR, or at least, out of this commit.

PR #7866, opened 2026-02-21
Show number of cards selected for Sell Patents

TarunTheo13:


TarunTheo13:
> Hope you don't mind me asking: this works?
> 
> My instinct is to find a way to pass the button label down to the child component. But if this works, let's do it.

It works when I test it locally for sure, I can try to do that if you want though

src/client/components/OrOptions.vue
```diff
@@ -14,11 +14,11 @@
                               :playerView="playerView"
                               :playerinput="option"
                               :onsave="playerFactorySaved(idx)"
-                              :showsave="false"
+                              :showsave="showsave && isMultiSelectCard(option)"
```
kberg: I suggest renaming `isMultiSelectCard` to something more meaningful, because it isn't clear to me what is happening. Something like `showChildSaveButton`.

Then a comment in `showChildSaveButton` which explains that this allows the child component to control the text in the save button, which is expected to be "Save _n_" or something.

kberg (commented):
This makes sense. Modulo one comment, please.

I'm OK with having a test.

src/client/components/OrOptions.vue
kberg: ping this, and then we can get this in.

kberg: Test can be another time.

kberg (commented):
Ready to move this forward.

kberg (approved):

kberg:
Thank you!

TarunTheo13:
Thanks for wrapping this up for me

PR #7867, opened 2026-02-21
Tuning start screen

alrusdi:
An attempt to remove that annoying grey background on starting screen
grayzone


kberg (approved):
Thanks!

alrusdi:
Confirmed on portrait displays where the screen height exceeds the start page content. I'll work on a fix.

PR #7890, opened 2026-02-28
More UI fixes

alrusdi:
More UI fixes. I don't sure when it was introduced. May be my recent PRs

ui-fixes

cards-list

### DIFF
--- a/src/client/components/LanguageSelectionDialog.vue
+++ b/src/client/components/LanguageSelectionDialog.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="preferences_panel">
+  <div class="preferences_panel preferences_panel--language-switcher">
     <div class="preferences_panel_item form-group">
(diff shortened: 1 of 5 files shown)


src/client/components/LanguageSelectionDialog.vue
```diff
@@ -9,7 +9,7 @@
             :title="LANGUAGES[lang][1]"
             @click="switchLanguageTo(lang)"
           />
-          <span class="language-text" @click="switchLanguageTo(lang)">{{ LANGUAGES[lang] }}</span>
```
kberg, line 12: The behavior that currently exists is intentional, just ugly and that it shows the name of the language in English and also its own tongue. It is ugly though. But I'm okay with your change.

kberg (approved):

kberg:
Thank you!

PR #8054, opened 2026-04-18
Return full current-generation logs when explicitly requested

rusliksu:
## Summary

When the client explicitly requests the current generation, the server should return the full generation log instead of truncating it to the last 50 messages.

The default log view still keeps the lightweight tail behavior.

## Why

In larger games, the current generation can exceed 50 log entries. That makes the beginning of the current generation unreachable even when the user clicks the current generation explicitly.

## Validation

- `npx mocha --import=tsx --require tests/testing/setup.ts tests/routes/ApiGameLogs.spec.ts`
- result: `13 passing`

### DIFF
--- a/src/server/routes/GameLogs.ts
+++ b/src/server/routes/GameLogs.ts
@@ -40,12 +40,12 @@ export class GameLogs {
       }
     };
 
-    // for most recent generation pull last 50 log messages
-    if (generation === null || Number(generation) === game.generation) {
+    // Default view keeps the payload small. An explicit generation request should
+    // always return the full generation, including the current one.
+    if (generation === null) {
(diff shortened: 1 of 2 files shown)


kberg (commented):
This looks reasonable. Have you manually tested this?

rusliksu:
Manually verified.

Broken on current upstream:
- https://terraforming-mars.herokuapp.com/spectator?id=s96c71e59285a

Fixed on preview:
- https://preview.tm.knightbyte.win/spectator?id=sc1bb3b986aae

If you scroll to the top of the Game log, current upstream shows only the tail of the current generation.
On preview, the current-generation log view shows the full generation log.

kberg (approved):

kberg:
Thanks again!

PR #8062, opened 2026-04-24
Add Russian translations for milestones and awards

Almaziko:
Hello! Translated all previously untranslated milestone and award cards into Russian.


kberg (approved):
Thank you!

PR #8075, opened 2026-04-29
Add tsc-alias to resolve path aliases after server build

kberg:


kberg:
@ilevit trying again.

ilevit:
> @ilevit trying again.

Great to see, fingers crossed it works with tsc-alias this time

Issue #8088, opened 2026-05-02
Air Raid should allow partial/no M€ steal while still spending a floater


rusliksu:
I think the recent “steal X” change may be stricter than the rules intend.

This came up in real play on my local server. Several players in my group were surprised when a steal card could not be used because the target had less than the full amount of M€. The clearest case for us was Air Raid, but I think the same concern may apply to the six cards changed in that commit: Bjorn, Air Raid, Ancient Shipyards, The Darkside of the Moon Syndicate, Space Privateers, and Subnautic Pirates.

Our reading is that “steal X M€” is still resource removal. If the target has less than X, you should steal what is available. If there is nothing useful to steal, the steal part can be skipped, unless the card makes it a cost.

The FAQ seems to support this:

https://github.com/terraforming-mars/faq/wiki#when-can-my-actions-affect-an-opponents-productionresources

It says resource removal can be partial or skipped. The Mons Insurance section also says this applies whether the card says “remove” or “steal”:

https://github.com/terraforming-mars/faq/wiki#mons-insurance-pro

The Air Raid entry is a good example:

https://github.com/terraforming-mars/faq/wiki#air-raid-c02-c

It says the floater is still lost even if the player decides not to steal any M€. So Air Raid should require paying the floater, but not require a target with at least 5 M€.

This is the commit I mean:

https://github.com/terraforming-mars/terraforming-mars/commit/7a4ddc14cc6c002b229010652b62ed13e30b79a5

Maybe I’m missing a steal-specific ruling, but from these references I would expect these effects to mean “steal up to X”, not “the full X must be available.”

kberg:
Thanks - I'm kind of busy but did see what you wrote. I will get to this in
at worst a few days.

TecoChN:
The change is not stricter then intended. 
First up: For example the card subnautic pirates got an up to so its possible to steal less then 7, for example 1. 
Second the rule is based on the rule: u have to do every action at its fullest otherwise the action / card is not possible to perform. 
You said removing resources is an exeption. Thats true. BUT stealing is not just removing resources

https://boardgamegeek.com/thread/1839646/article/26813906#26813906
In this thread Jacob Fryxelius wrote: stealing requires you to actually steal something
And he also wrote: The word 'STEAL' is part of the iconography, so to speak. Since it is not part of an explaining paranthesis, it should be considered in combination with the icons.

[...]

PR #8093, opened 2026-05-04
Add help annotation for Air Raid

rusliksu:
Refs #8088.

## Summary
- add `hasExternalHelp` to Air Raid so the card links to Card Details
- add a unit test for the annotation
- leave gameplay unchanged: in multiplayer Air Raid still requires a floater and an opponent with at least 5 M€

## Requested Card Details text
@kberg, could you add the corresponding Card Details text when this lands?

Suggested text:

> Air Raid requires removing 1 floater and stealing exactly 5 M€ from one opponent. These are treated as a package deal, so Air Raid cannot be played unless an opponent has at least 5 M€. If no opponent has 5 M€, the card is unavailable even if you have a floater.

## Testing
- `npx mocha --import=tsx --require tests/testing/setup.ts tests/cards/colonies/AirRaid.spec.ts`

### DIFF
--- a/src/server/cards/colonies/AirRaid.ts
+++ b/src/server/cards/colonies/AirRaid.ts
@@ -20,6 +20,7 @@ export class AirRaid extends Card implements IProjectCard {
 
       metadata: {
         cardNumber: 'C02',
+        hasExternalHelp: true,
         description: 'Requires that you lose 1 floater. Steal 5 M€ from any player.',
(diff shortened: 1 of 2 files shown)


rusliksu:
Closing for now; I want to stage this in my fork first before sending it upstream.

kberg:
OK. Next time, there's no need for the test.

PR #8095, opened 2026-05-06
Use an agent-agnostic document.

kberg:


kberg:
@ilevit FYI

ilevit:
Different tools have different defaults for what they use.  Gemini CLI, for example, seems to [document](https://geminicli.com/docs/cli/gemini-md/) itself as only looking at GEMINI.md.  Though it can be configured by adding a project-level .gemini/settings.json I guess.

Separately, from what I've been reading lately, the current advice on how to structure things seems to be to break out separate md files for different aspects of the codebase (e.g., docs/db.md, docs/ui.md, etc.) and then refer to them in the GEMINI.md type files which become thin dispatchers.  This prevents unnecessary filling up of context with information not relevant to the task (e.g., "If you are making a database change, refer to docs/db.md" prevents it having to read about how the UI works into memory).

kberg:
eeep. Well, I'm fine with one document for guidance, maybe two, but that's
a bit more than I feel comfortable with at the moment. 🤔

kberg:
@ilevit OK, I'm removing this commit. :(

PR #8116, opened 2026-05-11
Log ocean adjacency bonus

### LINKED ISSUE #7898: Add tile placement reward next to oceans to game log
Hello,

I have noticed that when placing a tile next to an ocean, the game log does not show that you gained x amount of megacredits for placing next to the ocean. Its just quietly added to the players megacredits.

Any chance this could be added to the game log?

"Player placed an ocean tile at E6"
"Player placed a greenery at F5"
"Player gained 2 M€"

I know this is not high priority at all. In my case for example our friendgroup never got the ocean placement bonus rule taught and only noticed it when checking the manual again after quite a few games. We played a lot of online games and never noticed there either because of this.

I think it would be quite neat to be added, since the game log tracks everything quite meticulously and I found it relatively odd that this isnt tracked like other placement bonuses.

Thanks a lot either way!

rusliksu:
## Summary

- Log the standard M€ bonus gained from placing a tile next to oceans.
- Keep the existing M€ amount unchanged; the bonus now goes through `stock.add(..., {log: true})` so it appears in the game log.

Fixes https://github.com/terraforming-mars/terraforming-mars/issues/7898

## Demo

Same scenario on both servers: Teractor uses Aquifer at `06` / A4, then places Greenery at `12` / B5 adjacent to that ocean.

Heroku production: https://terraforming-mars.herokuapp.com/player?id=pfee4416424f

Heroku production log

Preview branch: https://preview.tm.knightbyte.win/player?id=p3d704af83bcb

Preview branch log

## Testing

- `npx mocha --import=tsx --require tests/testing/setup.ts tests/Game.spec.ts --grep "ocean adjacency bonus logging"`
- `npx mocha --import=tsx --require tests/testing/setup.ts tests/Game.spec.ts`
- `npm run lint:server -- --quiet`
- `npm run build`
- Preview deploy/verify OK on `preview.tm.knightbyte.win`

### DIFF
--- a/src/server/Game.ts
+++ b/src/server/Game.ts
@@ -1382,11 +1382,12 @@ export class Game implements IGame, Logger {
       this.grantSpaceBonuses(player, space);
     }
 
-    this.board.getAdjacentSpaces(space).forEach((adjacentSpace) => {
-      if (Board.isOceanSpace(adjacentSpace)) {
-        player.megaCredits += player.oceanBonus;
-      }
-    });
+    const adjacentOceanCount = this.board.getAdjacentSpaces(space).filter(Board.isOceanSpace).length;
+    const oceanAdjacencyBonus = adjacentOceanCount * player.oceanBonus;
+    if (oceanAdjacencyBonus > 0) {
+      player.stock.add(Resource.MEGACREDITS, oceanAdjacencyBonus);
+      this.log('${0} gained ${1} M€ from ${2} ocean(s)', (b) => b.player(player).number(oceanAdjacencyBonus).number(adjacentOceanCount));
+    }
 
     // TODO(kberg): these might not apply for some bonuses, e.g. Frontier Town.
     // https://boardgamegeek.com/thread/3344366/article/44658730#44658730


kberg:
Ah, I thought I commented on the original bug, I did not. I didn't want to do this to prevent too much in the logs. Although the summary is nice.

The tests are overkill. Don't you think? I would like to know why you though those tests were worth their size.

The logging is also (imo) insufficient. There are so many things that happen when someone places a tile. "X gained 2mc" is not enough.

rusliksu:
That makes sense. My bias as a player is that the log is currently under-informative rather than too noisy; for post-game analysis it is useful to know where money/resources came from. On my own server I have been adding more logs for that reason, for example draft logs showing the picked card and passed cards:

custom draft log example

For this PR, I agree the current `X gained 2 M€` line is too generic. If you are open to logging this event at all, I can change it to an explicit message like `X gained 2 M€ from ocean adjacency` and reduce the tests to one focused log test plus one multiple-ocean case. If you would rather keep this event out of the log entirely, I can close the PR.

kberg:
Tell you what: how about ${0} gained ${1} MC from ${2} oceans.? OK, let's do it.

rusliksu:
Updated. I used \M€\ to match the existing log-message convention, and \ocean(s)\ to follow the existing \step(s)\ / \card(s)\ style. I left locale files unchanged for now; I can add Russian easily, but I’d leave the other languages to native translators.

kberg:
Thanks, I still think the log tests are very unnecessary.

rusliksu:
Removed the log tests. The PR diff is now only Game.ts, and CI is green.

kberg (approved):
Thank you.

PR #8139, opened 2026-05-16
Change "mars is terraformed" from an alert to persistent banner

hd256:
The "mars is terraformed" alert is a bit annoying and pops up on every refresh. Instead switch to a banner which is floating and persistent.

Example:

image


kberg (approved):
Thank you!

PR #8141, opened 2026-05-17
Update waiting-for to include player names instead of just colors

hd256:
More colorblind-friendly.

Example:

image

### DIFF
--- a/src/client/components/WaitingFor.vue
+++ b/src/client/components/WaitingFor.vue
@@ -3,7 +3,7 @@
   <template v-if="waitingfor === undefined">
     {{ $t('Not your turn to take any actions') }}
     <template v-if="playersWaitingFor.length > 0">
-      (⌛ <span v-for="color in playersWaitingFor" :class="playerColorClass(color, 'bg')" :key="color">&nbsp;&nbsp;&nbsp;</span>)
+      (⌛ <span v-for="player in playersWaitingFor" class="log-player" :class="'player_bg_color_' + player.color" :key="player.color"> {{ player.name }} </span>)
     </template>
   </template>
   <div v-else class="wf-root">
@@ -50,7 +50,7 @@ let ui_update_timeout_id: number | undefined;
 let documentTitleTimer: number | undefined;
 
 type DataModel = {
-  playersWaitingFor: Array<Color>
+  playersWaitingFor: Array<{name: string, color: Color}>
   suspend: boolean,
   savedPlayerView: PlayerViewModel | undefined;
 }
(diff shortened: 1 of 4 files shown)


src/client/components/WaitingFor.vue
kberg, line 1: All of this can be done in `WaitingFor.vue`. Since `WaitingFor` receives `players`, and each one has `color` the name can come from there.

src/client/components/WaitingFor.vue
```diff
@@ -3,7 +3,7 @@
   <template v-if="waitingfor === undefined">
     {{ $t('Not your turn to take any actions') }}
```
kberg, line 4: Side note: I wonder if this should be reduced to 'Not your turn'.

src/client/components/WaitingFor.vue
```diff
@@ -3,7 +3,7 @@
...
     <template v-if="playersWaitingFor.length > 0">
-      (⌛ <span v-for="color in playersWaitingFor" :class="playerColorClass(color, 'bg')" :key="color">&nbsp;&nbsp;&nbsp;</span>)
```
kberg, line 6: Why is `playerColorClass` unused?

kberg (changes requested):

src/client/components/WaitingFor.vue
hd256: Fixed to reuse `playerColorClass`

src/client/components/WaitingFor.vue
hd256: SGTM but i'll leave that for a different change. I'm not sure if there are other l18n changes that need to occur as a result.

src/client/components/WaitingFor.vue
hd256: Not sure what you mean. The waiting for data model only has `waitingFor: Array<Color>`, it does not pass the entire Player object, which is why I need to update that to also include the name.

hd256:
Fixed tests as well in latest push

src/client/components/WaitingFor.vue
kberg: You are correct about the API. However, the component has a `players` property that contains the names.

kberg: If you remove all the server-side stuff and add this method to `WaitingFor.vue`

```
    playerName(color: Color) {
      const player = this.players.find((p) => p.color === color);
      return player?.name ?? '';
    },
```

then the key change on line 6 could become something like

```
(⌛ <span v-for="color in playersWaitingFor" class="log-player" :class="playerColorClass(color, 'bg')" :key="color">
{{ playerName(color) }}
</span>)
```

and that would pretty much be all of it.

kberg (changes requested):

src/client/components/WaitingFor.vue
hd256: I see now. Done.

kberg (approved):
Thanks!

kberg:
Thank you!

PR #8143, opened 2026-05-17
Add Planetary track to player setup view and fix cutoff for details section in player setup view.

hd256:
1. add planetary track to "Board" section in the Player Setup accordion (if Pathfinders is enabled).
2. spectre accordions have a default a default max-height of 50rem which cuts off the board view in the player setup phase. Remove this max-height limit to allow all sections to render.

Before:

image

After:

image

### DIFF
--- a/src/client/components/PlayerSetupView.vue
+++ b/src/client/components/PlayerSetupView.vue
@@ -87,6 +87,8 @@
 
         <turmoil v-if="game.turmoil" :turmoil="game.turmoil"></turmoil>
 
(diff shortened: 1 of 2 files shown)


kberg:
I don't see a meaningful difference between the two screenshots as they pertain to Planetary tracks

kberg:
~~Also, I don't think we use anything related to spectre css. So what's the CSS change for?~~

kberg:
Oh wait we do use spectre. Huh

kberg (approved):
Ignore my comments. Thanks for this

Borbarad2:
Hi, great stuff here.
Could this be done for the moon cut off to player setup as well? see issue#3797
Thanks in advanced

hd256:
I think this fix resolves that issue as well.

PR #8146, opened 2026-05-17
Auto-select card for Soil Enrichment when only one is eligible

ilevit:
* Skip `SelectCard` prompt in `play` method if only one card has microbes to remove
* Add unit test to verify automatic resource removal and plant gain when one card is eligible


kberg (approved):
Good one. Thank you!

PR #8149, opened 2026-05-17
Log Greens policy greenery bonus

rusliksu:
## Summary

- Log the 4 M€ bonus gained from placing a greenery while the Greens ruling policy is active.
- Keep the existing M€ amount unchanged; the bonus now has an explicit source in the game log.

## Demo

Scenario: Greens ruling policy is active, blue places a greenery during the action phase.

Before: the player gained 4 M€ with no source line.

After:

`blue gained 4 M€ from Turmoil Greens policy`

## Testing

- `npx mocha --import=tsx --require tests/testing/setup.ts tests/turmoil/parties/Greens.spec.ts`
- `npm run build:server`


kberg:
Why'd you close it?

PR #8150, opened 2026-05-17
Log Greens policy greenery bonus

rusliksu:
## Summary

- Log the 4 M€ bonus gained from placing a greenery while the Greens ruling policy is active.
- Keep the existing M€ amount unchanged; the bonus now has an explicit source in the game log.

## Demo

Scenario: Greens ruling policy is active, blue uses the Greenery standard project and places a greenery on A1.

Before on Heroku: the player receives the +4 M€ policy bonus, but the game log does not show a source line.

Before: Heroku missing Greens policy log

After on preview: the game log includes the policy source line.

After: preview logs Greens policy source

Preview: https://preview.tm.knightbyte.win/player?id=p3c9f59981413

Before-fix comparison: https://terraforming-mars.herokuapp.com/player?id=p89853207ca42

## Testing

- `npx mocha --import=tsx --require tests/testing/setup.ts tests/turmoil/parties/Greens.spec.ts`
- `npm run build` during preview deploy
- Preview deploy/verify OK on `preview.tm.knightbyte.win`

### DIFF
--- a/src/server/turmoil/parties/Greens.ts
+++ b/src/server/turmoil/parties/Greens.ts
@@ -62,6 +62,8 @@ class GreensPolicy01 implements IPolicy {
   onTilePlaced(player: IPlayer, space: Space) {
     if (Board.isGreenerySpace(space) && player.game.phase === Phase.ACTION) {
       player.stock.add(Resource.MEGACREDITS, 4);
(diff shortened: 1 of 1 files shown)


kberg (approved):
This is good. You made me realize I should probably improve the message even more. So expect something that does this but on a larger scale.

Issue #4261, opened 2022-03-31
Undo: Government support and baying card


Buheda:
1. As I can see there is no Undo after Government support, as this step after finishing action. 
But as a fact this step can also be different, we still can decide which parameter should be updated. 

2. Maybe baying card also can be resumed ?

kberg:
@Buheda I'm sorry! I can't quite understand what you're asking. I think you're asking for more undo in some of the areas where it would help.

Buheda:
@kberg  Hello, in general - yes. More undo.
I asked about:
1. undo after Government support in Venus. We can decide which parameter can be updated.
2. And buying a card at the start of a generation. Some card was bought by mistake or bought not what wanted.

Buheda:
Hello. There is still no any action? I'm faced up again with this issue(((

kberg:
Well, it's been a while, but you're right. I haven't done anything here. I see why it might be useful to have undo support in those areas but I don't think it's really reasonable.

1. After World Government Terraforming, players are presented with cards from the top of the draw pile. It is not fair to allow the player to change WGT after people see project cards. You could argue that the player should just get an extra chance to confirm their WGT selection _before_ players see new project cards, but I don't think that's worth doing.

2. Once a player chooses which cards they want to buy, if other players haven't finished their choices the player might still have a chance to change their selection. But that will be tough to do.

Buheda:
@kberg  Thank you for answer. 
1. Yes, I thought about undo before new cards. On my opinion it is very logical, the similar as any other undo moves. In solo mode it have no influence on other players at all. Why WGT should be worst than any other actions?
2. I understand draft mode - it is impossible and not logical. But while game not started - can't understand. And again in solo mode also no impact to another players.

Or there is complexity with changing history to undo, as history page with undo starts from new buying cards?

kberg:
I understand your perspective, but I don't agree. In my opinion you should never be allowed to undo after you get new information.

Buheda:
in case above "By mistake I selected to increase Venus (to 8%) by Government support and didn't get card, and couldn't undo it" there was no new information.

kberg:
Oh? What was next?

Buheda:
What can be after WGT? - selecting cards to buy for next generation. There are many cases when we can undo moves after opening cards to buy. Why this case worst ?

kberg:
The drawn cards are the new information. Undo after that is not fair.

Buheda:
[...]

PR #8151, opened 2026-05-18
Support for re-picking drafted cards

hd256:
This change enables players to re-pick during drafts as long as not all players have chosen a card for the current round.

In accomplishes this by adding a "polling" field to inputs. When set to true, a given input will wait for updates while simultaniously giving players the defined input options. Players can either wait or click the option. This is necessary to make the game advance once all players have chosen a card, but still providing the option to "re-pick" in the meantime.

Polling inputs will not do the animations that typically trigger when waiting for a player input (animated title bar, animated side-bar, etc) since the input is optional.

Additionally, show the cards which the player did not choose for the current draft as disabled cards.

Example:

image


Borbarad2:
Hi, great stuff here.
Could you make sure that all text, which could be read by the players and all UI-text is i18n compatible. Which is needed for good translations.
Thanks in advanced

hd256:
> Could you make sure that all text, which could be read by the players and all UI-text is i18n compatible.

How do I do that? What exactly does it entail?

src/server/Draft.ts
```diff
@@ -107,6 +113,30 @@ export abstract class Draft {
...
+   */
+  private askPlayerToRepick(player: IPlayer): void {
+    const giveTo = this.givingTo(player);
+    const option = new SelectOption(
+      message('Passing cards to ${0}. Repick draft card?', (b) => b.player(giveTo)),
+      'Repick',
+      undefined, // warnings
+    );
+    option.polling = true;
+    player.setWaitingFor(
```
kberg: What happens if there is a race condition between a player choosing new cards, and the other players completing their draft pick? I suspect it's a problem.

src/server/Draft.ts
```diff
@@ -140,10 +171,25 @@ export abstract class Draft {
...
+      this.askPlayerToRepick(player);
       this.game.save();
       return;
     }
 
+    // Clear any pending Undo options for all players
+    let waitingForCleared = false;
+    for (const p of this.game.players) {
+      if (p.getWaitingFor()?.type === 'option') {
+        (p as any).waitingFor = undefined;
```
kberg: Please get rid of `as any`. What are you trying to achieve here?

src/server/Draft.ts
```diff
@@ -140,10 +171,25 @@ export abstract class Draft {
...
+        (p as any).waitingForCb = undefined;
+        waitingForCleared = true;
+      }
+    }
+    // Advance game age to ensure that any players still waiting will clear out the repick input.
+    if (waitingForCleared) {
+      this.game.gameAge++;
```
kberg: gameAge is tightly coupled to saves. Can you choose something else?

src/server/IPlayer.ts
```diff
@@ -119,6 +119,8 @@ export interface IPlayer {
   draftHand: Array<IProjectCard>;
   /** Cards this player has already chosen during this draft round */
   draftedCards: Array<IProjectCard>;
+  /** Cards this player has not chosen during this draft round */
```
kberg: So these are the ones that are going to be passed to the next player?

src/server/PlayerInput.ts
```diff
@@ -17,6 +17,11 @@ export interface PlayerInput {
      * Used only when this option is a child option of an OrOptions.
      */
     eligibleForDefault?: boolean;
+    /**
+     * When true, this indicates that the user should continue to poll for updates
+     * even if waiting for an input from that user.
```
kberg: from that user or from other users?

src/client/components/PlayerHome.vue
```diff
@@ -57,6 +57,9 @@
         <div v-for="card in playerView.draftedCards" :key="card.name" class="cardbox">
           <Card :card="card"/>
         </div>
+        <div v-for="card in playerView.unchosenDraftCards" :key="card.name" class="cardbox">
```
kberg: So if I understand it, let's say I'm part way through a draft. I've selected my second card. So I'll see two cards without shading, and two that are shaded. Is there a visual cue as to which card is locked in from the prior pass?

kberg (commented):
This is exciting. I think this PR cuts through a few abstractions, and I'd love to resolve them.

I need to look at all of this much more closely, but this is at least a good start.

kberg:
Hi -- what are your thoughts?

src/server/Draft.ts
hd256: I'm trying to clear out the waitingfor which I couldn't do via the interface. Added new fn `clearWaitingFor()` to IPlayer.

src/server/Draft.ts
hd256: Actually, this isn't needed, so I removed it. It isn't needed because we save after each draft round which advances the game age. In our custom fork, we use this in some places where a save doesn't happen which is why we use this, but it's not needed here.

src/server/IPlayer.ts
hd256: Correct.

src/server/PlayerInput.ts
hd256: From that user. From my understanding the client doesn't poll for updates via waitingFor if it's waiting for an input from a user. This changes that behavior such that if `polling` is set to true for an input, it will poll for updates while simultaneously also able to accept an input for the user.

src/client/components/PlayerHome.vue
hd256: Currently no, but I could add one. In our fork, we add highlighting for the most recently picked card, but another way to tell is that the most recently picked card will be the right-most card without shading.

Here's what the highlighting looks like in our server:

image

kberg:
Thanks for the feedback. This morning i thought of an alternative approach that might eliminate the need to store data and deal with the race condition. I forgot the details and it's late right now. Maybe this will just be like Fermat's Last Theorum ... kberg's last code reivew? (eep I hope not!)

have i mentioned I'm tired? :D

src/server/Draft.ts
hd256: This is a valid concern. I'm not yet familiar enough with the overall architecture of the game or TS development to know where to put a fix for this. I suspect some high level lock is probably sufficient here as performance isn't too big of a concern but even that I'm not sure the best place to put that. Let me know if you have ideas/guidance you can offer?

hd256:
sorry for the delay in my response. just been a bit busy with other stuff. Happy to hear your alternative approach

src/server/routes/ApiWaitingFor.ts
```diff
@@ -16,11 +16,20 @@ export class ApiWaitingFor extends Handler {
...
   private timeToGo(player: IPlayer): boolean {
-    return player.getWaitingFor() !== undefined || player.game.phase === Phase.END;
+    const input = player.getWaitingFor();
+    if (input !== undefined) {
+      return !input.polling;
+    }
+    return player.game.phase === Phase.END;
   }
 
   private playersWithInputs(game: IGame) {
```
kberg: The name isn't correct anymore is it?

src/server/routes/ApiWaitingFor.ts
kberg, line 1: This class is a little more confusing now with `polling`. For instance, have the meanings of `timeToGo` and `playersWithInputs` still doing the exact same thing they were?

src/server/Draft.ts
```diff
@@ -63,6 +64,7 @@ export abstract class Draft {
 
     for (const [player, draftHand] of zip(this.game.players, arrays)) {
       player.draftHand = draftHand;
+      player.unchosenDraftCards = [];
```
kberg: I don't like adding `unchosenDraftCards`. When I say "I don't like" I'm willing to hear it's the right option, but I really thing there's got to be something more elegant. Is it possible to get rid of this new field?

Let's see if comments below facilitate that.

src/server/Draft.ts
```diff
@@ -107,6 +113,30 @@ export abstract class Draft {
     return this.passDirection() === 'after' ? this.game.getPlayerAfter(player) : this.game.getPlayerBefore(player);
   }
 
+  /**
+   * Ask the player if they want to repick their card.
+   */
+  private askPlayerToRepick(player: IPlayer): void {
```
kberg: Perhaps instead just return SelectCard. Then the player can just choose another card. The actively selected card could probably be marked disabled or something so it is hidden and unselectable.

src/server/Draft.ts
```diff
@@ -140,10 +171,25 @@ export abstract class Draft {
...
     if (this.game.players.some((p) => p.needsToDraft)) {
+      this.askPlayerToRepick(player);
       this.game.save();
       return;
     }
 
+    // Clear any pending Undo options for all players
+    let waitingForCleared = false;
+    for (const p of this.game.players) {
+      if (p.getWaitingFor()?.type === 'option') {
```
kberg: Still using as any.

src/server/Draft.ts
```diff
@@ -140,10 +171,25 @@ export abstract class Draft {
 
     // If anybody still needs to draft, stop here.
     if (this.game.players.some((p) => p.needsToDraft)) {
+      this.askPlayerToRepick(player);
       this.game.save();
       return;
     }
 
+    // Clear any pending Undo options for all players
```
kberg: What does this comment mean? Undo has nothing to do with this.

tests/drafting.spec.ts
kberg, line 1: There's going to need to be a real test here. Hmm.

src/server/Draft.ts
kberg: Yeah, hmm. Yes. The SelectCard for drafting needs to block redrafting when that particular draft has concluded. I guess?

src/server/PlayerInput.ts
kberg: I wonder if it should poll when you're waiting for anyway, particularly during drafts, just to know who has selected stuff. (Drafts, and also certain actions like initial selection and one of the colonies, I can't remember which.

kberg (commented):
There may be some comments talking  over what you have said. Apologies if so.

This is definitely a little better now since gameAge is gone.

There's merge conflicts now, I'm about to resolve them.

hd256:
I like the idea of just having a consistent select option which just allows you to switch the drafted card instead of going back and repicking. I'll try to take that approach and see how it shapes up. But I don't have too much time so it might take a while. If you want to close this in the meantime to keep a clean queue, feel free. I think your comments are valid but because my initial impl was just for our fork, I didn't take a lot of these things into consideration.

kberg:
Well, help me this way: why did you add the undraftedCards attribute? Maybe I can finish it up.

hd256:
That's just to show the cards you didn't pick in the UI so you can see what cards you can re-pick without clicking repick. If we switch to just always allowing repicking, we wouldn't need it.

kberg:
Thanks. The other thing that's nice about getting rid of the SelectOption
is that it simplifies dealing with a race condition. I think.

src/server/Draft.ts
hd256: holdover from when this was called Undo in our fork, fixed comment

hd256:
I had some time today and was able to implement the consistent repicking with SelectCard and not SelectOption. PTAL. I also removed undraftedCards since it's no longer necessary. I did NOT fix the tests though or add new ones.

I also stop the timer from running when polling is set to true but TBH, I'm not fully confident if that change in Player is sufficient broadly.

src/server/Draft.ts
hd256: I think this is resolved by the new structure.

src/server/Draft.ts
hd256: removed

src/server/Draft.ts
hd256: fixed

hd256:
@kberg let me know if you have any large concerns? If not, i can add tests so that we can hopefully finalize this change and get it merged.

src/client/components/PlayerHome.vue
```diff
@@ -273,7 +273,7 @@ export default defineComponent({
   },
   methods: {
     isPlayerActing(playerView: PlayerViewModel) : boolean {
-      return playerView.players.length > 1 && playerView.waitingFor !== undefined;
+      return playerView.players.length > 1 && playerView.waitingFor !== undefined && !playerView.waitingFor.polling;
```
kberg: Note: I don't love `polling` but it's the right solution, I think. (No response required.)

src/server/routes/ApiWaitingFor.ts
kberg: ping

src/server/PlayerInput.ts
```diff
@@ -20,6 +20,11 @@ export interface PlayerInput {
...
+     */
+    polling?: boolean;
```
kberg: Let's rename this `continuePolling`.

src/server/Draft.ts
```diff
@@ -110,27 +114,53 @@ export abstract class Draft {
...
+    }
+
+    const messageTitle = repick ?
+      'You can change your pick until all players have chosen a card. Passing to ${0}' :
+      (cardsToKeep === 1 ?
+        'Select a card to keep and pass the rest to ${0}' :
+        'Select two cards to keep and pass the rest to ${0}');
+    const selectCard = new SelectCard(
+      message(messageTitle, (b) => b.player(giveTo)),
+      repick ? 'Repick' : 'Keep',
```
kberg: Image

There's already two terms for the same thing: Select and Keep. Let's not add a third. I think we should settle on one of these. I suggest "select".

src/server/Draft.ts
```diff
@@ -110,27 +114,53 @@ export abstract class Draft {
...
+      // one of the cards we previously picked plus a new card.
+      if (cardsToKeep === 1) {
+        enabled = cardsToConsider.map((_, idx) => idx < player.draftHand.length);
+      }
+    } else {
+      cardsToConsider = player.draftHand;
+    }
+
+    const messageTitle = repick ?
+      'You can change your pick until all players have chosen a card. Passing to ${0}' :
```
kberg: See below. In which case 'pick' becomes 'selection'

kberg (commented):
Tests pass now. This is a nice feature for sure.

Now next is tests. Tell me your thoughts. `drafting.spec.ts` is your friend here. If it's too complicated let me know. Before that existed, drafting broke all the time, so it has the benefit of being there, but might cause some issues.

src/client/components/PlayerHome.vue
hd256: i also no longer like `polling` since it encapsulates a lot more behavior changes than just polling. wdyt about `isInputOptional` or something along those lines?

kberg: Looking ahead, I think continuePolling works.

kberg: Actually how about just calling it 'optional'?

src/server/routes/ApiWaitingFor.ts
hd256: Renamed those fns

src/server/Draft.ts
hd256: done

src/server/Draft.ts
hd256: Done.

src/client/components/PlayerHome.vue
hd256: Done. changed to `optional`

hd256:
Added tests and fixed some of the naming. @kberg PTAL when you can and lmk if there's any other issues?

kberg (commented):
Code looks good! I'm going to read the tests carefully.

tests/drafting.spec.ts
```diff
@@ -48,7 +48,7 @@ describe('drafting', () => {
       CardName.HACKERS]);
 
     selectCard(player, CardName.BIOFERTILIZER_FACILITY);
-    cast(player.getWaitingFor(), undefined);
+    expect(cast(player.getWaitingFor(), SelectCard).buttonLabel).to.equal('Select');
```
kberg: These tests need to show that the selection option is actually the optional one for reselecting, because otherwise it could be a false-positive and easily be mistaken for selecting a new draft card.

tests/drafting.spec.ts
```diff
@@ -650,6 +747,96 @@ describe('drafting', () => {
     });
   });
 
+  it('2 player - initial draft - repick card', () => {
```
kberg: Tell me: why do you think having two tests (one for standard draft, and one for initial draft), are both necessary?

hd256: good point. given that the changes are not specific to a type of draft, one test is probably sufficient. removed one.

tests/drafting.spec.ts
hd256: updated to test for the input message as well to ensure it's the re-select prompt.

kberg (commented):
Image

Yow your giant merge really made it hard to finish the review.

hd256:
> Image
> Yow your giant merge really made it hard to finish the review.

gah, sorry, let me try to undo

kberg (changes requested):
Yeah please undo your merge; that's my responsibility.

hd256:
ok done, ptal

tests/drafting.spec.ts
```diff
@@ -1055,6 +1154,18 @@ describe('drafting', () => {
   }
 });
 
+// Asserts that the player is offered to reselect their draft card
```
kberg, line 1158: I rewrote this just a little. Let's see.

kberg:
I'll fix this and merge. It's good to go.

hd256:
> I'll fix this and merge. It's good to go.

thanks for the guidance!

kberg (approved):
Good work and thank you for your patience and effort.

kberg:
Fixes #4261

PR #8158, opened 2026-05-19
feat: terraformed css

ssimeonoff:
Check the formatting as well as well as had to locally create .prettierrc and at least set `{
  "singleQuote": true
}` 

Untitled

### DIFF
--- a/src/client/components/PlayerHome.vue
+++ b/src/client/components/PlayerHome.vue
@@ -1,31 +1,45 @@
 <template>
-  <div id="player-home" :class="game.turmoil ? 'with-turmoil' : ''">
+  <div id="player-home" :class="(game.turmoil ? 'with-turmoil': '')">
     <top-bar :playerView="playerView" />
 
     <div v-if="game.phase === 'end'">
       <div class="player_home_block">
-        <DynamicTitle title="This game is over!" :color="thisPlayer.color" />
-        <a :href="'the-end?id=' + playerView.id" v-i18n>Go to game results</a>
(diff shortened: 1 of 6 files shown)


kberg (commented):
This is alot of reformatting. Can you undo the reformatting and stick to the functional change?

This project doesn't use prettier, if that matters.

kberg:
I did a reformat and lint check. Hopefully this works.

kberg:
Where is the actual terraforming announcement?

kberg (approved):

kberg:
Thank you!

hd256:
Sorry, why was the banner removed as part of this change?

kberg:
The complaint was that it was too present, didn't scroll away. (That might not be 100% correct, but you can see some of the comments on the discord.)

hd256:
I think this indicator is far too subtle. There is nothing to indicate that this gradient means that Mars is terraformed. While experienced site users might quickly identify it, for new players, it'd be confusing and right now, there's no strong signal that end game is triggered. The only real indication is in the logs which can be easily missed if it's mixed with a flurry of other log messages.

From looking at the discord, this change is praised so maybe I'm in the minority here, but I think the endgame being triggered should be easily visible and understandable. This change only accomplishes that if you already know about it.

kberg:
I actually do agree with you. Also I apologize for letting the feature went
away without telling you. Let's get some version of a better visual feature
back (hopefully yours) but also let's address the concerns on Discord.

kberg:
I've restored the dropdown with a key addition: it only animates once per
game (per browser.) I hope that will eliminate the majority of issues.

>

PR #8178, opened 2026-05-28
Change desc on Hydrogen Processing Plant to clarify that it won't decrease max-out oxygen

hd256:
This isn't obvious. Even if this is well known behavior, it'd still be helpful to those who might not know.

### DIFF
--- a/src/server/cards/pathfinders/HydrogenProcessingPlant.ts
+++ b/src/server/cards/pathfinders/HydrogenProcessingPlant.ts
@@ -26,7 +26,7 @@ export class HydrogenProcessingPlant extends Card implements IProjectCard {
           b.production((pb) => pb.energy(1)).slash().oceans(2).br;
           b.minus().oxygen(1).asterix().br;
         }),
-        description: 'Oxygen level must be 3% or higher. Decrease oxygen level 1% (unless it\'s already maxed out). ' +
+        description: 'Oxygen level must be 3% or higher. Decrease oxygen level 1% ' +
           'Raise your energy production 1 step for every two ocean tiles on Mars.',
       },
     });
--- a/src/server/cards/pathfinders/HydrogenProcessingPlant.ts
+++ b/src/server/cards/pathfinders/HydrogenProcessingPlant.ts
@@ -24,9 +24,9 @@ export class HydrogenProcessingPlant extends Card implements IProjectCard {
         cardNumber: 'Pf19',
         renderData: CardRenderer.builder((b) => {
           b.production((pb) => pb.energy(1)).slash().oceans(2).br;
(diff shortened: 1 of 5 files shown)


kberg:
This is a fair complaint. However what I recommend is to use the warnings feature instead. Search for 'maxoceans' and you'll see how to add a other warning type (which here can be called, say 'maxoxygen-reduce'?)

That pattern can then be applied to other cards that lower global parameters.

Also note that changing any strings like this breaks any translation work already done, which would have been an issue only if we continued down the current path.

Let me know if you want some help making it happen.

hd256:
The warning already exists as part of Executor although it's not particularly clear in this case (it says "Note: the oxygen level is already at its goal."). Unfortunately, at that point it may already be too late. For example, you may buy the card during research phase only to find it doesn't work when trying to play it, but had you known that, you may not have bought it in the first place.

Given that this appears to be a common pattern, maybe it is just a rules thing which our playgroup didn't know because this effect of reducing a global param is not common enough whereas typically `max*` warnings apply when trying to increase a param beyond its max value which comes up a lot.

I don't feel too strongly about this and understand the translation implications so we can close this.

kberg:
> The warning already exists as part of Executor although it's not particularly clear in this case (it says "Note: the oxygen level is already at its goal.").

I'm suggesting adding a different warning.

Changing

```
      if (g.oxygen !== undefined && game.getOxygenLevel() >= MAX_OXYGEN_LEVEL) {
        card.warnings.add('maxoxygen');
      }
```

to

```
      if (g.oxygen !== undefined && game.getOxygenLevel() >= MAX_OXYGEN_LEVEL) {
        if (g.oxygen < 0) {
          card.warnings.add('maxoxygen-reducing');
        } else {
          card.warnings.add('maxoxygen');
        } else {
      }
```


You are right about buying the card. How about adding something to CardHelpText too?

hd256:
Gotcha. Made those changes, PTAL. thanks!

src/server/cards/pathfinders/HydrogenProcessingPlant.ts
```diff
@@ -24,7 +24,7 @@ export class HydrogenProcessingPlant extends Card implements IProjectCard {
         cardNumber: 'Pf19',
         renderData: CardRenderer.builder((b) => {
           b.production((pb) => pb.energy(1)).slash().oceans(2).br;
-          b.minus().oxygen(1).br;
+          b.minus().oxygen(1).asterix().br;
```
kberg, line 27: I'm going to approve this, but I expect someone will say that the asterisk is not necessary. I would agree with that.

I think you know this, but, the rule is laid out on Turmoil.

Image

kberg (approved):

PR #8199, opened 2026-06-06
Better and translatable agendas on cards list page

alrusdi:
better_agendas_look


kberg (approved):
Sometimes I intentionally do a mediocre job because I hope someone will come along and make it better. Thank you! 

Also, thanks for using the new style of component. I think perhaps we should give the styles a name that matches the component. No change now but if you have a thought about it, tell me.

PR #8200, opened 2026-06-07
Document title i18n

alrusdi:
Besides improving i18n support for page titles, I fixed a typo in an import and addressed your earlier feedback on the CSS class names in PR #8199. If you're not happy with the current naming, I can adjust it.


kberg (approved):
Nice! Thank you.

PR #8001, opened 2026-03-30
Add core MarsBot manager, stock overrides, and dependency stubs

ignac8:
## What this PR adds

The central MarsBot manager class and its supporting infrastructure. This is the main object that orchestrates an automa game - it owns the board, decks, turn resolver, and scoring.

- `src/server/automa/MarsBot.ts` - the MarsBot manager: manages game state, action/bonus decks, per-generation turns, VP tracking, serialization. Exposes `getCorpContext()` for corporation effects and `toModel()` for the client.
- `src/server/automa/MarsBotStock.ts` - overrides the player's `Stock` so resource queries (e.g. Hired Raiders "steal from opponent") report MarsBot's MC supply instead of zero. Uses `Object.defineProperty` getters on the resource fields.
- `src/server/automa/MarsBotTags.ts` - overrides tag counting so MarsBot's tags come from track positions rather than played cards.
- `src/common/automa/MarsBotCorpTypes.ts` - shared types for corporation cards: `IMarsBotCorp`, `MarsBotTrackCube`, `MarsBotCorpContext`, draft priority, per-gen effects.
- `src/server/SerializedGame.ts` - adds MarsBot serialization fields to the game save format.

Several files here are stubs that compile but don't do anything yet - they exist so MarsBot.ts can import them without circular dependency issues. Real implementations come in later PRs:

| Stub | Implemented in |
|------|---------------|
| MarsBotTilePlacer | PR 3 |
| MarsBotTurnResolver | PR 3 |
| MarsBotBonusCard/Deck/Resolver | PR 4 |
| MarsBotScoring | PR 5 |
| MarsBotCorpRegistry/Resolver | PR 7 |

PR #8203, opened 2026-06-09
Add MarsBot bonus card type and base card set

ignac8:
First of the one file at a time PRs discussed in #8001. Builds on the types from #7994

Adds the bonus card data type and the base game set (B01-B08). Each generation one bonus card gets shuffled into MarsBot's 4 card action deck and resolved when drawn. The deck handling itself comes in the next PR

Kept these as plain data rather than ICard since bonus cards are never played or held like project cards, they only carry an effect that MarsBot resolves and discards. Can rework that if you'd rather see them tied into ICard

Heads up that I'm out for the next week so I won't be able to address feedback until I'm back


src/server/automa/MarsBotBonusCard.ts
```diff
@@ -0,0 +1,26 @@
+import {BonusCardId} from '../../common/automa/AutomaTypes';
+
+export type MarsBotBonusCard = {
+  readonly id: BonusCardId;
+  readonly name: string;
+  /** If true, this card has been destroyed and cannot re-enter the game. */
+  destroyed: boolean;
```
kberg: If you make destroyed optional, then you can accomplish the same without the necessary additional attribute.

tests/automa/MarsBotBonusCard.spec.ts
kberg, line 1: I recommend removing this entire test file. This doesn't seem worth its weight.

src/server/automa/MarsBotBonusCard.ts
kberg: If the card is destroyed, why is it still in whatever data structure might apply? Wouldn't the card just be removed from active decks?

kberg (changes requested):

src/server/automa/MarsBotBonusCard.ts
ignac8: Went further and dropped the field entirely instead of making it optional, since nothing in this PR uses it. The type is just id and name now

tests/automa/MarsBotBonusCard.spec.ts
ignac8: Done, removed it

src/server/automa/MarsBotBonusCard.ts
ignac8: Right, it isn't kept anywhere. A bonus card only gets destroyed mid resolution while it's still in MarsBot's action deck, never in a deck pile, so destroying it just means not discarding it afterwards. The old serialize and count code that read the flag was scanning piles a destroyed card can never be in, so it always came back empty anyway. The destroy step moves to the deck/resolver PRs as just not discarding the card

src/server/automa/MarsBotBonusCard.ts
```diff
@@ -0,0 +1,24 @@
+import {BonusCardId} from '../../common/automa/AutomaTypes';
+
+export type MarsBotBonusCard = {
+  readonly id: BonusCardId;
+  readonly name: string;
+};
+
+export function bonusCard(id: BonusCardId, name: string): MarsBotBonusCard {
```
kberg, line 8: This still feels kinda light, but OK.

kberg (approved):

PR #8207, opened 2026-06-09
Add MarsBot corporation framework types

ignac8:
Another small one for the MarsBot series (#8001)

This one is types only, the contract for MarsBot corporations: what a corp defines (starting tags, track cubes, draft priority, effects, per generation actions) and the context object its effects run against. The implementations and the registry come in later PRs once the MarsBot core is in

Since this shapes the whole corp framework I'd rather get feedback on it early than after a dozen corps are built on top of it. No tests here since there's no logic, just two tiny helpers that get covered with their consumers


src/server/automa/MarsBotCorpTypes.ts
```diff
@@ -0,0 +1,96 @@
...
+};
+
+export function trackCubeKey(trackIndex: number, position: number): string {
+  return `${trackIndex}:${position}`;
+}
+
+export type IMarsBotCorp = {
+  readonly name: CardName;
+  readonly description: string;
+  readonly startingTags: ReadonlyArray<Tag>;
```
kberg: Why startingTags and not tags?

src/server/automa/MarsBotCorpTypes.ts
```diff
@@ -0,0 +1,96 @@
...
+  readonly draftPriority?: MarsBotDraftPriority;
+  readonly setup?: { resolve(ctx: MarsBotCorpContext): void };
```
kberg: Why not `setup(ctx: MarsBotCorpContext): void;`? Is this obvious later? If so feel free to point me to the PR and I'll look.

src/server/automa/MarsBotCorpTypes.ts
```diff
@@ -0,0 +1,96 @@
...
+  onColonyPlaced?(ctx: MarsBotCorpContext): void;
+  vpBonus?(ctx: MarsBotCorpContext): number;
+};
+
+export type MarsBotCorpPerGen = {
+  timing: 'roundStart' | 'beforeActionPhase';
+  resolve(ctx: MarsBotCorpContext): void;
+};
+
+export type MarsBotCorpCardRef = {
```
kberg: What's the difference between a CardRef and a ICard?

src/server/automa/MarsBotCorpTypes.ts
kberg, line 1: This class is going to need some comments. I'm going to ask you to please assemble the comments yourself so as to avoid any kind of AI slop, or at least give it a thorough cleaning before updating with comments.

src/server/automa/MarsBotCorpTypes.ts
```diff
@@ -0,0 +1,96 @@
+import {Tag} from '../../common/cards/Tag';
+import {CardName} from '../../common/cards/CardName';
+import {BonusCardId, CubeType} from '../../common/automa/AutomaTypes';
+export {CubeType};
```
kberg: Not necessary.

src/server/automa/MarsBotCorpTypes.ts
```diff
@@ -0,0 +1,96 @@
...
+  readonly cost: number;
+  readonly hasRequirements: boolean;
+  readonly victoryPoints: number;
+};
+
+export function toCorpCardRef(name: CardName, tags: ReadonlyArray<Tag>, cost: number, hasRequirements: boolean, victoryPoints: number): MarsBotCorpCardRef {
+  return {name, tags, cost, hasRequirements, victoryPoints};
+}
+
+export type MarsBotCorpContext = {
```
kberg: By and large, we do this with IPlayer. IPlayer contains not only the information specific to the player (in this case it would include stuff like `trackPosition`) but also a reference to `game`, which is manually installed and removed (this keeps memory from overflowing.) So, can this match that pattern?

kberg (changes requested):
Got a few questions on this one. I'll keep reading in the meantime.

src/server/automa/MarsBotCorpTypes.ts
ignac8: No good reason, they're the tags printed on the corp card. Renamed to tags.

src/server/automa/MarsBotCorpTypes.ts
ignac8: It copied perGeneration's {timing, resolve} shape, but setup has no timing, so the wrapper bought nothing. Flattened to a plain method. Nothing in the later PRs needs the object form either.

src/server/automa/MarsBotCorpTypes.ts
ignac8: A snapshot of the fields corp effects read, so handlers wouldn't hang on to live cards. Both call sites turn out to have a real IProjectCard in hand anyway, so it was one more shadow type. Gone, the hooks take IProjectCard now.

src/server/automa/MarsBotCorpTypes.ts
ignac8: Done. I went through the call sites while writing them so the hook comments say when things actually fire.

src/server/automa/MarsBotCorpTypes.ts
ignac8: Removed.

src/server/automa/MarsBotCorpTypes.ts
ignac8: Done. The context type is now IMarsBot: bot state next to game and player references, and the bot manager implements it directly instead of building a context object. Serialization already treats game the way IPlayer does, dropped on save and reattached on load.

src/server/automa/MarsBotCorpTypes.ts
```diff
@@ -0,0 +1,137 @@
...
+import {CardName} from '../../common/cards/CardName';
+import {BonusCardId, CubeType} from '../../common/automa/AutomaTypes';
+import {GlobalParameter} from '../../common/GlobalParameter';
+import {TileType} from '../../common/TileType';
+import {IGame} from '../IGame';
+import {IPlayer} from '../IPlayer';
+import {IProjectCard} from '../cards/IProjectCard';
+import {MarsBotBoard} from './MarsBotBoard';
+
+/**
```
kberg: This isn't JSDoc.

src/server/automa/MarsBotCorpTypes.ts
```diff
@@ -0,0 +1,137 @@
...
+ * Types for MarsBot corporations, the corp cards from the Automa expansion.
+ *
+ * A MarsBot corp names the real corporation it stands in for and defines what
```
kberg: I'd love to understand why you think this is a useful comment. I don't, but before you remove it, sell me on it.

src/server/automa/MarsBotCorpTypes.ts
```diff
@@ -0,0 +1,137 @@
...
+
+/** How the bot picks a card in the research draft. Each corp names one priority. */
+export type MarsBotDraftPriority =
+  | { type: 'tags'; tags: Array<Tag> }
+  | { type: 'mostExpensive' }
+  | { type: 'leastAdvancedTrack' }
+  | { type: 'mostTags' };
+
+/**
+ * A cube a corp places on the bot's board during setup. It triggers the corp's
```
kberg: This is too detailed. To simplify: "A cube on the corp's track, including its type and position within a track."

src/server/automa/MarsBotCorpTypes.ts
```diff
@@ -0,0 +1,137 @@
...
+  readonly draftPriority?: MarsBotDraftPriority;
+  /** Runs once, right after the bot's corporation is chosen. */
+  setup?(bot: IMarsBot): void;
+  readonly effect?: MarsBotCorpEffect;
+  readonly perGeneration?: MarsBotCorpPerGen;
+  /** Cubes seeded onto the bot's board tracks during setup. */
+  readonly trackCubes?: ReadonlyArray<MarsBotTrackCube>;
+};
+
+/** Corp abilities that react to game events. All hooks are optional. */
```
kberg: This can be removed

src/server/automa/MarsBotCorpTypes.ts
```diff
@@ -0,0 +1,137 @@
...
+  /** The bot's track marker reached one of this corp's cubes. */
+  onTrackCubeTrigger?(bot: IMarsBot, trackIndex: number, position: number, cubeType: CubeType): void;
+  /** The bot drew and resolved a project card. */
+  onProjectCardResolved?(bot: IMarsBot, card: IProjectCard): void;
+  /** The human player played a card. */
+  onHumanCardPlayed?(bot: IMarsBot, card: IProjectCard): void;
+  /** A tile landed on the board, placed by either side. */
+  onTilePlaced?(bot: IMarsBot, placedByMarsBot: boolean, tileType: TileType): void;
+  onVenusRaised?(bot: IMarsBot): void;
+  /** Fires before a global parameter raise. Return true to skip the raise. */
```
kberg: What does "skip the raise" mean?

onVenusRaised reads like venus has already been raised. Is this the right name?

src/server/automa/MarsBotCorpTypes.ts
```diff
@@ -0,0 +1,137 @@
...
+  onGlobalParameterRaised?(bot: IMarsBot, parameter: GlobalParameter): boolean;
+  /** Fires when the bot is about to gain M€. Returns the amount that actually reaches the supply. */
+  onMcGained?(bot: IMarsBot, amount: number): number;
```
kberg: Same, `on` calls usually refer to when something has already been done. onTilePlaced means the tile has been placed, not that it is about to be placed.

src/server/automa/MarsBotCorpTypes.ts
```diff
@@ -0,0 +1,137 @@
...
+  /** Fired when MarsBot places a colony on any tile (Colonies rule C-33). */
+  onColonyPlaced?(bot: IMarsBot): void;
+  /** Extra victory points added at final scoring. */
+  vpBonus?(bot: IMarsBot): number;
+};
+
+/** A corp action taken every generation, at the given point in the round. */
+export type MarsBotCorpPerGen = {
```
kberg: I can't tell, but I suspect this whole type is overkill

src/server/automa/MarsBotCorpTypes.ts
```diff
@@ -0,0 +1,137 @@
...
+  timing: 'roundStart' | 'beforeActionPhase';
+  resolve(bot: IMarsBot): void;
+};
+
+/**
+ * The bot as corp handlers see it, implemented by the bot manager. Like
```
kberg: Why are the drop and keep details in here? Does IPlayer's top comment do that? Is that a key piece of information about IMarsBot?

src/server/automa/MarsBotCorpTypes.ts
```diff
@@ -0,0 +1,137 @@
...
+  removeBonusCard(bonusCardId: BonusCardId): void;
+  /** Discards the action-deck card with the fewest tags. */
+  discardFewestTagsFromActionDeck(): void;
+
+  raiseTemperature(steps: 1 | 2 | 3): void;
+  placeOcean(): void;
+  placeCity(): void;
+  placeGreenery(): void;
+  /** Places a colony on a randomly selected eligible tile (Colonies rule C-15b). Returns true if placed. */
+  placeRandomColony(): boolean;
```
kberg: If it's optional, then call it `maybePlaceRandomColony`

src/server/automa/MarsBotCorpTypes.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  /** Draws project cards into the bot's action deck. */
+  drawProjectCardsToActionDeck(count: number): void;
+  /** Puts a bonus card in the action deck, pulling it from the bonus deck when present. */
+  addBonusCardToActionDeck(bonusCardId: BonusCardId): void;
+  addBonusCardToBonusDeck(bonusCardId: BonusCardId): void;
+  /** Removes a bonus card from the bonus deck and from the action deck. */
+  removeBonusCard(bonusCardId: BonusCardId): void;
+  /** Discards the action-deck card with the fewest tags. */
+  discardFewestTagsFromActionDeck(): void;
```
kberg: This name reads like it's discarding tags, not cards.

kberg (changes requested):
Thanks. I understand things better.

src/server/automa/MarsBotCorpTypes.ts
ignac8: Gone, along with the rest of the banner.

src/server/automa/MarsBotCorpTypes.ts
ignac8: No sale. It summarized the declarations below it. Removed.

src/server/automa/MarsBotCorpTypes.ts
ignac8: Took your wording.

src/server/automa/MarsBotCorpTypes.ts
ignac8: Removed.

src/server/automa/MarsBotCorpTypes.ts
ignac8: onVenusRaised does fire after the raise, the game hook calls it post-increase, so that name holds. The other one didn't: it fires before and can cancel, so it's now interceptGlobalParameterRaise and the comment says the parameter doesn't move when it returns true.

src/server/automa/MarsBotCorpTypes.ts
ignac8: Kept the name and fixed the moment instead: it now fires after the M€ lands and returns nothing. The only corp using it just counts what was earned, so the filter-style return had no users.

src/server/automa/MarsBotCorpTypes.ts
ignac8: It was. Replaced with two optional methods on the corp, roundStart and beforeActionPhase. They match the two boxes on the physical cards.

src/server/automa/MarsBotCorpTypes.ts
ignac8: Trimmed to the one line.

src/server/automa/MarsBotCorpTypes.ts
ignac8: Renamed to maybePlaceRandomColony.

src/server/automa/MarsBotCorpTypes.ts
ignac8: Renamed to discardCardWithFewestTags.

src/server/automa/MarsBotCorpTypes.ts
```diff
@@ -0,0 +1,116 @@
...
+  readonly effect?: MarsBotCorpEffect;
+  /** A corp action taken at the start of every generation. */
+  roundStart?(bot: IMarsBot): void;
+  /** A corp action taken every generation, right before the action phase. */
+  beforeActionPhase?(bot: IMarsBot): void;
+  /** Cubes seeded onto the bot's board tracks during setup. */
+  readonly trackCubes?: ReadonlyArray<MarsBotTrackCube>;
+};
+
+export type MarsBotCorpEffect = {
```
kberg, line 49: nit: this is Effects. What do you think? Is it expected that bots only have a single effect?

I kinda like how effects are separate in this way. ICard doesn't do that. 🤔

src/server/automa/MarsBotCorpTypes.ts
```diff
@@ -0,0 +1,116 @@
...
+  /** The bot's track marker reached one of this corp's cubes. */
+  onTrackCubeTrigger?(bot: IMarsBot, trackIndex: number, position: number, cubeType: CubeType): void;
+  /** The bot drew and resolved a project card. */
+  onProjectCardResolved?(bot: IMarsBot, card: IProjectCard): void;
+  /** The human player played a card. */
+  onHumanCardPlayed?(bot: IMarsBot, card: IProjectCard): void;
+  /** A tile landed on the board, placed by either side. */
+  onTilePlaced?(bot: IMarsBot, placedByMarsBot: boolean, tileType: TileType): void;
+  onVenusRaised?(bot: IMarsBot): void;
```
kberg, line 58: As long as this  is "called after venus is raised any number of steps" I'm good. :D

src/server/automa/MarsBotCorpTypes.ts
```diff
@@ -0,0 +1,116 @@
...
+  /** Called before the bot raises a global parameter. Return true to consume the raise: the parameter does not move. */
```
kberg, line 59: What does "the parameter does not move" mean?

src/server/automa/MarsBotCorpTypes.ts
```diff
@@ -0,0 +1,116 @@
...
+export interface IMarsBot {
+  readonly game: IGame;
+  /** The neutral player that holds the bot's terraform rating and owns its tiles. */
+  readonly player: IPlayer;
+  readonly board: MarsBotBoard;
+
+  /** The bot's M€ pool. */
+  mcSupply: number;
+  /** Floaters stored by Venus corps. */
+  readonly floaterCount: number;
```
kberg, line 79: s/floaterCount/floaters

src/server/automa/MarsBotCorpTypes.ts
```diff
@@ -0,0 +1,116 @@
...
+
+/** The bot as corp handlers see it, implemented by the bot manager. */
+export interface IMarsBot {
+  readonly game: IGame;
+  /** The neutral player that holds the bot's terraform rating and owns its tiles. */
+  readonly player: IPlayer;
+  readonly board: MarsBotBoard;
+
+  /** The bot's M€ pool. */
+  mcSupply: number;
```
kberg, line 77: s/mcSupply/megacredits?

src/server/automa/MarsBotCorpTypes.ts
```diff
@@ -0,0 +1,116 @@
...
+  addFloaters(count: number): void;
+  /** Removes floaters, stopping at zero. */
+  spendFloaters(count: number): void;
+
+  gainMc(amount: number): void;
+  /** Positive steps raise the bot's TR, negative steps lower it. */
+  raiseTR(steps: number): void;
+  advanceTrack(trackIndex: number): void;
+
+  /** Draws the top project card and resolves it as a bot action. False when the deck is empty. */
```
kberg, line 89: What happens if the deck is empty?

src/server/automa/MarsBotCorpTypes.ts
```diff
@@ -0,0 +1,116 @@
...
+  discardCardWithFewestTags(): void;
+
+  raiseTemperature(steps: 1 | 2 | 3): void;
+  placeOcean(): void;
+  placeCity(): void;
+  placeGreenery(): void;
+  /** Places a colony on a randomly selected eligible tile (Colonies rule C-15b). Returns true if placed. */
+  maybePlaceRandomColony(): boolean;
+
+  /** Corp-specific counters, serialized with the bot. Missing keys read as 0. */
```
kberg, line 113: I'm gonna have some questions about this as I see the implementations.

kberg (approved):
I still have these pending questions, but let's get this in and move to the next one. Can you respond to them and/or supply fixes later?a

src/server/automa/MarsBotCorpTypes.ts
ignac8: The corp card has a single effect box, and the handlers inside are the triggers that box can fire on, which is why it reads singular. Say the word if you want it plural.

src/server/automa/MarsBotCorpTypes.ts
ignac8: It is. Wrote that on the method in #8353.

src/server/automa/MarsBotCorpTypes.ts
ignac8: Badly worded on my part. It means the raise is cancelled: the global parameter stays at its current value and the corp takes its own bonus instead. Reworded in #8353.

src/server/automa/MarsBotCorpTypes.ts
ignac8: Renamed in #8353.

src/server/automa/MarsBotCorpTypes.ts
ignac8: Renamed in #8353, and MarsBotModel's field with it since PlayerModel already spells it megacredits.

src/server/automa/MarsBotCorpTypes.ts
ignac8: The deck reshuffles its discard pile when the draw pile runs out, so false only comes back when both piles are empty, and the bot skips that action. Documented on the method in #8353.

src/server/automa/MarsBotCorpTypes.ts
ignac8: Fair. They hold the per-corp counters, like Pristar's cube or Mining Guild's countdown. The manifests that use them are still queued behind these PRs.

PR #8208, opened 2026-06-09
Add MarsBot tile placement logic

ignac8:
Next one for the MarsBot series (#8001), independent of the other open ones

Adds the tile placement rules from the automa rulebook: greeneries go next to MarsBot cities and away from human ones, cities next to existing greenery, oceans on any reserved spot, with the ocean adjacency and placement bonus tiebreakers on top. Works against any two IPlayers so the tests just use a regular two player game


src/server/automa/MarsBotTilePlacer.ts
```diff
@@ -0,0 +1,201 @@
...
+ * Ocean: Any ocean-reserved space.
+ *
+ * Tiebreakers (after type-specific rules):
+ *   1. Adjacent to as many oceans as possible.
+ *   2. Cover the most placement bonus icons.
+ *   3. Random (flip a project card, use cost to count through tied spaces).
+ */
+export class MarsBotTilePlacer {
+  constructor(
+    private readonly game: IGame,
```
kberg, line 21: I'm curious if it's possible to infer the marsbot and human player from game.players. Later.

src/server/automa/MarsBotTilePlacer.ts
```diff
@@ -0,0 +1,201 @@
...
+      return undefined;
+    }
+    // Oceans have no type-specific primary sort, go straight to tiebreakers.
+    return this.selectBestSpace(spaces, () => 0);
+  }
+
+  /** Find a space for the Neural Instance tile: not adjacent to any tiles, not on edge, not on/adjacent to reserved spaces. */
+  public findNeuralInstanceSpace(): Space | undefined {
+    const board = this.game.board;
+    const isReserved = (s: Space) => s.spaceType === SpaceType.OCEAN || s.spaceType === SpaceType.RESTRICTED;
```
kberg: You can inline this, right?

src/server/automa/MarsBotTilePlacer.ts
```diff
@@ -0,0 +1,201 @@
...
+      return tied[0].space;
+    }
+
+    // Tiebreaker 3: random using project deck card cost
+    return this.randomTiebreak(tied.map((t) => t.space));
+  }
+
+  /** Break a tie by flipping a project card and counting through spaces. */
+  private randomTiebreak(spaces: ReadonlyArray<Space>): Space {
+    const card = this.game.projectDeck.draw(this.game);
```
kberg: use `drawOrThrow`

src/server/automa/MarsBotTilePlacer.ts
```diff
@@ -0,0 +1,201 @@
...
+    const cost = card?.cost ?? 0;
+    if (card) {
+      this.game.projectDeck.discardPile.push(card);
+    }
+    const index = cost % spaces.length;
+    return spaces[index];
+  }
+
+  /** Calculate MC MarsBot gains from placement bonuses (1 MC per icon covered). */
+  public getPlacementBonusMC(space: Space): number {
```
kberg, line 167: Is this a useful method? Perhaps, if it's invoked in multiple places.

(I realize a fuller PR would help me see that, still, thank you for breaking this up.)

src/server/automa/MarsBotTilePlacer.ts
```diff
@@ -0,0 +1,201 @@
...
+    return space.bonus.length; // 1 MC per icon
+  }
+
+  /** Calculate MC MarsBot gains from ocean adjacency (2 MC per adjacent ocean). */
+  public getOceanAdjacencyMC(space: Space): number {
+    const adj = this.game.board.getAdjacentSpaces(space);
+    const oceans = adj.filter((s) => Board.isOceanSpace(s) && s.tile !== undefined).length;
```
kberg: You don't care that this is not compatible with Ares, right? I don't mind, but let's make sure then that you can't create a game with bots and Ares.

tests/automa/MarsBotTilePlacer.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+      // Place two greeneries adjacent to center
+      let placed = 0;
+      for (const s of adj) {
+        if (s.tile === undefined && s.spaceType === SpaceType.LAND && placed < 2) {
+          game.simpleAddTile(marsBot, s, {tileType: TileType.GREENERY});
+          placed++;
+        }
+      }
+
+      if (placed >= 2) {
```
kberg: This test can be falsely positive. If placed is not 2, there's no assertions. Can you check the other tests accordingly, please? In this case I'd instead assert an expectation that placed is 2. Or if that's too unreliable that it's 2 or greater.

tests/automa/MarsBotTilePlacer.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    it('returns 0 for spaces not adjacent to oceans', () => {
+      // Find a space with no adjacent oceans
+      const space = game.board.spaces.find((s) =>
+        s.tile === undefined &&
+        s.spaceType === SpaceType.LAND &&
+        game.board.getAdjacentSpaces(s).every((adj) => adj.tile?.tileType !== TileType.OCEAN),
+      );
+      if (space) {
+        expect(tilePlacer.getOceanAdjacencyMC(space)).to.eq(0);
```
kberg: See my comment below about false positives. In this case, you can just remove the conditional and use `space!`. `!` isn't allowed in production code, but is allowed in tests.

kberg (changes requested):
The tests are not great. Interesting miss that the tests can run without asserting anything.

src/server/automa/MarsBotTilePlacer.ts
```diff
@@ -0,0 +1,201 @@
...
+  public findGreenerySpace(): Space | undefined {
+    const board = this.game.board;
+    const spaces = board.getAvailableSpacesForGreenery(this.marsBot);
+    if (spaces.length === 0) {
+      return undefined;
+    }
+
+    return this.selectBestSpace(spaces, (space) => {
+      const adj = board.getAdjacentSpaces(space);
+      const marsBotCities = adj.filter((s) => s.player === this.marsBot && Board.isCitySpace(s)).length;
```
kberg: Almost everywhere just use Board.ownedBy(this.marsBot) or Board.ownedBy(this.humanPlayer).

src/server/automa/MarsBotTilePlacer.ts
```diff
@@ -0,0 +1,201 @@
...
+    if (spaces.length === 1) {
+      return spaces[0];
+    }
+
+    const board = this.game.board;
+
+    const scored = spaces.map((space) => ({
+      space,
+      primary: primaryScore(space),
+      adjacentOceans: board.getAdjacentSpaces(space).filter((s) => Board.isOceanSpace(s) && s.tile !== undefined).length,
```
kberg: Remove s.tile !== undefined beause if it's an ocean space it's alrady got a tile. `isOceanSpace` misrepresents what it actually does, which is, that it has a tile that can represent an ocean. It's leftover from early coding and I haven't done the churn to rename it.

src/server/automa/MarsBotTilePlacer.ts
kberg: Simplify to 

```
const cities = adj.filter(Board.isCitySpace);
const botCities = cities.filter(Board.ownedBy(this.marsBot));
const humanCities = cities.filter(Board.ownedBy(this.humanPlayer));

...
```

src/server/automa/MarsBotTilePlacer.ts
```diff
@@ -0,0 +1,201 @@
...
+    primaryScore: (space: Space) => number,
+  ): Space | undefined {
+    if (spaces.length === 0) {
+      return undefined;
+    }
+    if (spaces.length === 1) {
+      return spaces[0];
+    }
+
+    const board = this.game.board;
```
kberg: No need for this local variable

src/server/automa/MarsBotTilePlacer.ts
```diff
@@ -0,0 +1,201 @@
...
+      return adj.filter((s) => Board.isGreenerySpace(s)).length;
+    });
+  }
+
+  /** Find the best space for MarsBot to place a city adjacent to ≥2 greenery/ocean tiles (B05). */
+  public findExpediteConstructionCitySpace(): Space | undefined {
+    const board = this.game.board;
+    const spaces = board.getAvailableSpacesForCity(this.marsBot);
+    const eligible = spaces.filter((space) => {
+      const adj = board.getAdjacentSpaces(space);
```
kberg: This algorithm is repeated within findExpediteConstructionCitySpace twice.

tests/automa/MarsBotTilePlacer.spec.ts
kberg, line 1: Do you ever test tiebreakers?

src/server/automa/MarsBotTilePlacer.ts
ignac8: Took your version.

src/server/automa/MarsBotTilePlacer.ts
ignac8: Removed, here and in getOceanAdjacencyMC. There's a test now for an ocean-reserved space with no tile on it paying 0, so the meaning of isOceanSpace is pinned if anyone is tempted to re-add the check.

src/server/automa/MarsBotTilePlacer.ts
ignac8: Dropped.

src/server/automa/MarsBotTilePlacer.ts
ignac8: One scorer now, used for both the filter and the sort.

src/server/automa/MarsBotTilePlacer.ts
ignac8: Inlined. The second use, `isReserved(space)` on the candidate itself, turned out to be dead: getAvailableSpacesOnLand only returns LAND spaces, so a candidate can never be ocean-reserved or restricted. With that gone there is one use left and it reads fine inline.

src/server/automa/MarsBotTilePlacer.ts
ignac8: Done, and the card goes through `projectDeck.discard` now instead of being pushed onto discardPile.

src/server/automa/MarsBotTilePlacer.ts
ignac8: Nine call sites once the rest of the bot lands: every tile the turn resolver and the bonus card resolver place pays placement M€ this way. Two of them want the halves separately, which is why those stayed public.

src/server/automa/MarsBotTilePlacer.ts
ignac8: Right, it isn't Ares-aware. Bot games can't turn Ares on: AutomaGameSetup.sanitizeGameOptions force-disables it, along with the Moon, Pathfinders, CEOs, Star Wars, underworld, community cards and the solar phase. That file comes in a later PR.

src/server/automa/MarsBotTilePlacer.ts
ignac8: Left for later as you suggested. The bot's player hangs off the automa hooks, so it should be small.

tests/automa/MarsBotTilePlacer.spec.ts
ignac8: Caught something real: six tests could run without asserting anything. Every conditional is gone. I also checked the new ones bite by breaking the implementation four ways, dropping the greenery preference, each of the two tiebreakers, and the expedite threshold. The threshold survived at first, because my "returns undefined" case used an empty board where a threshold of 1 also finds nothing, so there is now a case with exactly one greenery.

tests/automa/MarsBotTilePlacer.spec.ts
ignac8: Done.

tests/automa/MarsBotTilePlacer.spec.ts
ignac8: They weren't tested. Added the adjacent-ocean and bonus-icon tiebreaks, using findGreenerySpace on a board where the bot owns nothing so every candidate ties on the primary score, plus one showing a project card is drawn and discarded when spaces are still tied.

src/server/automa/MarsBotTilePlacer.ts
```diff
@@ -32,11 +32,11 @@ export class MarsBotTilePlacer {
     }
 
     return this.selectBestSpace(spaces, (space) => {
-      const adj = board.getAdjacentSpaces(space);
-      const marsBotCities = adj.filter((s) => s.player === this.marsBot && Board.isCitySpace(s)).length;
-      const humanCities = adj.filter((s) => s.player === this.humanPlayer && Board.isCitySpace(s)).length;
+      const cities = board.getAdjacentSpaces(space).filter(Board.isCitySpace);
```
kberg, line 35: The .length was fine, I didn't mean literally. But this is also fine.

tests/automa/MarsBotTilePlacer.spec.ts
```diff
@@ -17,130 +19,127 @@ describe('MarsBotTilePlacer', () => {
...
     });
 
-    it('returns undefined when no ocean spaces available', () => {
-      // Fill all ocean spaces
-      const oceanSpaces = game.board.getAvailableSpacesForOcean(marsBot);
-      for (const s of oceanSpaces) {
-        game.simpleAddTile(marsBot, s, {tileType: TileType.OCEAN});
+    it('returns undefined once every ocean space is taken', () => {
+      for (const space of game.board.getAvailableSpacesForOcean(marsBot)) {
+        game.simpleAddTile(marsBot, space, {tileType: TileType.OCEAN});
```
kberg: TestingUtils has maxOutOceans.

tests/automa/MarsBotTilePlacer.spec.ts
```diff
@@ -17,130 +19,127 @@ describe('MarsBotTilePlacer', () => {
...
       }
-      const space = tilePlacer.findOceanSpace();
-      expect(space).to.be.undefined;
+      expect(tilePlacer.findOceanSpace()).is.undefined;
```
kberg: Prefer cast(x, undefined) to expect(x).is.undefined.

When cast fails, it prints out a more useful message. Apply throughout.

tests/automa/MarsBotTilePlacer.spec.ts
```diff
@@ -17,130 +19,127 @@ describe('MarsBotTilePlacer', () => {
...
-        const adj = game.board.getAdjacentSpaces(citySpace);
-        const adjacentGreenery = adj.filter((s) => s.tile?.tileType === TileType.GREENERY);
-        // The chosen space should be adjacent to greenery (or at least the best available)
-        expect(adjacentGreenery.length).to.be.gte(0); // Basic assertion
-      }
+    it('prefers a space next to a greenery', () => {
+      game.simpleAddTile(human, inlandSpace(), {tileType: TileType.GREENERY});
+
+      const space = tilePlacer.findCitySpace();
+      expect(game.board.getAdjacentSpaces(space!).filter(Board.isGreenerySpace)).has.length(1);
```
kberg: Get the space id from inlandSpace and use .some(space.id === ...) repeating throughout

src/server/automa/MarsBotTilePlacer.ts
```diff
@@ -57,24 +57,16 @@ export class MarsBotTilePlacer {
   /** Find the best space for MarsBot to place a city adjacent to ≥2 greenery/ocean tiles (B05). */
   public findExpediteConstructionCitySpace(): Space | undefined {
```
kberg: What does expedite mean here?

tests/automa/MarsBotTilePlacer.spec.ts
```diff
@@ -17,130 +19,127 @@ describe('MarsBotTilePlacer', () => {
...
+
+    it('finds the space next to two greeneries', () => {
+      const [first, second] = game.board.getAdjacentSpaces(inlandSpace());
+      game.simpleAddTile(human, first, {tileType: TileType.GREENERY});
+      game.simpleAddTile(human, second, {tileType: TileType.GREENERY});
+
+      const space = tilePlacer.findExpediteConstructionCitySpace();
+      const adjacent = game.board.getAdjacentSpaces(space!)
+        .filter((s) => Board.isGreenerySpace(s) || Board.isOceanSpace(s));
+      expect(adjacent).has.length(2);
```
kberg: I think this is implied, but find the actual spaes and don't just count them.

tests/automa/MarsBotTilePlacer.spec.ts
```diff
@@ -17,130 +19,127 @@ describe('MarsBotTilePlacer', () => {
...
+
+      expect(tilePlacer.getOceanAdjacencyMC(neighbour)).to.eq(0);
+    });
+
+    it('adds both parts together', () => {
+      const ocean = game.board.getAvailableSpacesForOcean(marsBot)[0];
+      game.simpleAddTile(marsBot, ocean, {tileType: TileType.OCEAN});
+      const neighbour = game.board.getAdjacentSpaces(ocean).find((s) => game.board.canPlaceTile(s))!;
+
+      expect(tilePlacer.getTotalPlacementMC(neighbour)).to.eq(neighbour.bonus.length + 2);
```
kberg: assert the bonuses in the neighbor. Don't just randomly hope the test matches the neighbor's bonuses.

src/server/automa/MarsBotTilePlacer.ts
ignac8: Leaving it as is then.

tests/automa/MarsBotTilePlacer.spec.ts
ignac8: maxOutOceans stops at MAX_OCEAN_TILES, so it places 9 and leaves 3 of Tharsis's 12 ocean-reserved spaces empty. findOceanSpace still finds one of those. Passing 12 as toValue hangs instead: game.addOcean refuses the 10th ocean, getOceanSpaces().length stays at 9, and the loop never ends. Kept the explicit loop over getAvailableSpacesForOcean.

tests/automa/MarsBotTilePlacer.spec.ts
ignac8: Done, cast(x, undefined) at the three undefined checks.

tests/automa/MarsBotTilePlacer.spec.ts
ignac8: Done. The tests capture the space they set up and assert the result sits next to it by id.

src/server/automa/MarsBotTilePlacer.ts
ignac8: It is the Expedited Construction bonus card (B05). The method was missing the d. Renamed to findExpeditedConstructionCitySpace, and the comment names the card now.

tests/automa/MarsBotTilePlacer.spec.ts
ignac8: Done. It asserts the returned space is adjacent to both greeneries the test placed, by id.

tests/automa/MarsBotTilePlacer.spec.ts
ignac8: The neighbour it happened to pick had no bonus icons at all, so "adds both parts together" was asserting 0 + 2. The placement M€ tests now name their spaces and assert the bonuses: ocean on 04, two steel icons on 03, so the total is a literal 4.

tests/automa/MarsBotTilePlacer.spec.ts
kberg: Oh good point. Thanks.

src/server/automa/MarsBotTilePlacer.ts
kberg: As you can see I don't know the expansion. Eventually it shouldn't matter you an expect understanding of the expansion.

tests/automa/MarsBotTilePlacer.spec.ts
kberg: Oh that's funny

kberg (approved):
I made a commit with just a couple of minor churn things. Also you'll note I added a TODO to move inlandSpcae to TestingUtils, that looks super useful.

PR #8209, opened 2026-06-09
Add MarsBot draft card selection

ignac8:
Part of the MarsBot series (#8001). Builds on #8207 so that commit shows in the diff too, only MarsBotDraftResolver.ts and its test are new here

Adds how MarsBot picks cards during drafting: by corp draft priority when it has one (matching tags, most expensive, least advanced track, most tags) and random otherwise. Tests fabricate the cards and run against the Tharsis board with a seeded rng


src/server/automa/corps/MarsBotDraftResolver.ts
```diff
@@ -0,0 +1,149 @@
...
+import {Tag} from '../../../common/cards/Tag';
+import {Random} from '../../../common/utils/Random';
+import {MarsBotBoard} from '../MarsBotBoard';
+import {inplaceShuffle} from '../../utils/shuffle';
+
+/**
+ * Pick the item with the highest score, breaking ties randomly.
+ * If all scores are 0, returns undefined (caller handles fallback).
+ */
+function pickBest<T>(items: T[], scorer: (t: T) => number, rng: Random): T | undefined {
```
kberg: Array<T> is preferred over T[]

src/server/automa/corps/MarsBotDraftResolver.ts
```diff
@@ -0,0 +1,149 @@
...
+ */
+export class MarsBotDraftResolver {
+  /**
+   * Pick the best card for MarsBot from a hand based on the draft priority.
+   */
+  public static pickCardForMarsBot(
+    hand: IProjectCard[],
+    priority: MarsBotDraftPriority,
+    rng: Random,
+    board?: MarsBotBoard,
```
kberg: Could MarsBotBoard be renamed to MarsBotTracks? `board` is too easy to confuse for `game.board`.

src/server/automa/corps/MarsBotDraftResolver.ts
```diff
@@ -0,0 +1,149 @@
...
+  if (bestScore <= 0) {
+    return undefined;
+  }
+  return bestItems[rng.nextInt(bestItems.length)];
+}
+
+/**
+ * Handles MarsBot draft card selection and post-draft discard based on corp draft priority.
+ */
+export class MarsBotDraftResolver {
```
kberg, line 11: This file is a mix of sometimes using static class methods and sometimes using functions.  This reads like a code smell.

src/server/automa/corps/MarsBotDraftResolver.ts
```diff
@@ -0,0 +1,149 @@
...
+    }
+    if (hand.length === 1) {
+      return hand[0];
+    }
+
+    switch (priority.type) {
+    case 'tags':
+      return MarsBotDraftResolver.pickByTags(hand, priority.tags, rng);
+    case 'mostExpensive':
+      return pickBest(hand, (c) => c.cost, rng) ?? hand[rng.nextInt(hand.length)];
```
kberg: In what case will pickBest not return something? It's protecting a case that, if it occurred, would reveal a bigger problem.

src/server/automa/corps/MarsBotDraftResolver.ts
```diff
@@ -0,0 +1,149 @@
...
+    case 'leastAdvancedTrack':
+      return MarsBotDraftResolver.pickByLeastAdvancedTrack(hand, rng, board);
+    case 'mostTags':
+      return pickBest(hand, (c) => c.tags.filter((t) => t !== Tag.WILD).length, rng) ?? hand[rng.nextInt(hand.length)];
```
kberg: Same

src/server/automa/corps/MarsBotDraftResolver.ts
```diff
@@ -0,0 +1,149 @@
...
+    }
+  }
+
+  /**
+   * Post-draft discard: shuffle drafted cards, reveal from top.
+   * If card has NO matching priority tag → discard. If it HAS a match → stop.
+   * If all 4 checked → keep all.
+   */
+  public static postDraftDiscard(
```
kberg: What does post mean?

src/server/automa/corps/MarsBotDraftResolver.ts
```diff
@@ -0,0 +1,149 @@
...
+    drafted: IProjectCard[],
+    priority: MarsBotDraftPriority,
+    rng: Random,
+  ): {kept: IProjectCard[], discarded: IProjectCard[]} {
+    if (priority.type !== 'tags') {
```
kberg: Can you check this one out to see if it can be simpler?

kberg (changes requested):
I haven't read the test yet, or CorpTypes. I assume CorpTypes will be its own PR

tests/automa/corps/MarsBotDraftResolver.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+import {MarsBotDraftPriority} from '../../../src/server/automa/MarsBotCorpTypes';
+import {Tag} from '../../../src/common/cards/Tag';
+import {SeededRandom} from '../../../src/common/utils/Random';
+import {IProjectCard} from '../../../src/server/cards/IProjectCard';
+import {CardType} from '../../../src/common/cards/CardType';
+import {CardName} from '../../../src/common/cards/CardName';
+import {MarsBotBoard} from '../../../src/server/automa/MarsBotBoard';
+import {THARSIS_MARSBOT_BOARD} from '../../../src/server/automa/boards/TharsisMarsBot';
+
+function fakeCard(name: string, tags: Tag[], cost: number = 10): IProjectCard {
```
kberg: There's already a fakeCard mechanism. Don't make your own. Ask your AI what other constructs you're making that already exist.

tests/automa/corps/MarsBotDraftResolver.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+        fakeCard('A', [Tag.EARTH]),     // No match
+        fakeCard('B', [Tag.PLANT]),     // No match
+        fakeCard('C', [Tag.BUILDING]),  // Match
+        fakeCard('D', [Tag.EARTH]),     // After match — kept
+      ];
+
+      const result = MarsBotDraftResolver.postDraftDiscard(cards, priority, rng);
+      // After shuffle, order may vary, but logic should work:
+      // Non-matching before first match → discarded, rest → kept
+      const totalCards = result.kept.length + result.discarded.length;
```
kberg: These assertions seem reltaively weak and could b better.

tests/automa/corps/MarsBotDraftResolver.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+import {Tag} from '../../../src/common/cards/Tag';
+import {SeededRandom} from '../../../src/common/utils/Random';
+import {IProjectCard} from '../../../src/server/cards/IProjectCard';
+import {CardType} from '../../../src/common/cards/CardType';
+import {CardName} from '../../../src/common/cards/CardName';
+import {MarsBotBoard} from '../../../src/server/automa/MarsBotBoard';
+import {THARSIS_MARSBOT_BOARD} from '../../../src/server/automa/boards/TharsisMarsBot';
+
+function fakeCard(name: string, tags: Tag[], cost: number = 10): IProjectCard {
+  return {
```
kberg: Is there really a need for fakeCards. You can find plenty of cards that match the tags you need exactly. reate them once, and give them names that help understand why they're in your test (eg `buildingTagCard`)

tests/automa/corps/MarsBotDraftResolver.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+      const cards = [
+        fakeCard('A', [], 5),
+        fakeCard('B', [], 20),
+        fakeCard('C', [], 12),
+      ];
+      const picked = MarsBotDraftResolver.pickCardForMarsBot(cards, priority, rng);
+      expect(picked.name).to.eq('B' as CardName);
+    });
+
+    it('ties pick randomly', () => {
```
kberg: How does this verify that ties pick randomly?

tests/automa/corps/MarsBotDraftResolver.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+  return {
+    name: name as CardName,
+    tags,
+    cost,
+    type: CardType.ACTIVE,
+  } as unknown as IProjectCard;
+}
+
+describe('MarsBotDraftResolver', () => {
+  describe('pickCardForMarsBot - tags priority', () => {
```
kberg: If you need controlled shuffling, this is not a good way to do it because it hides what you want to know.

My suggestion is instead of injecting rng to inject a shuffle function. The real implementation will include the random shuffler, but you can then either set the order you like or just have it work like a no-op.

kberg (changes requested):

src/server/automa/corps/MarsBotDraftResolver.ts
ignac8: Done.

src/server/automa/corps/MarsBotDraftResolver.ts
ignac8: Renamed, in #8353, since MarsBotBoard is already merged. This PR picks up the new name once that lands.

src/server/automa/corps/MarsBotDraftResolver.ts
ignac8: Fair. It's a class with state now: the tracks and a shuffle go in the constructor, so no method carries them as parameters.

src/server/automa/corps/MarsBotDraftResolver.ts
ignac8: Whenever no card in the hand carries a priority tag, which is a legal hand, and the rule is then to pick at random. You're right that the guard hid that. pickBest is total now, treating a hand where nothing scores as one tie across every card, so the three `?? hand[rng.nextInt(...)]` fallbacks are gone.

src/server/automa/corps/MarsBotDraftResolver.ts
ignac8: Same answer as above: it covered the hand where nothing matches. pickBest returns a card in every case now.

src/server/automa/corps/MarsBotDraftResolver.ts
ignac8: Nothing useful. It's discardAfterDraft now.

src/server/automa/corps/MarsBotDraftResolver.ts
ignac8: Shuffle, findIndex the first card with a priority tag, discard what came before it, keep the rest. The foundMatch loop and its trailing special case are gone.

tests/automa/corps/MarsBotDraftResolver.spec.ts
ignac8: Mine is gone.

tests/automa/corps/MarsBotDraftResolver.spec.ts
ignac8: It asserts the exact kept and discarded lists now instead of counting them.

tests/automa/corps/MarsBotDraftResolver.spec.ts
ignac8: Done, real cards named for their role: buildingCard, spaceCard, buildingAndSpaceCard. fakeCard is left in two places, for wild tags, since no project card carries one (Research Coordination is a prelude).

tests/automa/corps/MarsBotDraftResolver.spec.ts
ignac8: It didn't. With the shuffle injected the tie set is visible: the same hand gives the first tied card under a no-op shuffle and the last under a reversing one, which also shows the untied card never wins.

tests/automa/corps/MarsBotDraftResolver.spec.ts
ignac8: Done, and it removed Random from the class entirely, since a random tie-break is just shuffling the tied cards and taking the first.

src/server/automa/corps/MarsBotDraftResolver.ts
```diff
@@ -0,0 +1,110 @@
...
+      return this.pickBest(hand, (card) => card.cost);
+    case 'leastAdvancedTrack':
+      return this.pickByLeastAdvancedTrack(hand);
+    case 'mostTags':
+      return this.pickBest(hand, (card) => this.countTags(card));
+    }
+  }
+
+  /**
+   * Reveals the drafted cards one at a time, discarding those without a priority tag and
```
kberg: "reveals the drafted cards one at a time" makes it seem like this method is logging publicly every card or something. But it's not really doing any of that, I think.

src/server/automa/corps/MarsBotDraftResolver.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    const firstMatch = cards.findIndex((card) => this.hasAnyTag(card, priority.tags));
+    if (firstMatch === -1) {
+      return {kept: cards, discarded: []};
+    }
+    return {kept: cards.slice(firstMatch), discarded: cards.slice(0, firstMatch)};
+  }
+
+  /**
+   * The highest scoring item, breaking ties at random. A hand where nothing scores at all
```
kberg: "Returns the highest scoring item". document functions like they are verbs and attributes like they are nouns.

src/server/automa/corps/MarsBotDraftResolver.ts
```diff
@@ -0,0 +1,110 @@
...
+   * is one big tie, which is how a card gets picked when none of them match.
+   */
+  private pickBest<T>(items: Array<T>, score: (item: T) => number): T {
```
kberg: I think this is a nit, but I think `scorer` would be a clearer name than `score` Then `value` below could be, you know, the `score` that comes from the `scorer`.

src/server/automa/corps/MarsBotDraftResolver.ts
```diff
@@ -0,0 +1,110 @@
...
+    let best = -1;
+    let tied: Array<T> = [];
+    for (const item of items) {
```
kberg, line 94: I'm surprised we don't have an algorithm like this already to reuse. But this is good.

src/server/automa/corps/MarsBotDraftResolver.ts
```diff
@@ -0,0 +1,110 @@
...
+    return this.pickBest(hand, (card) => this.scoreByTags(card, priorityTags));
+  }
+
+  private pickByLeastAdvancedTrack(hand: Array<IProjectCard>): IProjectCard {
+    const leastAdvanced = this.tracks.getLeastAdvancedTrackIndex();
+    return this.pickByTags(hand, this.tracks.data[leastAdvanced].tags);
+  }
+
+  /**
+   * Tags earlier in the priority list are worth more, so a card matching the first priority
```
kberg: Not to put too fine a point in it, this can have a leading sentence

> Return the score for `card` based on the priority tags.

src/server/automa/corps/MarsBotDraftResolver.ts
```diff
@@ -0,0 +1,110 @@
...
+   * beats one matching the second, and a card matching several beats one matching a single tag.
+   * Priorities only ever list printed tags, so a wild tag matches nothing and scores zero.
+   */
+  private scoreByTags(card: IProjectCard, priorityTags: ReadonlyArray<Tag>): number {
+    let score = 0;
+    for (const tag of card.tags) {
+      const priority = priorityTags.indexOf(tag);
+      if (priority >= 0) {
+        score += priorityTags.length - priority;
```
kberg: This feels a little like a magic incantation that doesn't explain how the number's score makes sense. But is this obvious from the automa's rules? If so, please move on.

src/server/automa/corps/MarsBotDraftResolver.ts
```diff
@@ -0,0 +1,110 @@
...
+      }
+    }
+    return score;
+  }
+
+  private hasAnyTag(card: IProjectCard, priorityTags: ReadonlyArray<Tag>): boolean {
+    return card.tags.some((tag) => priorityTags.includes(tag));
```
kberg: Use `utils.intersection`

tests/automa/corps/MarsBotDraftResolver.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+const spaceCard = new Asteroid(); // Space, 14 M€
+const buildingAndSpaceCard = new SpaceElevator(); // Space and Building, 27 M€
+const expensiveSpaceCard = new BigAsteroid(); // Space, 27 M€
+const earthCard = new AcquiredCompany(); // Earth, 10 M€
+const plantCard = new Algae(); // Plant, 10 M€
+const otherPlantCard = new Lichen(); // Plant, 7 M€
+
+/** Leaves the order alone, so the test sets it by writing the cards in the order it wants. */
+const keepOrder: Shuffler = () => {};
+/** Reverses the order, to show a choice really does follow the shuffle. */
```
kberg, line 31: Nice

tests/automa/corps/MarsBotDraftResolver.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+const reverseOrder: Shuffler = (items) => {
+  items.reverse();
+};
+
+function resolver(shuffle: Shuffler = keepOrder): MarsBotDraftResolver {
```
kberg: Name the variable shuffler.

tests/automa/corps/MarsBotDraftResolver.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+  return new MarsBotDraftResolver(new MarsBotBoard(THARSIS_MARSBOT_BOARD), shuffle);
+}
+
+describe('MarsBotDraftResolver', () => {
+  describe('pickCard, tags priority', () => {
+    const priority: MarsBotDraftPriority = {type: 'tags', tags: [Tag.BUILDING, Tag.SPACE]};
+
+    it('picks the card carrying a priority tag', () => {
```
kberg, line 44: This is really easy to read

kberg (commented):
Much much clearer tests

src/server/automa/corps/MarsBotDraftResolver.ts
ignac8: Reworded. Chasing the wording turned up a real problem behind it: the method discarded every card ahead of the first one carrying a priority tag, and the rules discard the first card *without* one and stop there, so at most one card ever leaves. The rulebook's aside that the action deck is "4 (in rare cases it can be 5) cards" only fits that reading, since five means nothing was discarded, which needs all four to match. Fixed, with tests.

src/server/automa/corps/MarsBotDraftResolver.ts
ignac8: Done. `pickCard`, `discardAfterDraft`, `pickBest` and `scoreByTags` all lead with a verb now.

src/server/automa/corps/MarsBotDraftResolver.ts
ignac8: Renamed. The parameter is `scorer` and the local it produces is `score`.

src/server/automa/corps/MarsBotDraftResolver.ts
ignac8: Nothing in `utils.ts` covers it. It is a little more than a max: it keeps every item tied for best and hands the tie to the injected shuffler, and the score is now a list compared entry by entry rather than a single number.

src/server/automa/corps/MarsBotDraftResolver.ts
ignac8: Added, close to your wording.

src/server/automa/corps/MarsBotDraftResolver.ts
ignac8: You were right to distrust it. It is not obvious from the rules because it does not follow them.

The corporations rulebook says to consider the first tag only, and to move to the second one only if no card matches the first, with a card carrying the first tag twice beating a card carrying the first and the second. That is a lexicographic comparison of per-tag counts, and a weighted sum is not the same thing: lower priority tags add up and overtake a higher priority one.

Celestic is the rulebook's own worked example and the old formula got it wrong. Venus over Jovian scored Venus 2 and Jovian 1, so a single Venus card and a card with two Jovian tags both scored 2 and it picked between them at random, where the example says take the Venus card. Over all 727 project cards the sum disagrees with the rules on 50 pairs for Celestic, 139 for Aphrodite, 46 for EcoTec and 156 for Tycho Magnetics.

`scoreByTags` now returns how many of each priority tag a card carries, in priority order, and `pickBest` compares those entry by entry. There is no constant left to explain.

src/server/automa/corps/MarsBotDraftResolver.ts
ignac8: Using `hasIntersection`, since the method returns a boolean and `intersection` would build the array to throw it away.

tests/automa/corps/MarsBotDraftResolver.spec.ts
ignac8: Renamed, in the spec and on the field it is passed to.

src/server/automa/corps/MarsBotDraftResolver.ts
kberg: Yeah I understand that, but there's more than one place where we rank items by score and look for first and second place, etc.

kberg (approved):

src/server/automa/corps/MarsBotDraftResolver.ts
ignac8: Right, Election, Revolution and the award scoring all walk first and second place the same way. Pulled it into rankedTiers in utils and refit the three in #8358.

PR #8210, opened 2026-06-09
Add MarsBot milestone and award evaluation

ignac8:
Next one for the MarsBot series (#8001), independent of the other open ones

Adds the table of how MarsBot qualifies for each milestone and how its award scores are computed, since it can't meet the normal card and resource based criteria. Everything works against a small context object so the 39 tests just fabricate contexts directly, no game needed


src/server/automa/MarsBotMilestoneAwardEval.ts
```diff
@@ -0,0 +1,214 @@
+import {MilestoneName} from '../../common/ma/MilestoneName';
+import {AwardName} from '../../common/ma/AwardName';
+
+/**
+ * Context provided to milestone/award evaluation functions for MarsBot.
+ */
+export type MarsBotMAContext = {
```
kberg: Is this full context necessary? Instead, like the app's awards and milestones, can you read the game and the bot?

kberg (changes requested):
I haven't done a full review at all, but this stood out and probably forces a significant change.

tests/automa/MarsBotMilestoneAwardEval.spec.ts
```diff
@@ -0,0 +1,252 @@
...
+      const ctx = mockContext({cityCount: 3});
+      expect(MILESTONE_EVALS.get('Mayor')!(ctx)).to.be.true;
+    });
+
+    it('Mayor: cityCount < 3 returns false', () => {
+      const ctx = mockContext({cityCount: 2});
+      expect(MILESTONE_EVALS.get('Mayor')!(ctx)).to.be.false;
+    });
+
+    it('Builder: trackPos(1) >= 8 returns true', () => {
```
kberg: trackPos(1) represents index 0?

src/server/automa/MarsBotMilestoneAwardEval.ts
ignac8: It was a shadow of state that already exists. The evals take IMarsBot now and read the game the way the real ones do, so Mayor is `game.board.getCities(bot.player).length`. IMarsBot gained two members for what only the bot knows, its played project cards and its temperature raises. Eight context fields had no reader at all and went with it, including a flood fill computing largestConnectedTileGroup for nobody.

Two bugs came out of it. Hydrologist counted ocean spaces owned by the bot, but simpleAddTile leaves ocean tiles unowned, so it was always 0 and the milestone could never be claimed; it reads globalParameterSteps now, like the real Hydrologist does. Visionary appended venusTrackPos to allTrackPositions, which already contains the Venus track when Venus is in play, so Venus was counted twice.

tests/automa/MarsBotMilestoneAwardEval.spec.ts
ignac8: No, the test name was wrong; the code checks index 0. The spec is rewritten and the names match what they assert.

src/server/automa/MarsBotMilestoneAwardEval.ts
```diff
@@ -0,0 +1,193 @@
...
+    bot.game.board.getGreeneries(bot.player).length >= 1 && trackPos(bot, 0) >= 5],
+  ['Legend4', (bot) => playedCards(bot, isEvent) >= 4],
+  ['Lobbyist', () => false],
+  ['Merchant', (bot) => allMarsTracksAtOrAbove(bot, 2)],
+  ['Metallurgist', (bot) => trackPos(bot, 0) + trackPos(bot, 1) >= 9],
+  ['Philantropist', (bot) => playedCards(bot, (card) => card.getVictoryPoints(bot.player) >= 0) >= 5],
+  ['Pioneer4', () => undefined],
+  ['Planetologist', () => false],
+  ['Producer', (bot) => topThreeTracksSum(bot) >= 16],
+  ['Researcher', (bot) => trackPos(bot, 3) >= 4],
```
kberg, line 126: These are hard-coded numbers for indexes, not maintainable.

src/server/automa/MarsBotMilestoneAwardEval.ts
```diff
@@ -0,0 +1,193 @@
...
+const isGreen = (card: IProjectCard) => card.type !== CardType.EVENT && card.type !== CardType.ACTIVE;
+
+/**
+ * How MarsBot claims a milestone. True when it qualifies, false when it does not, and
+ * undefined for the milestones it reads off the board like any other player.
+ */
+type MilestoneEval = (bot: IMarsBot) => boolean | undefined;
+export const MILESTONE_EVALS = new Map<MilestoneName, MilestoneEval>([
+  // Tharsis
+  ['Terraformer', (bot) => bot.player.terraformRating >= 35],
```
kberg, line 76: I don't think this is a good approach. You're creating a complete sideloaded set of code that could  be, for instance, made part of IMilestone and IAward. You're doubling up on milestone claim levels, and so on. Please suggest an alternative that works with IMilestone and IAward. Tests can then become part of those classese.

ignac8: I had the AI audit the table against the classes, then we checked rulebook C itself: pages 9-10 cover the official maps and 13-16 the Awards & Milestones expansion, so most non-Tharsis entries do have a source. It now works your way with those as overrides: the bot goes through canClaim and getScore with its player, IMilestone and IAward get optional marsBotCanClaim/marsBotScore, and the classes where the automa rules differ override them, tested in their class specs. Entries the rulebook marks Unchanged fall through to the normal rules. Checking the source also fixed a dozen spots the table got wrong (Fundraiser and Promoter read the wrong track, Trader and Lobbyist could never be claimed) and Terraformer under Turmoil, where the class knows the 26 threshold the table hardcoded as 35.

src/server/automa/MarsBotMilestoneAwardEval.ts
ignac8: Gone with the redesign. The overrides read tracks through getTrackIndexForTag, so no bare numbers are left.

src/server/automa/MarsBotCorpTypes.ts
```diff
@@ -77,6 +77,10 @@ export interface IMarsBot {
   mcSupply: number;
   /** Floaters stored by Venus corps. */
   readonly floaterCount: number;
+  /** Project cards the bot has drawn and resolved, in the order it played them. */
+  readonly playedProjectCards: ReadonlyArray<IProjectCard>;
+  /** How many times the bot has raised the temperature. */
+  readonly temperatureRaises: number;
```
kberg, line 83: Use `player.globalParameterSteps`

src/server/awards/modular/Administrator.ts
```diff
@@ -8,4 +9,9 @@ export class Administrator implements IAward {
   public getScore(player: IPlayer): number {
     return player.tags.numberOfCardsWithNoTags();
   }
+
+  /** MarsBot counts the tagless cards in its own played pile, plus 2. */
+  public marsBotScore(bot: IMarsBot): number {
+    return bot.playedProjectCards.filter((card) => card.tags.length === 0).length + 2;
```
kberg, line 15: Doesn't player have playedCards? I guess not?

src/server/automa/MarsBotMilestoneAwardEval.ts
```diff
@@ -0,0 +1,36 @@
...
+  return index === undefined ? 0 : bot.board.tracks[index].position;
+}
+
+/** Every track position, the Venus track included when in play. */
+export function marsBotAllTrackPositions(bot: IMarsBot): Array<number> {
+  return bot.board.tracks.map((track) => track.position);
+}
+
+/** The Mars track positions, leaving a Venus track out. */
+export function marsBotMarsTrackPositions(bot: IMarsBot): Array<number> {
```
kberg, line 32: This is a bad name for a function that filters Venus.

Have you reviewed this change? What do you think?

src/server/automa/MarsBotMilestoneAwardEval.ts
```diff
@@ -0,0 +1,36 @@
+import {Tag} from '../../common/cards/Tag';
+import {IMilestone} from '../milestones/IMilestone';
+import {IAward} from '../awards/IAward';
+import {IMarsBot} from './MarsBotCorpTypes';
+
+/**
+ * MarsBot claims milestones and scores awards through the ordinary IMilestone and IAward
```
kberg, line 7: Once again, poor comment style. Please review your PR manually.

src/server/awards/terraCimmeria/Biologist.ts
```diff
@@ -9,4 +11,9 @@ export class Biologist implements IAward {
   public getScore(player: IPlayer): number {
     return player.tags.multipleCount([Tag.MICROBE, Tag.PLANT, Tag.ANIMAL], 'award');
   }
+
+  /** The automa rules score MarsBot's Biologist as its bio track plus 5. */
```
kberg, line 15: Is this a useful comment?

kberg (changes requested):
Note my comment about that I think this requires another human pass. In addition, this is still too big for me to review. Can this be split down into maybe a couple of milestones, and/or a couple of awards? Then it will make more sense.

PR #8211, opened 2026-06-09
Add MarsBot Turmoil delegate helpers

ignac8:
Next one for the MarsBot series (#8001)

Adds the Turmoil delegate logic for MarsBot: which party it sends a delegate to (the T-7 priority list from the automa rules), party leader updates and delegate counting

Game setup is split out to #8369 and this PR sits on top of it: the spec now builds a real automa game via the automa option instead of a second player standing in as the bot


src/server/automa/turmoil/MarsBotTurmoilHelper.ts
```diff
@@ -0,0 +1,173 @@
+import {Turmoil} from '../../turmoil/Turmoil';
+import {IParty} from '../../turmoil/parties/IParty';
+import {PartyName} from '../../../common/turmoil/PartyName';
+import {IPlayer} from '../../IPlayer';
+import {IGame} from '../../IGame';
+
+/**
+ * T-7: Select which party MarsBot should place a delegate into.
```
kberg: This is not what the first sentence of a function JSdoc should be

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
```diff
@@ -0,0 +1,173 @@
...
+export function selectPartyForDelegate(
+  turmoil: Turmoil,
+  marsBotPlayer: IPlayer,
+  humanPlayer: IPlayer,
+): PartyName | undefined {
+  if (!turmoil.hasDelegatesInReserve(marsBotPlayer)) {
+    return undefined;
+  }
+
+  const allParties = turmoil.parties as ReadonlyArray<IParty>;
```
kberg: These two lines are awkward.

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
kberg, line 1: If it's not a class, don't use caps. Caps like this are reserved for classes, so here I would expect a MarsBotTurmoilHelper. `turmoilHelper.ts` is fine, though you might want marsBot..

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
```diff
@@ -0,0 +1,173 @@
...
+}
+
+/**
+ * After placing a MarsBot delegate, set MarsBot as party leader if it now has
+ * strictly more delegates than the current leader.
+ *
+ * The Party class's checkPartyLeader only iterates game.playersInGenerationOrder + NEUTRAL.
+ * MarsBot is not in that list, so we must update leadership manually.
+ */
+export function updatePartyLeaderForMarsBot(party: IParty, marsBotPlayer: IPlayer): void {
```
kberg: This is a case for a `maybe` titled method.

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
```diff
@@ -0,0 +1,173 @@
...
+ *
+ *  Tier 1: Party where +1 makes MarsBot become Party Leader AND the party become Dominant.
+ *  Tier 2: Party where +1 makes MarsBot become Party Leader.
+ *  Tier 3: Party where MarsBot is already Party Leader, and +1 makes it become Dominant
+ *          (party must NOT already be Dominant — both conditions require a state transition).
+ *  Tier 4: Party where the human player has fewest delegates (incl. zero).
+ *  Tier 5: Party where MarsBot has fewest delegates (incl. zero).
+ *  Tier 6: Next party clockwise from the Dominance marker (circling around). Always yields 1.
+ */
+export function selectPartyForDelegate(
```
kberg: _maybe_ if this was a class, though, you could make attributes out of turmoil and player etc, and so not have to nested parameter passing all the way down. Plus you can store attributes like "Would I be the party leader for Greens?" without having to call it so many times.

That last part is premature optimization, since I don't think we have a huge CPU problem, but it merits being pointed out.

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
```diff
@@ -0,0 +1,173 @@
...
+): ReadonlyArray<IParty> {
+  const filtered = candidates.filter(predicate);
+  return filtered.length >= 1 ? filtered : candidates;
+}
+
+/**
+ * T-7 + game engine: Place 1 MarsBot delegate from reserve into the selected party.
+ * Returns the party name the delegate was placed in, or undefined if no reserve.
+ */
+export function placeDelegateForMarsBot(
```
kberg: There's no need to pass all these parameters. They can be inferred by humanPlayer or marsBotPlayer.

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
```diff
@@ -0,0 +1,173 @@
...
+    party.delegates.get(party.partyLeader) :
+    0;
+  if (marsBotCount > leaderCount) {
+    party.partyLeader = marsBotPlayer;
+  }
+}
+
+/** Count all delegates for `player` in `party`. */
+export function totalDelegates(party: IParty, player: IPlayer): number {
+  return party.delegates.get(player);
```
kberg: This is not a helpful function

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
```diff
@@ -0,0 +1,173 @@
...
+  // Must not already be the dominant party ("becomes" implies a transition)
+  if (turmoil.dominantParty === party) {
+    return false;
+  }
+  const partyCountAfter = party.delegates.size + 1;
+  return allParties.every((p) => p === party || p.delegates.size < partyCountAfter);
+}
+
+/**
+ * Clockwise distance from `fromIndex` to `toIndex` in a circular array of `size`.
```
kberg, line 118: I suspect the first sentence is more than enough. Maybe first and third.

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
```diff
@@ -0,0 +1,173 @@
...
+ * Clockwise = decreasing-index direction (consistent with Turmoil.setNextPartyAsDominant).
+ * Distance from a party to itself is `size` (to put it last in clockwise sort).
+ */
+function clockwiseDistance(fromIndex: number, toIndex: number, size: number): number {
+  if (fromIndex === toIndex) {
+    return size;
+  } // same party — put last
```
kberg: That's not where the comment should go

kberg (changes requested):

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
kberg: also what's the point of `as` here?

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
```diff
@@ -0,0 +1,173 @@
...
+/**
+ * Narrowing helper: if any party in `candidates` passes `predicate`, return only those.
+ * Otherwise return `candidates` unchanged.
+ */
+function narrow(
+  candidates: ReadonlyArray<IParty>,
+  predicate: (p: IParty) => boolean,
+): ReadonlyArray<IParty> {
+  const filtered = candidates.filter(predicate);
+  return filtered.length >= 1 ? filtered : candidates;
```
kberg: `> 0`

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
```diff
@@ -0,0 +1,173 @@
...
+  const partyName = selectPartyForDelegate(turmoil, marsBotPlayer, humanPlayer);
+  if (partyName === undefined) {
+    return undefined;
+  }
+  turmoil.sendDelegateToParty(marsBotPlayer, partyName, game);
+  // Party.checkPartyLeader only considers game.playersInGenerationOrder + NEUTRAL, not MarsBot.
+  // Manually update the party leader if MarsBot now has the most delegates.
+  const party = turmoil.getPartyByName(partyName);
+  updatePartyLeaderForMarsBot(party, marsBotPlayer);
+  game.log('MarsBot places delegate in ${0} (Party Politics)', (b) => b.partyName(partyName));
```
kberg: Is PartyPolitics an automa thing?

I'm not sure, but I might ask if this can be more general `${0} adds a delegate in ${1}`,

kberg (changes requested):

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
ignac8: Rewritten, first sentence says what it does and the tier table follows.

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
ignac8: The cast did nothing: createParties already returns ReadonlyArray<IParty>. Gone, and the two lines with it.

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
ignac8: It's a class now, so the file name stands.

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
ignac8: Done. The game, turmoil and both players are constructor attributes, so nothing is threaded through the tier checks any more.

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
ignac8: Gone, maybePlaceDelegate takes none.

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
ignac8: Renamed to maybeUpdatePartyLeader. Its test also asserted `leader === marsBot || leader === before`, which covers every possible outcome; it now pins the three real cases.

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
ignac8: Deleted. It wrapped one call to party.delegates.get.

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
ignac8: Trimmed to the first and third, as you guessed.

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
ignac8: Moved above the branch.

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
ignac8: Done.

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
ignac8: Party Politics is the automa bonus card that triggers the placement, not an engine concept. The message is generic now: `${0} adds a delegate in ${1}`.

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
```diff
@@ -0,0 +1,138 @@
...
+   * Returns the party it went to, or undefined when the reserve is empty.
+   */
+  public maybePlaceDelegate(): PartyName | undefined {
+    const partyName = this.selectParty();
+    if (partyName === undefined) {
+      return undefined;
+    }
+    this.turmoil.sendDelegateToParty(this.marsBotPlayer, partyName, this.game);
+    this.maybeUpdatePartyLeader(this.turmoil.getPartyByName(partyName));
+    this.game.log('${0} adds a delegate in ${1}', (b) => b.player(this.marsBotPlayer).partyName(partyName));
```
kberg: Log messages typically use past tense, so this should be "added a delegate"

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
```diff
@@ -0,0 +1,138 @@
...
+   * Sends one delegate from MarsBot's reserve to the party it picks (T-7).
+   * Returns the party it went to, or undefined when the reserve is empty.
+   */
+  public maybePlaceDelegate(): PartyName | undefined {
+    const partyName = this.selectParty();
+    if (partyName === undefined) {
+      return undefined;
+    }
+    this.turmoil.sendDelegateToParty(this.marsBotPlayer, partyName, this.game);
+    this.maybeUpdatePartyLeader(this.turmoil.getPartyByName(partyName));
```
kberg: Is this line necessary? turmoil.sendDelegateToParty calls party.addDelegate, which calls its internal checkPartyLeader. (which I could rename `maybeUpdatePartyLeader`)

I suspect your maybeUpdatePartyLeader is entirely unnecessary.

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
```diff
@@ -0,0 +1,138 @@
...
+      return false;
+    }
+    return party.delegates.get(this.marsBotPlayer) + 1 > this.leaderCount(party);
+  }
+
+  /**
+   * Would placing one delegate make this party the new Dominant party?
+   * The party must not already be Dominant, since "becomes" implies a transition.
+   */
+  private wouldBecomeDominant(party: IParty): boolean {
```
kberg, line 95: These are actually pretty useful functions. I wonder if they can move to IParty (eventually)

tests/automa/turmoil/MarsBotTurmoilHelper.spec.ts
```diff
@@ -0,0 +1,203 @@
...
+    }
+    expect(turmoil.hasDelegatesInReserve(marsBotPlayer)).is.false;
+
+    expect(helper.selectParty()).is.undefined;
+  });
+});
+
+describe('maybePlaceDelegate', () => {
+  it('sends a delegate from the reserve to the chosen party', () => {
+    const {turmoil, marsBotPlayer, helper} = createTurmoilGame();
```
kberg: This is an odd way of doing it.

kberg (commented):
I need to review the tests some more but let's see how the code plays out after this review. Thanks.

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
ignac8: Changed to "added a delegate in".

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
ignac8: It is load bearing. `checkPartyLeader` chooses the new leader from `game.playersInGenerationOrder` plus NEUTRAL, and MarsBot's player is deliberately in neither, so it can raise `max` past the current leader's count and then find nobody to promote. With a human on 1 delegate and MarsBot on 2, the human stays party leader.

The spec did not catch that because it cast a regular second player as MarsBot, and `checkPartyLeader` does find that one. It now builds the player the way the game builds MarsBot's, outside `game.players` and with its own delegate reserve, and there is a test for the leadership that fails if this line goes.

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
ignac8: They would fit there. Neither one is MarsBot specific: both only ask what one more delegate would do to a party.

tests/automa/turmoil/MarsBotTurmoilHelper.spec.ts
ignac8: The odd part was load bearing in the wrong direction: a two player game was standing in for MarsBot, so `checkPartyLeader` treated the stand-in as a player it could promote and the party leader test above passed for a reason that does not hold in a real game. The setup now builds the player the way `AutomaGameSetup` does, and says why.

If what looked odd was the destructured bundle rather than the game, say so and I will inline `testGame` per test instead.

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
kberg: I don't know what you mean by it's load bearing. But I think you mean checkPartyLeader doesn't work with automa players.

Nonetheless, I think checkPartyLeader ought to be able to handle this.

tests/automa/turmoil/MarsBotTurmoilHelper.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+ * set up against the game, given its own delegate reserve, and left out of game.players.
+ *
+ * Standing a regular second player in for MarsBot would be easier to read but would test the
+ * wrong thing, because Party.checkPartyLeader can only ever promote someone it finds in
+ * game.playersInGenerationOrder.
+ */
+function createTurmoilGame() {
+  const [game, humanPlayer] = testGame(1, {turmoilExtension: true});
+  const turmoil = Turmoil.getTurmoil(game);
+  const marsBotPlayer = new Player('MarsBot', 'bronze', false, 0, 'p-marsbot' as PlayerId);
```
kberg: as is not necessary.

tests/automa/turmoil/MarsBotTurmoilHelper.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+/**
+ * Builds a Turmoil game whose MarsBot player is put together the way the game puts it together:
+ * set up against the game, given its own delegate reserve, and left out of game.players.
+ *
+ * Standing a regular second player in for MarsBot would be easier to read but would test the
+ * wrong thing, because Party.checkPartyLeader can only ever promote someone it finds in
+ * game.playersInGenerationOrder.
+ */
+function createTurmoilGame() {
+  const [game, humanPlayer] = testGame(1, {turmoilExtension: true});
```
kberg, line 15: Will game eventually create the marsBotPlayer?

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
```diff
@@ -0,0 +1,138 @@
...
+
+/**
+ * Places MarsBot's delegates and keeps its party leadership up to date.
+ *
+ * Party.checkPartyLeader only considers game.playersInGenerationOrder plus NEUTRAL, and MarsBot
+ * is in neither, so leadership has to be applied here instead.
+ */
+export class MarsBotTurmoilHelper {
+  constructor(
+    private readonly game: IGame,
```
kberg: humanPlayer and turmoil are obviously inferred from game. So should marsBotPlayer.

The original implementation passed player and game along everywhere, and it was too easy to accidentally send players and games that were not actually correlated. Same should happen here. If that requires another bit of code first, that's OK.

I'd strongly recommend you suggest an approach @ignac8 instead of the AI.

kberg (changes requested):

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
ignac8: Moved it into checkPartyLeader. It now also considers delegates that aren't game players when picking a new leader, after the players and NEUTRAL, so ordering stays the same for regular games. maybeUpdatePartyLeader is gone and the spec covers the overtake, the tie and the removal case through sendDelegateToParty.

tests/automa/turmoil/MarsBotTurmoilHelper.spec.ts
ignac8: Dropped. PlayerId is p${string} so the literal fits as is.

tests/automa/turmoil/MarsBotTurmoilHelper.spec.ts
ignac8: Yes. The automa setup creates it when the automa option is on, that part lands with the game integration. Until then the test builds the player itself and hangs it on game.automaHooks.

src/server/automa/turmoil/MarsBotTurmoilHelper.ts
ignac8: Fair point. I asked the AI for three options and picked the one that made most sense to me: the game gets an automaHooks field, same as my full branch already has. Upstream it starts as a small IAutomaGameHooks interface holding just the MarsBot player, and it grows as more automa parts land. The class on my branch implements it, so the two stay in line. The helper ctor is now just (game) and reads turmoil, the human and MarsBot from it.

The other two options were a single automaPlayer field on IGame, smaller but it leaves two automa surfaces once the hooks arrive anyway, and keeping the marsBotPlayer parameter until game integration lands, which you already answered. If you want a different shape or name, tell me and I will redo it.

src/server/automa/IAutomaGameHooks.ts
```diff
@@ -0,0 +1,10 @@
+import {IPlayer} from '../IPlayer';
+
+/**
+ * The game's automa (MarsBot) machinery, present when the automa variant is on.
+ * Grows as the automa port lands. For now it only names the MarsBot player,
+ * which is deliberately not part of game.players.
+ */
+export interface IAutomaGameHooks {
```
kberg, line 8: In this codebase, interfaces are only for when you create classes.

tests/automa/turmoil/MarsBotTurmoilHelper.spec.ts
```diff
@@ -0,0 +1,232 @@
...
+ * Builds a Turmoil game with its MarsBot player put together the way the game will put it
+ * together: set up against the game, registered in automaHooks, given its own delegate
+ * reserve, and deliberately not in game.players.
+ */
+function createTurmoilGame() {
+  const [game, humanPlayer] = testGame(1, {turmoilExtension: true});
+  const turmoil = Turmoil.getTurmoil(game);
+  const marsBotPlayer = new Player('MarsBot', 'bronze', false, 0, 'p-marsbot');
+  marsBotPlayer.setup(game);
+  game.automaHooks = {marsBotPlayer};
```
kberg, line 19: This kind of setup shouldn't be in the test, it should be part of creating the game.

I think game setup should be its own PR.

src/server/turmoil/parties/Party.ts
```diff
@@ -70,6 +70,13 @@ export abstract class Party {
       // Add NEUTRAL in the list
       playersToCheck.push('NEUTRAL');
 
+      // A delegate that is not a game player (MarsBot's) can still lead the party.
+      for (const other of new Set(this.delegates)) {
```
kberg, line 74: if (game.automaHooks.marsBotPlayer) {
  if (this.delegates.has(game.automaHooks.marsBotPlayer))
  ...

kberg (changes requested):

kberg:
Speaking of which, why isn't IMarsBot part of the automa hooks?

src/server/automa/IAutomaGameHooks.ts
ignac8: Made it a type. My branch has the class that ends up implementing it, but that is a later PR.

tests/automa/turmoil/MarsBotTurmoilHelper.spec.ts
ignac8: Split it out: #8369 adds the automa option and creates the MarsBot player during game setup, with the turmoil delegates and the reload path. This PR sits on top and the test builds a real automa game now.

src/server/turmoil/parties/Party.ts
ignac8: Took your shape. The bot goes in after NEUTRAL, so regular games keep their ordering.

ignac8:
It will be. IMarsBot is the corp handlers' view of the bot manager, and nothing in this PR can build one yet: the manager with its board and decks comes in later PRs. The turmoil helper only needs the bot's IPlayer for delegates, so the hooks carry just that for now. On my branch the hooks hold the full bot, and the type here grows to match as those PRs land.

PR #8225, opened 2026-06-13
Fix optional input timer accounting

dcep93:
## Summary
- Do not stop a player timer when processing an optional input.
- Do not stop a player timer when clearing an optional input.
- Add regression tests for both optional input paths.

## Root cause
PR #8151 made optional inputs skip `timer.start()`, but `Player.process` and `Player.clearWaitingFor` still called `timer.stop()` unconditionally. Since `Timer.stop()` calculates elapsed time from the prior `startedAt`, stopping an optional input can incorrectly add stale elapsed time to the player timer.

## Validation
- `npx mocha --import=tsx --require tests/testing/setup.ts tests/Player.spec.ts`
- `npm run make:json && npm run build:test`
- `npm run lint:server -- --no-cache src/server/Player.ts tests/Player.spec.ts`

Note: local npm emitted the existing engine warning because this machine is on Node v25.8.1 while package.json requires Node 22.x.

### DIFF
--- a/src/server/Player.ts
+++ b/src/server/Player.ts
@@ -1692,7 +1692,9 @@ export class Player implements IPlayer {
     this.waitingFor = undefined;
     this.waitingForCb = undefined;
     try {
-      this.timer.stop();
+      if (!waitingFor.optional) {
+        this.timer.stop();
+      }
       this.defer(waitingFor.process(input, this));
       waitingForCb();
     } catch (err) {
@@ -1747,9 +1749,12 @@ export class Player implements IPlayer {
   }
 
   public clearWaitingFor(): void {
+    const waitingFor = this.waitingFor;
     this.waitingFor = undefined;
     this.waitingForCb = undefined;
-    this.timer.stop();
+    if (waitingFor !== undefined && !waitingFor.optional) {
+      this.timer.stop();
+    }
   }
 
   public serialize(): SerializedPlayer {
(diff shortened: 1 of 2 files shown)


dcep93:
I haven't actually read or validated this yet, but I noticed the bug and had my AI identify and fix. Hopefully I'll get time to look at it properly, add comments, and check safety/correctness.

traneric94 (approved):

kberg (approved):
Thank you

kberg:
> I haven't actually read or validated this yet, but I noticed the bug and had my AI identify and fix. Hopefully I'll get time to look at it properly, add comments, and check safety/correctness.

Could you expedite this? I'd like to get a fix in.

dcep93:
https://github.com/user-attachments/assets/17947a3c-aa9e-4c9c-a089-b007cbe16eab

kberg:
Daniel, I didn't pay attention and now I realize you're not the original feature author. Thanks for stepping in! If you feel confident about this change based on your manual test (thanks for the video!) then remove the draft status and I'll merge it in.

dcep93:
I think this PR is good to land, thanks for taking a look!

Looks like the optional-updater change was undeployed in prod, so localhost:8081 shows the broken latest-main checkout, and localhost:8080 performs the same actions, validating the fix.

1. initialize a 2 player game with draft and timers, pick a starting hand, pass gen 1
2. red player picks their draft card immediately
3. green player waits 1 minute, then picks their first draft card
4. we expect that red player shouldnt have that step-3 minute against their clock

There's still an ambiguous gap, let me know if I should have my clanker patch it. If red *updates* their selection during step 3 above, should they be on the hook for that time they spent deciding on that update? Perhaps so, because they benefited from that time, but perhaps not, because no one was "waiting for them" during that time.

Specifically from codex:
> I don’t think we should use the static/global `Timer.lastStoppedAt` to implement that. If we want to cover this case, the safer version is to record a per-timer/per-player “last input stopped at” timestamp and charge optional submissions from that.

kberg:
Good thought. Let's leave it as is. Thanks again!

PR #8226, opened 2026-06-15
fix(i18n): German translation

Borbarad2:
Change, correct and add-ons for translation purpose


kberg:
Delta Project is in alpha, I'm not accepting translations.

This is the second time I am telling you this. It is also not the first time I have asked you not to do something but you do anyway, making work for me, so it's why I have run out of patience.

kberg:
Delta Project is in alpha, I'm not accepting translations.

This is the second time I am telling you this. It is also not the first time I have asked you not to do something but you do anyway, making work for me, so it's why I have run out of patience.

Borbarad2:
Do what ever you want to do with it.
But instead of deleting/closing it, leave it open for any time given, and merge it any time later you want and prefer. 

Most of that PR is **not** Delta Project, i.e. repicking drafted card, new messages for oceans etc., and all other lines refering to Delta Project I can delete if you like/prefer.

PR #8230, opened 2026-06-15
Smaller titles for ua cards.

kberg:
Matches the style for ru.


kberg:
FYI [Zapara](https://github.com/yzapara)

PR #8239, opened 2026-06-17
Add Delta Project logo

TarunTheo13:
Have added the logo, let me know if you want me to turn it into a circle (it's currently a square)


kberg:
A circle would be nice. Also note that a separate style is necessary for the card list: 
image

kberg:
But this is good enough for now.

kberg (approved):

PR #8240, opened 2026-06-17
Measure card titles to fit, gated behind experimental UI

kberg:
Replace the character-count heuristic for sizing card titles with a measurement-based fit: shrink the font until the rendered text stops overflowing its box, which is accurate across fonts and languages.

- titleFit.ts: measure-and-shrink, plus timing metrics.
- titleFontSizeCache.ts: per-title localStorage cache (2-day expiry via expireMs, set on write, not extended on read) so return visits skip the layout-forcing re-measurement.
- CardTitle.vue: fit only when experimental_ui is on; otherwise keep the length-based size classes.
- CardList.vue: log total resize time on /cards open.
- language_hacks.less: gate the title !important overrides to experimental_ui off, so they don't fight the measured size.


kberg:
The /cards endpoint is between 300 to 800 ms slower on first run but given that's for hundreds of cards, it's pretty efficient.

PR #8241, opened 2026-06-17
Update delta project icon

kberg:


kberg:
FYI @TarunTheo13 . An easy tweak.

kberg:
FYI @TarunTheo13 . An easy tweak.

PR #8244, opened 2026-06-18
Improve language switcher: highlight current language, faster tooltip, six per row

kberg:
## What

Three small improvements to the language switcher on the start screen:

- **Highlight the current language** with an oval ring matching the flag shape.
- **Faster tooltips.** Replace the native `title` attribute (which has a long, browser-controlled delay) with a custom CSS tooltip from `data-title` that appears immediately on hover.
- **Six flags per row, deterministically.** Previously the per-row count varied by locale (5 or 6) because spacing was built from three stacked sources — the flex `gap`, a per-flag right margin, and a literal `&nbsp;` between flags (a real flex item whose width depends on locale font metrics). This removes the margin and the `&nbsp;`, leaving spacing to the flex `gap` alone, and fixes the switcher width to fit exactly six flags regardless of language.

## Notes

Pure client/CSS change; no server or game-logic impact.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/client/components/LanguageSwitcher.vue
+++ b/src/client/components/LanguageSwitcher.vue
@@ -2,8 +2,8 @@
   <div class="language-switcher">
     <template v-for="lang in ALL_LANGUAGES" :key="lang">
     <div
-      :class="`language-icon language-icon--${lang} language-icon-for-switcher`"
-      :title="title(lang)"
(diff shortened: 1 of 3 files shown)


kberg:
image

PR #8249, opened 2026-06-19
Make Tosche Station card text clearer

TarunTheo13:
I've been thinking these changes would decrease confusion around this card. It was a very simple change so I thought I would just create a PR & you can be the judge.

Current card:

Screenshot 2026-06-19 at 06 09 54

Proposed change:

Screenshot 2026-06-19 at 06 09 18


kberg (approved):
It is better. Do you want to update the translation strings or would you like me to do it?

TarunTheo13:
> It is better. Do you want to update the translation strings or would you like me to do it?

I've never done it before. Maybe you do it and then I know for the future

kberg:
> > It is better. Do you want to update the translation strings or would you like me to do it?
> 
> I've never done it before. Maybe you do it and then I know for the future

Sure thing. Look at the commit I added to this PR.

PR #8255, opened 2026-06-20
Randomize Maria colony choices

rusliksu:
## Summary
- Make Maria draw a shuffled subset of unused colony tiles, then choose one tile from that drawn subset.
- Leave explicit "of your choice" colony-tile effects such as Aridor and Prospecting unchanged.
- Avoid changing discarded-colony serialization or the public game model.

## Demo
- Repro fixture: discarded colonies are `[Titan, Callisto]`, generation is 1, and the RNG is forced so the shuffled offer is `Callisto`.
- Expected behavior: Maria shows only the randomized drawn subset, then the player chooses one tile from that subset.

## Testing
- `npx mocha --import=tsx --require tests/testing/setup.ts tests/cards/ceo/Maria.spec.ts`
- `npx eslint src/server/cards/ceos/Maria.ts tests/cards/ceo/Maria.spec.ts`
- `npm run build`

## Notes
- Replaces the previous broader discarded-colony-order approach with the narrower Maria-only random draw suggested in review.
- No change to Aridor/Prospecting choice behavior.

### DIFF
--- a/tests/cards/ceo/Maria.spec.ts
+++ b/tests/cards/ceo/Maria.spec.ts
@@ -9,7 +9,9 @@ import {Venus} from '../../../src/server/cards/community/Venus';
 import {Celestic} from '../../../src/server/cards/venusNext/Celestic';
 import {IapetusII} from '../../../src/server/cards/pathfinders/IapetusII';
 import {CollegiumCopernicus} from '../../../src/server/cards/pathfinders/CollegiumCopernicus';
-import {cast} from '../../../src/common/utils/utils';
+import {Callisto} from '../../../src/server/colonies/Callisto';
+import {Titan} from '../../../src/server/colonies/Titan';
+import {cast, toName} from '../../../src/common/utils/utils';
 
 describe('Maria', () => {
   let card: Maria;
@@ -35,6 +37,19 @@ describe('Maria', () => {
     selectColony.cb(selectedColony);
     expect(game.colonies).to.contain(selectedColony);
     expect(game.colonies).has.length(coloniesInPlay + 1);
+  });
+
+  it('Draws from discarded colonies without mutating their order', () => {
+    game.discardedColonies = [new Titan(), new Callisto()];
+
+    cast(card.action(player), undefined);
+    runAllActions(player.game);
+
+    const selectColony = cast(player.popWaitingFor(), SelectColony);
+    const drawnColonyNames = selectColony.colonies.map(toName);
+    expect(drawnColonyNames).has.length(1);
+    expect(['Titan', 'Callisto']).contains(drawnColonyNames[0]);
+    expect(game.discardedColonies.map(toName)).deep.eq(['Titan', 'Callisto']);
   });
 
   it('Takes action in Generation 4', () => {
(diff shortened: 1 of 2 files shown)


kberg:
Hi -- thanks for this! What problem is this solving?

kberg:
Oh wait I didn't see the comment about Maria etc... let me read.

kberg:
Sorry, I'm still not clear. What's the bug?

src/server/colonies/ColonyDealer.ts
```diff
@@ -47,6 +47,14 @@ export class ColonyDealer {
     return gameOptions.customColoniesList.some((colonyName) => communityColonyNames.includes(colonyName));
   }
 
+  private validColoniesForGameOptions(): Array<IColony> {
+    const customColonies = this.gameOptions.customColoniesList;
+    if (customColonies.length > 0) {
+      return ALL_COLONIES_TILES.filter((c) => customColonies.includes(c.colonyName)).map((cf) => new cf.Factory());
+    }
+    return [...this.gameColonies];
```
kberg: Why the spread operator?

src/server/models/ServerModel.ts
```diff
@@ -58,7 +58,7 @@ export class Server {
       colonies: coloniesToModel(game, game.colonies, false, true),
       deckSize: game.projectDeck.drawPile.length,
       discardPileSize: game.projectDeck.discardPile.length,
-      discardedColonies: game.discardedColonies.map(toName),
+      discardedColonies: game.discardedColonies.map(toName).sort(),
```
kberg: What value is sorting this?

src/server/Game.ts
```diff
@@ -1771,8 +1772,13 @@ export class Game implements IGame, Logger {
     // Reload colonies elements if needed
     if (gameOptions.coloniesExtension) {
       game.colonies = ColonyDeserializer.deserializeAndFilter(d.colonies);
-      const colonyDealer = new ColonyDealer(rng, gameOptions);
-      colonyDealer.restore(game.colonies);
+      // Legacy saves do not have a stored discarded-colony order. Use a throwaway RNG so
```
kberg: What case is this optimization addressing?

kberg (changes requested):
After reading all of this, what if instead, Maria randomly chose `n` colonies from the unserialized `discardedColonies`?

rusliksu:
I replaced this with the narrower Maria-only approach you suggested. Maria now shuffles a copy of `discardedColonies` and offers only the drawn `n` tiles; Aridor and Prospecting keep the existing full-pool choice behavior because their text says "of your choice".

tests/cards/ceo/Maria.spec.ts
```diff
@@ -37,6 +39,23 @@ describe('Maria', () => {
     expect(game.colonies).has.length(coloniesInPlay + 1);
   });
 
+  it('Randomly draws from discarded colonies without mutating their order', () => {
+    game.discardedColonies = [new Titan(), new Callisto()];
+    const nextInt = game.rng.nextInt;
```
kberg: Instead of worrying about the randomizer, just use .to.have.members (or is it .contains?)

rusliksu: Done, I removed the RNG stub and now assert that Maria offers one drawn colony from the available discarded pool without mutating that pool before the player chooses.

kberg (approved):

PR #8258, opened 2026-06-21
Log Splice microbe additions

rusliksu:
## Summary
- Log Splice's Add microbe option using the existing card-resource log path.
- No gameplay change.

## Testing
- `npx mocha --import=tsx --require tests/testing/setup.ts tests/cards/promo/Splice.spec.ts`
- `git diff --check upstream/main`


tests/cards/promo/Splice.spec.ts
kberg, line 1: This is not worth a test. In the future please consider that when submitting PRs.

kberg:
For updating a log message, this is far, far too much description. I'm not going to read it. Please be mindful of all of that. If you're doing this by AI agent, you're stlil responsible for a reasonable amount of work around a code change.

kberg (approved):

PR #8260, opened 2026-06-21
Add Law Suit help text

rusliksu:
## Summary
- Add a `CardHelpText` entry clarifying who Law Suit may target.
- Leave Law Suit gameplay, player selection, and card models unchanged.

## Demo
Hover the help icon on Law Suit. It now explains that Law Suit may target any player who removed your resources or decreased your production this generation.

## Testing
- `npm run lint:client`
- `git diff --check upstream/main`

## Notes
- No gameplay change; this is a static help text entry following review feedback.

### DIFF
--- a/src/client/cards/CardHelpText.ts
+++ b/src/client/cards/CardHelpText.ts
@@ -4,6 +4,8 @@ import {CardName} from '@/common/cards/CardName';
 export const CARD_HELP_TEXT: Partial<Record<CardName, string>> = {
   [CardName.AIR_RAID]: `Air Raid requires removing 1 floater and stealing exactly 5 M€ from one opponent. These are treated as a package deal, so Air Raid cannot be played unless an opponent has at least 5 M€. If no opponent has 5 M€, the card is unavailable even if you have a floater.`,
 
+  [CardName.LAW_SUIT]: `Law Suit may target any player who removed your resources or decreased your production this generation.`,
+
   [CardName.ROBOTIC_WORKFORCE]:
 `From the unofficial FAQ 1.7:
 


kberg (changes requested):
Nice idea, but instead of all of this add an entry to CardHelpText.

src/client/cards/CardHelpText.ts
```diff
@@ -4,6 +4,8 @@ import {CardName} from '@/common/cards/CardName';
 export const CARD_HELP_TEXT: Partial<Record<CardName, string>> = {
   [CardName.AIR_RAID]: `Air Raid requires removing 1 floater and stealing exactly 5 M€ from one opponent. These are treated as a package deal, so Air Raid cannot be played unless an opponent has at least 5 M€. If no opponent has 5 M€, the card is unavailable even if you have a floater.`,
 
+  [CardName.LAW_SUIT]: `Law Suit may target any player who removed your resources or decreased your production this generation.`,
```
kberg, line 7: Are you done? Because this is just lazy. It repeats the card. I appreciate that you are trying to make contributions, but perhaps you would like to have a conversation. I certainly do not want to continue feeling like I'm just responding to a bot.

rusliksu: You're right. This line doesn't add useful information beyond the card text. I'll close this PR rather than iterate on it; if I revisit the target-selection issue, I'll bring a concrete UX proposal and reproduction first.

rusliksu: One clarification, since I explained the intent poorly: the useful part was not static help text. The user problem is that before playing Law Suit it's hard to know which player is currently eligible without scanning the game log.

If I revisit this, I'd propose a small pre-play hint listing current eligible target names, for example `Eligible targets: Alice, Bob`, with no gameplay change. Does that sound like a useful direction, or would you rather avoid this UI hint?

kberg: > One clarification, since I explained the intent poorly: the useful part was not static help text. The user problem is that before playing Law Suit it's hard to know which player is currently eligible without scanning the game log.
> 
> If I revisit this, I'd propose a small pre-play hint listing current eligible target names, for example `Eligible targets: Alice, Bob`, with no gameplay change. Does that sound like a useful direction, or would you rather avoid this UI hint?

:thinking: that's interesting.

PR #8265, opened 2026-06-23
i18n(ru): add Russian translation for Sterling Vents card

Almaziko:
Hi! A beautiful new map has appeared, and here is its translation!


kberg (approved):
Thank you!

PR #8273, opened 2026-06-25
Mark New Holland as covering an ocean

rusliksu:
## Summary
- Mark New Holland's ocean-upgrade placement as covering the existing ocean tile, same as other ocean upgrade placements.
- This prevents covered ocean placement bonuses from being treated as newly received by effects such as Geological Expedition.

## Reproduction
- Custom server game: https://tm.knightbyte.win/game?id=ge01f744258d4
- Tagir played New Holland on ocean space F5, then Geological Expedition granted the covered ocean's plant bonus.
- New Holland does not grant the covered ocean space bonus, so Geological Expedition should not grant an additional copy of that bonus either.

New Holland log showing the extra plant

Geological Expedition card text

## Testing
- `npx mocha --import=tsx --require tests/testing/setup.ts "tests/cards/pathfinders/GeologicalExpedition.spec.ts" "tests/cards/promo/NewHolland.spec.ts"`
- `npx eslint src/server/deferredActions/PlaceTile.ts tests/cards/pathfinders/GeologicalExpedition.spec.ts`
- `npm run build:server`

### DIFF
--- a/src/server/deferredActions/PlaceTile.ts
+++ b/src/server/deferredActions/PlaceTile.ts
@@ -32,7 +32,7 @@ export class PlaceTile extends DeferredAction<Space> {
     return new SelectSpace(title, availableSpaces)
       .andThen((space: Space) => {
         const tile: Tile = {...this.options.tile};
-        if (this.options.on === 'upgradeable-ocean') {
+        if (this.options.on === 'upgradeable-ocean' || this.options.on === 'upgradeable-ocean-new-holland') {
           tile.covers = space.tile;
         }
         game.addTile(this.player, space, tile);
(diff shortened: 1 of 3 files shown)


kberg (approved):
I rewrote the tests because the test was in the wrong place.

There are two behaviors you want to be sure of: 1) one is that New Holland is marked as covering a tile (which is now tested in New Holland's test) 2) and that Geologocal Expedition doesn't grant something when covering (the rewrittten test.)

This also verifies that you don't get the steel, either.

It also doesn't copy the test's antipattern of popping the waiting for even though there was nothing to pop.

PR #8275, opened 2026-06-25
Make activated Mars Maths survive undo

rusliksu:
## Summary
- Persist per-turn action budget across player serialization.
- Covers extra-action turns when the game is restored from a save, including undo.
- Older saves still fall back to the normal 2-action budget.

## Example
Self-contained reproduction covered by `MarsMaths.spec.ts`:

1. A player takes a first action.
2. The player uses Mars Maths, increasing this turn's action budget from 2 to 4.
3. The player takes another action and the game is serialized.
4. The player takes a third action.
5. Undo restores the serialized game state.

Actual before this change: the third action was rolled back, but the restored player had `actionsTakenThisRound = 2` and `availableActionsThisRound` reset to the default 2, so the server treated the turn as finished.

Expected: undo restores the player with `actionsTakenThisRound = 2` and `availableActionsThisRound = 4`, so it is still that player's turn.

## Testing
- `npx mocha --import=tsx --require tests/testing/setup.ts "tests/Player.spec.ts" "tests/cards/pathfinders/MarsMaths.spec.ts"`
- `npx eslint src/server/Player.ts src/server/SerializedPlayer.ts tests/Player.spec.ts tests/cards/pathfinders/MarsMaths.spec.ts`
- `C:\Users\Ruslan\tm\tools\tm-upstream-pr-check.ps1 -TestCommand 'npx mocha --import=tsx --require tests/testing/setup.ts "tests/Player.spec.ts" "tests/cards/pathfinders/MarsMaths.spec.ts"'`

### DIFF
--- a/tests/Player.spec.ts
+++ b/tests/Player.spec.ts
@@ -311,18 +311,6 @@ describe('Player', () => {
     expect(newPlayer.colonies.usedTradeFleets).eq(100);
   });
 
-  it('preserves the expanded action budget through serialization', () => {
-    const player = new Player('blue', 'blue', false, 0, 'p-blue');
-    Game.newInstance('gameid', [player], player, 'spectatorid');
-    player.actionsTakenThisRound = 2;
-    player.availableActionsThisRound = 4;
-
-    const newPlayer = Player.deserialize(player.serialize());
-
-    expect(newPlayer.actionsTakenThisRound).eq(2);
-    expect(newPlayer.availableActionsThisRound).eq(4);
-  });
-
   it('pulls self replicating robots target cards', () => {
     const player = new Player('blue', 'blue', false, 0, 'p-blue');
     expect(player.getSelfReplicatingRobotsTargetCards()).is.empty;
--- a/tests/Player.spec.ts
+++ b/tests/Player.spec.ts
@@ -311,6 +311,18 @@ describe('Player', () => {
     expect(newPlayer.colonies.usedTradeFleets).eq(100);
   });
 
+  it('preserves the expanded action budget through serialization', () => {
+    const player = new Player('blue', 'blue', false, 0, 'p-blue');
+    Game.newInstance('gameid', [player], player, 'spectatorid');
+    player.actionsTakenThisRound = 2;
+    player.availableActionsThisRound = 4;
+
+    const newPlayer = Player.deserialize(player.serialize());
+
+    expect(newPlayer.actionsTakenThisRound).eq(2);
+    expect(newPlayer.availableActionsThisRound).eq(4);
+  });
+
   it('pulls self replicating robots target cards', () => {
     const player = new Player('blue', 'blue', false, 0, 'p-blue');
     expect(player.getSelfReplicatingRobotsTargetCards()).is.empty;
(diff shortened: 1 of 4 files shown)


tests/cards/pathfinders/MarsMaths.spec.ts
```diff
@@ -22,6 +23,17 @@ describe('MarsMaths', () => {
     expect(player.availableActionsThisRound).eq(previousActions + 2);
   });
 
+  it('preserves additional actions through serialization', () => {
```
kberg: This is a useless test and even still this tests Player, not Mars Maths. If there is a behavior that was broken but now is fixed, test that.

kberg (changes requested):
Please explain in your own words the issue that you found and how this fixes the issue. Please write it in your own language and don't worry about translating it.

kberg:
> Please explain in your own words the issue that you found and how this fixes the issue. Please write it in your own language and don't worry about translating it.

Please excuse me. I did not see the reproduction steps.

It looks as though you have the makings of a test. Keep in mind that Undo is now best effort only. But I'm happy to fix bugs.

rusliksu:
The issue was not Mars Maths itself, but the player runtime state after restore/undo.

Mars Maths temporarily raises `availableActionsThisRound` from 2 to 4. Before this PR, only `actionsTakenThisRound` was serialized. After restore, a player could come back with `actionsTakenThisRound = 2` while `availableActionsThisRound` fell back to the default 2, so the server treated the turn as finished. This change serializes and restores `availableActionsThisRound`; older saves without that field still fall back to 2.

I also moved the regression test from `MarsMaths.spec.ts` to `Player.spec.ts`, so it now tests player serialization behavior directly.

tests/cards/pathfinders/MarsMaths.spec.ts
rusliksu: Done. I removed the Player serialization check from `MarsMaths.spec.ts` and moved the regression to `Player.spec.ts`, where it verifies that `availableActionsThisRound` survives `Player.serialize()` / `Player.deserialize()` alongside `actionsTakenThisRound`.

kberg: It's still a useless test. Can you remove it please?

kberg:
I've added the necessary tests. Take a look.

tests/Player.spec.ts
```diff
@@ -311,6 +311,18 @@ describe('Player', () => {
     expect(newPlayer.colonies.usedTradeFleets).eq(100);
   });
 
+  it('preserves the expanded action budget through serialization', () => {
```
kberg: Sorry, I still think this is a useless test, no matter where you put it.

Do you know the saying "test until fear turns to boredom?" AIs don't know what that means so they'll just add tests that verify the micro-behavior not worrying whether a human being even cares about that level of detail. Please remove this test.

kberg (changes requested):
I gave you the necessary tests. Once the final change is made this can go in.

rusliksu:
Removed the extra player serialization test and updated the PR description, so it no longer depends on the old custom-server game link

Thanks for adding the Mars Maths behavior tests

kberg:
> Removed the extra player serialization test and updated the PR description, so it no longer depends on the old custom-server game link
> 
> Thanks for adding the Mars Maths behavior tests

The tests should have been there before, but that card was written at a time when it wasn't too obvious how to do it smartly.

kberg (approved):
Thank you. This is getting a little easier.

PR #8281, opened 2026-06-28
i18n(de): add German translation

Borbarad2:
add new translation according to F12 debug screen (except NO DELTA PROJECT AT ALL)


kberg (approved):

Borbarad2:
Thank you.

PR #8296, opened 2026-07-03
Skip Floating Trade Hub target prompt for single card

rusliksu:
## Summary
- Auto-select the only available floater card when Floating Trade Hub adds 2 floaters.
- Keep the multi-target flow unchanged, and keep the convert-floaters branch available when Floating Trade Hub already has floaters.

## Demo
Scenario: play Floating Trade Hub when it is the only card that can hold floaters. Using the action now adds 2 floaters immediately instead of asking which card should receive them.

If another floater card is in play, the existing target-card prompt still appears.

## Testing
- `npx mocha --import=tsx --require tests/testing/setup.ts tests/cards/prelude2/FloatingTradeHub.spec.ts`
- `npx eslint src/server/cards/prelude2/FloatingTradeHub.ts tests/cards/prelude2/FloatingTradeHub.spec.ts`
- `npm run make:static && npm run build:tests`
- proof PR CI passed: build-linux, build-windows, build-docker

## Notes
- No resource outcome change; this only skips a redundant prompt when there is exactly one legal target.

### DIFF
--- a/src/server/cards/prelude2/FloatingTradeHub.ts
+++ b/src/server/cards/prelude2/FloatingTradeHub.ts
@@ -37,11 +37,11 @@ export class FloatingTradeHub extends PreludeCard implements IActionCard {
 
   public action(player: IPlayer) {
     const floaterCards = player.getResourceCards(CardResource.FLOATER);
-    if (this.resourceCount === 0 && floaterCards.length === 1) {
-      this.add2Floaters(player, floaterCards[0]);
-      return undefined;
+    const floatingTradeHub = floaterCards.find((card) => card.name === this.name);
+    if (this.resourceCount === 0 && floaterCards.length === 1 && floatingTradeHub !== undefined) {
+      return this.add2Floaters(player, floatingTradeHub);
     }
-    const add2Floaters = this.add2FloatersOption(player, floaterCards);
+    const add2Floaters = this.add2FloatersOption(player, floaterCards, floatingTradeHub);
     const selectResource = new SelectResource('Select resource to gain');
     const selectAmount = new SelectAmount('Select amount of floaters to remove', undefined, 1, this.resourceCount, true);
     const removeFloaters = new AndOptions(selectAmount, selectResource)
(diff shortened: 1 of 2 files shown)


tests/cards/prelude2/FloatingTradeHub.spec.ts
```diff
@@ -64,4 +63,15 @@ describe('FloatingTradeHub', () => {
     expect(player.stock.asUnits()).deep.eq(Units.of({plants: 4}));
     expect(card.resourceCount).to.eq(1);
   });
+
+  it('Act - add resources branch autoselects the only target card', () => {
+    card.resourceCount = 5;
+
+    const orOptions = cast(card.action(player), OrOptions);
+    const addFloaters = cast(orOptions.options[0], SelectOption);
```
kberg: s/addFloaters/selectOption

tests/cards/prelude2/FloatingTradeHub.spec.ts
kberg, line 1: Take a minute to read this test. What feature is no longer being tested?

src/server/cards/prelude2/FloatingTradeHub.ts
```diff
@@ -35,10 +36,12 @@ export class FloatingTradeHub extends PreludeCard implements IActionCard {
   }
 
   public action(player: IPlayer) {
-    const add2Floaters = new SelectCard('Select card to gain 2 floaters', undefined, player.getResourceCards(CardResource.FLOATER)).andThen(([card]) => {
-      player.addResourceTo(card, {qty: 2, log: true});
+    const floaterCards = player.getResourceCards(CardResource.FLOATER);
+    if (this.resourceCount === 0 && floaterCards.length === 1) {
```
kberg: You are inferring that floaterCards[0].name is this.name, but I would be more explicit about it, both here and in add2FloatersOption.

src/server/cards/prelude2/FloatingTradeHub.ts
```diff
@@ -35,10 +36,12 @@ export class FloatingTradeHub extends PreludeCard implements IActionCard {
...
+      this.add2Floaters(player, floaterCards[0]);
```
kberg: I don't think this line actually works, since it doesn't return anything.

kberg (commented):
Thank you for the contribution. There are still a couple of issues to resolve.

rusliksu:
I updated the PR based on your comments

tests/cards/prelude2/FloatingTradeHub.spec.ts
```diff
@@ -64,4 +65,29 @@ describe('FloatingTradeHub', () => {
     expect(player.stock.asUnits()).deep.eq(Units.of({plants: 4}));
     expect(card.resourceCount).to.eq(1);
   });
+
+  it('Act - add resources branch autoselects the only target card', () => {
```
kberg: I'm reading this some more -- help me understand: how are these tests different from the two above?

tests/cards/prelude2/FloatingTradeHub.spec.ts
```diff
@@ -24,70 +24,72 @@ describe('FloatingTradeHub', () => {
...
+      selectCard.cb([card]);
 
-    expect(card.resourceCount).eq(2);
+      expect(card.resourceCount).eq(2);
+    });
   });
 
-  it('Act - select resource', () => {
-    card.resourceCount = 5;
+  describe('when converting floaters is available', () => {
```
kberg: We only really use subdescribes when setup varies greatly and then rarely, not to group by functional behavior.

rusliksu: Done, I removed the nested describes

kberg (approved):
I made minor edits. This is good to go. Thank you!

kberg (approved):
Thank you

PR #8307, opened 2026-07-10
Skip Floating Refinery target prompt for single card

rusliksu:
## Summary

- Skip the second target-selection prompt when only one card can spend two floaters.
- Keep the card chooser when multiple eligible targets exist.

## Demo

With Floating Refinery, choose **Remove floaters** while exactly one card has at least two floaters. The two floaters are removed immediately and the player gains 1 titanium and 2 M€.

When multiple cards can pay, the existing target-card prompt is unchanged.

## Testing

- `npx mocha --import=tsx --require tests/testing/setup.ts tests/cards/prelude2/FloatingRefinery.spec.ts`
- `npm run lint:server`
- `npm run lint:client`
- `npm run build`

## Notes

- No resource or payment behavior changes; this only skips a redundant prompt.

### DIFF
--- a/tests/cards/prelude2/FloatingRefinery.spec.ts
+++ b/tests/cards/prelude2/FloatingRefinery.spec.ts
@@ -59,6 +59,7 @@ describe('FloatingRefinery', () => {
 
   it('act - removes from another card when it is the only eligible target', () => {
     card.resourceCount = 1;
+    // floater1 is the only eligible card.
     floater1.resourceCount = 2;
     floater2.resourceCount = 1;
     other.resourceCount = 1;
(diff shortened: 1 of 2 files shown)


tests/cards/prelude2/FloatingRefinery.spec.ts
```diff
@@ -46,19 +45,35 @@ describe('FloatingRefinery', () => {
...
+    cast(orOptions.options[0].cb(), undefined);
+
     expect(player.stock.titanium).to.eq(1);
     expect(player.stock.megacredits).to.eq(2);
     expect(card.resourceCount).to.eq(1);
   });
 
+  it('act - removes from another card when it is the only eligible target', () => {
+    card.resourceCount = 1;
+    floater1.resourceCount = 2;
```
kberg, line 63: Just add a comment saying that `floater1` is the only card that can be acted upon. This is the kind of thing that AI will really overdo. If it's more than 7 words it's too much. :D

kberg (approved):

kberg:
Thanks

PR #8311, opened 2026-07-12
Skip single-card resource target prompts

rusliksu:
## Summary
- Add `SelectCard.maybeConvertToSelectOption` for one eligible card.
- Use it in three missing flows and four existing equivalents.

## Demo
One eligible card is shown as a direct option. Multiple cards still open the normal selector.

## Testing
- `npm run build:tests`
- 38 focused tests for single and multiple targets
- `npm run lint:server`
- `npm run lint:client`
- `npm run build`


kberg:
Hi -- how many cards would you like to try to do this with? Because I might want to just come up with a way to shorthand this for less code, and with clear code.

rusliksu:
I found three live cases that still needed this: Ecotec, the Venus alternate-track wild resource bonus, and the Underworld any-resource reward. I added a small `selectCardOrOption` shorthand and also used it to replace the existing manual branches in Imported Hydrogen, Large Convoy, Local Heat Trapping, Extreme Cold Fungus, and Bio-Printing Facility.

rusliksu:
Hi, this is ready for another look if you have time

kberg:
I haven't forgotten this, but I'm looking to make this much simpler. I'm playing with an alternate mechanism. Hope that's OK.

rusliksu:
Sure, sounds good.

src/server/cards/base/ExtremeColdFungus.ts
```diff
@@ -51,27 +51,16 @@ export class ExtremeColdFungus extends Card implements IActionCard, IProjectCard
...
-      return new OrOptions(
-        new SelectOption(message('Add ${0} microbes to ${1}', (b) => b.number(2).card(targetCard)), 'Add microbes').andThen(() => {
-          player.addResourceTo(targetCard, {qty: 2, log: true});
-          return undefined;
-        }),
-        gainPlantOption,
-      );
-    }
-
     return new OrOptions(
```
kberg, line 54: OK, thank you for working on this. I don't ... love it, but it's made me think.

Here's what I suggest.

Add a method to SelectCard:

```
publicMaybeConvertToSelectOption(title: string | Message): SelectCard<T> | SelectOption {
  if (this.cards.length !== 1) {
    return this;
  }
  return new SelectOption(title).andThen(() => {
    this.cb(this.cards[0]);
  });
}
```

I don't like the method name. I am not sure I like the use of a suffix type method. But it is a little less confusing. And I believe it the commit will be simpler.

Honestly I think the biggest problem is that we rely too much on "new" here. But I'm not ready for a bigger refactoring / churn. It doesn't merit it yet.

kberg (changes requested):

src/server/cards/base/ExtremeColdFungus.ts
rusliksu: updated this using a SelectCard method. Also merged current main

rusliksu: Bio-Printing Facility conflicted with main, so I kept new ActionCard version because it already autoselects a single target

kberg: Yeah, I think this is better. Easier to read.

src/server/cards/base/ImportedHydrogen.ts
```diff
@@ -58,34 +58,22 @@ export class ImportedHydrogen extends Card implements IProjectCard {
     const gainPlantsOption = new SelectOption('Gain 3 plants', 'Gain plants').andThen(gainPlants);
```
kberg, line 58: The code can be further simplified by finding a way to handle zero card counts. That will only reduce code further, and we can deal with that later. Don't deal with it now, please.

src/server/deferredActions/GainAnyResourceButScienceDeferred.ts
```diff
@@ -19,11 +20,12 @@ export class GainAnyResourceButScienceDeferred extends DeferredAction {
 
     const cards = this.player.getResourceCards(undefined).filter((card) => card.resourceType !== CardResource.SCIENCE);
     if (cards.length > 0) {
-      orOptions.options.push(new SelectCard('Gain 1 card resource', undefined, cards)
+      orOptions.options.push(new SelectCard('Gain 1 card resource', 'Add resource', cards)
```
kberg: Why did this change?

tests/cards/prelude2/Ecotec.spec.ts
kberg, line 1: Sorry, I just wrote tests for all cards without tests, which included Ecotec. Please sync and resolve.

tests/inputs/SelectCard.spec.ts
```diff
@@ -37,6 +38,27 @@ describe('SelectCard', () => {
     expect(selected).deep.eq([ioMiningIndustries]);
   });
 
+  it('Converts a single card to an option', () => {
+    const selectCard = new SelectCard('Select card', 'Add resource', [aquiferPumping])
+      .andThen(cb);
+
+    const result = selectCard.maybeConvertToSelectOption('Add resource to Aquifer Pumping');
+
+    expect(result).instanceOf(SelectOption);
```
kberg: Don't use expect.instanceof. See how existing tests use cast instead.

tests/inputs/SelectCard.spec.ts
```diff
@@ -37,6 +38,27 @@ describe('SelectCard', () => {
...
+    expect(result.buttonLabel).eq('Add resource');
+    result.process({type: 'option'});
```
kberg: result.cb() is fine.

tests/inputs/SelectCard.spec.ts
```diff
@@ -37,6 +38,27 @@ describe('SelectCard', () => {
     expect(selected).deep.eq([ioMiningIndustries]);
   });
 
+  it('Converts a single card to an option', () => {
+    const selectCard = new SelectCard('Select card', 'Add resource', [aquiferPumping])
+      .andThen(cb);
+
+    const result = selectCard.maybeConvertToSelectOption('Add resource to Aquifer Pumping');
```
kberg: Merge this with the line above. Call it `input` and then with the `cast` I mentioned below, call the new one `selectOption`.

tests/inputs/SelectCard.spec.ts
```diff
@@ -37,6 +38,27 @@ describe('SelectCard', () => {
...
+
+    expect(result).instanceOf(SelectOption);
+    expect(result.buttonLabel).eq('Add resource');
+    result.process({type: 'option'});
+    expect(selected).deep.eq([aquiferPumping]);
+  });
+
+  it('Keeps a card selection for multiple cards', () => {
+    const selectCard = new SelectCard('Select card', 'Add resource', [aquiferPumping, ioMiningIndustries])
```
kberg: Same as above, and include cast(... selectCard)

tests/deferredActions/GainAnyResourceButScienceDeferred.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+import {Tardigrades} from '../../src/server/cards/base/Tardigrades';
+import {GainAnyResourceButScienceDeferred} from '../../src/server/deferredActions/GainAnyResourceButScienceDeferred';
+import {OrOptions} from '../../src/server/inputs/OrOptions';
+import {SelectOption} from '../../src/server/inputs/SelectOption';
+import {cast} from '../../src/common/utils/utils';
+import {testGame} from '../TestGame';
+
+describe('GainAnyResourceButScienceDeferred', () => {
+  it('offers the only resource card directly', () => {
+    const [/* game */, player] = testGame(1, {underworldExpansion: true});
```
kberg: Why is underworld part of this test

tests/deferredActions/GainAnyResourceButScienceDeferred.spec.ts
```diff
@@ -0,0 +1,21 @@
+import {expect} from 'chai';
+import {Tardigrades} from '../../src/server/cards/base/Tardigrades';
+import {GainAnyResourceButScienceDeferred} from '../../src/server/deferredActions/GainAnyResourceButScienceDeferred';
+import {OrOptions} from '../../src/server/inputs/OrOptions';
+import {SelectOption} from '../../src/server/inputs/SelectOption';
+import {cast} from '../../src/common/utils/utils';
+import {testGame} from '../TestGame';
+
+describe('GainAnyResourceButScienceDeferred', () => {
+  it('offers the only resource card directly', () => {
```
kberg, line 30: I appreciate this test. Can you add another one above it that tests the card in its more natural case? Remember to include a case with a science, to see it is excluded.

tests/venusNext/GrantVenusAltTrackBonusDeferred.spec.ts
```diff
@@ -70,4 +71,15 @@ describe('GrantVenusAltTrackBonusDeferred', () => {
     // The second option is the standard resource section.
     expect(input.options[1]).instanceof(AndOptions);
   });
+
+  it('offers the only resource card directly', () => {
+    const card = new Tardigrades();
+    player.playedCards.push(card);
+
+    const input = cast(new GrantVenusAltTrackBonusDeferred(player, 0, true).execute(), OrOptions);
```
kberg: rename input to orOptions, since you have cast its type

kberg (commented):
Seems like alot but I think it's not so much

tests/cards/prelude2/Ecotec.spec.ts
rusliksu: synced main and kept the updated Ecotec tests

tests/deferredActions/GainAnyResourceButScienceDeferred.spec.ts
rusliksu: added this case including a science card

kberg (approved):
This is a nice incremental improvement. Thank you.

PR #8319, opened 2026-07-17
Create promo-cards.json

diegoequisde:
Add new file promo-cards.json to translate to spanish.


kberg (approved):
This is good, thank you!

I did make a small edit. The app doesn't allow you to repeat the English translation. This is because some translations just repeated all the English with no real value. By replacing "Astra Mechanica" with "" it says "Just use the English value" and not "this string should be empty."

PR #8322, opened 2026-07-18
Stabilize hand card drag insertion

rusliksu:
## Summary
- Insert dragged hand cards before or after the hovered card based on pointer position.
- Keep the experimental point-and-click reorder UI unchanged.

## Demo
Drag a card over either half of another card; it lands on the corresponding side.

## Testing
- SortableCards component tests
- `npm run lint:server`
- `npm run lint:client`
- `npm run build`

### DIFF
--- a/tests/client/components/SortableCards.spec.ts
+++ b/tests/client/components/SortableCards.spec.ts
@@ -35,10 +35,9 @@ describe('SortableCards', () => {
     expect(cards[0].props().card.name).to.eq(CardName.ANTS);
     expect(cards[1].props().card.name).to.eq(CardName.CARTEL);
     const draggers = sortable.findAll('[draggable=true]');
+    draggers[0].element.getBoundingClientRect = () => ({left: 0, width: 200} as DOMRect);
     await draggers[1].trigger('dragstart');
-    await sortable.vm.$nextTick();
-    const droppers = sortable.findAll('.drop-target');
-    await droppers[0].trigger('dragover');
+    await draggers[0].trigger('dragover', {clientX: 50});
     await draggers[1].trigger('dragend');
     cards = sortable.findAllComponents({
       name: 'Card',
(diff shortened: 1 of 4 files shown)


kberg:
I don't understand what this change does. Also, if I'm reading it correctly, you wrote a drag-and-drop feature, and your testing is just running the automated tests. Is that correct? Because I would expect you to manually verify the change.

rusliksu:
The existing drag-and-drop on main uses 10px drop zones between cards; this change uses the whole card, with the left and right halves inserting before or after it, while Reorder Cards is unchanged. I tested both modes manually, and the same seeded game is available on both servers for comparison:
Main: https://terraforming-mars.herokuapp.com/player?id=pf4caecb75189
PR preview: https://preview.tm.knightbyte.win/player?id=pa98d222ec809

tests/client/components/SortableCards.spec.ts
```diff
@@ -35,10 +35,9 @@ describe('SortableCards', () => {
     expect(cards[0].props().card.name).to.eq(CardName.ANTS);
     expect(cards[1].props().card.name).to.eq(CardName.CARTEL);
     const draggers = sortable.findAll('[draggable=true]');
+    draggers[0].element.getBoundingClientRect = () => ({left: 0, width: 200} as DOMRect);
```
kberg: Why is this necessary now? This will not be clear to someone trying to understand what is happening.

tests/client/components/SortableCards.spec.ts
```diff
@@ -79,23 +78,22 @@ describe('SortableCards', () => {
...
     await draggers[0].trigger('dragstart');
-    await sortable.vm.$nextTick();
-    const droppers = sortable.findAll('.drop-target');
-    await droppers[2].trigger('dragover');
+    await draggers[2].trigger('dragover', {clientX: 150});
     await draggers[0].trigger('dragend');
     cards = sortable.findAllComponents({
       name: 'Card',
     });
     expect(cards[0].props().card.name).to.eq(CardName.ANTS);
```
kberg: While this test is not broken, it tests something different. Why isn't this testing the prior order?

tests/client/components/SortableCards.spec.ts
kberg, line 1: This test used to be easier to understand because it was dragging and dropping over cards with a known position in the order. Now it's using magic numbers in the form of pixels.

kberg (changes requested):
Just a reminder: I want responses from you, and not the AI.

tests/client/components/SortableCards.spec.ts
rusliksu: this is needed because jsdom doesn't calculate element bounds

tests/client/components/SortableCards.spec.ts
rusliksu: restored previous card order

tests/client/components/SortableCards.spec.ts
rusliksu: replaced pixel values into before/after helper

tests/client/components/SortableCards.spec.ts
```diff
@@ -1,10 +1,24 @@
...
 import {CardName} from '@/common/cards/CardName';
 import SortableCards from '@/client/components/SortableCards.vue';
 import {FakeLocalStorage} from './FakeLocalStorage';
 
+type DropPosition = 'before' | 'after';
+
+async function dragCard(sortable: VueWrapper<InstanceType<typeof SortableCards>>, sourceIndex: number, targetIndex: number, position: DropPosition) {
+  const draggers = sortable.findAll('[draggable=true]');
+  const target = draggers[targetIndex];
+  const bounds = {left: 0, width: 2};
```
kberg: These numbers are still magic. Why are  they 0, 2 instead of 0, 200?

Do you understand what this test is doing?

rusliksu: Yes, code compares clientX with the middel of the card. I used width 2 only to make middle 1? but i see, this leaves unexplained test values

kberg:
I wonder why it's dragging a little bit more: 
image

kberg:
I've updated the test. If you can fix the visual error, this can go in.

rusliksu:
Fixed

kberg (approved):

kberg:
This is a good change.

PR #8323, opened 2026-07-18
Keep timer stop time monotonic

rusliksu:
## Summary
- Keep the shared timer stop timestamp monotonic when deserializing games.
- Prevent loading an older saved game from moving subsequent timer starts backwards.

## Demo
1. Deserialize a timer with `lastStoppedAt = 5000`.
2. Deserialize another timer with `lastStoppedAt = 1000`.
3. Start the first timer; its `startedAt` remains `5000`.

## Testing
- `npx mocha --import=tsx --require tests/testing/setup.ts tests/common/Timer.spec.ts`
- `npm run lint:server`
- `npm run lint:client`
- `npm run build`

## Notes
- No UI or gameplay-rule change.

### DIFF
--- a/src/common/Timer.ts
+++ b/src/common/Timer.ts
@@ -68,8 +68,7 @@ export class Timer {
     timer.running = d.running;
     timer.afterFirstAction = d.afterFirstAction;
 
-    // Should this be `Math.max(Timer.lastStoppedAt, d.lastStoppedAt)`?
-    Timer.lastStoppedAt = d.lastStoppedAt;
+    Timer.lastStoppedAt = Math.max(Timer.lastStoppedAt, d.lastStoppedAt);
     return timer;
   }
 
--- a/src/common/Timer.ts
+++ b/src/common/Timer.ts
@@ -68,6 +68,7 @@ export class Timer {
     timer.running = d.running;
     timer.afterFirstAction = d.afterFirstAction;
 
+    // Never move the shared timer backwards.
     Timer.lastStoppedAt = Math.max(Timer.lastStoppedAt, d.lastStoppedAt);
     return timer;
   }
@@ -140,4 +141,3 @@ export class Timer {
     return elapsedDate.toISOString().substr(14, 5);
   }
 }
-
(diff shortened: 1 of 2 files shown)


kberg:
Can you explain to me the bug you are fixing? I don't quite follow.

rusliksu:
`Timer.lastStoppedAt` is a shared static value across all loaded games. Loading an older save could overwrite it with an earlier timestamp, effectively moving the shared clock backwards.

For example, if Game A updates `lastStoppedAt` to **18:00**, then loading Game B with a saved value of **17:30** would replace it with the older timestamp. The next timer update in Game A would measure elapsed time from **17:30** instead of **18:00**, giving the player extra time.

Using `Math.max` preserves the newest timestamp, so loading an older save cannot move the shared timer backwards while still allowing newer timestamps to be restored.

kberg:
> `Timer.lastStoppedAt` is a shared static value across all loaded games. Loading an older save could overwrite it with an earlier timestamp, effectively moving the shared clock backwards.
> 
> For example, if Game A updates `lastStoppedAt` to **18:00**, then loading Game B with a saved value of **17:30** would replace it with the older timestamp. The next timer update in Game A would measure elapsed time from **17:30** instead of **18:00**, giving the player extra time.
> 
> Using `Math.max` preserves the newest timestamp, so loading an older save cannot move the shared timer backwards while still allowing newer timestamps to be restored.

This reads like the AI wrote it.

src/common/Timer.ts
```diff
@@ -68,8 +68,7 @@ export class Timer {
     timer.running = d.running;
     timer.afterFirstAction = d.afterFirstAction;
 
-    // Should this be `Math.max(Timer.lastStoppedAt, d.lastStoppedAt)`?
-    Timer.lastStoppedAt = d.lastStoppedAt;
+    Timer.lastStoppedAt = Math.max(Timer.lastStoppedAt, d.lastStoppedAt);
```
kberg, line 72: I understand it now. Please Add a short comment explaining why this is correct. One sentence max. Then I can merge this.

rusliksu: Added a short comment

kberg (approved):
This makes alot of sense now. Thank you!

PR #8206, opened 2026-06-09
Add MarsBot bonus deck

ignac8:
One more for the MarsBot series (#8001). This one builds on #8203 so the diff shows that commit too, only MarsBotBonusDeck.ts and its test are new here. Will rebase once #8203 lands

Adds the bonus deck itself: shuffled draw pile, discard pile, reshuffle when the draw pile runs out, and destroyed cards staying out of the game for good. The expansion variants of the deck come later with their expansions

I'm out for the next week so replies will be slow

PR #8342, opened 2026-07-28
Trim bonus deck tests covered by Deck.spec

ignac8:
Follow-up to #8206. The draw, discard and reshuffle tests exercised plain Deck behavior that Deck.spec.ts already covers. Kept the MarsBot part: the createBase contents and the deserialize round trip.


kberg (approved):

kberg:
Thanks

PR #8353, opened 2026-08-03
Fix Automa card names and rename IMarsBot members

ignac8:
Follow-ups to #8206 and #8207 from review comments on both, plus the `MarsBotBoard` rename you asked about on #8209.

- The eight `Automa: ...` card names rendered as "Automa" in the client, which does `name.split(':')[0]` in CardTitle, Card.vue and LogMessageComponent. Moved them to the suffix form used by `Deimos Down:promo` and the Ares cards.
- Removed the bonus deck's flow comment.
- `IMarsBot.mcSupply` and `floaterCount` are now `megacredits` and `floaters`. `MarsBotModel`'s field renamed to match `PlayerModel`.
- Reworded the hook comments you asked about: `onVenusRaised`, `interceptGlobalParameterRaise`, and what an empty deck means.
- `MarsBotBoard` is now `MarsBotTracks`. Its `tracks` array became `all`, so callers read `bot.tracks.all[i]` instead of `tracks.tracks[i]`.


src/common/cards/CardName.ts
```diff
@@ -611,14 +611,14 @@ export enum CardName {
...
   // Automa
-  AUTOMA_METEOR_SHOWER = 'Automa: Meteor Shower',
-  AUTOMA_INVASIVE_SPECIES = 'Automa: Invasive Species',
-  AUTOMA_RESEARCH_AND_DEVELOPMENT = 'Automa: Research and Development',
-  AUTOMA_OVERACHIEVEMENT = 'Automa: Overachievement',
-  AUTOMA_EXPEDITED_CONSTRUCTION = 'Automa: Expedited Construction',
-  AUTOMA_LOBBYISTS = 'Automa: Lobbyists',
-  AUTOMA_LOCAL_NEURAL_INSTANCE = 'Automa: Local Neural Instance',
-  AUTOMA_CORPORATE_COMPETITION = 'Automa: Corporate Competition',
+  AUTOMA_METEOR_SHOWER = 'Meteor Shower:automa',
```
kberg, line 617: Are these cards in conflict with others in the cardname set?

src/server/automa/MarsBotCorpTypes.ts
```diff
@@ -86,11 +87,14 @@ export interface IMarsBot {
   raiseTR(steps: number): void;
   advanceTrack(trackIndex: number): void;
 
-  /** Draws the top project card and resolves it as a bot action. False when the deck is empty. */
+  /**
+   * Draws the top project card and resolves it as a bot action. The deck reshuffles its
+   * discard pile when it runs out, so false means both piles were empty.
+   */
   drawAndResolveProjectCard(): boolean;
```
kberg: What happens when both piles are empty?

src/server/automa/MarsBotTracks.ts
```diff
@@ -53,13 +53,13 @@ export class MarsBotTrack {
   }
 }
 
-/** The MarsBot board with tracks. Handles tag-to-track mapping and track state. */
-export class MarsBotBoard {
-  public readonly tracks: ReadonlyArray<MarsBotTrack>;
+/** MarsBot's tracks. Handles tag-to-track mapping and track state. */
+export class MarsBotTracks {
+  public readonly all: ReadonlyArray<MarsBotTrack>;
```
kberg: Why not `tracks`? That's a good name

src/common/cards/CardName.ts
ignac8: No. None of the eight collide with another value in the enum, and no display name collides either. The suffix is there because the old "Automa: X" form rendered as "Automa" in the client, which keeps what precedes the colon. It also matches the existing groups like Capital:ares and Special Design:proxy. Happy to drop it if you would rather these read as plain names.

src/server/automa/MarsBotCorpTypes.ts
ignac8: Nothing happens, and no caller noticed: all five call sites ignore the boolean. Dropped it. The three methods return void and take the maybe prefix per STYLE.md, and the comment says the call does nothing when both piles are empty.

src/server/automa/MarsBotTracks.ts
ignac8: Because the class is MarsBotTracks, so the field would read bot.tracks.tracks[i] at every call site, 256 of them. all keeps it at bot.tracks.all[i]. If you prefer tracks anyway, I will rename it.

kberg: Ugh, I'm reading more carefully. Sorry for this. I see now why you called it board. MarsBotBoard is fine, just don't call it `board`, but refer to it as marsBotBoard as a variable. please.

src/server/automa/MarsBotTracks.ts
```diff
@@ -53,13 +53,13 @@ export class MarsBotTrack {
...
   private readonly tagToTrack: Map<Tag, number>;
 
   constructor(public readonly data: ReadonlyArray<TrackDefinition>) {
```
kberg: s/data/definitions

src/server/automa/MarsBotTracks.ts
```diff
@@ -53,13 +53,13 @@ export class MarsBotTrack {
...
-    this.tracks = data.map((def) => new MarsBotTrack(def));
+    this.all = data.map((def) => new MarsBotTrack(def));
     this.tagToTrack = new Map();
```
kberg: this.tagToTrack: Record<Tag, number> = {};

etc.

More efficient, less memory, faster.

src/server/automa/MarsBotTracks.ts
kberg: If tagToTrack is public, this these can be removed.

src/server/automa/MarsBotTracks.ts
ignac8: Renamed back to MarsBotBoard, the array is tracks again, and variables say marsBotBoard.

src/server/automa/MarsBotTracks.ts
ignac8: Renamed.

src/server/automa/MarsBotTracks.ts
ignac8: Done, as Partial<Record<Tag, number>>. A bare Record wants every tag as a key.

src/server/automa/MarsBotTracks.ts
ignac8: Public now, both accessors removed.

src/common/cards/CardName.ts
kberg: Umm... thanks, I think, no, this is OK, thank you.

kberg (approved):
Thanks

PR #8358, opened 2026-08-04
Share the first and second place ranking walk

ignac8:
Election, Revolution and the award scoring each carry their own copy of the same walk: sort players by score, prize for first place, prize for second, ties are friendly. This pulls the grouping into `rankedTiers` in utils and refits the three sites. The old loops also recomputed each score on every comparison, the helper calls the scorer once per player.

Adds the tests the walk was missing: a tie for first in both global events, Election with two players, and a small `calculateVictoryPoints` spec proving two player games award no second place, which nothing covered before.

Follows kberg's note in #8209 that we rank items by score in more than one place.

### DIFF
--- a/src/common/utils/utils.ts
+++ b/src/common/utils/utils.ts
@@ -134,6 +134,27 @@ export function zip<S, T>(first: ReadonlyArray<S>, second: ReadonlyArray<T>): Ar
 }
 
 /**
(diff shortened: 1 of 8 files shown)


kberg:
I appreciate the excitement but we have a lot of code to get through, I'd
like to avoid this side quest.

PR #8360, opened 2026-08-08
Preserve log view across game refresh

rusliksu:
## Summary

Keeps the log where you were reading after game updates. The arrow returns to the latest logs.

## Demo

| Desktop | Mobile |
| --- | --- |
| Desktop | Mobile |

## Testing

- LogPanel tests
- lint and build

No gameplay changes.

### DIFF
--- a/src/client/components/logpanel/LogPanel.vue
+++ b/src/client/components/logpanel/LogPanel.vue
@@ -25,7 +25,11 @@
         title="Latest logs"
         data-test="log-latest"
         @click="showLatestLogs"
-      ><span class="log-latest-button-icon" aria-hidden="true"></span></button>
+      >
+        <svg class="log-latest-button-icon" viewBox="0 0 24 24" fill="none" aria-hidden="true" focusable="false">
+          <path d="M12 5v14M19 12l-7 7-7-7"/>
+        </svg>
+      </button>
       <div class='debugid'>(debugid {{step}})</div>
     </div>
     <CardPanel v-if="selectedMessage !== undefined" :message="selectedMessage" :players="players" @hide="selectedMessage = undefined"/>
--- a/src/client/components/logpanel/LogPanel.vue
+++ b/src/client/components/logpanel/LogPanel.vue
@@ -18,16 +18,16 @@
           <LogMessageComponent v-for="(message, index) in messages" :key="index" :message="message" :viewModel="viewModel" @click="messageClicked(message)" @spaceClicked="spaceClicked"/>
         </ul>
       </div>
+      <button
+        type="button"
+        class="log-latest-button"
+        aria-label="Latest logs"
+        title="Latest logs"
+        data-test="log-latest"
+        @click="showLatestLogs"
+      ><span class="log-latest-button-icon" aria-hidden="true"></span></button>
       <div class='debugid'>(debugid {{step}})</div>
     </div>
-    <button
-      type="button"
-      class="log-latest-button"
-      aria-label="Latest logs"
-      title="Latest logs"
-      data-test="log-latest"
-      @click="showLatestLogs"
-    >&#x2193;</button>
     <CardPanel v-if="selectedMessage !== undefined" :message="selectedMessage" :players="players" @hide="selectedMessage = undefined"/>
   </div>
 </template>
--- a/src/client/components/logpanel/LogPanel.vue
+++ b/src/client/components/logpanel/LogPanel.vue
@@ -13,12 +13,13 @@
       <span class="label-additional" v-if="players.length === 1"><span :class="lastGenerationClass" v-i18n>of {{lastSoloGeneration}}</span></span>
     </div>
     <div class="panel log-panel">
-      <div id="logpanel-scrollable" class="panel-body">
+      <div id="logpanel-scrollable" class="panel-body" @scroll="updateScrollState">
         <ul v-if="messages">
           <LogMessageComponent v-for="(message, index) in messages" :key="index" :message="message" :viewModel="viewModel" @click="messageClicked(message)" @spaceClicked="spaceClicked"/>
         </ul>
       </div>
       <button
+        v-show="showScrollToBottomButton"
         type="button"
         class="log-latest-button"
         aria-label="Latest logs"
@@ -68,6 +69,7 @@ type LogPanelModel = {
   messages: Array<LogMessage>,
   selectedGeneration: number,
   selectedMessage: LogMessage | undefined,
+  showScrollToBottomButton: boolean,
 };
 
 export default defineComponent({
@@ -92,6 +94,7 @@ export default defineComponent({
       messages: [],
       selectedGeneration: -1,
       selectedMessage: undefined,
+      showScrollToBottomButton: false,
     };
   },
   components: {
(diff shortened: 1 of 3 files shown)


kberg:
Do you understand how and why it works?

rusliksu:
Yes. The log panel is recreated on every game update, so it saves the current position before that and restores it after. This is useful when reading older turns while other players keep playing, because new logs don’t push you back to the bottom

kberg:
> Yes. The log panel is recreated on every game update, so it saves the current position before that and restores it after. This is useful when reading older turns while other players keep playing, because new logs don’t push you back to the bottom

Oh, I get it. OK, thank you.

src/client/components/logpanel/LogPanel.vue
```diff
@@ -41,6 +53,15 @@ import {isMarsSpace} from '@/common/boards/spaces';
 
 let logAbortController: AbortController | undefined;
 
+type LogPanelViewState = {
+  participantId: ParticipantId,
```
kberg: Why is participantId necessary?

src/client/components/logpanel/LogPanel.vue
```diff
@@ -41,6 +53,15 @@ import {isMarsSpace} from '@/common/boards/spaces';
...
+  selectedGeneration: number,
+  scrollTop: number,
```
kberg: I don't know what this attribute or the next one do.

src/client/components/logpanel/LogPanel.vue
```diff
@@ -137,8 +162,10 @@ export default defineComponent({
           if (getPreferences().enable_sounds && window.location.search.includes('experimental=1') ) {
             SoundManager.newLog();
           }
-          if (generation === this.generation) {
+          if (forceScrollToEnd || restoredState?.stickToBottom === true || (restoredState === undefined && generation === this.generation)) {
```
kberg: This suddenly got alot more complicated. Help me understand what it is doing.

src/client/components/logpanel/LogPanel.vue
```diff
@@ -150,11 +177,28 @@ export default defineComponent({
...
         scrollablePanel.scrollTop = scrollablePanel.scrollHeight;
       }
     },
+    restoreScrollTop(scrollTop: number) {
+      const scrollablePanel = this.getScrollablePanel();
+      if (scrollablePanel !== null) {
+        scrollablePanel.scrollTop = scrollTop;
+      }
+    },
+    getScrollablePanel(): HTMLElement | null {
```
kberg: Can this be computed? It should never change.

src/client/components/logpanel/LogPanel.vue
```diff
@@ -150,11 +177,28 @@ export default defineComponent({
...
+      return document.getElementById('logpanel-scrollable');
+    },
+    isNearBottom(): boolean {
+      const scrollablePanel = this.getScrollablePanel();
+      if (scrollablePanel === null) {
+        return true;
+      }
+      const remaining = scrollablePanel.scrollHeight - scrollablePanel.clientHeight - scrollablePanel.scrollTop;
+      return remaining <= 24;
```
kberg: What does 24 mean?

src/client/components/logpanel/LogPanel.vue
kberg, line 1: This isn't exactly your problem, but this component seems to now be more complicated, such that it's hard to tell all of this apart. I feel like there may be some way to split this into two components, or classes or something: one that manages rendering the panel, including these special visual behaviors, and one for game-specific stuff.

I'm not asking you to do that.

rusliksu:
updated this, ready for another look

src/client/components/logpanel/LogPanel.vue
rusliksu: you are right, it isnt needed. The component only remounts for the same participant and opening another player reloads the page. I'll remove it

src/client/components/logpanel/LogPanel.vue
kberg: I'm little concerned about the complexity in here. Even though it's just a couple hundred lines. So, this won't go in on its own. Hold off, I'm going to think about this some.

kberg:
PR #8393 is going to meaningfully change this PR. I'll come back to you with more thoughts in a bit.

Note also that there's already a piece of UI similar to the scroll button in CardList.vue

src/client/components/logpanel/LogPanel.vue
```diff
@@ -104,11 +127,15 @@ export default defineComponent({
     },
     selectGeneration(gen: number): void {
       if (gen !== this.selectedGeneration) {
-        this.getLogsForGeneration(gen);
+        this.getLogsForGeneration(gen, gen === this.generation ? 'bottom' : undefined);
       }
       this.selectedGeneration = gen;
     },
-    getLogsForGeneration(generation: number): void {
+    showLatestLogs(): void {
```
kberg, line 141: Rename to `scrolltobottom`

src/client/components/logpanel/LogPanel.vue
```diff
@@ -104,11 +127,15 @@ export default defineComponent({
...
+      this.selectedGeneration = this.generation;
+      this.getLogsForGeneration(this.generation, 'bottom');
```
kberg, line 143: Why does showLatestLogs have to fetch logs from the web? Maybe this is one of those things I'm missing. Why doesn't it just scroll to the bottom?

kberg (commented):
Image

I am in gen 3. However, I was never brought to gen 3. You can see that there is no reason that I would have scrolled.

kberg:
image

1. Please don't show the button when it is already at the bottom.
2. Please don't hide the scroll bar.

kberg:
OK, if you can fix those and address my comments I will take another look. This is starting to make sense.

kberg:
This is now going to get more complicated.

Are you familiar with the "suspend" feature? It's not very easy to discover, but it suspends fetching the log from the server.

image

and provides a very similar behavior. We don't need both of these.

kberg:
I'll have to think about this some more.

src/client/components/logpanel/LogPanel.vue
rusliksu: when older generation selected, messages have only this generation, so it fetch current logs first and then scroll to bottom

src/client/components/logpanel/LogPanel.vue
rusliksu: didnt rename it yet because it also fetch current generation, will wait for decision about suspend and #8393

rusliksu:
fixed both visual issues, button is hidden at bottom and scrollbar stays visible, didnt change suspend behavior yet while you think about it, preview https://preview.tm.knightbyte.win/spectator?id=s64edfe038a06

kberg:
> fixed both visual issues, button is hidden at bottom and scrollbar stays visible, didnt change suspend behavior yet while you think about it, preview https://preview.tm.knightbyte.win/spectator?id=s64edfe038a06

Thanks for working with me.

Thinking about this some more, I like your UI experience much better. But I also think the whole thing needs a little bit of a cleanup. For instance, a proper log view would be smoother, and show history in a streaming way. Don't do anything. I think in the end, I'm going to adopt your change (if the functional issue I noticed is resolved, or perhaps I was mistaken.) And then I can deal with breaking pieces apart etc etc.

src/client/components/logpanel/LogPanel.vue
kberg: Makes sense

src/client/components/logpanel/LogPanel.vue
kberg: k. I really hope this becomes clearer with a rewrite.

src/client/components/logpanel/LogPanel.vue
kberg: I'm adding comments describing them.

kberg:
> Image
> I am in gen 3. However, I was never brought to gen 3. You can see that there is no reason that I would have scrolled.

If this can be fixed / addressed, this can be merged in.

kberg (changes requested):

kberg:
Never mind, I see what's going on now.

kberg (approved):
This can go in. Thanks!

kberg:
There are still a couple of bugs but I'd rather do a wholesale update and deal with it then. Thank you.

PR #8367, opened 2026-08-09
Add removeResourcesFromAnAnyCard to behavior.

kberg:


src/server/behavior/Executor.ts
```diff
@@ -423,6 +425,22 @@ export class Executor implements BehaviorExecutor {
       }
     }
 
+    if (behavior.removeResourcesFromAnyCard !== undefined) {
+      const r = behavior.removeResourcesFromAnyCard;
+      const remainder = {...behavior};
+      delete remainder['removeResourcesFromAnyCard'];
```
kberg, line 431: I don't know if this is a good plan.

kberg: Yep, it's correct.

PR #8368, opened 2026-08-10
Add agent-readable card database export (base, corpera, prelude)

ryoku:
## Summary
- Adds `npm run make:card-db`, which walks the existing card manifests at runtime (same technique as `export_card_rendering.ts`) to produce a normalized, agent-readable JSON database at `data/cards/cards.json` (+ `index.json` lookup table + `README.md`), covering the `base`, `corpera` and `prelude` modules.
- Pure normalizer functions (`normalizeRequirements.ts`, `normalizeEffect.ts`) translate the game's `CardRequirementDescriptor` / `Behavior` DSLs into a flat schema (`CardDatabaseTypes.ts`), throwing rather than silently dropping data for any mechanic outside the three covered modules.
- Cards whose behavior lives in TypeScript rather than declarative data are detected mechanically via a prototype-chain walk (`hasBespokeLogic.ts`) and their semantics are supplied by a hand-authored `overlay.ts`, keyed by `CardName`. A test (`buildCardDatabase.spec.ts`) fails the build if any mechanically-detected card lacks an overlay entry, or if an overlay entry is no longer needed — the completeness check is exhaustive and currently green.
- Adds two small getters to `Card.ts` (`actionBehavior`, `globalParameterRequirementBonus`) so the exporter can read data otherwise only reachable via the protected `properties` field.
- No new dependencies.

## Test plan
- [x] `npm run test:server` — 7304 passing
- [x] `npm run lint:server` — clean
- [x] `npx tsc --noEmit -p src/tsconfig.json` — clean (pre-existing unrelated genfiles errors excluded)
- [x] `npm run make:card-db` regenerates `data/cards/*` byte-identically on repeated runs (verified via diff)
- [x] `git status --short data/cards` clean after regenerating, confirming the committed output matches the generator

### DIFF
--- a/data/cards/README.md
+++ b/data/cards/README.md
@@ -1,0 +1,72 @@
+# Terraforming Mars card database
+
+Generated from the game source by `npm run make:card-db`. Do not edit by hand —
+regenerate instead. The schema of record is
+`src/server/tools/cardDatabase/CardDatabaseTypes.ts`.
+
+## Files
+
+- `cards.json` — an array of every card in the `base`, `corpera` (Corporate Era)
+  and `prelude` sets. Sorted by set, then by `id`. This is the whole database;
+  read it in one go.
+- `index.json` — the same cards reduced to `id`, `name`, `set`, `kind`, `type`,
+  `cost`, `tags`, `vp` and `bespoke`. Use it to find a card, then look the full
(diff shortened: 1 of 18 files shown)


kberg:
What did I miss? :D

PR #8396, opened 2026-08-23
Log selected behavior resource spends in Executor

rusliksu:
## Summary

- add an opt-in `spend.log` flag to the declarative behavior DSL and emit the selected direct-resource spend from `Executor`
- enable that flag for Electro Catapult's plant and steel alternatives without overriding the card action
- preserve `player.pay` for steel/titanium and leave every unannotated `behavior.spend` unchanged

## Demo

Before, a complete Electro Catapult action only exposed the gain:

```text
blue gained 7 M€
```

On this commit, the full action produces:

```text
plant: blue spent 1 plant | blue gained 7 M€
steel: blue spent 1 steel | blue gained 7 M€
```

## Testing

- focused Executor + Electro Catapult tests: 86 passing
- full server suite: 7261 passing
- `npm run build:tests`
- `npm run lint:server`
- `npm run lint:client`
- `npm run build`
- upstream PR gate passed for `61826f87`

## Notes

This replaces the card-specific implementation from closed PR #8395 with the Executor-owned approach requested by the maintainer. Gameplay and resource selection are unchanged.

### DIFF
--- a/src/server/behavior/Behavior.ts
+++ b/src/server/behavior/Behavior.ts
@@ -47,6 +47,8 @@ export type Behavior = {
   spend?: Partial<OneOfType<Spend>> & {
     canUseSteel?: boolean,
     canUseTitanium?: boolean,
+    /** Log a direct steel, titanium, plant, or energy spend. */
+    log?: true,
   };
 
   /**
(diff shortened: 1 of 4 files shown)


kberg (commented):
This doesn't solve the problem of the message that is lost. I'll take care of this, thanks.

PR #8397, opened 2026-08-24
Feature/mobile new game screen

quintavallechristian:
Improved mobile interface for CreateGameForm.
From: 
image

To:
image

And GameHome screen

From:
image

To:
image

- I didn't switch to <script setup> because, expecially for CreateGameForm, because it would have changed a lot of lines, not related to responsiveness
- I added a sticky behavior for the main buttons, placed at the bottom of the screen
- i increased the z-index of the preferencesIcon to make it visible on top of the other elements

### DIFF
--- a/src/client/components/create/CreateGameForm.vue
+++ b/src/client/components/create/CreateGameForm.vue
@@ -613,6 +613,7 @@ type Refs = {
 type FormModel = {
   preludeToggled: boolean;
   uploading: boolean;
+  previousViewport: string;
 };
 
 export default defineComponent({
@@ -622,6 +623,7 @@ export default defineComponent({
       ...defaultCreateGameModel(),
       preludeToggled: false,
       uploading: false,
+      previousViewport: '',
     };
   },
   components: {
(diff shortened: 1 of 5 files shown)


src/styles/preferences.less
```diff
@@ -180,6 +180,7 @@
   position: fixed;
   bottom: 4px;
   height: 46px;
+  z-index: 110;
```
kberg, line 183: Curious what this is doing.

src/client/components/create/CreateGameForm.vue
```diff
@@ -680,6 +682,24 @@ export default defineComponent({
   mounted() {
     setDocumentTitle('Create New Game');
     this.restoreLastSettings();
+
+    // Set the viewport width to width=device-width on the create game form so mobile browsers use their actual CSS viewport width.
+    // The current global viewport is width=1260, which prevents the create game form from using the device width on phones.
+    // This is a temporary solution in order to make this edit scoped to the create game form.
+    // TODO: Once responsiveness covers the whole project, this code should be removed and the tag in index.html should be updated directly.
```
kberg, line 689: How many more times is this going to be applied? Should it just become a function in a file on its own?

kberg (approved):
Looks good. Just let me know your thoughts on my two questions.

src/styles/preferences.less
quintavallechristian: it's the point I highlighted in the description:
"i increased the z-index of the preferencesIcon to make it visible on top of the other elements"

image
It places the gear on top of the buttons (and the modal as well). Otherwise, in some screens, you cannot press the gear because buttons are on top.

src/client/components/create/CreateGameForm.vue
quintavallechristian: I think the next step will be the responsiveness of the main game page. When that is done, we can get rid of this procedure and fulfill the TODO.

src/styles/preferences.less
kberg: Apologies: I've been a bit disillusioned by descriptions lately (assuming AI generation), and will read yours more carefully.

kberg (approved):
Thanks. This is a good change. I will push this and your prior change shortly.

PR #8435, opened 2026-09-15
Add hover tooltips and expandable descriptions for Turmoil agendas

JessyIsCute:
TurmoilAgenda.vue gains a hover tooltip (teleported to <body> so it isn't clipped by the party accordion's overflow:hidden/scale wrapper) showing the bonus/policy's plain-text effect. TurmoilAgendaContainer.vue (used on the /cards reference page) gets a click-to-expand description below each card using the same text.


src/common/turmoil/AgendaDescriptions.ts
```diff
@@ -0,0 +1,56 @@
...
+ * (hover tooltips on the Turmoil board, and the /cards reference page). The authoritative
+ * logic lives in the matching `src/server/turmoil/parties/*.ts` Bonus/Policy classes --
+ * these strings are copied from their `description` fields, since the client can't import
+ * server-only game logic.
+ *
+ * Kelvinists' kp01 description is normally dynamic (cheaper with High-Temp
+ * Superconductors in play); this shows its base cost, matching the server's own fallback
+ * when no player context is available.
+ */
+export const AGENDA_DESCRIPTIONS: Partial<Record<BonusId | PolicyId, string>> = {
```
kberg: These are already defined in the policies and bonuses. It will be better to get them in one place. It wouldn't be so bad if this was the place, but there's already a mechanism for strings shared between server and client. 🤔

src/common/turmoil/AgendaDescriptions.ts
```diff
@@ -0,0 +1,56 @@
+import {BonusId, PolicyId} from './Types';
+
+/**
+ * Plain-text descriptions for every bonus/policy id, for client-side reference display
+ * (hover tooltips on the Turmoil board, and the /cards reference page). The authoritative
+ * logic lives in the matching `src/server/turmoil/parties/*.ts` Bonus/Policy classes --
+ * these strings are copied from their `description` fields, since the client can't import
+ * server-only game logic.
+ *
+ * Kelvinists' kp01 description is normally dynamic (cheaper with High-Temp
```
kberg: Unnecessary comment

src/styles/turmoil.less
```diff
@@ -691,3 +691,24 @@ input[type="radio"]:checked + .party-container::after {
   background-size: 50px 46px;
   width: 50px;
 }
+
+// This is teleported to <body> and positioned in TurmoilAgenda.vue's onMouseEnter (viewport
```
kberg: I don't understand this. What is it saying?

src/client/components/cardlist/TurmoilAgendaContainer.vue
```diff
@@ -1,16 +1,18 @@
 <template>
-  <div class="container">
+  <div class="container" @click="showDescription = !showDescription">
   <TurmoilAgenda :id="agendaId" />
   <div class="line1">{{ agendaId }}</div>
   <div class="line2">
     {{ $t(agenda.name) }} {{ $t(agenda.type) }} {{ agenda.num }}
   </div>
+  <div class="description" v-if="showDescription">{{ $t(description) }}</div>
```
kberg: Use v-i18n instead of $t

src/client/components/cardlist/TurmoilAgendaContainer.vue
```diff
@@ -42,7 +47,10 @@ const agenda = computed<AgendaInfo>(() => agendaInfoById(props.agendaId));
     margin-left: 80px !important;
   }
 
-  > div:first-child {
+  // TurmoilAgenda's template root is no longer a single element (it now has a sibling
```
kberg: "is no longer?" Why is it being described this way? I don't really understand this comment. What is it saying?

kberg (commented):
Thanks for the PR. I have a couple of questions about CSS which I don't understand very well.

src/common/turmoil/AgendaDescriptions.ts
JessyIsCute: Comment Removed.

src/client/components/cardlist/TurmoilAgendaContainer.vue
JessyIsCute: Updated to use v-i18n.

src/styles/turmoil.less
JessyIsCute: Basically the turmoil board's container clips anything that pokes outside its edges. Teleport moves the tooltip's actual HTML out of that container. so it's going ontop of everything. I didn't want to mess with the clipping cause it seems like it's neeeded for expand/collapsing animation?

src/common/turmoil/AgendaDescriptions.ts
JessyIsCute: Solved.

src/common/turmoil/IClientAgenda.ts
```diff
@@ -0,0 +1,6 @@
+import {BonusId, PolicyId} from './Types';
+
+export interface IClientAgenda {
```
kberg: You did the right thing by following the pattern, but in this project interfaces are becoming things which provide an abstraction for classes. Can you use `type`? I could do this as a full change later, but while you're here...

src/common/turmoil/IClientAgenda.ts
```diff
@@ -0,0 +1,6 @@
...
+  id: BonusId | PolicyId;
```
kberg: This field doesn't look necessary. Is it? It's OK if the type has only one field in it. I could see doing something nice like adding rendering instructions here later.

src/client/turmoil/ClientAgendaManifest.ts
```diff
@@ -0,0 +1,13 @@
...
+// @ts-ignore agendas.json doesn't exist during npm run build
+import agendaJson from '@/genfiles/agendas.json';
+
+const agendas: Map<BonusId | PolicyId, IClientAgenda> = new Map();
+(agendaJson as any as Array<IClientAgenda>).forEach((agenda) => {
+  agendas.set(agenda.id, agenda);
+});
+
+export function getAgendaDescription(id: BonusId | PolicyId): string {
+  return agendas.get(id)?.description ?? `Unknown agenda ${id}`;
```
kberg: Match the other APIs by returning the ~I~`ClientAgenda` and throwing if it does not exist. Optional: can name `getAgendaOrThrow`.

src/common/turmoil/AgendaDescriptions.ts
kberg: OK. I have some comments, but yeah that's good.

src/styles/turmoil.less
kberg: I'm new to Teleport, so, cool. Thanks.

kberg (approved):
Some minor tweaks but this is a good change.

src/client/components/cardlist/TurmoilAgendaContainer.vue
JessyIsCute: added a extra div which makes this fix unnessairly and its now much simpler.

src/common/turmoil/IClientAgenda.ts
```diff
@@ -1,6 +1,3 @@
-import {BonusId, PolicyId} from './Types';
-
-export interface IClientAgenda {
-  id: BonusId | PolicyId;
+export type IClientAgenda = {
```
kberg: Rename the type to ClientAgenda. Rename the file to ClientAgenda.

PR #8438, opened 2026-09-16
Make Neptunian Power Consultants compatible with Flooding

### LINKED ISSUE #8436: When Flooding triggers Neptunian Power Consultants
Ideally the Flooding player should get the -4 MC option first before the NPC player gets to make their choice.

(Officially, of course, the active player gets to make all ordering decisions, including that they could opt to wait to do the -4 until after, but if you don't want to make it a choice it probably is more often what they want to do the -4 first. I think.)

kberg:
Fixes #8436


kevinb9n:
nice, thanks!

PR #8444, opened 2026-09-18
/Help Modular and Policital Party listing

JessyIsCute:
image


src/client/components/help/HelpTurmoilParties.vue
```diff
@@ -0,0 +1,77 @@
+<template>
+    <div class="help-turmoil-parties-container">
+      <h2 v-i18n>Political Parties</h2>
+      <p v-i18n>The ruling party's bonus is granted to every player at the end of each generation (scaled by how well each player matches it); its policy applies for the whole generation it rules.</p>
```
kberg: Policies don't apply for the whole generation.  Where did you get that?

Also this reads like AI wrote it.

src/client/components/help/HelpTurmoilParties.vue
```diff
@@ -0,0 +1,77 @@
+<template>
```
kberg, line 1: Please use the <script setup> style for new components.

kberg (commented):
I'll have to run this locally to see what's happening. Just these for now

kberg:
image

This is not going to look good when compressed. How is it going to look on a mobile phone? Any thoughts?

src/client/components/help/HelpTurmoilParties.vue
JessyIsCute: Policies do apply for the whole generations? I thought, Policies generation wide, Bonuses start of generation?

kberg: > Policies do apply for the whole generations? I thought, Policies generation wide, Bonuses start of generation?

image

image

Hope this helps

JessyIsCute: I dont understand, if green become active you get 3 mc for every greenery you palce during the whole generation? what am i missing?

kberg: What you're missing is that it's only during the **action phase**, which is not the whole generation. It doesn't happen during final greenery, it doesn't happen during prelude phase.

That surprised me too, by the way.

kberg (commented):
Just update the description of policies and bonuses (if necessary? I forget) and we're good to go.

